PR #85, opened 2023-12-31
TTS & SentenceTokenizer impl: OpenAI, ElevenLabs & StreamAdapter

theomonnom:
## Summary
The `SimulationDispatch` was passed in the agent's dispatch metadata (`RoomAgentDispatch.Metadata` → `job.metadata`). That surface belongs to users: an agent reading its own dispatch metadata for config would see our simulation payload under simulation and behave differently than in production.

The dispatch now rides in the simulator participant's **`lk.simulation.dispatch` attribute** (LiveKit-reserved `lk.` namespace, next to the existing `lk.simulator` marker); the agent dispatch carries only the agent name.

Counterpart SDK change: livekit/agents#6053 (reads the attribute first, falls back to job/room metadata for older servers).

**Rollout note:** merge and release livekit/agents#6053 first — once this deploys, agents on older livekit-agents versions won't resolve a `SimulationContext` (simulations still run and get judged; `on_simulation_end` / `scenario.userdata` won't reach the agent).


davidzhao (commented):
Nice work! A few comments.

PR #137, opened 2024-01-26
Add LMNT TTS plugin

zachoverflow:
I have added support for both the low-latency streaming and non-streaming versions of the LMNT API, following the existing TTS structure in adjacent plugins.

One note:
* I have included a dependency on torchaudio. While it is a bit heavyweight, I noticed that livekit-plugins-openai also takes a dependency on it so I have done the same. (It makes implementation slightly easier.)

While I'm here, I also am fixing a minor README issue to help future folks to run the kitt demo:
* The click configuration inside run_app attaches the LiveKit args to cli() instead of the subcommand start(). This means click expects us to put the subcommand 'start' after the LiveKit args instead of before them.

### DIFF
--- a/livekit-plugins/livekit-plugins-lmnt/livekit/plugins/lmnt/tts.py
+++ b/livekit-plugins/livekit-plugins-lmnt/livekit/plugins/lmnt/tts.py
@@ -1,0 +1,285 @@
+# Copyright 2023 LiveKit, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import aiohttp
+import asyncio
+import base64
+import contextlib
+import io
+import json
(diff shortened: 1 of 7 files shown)


livekit-plugins/livekit-plugins-lmnt/livekit/plugins/lmnt/tts.py
```diff
@@ -0,0 +1,285 @@
...
+        )
+    return voices
+
+
+def _float_to_int16(array):
+    return np.clip(np.floor(array * 32767.5), -32768, 32767).astype(np.int16)
+
+
+def decode_mp3_to_frame(audio: bytes) -> rtc.AudioFrame:
+    tensor, sample_rate = torchaudio.load(io.BytesIO(audio), format="mp3")
```
keepingitneil, line 277: Do/Can we know ahead of time what the sample_rate will be?

When publishing an audio source in an agent, the sample rate (and num channels) must be known: https://github.com/livekit/agents/blob/main/examples/kitt/kitt.py#L76

It would be nice to publish at the same sample rate that lmnt is producing to avoid having to do a resample. This is especially impactful because, currently, LiveKit's resampler doesn't maintain continuity between subsequent audio frames so audio quality and artifacts can be introduced.

keepingitneil: Spoke about this internally - we may address this at the library level so we can leave this as-is

keepingitneil (approved):

zachoverflow:
Went ahead and fixed the sample rate when creating the audio frame to 24kHz as that's what we're always sending over, so no need to work around internally for us. Thanks!

PR #141, opened 2024-01-29
rework how parameters are set and checked by agent Worker

dsa:


livekit-agents/livekit/agents/worker.py
```diff
@@ -390,11 +391,23 @@ def run_app(worker: Worker) -> None:
     )
     @click.option(
         "--url",
-        help="The websocket URL",
-        default=worker._rtc_url,
+        required=True,
```
keepingitneil, line 394: required=True means it would fail for users who want to use `dotenv`

dsa: No, it doesn't, because LIVEKIT_URL is read from dotenv before the agent start method is called.

keepingitneil: ahhh ok yeah!

keepingitneil (approved):

PR #144, opened 2024-02-01
Add missing requirements + version constraints

keepingitneil:
sorry :)


keepingitneil:
Bypassing branch protections because it's low stakes and might need more iteration as I deploy the Dockerfile

PR #145, opened 2024-02-01
Update deepgram endpointing 

seanmuirhead:
Notes: 
- Added `speech_final` to `SpeechEvent`. Defaulted to `False` because looks like it's being used elsewhere. 
- Return `speech_final` for each `SpeechEvent`. 
- Add `utterance_end_ms` for more control over VAD

Relevant Docs:
https://developers.deepgram.com/docs/understanding-end-of-speech-detection 
https://developers.deepgram.com/docs/understand-endpointing-interim-results

### DIFF
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -18,7 +18,7 @@ class SpeechData:
 class SpeechEvent:
     is_final: bool
     alternatives: List[SpeechData]
-    speech_final: bool = False
+    end_of_speech: bool = False
 
 
 class STT(ABC):
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -17,6 +17,7 @@ class SpeechData:
 @dataclass
 class SpeechEvent:
     is_final: bool
+    speech_final: bool
     alternatives: List[SpeechData]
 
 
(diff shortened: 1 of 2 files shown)


theomonnom (approved):
Awesome, thanks for contributing!

Vibrat:
Hi @seanmuirhead,

I come from a similar PR #148. 

We are sharing some changes on the codebase. I think the end result of both `endpointing` and `interim` is to have `speech_final` and `is_final` as the final output. Would be nice if we can rebase my change onto your PR so that I can close my ticket and we continue the discussion here

davidzhao:
what do you think about exposing `end_of_speech` as the field instead of `speech_final`? It seems a bit more generic to signal the end of input.. and also trying to avoid having both speech_final and is_final.

seanmuirhead:
@davidzhao Ya I agree, much cleaner that way. Will include in rebase

seanmuirhead:
@davidzhao @Vibrat done! Removed `utterance_end_ms` according to Deepgram docs that were updated ~6hrs ago. Two notes for comment: 

1. @Vibrat 's `min_silence_duration` sound good for the attribute name? 
2. Deepgram has `endpointing` cast as a string, which is why we've done it here, but int works just as fine for the record.

davidzhao (approved):
this looks good! we'll get it merged and update the other plugins to behave the same way.

PR #150, opened 2024-02-04
properly closes connection with elevenlabs when completing stream

CalinR:
Currently the stream isn't closed when elevenlabs finishes getting data. This is due to not breaking out of the parent `while True` loop on line 182 of `tts.py`. Since we only break out of the inner `while not ws.closed` loop we never get to the bottom of _run which sets `self._closed` to true. Because of this, we never get a `StopAsyncIteration` in `__anext__`. This prevents the kitt example from running an aclose on the tts stream.

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -184,7 +184,9 @@ class SynthesizeStream(tts.SynthesizeStream):
         retry_count = 0
         listen_task: Optional[asyncio.Task] = None
         ws: Optional[aiohttp.ClientWebSocketResponse] = None
-        while True:
+        completed = False
+
+        while not completed:
             try:
                 ws = await self._try_connect()
                 retry_count = 0  # reset retry count
@@ -212,6 +214,7 @@ class SynthesizeStream(tts.SynthesizeStream):
                         self._event_queue.put_nowait(
                             tts.SynthesisEvent(type=tts.SynthesisEventType.FINISHED)
                         )
+                        completed = True
                         break
 
             except asyncio.CancelledError:


TasseDeCafe:
Here I thought I was doing something wrong, thank you.

CalinR:
I need to make a modification to this. This solution will break the retry functionality

CalinR:
I've updated this commit to handle retries

keepingitneil:
Hrmm so this plugin is designed to be able to keep using the same stream for the lifetime of the application (i.e. not completing it every time we flush to get a response from elevenlabs).

If I'm reading things correctly, calling `aclose` should "cancel" the main task which would send it into the `asyncio.CancelledError` exception flow which should break out of the top level while loop and set closed to true. Perhaps there is a bug in that flow?

CalinR:
Interesting I have definitely misunderstood how this plugin should function and that makes more sense. Currently what I'm seeing is a flood of "didn't receive anything in the last 20 seconds" errors from elevenlabs. And I've seen this error when I've been constantly sending text to it so I'm curious now as to why that error is happening. I'll have to investigate further

keepingitneil:
Sounds good @CalinR -  yeah I'm addressing another bug here: https://github.com/livekit/agents/pull/152 which may be related.

CalinR:
Sounds good. @keepingitneil if I'm reading the code correctly it also does look like in the Kitt example we are doing a new tts_plugin.stream() every time we get a new event that is final from the STT which means we're opening a new connection with elevenlabs. This would explain why we're getting disconnected errors since the last connection was never closed and was allowed to timed out.

So I think if the intention was for the connection to live for the life of the application we would need to move where tts_plugin.stream() is called.

keepingitneil:
Yeah so the plugin is designed to work with both strategies. So in the case where `aclose` is used - I believe the CancelledError flow should get triggered. Are you seeing issues with that flow?

CalinR:
I am in the Kitt example. This is because aclose is called after the stream loop has completed in send_audio_stream but since the iterator never returns that it has completed it never gets to that code block.

Calling aclose from inside the stream loop `async for e in tts_stream` if we received a finished event does work as expected though.

keepingitneil:
Closing because we've changed the APIs a bit since this PR. Among those changes are fixes to the connection handling issues. Sorry for the delay here and thanks for surfacing this issue 🙏

PR #152, opened 2024-02-05
Retry text that failed to send to elevenlabs + fix skipped task_done() call

keepingitneil:
Elevenlabs closes websockets when you don't send text fast enough. We would currently drop that text and short circuit out of the `task_done()` call.


davidzhao (approved):
lg

PR #153, opened 2024-02-06
Update KITT and other plugins to use end_of_speech field

davidzhao:
Tested with KITT. It significantly improves the end of speech behavior so that we are giving it a 1s wait before starting to process user input.


theomonnom (approved):
lgtm!

PR #156, opened 2024-02-07
Rewrite Deepgram to use WebSocket API

keepingitneil:


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -40,12 +47,17 @@ def __init__(
...
         super().__init__(streaming_supported=True)
-        api_key = api_key or os.environ.get("DEEPGRAM_API_KEY")
-        if not api_key:
+        self._api_key = api_key or os.environ.get("DEEPGRAM_API_KEY")
+        if not self._api_key:
             raise ValueError("Deepgram API key is required")
 
-        dg_opts = deepgram.DeepgramClientOptions(api_key=api_key, url=api_url or "")
-        self._client = deepgram.DeepgramClient(config=dg_opts)
+        self._session = aiohttp.ClientSession(
```
theomonnom, line 54: Session is leaked here. I think this should be moved into the Stream ctor

keepingitneil: changed to use the more canonical `async with aiohttp.ClientSession()`. No downsides to this in the stream case but added latency in the recognize case.

PR #171, opened 2024-02-17
fixes audio and video frame events in examples

CalinR:
There was changes made to `rtc.AudioStream` and `rtc.VideoStream` to return an FrameEvent object instead of the Frame in this pull request https://github.com/livekit/python-sdks/pull/155. This causes issues in the livekit examples as they are still expecting a Frame object and then throws an error that the attribute `remix_and_resample` doesn't exist on the object.

I've updated all examples impacted by this change.


davidzhao:
Thanks for the fix here @CalinR! This is due to our recent update to the Python SDK that have not fully migrated over yet.

davidzhao (approved):

PR #175, opened 2024-02-21
Fix openai tts

keepingitneil:
torchaudio gives normalized float32


theomonnom:
should we remove torchaudio instead?
Also supporting chunked output?

PR #176, opened 2024-02-21
enh: Adding google TTS

brightsparc:
Add support for [google text to speech](https://cloud.google.com/text-to-speech/docs/libraries) synthesize


brightsparc:
Seems tests failed due to missing ENV vars.

brightsparc:
Can we please get this landed now that agent refactor is complete?

davidzhao:
Hey @brightsparc sorry for the delays.. we'll take a look at this PR in the next day or so now that the major refactor is done.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -0,0 +1,123 @@
...
+        language: LgType = "en-US",
+        gender: GenderType = "neutral",
+        voice_name: str = "",  # Not required
+        audio_encoding: AudioEncodingType = "wav",
+        sample_rate: int = 24000,
+        speaking_rate: float = 1.0,
+        credentials_info: Optional[dict] = None,
+        credentials_file: Optional[str] = None,
+    ) -> None:
+        super().__init__(streaming_supported=False)
```
keepingitneil, line 52: can you include sample_rate and num_channels to this?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -0,0 +1,123 @@
...
+    ) -> tts.SynthesizedAudio:
+        # Perform the text-to-speech request on the text input with the selected
+        # voice parameters and audio file type
+        response: SynthesizeSpeechResponse = await self._client.synthesize_speech(
+            input=texttospeech.SynthesisInput(text=text),
+            voice=self._config.voice,
+            audio_config=self._config.audio_config,
+        )
+
+        data = response.audio_content
```
keepingitneil, line 114: This only works for linear16 right? Other formats would need to be decoded

brightsparc: If the data needs to be in linear16, yes we would probably need a conversion from other formats.  Do you have any suggestions on libraries to use for this, there is [opuspy](https://github.com/elevenlabs/opuspy) that wraps the opus cpp library for example.

Perhaps in the short term we could restrict to only linear16 format?

brightsparc: @keepingitneil i've update the google TTS to add support for mp3 codec, and let the others commented out until we want to add support for them in the agents codecs library.  I've also fixed tests to return async generator.

brightsparc:
@keepingitneil how should I be passing Google credentials for these tests. Should there already be something set in the env or do I need to pass a reference to a  json file?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
keepingitneil: This looks good! Thanks for doing this! I'll merge it, probably won't make it into a release for a week or so (we need to fix some broken tests on our end)

keepingitneil:
@brightsparc creds are on our end. No further action needed!

PR #182, opened 2024-02-26
Introduce codecs optional dependency, Mp3StreamDecoder, and Streaming OAI TTS Response

keepingitneil:


livekit-agents/livekit/agents/codecs/mp3.py
```diff
@@ -0,0 +1,82 @@
...
+
+    async def __anext__(self):
+        while True:
+            try:
+                packet = self._output_queue.get_nowait()
+                if packet is None:
+                    raise StopAsyncIteration
+                return packet
+            except queue.Empty:
+                await asyncio.sleep(0.01)
```
theomonnom, line 82: Not fan of waiting inside a loop, how slow is av.CodecContext if we want to use it inside the main thread?

Is there a reason to not use asyncio.to_thread?

theomonnom: No I mean, can use use asyncio.to_thread to do the decode?

keepingitneil: Oh hrmm wdym?

So the input here is intended to be chunks of bytes that don't necessarily correspond to mp3 chunks. So we push them in a queue for a thread to consume them and create mp3 packets. Those mp3 packets then go into another queue (because they must stay in order) for decoding.

Both parsing and decoding are blocking and too slow for something you'd do on the same thread as the asyncio event loop.

theomonnom: So I was thinking about creating a new task, but using to_thread inside it when starting to decode smth

keepingitneil: Ahh I see, made that adjustment. I also got rid of flush and made this class ephemeral. As I was using it, it was the more common use case (otherwise we'd have to introduce "started" and "finished" events)

davidzhao (approved):
lgtm

PR #185, opened 2024-02-29
Fix bug in nltk plugin

keepingitneil:
Right now iteration ends immediately

### DIFF
--- a/livekit-plugins/livekit-plugins-nltk/livekit/plugins/nltk/sentence_tokenizer.py
+++ b/livekit-plugins/livekit-plugins-nltk/livekit/plugins/nltk/sentence_tokenizer.py
@@ -1,9 +1,11 @@
-from typing import List
-from livekit import agents
-from dataclasses import dataclass
-import dataclasses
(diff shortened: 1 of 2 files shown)


theomonnom:
Just thinking out loud, we need a way to match the input and output sentence after a flush.
Maybe the right thing to do here is to recreate a stream every time?

keepingitneil:
> Just thinking out loud, we need a way to match the input and output sentence after a flush. Maybe the right thing to do here is to recreate a stream every time? Or use events to be consistent with TTS/STT/VAD

yeah...even events mean you have to keep track (by counting). This is true for TTS/STT/VAD as well. I've been messing around with interruptions to get a feel for how our APIs could work with them and creating new streams every time is the easiest way to "ignore" previous inputs/outputs and move on to the next

Maybe we can revisit the pattern but IMO serious use cases will be creating new streams. wdyt?

theomonnom (approved):
lgtm for now, we can discuss more on Slack for the dev branch

PR #191, opened 2024-03-04
Stronger tests + tweaks to mp3 codec

keepingitneil:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/tts.py
```diff
@@ -68,22 +68,8 @@ async def fetch():
...
-                            sample_rate=OPENAI_TTS_SAMPLE_RATE,
-                            num_channels=OPENAI_TTS_CHANNELS,
-                            samples_per_channel=len(data) // 2,
-                        ),
-                    )
-                )
-            results.close()
+        fetch_task = asyncio.create_task(fetch())
+        async for data in decoder:
+            yield tts.SynthesizedAudio(text=text, data=data)
```
theomonnom, line 73: This isn't streamed to the users. This is waiting for the "download" to finish before yielding anything to the user

theomonnom: You could use asyncio.select or __anext__ after a http chunk

theomonnom (changes requested):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/tts.py
keepingitneil: Hrmm so it isn't waiting for the download to complete. If 4096 bytes come in the decoder starts decoding them. It's possible (and probably) for a few http chunks to come in before the first decoded result but I think that's ok.

I do think decoding should happen on another thread (but maybe just one other thread instead of constantly doing `to_thread`?). I think the code-complexity tradeoff is worth it. Also regardless of internal implementation, I think the decoder api should remain push_chunk + async iterator. 

I'm assuming there will be slower decoders (video?) in the future and having a consistent api probably makes sense, especially when the only downside is a few lines of code. I'm also admittedly a bit scared of debugging issues of users running on cheap hardware decoding longer-than-anticipated audio where the symptoms manifest in non-obvious ways.

I know you've been pretty in favor of sync decoding - are there other reasons that I'm not seeing for your preference?

keepingitneil: > EDIT: You're creating a new task, I think this should be avoided wdyt?

Hrmm, so I don't have a strong opinion on task creation and whether or not we should try and avoid it. This is moreso a discussion about what should the mp3.StreamDecoder API look like. And _possibly_ a discussion about what should all decoders look like.

I went with push_chunk + a stream decoder mainly because it felt like it communicated that this decoder is stateful. The other option is `async decode_chunk(<some http chunk>) -> rtc.AudioFrame` I'm not totally against but it "feels" stateless even though it's not.

tests/change-sophie.wav
keepingitneil, line 1: moved to lfs

livekit-agents/livekit/agents/tts/stream_adapter.py
```diff
@@ -34,9 +34,16 @@ async def _run(self) -> None:
...
+                self._event_queue.put_nowait(
+                    SynthesisEvent(type=SynthesisEventType.STARTED)
+                )
                 self._event_queue.put_nowait(
                     SynthesisEvent(type=SynthesisEventType.AUDIO, audio=audio)
                 )
+                self._event_queue.put_nowait(
+                    SynthesisEvent(type=SynthesisEventType.FINISHED)
+                )
+                self._event_queue.put_nowait(None)
```
theomonnom, line 46:
```suggestion

```

livekit-agents/livekit/agents/tts/stream_adapter.py
```diff
@@ -51,15 +58,17 @@ async def flush(self) -> None:
     async def aclose(self) -> None:
         self._main_task.cancel()
         try:
+            await self._event_queue.put(None)
```
theomonnom, line 61:
```suggestion
           self._event_queue.put_nowait(None)
```

livekit-agents/livekit/agents/codecs/mp3.py
```diff
@@ -32,8 +35,8 @@ def __init__(self):
                 "You haven't included the decoder_utils optional dependencies. Please install the decoder_utils extra by running `pip install livekit-agents[decoder_utils]`"
             )
         self._closed = False
-        self._input_queue = asyncio.Queue()
-        self._output_queue = asyncio.Queue()
+        self._input_queue = asyncio.Queue[bytes]()
```
theomonnom, line 38: Let's not type here, bytes and rtc.AudioFrame are wrong since you're also pushing None

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -302,6 +303,36 @@ async def recv_task():
 
         await asyncio.gather(send_task(), recv_task(), keepalive_task())
 
+    def _commit_all_final_events(self) -> None:
```
theomonnom, line 306: What is the motivation for this behaviour, if you're closing VAD and not sending enough data before, if the user is still speaking then the END_OF_SPEECH event not being sent is expected IMO

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -202,6 +202,7 @@ async def _run(self, max_retry: int) -> None:
                         "vad_events": True,
                         "channels": self._num_channels,
                         "endpointing": self._config.endpointing,
+                        "utterance_end_ms": 1000,
```
theomonnom, line 205: Should this be hardcoded?

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -311,6 +342,10 @@ def _process_stream_event(self, data: dict) -> None:
             self._event_queue.put_nowait(start_event)
             return
 
+        if data["type"] == "UtteranceEnd":
+            self._commit_all_final_events()
```
theomonnom, line 346: Can u add a link that mentions the docs?
The event handling is becoming quite complex here and I'm not sure if UterranceEnd sufficient by himself.
e.g we're warning when there is no _final_events, maybe this will happen and this is expected

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
keepingitneil: hrmm probably not. I think it can be Min(1000, endpointing).

> It is recommended that you set the value of utterance_end_ms to be 1000 ms or higher. This is because UtteranceEnd relies on Deepgram's interim_results feature. Deepgram's Interim Results are typically sent every one second, so using a value of less than one second for utterance_end_ms will not offer any benefits.

livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
```diff
@@ -148,12 +149,44 @@ async def _run(self):
                         break
 
                     await asyncio.shield(self._run_inference())
+        except Exception as e:
+            logging.exception(f"silero vad failed: {e}")
         finally:
             self._event_queue.put_nowait(None)
 
-    async def _run_inference(self) -> None:
+    async def _run_inference(self, force_end_of_speech=False) -> None:
```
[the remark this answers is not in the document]
keepingitneil: The issue addressed here is that we use the VAD END_OF_SPEECH event in the stt stream adapter. In the close case END_OF_SPEECH was never getting called so the stream adapter was never running its inference.

I see your point but I can see it the other way too. What would your preferrered solution be? Collect audio frames in the stream_adapter itself?

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -302,15 +304,59 @@ async def recv_task():
...
 
+    def _commit_all_final_events(self) -> None:
+        if not self._speaking:
+            logging.warning(
+                "trying to commit final events without being in the speaking state"
+            )
+            return
+
+        if len(self._final_events) == 0:
+            logging.warning("received end of speech without any final transcription")
```
keepingitneil, line 315: @theomonnom I removed the return in this case so that a START_OF_SPEECH event is matched with an END_OF_SPEECH event even if it's likely a bug if we hit this condition

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
keepingitneil: Made some adjustments, lmk if it makes sense. This logic is definitely confusing - It could be worth writing a custom test just for deepgram because they have two "VAD" systems that work differently. One is the endpointing and it's pure DSP. Background noise will trigger it. the other is utterance end which is more NLP based.

tests/test_stt.py
```diff
@@ -24,29 +67,50 @@ def read_wav_file(filename: str) -> rtc.AudioFrame:
...
+        "transcript": TEST_AUDIO_TRANSCRIPT,
+    },
+    "long": {
+        "audio": read_mp3_file(TEST_AUDIO_FILEPATH_2),
+        "transcript": TEST_AUDIO_TRANSCRIPT_2,
+    },
+}
+
+# Google stt performing poorly
+threshold_overrides = {
```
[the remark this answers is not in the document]
keepingitneil: Removed the short test constraint because it was passing. But even though this is threshold looks agressive, it's not really:

```
=========================== short test summary info ============================
FAILED tests/test_stt.py::test_recognize[google-long] - assert 0.164634[146](https://github.com/livekit/agents/actions/runs/8160618138/job/22307572857#step:6:147)34146342 > 0.5
 +  where 0.16463414634146342 = <bound method SequenceMatcher.ratio of <difflib.SequenceMatcher object at 0x7f8cdf75d4c0>>()
 +    where <bound method SequenceMatcher.ratio of <difflib.SequenceMatcher object at 0x7f8cdf75d4c0>> = <difflib.SequenceMatcher object at 0x7f8cdf75d4c0>.ratio
 +      where <difflib.SequenceMatcher object at 0x7f8cdf75d4c0> = SequenceMatcher(None, "My girlfriend is asleep. So I can't talk loud, but that's probably pretty good for this test. This is a long test for", "My girlfriend is asleep so I can't talk loud, but that's probably pretty good for this test. This is a long test for speech-to-text, it has some pauses in it, it might have some background noise in it, we'll see")
FAILED tests/test_stt.py::test_stream[google-long] - assert 0.14594594594594595 > 0.5
 +  where 0.14594594594594595 = <bound method SequenceMatcher.ratio of <difflib.SequenceMatcher object at 0x7f8cd4f50050>>()
 +    where <bound method SequenceMatcher.ratio of <difflib.SequenceMatcher object at 0x7f8cd4f50050>> = <difflib.SequenceMatcher object at 0x7f8cd4f50050>.ratio
 +      where <difflib.SequenceMatcher object at 0x7f8cd4f50050> = SequenceMatcher(None, " My girlfriend is asleep. So I can't talk loud, but that's probably pretty good for this test. This is a long test for Speech to text it has some pauses in it.", "My girlfriend is asleep so I can't talk loud, but that's probably pretty good for this test. This is a long test for speech-to-text, it has some pauses in it, it might have some background noise in it, we'll see")
============= 2 failed, 14 passed, 4 warnings in 111.71s (0:01:51) =============
```

Our diffing algorithm isn't very strong. This is to me looks like >0.16 match

livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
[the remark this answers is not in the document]
keepingitneil: I removed this and added silence to the end of the other audio file. PR was too big and this is a different existing issue that can be addressed in a future PR

livekit-agents/livekit/agents/codecs/mp3.py
```diff
@@ -53,31 +62,42 @@ async def _run(self):
...
 
     def _decode_input(self, input: bytes):
         packets = self._codec.parse(input)
-        result = b""
+        result_frames: List[rtc.AudioFrame] = []
         for packet in packets:
             try:
                 decoded = self._codec.decode(packet)
                 for frame in decoded:
+                    nchannels = len(frame.layout.channels)
```
theomonnom, line 76: This isn't valid, you're stil using the frames with nchannels > 1
I think making the API sync will makes everything much simpler. Error handling is starting to be hard here no?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -101,43 +101,34 @@ async def list_voices(self) -> List[Voice]:
             data = await resp.json()
             return dict_to_voices_list(data)
 
-    def synthesize(
+    async def synthesize(
```
theomonnom, line 104: This remains unchanged inside the base class
function signatures are not consistent

tests/utils.py
```diff
@@ -0,0 +1,32 @@
+import re
+from collections import defaultdict
+
+
+def compare_word_counts(actual: str, expected: str):
```
theomonnom, line 5: The order isn't compared?

tests/test_stt.py
```diff
@@ -1,12 +1,42 @@
...
+
+
+def read_mp3_file(filename: str) -> List[rtc.AudioFrame]:
+    async def decode():
+        mp3 = agents.codecs.Mp3StreamDecoder()
+        with open(filename, "rb") as file:
+            for chunk in iter(lambda: file.read(4096), b""):
+                mp3.push_chunk(chunk)
+
+        mp3.close()
```
theomonnom, line 32: API feels a little bit weird, what does it means to close?

theomonnom (approved):
ok to merge once these are resolved.
but we still need to think before merging with main the API

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -302,23 +304,67 @@ async def recv_task():
...
+            self._speaking = True
             start_event = stt.SpeechEvent(type=stt.SpeechEventType.START_OF_SPEECH)
             self._event_queue.put_nowait(start_event)
-            return
+        elif data["type"] == "UtteranceEnd":
+            # https://developers.deepgram.com/docs/understanding-end-of-speech-detection#using-utteranceend-and-endpointing
+            # If UtterenceEnd comes before SpeechFinal, we use it to commit the final events
+            # otherwise, we ignore it.
+            if self._speaking:
+                self._commit_all_final_events()
```
theomonnom, line 358: Can UtteranceEnd happens without SpeechStart?

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -302,23 +304,67 @@ async def recv_task():
 
         await asyncio.gather(send_task(), recv_task(), keepalive_task())
 
+    def _commit_all_final_events(self) -> None:
```
theomonnom, line 307:
```suggestion
    def _end_speech(self) -> None:
```

tests/utils.py
keepingitneil: nah, in the future we can do a stronger vector based approach if we need it but I think comparing lexicons is strong enough for now

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
keepingitneil: Shouldn't, unless you receive transcribed words without a speech start. Utterance end is based on transcribed words timing:
https://developers.deepgram.com/docs/understanding-end-of-speech-detection#using-utteranceend

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -302,23 +304,67 @@ async def recv_task():
...
+        )
+        self._event_queue.put_nowait(end_event)
+        self._final_events = []
+
     def _process_stream_event(self, data: dict) -> None:
         assert self._config.language is not None
 
-        # https://developers.deepgram.com/docs/speech-started
         if data["type"] == "SpeechStarted":
+            if self._speaking:
```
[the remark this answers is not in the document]
keepingitneil: I think we can keep until we have more confidence that this logic is correct and aligned with how deepgram works

livekit-agents/livekit/agents/codecs/mp3.py
keepingitneil: fair, i'll just make it all sync in a fast-follow PR

PR #196, opened 2024-03-06
Make mp3 decoder synchronous + Deepgram Tweaks

keepingitneil:
- make the mp3 decoder synchronous
- deepgram APIs are pretty annoying. Everything seems to have its own brain:
  - speechstarted doesn't care about speech_final or utterance end
  - utterance end doesn't care if there were any transcripts with non-empty text.
  - I'm accounting for them in this PR now. This keeps our API such that START is always matched with END even though deepgram doesn't do that.

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -325,13 +325,21 @@ class SpeechStream(stt.SpeechStream):
         if len(self._final_events) > 0:
             confidence /= len(self._final_events)  # avg. of confidence
 
+        # In rare cases, an utterance end event can come
+        # even if there was no transcribed text
+        start_time = 0
+        end_time = 0
+        if len(self._final_events) > 0:
+            start_time = self._final_events[0].alternatives[0].start_time
+            end_time = self._final_events[-1].alternatives[0].end_time
+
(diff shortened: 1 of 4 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -311,8 +311,7 @@ def _end_speech(self) -> None:
             )
             return
 
-        if len(self._final_events) == 0:
```
keepingitneil, line 314: This is actually a normal case. deepgram SpeechStarted (and VAD generally) does not mean it found transcribable audio

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/tts.py
```diff
@@ -45,30 +42,23 @@ def __init__(self, api_key: Optional[str] = None) -> None:
...
+                "input": text,
+                "model": model,
+                "voice": voice,
+                "response_format": "mp3",
+            },
+        ) as resp:
+            async for data in resp.content.iter_chunked(4096):
+                frames = decoder.decode_chunk(data)
+                if len(frames) > 0:
+                    frame = utils.merge_frames(frames)
```
theomonnom, line 63: Merging isn't needed, let's yield each time

theomonnom (approved):

PR #198, opened 2024-03-07
job isolation with sub-processes

theomonnom:


theomonnom:
Still having some deadlock but should be ready for review while I'm working on it

livekit-agents/livekit/agents/ipc/client.py
```diff
@@ -0,0 +1,136 @@
...
+
+    async def send(self, msg: worker.IPCJobMessage) -> None:
+        if self._closed:
+            raise Exception("ipc client is closed")
+
+        await self._send_queue.put(msg)
+
+    async def _run_conn(self) -> None:
+        process_logger.debug("connecting to ipc server")
+        rx, tx = await asyncio.open_connection("127.0.0.1", IPC_PORT)
```
keepingitneil, line 109: I haven't done a lot of research so if it's a bad idea, just ignore me. But can you use shared memory: https://docs.python.org/3.10/library/multiprocessing.shared_memory.html ?

keepingitneil (approved):

livekit-agents/livekit/agents/ipc/job_context.py
```diff
@@ -0,0 +1,104 @@
...
+from livekit import api, rtc
+from livekit.protocol import worker, agent
+
+import asyncio
+import contextlib
+import logging
+from .client import IPCClient
+
+
+class JobContext:
```
davidzhao, line 27: nit: it's a bit strange why the context is located in the ipc package. two reasons:
* the context seems like a piece of code the end user would have to access frequently (I know I've had to reference it a few times when creating an agent)
* while it does communicate with parent process, it feels like that's not the core of this class

davidzhao (approved):
lg!

PR #204, opened 2024-03-27
Static Type Checking For Plugins

keepingitneil:


.github/workflows/check-types.yml
```diff
@@ -0,0 +1,43 @@
+name: Check Types 
+
+on: push
```
davidzhao, line 3: I think we typically run on both pull-request & pushes to main.

livekit-plugins/livekit-plugins-deepgram/setup.py
```diff
@@ -46,13 +46,15 @@
     keywords=["webrtc", "realtime", "audio", "video", "livekit"],
     license="Apache-2.0",
     packages=setuptools.find_namespace_packages(include=["livekit.*"]),
-    python_requires=">=3.10.0",  # deepgram-sdk requires 3.10
+    python_requires=">=3.9.0",  # deepgram-sdk requires 3.10
```
davidzhao, line 49: this isn't matching the comment.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/dalle3.py
```diff
@@ -1,66 +0,0 @@
-# Copyright 2023 LiveKit, Inc.
```
davidzhao, line 1: deleting this would break agent quickstart guide demo. any reasons to remove this?

I guess it'll be fine to put this as an example somewhere too in order to show people how to take the output and creating a VideoFrame out of it

davidzhao (commented):
Nice! just a couple of nits

.github/workflows/check-types.yml
keepingitneil: so this will run on push to any branch. This is how the ruff check is configured and I figured it makes sense for type checking to have the same policy

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/dalle3.py
keepingitneil: yeah we talked about this being an example instead of in the plugin (which i'm in favor of). Good callout re: quickstart.

.github/workflows/check-types.yml
davidzhao: gotcha.. it'd be good to standardize.. the rest of our stuff all run on PRs.. so it's possible to have a private branch without triggering all the CI. and I think it'll be possible to require checks to pass before allowing merges.. I think we need that.

davidzhao: ok.. I think we can already do the latter.. that's a configuration.. I'll enable.

keepingitneil: cool, I changed it to run on push to main or dev and pr onto main or dev

davidzhao (approved):
lgtm!

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -69,8 +69,12 @@ def __init__(self, *, streaming_supported: bool) -> None:
         self._streaming_supported = streaming_supported
 
     @abstractmethod
-    async def synthesize(self, *, text: str) -> AsyncIterable[SynthesizedAudio]:
-        pass
+    async def synthesize(self, text: str, **_: Any) -> AsyncIterable[SynthesizedAudio]:
```
theomonnom, line 72: We shouldn't add , **_: Any

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -105,8 +105,8 @@ async def list_voices(self) -> List[Voice]:
 
     async def synthesize(
         self,
-        *,
         text: str,
+        **_,
```
theomonnom, line 109: What is the reason for **_, here?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -286,9 +286,9 @@ async def _listen_task(self, ws: aiohttp.ClientWebSocketResponse) -> None:
             if msg.type != aiohttp.WSMsgType.TEXT:
                 continue
 
-            msg = json.loads(msg.data)
-            if msg.get("audio"):
-                data = base64.b64decode(msg["audio"])
+            jsonMessage: Dict[str, Any] = json.loads(str(msg.data))
```
theomonnom, line 289: nit: use snakecase

livekit-plugins/livekit-plugins-nltk/livekit/plugins/nltk/sentence_tokenizer.py
```diff
@@ -6,7 +8,7 @@
 
 from livekit import agents
 
-import nltk
+import nltk  # type: ignore
```
theomonnom, line 11: what is the issue with nltk types?

livekit-agents/livekit/agents/tts/tts.py
theomonnom: I see the TTS of OpenAI has more arguments. The base class should not allow for more arguments.
So the change needs to be made in the OpenAI TTS, not the base class

keepingitneil: Hrmm so is the preferred pattern to make new TTS objects for new configurations (i.e. put configuration in the constructor)?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
keepingitneil: This was to match signature with base class. But we can take that discussion to the other thread

livekit-agents/livekit/agents/tts/tts.py
keepingitneil: actually as I think about it more, I think you're right and requiring configuration in the ctor makes more sense even if it requires more TTS objects per-configuration. Will make the adjustments in a PR soon

PR #219, opened 2024-04-07
Cleaner KITT

keepingitneil:


examples/kitt/agent.py
```diff
@@ -0,0 +1,179 @@
...
+    audio_stream_future = asyncio.Future[rtc.AudioStream]()
+
+    def on_track_subscribed(track: rtc.Track, *_):
+        if track.kind == rtc.TrackKind.KIND_AUDIO:
+            audio_stream_future.set_result(rtc.AudioStream(track))
+
+    def on_chat_message(message: rtc.ChatMessage):
+        print("Got chat message: ", message)
+
+    # IMPORTANT: Subscribe to audio tracks immediately. If you don't,
```
davidzhao, line 54: are we using autosubscribe in this case? If so this comment wouldn't be relevant.

if we are not using autoSubscribe, hoping there are some mechanics built-in that can help us subscribe to tracks. feels like learning about this published thing is really confusing

examples/kitt/agent.py
```diff
@@ -0,0 +1,179 @@
...
+                working_text += delta
+                print("Working text: ", working_text)
+                working_message = ChatMessage(role=ChatRole.USER, text=working_text)
+                chat_context = agents.llm.ChatContext(
+                    messages=committed_messages + [working_message]
+                )
+                # Cancel exisiting LLM inference
+                if latest_llm_stream is not None:
+                    await latest_llm_stream.aclose(wait=False)
+                latest_llm_stream = await llm.chat(history=chat_context)
```
davidzhao, line 96: I wonder if this could be chained somehow.. where each step only cancels the next step.. instead of having to worry about the entire chain.

examples/kitt/agent.py
```diff
@@ -0,0 +1,179 @@
...
+                # Cancel existing tts inference
+                await latest_tts_stream.aclose(wait=False)
+                # Pre-warm the tts
+                latest_tts_stream = tts.stream()
+                await llm_stream_queue.put(latest_llm_stream)
+
+    async def chat_stream_task():
+        while True:
```
davidzhao, line 104: trying to understand the logic here. when the user does not say a word.. `llm_stream` would be None and it would have broke out?

what happens when the user starts speaking but then stops.. would it keep waiting in `async for chunk...`?

finally, what's the significance of .flush()?

examples/kitt/agent.py
```diff
@@ -0,0 +1,179 @@
...
+            agent_speaking = True
+            committed_message = False
+            async for event in tts_stream:
+                if event.type == agents.tts.SynthesisEventType.AUDIO:
+                    # As soon as the agent speaks, we commit the working text
+                    # and the agent's speech to the chat.
+
+                    if not committed_message:
+                        committed_messages.append(
+                            ChatMessage(role=ChatRole.USER, text=working_text),
```
davidzhao, line 130: this seems a bit dangerous to access working_text from outside of the current task. we don't really know the sequencing of things with this right? i.e. is it possible to have:

1. user speaking a paragraph (p1)
2. llm processed p1
3. tts streaming response to that
4. user spoke more (p2a)
5. we are still processing tts_stream, and p2a ends up getting added to the history in the wrong order.

examples/kitt/agent.py
```diff
@@ -0,0 +1,179 @@
...
+            if tts_stream is None:
+                break
+            agent_speaking = True
+            committed_message = False
+            async for event in tts_stream:
+                if event.type == agents.tts.SynthesisEventType.AUDIO:
+                    # As soon as the agent speaks, we commit the working text
+                    # and the agent's speech to the chat.
+
+                    if not committed_message:
```
davidzhao, line 128: nit: renaming to distinguish visually from `committed_messages`

```suggestion
                    if not has_committed:
```

PR #224, opened 2024-04-10
Bump versions + Add READMEs to examples

keepingitneil:


theomonnom:
can you add the new deps to livekit-agents?

- attrs
- watchfiles
- colorlog
- pythonjsonlogger

examples/speech-to-text/agent.py
```diff
@@ -16,23 +16,33 @@ async def entrypoint(job: JobContext):
...
 
-    def on_track_subscribed(
-        track: rtc.Track,
-        pub: rtc.TrackPublication,
-        rp: rtc.RemoteParticipant,
-    ):
-        asyncio.create_task(process_track(track))
+    def on_track_subscribed(track: rtc.Track, *_):
+        if track.kind == rtc.TrackKind.KIND_AUDIO:
+            audio_stream_future.set_result(rtc.AudioStream(track))
```
theomonnom, line 23: i think it is a bad idea to use a future here, it'll throw if it is already done

keepingitneil: Hrmm yeah I guess for stt we can handle multiple tracks but for kitt we really can only handle one track. So I can wrap it in try/except and log an error if there are other tracks that come. Wdyt?

theomonnom: if you can find something simpler it would be cool, the goal here is to have the minimum amount of code
why not starting the function here?

theomonnom: the entry point doesn't have to be long-lived

keepingitneil: ahh really? I was under the impression that when it exits the agent is finished. But hrmm the tasks ought to go somewhere right. Fire-and-forget is probably not the best example code.

livekit-plugins/livekit-plugins-elevenlabs/setup.py
```diff
@@ -51,7 +51,7 @@
     python_requires=">=3.9.0",
     install_requires=[
         "livekit >= 0.9.2",
-        "livekit-agents >= 0.3.0",
+        "livekit-agents >= 0.5, < 0.6",
```
theomonnom, line 54: We don’t want to put a restriction on the maximum used version

theomonnom: Use ~=

livekit-agents/setup.py
```diff
@@ -54,6 +54,9 @@
         "protobuf>=3",
         "types-protobuf>=4,<5",
         "python-json-logger>=2,<3",
+        "attrs>=23.0,<24.0",
```
theomonnom, line 57: Can also use ~=

examples/speech-to-text/agent.py
theomonnom: The important thing is to keep a reference to it. These tasks will be cancelled when the job process ends

keepingitneil: wdyt about what I just updated it to? not simple enough?

livekit-plugins/livekit-plugins-elevenlabs/setup.py
keepingitneil: I think we do, we've had issues in the past where we've made a breaking API change (for example to tts) where there is a maximum version of agents that is supported.

For example:
- we change agents tts interface, bump agents to 0.6
- user is on this version of elevenlabs (maybe pinned in their requirements.txt) but maybe they didn't pin agents
- user builds new Docker image. Now they have 11labs at a version that isn't compatible with their version of agents.

Maybe it's an edge case but I think it happens. Especially in local environments where users aren't installing fresh every time.

theomonnom: Yes but it is because we break the API and did a patch instead of a major version

keepingitneil: No i mean if the user pins their requirements to elevenlabs and elevenlabs has no max. They can pull a version of agents that isn't compatible with elevenlabs

theomonnom: It is ok if you use ~= 0.6

keepingitneil: you mean ~= 0.5? yeah I think that is ok. Although we've needed to force an upgrade a few times in the history of agents with patches which is why I have >= but I think ~ will be ok. I'll change it

livekit-agents/setup.py
```diff
@@ -54,6 +54,9 @@
         "protobuf>=3",
         "types-protobuf>=4,<5",
         "python-json-logger>=2,<3",
+        "attrs~=23",
```
theomonnom, line 57: not in the diff but can you do the same for python-json-logger?

livekit-plugins/livekit-plugins-nltk/setup.py
```diff
@@ -46,7 +46,11 @@
     license="Apache-2.0",
     packages=setuptools.find_namespace_packages(include=["livekit.*"]),
     python_requires=">=3.9.0",
-    install_requires=["livekit >= 0.9.2", "nltk >= 3, < 4"],
+    install_requires=[
+        "livekit >= 0.9.2",
```
theomonnom, line 50: also use for livekit ~= 
otherwise this will break use code when adding breaking changes..

theomonnom:
can you add onnxruntime to vad and defaults to it

examples/speech-to-text/agent.py
```diff
@@ -13,26 +13,38 @@
...
+    def on_track_subscribed(track: rtc.Track, *_):
+        if track.kind == rtc.TrackKind.KIND_AUDIO:
+            stream_queue.put_nowait(rtc.AudioStream(track))
 
     job.room.on("track_subscribed", on_track_subscribed)
 
-    async for stt_event in stt_stream:
-        if stt_event.type == agents.stt.SpeechEventType.FINAL_TRANSCRIPT:
-            logging.info("Got transcript: %s", stt_event.alternatives[0].text)
+    for participant in job.room.participants.values():
```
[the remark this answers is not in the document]
keepingitneil: unfortunately we make no guarantees that on_track_subscribed will be called in this setup. I think the timing of things has changed in a way where it was working before but now we must iterate as well. As a followup, we should explore introducing a helper class that does this for you automatically in one-line of code

examples/speech-to-text/agent.py
```diff
@@ -13,26 +13,38 @@
...
+            stream_queue.put_nowait(rtc.AudioStream(track_pub.track))
+
+    async def process_track(audio_stream: rtc.AudioStream):
+        async for audio_frame_event in audio_stream:
+            stt_stream.push_frame(audio_frame_event.frame)
+
+    async def process_stt(stt_stream: agents.stt.STTStream):
+        async for stt_event in stt_stream:
+            if stt_event.type == agents.stt.SpeechEventType.FINAL_TRANSCRIPT:
+                logging.info("Got transcript: %s", stt_event.alternatives[0].text)
```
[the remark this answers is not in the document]
theomonnom: just print here, it's a tiny example

PR #225, opened 2024-04-10
Implement AutoSubscribe

keepingitneil:


livekit-agents/livekit/agents/ipc/job_main.py
```diff
@@ -35,7 +36,26 @@ async def _start(
 ) -> None:
     close_tx, close_rx = aio.channel()  # used by the JobContext to signal shutdown
 
-    cnt = room.connect(args.url, args.token)
+    auto_subscribe = args.auto_subscribe
+    opts = rtc.RoomOptions()
+    if auto_subscribe != AutoSubscribe.SUBSCRIBE_ALL:
+        opts.auto_subscribe = False
```
theomonnom, line 42: always false?

theomonnom: ah it defaults to True, can you make it explicit?

livekit-agents/livekit/agents/worker.py
```diff
@@ -292,7 +292,9 @@ def _reload_jobs(self, jobs: list[ActiveJob]):
     def _start_process(
         self, job: agent.Job, url: str, token: str, accept_data: AcceptData
     ):
-        proc = ipc.JobProcess(job, url, token, accept_data.entry)
```
theomonnom, line 295: let's add AcceptData to the ctor here, it'll be cleaner

livekit-agents/livekit/agents/ipc/job_main.py
```diff
@@ -6,6 +6,7 @@
 import os
 
 from livekit import rtc
+from livekit.agents.job_request import AutoSubscribe
```
theomonnom, line 9: Use relative import to avoid cyclic dependencies

livekit-agents/livekit/agents/ipc/job_process.py
```diff
@@ -5,6 +5,7 @@
 import threading
 from typing import Callable
 
+from livekit.agents.job_request import AutoSubscribe
```
theomonnom, line 8: Same here, use relative imports

livekit-agents/livekit/agents/ipc/protocol.py
```diff
@@ -5,6 +5,7 @@
 from typing import Callable, ClassVar
 
 from attrs import define
+from livekit.agents.job_request import AutoSubscribe
```
theomonnom, line 8: Same here, use relative imports

livekit-agents/livekit/agents/ipc/job_main.py
keepingitneil: I think it's right.
SUBSCRIBE_NONE != SUBSCRIBE_ALL so we set `opts.auto_subscribe = False`. Then in on_track_published we never subscribe. I will adjust so that the handler if never registered if you'd like

theomonnom (approved):

livekit-agents/livekit/agents/ipc/job_main.py
```diff
@@ -35,7 +36,27 @@ async def _start(
...
-    cnt = room.connect(args.url, args.token)
+    auto_subscribe = args.accept_data.auto_subscribe
+    opts = rtc.RoomOptions(auto_subscribe=True)
+    if auto_subscribe != AutoSubscribe.SUBSCRIBE_ALL:
+        opts.auto_subscribe = False
+
+        def on_track_published(pub: rtc.RemoteTrackPublication, *_):
+            if (
+                pub.kind == rtc.TrackKind.KIND_AUDIO
+                and auto_subscribe == AutoSubscribe.AUDIO_ONLY
```
davidzhao, line 47: for publisher kind, we should only subscribe if publisher identity matches

PR #228, opened 2024-04-11
TTS API remove double async

keepingitneil:


keepingitneil:
oh woops forgot to commit, one moment

Issue #202, opened 2024-03-26
deepgram transcribe quality in livekit much lower than when transcribing on deepgram demo website


aviadr1:
please compare the transcribing the accuracy of running the livekit+deepgram agent demo with running the deepgram demo on their website, you will see quality of transcription much higher on deepgram.

deepgram demo:
https://console.deepgram.com/project/<YOUR_PROJECT_ID>/mission/transcribe-your-voice-in-realtime

livekit+agent demo:
https://github.com/livekit/agents/tree/main/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram

davidzhao:
which model are you using on the website vs the plugin?

can you share any reproductions of this? without examples or reproductions, it's not clear to us what the issue is.

aviadr1:
the deepgram demo uses the defaults for every config and the "nova" model.
the livekit+agent demo uses modified code and configs.

I have been able to get similar performance to the deepgram demo by reverting back to the defaults using this code to replace `livekit.plugins.deepgram.stt.STT` : 

class MySTT(STT):
    def stream(self, *, language: Optional[str] = None) -> "SpeechStream":
        config = self._sanitize_options(language=language)
        return SpeechStream(
            config,
            api_key=self._api_key,
            sample_rate=24000,
        )

    async def _try_connect(self, session: aiohttp.ClientSession) -> aiohttp.ClientWebSocketResponse:
        # TODO: seems like using the default parameters as are available in in the STT demo
        #       https://console.deepgram.com/project/9ea3e34f-5776-47e1-9366-232fa0c5a887/mission/transcribe-your-voice-in-realtime
        #       works better somehow. the github for this is https://github.com/deepgram-devs/js-live-example/blob/main/public/client.js
        #       I'm not sure which of this parameters is the one that is causing the issue.
        #       need to experiment with it.
        live_config = {
            # "model": self._config.model,
            "model": "nova",
            # "punctuate": self._config.punctuate,
            # "smart_format": self._config.smart_format,
            # "interim_results": self._config.interim_results,
            "encoding": "linear16",
            "sample_rate": self._sample_rate,
            "channels": self._num_channels,
            # "endpointing": str(self._config.endpointing or "10"),
        }

        if self._config.language:
            live_config["language"] = self._config.language

        query_params = urlencode(live_config).lower()

        url = f"wss://api.deepgram.com/v1/listen?{query_params}"
[...]

PR #245, opened 2024-04-17
Uses default sample rate and original audio without resampling

vanics:
Uses default sample rate and original audio without resampling sample for depgram live stream

This addresses the issue highlighted here:
https://github.com/livekit/agents/issues/202

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -149,7 +149,7 @@ class SpeechStream(stt.SpeechStream):
         self,
         config: STTOptions,
         api_key: str,
-        sample_rate: int = 16000,
+        sample_rate: int = 44000,
         num_channels: int = 1,
         max_retry: int = 32,
     ) -> None:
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -279,9 +279,12 @@ async def send_task():
                 if isinstance(data, rtc.AudioFrame):
                     # TODO(theomonnom): The remix_and_resample method is low quality
                     # and should be replaced with a continuous resampling
-                    frame = data.remix_and_resample(
-                        self._sample_rate, self._num_channels
-                    )
+                    # frame = data.remix_and_resample(
```
vanics, line 282: This is sort of an anti-pattern but wanted to leave here to check the intention of still reworking this.

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -149,7 +149,7 @@ def __init__(
         self,
         config: STTOptions,
         api_key: str,
-        sample_rate: int = 16000,
+        sample_rate: int = 44000,
```
keepingitneil, line 152: 48000 is the common sample rate

vanics:
@aviadr1 Why would it be 44100? I believe it is 48K. I just fixed it!

aviadr1:
@vanics Oh, I am not sure what the original livekit sample rate is, but if it is not 48khz, then under that the customary sampling rate is 44100hz (and not 44000hz).

BTW is the livekit sampling rate ALWAYS 48000z?

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
keepingitneil: hey @vanics can we actually keep the remix_and_resample method. If the sample rates match, it's no-op but we don't want it to completely break if users override the default.

keepingitneil: Actually i'll merge and do a followup. Thanks for making this PR!

PR #251, opened 2024-04-24
improve logging

theomonnom:


livekit-agents/livekit/agents/ipc/job_main.py
```diff
@@ -54,8 +59,13 @@ async def _start(
     async def _start_if_valid():
         nonlocal usertask
         if start_req and room.isconnected():
```
keepingitneil, line 61: might as well do an early-out here too

keepingitneil (approved):

PR #259, opened 2024-04-29
VoiceAssistant - support function callings ootb

theomonnom:


examples/voice-assistant/assistant_calls.py
```diff
@@ -0,0 +1,117 @@
...
+    assistant = VoiceAssistant(
+        vad=silero.VAD(),
+        stt=deepgram.STT(),
+        llm=gpt,
+        tts=elevenlabs.TTS(),
+        fnc_ctx=AssistantFnc(),
+    )
+
+    @assistant.on("will_synthesize_llm")
+    def _will_synthesize_llm(chat_ctx: llm.ChatContext, user_msg: llm.ChatMessage):
```
theomonnom, line 57: this may be weird, I'm thinking about adding the chat_context to the constructor. But it may be confusing on whoever own the object

examples/voice-assistant/assistant_calls.py
```diff
@@ -0,0 +1,117 @@
...
+    ):
+        logging.info("toggle_light %s %s", room, status)
+        ctx = AssistantContext.get_current()
+        key = "enabled_rooms" if status else "disabled_rooms"
+        li = ctx.get_metadata(key, [])
+        li.append(room)
+        ctx.store_metadata(key, li)
+
+    @llm.ai_callable(desc="User want the assistant to stop/pause speaking")
+    def stop_speaking(self):
```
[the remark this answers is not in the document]
theomonnom: When you ask the LLM to stop talking when telling a story, you don't want it to answer. Instead you just call this function and don't call Assistant.say

examples/voice-assistant/assistant_calls.py
```diff
@@ -0,0 +1,117 @@
...
+
+class Room(Enum):
+    BEDROOM = "bedroom"
+    LIVING_ROOM = "living room"
+    KITCHEN = "kitchen"
+    BATHROOM = "bathroom"
+    OFFICE = "office"
+
+
+class AssistantFnc(llm.FunctionContext):
```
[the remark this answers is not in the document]
theomonnom: Yeah agree, I think right now it makes sense to separate it inside a class.
We could change different function context on the fly in the future. And having the functions scoped is a good practice IMO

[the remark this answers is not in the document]
theomonnom: Nothing prevents us from supporting this design in the future (no blocking changes)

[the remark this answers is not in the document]
davidzhao: does it have to inherit function context? or could it be passed in as a param? i.e.

```
class Assistant:
    def __init__(self, fctx):
       self.fctx.register(self.my_func, desc="blah blah")
```

examples/voice-assistant/assistant_calls.py
davidzhao: how often is this called? Is it a single time before the llm is initialized? in that case, it'd be better to call the function directly instead of a callback.

examples/voice-assistant/assistant_calls.py
```diff
@@ -0,0 +1,117 @@
...
+            0,
+            llm.ChatMessage(
+                role=llm.ChatRole.SYSTEM,
+                text="You are a voice assistant created by LiveKit. Your interface with users will be voice. You should use short and concise responses, and avoiding usage of unpronouncable punctuation.",
+            ),
+        )
+
+    @assistant.on("agent_speech_interrupted")
+    def _agent_speech_interrupted(chat_ctx: llm.ChatContext, msg: llm.ChatMessage):
+        msg.text += "... (user interrupted you)"
```
davidzhao, line 69: how would this be reflected in the chat history? how do we output part of the agent's text?

examples/voice-assistant/assistant_calls.py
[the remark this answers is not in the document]
keepingitneil: I see, so what is the behavior of the assistant without this `stop_speaking` fn? Since we already support interruption at the agent level and we already are handling the interruption with `@assistant.on("agent_speech_interrupted")` is this fn needed?

BTW I'm asking because I don't know the answer and not because I'm recommending something different. But maybe since I'm confused others will be too.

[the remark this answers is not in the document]
theomonnom: This is a very easy way to detect when to stop talking vs we interrupt to ask a new question

examples/voice-assistant/assistant_calls.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @assistant.on("function_calls_done")
```
davidzhao, line 71: nit: just wording.

```suggestion
    @assistant.on("function_calls_finished")
```

examples/voice-assistant/assistant_calls.py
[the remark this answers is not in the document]
theomonnom: When users are interrupting, the new speech is automatically sent to the LLM and we will get a new answer, this is how VoiceAssistant is working, wdyt?

examples/voice-assistant/assistant_calls.py
```diff
@@ -0,0 +1,117 @@
...
+    def _function_calls_done(ctx: AssistantContext):
+        logging.info("function_calls_done %s", ctx)
+        enabled_rooms = ctx.get_metadata("enabled_rooms", [])
+        disabled_rooms = ctx.get_metadata("disabled_rooms", [])
+
+        async def _handle_answer():
+            prompt = "Make a summary of the following actions you did:"
+            if enabled_rooms:
+                enabled_rooms_str = ", ".join(enabled_rooms)
+                prompt += f"\n - You enabled the lights in the following rooms: {enabled_rooms_str}"
```
davidzhao, line 81: is this to help the agent keep track of state? wanna add some docs to indicate the intent?

examples/voice-assistant/assistant_calls.py
```diff
@@ -0,0 +1,117 @@
...
+
+            stream = await gpt.chat(chat_ctx)
+            await assistant.say(stream)
+
+        if enabled_rooms or disabled_rooms:
+            asyncio.ensure_future(_handle_answer())
+
+    # start the assistant with the first participant found inside the room
+
+    @ctx.room.on("participant_connected")
```
davidzhao, line 99: i wonder if it's better to have `assistant.start` handle this? i.e. if you don't pass in a participant, it'll automatically wait for the first participant. wdyt?

examples/voice-assistant/assistant_calls.py
davidzhao, line 1: nit: perhaps update to function_calling.py

examples/voice-assistant/assistant_calls.py
theomonnom: Just wondering the case where there is multiple people in the room. but I guess it is OK as a default impl

examples/voice-assistant/assistant_calls.py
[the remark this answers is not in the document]
theomonnom: For now it has to inherit

livekit-agents/livekit/agents/voice_assistant/assistant.py
```diff
@@ -553,41 +621,38 @@ async def _play_speech_if_validated(self, data: _SpeechData) -> None:
...
+
+            msg = allm.ChatMessage(
+                text=data.collected_text,
+                role=allm.ChatRole.ASSISTANT,
+            )
+
+            if data.add_to_ctx:
+                self._chat_ctx.messages.append(msg)
+                if data.interrupted:
+                    self.emit("agent_speech_interrupted", self._chat_ctx, msg)
```
davidzhao, line 653: can we use constants for events? instead of strings?

davidzhao (approved):
LG 🔥 🔥 🔥 

just couple of nits here and there,

keepingitneil (approved):

examples/voice-assistant/assistant_calls.py
theomonnom: will followup in another PR for dynamic chat context

PR #262, opened 2024-04-30
transcription forwarder & add transcription to VoiceAssistant

keepingitneil:


livekit-agents/livekit/agents/voice_assistant/transcription_manager.py
```diff
@@ -0,0 +1,68 @@
...
+from livekit import rtc
+
+
+class TranscriptionManager:
+    def __init__(self, room: rtc.Room, participant_identity: str):
+        self._participant_identity = participant_identity
+        self._segments: List[rtc.TranscriptionSegment] = []
+        self._track_sid = ""
+        self._room = room
+        self._current_id = 1
```
davidzhao, line 14: this should be an uuid, we don't want it to conflict between TranscriptionManager instances.

livekit-agents/livekit/agents/voice_assistant/transcription_manager.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def update_track_sid(self, track_sid: str) -> None:
+        self._track_sid = track_sid
+
+    def reset_transcription(self) -> None:
+        self._segments = []
+
+    def final_transcription(self, text: str, language: str = "english"):
```
davidzhao, line 22: seems strange to have two diff methods for this? wdyt about `send_transcription(..., final=True)` ?

davidzhao: also feels like as an interface.. the following would be cleaner:

```
segment = transcription_manager.start_segment(identity, trackID)
segment.update(full_text, final=True, duration=..., lang=...)
# once final, further .updates would throw an exception
```

It feels like this way, we do not have to create separate instances of TranscriptionManager for every user in the room. feels more natural especially with a dynamic number of speakers

livekit-agents/livekit/agents/voice_assistant/transcription_manager.py
theomonnom, line 1: transcription manager shouldn't be inside voice_assistant

theomonnom (changes requested):
The TranscriptionManager is per room, so users don't need to keep track of too much states 

### TTS Usecase

```python
tr_man = TranscriptionManager(room)
segment = tr_man.start_segment(identity=participant_identity)
async for ev in stt_stream:
    segment.update(event)
```

<details>
  <summary>Signature</summary>

```python
def start_stt_stream(self, identity: str, track_id: str | None = None) -> SegmentBuilder: ...
```
Note that the track_id is optional because it will automatically be guessed, let's get the first MICROPHONE track
</details>



### Other sources, whatever str generator (e.g LLMStream)

Similar to @davidzhao proposition 
```python
tr_man = TranscriptionManager(room)
segment = tr_man.start_segment(identity=participant_identity, language='en')
full_text = ""
async for s in text_generator:
    full_text += s
    segment.update(full_text)

segment.end() # apply final=True 
```
I'm not sure about the full_text variable, maybe the update method automatically concat the strings internally?

<details>
  <summary>Signature</summary>

```python
def start_segment(self, identity: str, language: str, track_id: str | None = None) -> SegmentBuilder: ...
```
</details>

livekit-agents/livekit/agents/voice_assistant/assistant.py
```diff
@@ -685,6 +708,12 @@ async def _synthesize_task(
             _first_frame = True
             async for audio in self._tts.synthesize(data.source):
                 if _first_frame:
+                    if self._agent_transcription_manager:
```
theomonnom, line 711: This isn't the right place to do it, this should maybe be inside the maybe_play_speech and only triggered if validated channel is closed

livekit-agents/livekit/agents/utils/transcription_manager.py
```diff
@@ -0,0 +1,65 @@
...
+        self._language = language
+        self._track_id = track_id
+        if isinstance(participant, rtc.Participant):
+            self._participant_identity = participant.identity
+        else:
+            self._participant_identity = participant
+        self._start_time = time.time_ns()
+        self._text = ""
+
+    def update(self, text: str):
```
davidzhao, line 44: hmm shouldn't update publish as well, otherwise it wouldn't be real-time updates?

I think it might be better to have final as a flag on `update`, versus final vs commit. If we handled the STTEvent, then it could be handled automatically.

the other thing is thinking about speed of updates.. I'm wondering if the updates should be pushed into a background queue, so we can do some sort of automatically throttling.. for example, if I did:

```
segment.update('a')
segment.update('a b')
```

this should only trigger a single update. (we can update this in a future PR, but just thinking out loud)

vanics:
I love the work that is happening here. :)

livekit-agents/livekit/agents/transcription/tts_forwarder.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+
+        self._seg_queue = asyncio.Queue[_SegmentData | None]()
+        self._seg_queue.put_nowait(first_segment)
+
+        # the forwarding of the transcription is started as soon as the user push the first audio frame.
+        self._start_future = asyncio.Future()
+
+    def validate(self) -> None:
+        """Validate must be called once we started to playout the audio to the user"""
```
theomonnom, line 75: mb start is a better name?

livekit-agents/livekit/agents/transcription/transcription.py
```diff
@@ -0,0 +1,87 @@
...
+        language: str = "",
+        track_id: str | None = None,
+        speed: float = 4,
+        word_tokenizer: tokenize.WordTokenizer = tokenize.basic.WordTokenizer(),
+        sentence_tokenizer: tokenize.SentenceTokenizer = tokenize.basic.SentenceTokenizer(),
+        hyphenate_word: Callable[[str], list[str]] = tokenize.basic.hyphenate_word,
+    ):
+        """
+        Forward TTS transcription to the users. This function tries to imitate the right timing of
+        speech with the synthesized text. The first estimation is based on the speed parameter. Omce
```
keepingitneil, line 43: NIT: "Omce" -> "Once"

livekit-agents/livekit/agents/transcription/transcription.py
```diff
@@ -0,0 +1,87 @@
...
+        we have received all the audio of a specific text segment, we recalculate the speed using
+        the length of the text & audio.
+
+        The word_tokenizer is used to forward the transcription word by word.
+        The sentence_tokenizer is used to split the streamted text into mulitple transcription
+        segments.
+        hyphenate_word is used to
```
keepingitneil, line 50: Incomplete?

livekit-agents/livekit/agents/transcription/tts_forwarder.py
keepingitneil: Wondering if we need this at all? Under what conditions would we push the first audio frame and not want to call validate?

theomonnom:
> Wondering if we need this at all? Under what conditions would we push the first audio frame and not want to call validate?

VoiceAssistant synthesize the audio frames but the playout may not be validated yet

keepingitneil:
> > Wondering if we need this at all? Under what conditions would we push the first audio frame and not want to call validate?
> 
> VoiceAssistant synthesize the audio frames but the playout may not be validated yet

Ahh right, I'm a little worried that it's too easy to misuse the `TTSSegmentsForwarder` class. Wondering if it makes sense to start a segment and get a handle to pass around: `handle = tr_forwarder.start()`. `handle.push_text(), handle.validate(), handle.discard(), etc etc`

theomonnom:
This isn't the goal of TTSSegmentsForwarder, here we only know the text we're pushing to (e.g LLM response, and some streamed audio).

This class tries to sync stuff

theomonnom:
But yeah agree we should have a easy way to create segments in a builder pattern for other use cases. But it don't think I should do it inside this PR? Also I'm not seeing an use case yet

theomonnom:
For now the target was TTS & STT transcriptions

keepingitneil:
I was less worried about the builder pattern itself and more worried about how easy it is to to accidentally mismatch push_text, push_audio, mark_*_end. Although my suggestion above doesn't actually prevent that...

Anyways yeah this looks good

theomonnom:
Ah yeah good point, I didn't find a better API for that. Happy to get suggestions. the mark_*_end stuff is a source of confusion for sure.

keepingitneil:
Yeah, for example, we should always push text before pushing audio right? It's easy to accidentally push text, mark segment end, push audio. So if we start a segment and get a handle, at least we know we won't accidentally push audio without text right? Idk, maybe _enter_,_exit_ could also be helpful to automatically end?

Anyways maybe for a future iteration to think about these things.

theomonnom:
pushing audio before text is OK. The only thing that isn't ok and will raise an exception is to mark_audio_end before ending the text segment

livekit-agents/livekit/agents/tokenize/basic.py
```diff
@@ -34,7 +34,7 @@ def __init__(
             stream_context_len=stream_context_len,
         )
 
-    def tokenize(self, *, text: str, language: str | None = None) -> list[str]:
+    def tokenize(self, text: str, language: str | None = None) -> list[str]:
```
davidzhao, line 37: this is a bit inconsistent with `stream`, where language is required to be a kwarg.. I would say let's be consistent here.

livekit-agents/livekit/agents/transcription/stt_forwarder.py
```diff
@@ -0,0 +1,87 @@
...
+        self._current_id = _uuid()
+
+    async def _run(self):
+        try:
+            while True:
+                seg = await self._queue.get()
+                if seg is None:
+                    break
+
+                transcription = rtc.Transcription(
```
davidzhao, line 41: should there be rate limiting (i.e. max once a second) and bundling? where if there are multiple updates queued for the same segment_id, it feels like we shouldn't send multiple updates.. (TODO is fine too)

livekit-agents/livekit/agents/transcription/transcription.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def forward_stt_transcription(
+        self, *, participant: rtc.Participant | str, track_id: str | None = None
+    ) -> "STTSegmentsForwarder":
+        """
+        Forward STT transcription to the users. (Useful for client-side rendering)
+        """
+        identity = participant if isinstance(participant, str) else participant.identity
+        if track_id is None:
+            track_id = self._find_micro_track_id(identity)
```
davidzhao, line 22: is the idea that it defaults to the first microphone track? I see that we are always passing in track_id, is the thinking that end-users would not?

livekit-agents/livekit/agents/transcription/transcription.py
```diff
@@ -0,0 +1,87 @@
...
+from .. import tokenize
+from .stt_forwarder import STTSegmentsForwarder
+from .tts_forwarder import TTSOptions, TTSSegmentsForwarder
+
+
+class TranscriptionManager:
+    def __init__(self, room: rtc.Room):
+        self._room = room
+
+    def forward_stt_transcription(
```
davidzhao, line 14: `forward*` sounds like it's actually performing the forwarding, which it does not.

```suggestion
    def create_stt_forwarder(
```

livekit-agents/livekit/agents/transcription/transcription.py
```diff
@@ -0,0 +1,87 @@
...
+        if track_id is None:
+            track_id = self._find_micro_track_id(identity)
+
+        return STTSegmentsForwarder(
+            room=self._room,
+            participant_identity=identity,
+            track_id=track_id,
+        )
+
+    def forward_tts_transcription(
```
davidzhao, line 30:
```suggestion
    def create_tts_forwarder(
```

livekit-agents/livekit/agents/transcription/tts_forwarder.py
davidzhao: same question as neil.. I'm not sure why we would make a user call a required function?

livekit-agents/livekit/agents/transcription/tts_forwarder.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def push_text(self, text: str | None) -> None:
+        if text is not None:
+            cur_seg = self._pending_segment.cur_text
+            cur_seg.text += text
+            cur_seg.sentence_stream.push_text(text)
+        else:
+            self.mark_segment_end()
+
+    def mark_segment_end(self) -> None:
```
davidzhao, line 105:
```suggestion
    def mark_text_segment_end(self) -> None:
```

davidzhao: or maybe they should be called: `end_text_segment` and `end_audio_segment`

livekit-agents/livekit/agents/transcription/tts_forwarder.py
```diff
@@ -0,0 +1,230 @@
...
+            seg = self._pending_segment.q.popleft()
+        except IndexError:
+            raise IndexError(
+                "mark_audio_segment_end called before any mark_segment_end"
+            )
+
+        seg.avg_speed = len(self._calc_hyphenes(seg.text)) / seg.audio_duration
+        self._pending_segment.cur_audio = self._pending_segment.q[0]
+
+    def push_text(self, text: str | None) -> None:
```
davidzhao, line 97: can we add docstrings on usage? also I think for non-streaming LLMs, users would need to explicitly call `end_text_segment`.. this feels cleaner to call `push_text(text, end_segment=True)`, versus assuming None ends a segment..

livekit-agents/livekit/agents/voice_assistant/assistant.py
```diff
@@ -175,8 +175,8 @@ def on(self, event: EventTypes, callback: Callable | None = None) -> Callable:
                 - user_speech_committed: the user speech was committed to the chat context
                 - agent_speech_committed: the agent speech was committed to the chat context
                 - agent_speech_interrupted: the agent speech was interrupted
-                - function_calls_completed: all function calls have been completed
-                - will_synthesize_llm: the assistant will synthesize the LLM output
+                - function_calls_collected: all function calls have been completed
```
davidzhao, line 178: TBH `collected` does not sound as clear as `completed`

livekit-agents/livekit/agents/transcription/stt_forwarder.py
theomonnom: I don't think STT often change the segment, I think it is OK to forward everything we get from them

livekit-agents/livekit/agents/transcription/transcription.py
theomonnom: The thinking is to make it optional, so you don't have to think about it. It'll use the microphone by default, But yeah feel free for the users to pass their own track_id

livekit-agents/livekit/agents/voice_assistant/assistant.py
theomonnom: function_calls_finished means that all functions have been called as well as their execution.
function_calls_collected means that we just got all functions we need to launch

livekit-agents/livekit/agents/transcription/tts_forwarder.py
theomonnom: TTS has the same method, I think it was OK to do it this way because we have mark_end_segment

livekit-agents/livekit/agents/tokenize/basic.py
theomonnom: ah, language should be removed from there actually, the right place is the constructor

theomonnom: will follow the STT API

examples/text-to-speech/transcription_protocol.py
```diff
@@ -0,0 +1,168 @@
...
+        async for ev in tts_stream:
+            if ev.type != tts.SynthesisEventType.AUDIO:
+                continue
+            assert ev.audio is not None
+            playout_q.put_nowait(ev.audio.data)
+
+        playout_q.put_nowait(None)
+
+    synth_task = asyncio.create_task(_synth_task())
+    playout_task = asyncio.create_task(_playout_task(playout_q, source))
```
davidzhao: how do we signal the start/end of audio here?

examples/text-to-speech/sync_tts_transcription.py
```diff
@@ -0,0 +1,168 @@
...
+    tts_11labs = elevenlabs.TTS()
+
+    # publish an audio track
+    source = rtc.AudioSource(tts_11labs.sample_rate, tts_11labs.num_channels)
+    track = rtc.LocalAudioTrack.create_audio_track("agent-mic", source)
+    options = rtc.TrackPublishOptions(source=rtc.TrackSource.SOURCE_MICROPHONE)
+    await job.room.local_participant.publish_track(track, options)
+
+    #
+    # 1. example using a tts stream (we split text into chunks just for the example)
```
davidzhao, line 55: these are separate examples, can we put them in three separate functions?
i.e. 
* streaming multiple segments
* single_segment
* deferred playout

livekit-agents/livekit/agents/transcription/_utils.py
```diff
@@ -0,0 +1,28 @@
...
+            break
+
+    if track_id is None:
+        raise ValueError(f"participant {identity} does not have a microphone track")
+
+    return track_id
+
+
+def segment_uuid() -> str:
+    return "TR_" + str(uuid.uuid4())[:12]
```
davidzhao, line 28: does uuid4 come with `-`? if so can we strip?

```suggestion
    return "SG_" + str(uuid.uuid4())[:12]
```

livekit-agents/livekit/agents/voice_assistant/assistant.py
davidzhao: got it.. in that case can we update the comment here? it still says: `all function calls have been completed`

davidzhao (commented):
a couple of minor ones. otherwise lgtm!

davidzhao (approved):

PR #288, opened 2024-05-14
Bump versions

keepingitneil:


livekit-plugins/livekit-plugins-nltk/livekit/plugins/nltk/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.5.dev1"
+__version__ = "0.5.0"
```
davidzhao, line 15: interesting.. this is the only one on a higher version than the rest.. I wonder why.

davidzhao (approved):

PR #290, opened 2024-05-16
add chunked stream

theomonnom:


livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -31,6 +33,33 @@ class SynthesisEvent:
...
+        Utility method to collect every frame in a single call
+        """
+        frames = []
+        async for frame in self:
+            frames.append(frame)
+
+        return misc.merge_frames(frames)
+
+    @abstractmethod
+    async def __anext__(self) -> rtc.AudioFrame: ...
```
theomonnom, line 52: Should this returns SynthesizedAudio instead?

keepingitneil (approved):

theomonnom:
This isn't a breaking change anymore. I'm returning tts.SynthesizedAudio.

Google TTS has breaking changes because the API wasn't correct.

PR #296, opened 2024-05-17
better reconnection & ignore irrelevant messages

theomonnom:
also fix slow reconnection & make livekit-agents work on python 3.9 (still need some work for plugins)

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -218,10 +218,10 @@ class Worker(utils.EventEmitter[EventTypes]):
 
         await asyncio.gather(*close_co, return_exceptions=True)
 
+        self._msg_chan.close()
+
         await self._http_server.aclose()
         await self._session.close()
-
-        self._msg_chan.close()
         await self._close_future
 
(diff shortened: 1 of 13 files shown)


theomonnom:
main difference is that we're deciding to drop a message if we're currently reconnecting depending on some conditions.

and we also wait for worker registration before validating a ws connection (we should implement worker reconnection/resume in the future)

livekit-agents/livekit/agents/worker.py
```diff
@@ -226,75 +261,93 @@ async def _worker_task(self):
...
+                    agent_url, headers=headers, autoping=True
+                )
+
                 retry_count = 0
 
+                # do worker registration
+                req = agent.WorkerMessage()
+                req.register.type = self._opts.worker_type
+                req.register.allowed_permissions.CopyFrom(
+                    models.ParticipantPermission(
```
davidzhao, line 277: I don't believe `agent` needs to be set.. and we also need to copy the [can_publish_sources](https://github.com/livekit/protocol/blob/main/protobufs/livekit_models.proto#L80) field

davidzhao (approved):

PR #297, opened 2024-05-17
add optional global http session

theomonnom:
also fix leaked http session on plugins


livekit-agents/livekit/agents/utils/http_session.py
davidzhao, line 1: nice!

livekit-agents/livekit/agents/worker.py
```diff
@@ -109,30 +109,40 @@ def __init__(
...
+        self._session: aiohttp.ClientSession | None = None
+        self._api: api.LiveKitAPI | None = None
 
     async def run(self):
+        if not self._closed:
+            raise Exception("worker is already running")
+
         logger.info("starting worker", extra={"version": __version__})
+
+        # this LiveKit API object is only really useful in non-third-party workers
```
davidzhao: nit: what does non third party workers mean? standard?

davidzhao (approved):
lgtm!

livekit-agents/livekit/agents/worker.py
theomonnom: yes standard

theomonnom: will rename

PR #310, opened 2024-05-21
fix utils.http_session creating multiple clients  

theomonnom:
ContextVar was OK is JobProcesses because we have one main task.
But this isn't true inside tests and maybe on usercode inside the main worker process


theomonnom:
mmh may require some more work, each test has its own event loop?

davidzhao:
seems like moving to global will be worse? i.e. multiple runloops sharing the same var?

davidzhao (approved):

theomonnom:
http_session() ont its own now initialize nothing (we know have predictable path of where this should be used)
for now:
 - inside the usercb of jobs
 - tests

PR #313, opened 2024-05-22
Fix python regex in basic_sent. Also added GPT-4o to OAI models

davidzhao:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -25,7 +25,7 @@ class LLM(llm.LLM):
     def __init__(
         self,
         *,
-        model: str | ChatModels = "gpt-4-turbo",
+        model: str | ChatModels = "gpt-4o",
```
theomonnom, line 28: 🤖🚀

theomonnom (approved):

PR #319, opened 2024-05-23
Bump versions

keepingitneil:


examples/speech-to-text/deepgram_stt.py
```diff
@@ -35,9 +35,8 @@ async def entrypoint(job: JobContext):
     async def transcribe_track(track: rtc.Track):
         audio_stream = rtc.AudioStream(track)
         stt_forwarder = transcription.STTSegmentsForwarder(
-            room=job.room, participant=job.room.local_participant
+            room=job.room, participant=job.room.local_participant, track=track
```
keepingitneil, line 38: seems like there was a race in the logic where it couldn't find the microphone track even after the track subscribed event triggered with the microphone track. So here I explicitly set the track to get the example working.

theomonnom (approved):

davidzhao (approved):

PR #324, opened 2024-05-27
improvements to AI functions

theomonnom:
- allow to add new AI functions at runtime
    - Useful if the choices are loaded from a database for example
- voice-assistant/function_calling.py now reflects a more realworld scenario
    - forward current states to llm context
    - add a bunch of comments 
    - show how to add aifunction at runtime
- polish some API stuff (breaking changes)
    - renamed history to fnc_ctx inside LLM
- added `will_create_llm_stream` to VoiceAssistant (this isn't an event)
    - It allows to inject stuff before any LLM call
    - if a custom LLM stream is returned, it will be used instead of assistant creating its own
- added USE_DOCSTRING to ai_function
    - this is cool when u have a long description for an ai function
- added a bunch of getters for tts, stt, vad, llm to VoiceAssistant
- added fnc_ctx getter/setter to VoiceAssistant
- renamed AssistantContext to AssistantCallContext
- add array support
- add choices to TypeInfo (shortcut for enum)
- fix int and float parameters (treated as number by OAI)


livekit-agents/livekit/agents/llm/function_context.py
```diff
@@ -14,94 +14,156 @@
...
+        description: str | _UseDocMarker | None = None,
+        auto_retry: bool = True,
+    ) -> Callable:
+        def deco(f):
+            _set_metadata(f, name=name, desc=description, auto_retry=auto_retry)
+            self._register_ai_function(f)
+
+        return deco
+
+    def _register_ai_function(self, fnc: Callable) -> None:
```
keepingitneil, line 102: Should we have a public version of this? Should we have an unregister method? Maybe out of scope for this pr

keepingitneil (approved):

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -29,35 +29,33 @@ class LLM(abc.ABC):
...
+
+    def __aiter__(self) -> AsyncIterator[ChatChunk]:
+        return self
 
     @abc.abstractmethod
     async def __anext__(self) -> ChatChunk: ...
 
     @abc.abstractmethod
-    async def aclose(self, wait: bool = True) -> None: ...
+    async def aclose(self) -> None: ...
```
davidzhao, line 61: how do we ensure that the reader has consumed all of the content then?

theomonnom: we never have to for LLM, you push once. Also when the stream ends, you know you have consumed everything.

There is also the async function gather_function_results that is used to make sure we wait for all function calls to resolve

examples/voice-assistant/function_calling.py
```diff
@@ -23,81 +37,181 @@ class Room(Enum):
...
+            str,
+            llm.TypeInfo(
+                desc="The name or identifier of the contact to whom the message will be sent.",
+                choices=contacts,
+            ),
+        ],
+        message: str,
+    ):
+        logging.info("sending a message to %s: %s", contact, message)
+        call_ctx = AssistantCallContext.get_current()
```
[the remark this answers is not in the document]
theomonnom: This is based on [contextvars](https://docs.python.org/3/library/contextvars.html) which should be pythonic.
The good news is that we don't really care about this context for most usecases now. You can directly return values. But agree that we could infer the parameter and automatically fill the value. But this means the specific FunctionContext must only be used with VoiceAssistant.

I don't have strong opinions here, I think it is OK to ignore as it seems like a feature that we can add later

livekit-agents/livekit/agents/llm/llm.py
theomonnom: Also I'll make a PR and a changelog for it, but I have plans to remove this argument for every baseclass we have.
I think it is better to have an explicit function in most cases.

PR #332, opened 2024-05-30
interrupt stream if trigger say() while waiting for play task

seanmuirhead:
**UPDATE (05/31/24):** This has mitigated the issue where a say() task sneaks in while the LLMStream is waiting. However, this fix may have side effects as we (Assort) just ran into the same issue, but with 2 LLMStreams being synthesized at the same time.

Ran into an issue where we have two `self._playing_task`s running at the same time, resulting in audio issues. 

**To Reproduce This Issue:**

1. Add the following listener to the `minimal_assistant.py` script: 
  `assistant.on("agent_stopped_speaking", lambda: asyncio.create_task(assistant.say("This is a long message that run in parallel with the following LLMStream. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam nec purus nec nunc ultricies ultricies.")))`
  
2. Start the agent and say some things, making sure that your interrupting it a couple of times. Eventually, you should hear the agent become inaudible (it will sound like it's trying to say 2 things at once).

Please lmk if there are better ways to approach this.

### DIFF
--- a/livekit-agents/livekit/agents/voice_assistant/assistant.py
+++ b/livekit-agents/livekit/agents/voice_assistant/assistant.py
@@ -162,6 +162,7 @@ class VoiceAssistant(utils.EventEmitter[EventTypes]):
 
         # playout state
         self._maybe_answer_task: asyncio.Task | None = None
+        self._data_waiting_to_play: _SpeechData | None = None
         self._playing_speech: _SpeechData | None = None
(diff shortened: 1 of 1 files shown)


seanmuirhead:
Update: Updating to the most recent `livekit-agents 0.7.0` seems to have fixed this issue. Closing this out

PR #335, opened 2024-05-31
Introduce vision capabilities to LLM

keepingitneil:
Draft PR for API feedback:
- New `livekit-agents[images]` optional package for image utilities (jpg, png, resizing, etc)
- ChatMessages can now take a list of images
- OpenAI LLM implementation (untested RN but take a look at the strategy and APIs)

### DIFF
--- a/examples/_deployed/kitt/kitt.py
+++ b/examples/_deployed/kitt/kitt.py
@@ -2,6 +2,7 @@ import asyncio
 import copy
 import logging
 from collections import deque
+from typing import Annotated
 
 from livekit import agents, rtc
 from livekit.agents import JobContext, JobRequest, WorkerOptions, cli
@@ -18,10 +19,17 @@ MAX_IMAGES = 3
 
 class AssistantFnc(agents.llm.FunctionContext):
     @agents.llm.ai_callable(
-        desc="Called when asked to look at something or see something or evaulate a visual using your vision capabilities"
+        desc="Called when asked to evaluate something that would require vision capabilities."
     )
-    async def image(self):
-        pass
+    async def image(
+        self,
+        user_msg: Annotated[
+            str,
+            agents.llm.TypeInfo(desc="The user message that triggered this function"),
+        ],
+    ):
+        ctx = AssistantContext.get_current()
+        ctx.store_metadata("user_msg", user_msg)
 
 
 async def entrypoint(ctx: JobContext):
@@ -37,7 +45,9 @@ async def entrypoint(ctx: JobContext):
         ]
     )
 
-    gpt = openai.LLM()
+    gpt = openai.LLM(
+        model="gpt-4o",
+    )
     latest_image: rtc.VideoFrame | None = None
     img_msg_queue: deque[agents.llm.ChatMessage] = deque()
     assistant = VoiceAssistant(
@@ -45,6 +55,7 @@ async def entrypoint(ctx: JobContext):
         stt=deepgram.STT(),
         llm=gpt,
         tts=elevenlabs.TTS(encoding="pcm_44100"),
+        fnc_ctx=AssistantFnc(),
         chat_ctx=initial_ctx,
     )
 
@@ -64,21 +75,29 @@ async def entrypoint(ctx: JobContext):
 
         asyncio.create_task(_answer_from_text(msg.message))
 
-    async def respond_to_image():
-        nonlocal latest_image, img_msg_queue
+    async def respond_to_image(user_msg: str):
+        nonlocal latest_image, img_msg_queue, initial_ctx
         if not latest_image:
             return
-        initial_ctx.messages[-1].images.append(
-            agents.llm.ChatImage(image=latest_image, dimensions=(128, 128))
+
+        initial_ctx.messages.append(
+            agents.llm.ChatMessage(
+                role=agents.llm.ChatRole.USER,
+                text=user_msg,
+                images=[
+                    agents.llm.ChatImage(image=latest_image, dimensions=(128, 128))
+                ],
+            )
         )
         img_msg_queue.append(initial_ctx.messages[-1])
         if len(img_msg_queue) >= MAX_IMAGES:
-            img_msg_queue.popleft()
+            msg = img_msg_queue.popleft()
+            msg.images = []
 
         stream = await gpt.chat(initial_ctx)
-        await assistant.say(stream)
+        await assistant.say(stream, allow_interruptions=True, add_to_ctx=True)
 
-    async def video_track_process():
+    async def process_video_track():
         nonlocal latest_image
         while True:
             await asyncio.sleep(0.1)
@@ -98,13 +117,16 @@ async def entrypoint(ctx: JobContext):
 
     @assistant.on("function_calls_finished")
     def _function_calls_done(ctx: AssistantContext):
-        asyncio.ensure_future(respond_to_image())
+        user_msg = ctx.get_metadata("user_msg")
+        if not user_msg:
+            return
+        asyncio.ensure_future(respond_to_image(user_msg))
 
     assistant.start(ctx.room)
 
     await asyncio.sleep(3)
     await assistant.say("Hey, how can I help you today?", allow_interruptions=True)
-    await video_track_process()
+    await process_video_track()
 
 
 async def request_fnc(req: JobRequest) -> None:
--- a/examples/_deployed/kitt/kitt.py
+++ b/examples/_deployed/kitt/kitt.py
@@ -1,16 +1,27 @@
 import asyncio
 import copy
 import logging
+from collections import deque
 
-from livekit import rtc
+from livekit import agents, rtc
 from livekit.agents import JobContext, JobRequest, WorkerOptions, cli
 from livekit.agents.llm import (
     ChatContext,
     ChatMessage,
     ChatRole,
 )
-from livekit.agents.voice_assistant import VoiceAssistant
+from livekit.agents.voice_assistant import AssistantContext, VoiceAssistant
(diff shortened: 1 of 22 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -54,6 +72,56 @@ async def chat(
...
+        ]
+        seen_images = set(video_frame_images)
+        (w, h) = ImageDetailsResizeValues[self._opts.image_detail]
+        encode_options = images.EncodeOptions(
+            format="JPEG",
+            resize_options=images.ResizeOptions(
+                width=w, height=h, strategy="center_aspect_fit"
+            ),
+        )
+        for img in video_frame_images:
```
davidzhao, line 90: wouldn't this reencode everything each time?

it seems like the algo would be better to:
* in ChatContext, allow users to specify the following
  * max_images
  * max_messages
* truncate context before syncing image cache
* while syncing, iterate through the entire list
* for any image not already cached, encode it
* while iterating, count how many images there has been
* truncate earlier images

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
```diff
@@ -26,3 +26,9 @@
     "gpt-3.5-turbo-1106",
     "gpt-3.5-turbo-16k-0613",
 ]
+ImageDetail = Literal["low", "medium", "high"]
+ImageDetailsResizeValues = {
```
davidzhao, line 30: OAI uses the following in their assistant API: https://platform.openai.com/docs/assistants/how-it-works/low-or-high-fidelity-image-understanding

it seems that our `high` is just matching their `low`? GPT-4o limits things to 768px

davidzhao (commented):
looks p clean!

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -0,0 +1,48 @@
...
+    images: list[Union["ChatMessageVideoFrameImage", "ChatMessageURLImage"]] = []
+
+
+@define
+class ChatContext:
+    messages: list[ChatMessage] = []
+
+
+@define
+class ChatMessageVideoFrameImage:
```
theomonnom, line 42: Not fan of the long names here what about smthg like:
```
@define
class ChatImage:
   frame: url.URL | str | rtc.VideoFrame
   detail: Literal["low", "high", "auto"]
   
```

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -27,24 +43,26 @@ def __init__(
         *,
         model: str | ChatModels = "gpt-4o",
         client: openai.AsyncClient | None = None,
+        image_detail: ImageDetail = "low",
```
theomonnom, line 46: This shouldn't in the LLM constructor but in the ChatImage 
See above (I added a `detail` field)

livekit-plugins/livekit-plugins-google/livekit/plugins/google/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.5.0"
+__version__ = "0.6.dev0"
```
theomonnom, line 15: I don't think we should bump unchanged packages

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
theomonnom: @keepingitneil when sending an image to OAI and setting the fidelity/detail field, can they resize the image on their end?

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -0,0 +1,48 @@
...
+@define
+class ChatMessage:
+    role: ChatRole
+    text: str
+    images: list[Union["ChatMessageVideoFrameImage", "ChatMessageURLImage"]] = []
+
+
+@define
+class ChatContext:
+    messages: list[ChatMessage] = []
```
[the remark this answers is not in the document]
theomonnom: Agree with that but let's think about an API design in another PR

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
davidzhao: I believe they can but we still want to resize to avoid sending a large payload. that'll increase latency

livekit-plugins/livekit-plugins-google/livekit/plugins/google/version.py
keepingitneil: I think we may want to start using >= and <= in our setup.py files instead of ~

livekit-agents/livekit/agents/llm/chat_context.py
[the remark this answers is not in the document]
keepingitneil: ^^ same. My idea for this first pass was to leave it to agent logic to manage messages and manage image context history

livekit-agents/livekit/agents/images/image.py
```diff
@@ -0,0 +1,105 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import io
+from dataclasses import dataclass
+from typing import Literal, Optional
+
+import PIL.Image
```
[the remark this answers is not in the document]
keepingitneil: yeah, would be ideal 👍

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
keepingitneil: Fixed the logic bug here. In this PR I purposely don't do any extra management of the context and leave that up to users. I'm not sure if users might want to more sophisticated treatment of images. Idk, maybe there is a use case where they always want the first image and then the last N images.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
keepingitneil: Yeah so low | medium | high is really only relevant for url images. I re-use it for sizing base64 images but choose different values to better keep the payload size in check. I'll rethink this a bit

livekit-agents/livekit/agents/llm/chat_context.py
keepingitneil: yeah I went back and forth on this. I'll go back to it. I won't have detail there though because that might be LLM specific.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
keepingitneil: I'm not sure if this is LLM specific or not. Although I do recognize that images ought to be able to have different sizing. Maybe ChatImage can simply have an LLM-agnostic width x height and OAI llm maps that to the nearest `low | medium | high`?

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -1,36 +1,18 @@
...
-
-@define
-class ChatMessage:
-    role: ChatRole
-    text: str
-
-
-@define
-class ChatContext:
-    messages: list[ChatMessage] = []
```
keepingitneil, line 27: This was using a shared list for all ChatContext instances. Moved to a field with factory in chat_context.py

examples/_deployed/kitt/kitt.py
```diff
@@ -51,10 +75,58 @@ def on_chat_received(msg: rtc.ChatMessage):
 
         asyncio.create_task(_answer_from_text(msg.message))
 
+    async def respond_to_image(user_msg: str):
```
davidzhao, line 78: would be good to have some comments in the example to walk users through the flow

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -0,0 +1,48 @@
...
+
+@dataclass
+class ChatContext:
+    messages: list[ChatMessage] = field(default_factory=list)
+
+
+@dataclass
+class ChatImage:
+    image: str | rtc.VideoFrame
+    dimensions: Tuple[int, int]
```
davidzhao, line 45: do we want a default here?

davidzhao (approved):

livekit-agents/livekit/agents/llm/chat_context.py
theomonnom: dimensions is confusing with the width and height already encoded inside rtc.VideoFrame.
We should rename dimensions to inference_dimenstions? 
@keepingitneil @davidzhao 

Also using a tuple is less explicit in situations you're not using typing. not a strong opinion but I tend to prefer having two separate fields. Not a big deal since ChatImage is a struct anyway

theomonnom: re, it makes stuff hard to read. e.g: https://github.com/livekit/agents/blob/cf3c5344958b3658d577c4c57fb909dfe41b39a7/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py#L296

theomonnom: inference_width
inference_height 

I think these may be better, also let's default to None the values

PR #337, opened 2024-06-01
plugins: add azure stt,tts

ty-elastic:
I had a need to specifically use [Azure Cognitive Speech](https://azure.microsoft.com/en-us/products/ai-services/ai-speech) services for TTS, STT. The Azure SDK is kind of interesting in that it doesn't natively support asyncio, and callbacks happen on a separate thread (which requires some careful handling of event queues).

I've tested this contribution both against the built-in tests, as well as within my own assistant app. Given that this project is essentially a "hobby", I can't really commit to maintaining this module should you significantly rev the dependent frameworks/APIs. Perhaps, though, you'll want to add and maintain it amongst your other supported drivers.

Thank you for building the agent framework - it is really well done (letting you focus on building new apps, not on "plumbing").


theomonnom:
Hey, this is awesome! I'll take a look at merging this early this week 😀🔥

davidzhao (approved):

theomonnom:
Hey, awesome @ty-elastic, I've cleaned up the API and fixed the event loop being blocked on several part of the code.
I also removed the Stream implementation of TTS, we have the concept of StreamAdapter that automatically synthesize sentence by sentence

Thanks for your contribution

PR #338, opened 2024-06-01
Switch to using OAI as default TTS in minimal_assistant

davidzhao:


theomonnom (approved):
lg, u can use `ruff check --fix` to sort imports

PR #357, opened 2024-06-06
API improvements to Vision

keepingitneil:


livekit-agents/livekit/agents/voice_assistant/assistant.py
```diff
@@ -481,13 +480,15 @@ def _on_final_transcript(self, text: str) -> None:
 
         # this speech may not be validated, so we create a copy
         # of our context to add the new user message
-        chat_ctx = copy.deepcopy(self._chat_ctx)
```
keepingitneil, line 484: opted for shallow copy because deepcopy has issues with images (i.e. rtc.VideoFrame isn't picklable and we don't want to copy those anyways)

theomonnom: Ah, but we do want a copy, otherwise the code isn't right and all the final transcription will stack up as user message in the ChatContext. 

I think we can implement a internal copy fnc for ChatContext?

examples/_deployed/kitt/kitt.py
```diff
@@ -16,6 +16,11 @@
 
 MAX_IMAGES = 3
 
+NO_IMAGE_MESSAGE_SIP = "I'm sorry, I can't see anything when you're calling from a phone. Try the web demo."
```
theomonnom, line 19: What about just removing the FunctionContext when using SIP? So we don't need this message and it simplifies the code

livekit-agents/livekit/agents/voice_assistant/assistant.py
keepingitneil: We need a deep copy?

theomonnom: There is also a deepcopy inside _answer_from_text, I think you missed it

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -27,22 +27,24 @@ class ChatRole(enum.Enum):
     TOOL = "tool"
 
 
+@dataclass
+class ChatImage:
+    image: str | rtc.VideoFrame
+    inference_width: int = 128
+    inference_height: int = 128
+    _cache: Dict[object, Any] = field(default_factory=dict)
```
theomonnom, line 35: Not fan of having private variables in a dataclass, but it is OK for now, we can iterate with a new ChatContext API.
For now let's correctly hide the field
```suggestion
    _cache: dict = field(default_factory=dict, init=False, repr=False)
```

theomonnom: I see we're not using a LLM base class for the keys. I guess the reason is to avoid import cycles. 
Not sure if moving this stuff inside a separate file is really necessary?

keepingitneil: Yeah I guess since it's only for a type I can circular import with TYPE_CHECKING check. I think the separate file has nice symmetry with function_context.py and since this is potentially still going to grow I think it makes sense.

livekit-agents/livekit/agents/voice_assistant/assistant.py
keepingitneil: Is this to prevent a footgun for users or are we mutating ChatMessages in our implementation?

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -27,22 +27,24 @@ class ChatRole(enum.Enum):
     TOOL = "tool"
 
 
+@dataclass
+class ChatImage:
+    image: str | rtc.VideoFrame
+    inference_width: int = 128
+    inference_height: int = 128
```
theomonnom, line 34: Motivation for changing to None is to allow to detect "unspecified". So we can chose strong default depending on which LLM is being used. For example Gemini doesn't need any specific size 
```suggestion
    inference_width: int | None = None
    inference_height: int | None = None
```

livekit-agents/livekit/agents/voice_assistant/assistant.py
theomonnom: We're mutating inside our impl

examples/_deployed/kitt/kitt.py
keepingitneil: Yeah makes sense.

livekit-agents/livekit/agents/llm/chat_context.py
keepingitneil: Interesting, yeah I was mostly thinking about a sane default in terms of sending images over the wire (and not thinking as much about the URL case). No strong opinion here, None is fine.

examples/_deployed/kitt/kitt.py
```diff
@@ -102,24 +102,42 @@ async def respond_to_image(user_msg: str):
...
+
+    async def get_human_video_track():
+        remote_video_tracks = [
+            t_pub.track
+            for _, p in ctx.room.participants.items()
+            for _, t_pub in p.tracks.items()
+            if t_pub.track is not None
+            and t_pub.kind == rtc.TrackKind.KIND_VIDEO
+            and isinstance(t_pub.track, rtc.RemoteVideoTrack)
+        ]
```
theomonnom, line 115: I feel like this is harder to read than just having:
- a function called `_handle_video_track(track) ` that will create a VideoStream and iterate over the frame
- register a callback for track_subscribed and call this _handle_video_track when needed 
- iterate after the assistant start over all participant and their tracks to see if smthg already match

livekit-agents/livekit/agents/voice_assistant/assistant.py
keepingitneil: Added a copy. Couldn't find any other copies that I missed...

examples/_deployed/kitt/kitt.py
keepingitneil: Hrmm, well the strategy is the same but I like having both iteration and callback in one awaitable function because:
- it's self documenting: `get_human_video_track()` IMO is easier to understand than two different code paths (iteration & callback) that spawn a process_video_track task.
- it's easy to throw in a loop that ensures only one video track is being processed at a time
- in this case, it avoids having to create a new task and cancelling tasks if new video tracks come in

I've moved it out of main entrypoint for better readability (because really it's a utility function) and removed the process_video_track function and just inlined the video track at the end.

theomonnom: no strong opinion here, either way is OK. I was more looking into this piece of code that is harder to read than a normal loop (at least for me)
Screenshot 2024-06-07 at 11 24 11 PM

theomonnom (approved):

examples/_deployed/kitt/kitt.py
keepingitneil: ahh ok fair - updated to a loop

nbsp: i feel like having a list comprehension is more idiomatic for Python than callbacks, so i'm with Neil on this, if you need a tiebreaker :)

keepingitneil: too late haha. Anyways since agents is only on python for now, we might have enough non python folks reading it so 🤷‍♂️ We'll be making this kind of stuff easier soon enough anyways!

PR #358, opened 2024-06-06
[DRAFT] CPU only torch

albertpurnama:


livekit-plugins/livekit-plugins-silero/setup.py
```diff
@@ -51,12 +51,15 @@
...
-        "torch >= 2, < 3",
-        "torchaudio >= 2",
+        "torch==2.3.1+cpu",
+        "torchaudio==2.3.1+cpu",
         "numpy >= 1, < 2",
         "onnxruntime~=1.17.0",
     ],
     package_data={"livekit.plugins.silero": ["files/silero_vad.jit"]},
+    dependency_links=[
+        "https://download.pytorch.org/whl/torch_stable.html"
```
albertpurnama, line 61: I'm currently trying to fix this. Seems like the use of dependency_list is deprecated. As mentioned in this long issue https://github.com/pypa/pip/issues/4187.

However, cpu version of torch (specifically `torch==2.3.1+cpu` and `torchaudio==2.3.1+cpu` doesn't seem to be installed through PyPi. I will try to see alternatives on PyPi to make it simpler.

keepingitneil: Tried some things out. In a requirements.txt file it would be installed like this:
```
torchaudio==2.3.1;sys_platform!='linux'
--extra-index-url https://download.pytorch.org/whl/cpu
torch==2.3.1;sys_platform=='linux'
--extra-index-url https://download.pytorch.org/whl/cpu
torchaudio==2.3.1;sys_platform=='linux'
```

Not sure how that translates to setup.py but maybe it's helpful

keepingitneil: actually ignore the sys_platform stuff, I don't think it's really needed.

keepingitneil: Edit:
```
torch==2.3.1 --index-url https://download.pytorch.org/whl/cpu
torchaudio==2.3.1 --index-url https://download.pytorch.org/whl/cpu
```
this is what I landed on for working requirements.txt. Haven't looked into how this translates to setup.py

albertpurnama: Thanks @keepingitneil , I did it with another approach, I added `-f https://download.pytorch.org/whl/torch_stable.html` in the beginning of my `requirements.txt` file and it works.

But working towards adding it to setup.py

keepingitneil: Hrmm, maybe the best solution here is to just pre-install torch and torchaudio in the Dockerfile itself...

keepingitneil: Ok so did some testing on my mac creating docker images:
```
kitt-torchcuda                latest    8fa9a21894cd   6 seconds ago   1.23GB
kitt-torchcpu-nosilero        latest    5ab95344eefd   2 minutes ago   1.21GB
kitt-without-silero           latest    2ce8716a7544   4 minutes ago   629MB
kitt-without-requirementstxt  latest    90ef8f62379b   6 minutes ago   453MB
kitt                          latest    e7eaf61cc045   6 minutes ago   1.19GB
kitt-torchcpu                 latest    119601aabe5c   8 minutes ago   1.29GB
```

IMO these values are all reasonable. The question is: how did you end up with an 8GB build?! 🤔

albertpurnama: Will share a repo for reproduction soon @keepingitneil

andrewjhogue:
@keepingitneil @albertpurnama +1 on this - am also seeing a 5GB+ build when deploying to Docker.

In case it helps with triage - resulting size differed quite a bit between builds for amd64 vs x86-64 (eg > 5GB vs ~1GB).

theomonnom:
I think we can completely remove the torch dependency and use onnxruntime by default?

albertpurnama:
@andrewjhogue I can confirm that building with `--platform linux/amd64` is 5GB and building for mac system (default without the `--platform` is ~1GB

Command I use to run default ~1GB build is `docker build -t your-img-name:latest .`

While the command I use to run the linux/amd64 build is `docker buildx build --platform linux/amd64 -t your-img-name:latest .

theomonnom:
Hey, I started some work here https://github.com/livekit/agents/pull/380

theomonnom:
Closing this PR in favour of the new one. I should be able to merge it in the following days

albertpurnama:
Hey @theomonnom thanks! awesome work!

PR #360, opened 2024-06-07
Update stt.py support custom openai_base_url

JARVISMindEngineer:


davidzhao:
implemented in #364, thanks for the suggestion and PR!

PR #364, opened 2024-06-07
Allow custom base_url to be used with OpenAI plugins

davidzhao:
Related to #359 #360

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -31,6 +31,7 @@ import openai
 
 from .log import logger
 from .models import ChatModels
+from .utils import get_base_url
 
 IMAGE_DETAIL_DIMENSIONS: List[Tuple[int, str]] = [
     (512, "low"),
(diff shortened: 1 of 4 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -49,10 +49,11 @@ def __init__(
         self,
         *,
         model: str | ChatModels = "gpt-4o",
+        base_url: str | None = None,
         client: openai.AsyncClient | None = None,
     ) -> None:
         self._opts = LLMOptions(model=model)
-        self._client = client or openai.AsyncClient()
+        self._client = client or openai.AsyncClient(base_url=base_url)
```
theomonnom, line 56: get_base_url here too for the env var?

theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
davidzhao: AsyncClient handles it internally

davidzhao: though probably good to make it consistent in case they change behavior

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py
```diff
@@ -0,0 +1,7 @@
+import os
```
theomonnom, line 1:
```suggestion
from __future__ import annotations

import os
```

davidzhao: ah i just started using `Optional`.. will keep in mind in the future. def seems easier

PR #370, opened 2024-06-11
add more context to logs

theomonnom:
- extra is shown by default in both, dev and prod
- removed colorlog and python-json-logger dependencies (current code is based on them)

### DIFF
--- a/livekit-agents/livekit/agents/cli/log.py
+++ b/livekit-agents/livekit/agents/cli/log.py
@@ -1,54 +1,206 @@
+from __future__ import annotations
+
+import json
 import logging
+import re
+import traceback
+from collections import OrderedDict
+from datetime import date, datetime, time, timezone
+from inspect import istraceback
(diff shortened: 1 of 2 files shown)


theomonnom:
agent_identity will be in another PR

davidzhao (approved):
lg!

should some of this be in the python SDK? seems like we might want the json/colored loggers in multiple places

theomonnom:
> lg!
> 
> should some of this be in the python SDK? seems like we might want the json/colored loggers in multiple places

I think python-sdks shouldn't be too opinionated. In agents we're doing it but I'm going to allow to bypass these configurations.

davidzhao:
> I think python-sdks shouldn't be too opinionated. In agents we're doing it but I'm going to allow to bypass these configurations.

I think practically, users would just want sane defaults for everything. it'd be great to standardize things like logging on the python stack. but we that doesn't have to block this PR

PR #371, opened 2024-06-11
add cartesia tts

theomonnom:
further improvement would be to use websockets (advantage is to have a prewarmed connection for the first speech)


davidzhao (approved):
lgtm!

PR #378, opened 2024-06-12
remove attrs dependency

theomonnom:
We only use attrs for the kw_only feature because Python 3.9 doesn't support it (3.10+)
It isn't that important..

### DIFF
--- a/livekit-agents/livekit/agents/ipc/protocol.py
+++ b/livekit-agents/livekit/agents/ipc/protocol.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import io
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import ClassVar
 
 from livekit.protocol import agent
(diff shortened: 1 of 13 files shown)


livekit-agents/livekit/agents/ipc/protocol.py
```diff
@@ -19,7 +19,7 @@ class JobMainArgs:
     asyncio_debug: bool
 
 
-@define(kw_only=True)
+@dataclass
 class StartJobRequest:
     MSG_ID: ClassVar[int] = 0
     job: agent.Job = agent.Job()
```
keepingitneil, line 25: replace with factory?

davidzhao (approved):

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -1,27 +1,26 @@
...
+@dataclass
 class Choice:
     delta: ChoiceDelta
     index: int = 0
 
 
-@define
+@dataclass
 class ChatChunk:
     choices: list[Choice] = []
```
keepingitneil, line 25: replace with factory

theomonnom: yep good catch

PR #380, opened 2024-06-14
rework silero vad

theomonnom:
- remove pytorch dependency
   - we only use onnxruntime now (avg size of onnxruntime is 7MB, silero model is 1.5MB)
- use right window sizes (enforce the same one used while training the model)
- allow 16Khz*X sample rates (not technically right, we sample each sr//16000 step)
- fixed warning when loading the silero model
- the model is now embedded inside the pypi package
- move vad paramters to silero.VAD constructor
- fix duplicate final transcripts on the stt.StreamAdapter 

EDIT: the tests should now be fixed

### DIFF
--- a/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/onnx_model.py
+++ b/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/onnx_model.py
@@ -1,0 +1,69 @@
+import asyncio
+import importlib.resources
+import atexit
+import onnxruntime
+import numpy as np
+
+from typing import Any
+from contextlib import ExitStack
+from .log import logger
+
+_resource_files = ExitStack()
+atexit.register(_resource_files.close)
+
+
+SUPPORTED_SAMPLE_RATES = [8000, 16000]
+
+
+def new_inference_session(force_cpu: bool) -> onnxruntime.InferenceSession:
+    res = (
+        importlib.resources.files("livekit.plugins.silero.resources")
+        / "silero_vad.onnx"
+    )
+    ctx = importlib.resources.as_file(res)
+    path = _resource_files.enter_context(ctx)
+
+    opts = onnxruntime.SessionOptions()
+    opts.inter_op_num_threads = 1
+    opts.intra_op_num_threads = 1
+    opts.graph_optimization_level = onnxruntime.GraphOptimizationLevel.ORT_ENABLE_ALL
+
+    if force_cpu and "CPUExecutionProvider" in onnxruntime.get_available_providers():
+        session = onnxruntime.InferenceSession(
+            str(path), providers=["CPUExecutionProvider"], ess_options=opts
+        )
+    else:
+        session = onnxruntime.InferenceSession(str(path), sess_options=opts)
+
+    return session
+
+
+class OnnxModel:
+    def __init__(
+        self, *, onnx_session: onnxruntime.InferenceSession, sample_rate: int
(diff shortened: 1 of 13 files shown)


davidzhao (commented):
this lgtm, this should help get the image size to be much smaller

theomonnom:
I still need to plot the accuracy before merging

theomonnom:
Note that I'm using a default window size of 1024 at 16Khz (roughly 60ms)

livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/onnx_model.py
```diff
@@ -0,0 +1,66 @@
...
+import onnxruntime  # type: ignore
+
+_resource_files = ExitStack()
+atexit.register(_resource_files.close)
+
+
+SUPPORTED_SAMPLE_RATES = [8000, 16000]
+
+
+def new_inference_session(force_cpu: bool) -> onnxruntime.InferenceSession:
```
keepingitneil, line 15: Should this be async? I think loading the weights for the first time takes a while

tests/test_vad.py
keepingitneil, line 1: ?

theomonnom: I may write another PR for that since I'm late on plans

livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/onnx_model.py
theomonnom: I don't think we should, prewarming stuff will allow blocking ops

keepingitneil (approved):

tests/test_stt.py
```diff
@@ -35,7 +35,7 @@ def read_mp3_file(filename: str) -> rtc.AudioFrame:
     return agents.utils.merge_frames(frames)
 
 
-RECOGNIZE_STT = [deepgram.STT(), google.STT(), openai.STT(), azure.STT()]
+RECOGNIZE_STT = [deepgram.STT(), google.STT(), openai.STT()]
```
davidzhao, line 38: we removed azure STT from tests?

theomonnom: Only for the recognize method that we don't support. We only support streaming for it and it is correctly being testing at the end of the file

davidzhao: got it. interesting they only support streaming mode.

theomonnom: They do support it: https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/d7e66d66850ecee97d4e8c01950a100b441364d7/samples/python/console/speech_sample.py#L42

Fwiw recognize method is never used when streaming is supported. I will add a ticket for it but this is low priority

livekit-plugins/livekit-plugins-silero/setup.py
```diff
@@ -49,12 +49,11 @@
     python_requires=">=3.9.0",
     install_requires=[
         "livekit-agents~=0.7",
-        "torch >= 2, < 3",
-        "torchaudio >= 2",
-        "numpy >= 1, < 2",
-        "onnxruntime~=1.17.0",
+        "onnxruntime~=1.18",
+        "numpy~=1.26",
```
hauntsaninja, line 53: This prevents installation of numpy>=2, effectively reverting https://github.com/livekit/agents/pull/398/files. Was this pin intentional?

PR #384, opened 2024-06-18
Fix token on job reload in dev mode

keepingitneil:
This bug prevented reloaded agents from publishing transcriptions and from updating room metadata. 

Cause: The SFU does more than set video grants when generating tokens: https://github.com/livekit/protocol/blob/main/agent/token.go and it has apis that are not available in the client library. (i.e. SetKind)

Fix: In reloaded jobs now we simply reuse the previous token and extend its expiration.

### DIFF
--- a/livekit-agents/livekit/agents/cli/protocol.py
+++ b/livekit-agents/livekit/agents/cli/protocol.py
@@ -47,6 +47,7 @@ class ActiveJobsResponse:
             ipc_enc._write_bytes(b, job_s)
             accept_s = pickle.dumps(aj.accept_data)
             ipc_enc._write_bytes(b, accept_s)
+            ipc_enc._write_string(b, aj.token)
 
     def read(self, b: io.BytesIO) -> None:
         job_count = ipc_enc._read_int(b)
@@ -56,7 +57,8 @@ class ActiveJobsResponse:
             job.ParseFromString(job_s)
             accept_s = ipc_enc._read_bytes(b)
             accept_data = pickle.loads(accept_s)
-            self.jobs.append(ActiveJob(job=job, accept_data=accept_data))
+            token = ipc_enc._read_string(b)
+            self.jobs.append(ActiveJob(job=job, accept_data=accept_data, token=token))
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/worker.py
```diff
@@ -263,7 +265,12 @@ async def _worker_task(self) -> None:
                 if scheme.startswith("http"):
                     scheme = scheme.replace("http", "ws")
 
-                agent_url = f"{scheme}://{parse.netloc}/{parse.path.rstrip('/')}/agent"
+                path_parts = [
```
keepingitneil, line 268: This is unrelated but I was seeing `ws://localhost:7880//agent` in the logs so it's a small fix

livekit-agents/livekit/agents/cli/protocol.py
```diff
@@ -56,7 +57,8 @@ def read(self, b: io.BytesIO) -> None:
             job.ParseFromString(job_s)
             accept_s = ipc_enc._read_bytes(b)
             accept_data = pickle.loads(accept_s)
-            self.jobs.append(ActiveJob(job=job, accept_data=accept_data))
+            token = ipc_enc._read_string(b)
```
keepingitneil, line 60: @theomonnom please verify my pickle logic. I didn't do a deep dive into the ipc_enc workings.

theomonnom:
I think we could add https://github.com/livekit/protocol/blob/32c577d805ed1c885a4fa3fff3fc37e10f920111/auth/accesstoken.go#L60 to our Python SDK on livekit-api? wdyt?

keepingitneil:
> I think we could add [SetParticipantKind](https://github.com/livekit/protocol/blob/32c577d805ed1c885a4fa3fff3fc37e10f920111/auth/accesstoken.go#L60) to our Python SDK on livekit-api? wdyt?

I'm not sure. I wasn't sure of the reason we rely on kind in the SFU even though video grants has an agent property. But generally, I kind of like just having the SFU be the source of truth for token generation. It's reasonable to assume that there will be more/different permissions in the future that breaks this again. 

We should probably refresh the token periodically but I'm leaning towards reusing SFU generated tokens. Another approach, since the agent has the API Key/Secret. Is just copying the fields of the old jwt into a new one with a new ttl. Wdyt?

theomonnom:
> We should probably refresh the token periodically but I'm leaning towards reusing SFU generated tokens. Another approach, since the agent has the API Key/Secret. Is just copying the fields of the old jwt into a new one with a new ttl. Wdyt?
Copying the JWT data into a new one seems like the best solution in this case.

theomonnom (approved):

PR #401, opened 2024-06-30
Add example docker file for building agents

davidzhao:


nbsp (approved):
unfamiliar with Docker but this looks okay

PR #409, opened 2024-07-03
use changesets/changesets for versioning

nbsp:


scripts/bump_version
theomonnom, line 1: Can you move this file to .github

theomonnom:
Not familiar with this CI, can it generate a CHANGELOG file for every package we have?

livekit-plugins/livekit-plugins-elevenlabs/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "@livekit/agents-py-plugin-elevenlabs",
```
theomonnom, line 2: In the best scenario, this is only used as a placeholder, but nothing is really published to npm

theomonnom (approved):
otherwise lgtm!

livekit-plugins/livekit-plugins-elevenlabs/package.json
nbsp: yep it's just because package.json's schema requires a name. it will do nothing

livekit-agents/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "@livekit/agents-py",
```
lukasIO, line 2: it probably makes sense to use the same name here as the pypi package has. 
The GH release notes will use this name and it would feel less confusing that way?

lukasIO: it probably makes sense to use the same name here as the pypi package has.
The GH release notes will use this name and it would feel less confusing that way?

livekit-plugins/livekit-plugins-elevenlabs/package.json
lukasIO: I still think we should change the name, see my comment above

.github/workflows/publish-package.yml
```diff
@@ -74,6 +74,27 @@ jobs:
         name: python-package-distributions
         path: dist/
 
+    - uses: pnpm/action-setup@v2
+      with:
+        version: 9
```
lukasIO, line 79: because we're specifying the pnpm version in the root package.json we have to remove the version number here, otherwise the pipeline will fail

theomonnom:
can you add livekit-plugins-rag and livekit-plugins-minimal before merging? 🙏

livekit-plugins/livekit-plugins-rag/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "livekit-plugins/livekit-plugins-rag",
+  "private": true,
+  "version": "0.6.0"
```
theomonnom, line 4: ah, mistake from my side, this should be 0.1.0

nbsp:
> can you add livekit-plugins-rag and livekit-plugins-minimal before merging? 🙏

not merging until Lukas gets another look at this

livekit-agents/livekit/agents/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.7.2"
+__version__ = "0.0.0"
```
theomonnom, line 15: is this a mistake?

nbsp: oh oops yeah this is from testing

livekit-plugins/livekit-plugins-cartesia/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "livekit-plugins/livekit-plugins-cartesia",
```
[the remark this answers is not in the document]
nbsp: this is because we don't know which packages are and which aren't under livekit-plugins unless we do this. it's easier to strip it in the CI than to figure out whether we need to add it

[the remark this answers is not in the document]
nbsp: it's used programmatically to know where to cd in the CI

.github/workflows/publish-package.yml
```diff
@@ -29,11 +29,41 @@ permissions:
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  
+  build:
+    needs:
+    - bump
+    strategy:
+      matrix:
+        package: ${{ needs.bump.outputs.packages }}
```
lukasIO, line 63: I'm guessing we need to parse the string to JSON for the matrix by wrapping it in `fromJSON`

livekit-plugins/livekit-plugins-cartesia/package.json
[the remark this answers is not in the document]
lukasIO: maybe we can add a comment either in the publish-package.yml (or in the README) about what naming format we expect/need for this logic to work. Just as a reference once we add more lateron.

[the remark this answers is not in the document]
nbsp: maybe in the README for livekit-plugins-minimal, since that might be copied over as boilerplate for new packages?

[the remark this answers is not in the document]
lukasIO: sounds good, the root README sounds also appropriate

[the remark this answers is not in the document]
nbsp: having trouble fitting this anywhere in the root README, it feels too end-user focused and this is internal dev doc
i think the plugin part is good, i don't assume we'll add anything to anywhere but the plugin directory in this repository

.github/workflows/publish-package.yml
```diff
@@ -29,11 +29,41 @@ permissions:
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  
+  build:
+    needs:
+    - bump
+    strategy:
+      matrix:
+        package: ${{ fromJson(needs.bump.outputs.packages) }}
```
lukasIO, line 63: not sure if this will work, the reference calls for JSON to be capitalized
```suggestion
        package: ${{ fromJSON(needs.bump.outputs.packages) }}
```

lukasIO: https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson

nbsp: no it will, this is the canonical name for it actually, github documentation just doesn't use it

lukasIO (approved):

theomonnom:
Is it possible to push dev versions with it?
I want to bump livekit-agents to v0.8.0.dev0

PR #412, opened 2024-07-04
Support speed param when calling openai TTS

minhpq331:
This PR adds support for speed param when calling OpenAI TTS

Ref: https://platform.openai.com/docs/api-reference/audio/createSpeech#audio-createspeech-speed


theomonnom (approved):
Thanks!

PR #418, opened 2024-07-05
stt capabilities

theomonnom:


livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -2,26 +2,29 @@
...
+    if the STT doesn't support this event, this will be emitted as the same time as the first INTERIM_TRANSCRIPT"""
+    INTERIM_TRANSCRIPT = "interim_transcript"
+    """interim transcript, useful for real-time transcription"""
+    FINAL_TRANSCRIPT = "final_transcript"
+    """inal transcript, emitted when the STT is confident enough that a certain
+    portion of speech will not change"""
+    END_OF_SPEECH = "end_of_speech"
+    """indicate the end of speech, emitted when the user stops speaking
+    the first alternative is a combination of all the previous FINAL_TRANSCRIPT events"""
+    END_OF_UTTERANCE = "end_of_utterance"
```
davidzhao, line 26: what's the difference between `end_of_utterance` and `end_of_speech`? From DeepGram's docs, it seems like they are the same thing: https://developers.deepgram.com/docs/utterance-end

theomonnom: Deepgram doesn't have end_of_speech (We simulate it inside our integration) using the final_transcripts (endpoints). The endpoints are based on a VAD

end of utterance isn't based on VAD. So even if there is a lot of noise or inaccuracy on their VAD. It'll still be fired

theomonnom: end_of_speech is here for consistency with other STTs

davidzhao: got it, so what is the benefit of adding another event? vs using the current approach of translating end of utterance to `end_of_speech` event?

theomonnom: end_of_speech event will certainly be called sooner than end_of_utterance as the STT may do more checks (wait more or some NLP based detection to see if a sentence is finished)

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -39,13 +39,23 @@ class SpeechEvent:
     alternatives: List[SpeechData] = field(default_factory=list)
 
 
+@dataclass
+class STTCapabilities:
+    streaming_supported: bool
```
davidzhao, line 44:
```suggestion
    streaming: bool
```
removing _supported since everything in capabilities are supported

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -39,13 +39,23 @@ class SpeechEvent:
...
+    interim_results: bool
+
+
 class STT(ABC):
```
davidzhao, line 48: I sometimes wonder if our STT should be an abstract superclass, so stuff like this could be more easily inherited.

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -66,9 +76,11 @@ def stream(
             a different STT or use a StreamAdapter"
         )
 
-    @property
-    def streaming_supported(self) -> bool:
```
davidzhao, line 70: might be nice to keep this.. and just return `.capabilities.streaming`

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -39,13 +39,23 @@ class SpeechEvent:
...
+    streaming_supported: bool
+    interim_results: bool
+
+
 class STT(ABC):
     def __init__(
         self,
         *,
-        streaming_supported: bool,
+        capabilities: STTCapabilities,
```
davidzhao, line 52: should this be required? that'll break all existing plugins?

nbsp: i think we should give at least one version for deprecating this and still keeping the API open. maybe throw a warn

PR #419, opened 2024-07-05
plugins: add AssemblyAI STT plugin

cch41:
I've tested this contribution both against the built-in tests, as well as within our Assort Health voice assistant app. Every so often it struggles with picking up short utterances, but that may have more to do with its integration specifically with our app. `tests/test_stt` test results:
Screenshot 2024-07-05 at 4 12 07 PM

We're upstreaming so that others can use our AssemblyAI plugin, however we can't commit right now to maintaining it.

Two notes: The below test failures are unrelated, and this just implements streaming

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -1,0 +1,408 @@
+# Copyright 2023 LiveKit, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from __future__ import annotations
+
+import asyncio
+import dataclasses
+import json
+import os
+import time
+from contextlib import suppress
+from dataclasses import dataclass
+from typing import List, Optional, Union
+from urllib.parse import urlencode
+
+import aiohttp
+from livekit import rtc
+from livekit.agents import stt, utils
+from livekit.agents.utils import AudioBuffer
+
+
+ENGLISH = "en"
+
+
+@dataclass
+class STTOptions:
+    sample_rate: int | None = None
+    word_boost: str | None = None
+    encoding: str | None = None  # Allowed values: pcm_s16le, pcm_mulaw
+    disable_partial_transcripts: bool = False
+    enable_extra_session_information: bool = False
+    end_utterance_silence_threshold: int = None
+    buffer_size_seconds: int = None
+    token_expires_in: int = None
+
+
+class STT(stt.STT):
+    def __init__(
+        self,
+        *,
+        api_key: str | None = None,
+        sample_rate: int | None = 16000,
+        word_boost: str | None = None,
+        encoding: str | None = "pcm_s16le",
+        disable_partial_transcripts: bool = False,
+        enable_extra_session_information: bool = False,
+        end_utterance_silence_threshold: int = 1000,
+        http_session: aiohttp.ClientSession | None = None,
+        token_expires_in: int = 3600,
+        buffer_size_seconds: int = 0.2,
+    ):
+        super().__init__(streaming_supported=True)
+        api_key = api_key or os.environ.get("ASSEMBLYAI_API_KEY")
+        if api_key is None:
+            raise ValueError("Assembly AI API key is required")
+        self._api_key = api_key
+
+        self._opts = STTOptions(
+            sample_rate=sample_rate,
+            word_boost=word_boost,
+            encoding=encoding,
+            disable_partial_transcripts=disable_partial_transcripts,
+            enable_extra_session_information=enable_extra_session_information,
+            buffer_size_seconds=buffer_size_seconds,
+            token_expires_in=token_expires_in,
+            end_utterance_silence_threshold=end_utterance_silence_threshold,
+        )
+        self._session = http_session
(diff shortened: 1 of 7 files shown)


livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,408 @@
...
+        buffer: AudioBuffer,
+    ) -> stt.SpeechEvent:
+        raise NotImplementedError("Not implemented")
+
+    def stream(
+        self,
+        *,
+        language: str | None = None,
+    ) -> "SpeechStream":
+        config = dataclasses.replace(self._opts)
```
oconnoob, line 101: Why not just directly pass in `self._opts` here? It looks like `SpeechStream` doesn't modify this object

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,408 @@
...
+        token_expires_in: int = 3600,
+        buffer_size_seconds: int = 0.2,
+    ):
+        super().__init__(streaming_supported=True)
+        api_key = api_key or os.environ.get("ASSEMBLYAI_API_KEY")
+        if api_key is None:
+            raise ValueError("Assembly AI API key is required")
+        self._api_key = api_key
+
+        self._opts = STTOptions(
```
oconnoob, line 70: Wondering why you don't use dependency injection here as with `SpeechStream`?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,408 @@
...
+            disable_partial_transcripts=disable_partial_transcripts,
+            enable_extra_session_information=enable_extra_session_information,
+            buffer_size_seconds=buffer_size_seconds,
+            token_expires_in=token_expires_in,
+            end_utterance_silence_threshold=end_utterance_silence_threshold,
+        )
+        self._session = http_session
+        self._token_expires_in = token_expires_in
+
+    def _ensure_session(self) -> aiohttp.ClientSession:
```
oconnoob, line 83: In the future might make sense to just make this a property and set `self._session = utils.http_session()` in the getter

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,408 @@
...
+        when something went wrong.
+        """
+        if self._token is None or time.time() > self._token_expires_at:
+            await self._get_temporary_token()
+
+        try:
+            retry_count = 0
+            while not self._closed:
+                try:
+                    live_config = {
```
oconnoob, line 198: can we pull this definition out of the while loop?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,408 @@
...
+        async def send_task():
+            nonlocal closing_ws
+            # forward inputs to Assembly AI
+            # if we receive a close message, signal it to Assembly AI and break.
+            # the recv task will then make sure to process the remaining audio and stop
+            while True:
+                data = await self._queue.get()
+                self._queue.task_done()
+
+                if isinstance(data, rtc.AudioFrame):
```
oconnoob, line 250: worth adding an `else` case here to raise an Exception that there was an unexpected data type?

cch41:
Thanks @oconnoob for the comments, I should be able to get to them mid week

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,408 @@
...
+    token_expires_in: int = None
+
+
+class STT(stt.STT):
+    def __init__(
+        self,
+        *,
+        api_key: str | None = None,
+        sample_rate: int | None = 16000,
+        word_boost: str | None = None,
```
ploeber, line 55: I assume this is what the user sets, right? In that case I'd make it a `List[str] = []`. According to our docs, *"The parameter value must be a JSON encoded array of strings."*

In our SDK the user can set e.g. `word_boost = ["foo", "bar"]`, and under the hood we do:
`params["word_boost"] = json.dumps(self._word_boost)`, and later urlencode the params. Compare with Python SDK code: https://github.com/AssemblyAI/assemblyai-python-sdk/blob/master/assemblyai/transcriber.py#L1094.

A list will be easier for the user experience, otherwise they have to ensure this is already in the correct string format, e.g `'["foo", "bar"]'`

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,408 @@
...
+        enable_extra_session_information: bool = False,
+        end_utterance_silence_threshold: int = 1000,
+        http_session: aiohttp.ClientSession | None = None,
+        token_expires_in: int = 3600,
+        buffer_size_seconds: int = 0.2,
+    ):
+        super().__init__(streaming_supported=True)
+        api_key = api_key or os.environ.get("ASSEMBLYAI_API_KEY")
+        if api_key is None:
+            raise ValueError("Assembly AI API key is required")
```
ploeber, line 67: Nit pick: Should we change all strings to "AssemblyAI" instead of "Assembly AI" ? AssemblyAI is the correct spelling

ploeber (changes requested):
Hi, Patrick here from AssemblyAI. I also left some small comments, but the rest looks good! Thank you so much for working on this 🙏

ploeber (changes requested):
I did a second code review with some more improvements. I was told you're having troubles sending small chunks so hopefully this small logic fix to calculate the duration can help send chunks earlier.

theomonnom:
Hey, this pull request was automatically closed when we deleted the dev branch.
Can you reopen this PR against the main branch. We're looking forward to merge this PR

PR #422, opened 2024-07-07
Update README.md simulate job command

vanics:
The command to connect to an existing room has changed to `connect` but the documentation is still referring to the old command `simulate-job`.


theomonnom (approved):
ah! Thanks :)

PR #423, opened 2024-07-07
move aio to utils

theomonnom:


theomonnom:
The root module should be easy to understand (less is better here)

PR #427, opened 2024-07-08
fix types & lint

keepingitneil:


livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -471,21 +478,18 @@ async def _play_speech(self, speech_info: _SpeechInfo) -> None:
...
                     continue
 
-                tool_calls.append(called_fnc.call_info)
                 tool_calls_results.append(
-                    llm.ChatMessage.create_tool_from_called_function(called_fnc)
+                    ChatMessage.create_tool_from_called_function(called_fnc)
                 )
 
             chat_ctx = speech_info.source.chat_ctx.copy()
-            chat_ctx.messages.extend(tool_calls)
```
keepingitneil, line 488: @theomonnom Was this correct? Types were wrong

livekit-agents/livekit/agents/stt/stream_adapter.py
```diff
@@ -24,7 +25,9 @@ def __init__(
         stt: STT,
         vad: VAD,
     ) -> None:
-        super().__init__(streaming_supported=True)
+        super().__init__(
+            capabilities=STTCapabilities(streaming=True, interim_results=False)
```
davidzhao, line 29: should this reflect the capabilities of the thing it's wrapping? i.e. copy(stt.capabilities.copy) but update `.streaming=True`

keepingitneil: yeah makes sense. interim_results will always be False in the thing it's wrapping (I think) but anyways I think it's cleaner to do the copy. Will fix

davidzhao: sg, yeah I guess it makes no difference today.. but might make a diff when more capabilities are added.

theomonnom: the StreamAdapter *upgrades* the capability of a specific STT. 
If you want to get the real capabilities of the STT, you can use the wrapped_stt property and retrieve capabilities

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -533,8 +537,15 @@ def __init__(
...
-        self._last_final_transcript = user_msg.content
+    def on_new_synthesis(self, user_msg: ChatMessage) -> None:
+        str_content: str = ""
+        if isinstance(user_msg.content, str):
+            str_content = user_msg.content
+        elif isinstance(user_msg.content, list):
+            str_items = [item for item in user_msg.content if isinstance(item, str)]
+            str_content = " ".join(str_items)
+
+        self._last_final_transcript = str_content
```
theomonnom, line 549: We can assume it is always str here, I removed the code

theomonnom:
tests aren't passing for llm, let's do that in another PR. The reason is that the API changed

theomonnom (approved):
lgtm, thx
just fixed the stuff I just pushed

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
theomonnom: ah, fixed it thx.

PR #429, opened 2024-07-09
Fix llm test

keepingitneil:


theomonnom (approved):

tests/test_llm.py
```diff
@@ -159,6 +159,12 @@ async def test_cancelled_calls():
     stream = await _request_fnc_call(
         llm, "Turn off the lights in the Theo's bedroom", fnc_ctx
     )
+    stream.execute_functions()
+
+    # TODO (?): need to wait for the task to be scheduled but this might be
+    # a design flaw. Users would not be able to respond to cancelled tasks
+    # if they are never started.
```
theomonnom, line 166: I think we should be able to get the specific tasks handle when iterating over the stream

tests/test_llm.py
```diff
@@ -159,6 +159,12 @@ async def test_cancelled_calls():
...
+    await asyncio.sleep(0.1)
```
theomonnom, line 167:
```suggestion
    await asyncio.sleep(0)
```

tests/test_llm.py
keepingitneil: Oh yes true, maybe my comment can be updated. Really what it means is that it's possible for tasks to never start but idk if it's an issue 🤔

davidzhao:
nice! back to greeeen

PR #434, opened 2024-07-10
[Proposed API] Introduce a RetryTTS wrapper

keepingitneil:
Introduces a TTS Implementation that wraps a provided TTS implemementation but adds a retry policy.


vanics:
I love this!

PR #443, opened 2024-07-17
enable changesets ci

theomonnom:


keepingitneil (approved):

nbsp (commented):
looks good, but i'm unsure how to handle prereleases within this.
maybe we can make it run on a `dev` branch as well, and make it run an additional step there where it goes `pnpm exec changeset pre set dev`? look at the changeset docs for exact instructions

PR #450, opened 2024-07-18
Allow on demand dev releases

lukasIO:


.github/workflows/publish-package.yml
```diff
@@ -84,6 +84,7 @@ jobs:
       - name: Bump version.py
         run: |
           export module=$(sed 's/^.*\///' <<< ${{ matrix.package.name }} | tr '-' '/')
+          export version=$(sed -E 's|-(\d*)|${{  github.ref == 'refs/heads/main' && '' || '-dev' }}\1|g' <<< ${{ matrix.package.version }})
```
nbsp, line 87: `pnpm ci:version --snapshot dev` already produces `x.y.z-dev-TIMESTAMP`, there's no need to add another `-dev` to it, just trim down the last dash.

nbsp (changes requested):

.github/workflows/publish-package.yml
lukasIO: weirdly enough I tried that and then the output when running the command locally is wrong. So I had to add it in order to get the correct output of `x.y.z-devTIMESTAMP` for pypi

lukasIO: ah, thanks! will do

lukasIO: pushed the change, it looks a bit ugly as I read that nested expressions aren't supported. If you were thinking of a more elegant solution feel free to commit it to this branch!

nbsp: i think this one is clearer: `'s|-\([0-9]*\)$|\1|'` (take dash-numbers-eol and turn it into numbers-eol)
also note that sed regexes require `\(` `\)` for grouping, with literal backslashes

lukasIO: yeah, that seems cleaner! 
the `-E` argument allows for braces not to be escaped, but can go with the regular version of escaping it.

theomonnom: Can pypi recognize `v0.1.0-dev` too? I've only used `v0.1.0.dev0`

theomonnom: https://peps.python.org/pep-0440/
> Epoch segment: N!
> Release segment: N(.N)*
> Pre-release segment: {a|b|rc}N
> Post-release segment: .postN
> Development release segment: .devN
>

nbsp: the timestamp is only numbers, so even though it doesn't follow the "spirit" of the spec, i think it technically abides by it. version numbers aren't consecutive but they are ordered

theomonnom (approved):
Just did some testing with the official regex:

```
>>> import re
>>> VERSION_PATTERN = r"""
...     v?
...     (?:
...         (?:(?P<epoch>[0-9]+)!)?                           # epoch
...         (?P<release>[0-9]+(?:\.[0-9]+)*)                  # release segment
...         (?P<pre>                                          # pre-release
...             [-_\.]?
...             (?P<pre_l>(a|b|c|rc|alpha|beta|pre|preview))
...             [-_\.]?
...             (?P<pre_n>[0-9]+)?
...         )?
...         (?P<post>                                         # post release
...             (?:-(?P<post_n1>[0-9]+))
...             |
...             (?:
...                 [-_\.]?
...                 (?P<post_l>post|rev|r)
...                 [-_\.]?
...                 (?P<post_n2>[0-9]+)?
...             )
...         )?
...         (?P<dev>                                          # dev release
...             [-_\.]?
...             (?P<dev_l>dev)
...             [-_\.]?
...             (?P<dev_n>[0-9]+)?
...         )?
...     )
...     (?:\+(?P<local>[a-z0-9]+(?:[-_\.][a-z0-9]+)*))?       # local version
... """
>>> 
>>> _regex = re.compile(
...     r"^\s*" + VERSION_PATTERN + r"\s*$",
...     re.VERBOSE | re.IGNORECASE,
... )
>>> 
>>> 
>>> _regex.match("v0.1.0.dev0")
<re.Match object; span=(0, 11), match='v0.1.0.dev0'>
>>> _regex.match("v0.1.0.dev-0")
<re.Match object; span=(0, 12), match='v0.1.0.dev-0'>
>>> _regex.match("v0.1.0.dev-1")
<re.Match object; span=(0, 12), match='v0.1.0.dev-1'>
>>> _regex.match("v0.1.0.dev-143535365346")
<re.Match object; span=(0, 23), match='v0.1.0.dev-143535365346'>
>>> _regex.match("v0.1.0-dev-143535365346")
<re.Match object; span=(0, 23), match='v0.1.0-dev-143535365346'>
>>> _regex.match("v0.1.0-dev-153562")
<re.Match object; span=(0, 17), match='v0.1.0-dev-153562'>

```
Seems like the current format is OK, lgtm! Thanks!

theomonnom:
Note that this one is not valid tho:
`_regex.match("v0.1.0-dev1-1234532")`

nbsp (approved):

lukasIO:
should I merge this into `main` or into `dev` ? 👀 
In the end we'll need it in both anyways as the action won't show up in the actions tab if it's not present in the default branch of the repo, just wondering about the ordering mostly.

theomonnom:
merging to main lgtm

PR #458, opened 2024-07-21
use new procpool api 

theomonnom:


examples/voice-assistant/minimal_assistant.py
```diff
@@ -29,10 +29,10 @@ async def entrypoint(ctx: JobContext):
...
-async def request_fnc(req: JobRequest) -> None:
+async def job_request_fnc(req: JobRequest) -> None:
     logging.info("received request %s", req)
-    await req.accept(entrypoint)
+    await req.accept()
 
 
 if __name__ == "__main__":
-    cli.run_app(WorkerOptions(request_fnc))
+    cli.run_app(WorkerOptions(job_request_fnc=job_request_fnc, job_entrypoint_fnc=entrypoint))
```
davidzhao, line 38: having `job_` prefix seems redundant? also, could it be left blank to auto accept?

```suggestion
    cli.run_app(WorkerOptions(entrypoint_fnc=entrypoint))
```

livekit-agents/livekit/agents/job.py
```diff
@@ -44,8 +43,8 @@ class JobAcceptArguments:
 
 @dataclass
 class RunningJobInfo:
+    accept_arguments: JobAcceptArguments
```
davidzhao, line 46: what's the thinking in making these longer?

examples/voice-assistant/minimal_assistant.py
theomonnom: we can discuss more about it tmrw

livekit-agents/livekit/agents/job.py
theomonnom: this was public but not exposed to the users

PR #465, opened 2024-07-22
openai: reduce log noise

nbsp:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -56,6 +57,7 @@ def __init__(
         self._opts = LLMOptions(model=model)
         self._client = client or openai.AsyncClient(base_url=get_base_url(base_url))
         self._running_fncs: MutableSet[asyncio.Task] = set()
+        logging.getLogger("httpx").setLevel(logging.WARNING)
```
davidzhao, line 60: should the user have control over this?

nbsp: the most control OpenAI *lets* the user have is debug/info, both of which are too noisy for us. we could take control of the `OPENAI_LOG` envvar to ostensibly rewrite some of their logging code, but honestly i feel like that energy might be better spent sending a PR fixing it on their end?

davidzhao: I understand.. but seems odd that we are setting this behavior on behalf of their SDK.. if a user wants to suppress their httpx logs, wouldn't they just do it directly?

nbsp: well, yes, and that's still possible. we're just setting the default to warn because currently the logs are too noisy by default

davidzhao: makes sense.. this is subjective, but for me the info logs have been pretty helpful. I'm fine ok with the change to be consistent with the other plugins though

theomonnom:
https://github.com/livekit/agents/commit/18fef398e62028e5fe4681d338496fc85c7a6436

PR #505, opened 2024-07-24
fix wrong load reports & task leaks

theomonnom:
basically https://github.com/livekit/agents/pull/459

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -50,7 +50,8 @@ async def _default_request_fnc(ctx: JobRequest) -> None:
 
 
 def _default_cpu_load_fnc() -> float:
-    return psutil.cpu_percent() / 100
+    percent = psutil.cpu_percent(1.0)
+    return percent / 100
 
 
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/worker.py
```diff
@@ -313,9 +314,11 @@ async def _worker_task(self) -> None:
                     break
 
                 if retry_count >= self._opts.max_retry:
-                    raise Exception(
-                        f"failed to connect to livekit after {retry_count} attempts: {e}"
+                    logger.error(
```
davidzhao, line 317: we don't want to throw?

theomonnom: the reason is that the error were showing twice, because the cli also log exception on worker.run()

theomonnom: I'll double check to see if it introduced an issue where some code was waiting for an exc

theomonnom: yeah you're right, reverting to an exception

PR #524, opened 2024-07-25
livekit-plugins-openai: set timeout to 5 seconds

nbsp:


theomonnom:
https://www.python-httpx.org/advanced/resource-limits/ Let's increase the keepalive_expiry too

nbsp:
> [python-httpx.org/advanced/resource-limits](https://www.python-httpx.org/advanced/resource-limits/) Let's increase the keepalive_expiry too

added it with 15 seconds timeout (other attributes are openai defaults)

theomonnom:
tests aren't passing. Let's set them to 120seconds. I'm thinking about what would be the difference between two LLM calls

theomonnom (approved):

PR #525, opened 2024-07-25
Lazy create http session if needed

keepingitneil:
Some users are using the plugins by themselves without the agent subprocess/worker code. In these cases, a global http_session is never created. IMO this use case is perfectly legitimate.

### DIFF
--- a/livekit-agents/livekit/agents/utils/http_context.py
+++ b/livekit-agents/livekit/agents/utils/http_context.py
@@ -33,7 +33,7 @@ def http_session() -> aiohttp.ClientSession:
     val = _ContextVar.get(None)  # type: ignore
     if val is None:
         raise RuntimeError(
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/utils/http_context.py
```diff
@@ -32,7 +32,9 @@ def http_session() -> aiohttp.ClientSession:
 
     val = _ContextVar.get(None)  # type: ignore
     if val is None:
-        raise RuntimeError("no http_session() context available")
+        raise RuntimeError(
+            "Attempted to use an http session outside of a job context. This is probably because you are trying to use a plugin without also using the agent worker api. You may need to create your own aiohttp.ClientSession, pass it into the plugin constructor as a kwarg, and manage its lifecycle."
```
theomonnom, line 36: is the "also" a typo?

theomonnom (approved):
lgtm thanks

livekit-agents/livekit/agents/utils/http_context.py
keepingitneil: wasn't a typo IMO but also doesn't add value so I'll remove

PR #529, opened 2024-07-26
commit version bumps directly inside the changesets PR 

theomonnom:


lukasIO (commented):
looks good, but we'll also want to delete the steps that currently do the version bump in the release workflow

.github/workflows/publish-package.yml
```diff
@@ -103,12 +103,6 @@ jobs:
           python -m pip install --upgrade pip
           pip install build
 
-      - name: Bump version.py
-        run: |
-          export module=$(echo "${{ matrix.package.name }}" | tr '-' '/')
-          export version=$(sed 's|-\([0-9]*\)$|\1|' <<< ${{ matrix.package.version }})
-          sed -i "s/__version__.*/__version__ = \"$version\"/" $module/version.py
-
       - name: Commit changes to
```
lukasIO, line 106: also this one should be deleted

theomonnom: ah ok 👍

lukasIO (approved):

PR #543, opened 2024-07-29
initial prometheus metrics

theomonnom:
- multiprocessing & some basic metrics (running jobs, prewarm time)

still draft, needs reported cpu & memory metrics before merging.
Assistant stats will use another protocol


royatanu94:
@theomonnom 
 Can we use this ? Does this implementation works? TIA

PR #544, opened 2024-07-30
cartesia: support websocket connection & speech continuation 

sauhardjain:
draft because haven't set up testing yet


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -49,6 +54,7 @@ def __init__(
         sample_rate: int = 24000,
         api_key: str | None = None,
         http_session: aiohttp.ClientSession | None = None,
+        sentence_tokenizer: tokenize.SentenceTokenizer = tokenize.basic.SentenceTokenizer(),
```
theomonnom, line 57: Do we need a sentence_tokenizer when using the Cartesia API?
Can't we just send words and Cartesia directly?

sauhardjain: for continuations, the pacing of speech does improve a bit if we pass in sentences. alternatively, we could buffer in the first 4-5 words and then stream in words to the websocket as they come in. 

between these two approaches, what would work better within the livekit agents framework?

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -137,3 +148,176 @@ async def _main_task(self):
...
+                )
+                break
+            except Exception as e:
+                logger.warning(
+                    f"failed to connect to Cartesia, retrying in {retry_delay}s",
+                    exc_info=e,
+                )
+
+        if ws_conn is None:
+            raise Exception(f"failed to connect to Cartesia after {max_retry} retries")
```
sauhardjain, line 224: @theomonnom this follows the pattern set by the 11labs websocket plugin, however, i'm not sure if making a new websocket connection on every `sentence_stream` or `word_stream` is the right approach because that introduces latency associated with the websocket handshake

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
theomonnom: ah ok, I think it is OK to keep sentences. Just wondering what is the best for Cartesia

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
theomonnom: The reason why the 11labs connection is closed is because they don't support to make a new request in the same connection after a flush

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
sauhardjain: yep, generally slightly longer outputs have better prosody and pacing!

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
sauhardjain: gotcha! we should tweak this to maintain the same WS connection and refresh it on each call.

theomonnom:
Hey, I wanted to add tests but it seems like I don't have access to push changes to your branch.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
theomonnom: For long sentences coming from an LLM. I'm wondering if it should be better to send words instead?
Seems like this could improve the time to first frame?

theomonnom: We are also currently making a new request for each sentence (using the same WS). Does cartesia support multiple sentences in one single request instead (context_id)?

sauhardjain: Yes! So all the requests sent using one `context_id` are treated as part of a "continuation" which means that the speech for this group of requests will sound contiguous.

sauhardjain: > For long sentences coming from a LLM. I'm wondering if it should be better to send words instead?
Seems like this could improve the time to first frame?

Tokenizing on sentences usually leads to more natural-sounding speech. Words will work fine too, but for the first request sent to a `context_id`, we should buffer to 4-5 words.

theomonnom (approved):

PR #546, opened 2024-07-30
voiceassistant: interrupt on final transcript

theomonnom:


livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -381,6 +381,14 @@ def _on_final_transcript(ev: stt.SpeechEvent) -> None:
 
             self._deferred_validation.on_human_final_transcript(ev.alternatives[0].text)
 
+            words = self._opts.transcription.word_tokenizer.tokenize(
+                text=ev.alternatives[0].text
+            )
+            if len(words) >= 3:
```
theomonnom, line 387: amount of word isn't exposed for now. the behaviour can be confused with min_int_words.

keepingitneil (approved):

PR #553, opened 2024-07-31
forgot to add changesets

theomonnom:


.changeset/perfect-doors-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"livekit-plugins-deepgram": patch
+---
+
+deepgram: segment audio frames into 200ms intervals before sending to the websocket
```
lukasIO, line 5: for the changelog it helps to mention the PR that included the change e.g. 
```suggestion
deepgram: segment audio frames into 200ms intervals before sending to the websocket #549
```

lukasIO: (this is only needed when adding changesets retroactively, as they will by default always be associated with the PR that got merged into main)

lukasIO (approved):

PR #554, opened 2024-07-31
voiceassistant: run function calls sequentially

theomonnom:


livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -567,9 +567,17 @@ def _commit_user_question_if_needed() -> None:
...
-            await asyncio.gather(*tasks, return_exceptions=True)
+            called_fncs_info = speech_info.source.function_calls
+
+            called_fncs = []
+            for fnc in called_fncs_info:
+                called_fnc = fnc.execute()
+                called_fncs.append(called_fnc)
+                try:
+                    await called_fnc.task
+                except Exception:
```
keepingitneil, line 578: Should we log something here?

theomonnom: no, the ai callables errors will be forwarded to the LLM.
But I think we should create an AIError type, so we can still log unexpected errors

keepingitneil: It might still make sense to log something no? When debugging an old session it could be useful

theomonnom: we didn't before, I think this should be handled in a new PR, because we need to introduce a new special exception type that are marked as "expected"

keepingitneil (approved):

PR #562, opened 2024-08-01
silero: bring back expfilter

theomonnom:


livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
```diff
@@ -48,6 +50,7 @@ def load(
         max_buffered_speech: float = 60.0,
         activation_threshold: float = 0.25,
         sample_rate: int = 16000,
+        probability_alpha: float = 0.95,
```
theomonnom, line 53: mb a better name?

keepingitneil (approved):

PR #566, opened 2024-08-02
fix(voice assistant): missing spaces between transcript chunks

egoldschmidt:
If a user speaks slowly or trails off a bit we are finding that Deepgram will return multiple chunks of "final" transcripts, which are concatenated without a space creating things like this:

> Yeah. I mean, I I haven't packedyet at all, but she's got, Harry all packed up, so that's the most important part.Know,

(Notice lack of space between "packed" and "yet" and before "Know")

### DIFF
--- a/.changeset/empty-rings-begin.md
+++ b/.changeset/empty-rings-begin.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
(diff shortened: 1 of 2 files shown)


theomonnom (approved):
Thanks! Can you add a changesets so I can merge the PR

theomonnom (approved):

PR #567, opened 2024-08-02
elevenlabs: error on non-PCM data

nbsp:
will now output the following and skip synthesis instead of attempting to process data as PCM:

```
2024-08-02 12:01:38,006 - ERROR livekit.plugins.elevenlabs - 11labs returned non-PCM data: {"detail":{"status":"invalid_api_key","message":"Invalid API key"}} {"pid": 59945, "job_id": "AJ_5oARRaVErZCW"}
```

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -181,6 +181,10 @@ class ChunkedStream(tts.ChunkedStream):
             headers={AUTHORIZATION_HEADER: self._opts.api_key},
             json=data,
         ) as resp:
+            if resp.content_type != "audio/pcm":
+                content = await resp.text()
+                logger.error("11labs returned non-PCM data: %s", content)
+                return
             async for bytes_data, _ in resp.content.iter_chunks():
                 for frame in bstream.write(bytes_data):
                     self._event_ch.send_nowait(
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -156,6 +156,7 @@ class ChunkedStream(tts.ChunkedStream):
     ) -> None:
         super().__init__()
         self._text, self._opts, self._session = text, opts, session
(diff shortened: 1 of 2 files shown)


theomonnom:
ah that's nice, I think we should also add support for mp3 encoding on the synthesize method (like we do for the streaming method)
https://github.com/livekit/agents/blob/8ce90df03ec2e81dad0a56f66abb21b086704803/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py#L344

I think some free tiers of 11labs don't allow PCM.

nbsp:
> I think some free tiers of 11labs don't allow PCM.

i think checking for starting with `audio/` should be enough. errors would return `text/plain` and anything else should not be handled either way

theomonnom:
No I mean to add support for Mp3Decoder inside this PR, it should be very easy to do.
Take a look at https://github.com/livekit/agents/blob/8ce90df03ec2e81dad0a56f66abb21b086704803/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py#L345

theomonnom (approved):
awesome, just need to fix the CI then it's good to merge :)

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -156,6 +156,7 @@ def __init__(
     ) -> None:
         super().__init__()
         self._text, self._opts, self._session = text, opts, session
+        self._mp3_decoder = utils.codecs.Mp3StreamDecoder()
```
theomonnom, line 159:
```suggestion
        self._mp3_decoder = utils.codecs.Mp3StreamDecoder()
```
not: the mp3 decoder isn't available without using livekit-agents[codec]. Can you initialize it inside the condition `if encoding == "mp3":`

nbsp:
synthesize seems to return `audio/pcm` even when mp3 is requested, i don't think we need to decode it. that seems to fail

theomonnom:
Are we using the right parameter? https://elevenlabs.io/docs/api-reference/text-to-speech

theomonnom:
The issue is [here](https://github.com/livekit/agents/blob/88e75d71615a10d6f4d62b77764edba997e73499/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py#L395)
for synthesize_url we always assume "pcm_"

PR #572, opened 2024-08-03
0.8 migration guide

davidzhao:


README.md
```diff
@@ -81,14 +84,20 @@ The framework exposes a CLI interface to run your agent. To get started, you'll
...
 - LIVEKIT_API_SECRET
 
-### Running the worker
+### Starting the worker
 
 This will start the worker and wait for users to connect to your LiveKit server:
 
 ```bash
 python my_agent.py start
 ```
```
theomonnom, line 94:
```suggestion
```
Now that I think about it, I think we not show start here I'm seeing too many people never using dev.
Better place would be inside our docs?

theomonnom (approved):

README.md
davidzhao: I don't think `dev` should be the default here.. you don't really *need* dev mode.. but when users eventually move to production, they have to understand `start`

PR #575, opened 2024-08-04
openai: fix incorrect API urls on Windows

theomonnom:
os.path.join uses backslash on windows. the appropriate fnc to use is urllib.parse.urljoin


nbsp:
@theomonnom ping me when this one's ready for review, this is all that's needed for the CI updates to pass as well

keepingitneil (approved):

PR #577, opened 2024-08-05
Add single space between transcribed speech fragments

hari01584:
Initially, when user was talking to agent in slow manner..

Ie something like Hello...I..am..doing..good, then it was registering those as different chunks and the final text that was being formed was "Iamdoinggood" (See No Spaces)

However if user speak the same sentence fast and fluent like "I am doing good" then transcription had proper spaces.

This was due to the logic to simply append transcription without adding any additional space, this patch fixes it and make the text better for OpenAI to parse

### DIFF
--- a/.changeset/big-insects-dance.md
+++ b/.changeset/big-insects-dance.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+voiceassistant: add single space between transcribed speech fragments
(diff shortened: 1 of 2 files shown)


theomonnom:
Hey, seems like it is a duplicate of #566

hari01584:
Yes, thanks for merge!

PR #578, opened 2024-08-05
elevenlabs: update default model to eleven_turbo_v2_5

theomonnom:
This model should have a faster inference time (the first time to first frame didn't really reduced tho)


keepingitneil (approved):
looks like ruff failed but lgtm

PR #583, opened 2024-08-05
openai: use openai client for stt

theomonnom:
- Primarily for consistency with other openai interfaces
- also fix tests

http_session has been replaced by the openai client


davidzhao (approved):
lg

PR #586, opened 2024-08-06
Fix elevenlabs voice settings breaking

nbsp:
11labs API requires `style` float and `use_speaker_boost` boolean. defaulting to None gets translated to JSON as `null` type. this sets them to their default values noted in the API here: https://api.elevenlabs.io/v1/voices/settings/default

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -168,7 +168,7 @@ class ChunkedStream(tts.ChunkedStream):
         segment_id = utils.shortuuid()
 
         voice_settings = (
-            dataclasses.asdict(self._opts.voice.settings)
+            _strip_nones(dataclasses.asdict(self._opts.voice.settings))
             if self._opts.voice.settings
             else None
         )
@@ -300,7 +300,7 @@ class SynthesizeStream(tts.SynthesizeStream):
         init_pkt = dict(
             text=" ",
             try_trigger_generation=True,
-            voice_settings=dataclasses.asdict(self._opts.voice.settings)
+            voice_settings=_strip_nones(dataclasses.asdict(self._opts.voice.settings))
             if self._opts.voice.settings
             else None,
             generation_config=dict(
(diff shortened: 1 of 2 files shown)


nbsp:
just noticed that this does break our API, as the default for speaker boosting is now `True` and it used to be `False`; do we keep it as it was?

theomonnom:
Are the defaults safe when listing the voices using this fnc:
  async def list_voices(self) -> List[Voice]:

nbsp:
> Are the defaults safe when listing the voices using this fnc?
> `async def list_voices(self) -> List[Voice]:`

that function sets `settings=None`, which changes the behaviour of the stock voices now, yes

theomonnom:
Looking at their docs, they seem to explicitly use it on the examples? https://github.com/elevenlabs/elevenlabs-python/blob/a75c7e7e0b2aa8d04225ca34e2e387ccd0ba13ac/README.md#%EF%B8%8F-voices

On their 11labs python-client, I see that they also default these stuff to None

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -50,8 +50,8 @@ def _encoding_from_format(output_format: TTSEncoding) -> _Encoding:
 class VoiceSettings:
     stability: float  # [0.0 - 1.0]
     similarity_boost: float  # [0.0 - 1.0]
-    style: float | None = None  # [0.0 - 1.0]
-    use_speaker_boost: bool | None = False
+    style: float = 0.0  # [0.0 - 1.0]
+    use_speaker_boost: bool = True
```
keepingitneil, line 54: should we keep the same default as before?

theomonnom (approved):

nbsp:
new method just strips the `None`s from the object before passing it to the JSON encoder, and doesn't mess with our defaults, so everything should be the same.


> Looking at their docs, they seem to explicitly set it on the examples? [elevenlabs/elevenlabs-python@`a75c7e7`/README.md#%EF%B8%8F-voices](https://github.com/elevenlabs/elevenlabs-python/blob/a75c7e7e0b2aa8d04225ca34e2e387ccd0ba13ac/README.md#%EF%B8%8F-voices)
> 
> On their 11labs python-client, I see that they also default these stuff to None

i don't think elevenlabs' Python client even supports professional or cloned voices, they errored as well.

PR #588, opened 2024-08-06
voiceassistant: fix synthesis continuing after interruption

theomonnom:
the synthesis of speech was still happening even after an interruption. 
this PR also adds more logs for ttft


davidzhao (approved):
lg!

PR #589, opened 2024-08-06
voiceassistant: only commit the spoken words in the chat context

theomonnom:
This is useful when the speech was interrupted. This makes sure the LLM has the same "context" as what the user heard

### DIFF
--- a/livekit-agents/livekit/agents/voice_assistant/agent_output.py
+++ b/livekit-agents/livekit/agents/voice_assistant/agent_output.py
@@ -41,7 +41,7 @@ class SynthesisHandle:
         return self._speech_id
 
     @property
-    def tr_fwd(self) -> agent_transcription.TTSSegmentsForwarder:
+    def tts_forwarder(self) -> agent_transcription.TTSSegmentsForwarder:
         return self._tr_fwd
 
     @property
@@ -173,9 +173,9 @@ class AgentOutput:
 @utils.log_exceptions(logger=logger)
 async def _str_synthesis_task(text: str, handle: SynthesisHandle) -> None:
     """synthesize speech from a string"""
-    if not handle.tr_fwd.closed:
-        handle.tr_fwd.push_text(text)
-        handle.tr_fwd.mark_text_segment_end()
+    if not handle.tts_forwarder.closed:
+        handle.tts_forwarder.push_text(text)
+        handle.tts_forwarder.mark_text_segment_end()
 
     start_time = time.time()
     first_frame = True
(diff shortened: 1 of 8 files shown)


livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -637,12 +637,15 @@ def _commit_user_question_if_needed() -> None:
...
 
-                collected_text = answer_synthesis.collected_text
+                collected_text = answer_synthesis.tr_fwd.played_text
                 interrupted = answer_synthesis.interrupted
 
         if speech_info.add_to_chat_ctx and (not user_question or user_speech_committed):
             self._chat_ctx.messages.extend(extra_tools_messages)
 
+            if interrupted:
+                collected_text += "..."
```
theomonnom, line 647: This is kind of opinionated, wdyt?

nbsp: maybe the singular ellipsis `…` instead of `...` works better?

davidzhao: I think it's fine.. perhaps make this a constant? (if folks want to customize, we could offer it as an option)

livekit-agents/livekit/agents/voice_assistant/agent_output.py
```diff
@@ -34,13 +34,16 @@ def __init__(
...
         self._interrupt_fut = asyncio.Future[None]()
-        self._collected_text = ""  # collected text from the async stream
         self._speech_id = speech_id
 
     @property
     def speech_id(self) -> str:
         return self._speech_id
 
+    @property
+    def tr_fwd(self) -> agent_transcription.TTSSegmentsForwarder:
```
nbsp, line 44: make this more understandable in the public API at least

```suggestion
    def transcription_forwarder(self) -> agent_transcription.TTSSegmentsForwarder:
```

davidzhao (approved):

livekit-agents/livekit/agents/voice_assistant/agent_output.py
davidzhao: good point, +1 making the public API clear vs abbreviated

theomonnom: this isn't public, this is still internal

davidzhao: maybe `tts_forwarder` ?

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
theomonnom: if folks want to customize it, they can listen to the agent_speech_interrupted event

Toubat: I'd like this to be customized :)

PR #592, opened 2024-08-07
tts,stt: add FallbackAdapter

nbsp:
usage:

```python
text_to_speech = tts.FallbackAdapter(
    openai.TTS(),
    cartesia.TTS(),
    # any others that you may want...
    cooldown=30,
)
```

notes:
- [TTS] relies on sample rate and channel being the exact same across all TTS providers. won't initialize if not the case
- only sets `capabilities.streaming` to true if all providers support it
- ~~doubles as a timeout adapter, if you only set one provider~~ (**update**: we've decided providers should handle their own timeouts)
- if `cooldown > 0` then it reactivates a provider after `cooldown` seconds

issues:
- [x] ~~doesn't work with openai for some reason? it publishes data fine but then after it's done RTC complains about mismatched audio sources~~
- [x] TTS streaming throws an error after it's done (see note in source)
- [x] i haven't _actually_ checked that it switches when a provider fails, yet. dunno how to make that happen
  - [ ] switches tested working between transcriptions, still haven't tested what happens mid-transcription
- [x] ~~timeouts are currently not preset to anything, need to figure out what a good threshold is~~ set to 5s/120s to match our OpenAI timeout
- [x] STT FallbackAdapter untested with plain `recognize`, will test later, but streaming works great
- [x] need to remember to write tests when it's ready

feedback, nits, etc appreciated


livekit-agents/livekit/agents/tts/fallback_adapter.py
```diff
@@ -0,0 +1,153 @@
...
+                    # TODO(nbsp): elevenlabs _stream._run_ws()#_recv_task() simply exits when it's
+                    #             done, and the iterator hangs indefinitely until it times out.
+                    #             probably needs a change to SynthesizeStream to make it work
+                    item = await asyncio.wait_for(
+                        self._stream.__anext__(), self._timeout
+                    )
+                    self._timeout = self._keepalive_timeout
+                    self._event_ch.send_nowait(item)
+                except Exception as e:
+                    if isinstance(e, asyncio.TimeoutError) or isinstance(
```
theomonnom, line 134: This exception is never catched. asyncio.TimeoutError doesn't inherit from Exception

theomonnom, line 134: are asyncio.TimeoutError & TimeoutError different?

nbsp: python 3.11 notes state that the exception returned changed to plain AsyncTimeout, so this looks for both for compatibility

nbsp:
this is almost ready to review, just one bug left to squash:
with unlucky enough timing, a new sentence could be pushed to the stream just as the timeout hits 0, but since the stream is active it thinks it timed out. need to figure out a way to refresh the timer

livekit-agents/livekit/agents/stt/fallback_adapter.py
```diff
@@ -0,0 +1,153 @@
...
+    async def _main_task(self) -> None:
+        async def _pass_input():
+            async for input in self._input_ch:
+                self._stream.push_frame(input)
+
+        async def _pass_output():
+            while True:
+                try:
+                    item = await asyncio.wait_for(
+                        self._stream.__anext__(), self._timeout
```
theomonnom, line 133: Not sure if timeout makes sense for a streamed STT. The user could just "not speak" and we would reach this timeout

nbsp: this is kinda the same issue as the other StreamAdapter; we don't have a keepalive function so we can't discern between standby and timeout. i want to figure out a good solution for this on the TTS side first then port it here

davidzhao: we still have a way of detecting if the connection is alive right? maybe it's better to rely on connection-level timeout first, to eliminate false positives

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -256,7 +256,9 @@ async def _tokenize_input():
         @utils.log_exceptions(logger=logger)
         async def _run():
             async for word_stream in self._segments_ch:
+                self._pending += 1
                 await self._run_ws(word_stream)
+                self._pending -= 1
```
davidzhao, line 261: should this be in a finally block?

livekit-agents/livekit/agents/stt/fallback_adapter.py
nbsp: `StreamAdapter`s don't use a websocket, because they just map normal requests to fit the `SynthesizeStream` object, so we can't check for that either unfortunately

nbsp:
this should be the entire FallbackAdapter implementation done. can be merged once streams handle their own timeouts, which i'll do in a different PR

bradyneal:
Awesome. Cartesia went down during a user test yesterday, so I'm excited for this!

bradyneal:
Any update on this?

nbsp:
we're now working on this over here: https://github.com/livekit/agents/pull/1074

PR #597, opened 2024-08-07
disable http server by default

theomonnom:
We don't have much usage for it today (except health_check) 

Also reduce UPDATE_LOAD_INTERVAL from 10.0 to 5.0

### DIFF
--- a/.changeset/afraid-actors-wonder.md
+++ b/.changeset/afraid-actors-wonder.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+disable http server by default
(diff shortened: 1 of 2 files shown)


nbsp (approved):

davidzhao (approved):
looks fine, but I think we would benefit from a more dynamic update mechanism.. where we run the loop more frequently, but send an update if:
* load changes by more than 10%
* it's been more than 5 seconds

that'll allow us to react faster to load changes

davidzhao:
@theomonnom I'm also not sure about disabling the server by default.. perhaps don't fail the process if it's in conflict?

theomonnom:
> looks fine, but I think we would benefit from a more dynamic update mechanism.. where we run the loop more frequently, but send an update if:
> 
> * load changes by more than 10%
> * it's been more than 5 seconds
> 
> that'll allow us to react faster to load changes

Makes sense, will plan this one for another PR

davidzhao:
default behaviors are tricky to change.. as it breaks user's expectations

PR #601, opened 2024-08-07
Support OpenAI Assistants API

keepingitneil:


examples/voice-assistant/minimal_assistant.py
keepingitneil, line 1: No need to review this file. It won't be included in the PR. Just using it for testing

examples/voice-assistant/minimal_assistant.py
```diff
@@ -10,20 +10,20 @@
...
+    initial_ctx = llm.ChatContext()
 
     await ctx.connect(auto_subscribe=AutoSubscribe.AUDIO_ONLY)
 
     assistant = VoiceAssistant(
         vad=silero.VAD.load(),
         stt=deepgram.STT(),
-        llm=openai.LLM(),
+        llm=openai.AssistantLLM(
+            assistant_opts=openai.AssistantCreateOptions(
```
davidzhao, line 21: if I'm already using assistant API today.. is there a one to one mapping for the options?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/assistant_llm.py
```diff
@@ -0,0 +1,246 @@
...
+                timeout=5.0,
+                follow_redirects=True,
+                limits=httpx.Limits(
+                    max_connections=1000,
+                    max_keepalive_connections=100,
+                    keepalive_expiry=120,
+                ),
+            ),
+        )
+        self._assistant_opts = assistant_opts
```
davidzhao, line 78: nit: a bit confusing to set both types to the same var, and then to check the same var later.. not knowing what type it is. seems cleaner to keep them separate and only sync if it's a `AssistantCreateOptions`

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/assistant_llm.py
```diff
@@ -0,0 +1,246 @@
...
+    def chat(
+        self,
+        *,
+        chat_ctx: llm.ChatContext,
+        fnc_ctx: llm.FunctionContext | None = None,
+        temperature: float | None = None,
+        n: int | None = 1,
+        parallel_tool_calls: bool | None = None,
+    ):
+        return AssistantLLMStream(
```
davidzhao, line 105: would there be race issues if a user does:

```
llm.chat(message1)
llm.chat(message2)
```

is it possible for message1 to override the message context from message2 there? should the `chat` function limit only a single active stream at any given time?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/assistant_llm.py
```diff
@@ -0,0 +1,246 @@
...
+
+            added_messages_set: set[str] = self._chat_ctx._metadata[MESSAGES_ADDED_KEY][
+                self._thread_id
+            ]
+            # Note: this will add latency unfortunately. Usually it's just one message so we loop it but
+            # it will create an extra round trip to OpenAI before being able to run inference.
+            for msg in self._chat_ctx.messages:
+                msg_id = msg._metadata.get(MESSAGE_ID_KEY)
+                if msg_id and msg_id not in added_messages_set:
+                    await self._client.beta.threads.messages.delete(
```
davidzhao, line 195: can we run a task to sync all the changes that needed to be made vs await one by one?

davidzhao (commented):
the interfaces seem to fit in pretty well

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/assistant_llm.py
keepingitneil: This is a limitation of the OpenAI api. For any given thread_id, only one stream can be running at a time. Also, messages can not be mutated while that stream is running. The llm instance represents the thread. 

Logically it makes more sense for the chat_ctx to be the object that represents a thread but since anyone can mutate and/or copy the chat_ctx at any time (VoiceAssistant), we needed additional logic and state which couldn't really live anywhere else.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/assistant_llm.py
keepingitneil: There isn't a way to order messages so it was one by one. BUT!

I just made an adjustment by using the `additional_messages` which creates the messages at stream time. This introduced a fair amount of additional complexity because now we don't know the open_ai message_id up front. But definitely worth it for latency.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/assistant_llm.py
```diff
@@ -0,0 +1,346 @@
...
+class AssistantLoadOptions:
+    assistant_id: str
+    thread_id: str
+
+
+class AssistantLLM(llm.LLM):
+    def __init__(
+        self,
+        *,
+        assistant_opts: AssistantOptions,
```
theomonnom, line 72: What about `AssistantCreateOptions | AssistantLoadOptions` instead?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/assistant_llm.py
```diff
@@ -0,0 +1,346 @@
...
+from openai.types.beta.threads import Text, TextDelta
+from openai.types.beta.threads.run_create_params import AdditionalMessage
+from openai.types.beta.threads.runs import ToolCall, ToolCallDelta
+
+from .log import logger
+from .models import AssistantTools, ChatModels
+from .utils import build_oai_message
+
+DEFAULT_MODEL = "gpt-4o"
+OPEN_AI_MESSAGE_ID_KEY = "__openai_message_id__"
```
theomonnom, line 35: nit
```suggestion
OPENAI_MESSAGE_ID_KEY = "__openai_message_id__"
```

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/assistant_llm.py
```diff
@@ -0,0 +1,346 @@
...
+    def chat(
+        self,
+        *,
+        chat_ctx: llm.ChatContext,
+        fnc_ctx: llm.FunctionContext | None = None,
+        temperature: float | None = None,
+        n: int | None = 1,
+        parallel_tool_calls: bool | None = None,
+    ):
+        if n is not None:
```
theomonnom, line 126: Seems like this is always going to this condition? 
n defaults to 1 here

theomonnom (approved):
awesome!!

PR #608, opened 2024-08-09
Switch Cartesia to a sentence tokenizer

keepingitneil:


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -263,16 +240,16 @@ async def recv_task():
                             )
                         )
 
-                    pending_segments.remove(segment_id)
-                    if len(pending_segments) == 0 and self._input_ch.closed:
+                    if segment_id == segment_id:
```
theomonnom, line 243: this is always true?

theomonnom: I think you still need a list of pending segments (There could be multiple flush and the audio is always late)

keepingitneil: woops yes, funny because it actually might not matter that we go the segment id check because only one is used per connection right?

keepingitneil: Does the old code actually do this? Multiple context_ids actually causes issues. Cartesia runs them in parallel resulting in overlapping audio

livekit-agents/livekit/agents/tokenize/token_stream.py
```diff
@@ -59,7 +62,10 @@ def flush(self) -> None:
             else:
                 buf = self._buf
 
-            self._event_ch.send_nowait(TokenData(token=buf))
+            self._event_ch.send_nowait(
+                TokenData(token=buf, segment_id=self._current_segment_id)
```
theomonnom, line 66: 💯

theomonnom (approved):

sauhardjain:
looks great!

theomonnom (approved):

PR #609, opened 2024-08-09
added alternative ChatImage data field: base64_string

willsmanley:
probably could be a more robust change including implementation on other plugins, stricter typing to enforce mutual exclusivity with `image` field, etc

but wanted to put it out to get initial thoughts.

alternative approach could be to overload the `image` field and use a different and use a different property to signify that it is base64 instead of URL.

issue: https://github.com/livekit/agents/issues/610


willsmanley:
PS if any maintainers have time to look at this livekit flutter SDK echo cancellation bug, it would be amazing: https://github.com/livekit/client-sdk-flutter/issues/565 

I made a quick repro repo for it

PR #611, opened 2024-08-10
openai.LLM: add common API-compatible adapters

nbsp:
supports Ollama, Fireworks, Perplexity, Groq, Together, Octo.

Fireworks and Ollama accept arbitrary strings for models, checking against a list that gets updated too frequently for us to hardcode.

Ollama doesn't read the API key, but `openai` requires it regardless, so the guide says to supply it with an arbitrary blob.

Perplexity doesn't seem to need `/v1/` appended to its API URL.

currently untested, but all of these projects' documentations imply they work as a drop-in replacement.

~~**NOTE**: this is a minor bump! maybe let's not merge this until we want to release a v0.9.0~~


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -62,6 +62,45 @@ def __init__(
...
+        *,
+        model: str | PerplexityChatModels = "llama-3.1-sonar-small-128k-chat",
+        api_key: str | None = None,
+        base_url: str | None = "https://api.perplexity.ai",
+        client: openai.AsyncClient | None = None,
+    ) -> LLM:
+        return LLM(model=model, api_key=api_key, base_url=base_url, client=client)
+
     @staticmethod
     def create_azure_client(
```
theomonnom, line 105: It may not be the best name we picked for Azure, but should we keep the names consistent? cc @livekit/agent-devs

theomonnom, line 105: It may not be the best name we picked for Azure, but should we keep the names consistent? cc @livekit/agent-devs

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
```diff
@@ -30,3 +30,29 @@
...
+    "llama-3.1-8b-instant",
+    "llama3-groq-70b-8192-tool-use-preview",
+    "llama3-groq-8b-8192-tool-use-preview",
+    "llama-guard-3-8b",
+    "llama3-70b-8192",
+    "llama3-8b-8192",
+    "mixtral-8x7b-32768",
+    "gemma-7b-it",
+    "gemma2-9b-it",
+    "whisper-large-v3",
```
theomonnom, line 57:
```suggestion
```

theomonnom:
Let's add https://www.together.ai/ I know Shayne is using it :) 
https://octo.ai/ may also be relevant

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
nbsp: i think an important difference here is that Azure is still an OpenAI usecase, while the rest are other services supporting the same API. so `create_perplexity_client` makes less sense than it does for azure imo. i'm fine with changing it if there's consensus though

nbsp:
> Let's add [together.ai](https://www.together.ai/) I know Shayne is using it :) [octo.ai](https://octo.ai/) may also be relevant

added both of those as well!

noting something here that i just edited onto the top message: we're going to have to release 0.9.0 for this, so let's not merge this until 0.8 stabilizes

davidzhao (commented):
awesome! just to resolve the consistency question

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
nbsp: got it. just another potential issue: do we go with `ai` at the end of the names? or not? or depending on what the provider calls itself? for example `groq`/`fireworks` vs `octo`/`together`

theomonnom:
Why does this PR can't be a patch instead of a minor version?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
davidzhao: I think without is better. AI is kind of redundant. let's also alias the `create_azure_client` to `with_azure`, while marking the former deprecated.

nbsp:
> > noting something here that i just edited onto the top message: we're going to have to release 0.9.0 for this, so let's not merge this until 0.8 stabilizes
> 
> Why does this PR can't be a patch instead of a minor version? I know we're releasing new features but since we're not breaking anything etc.. no need to wait

your call. thinking about it from a semver compliance POV, but it doesn't matter

nbsp:
> I think without is better. AI is kind of redundant. let's also alias the `create_azure_client` to `with_azure`, while marking the former deprecated.

https://github.com/livekit/agents/pull/611#discussion_r1712680890
if we change the function name for azure we have to bump minor anyway

davidzhao (approved):
lg!

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -100,6 +107,92 @@ def create_azure_client(
...
+        azure_deployment: str | None = None,
+        api_version: str | None = None,
+        api_key: str | None = None,
+        azure_ad_token: str | None = None,
+        azure_ad_token_provider: AsyncAzureADTokenProvider | None = None,
+        organization: str | None = None,
+        project: str | None = None,
+        base_url: str | None = None,
+    ) -> LLM:
+        warnings.warn("This alias is deprecated. Use LLM.with_azure() instead")
```
theomonnom, line 183: Can you use the logger from the plugin

someshfengde:
can I use Groq with livfekit?

billyg88:
> can I use Groq with livekit?

Try this ```with_groq``` there is a specific method for that... look at the code in ```https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py```

PR #621, opened 2024-08-13
limit simultaneous process initialization

theomonnom:
- avoid stressing the OS too much

### DIFF
--- a/livekit-agents/livekit/agents/ipc/proc_pool.py
+++ b/livekit-agents/livekit/agents/ipc/proc_pool.py
@@ -14,6 +14,8 @@ EventTypes = Literal[
     "process_created", "process_started", "process_ready", "process_closed"
 ]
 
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/ipc/proc_pool.py
```diff
@@ -14,6 +14,8 @@
     "process_created", "process_started", "process_ready", "process_closed"
 ]
 
+MAX_CONCURRENT_INITIALIZATIONS = 3
```
theomonnom, line 17: I'm not exposing this value for now (simplicity)

theomonnom: It generally happens at startup. Otherwise it happens when you start to receive more jobs than the num_idle_processes

davidzhao (approved):

PR #624, opened 2024-08-14
Cartesia bump to 0.4.0

davidzhao:


nbsp:
Changesets should handle this on its own. looking into why, but we shouldn't merge this, it might break some things

nbsp:
if we want to bump cartesia to 0.4.0 instead of 0.3.1, we should instead change the `.changeset/five-planes-drum.md` file to say `minor` instead of `patch`

davidzhao:
> if we want to bump cartesia to 0.4.0 instead of 0.3.1, we should instead change the `.changeset/five-planes-drum.md` file to say `minor` instead of `patch`

I'm going to release this one manually, since agents isn't ready for the next release yet.

keepingitneil (approved):

PR #629, opened 2024-08-15
Modified RAG plugin

afigar:
I just added:

    def download_files(self) -> None:
        pass

and It worked!


theomonnom (approved):

theomonnom:
Thanks!

PR #633, opened 2024-08-15
openai: allow sending user IDs

### LINKED ISSUE #612: Add Support for Sending End-user IDS to Openai. This will prevent Openai account suspention incase of misuse by end user.
Hi as referenced [here](https://platform.openai.com/docs/guides/safety-best-practices/end-user-ids) OpenAI accepts End user ids. It might be useful tool to help OpenAI monitor and detect abuse. This allows OpenAI to provide your team with more actionable feedback in the event that we detect any policy violations in your application.

nbsp:
Fixes #612

chose to add it to `chat()` but might make sense to add it to `LLM()`.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -200,6 +200,7 @@ class LLM(llm.LLM):
         temperature: float | None = None,
         n: int | None = 1,
         parallel_tool_calls: bool | None = None,
+        user: str | None = None,
     ) -> "LLMStream":
         opts: dict[str, Any] = dict()
         if fnc_ctx and len(fnc_ctx.ai_functions) > 0:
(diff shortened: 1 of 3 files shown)


davidzhao:
I think adding to LLM constructor would be better. VoiceAssistant will be issuing the `chat` calls and it wouldn't be passing in additional details such as this.

davidzhao (approved):

davidzhao:
wanna fix the type check while you are there?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -212,13 +232,16 @@ def chat(
             if fnc_ctx and parallel_tool_calls is not None:
                 opts["parallel_tool_calls"] = parallel_tool_calls
 
+        user = self._opts.user or openai.NOT_GIVEN
```
theomonnom, line 235: is openai.NOT_GIVEN exported? last time I checked it wasn't

nbsp: https://github.com/openai/openai-python/blob/3bf03f2f7ee1a9e214484524000774532d02cba7/src/openai/__init__.py#L45

PR #636, opened 2024-08-16
Add agent_name to WorkerOptions

keepingitneil:


livekit-agents/livekit/agents/worker.py
```diff
@@ -583,7 +589,11 @@ async def _on_accept(args: JobAcceptArguments) -> None:
 
         logger.info(
             "received job request",
-            extra={"job_request": msg.job, "resuming": msg.resuming},
+            extra={
+                "job_request": msg.job,
+                "resuming": msg.resuming,
+                "agent_name": self._opts.agent_name,
```
theomonnom, line 595: I think if the worker can only have one agent_name. It isn't useful to relog it every time?

livekit-agents/livekit/agents/job.py
```diff
@@ -79,7 +79,9 @@ def room(self) -> rtc.Room:
     def agent(self) -> rtc.LocalParticipant:
         return self._room.local_participant
 
-    def add_shutdown_callback(self, callback: Callable[[], Coroutine]) -> None:
+    def add_shutdown_callback(
+        self, callback: Callable[[], Coroutine[None, None, None]]
```
theomonnom, line 83: Wondering what is the right type here.
Awaitable[Any]?

keepingitneil: Awaitable doesn't work downstream when we do an `asyncio.create_task` which wants a Coroutine

livekit-agents/livekit/agents/worker.py
keepingitneil: IMO actually we should log it even more...for example, imagine you wanted to do a datadoq query against specific agent_names. Maybe you want to see if VAD takes too long on certain agent_names

theomonnom: I would use the WorkerID no?
all the logs coming from the child processes will not print agent_names anyway.
Think we should add worker_id everywhere tho

livekit-agents/livekit/agents/worker.py
```diff
@@ -118,6 +118,8 @@ class WorkerOptions:
     """Maximum amount of time to wait for a process to initialize/prewarm"""
     permissions: WorkerPermissions = field(default_factory=WorkerPermissions)
     """Permissions that the agent should join the room with."""
+    agent_name: str = "default"
```
theomonnom, line 121: is this the right default to have? wondering how the server code is treating it, maybe there is no distinction

livekit-agents/livekit/agents/worker.py
keepingitneil: I think it would be useful to have agent_name as well. At least I can imagine it being valuable for agent_name to indexed in logging platforms (Datadogs, ELK, w/e). For example, being able to see that a bunch of error logs only happen on a specific agent_name would be useful to know

davidzhao: agreed. each log line should have sufficient context for a search query. context is super helpful when users are searching.

theomonnom (approved):

PR #643, opened 2024-08-18
stt,tts: add timeout functionality

nbsp:
introduces a new named argument to TTS, STT, and VAD: `timeout: float`. if set to 0, it will be ignored. if not, the abstract `async __anext__()` gets wrapped in a timeout.
`STT.recognize` is not iterable so its timeout is implemented at the plugin level.

defaults to 0 to preserve backwards compatibility.


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/tts.py
```diff
@@ -47,20 +47,22 @@ def __init__(
...
         client: openai.AsyncClient | None = None,
+        timeout: float = 0,
     ) -> None:
         super().__init__(
             capabilities=tts.TTSCapabilities(
                 streaming=False,
             ),
             sample_rate=OPENAI_TTS_SAMPLE_RATE,
             num_channels=OPENAI_TTS_CHANNELS,
+            timeout=timeout,
```
theomonnom, line 58: Seems like timeout=0 isn't a good default when not using the FallbackAdapter?

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -82,7 +89,13 @@ async def aclose(self) -> None:
         self._event_ch.close()
 
     async def __anext__(self) -> SynthesizedAudio:
-        return await self._event_ch.__anext__()
+        if self._timeout > 0:
+            try:
+                return await asyncio.wait_for(self._event_ch.__anext__(), self._timeout)
```
theomonnom, line 94: asyncio.wait_for is creating a new task for each call to self._event_ch.__anext__().
It seems like we're going to leak one task when the timeout is reached

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -82,7 +89,13 @@ async def aclose(self) -> None:
...
+            except (TimeoutError, asyncio.TimeoutError) as e:
```
theomonnom, line 95:
```suggestion
            except asyncio.TimeoutError as e:
```
asyncio.TimeoutError will cover both (Python >= 3.9)

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -47,8 +47,14 @@ class STTCapabilities:
 
 
 class STT(ABC):
-    def __init__(self, *, capabilities: STTCapabilities) -> None:
+    def __init__(
+        self,
+        *,
+        capabilities: STTCapabilities,
+        timeout: float = 0,
```
theomonnom, line 54: Should we use ` timeout: float = 0`, or  `timeout: float | None = None`
cc @livekit/agent-devs for preferences, I know python use float | None for some of his methods (like asyncio.wait_for). So I would vote to do what Python is doing

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/tts.py
nbsp: we kinda have to keep it this way, timeout=0 just disables the timeout which is the way it has been. for backwards compatibility we have to default to 0

livekit-agents/livekit/agents/tts/tts.py
nbsp: if a timeout occurs then the function inside the `wait_for` gets cancelled immediately before raising `TimeoutError`, so this should be fine

livekit-agents/livekit/agents/stt/stt.py
theomonnom: let's change to `timeout: float | None = None`

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -82,7 +89,13 @@ async def aclose(self) -> None:
         self._event_ch.close()
 
     async def __anext__(self) -> SynthesizedAudio:
-        return await self._event_ch.__anext__()
+        if self._timeout > 0:
+            try:
+                return await asyncio.wait_for(self._event_ch.__anext__(), self._timeout)
+            except asyncio.TimeoutError as e:
+                raise e.__class__("synthesis timed out")
+        else:
```
theomonnom, line 97: the else statement will not be needed once timeout accepts None

theomonnom:
There is some (http) clients that don't follow the timeout provided in the constructor (e.g google and azure)

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -82,7 +89,13 @@ async def aclose(self) -> None:
         self._event_ch.close()
 
     async def __anext__(self) -> SynthesizedAudio:
-        return await self._event_ch.__anext__()
+        if self._timeout > 0:
+            try:
+                return await asyncio.wait_for(self._event_ch.__anext__(), self._timeout)
+            except asyncio.TimeoutError as e:
+                raise e.__class__("synthesis timed out")
```
theomonnom, line 96: reraise instead?

Can we also make sure we have the same ErrorType when the plugins client timeout, otherwise the users may end up with 2 different errors to catch for the same timeout errors.

nbsp:
> There is some (http) clients that don't follow the timeout provided in the constructor (e.g google and azure)

azure SDK can't set a timeout, as far as i can tell

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/tts.py
```diff
@@ -117,15 +119,23 @@ def create_azure_client(
...
+            )
+
+            return ChunkedStream(
+                stream,
+                text,
+                self._opts,
+                timeout=self._timeout,
+            )
+        except openai.APITimeoutError as e:
+            raise TimeoutError from e
```
theomonnom, line 138: This isn't waiting, the code above is sync (no await)

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -112,6 +124,7 @@ def flush(self) -> None:
         self._check_input_not_ended()
         self._check_not_closed()
         self._input_ch.send_nowait(self._FlushSentinel())
+        self._pending += 1
```
theomonnom, line 127: What is _pending used for?

nbsp: every flush pending gets incremented, and every response it gets decremented. if pending is 0, then we're not waiting for new data, and we can wait forever. if pending is positive, then something is processing, and we need to time out after a while.

nbsp: an alternative would be to have another channel, that gets input on each flush:

```python
    def flush(self) -> None:
        """Mark the end of the current segment"""
        self._check_input_not_ended()
        self._check_not_closed()
        self._input_ch.send_nowait(self._FlushSentinel())
        self._req_ch.send_nowait(None)

    async def __anext__(self) -> SynthesizedAudio:
        async for _ in self._req_ch:
            return await asyncio.wait_for(
                self._event_ch.__anext__(), self._timeout
            )
```

this is cleaner, actually. let me know if it looks good and i'll implement it

theomonnom: One thing to note is that a flush can still be done even if no data was pushed

nbsp: this doesn't matter, as long as the ratio of flushes to expected events is 1:1.

nbsp: we can move the `_req_ch.send_nowait` to the plugin side, to only send if FlushSentinel was sent after data and not by itself

nbsp:
i think the issue with google STT failing is that it takes it more than 10 seconds to warm up. setting it to 30 seems to do the trick

theomonnom (changes requested):
Can you also add timeouts to Azure? (e.g for tts https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/61d7a46a53c92bd2ac652f19a383e888158a9395/samples/python/tts-text-stream/text_stream_sample.py#L20)

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
```diff
@@ -111,6 +121,7 @@ async def _main_task(self) -> None:
         try:
             async for input in self._input_ch:
                 if isinstance(input, rtc.AudioFrame):
+                    self._req_ch.send_nowait(None)
```
theomonnom, line 124: This doesn't seems right?

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -76,14 +82,17 @@ class SpeechStream(ABC):
     class _FlushSentinel:
         pass
 
-    def __init__(self):
+    def __init__(self, *, timeout: float | None = 10.0):
         self._input_ch = aio.Chan[Union[rtc.AudioFrame, SpeechStream._FlushSentinel]]()
         self._event_ch = aio.Chan[SpeechEvent]()
+        self._req_ch = aio.Chan[None]()
```
theomonnom, line 88: req_ch is confusing. Is there a way to keep this logic only inside stt/tts base class?

for e.g inside push_frame & flush?

theomonnom: Or if this isn't possible, we could still add a private function that serves as a callback for the baseclass logic like:

self._request_started()

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -200,6 +223,7 @@ async def sentence_stream_task():
         async def input_task():
             async for data in self._input_ch:
                 if isinstance(data, self._FlushSentinel):
+                    self._req_ch.send_nowait(None)
```
theomonnom, line 226: from a user perspective implementing a tts/stt, this is unclear why this is needed

theomonnom:
For the STT, I don't think the timeout should be handled in the baseclass like we do for TTS.
The reason is that we don't know when the user stopped speaking. So if the user speaks for more than 10s and the STT doesn't support interim results. It is going to raise a TimeoutError

For the STT, I think the timeout handling is specific to each STT?

There is the timeout of the Connection (e.g Websocket). And the timeout waiting for the first transcript.
The latter is hard to correctly detect. This is actually what's happening for Google STT (bumping the timeout to 30s isn't the right solution here).

nbsp:
non-streaming STT seems to be fine, even with large frames. for streaming, you're right, we need to implement that on the plugin side as well. `ws_connect(heartbeat: ...)` works for Deepgram TTS, but both Azure and Google are accessed via SDKs, which as far as i can tell don't expose these heartbeat methods.

nbsp:
this _should_ (🤞) be ready for final review now, while i figure out two remaining issues: the [Azure](https://github.com/livekit/agents/pull/643#pullrequestreview-2266996240) timeouts, and the fact that a lot of scary (but safe!) RuntimeErrors appear when a transcription times out.

nbsp:
re-added STT timeouts:

non-streaming STTs simply time out with asyncio if they don't get the result in time.

for streaming STTs, there is a `_timeout_exc` attribute that is either `None` or `TimeoutError`, and is checked on every `__anext__()`. circumstances differ depending on implementation:
- Deepgram uses `aiohttp`'s `heartbeat` attribute for websockets. if the websocket receives `WSMsgTypeError` of type `ServerTimeoutError`, it sets `_timeout_exc` and breaks from `_main_task`
- Azure starts a timeout on every first `AudioFrame` after a `SpeechEvent` and clears it every `SpeechEvent`
- Google and `StreamAdapter` both just await the next SpeechEvent and time out accordingly

making sure a failed STT prompts "sorry, i didn't hear that" is the FallbackAdapter's job, and will be handled in that PR once this one is merged.

@livekit/agent-devs please review and nitpick

davidzhao (dismissed):
nice work!

PR #648, opened 2024-08-19
voiceassistant: keep punctuations when sending agent transcription

theomonnom:


livekit-agents/livekit/agents/tokenize/_basic_word.py
```diff
@@ -1,22 +1,35 @@
...
+    text: str, ignore_punctuation: bool = True
+) -> list[tuple[str, int, int]]:
+    """
+    Split the text into words.
+    Returns a list of words with their start and end indices of the original text.
+    """
     # fmt: off
-    punctuations = [".", ",", "!", "?", ";", ":", "'", '"', "(", ")", "[", "]", "{", "}", "<", ">",
-                    "—"]
+    punctuations = ['!', '"', '#', '$', '%', '&', "'", '(', ')', '*', '+', ',', '-', '.', '/', ':', ';', '<', '=', '>',
```
nbsp, line 12: maybe overkill but for completeness' sake maybe we should take a slice of the unicode punctuations section?

instead of checking for each of the characters we just take its code and check if it's `0x2000 <-> 0x206F`
would be shorter too

livekit-agents/livekit/agents/tokenize/_basic_sent.py
```diff
@@ -1,9 +1,13 @@
...
 
-# rule based segmentation from https://stackoverflow.com/a/31505798, works surprisingly well
-def split_sentences(text: str, min_sentence_len: int = 20) -> list[str]:
-    """the text can't contains substrings "<prd>" or "<stop>"""
+# rule based segmentation based on https://stackoverflow.com/a/31505798, works surprisingly well
+def split_sentences(
+    text: str, min_sentence_len: int = 20
+) -> list[tuple[str, int, int]]:
+    """
+    the text can't contains substrings "<prd>" or "<stop>
```
nbsp, line 9: nit
```suggestion
    the text may not contain substrings "<prd>" or "<stop>"
```

livekit-agents/livekit/agents/tokenize/_basic_word.py
theomonnom: not sure how I can make this work with str.maketrans, not a big deal, let's keep stuff explicit :)

davidzhao (approved):
when you are including punctuation, just wanted to make sure it's coming with the latter part of the token. i.e. when `hello, world`, we'd want to show `hello`, then `hello, world`

PR #649, opened 2024-08-20
use rtc with LocalTrackSubscribed, update examples

nbsp:


nbsp:
superseded by livekit/python-sdks#238

davidzhao:
we still need to use this in voice_assistant and other examples

nbsp:
yep, but we still haven't cut a release in python-sdks for that change

PR #654, opened 2024-08-22
voiceassistant: add VoiceAssistantState

lukasIO:


livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -329,10 +335,11 @@ def _link_participant(self, identity: str) -> None:
             transcription=self._opts.transcription.user_transcription,
         )
 
-        def _on_start_of_speech(ev: vad.VADEvent) -> None:
+        async def _on_start_of_speech(ev: vad.VADEvent) -> None:
```
keepingitneil, line 338: remove async, and make set_state sync (outlined abovte)

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -359,6 +372,7 @@ def _on_end_of_speech(ev: vad.VADEvent) -> None:
             self.emit("user_stopped_speaking")
             self._deferred_validation.on_human_end_of_speech(ev)
             self._last_end_of_speech_time = time.time()
+            self.set_state("thinking")
```
theomonnom, line 375: end_of_speech could still happens without necessarily thinking (e.g interruption is disable but the user is speaking).

What you can do is the check if self._playing_speech is present and not done

lukasIO: thanks! but do you think this callback is generally the right place to set the `thinking` state or is there a better place to put this?

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -301,6 +305,14 @@ async def say(
         new_handle.initialize(source=source, synthesis_handle=synthesis_handle)
         self._add_speech_for_playout(new_handle)
 
+    def set_state(self, state: VoiceAssistantState):
+        """Set the current state of the voice assistant"""
+        t = asyncio.create_task(
+            self._room.local_participant.set_attributes({"agent.state": state})
```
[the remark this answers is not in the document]
lukasIO: makes sense, updated the attribute accordingly

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -301,6 +305,16 @@ async def say(
...
+    def set_state(self, state: VoiceAssistantState):
+        """Set the current state of the voice assistant"""
+        t = asyncio.create_task(
+            self._room.local_participant.set_attributes(
+                {"voice_assistant.state": state}
+            )
+        )
+        self._state_tasks.add(t)
+        t.add_done_callback(self._state_tasks.remove)
+
```
theomonnom, line 317: Let's keep this function private. the user should not be using it (renamed to _update_state)
What we can do to avoid state-flickering, is to add a debounce parameter to this function. (delay) 


```suggestion
    def _update_state(self, state: VoiceAssistantState, delay: float = 0.0):
        """Set the current state of the voice assistant"""

        @utils.log_exceptions(logger=logger)
        async def _run_task(delay: float) -> None:
            await asyncio.sleep(delay)

            if self._room.isconnected():
                await self._room.local_participant.set_attributes(
                    {"voice_assistant.state": state}
                )

        if self._update_state_task is not None:
            self._update_state_task.cancel()

        self._update_state_task = asyncio.create_task(_run_task(delay))
```

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -201,6 +203,8 @@ def __init__(
 
         self._last_end_of_speech_time: float | None = None
 
+        self._state_tasks = set[asyncio.Task[None]]()
```
theomonnom, line 206:
```suggestion
        self._update_state_task: asyncio.Task | None = None
```

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
theomonnom: It is set to 0.0 by default, I think we should only need it before switching to the "thinking" state (because VAD can be noisy)

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
theomonnom: Yes it isn't the right place to put it. VAD can be noisy and after a end of speech. There may be no transcript be (so nothing to answer). I think you can put this inside the _synthesize_agent_reply function.

You can check if the user is still speaking there and only update if not.

lukasIO: thanks! I'll leave the debounce delay at `0.0` for now and we can adjust if we deem it to be too noisy?

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -439,6 +464,9 @@ def _synthesize_agent_reply(self) -> None:
         if self._pending_agent_reply is not None:
             self._pending_agent_reply.interrupt()
 
+        if self._human_input is None or not self._human_input.speaking:
+            self._update_state("thinking")
```
theomonnom, line 468:
```suggestion
            self._update_state("thinking", 0.2)
```

theomonnom (approved):

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -439,6 +464,9 @@ def _synthesize_agent_reply(self) -> None:
         if self._pending_agent_reply is not None:
             self._pending_agent_reply.interrupt()
 
+        if self._human_input is None or not self._human_input.speaking:
```
theomonnom, line 467:
```suggestion
        if self._human_input is not None and not self._human_input.speaking:
```

PR #678, opened 2024-08-27
update livekit to 0.14.0 and await tracksubscribed

nbsp:
awaited in voice_assistant, examples updated for tts.


livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -418,6 +418,7 @@ def _on_playout_stopped(interrupted: bool) -> None:
         agent_playout.on("playout_started", _on_playout_started)
         agent_playout.on("playout_stopped", _on_playout_stopped)
 
+        await self._agent_publication.wait_for_subscription()
```
theomonnom, line 421: this may not be the right place, as we most likely want to synthesize a speech as fast as possible.
we should only wait for this before playing (I think it should be inside the _play_speech fnc)

keepingitneil (approved):

PR #679, opened 2024-08-27
Introduce easy api for starting tasks for remote participants

keepingitneil:


livekit-agents/livekit/agents/job.py
```diff
@@ -99,12 +107,74 @@ async def connect(
...
+        filter_fnc: Callable[[rtc.RemoteParticipant], bool] | None = None,
+    ):
+        """Adds a task to be run when a participant that matches the filter joins the room. In cases where
+        the participant has already joined, the task will be run immediately. The task will only be run once
+        per matching participant. In cases where a participant leaves and then rejoins with the same identity,
+        the new task will not be started until the old task has completed.
+        """
+
+        def _filter_fnc(p: rtc.RemoteParticipant) -> bool:
+            if filter_fnc and identity:
```
davidzhao, line 141: move this outside of `_filter_fnc`? to ensure that we can raise on `add_participant_task` vs later

davidzhao: a simpler way is to avoid this inner function.. but instead to set:

```python
if filter_fnc is None:
    if identity:
        filter_fnc = lambda p: p.identity == identity
    else:
        filter_fnc = lambda p: True
```

livekit-agents/livekit/agents/job.py
```diff
@@ -99,12 +107,74 @@ async def connect(
...
+                return filter_fnc(p)
+
+            return True
+
+        async def _coro_wrapper(p: rtc.RemoteParticipant):
+            if p.identity in self._participant_tasks:
+                t = self._participant_tasks[p.identity]
+
+                try:
+                    async with asyncio.timeout(5):
```
davidzhao, line 157: could be `await.asyncio.wait_for(t, 5)` ?

examples/participant/participant_task.py
```diff
@@ -0,0 +1,29 @@
...
+    async def participant_task(p: rtc.RemoteParticipant):
+        logger.info(f"participant task starting for {p.identity}")
+        # Do something with p.attributes, p.identity, p.metadata, etc.
+        # my_stuff = await fetch_stuff_from_my_db(p)
+
+        # Do something
+        await asyncio.sleep(60)
+        logger.info(f"participant task done for {p.identity}")
+
+    await ctx.add_participant_task(task_fnc=participant_task, filter_fnc=lambda p: True)
```
davidzhao, line 25: no need to await this

livekit-agents/livekit/agents/job.py
```diff
@@ -99,12 +107,74 @@ async def connect(
...
         _apply_auto_subscribe_opts(self._room, auto_subscribe)
 
     def shutdown(self, reason: str = "") -> None:
         self._on_shutdown(reason)
 
+    def _on_participant_connected(self, p: rtc.RemoteParticipant) -> None:
+        for filter, coro in self._participant_coro_lookup.items():
+            if filter(p):
+                task = asyncio.create_task(coro(p))
+                self._participant_tasks[p.identity] = task
```
davidzhao, line 122: this confused me a bit in terms of the sequencing (i.e. intention is to run them serially vs override). Does it make sense to enforce serial dispatch for two unrelated callbacks?

imagine this use cases:

```python
async def task1(p: rtc.RemoteParticipant):
    for frame in p.audio_stream:
       pass

async def task2(p: rtc.RemoteParticipant):
    for frame in p.video_stream:
       pass

ctx.add_participant_task(task1)
ctx.add_participant_task(task2)
```

davidzhao (commented):
I like the ergonomics of ctx.add_participant_task

davidzhao:
a couple of other thoughts/questions:

1. should this be in rtc? it doesn't look like any job specific functions are used. i.e. why not `room.add_participant_task`?
2. lifecycle-wise, how does it end when either local or remote participant becomes disconnected? is there an easy way for the user to tell? or maybe the answer is just to listen to `participant_disconnected`
3. is there a way to limit only the first participant?

livekit-agents/livekit/agents/job.py
keepingitneil: spoke with theo about just removing the timeout altogether and logging a warning if the previous task is running at all

examples/participant/participant_task.py
keepingitneil: bug on my end, fixed

livekit-agents/livekit/agents/job.py
keepingitneil: Hrmm, I suppose we could actually key off of the (identity,callback) tuple and say that each different callback runs in parallel but only one of each per participant runs at a time.

keepingitneil:
> a couple of other thoughts/questions:
> 
> 1. should this be in rtc? it doesn't look like any job specific functions are used. i.e. why not `room.add_participant_task`?
> 2. lifecycle-wise, how does it end when either local or remote participant becomes disconnected? is there an easy way for the user to tell? or maybe the answer is just to listen to `participant_disconnected`
> 3. is there a way to limit only the first participant?

1. I leaned toward agents because it has some overlap `participant_(dis)connected` callbacks but behaves differently. I worry it could be confused as two different ways to do the same thing (even though they are different). A _very_ rough symmetry argument could be made with other client libs: this feels like convenience on top of the core client lib, similar to the hooks you'd find in components-js.

For 2. it's just up to the users to end the task. So, for example, a task iterating audio frames would simply exit when the participant disconnects.

3. Hrmm no not really, do you think it's needed? The user would probably know and be able to filter out participants they don't want. Of course, they could also use a closure-scoped variable to keep track of it if they really only wanted the first participant without caring about any other filtering criteria.

livekit-agents/livekit/agents/job.py
```diff
@@ -99,12 +106,41 @@ async def connect(
...
     def shutdown(self, reason: str = "") -> None:
         self._on_shutdown(reason)
 
+    def _on_participant_connected(self, p: rtc.RemoteParticipant) -> None:
+        for coro in self._participant_entrypoints:
+            if (p.identity, coro) in self._participant_tasks:
+                logger.warning(
+                    f"a participant has joined before a prior participant task matching the same identity has finished: '{p.identity}'"
+                )
+            task = asyncio.create_task(coro(p))
```
theomonnom, line 123: nit, add a taskname 

like "participant_entrypoint"

livekit-agents/livekit/agents/job.py
```diff
@@ -99,12 +106,41 @@ async def connect(
...
+            self._participant_tasks[(p.identity, coro)] = task
+            task.add_done_callback(
+                lambda _: self._participant_tasks.pop((p.identity, coro))
+            )
+
+    def add_participant_entrypoint(
+        self,
+        *,
```
theomonnom, line 131: let's allow to put it directly the callback instead of using * ?

theomonnom (approved):

davidzhao (approved):
nice!

PR #680, opened 2024-08-27
Fix Google STT exception when no valid speech is recognized

davidzhao:


keepingitneil (approved):

mike-r-mclaughlin (commented):
Nice.  Thanks for the quick fix!

PR #684, opened 2024-08-28
add threaded job runners

theomonnom:
- this is a requirement when running very large model for better devex 
     - the other option is to do IPC between a model runner process & the job processes but this is too much complexity for a user to handle.
- generally, workers running big models don't serve a lot of concurrent jobs?

This is disabled by default and should only be used with strong reasons


davidzhao (commented):
looks like tests aren't passing.. but otherwise it lg

livekit-agents/livekit/agents/ipc/job_runner.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+from typing import Any, Protocol
+
+from ..job import RunningJobInfo
+
+
+class JobRunner(Protocol):
```
keepingitneil, line 8: A little confusing reading the code that there is a `JobRunner` and a `JobExecutor`. Maybe this  becomes `JobExecutor` and the worker accepts a JobExecutor instance? For which we can have a default

theomonnom: I'm using JobExecutorType

davidzhao (approved):

PR #685, opened 2024-08-29
examples: document how to log chats

nbsp:
mostly copying the minimal assistant example, with this section added:

```python
async def log_context():
    with open("transcriptions.log", "w") as f:
        chat_ctx = assistant.chat_ctx.copy()
        for msg in chat_ctx.messages:
            f.write(msg.role + "\n")
            f.write(msg.content + "\n\n")

ctx.add_shutdown_callback(log_context)
```


examples/voice-assistant/log_transcriptions.py
```diff
@@ -0,0 +1,70 @@
...
+        pass
+
+    @assistant.on("user_speech_committed")
+    def on_user_speech_committed(msg: llm.ChatMessage):
+        # convert string lists to strings, drop images
+        if isinstance(msg.content, list):
+            msg.content = "\n".join(
+                "[image]" if isinstance(x, llm.ChatImage) else x for x in msg
+            )
+        with open("transcriptions.log", "a+") as f:
```
theomonnom, line 58: let's avoid blocking ops.
we could also just get the assistant.chat_ctx in a shutdown callback and write everything at once?

davidzhao: hmm I think we should not assume that everything will be buffered. it's better to demonstrate control where user may choose how to buffer

nbsp: so should we have both a shutdown log and a real-time log?

davidzhao: for example, what if I'd like to send transcriptions to my backend service on the fly to log? (in order to improve latency and minimize risk of losing data). 

for really long sessions (think 2h+) we also don't want to buffer everything in memory.

ideally we'd also expose timing data (by emitting them in realtime, we can basically avoid having to tag timestamps)

additionally, some users may use summarization techniques with the LLM, so the chatcontext isn't an accurate representation of each sentence being said by the user/agent.

nbsp: new commit reverts the changes, going back to realtime, and also makes it non-blocking.
i also print the current datetime in the header for each message.

examples/voice-assistant/log_transcriptions.py
```diff
@@ -0,0 +1,77 @@
...
+            f.write(f"[{datetime.now()}] USER:\n{msg.content}\n\n")
+
+    async def log_agent_speech(msg: llm.ChatMessage):
+        with open("transcriptions.log", "a+") as f:
```
theomonnom, line 62: This is still blocking the event loop.
You should use aiofile

examples/voice-assistant/log_transcriptions.py
```diff
@@ -0,0 +1,78 @@
...
+        # convert string lists to strings, drop images
+        if isinstance(msg.content, list):
+            msg.content = "\n".join(
+                "[image]" if isinstance(x, llm.ChatImage) else x for x in msg
+            )
+        async with open("transcriptions.log", "a+") as f:
+            await f.write(f"[{datetime.now()}] USER:\n{msg.content}\n\n")
+
+    async def log_agent_speech(msg: llm.ChatMessage):
+        async with open("transcriptions.log", "a+") as f:
```
davidzhao, line 63: what about using a queue here instead of opening the file each time?

IMO this would be a cleaner design:
* log_user_speech and log_agent_speech push a dict with details to a asyncio.Queue
* spin up only a single task to write to file
* the task should flush periodically and close itself when disconnected
* we can avoid creating separate tasks for these callbacks due to synchronization issues (i.e. if both things write at the same time, chatlog is corrupted)

nbsp: sure, but this isn't production code. it shows that the agent and user texts are reachable like *this* and then you can do whatever with them: log to a file, append to a database, forward to somewhere else, etc. minimal_assistant isn't a demo you're gonna run on prod, having a lot of syscalls doesn't matter here imo

examples/voice-assistant/log_transcriptions.py
```diff
@@ -0,0 +1,80 @@
...
+            )
+        log_queue.put_nowait(f"[{datetime.now()}] USER:\n{msg.content}\n\n")
+
+    @assistant.on("agent_speech_committed")
+    def on_agent_speech_committed(msg: llm.ChatMessage):
+        log_queue.put_nowait(f"[{datetime.now()}] AGENT:\n{msg.content}\n\n")
+
+    @ctx.room.on("disconnected")
+    def on_disconnected(*args):
+        log_queue.task_done()
```
theomonnom, line 66: task_done is doing nothing on its own. This is only relevant when using queue.join().

I would also prefer to use ctx.add_shutdown_callback instead of room disconnect

examples/voice-assistant/log_transcriptions.py
```diff
@@ -0,0 +1,80 @@
...
+
+    async def write_transcription():
+        async with open("transcriptions.log", "w") as f:
+            while True:
```
theomonnom, line 70: This task never ends, so the file is also never closed. I think you can use aio.Chan instead of a Queue. 
But to keep this user friendly, you can just queue "None" and break when receiving it

theomonnom: Then to make sure the task consumed everything, you can await the task inside the add_shutdown_callback

examples/voice-assistant/log_transcriptions.py
```diff
@@ -0,0 +1,80 @@
...
+                msg = await log_queue.get()
+                await f.write(msg)
+
+    asyncio.shield(asyncio.create_task(write_transcription()))
```
theomonnom, line 74: this shouldn't be needed, nobody is going to cancel this since we're not awaiting it/neither keeping a reference to it

davidzhao (approved):
lg

theomonnom (approved):

examples/voice-assistant/log_transcriptions.py
theomonnom, line 1: nit: rename the file to save_transcriptions? or save_chatctx?

PR #687, opened 2024-08-30
Add AssemblyAI Plugin

oconnoob:
Extended from PR 419 - https://github.com/livekit/agents/pull/419


livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,423 @@
...
+        buffer: AudioBuffer,
+    ) -> stt.SpeechEvent:
+        raise NotImplementedError("Not implemented")
+
+    def stream(
+        self,
+        *,
+        language: Optional[str] = None,
+    ) -> "SpeechStream":
+        config = dataclasses.replace(self._opts)
```
oconnoob, line 109: Is a copy of `self._opts` (rather than directly passing) used just to prevent future bugs?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,423 @@
...
+        self._opts = STTOptions(
+            sample_rate=sample_rate,
+            word_boost=word_boost,
+            encoding=encoding,
+            disable_partial_transcripts=disable_partial_transcripts,
+            enable_extra_session_information=enable_extra_session_information,
+            buffer_size_seconds=buffer_size_seconds,
+            token_expires_in=token_expires_in,
+            end_utterance_silence_threshold=end_utterance_silence_threshold,
+        )
```
oconnoob, line 87: Why not use dependency injection? Because higher-level class than e.g. `SpeechStream` which does use DI?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,423 @@
...
+            raise ValueError("Failed to get temporary token")
+        self._token_expires_at = time.time() + self._token_expires_in
+
+    async def _run(self, max_retry: int) -> None:
+        """
+        Run a single websocket connection to AssemblyAI and make sure to reconnect
+        when something went wrong.
+        """
+        if self._token is None or time.time() > self._token_expires_at:
+            await self._get_temporary_token()
```
oconnoob, line 201: should we move this into the while loop in case the token happens to expire during a retry?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,423 @@
...
+                    buffer_duration = total_frames / samples_per_second
+
+                    if buffer_duration >= self._buffer_size_seconds:
+                        await ws.send_bytes(bytes(buffer))
+                        buffer.clear()
+                        buffer_duration = 0.0
+                elif data == SpeechStream._CLOSE_MSG:
+                    closing_ws = True
+                    await ws.send_str(data)  # tell AssemblyAI we are done with inputs
+                    break
```
oconnoob, line 284: should we add `else` to raise Exception if there was an unexpected data type?

oconnoob (commented):
@ploeber could you review when you get a chance?

tests/test_stt.py
```diff
@@ -9,7 +9,7 @@
 
 import pytest
 from livekit import agents, rtc
-from livekit.plugins import azure, deepgram, google, openai, silero
+from livekit.plugins import assemblyai, azure, deepgram, google, openai, silero
```
ploeber, line 12: since we add this new dependency, we need to add `./livekit-plugins/livekit-plugins-assemblyai \` to the `tests.yml` file

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,423 @@
...
+        word_boost: Optional[List[str]] = None,
+        encoding: Optional[str] = "pcm_s16le",
+        disable_partial_transcripts: bool = False,
+        enable_extra_session_information: bool = False,
+        end_utterance_silence_threshold: int = 1000,
+        http_session: Optional[aiohttp.ClientSession] = None,
+        token_expires_in: int = 3600,
+        buffer_size_seconds: float = 0.2,
+    ):
+        super().__init__(streaming_supported=True)
```
ploeber, line 70: This parameter is no longer working with the newest versions. I assume the interface was changed.
This worked for me:
```python
capabilities = stt.STTCapabilities(streaming=True, interim_results=True)
super().__init__(capabilities=capabilities)
```

ploeber (changes requested):
We also need to update the README.md and add our plugin

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
ploeber: Not sure what you mean?

oconnoob: Why don't we just pass in `STTOptions` directly as we do in `SpeechStream`'s `init`?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
oconnoob: Yes, I think we should just use the API key

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,423 @@
...
+
+class SpeechStream(stt.SpeechStream):
+    _CLOSE_MSG: str = json.dumps({"terminate_session": True})
+
+    def __init__(
+        self,
+        opts: STTOptions,
+        api_key: str,
+        http_session: aiohttp.ClientSession,
+        sample_rate: int = 16000,
```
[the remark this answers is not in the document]
oconnoob: good call

oconnoob:
@ploeber the `STT` interface requires that the abstract async method `_main_task` is implemented. Currently `_main_task` exists as an instance attribute of `SpeechStream` - I'm not sure how to reconcile this

tests/test_stt.py
keepingitneil: Nice work so far, care to join a shared slack channel to help get this over the finish line? Just need your emails

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,384 @@
...
+        self._event_queue.put_nowait(end_event)
+        self._final_events = []
+
+    def _process_stream_event(self, data: dict) -> None:
+        # see this page:
+        # https://www.assemblyai.com/docs/api-reference/streaming/realtime
+        # for more information about the different types of events
+        if data["message_type"] == "SessionBegins":
+            self._speaking = True
+            start_event = stt.SpeechEvent(type=stt.SpeechEventType.START_OF_SPEECH)
```
theomonnom, line 338: Hey, since this is only being fired once, I think it is OK to not fire it at all.
I'll add a "vad" capability boolean to the STT abc in the future.
Otherwise it lgtm thanks!
We just need to add our AssemblyAI token before merging

theomonnom (approved):

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,384 @@
...
+                    ws_url = "wss://api.assemblyai.com/v2/realtime/ws"
+                    url = f"{ws_url}?{urlencode(live_config).lower()}"
+                    ws = await self._session.ws_connect(url, headers=headers)
+                    retry_count = 0  # connected successfully, reset the retry_count
+
+                    await self._run_ws(ws)
+                except Exception:
+                    # Something went wrong, retry the connection
+                    if retry_count >= max_retry:
+                        print(
```
egoldschmidt, line 199: Could we use `logging.Logger`s instead so we can get structured logging, job_id, etc for better debugging? (here and below, there are a few prints floating around)

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,384 @@
...
+        finally:
+            self._event_queue.put_nowait(None)
+
+    async def _run_ws(self, ws: aiohttp.ClientWebSocketResponse) -> None:
+        """
+        This method can throw ws errors, these are handled inside the _run method
+        """
+
+        closing_ws = False
+        END_UTTERANCE_SILENCE_THRESHOLD_MSG = json.dumps(
```
egoldschmidt, line 222: One note: on line 263 below you'll raise an error if the data going over the queue is neither an `rtc.AudioFrame` nor a `_CLOSE_MSG`, so sending this message (which is a string) causes the loop to always fail

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,384 @@
...
+
+                    retry_delay = min(retry_count * 2, 10)  # max 10s
+                    retry_count += 1  # increment after calculating the delay, the first retry should happen directly
+
+                    print(
+                        f"AssemblyAI connection failed, retrying in {retry_delay}s",
+                    )
+                    await asyncio.sleep(retry_delay)
+        except Exception:
+            print("AssemblyAI task failed")
```
egoldschmidt, line 212: Might consider just wrapping this fn in `@utils.log_exceptions`

egoldschmidt (commented):
Awesome stuff, really excited to see this contribution! I've been wanting to test out AssemblyAI so I ran the plugin locally and found a few bugs/issues. Below I left you some notes on how I got things working smoothly - hopefully this is helpful! Looking forward to this being merged, we'd love to use it.

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
oconnoob: Hey @theomonnom ! Can you clarify what you mean? This stream event puts a start event on the event queue which signifies that a websocket connection with AssemblyAI has been established and a session has begun, not that speech itself has begun

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,384 @@
...
+        if not wait:
+            self._main_task.cancel()
+
+        with suppress(asyncio.CancelledError):
+            await self._main_task
+
+        await self._session.close()
+
+    async def _main_task(self) -> None:
+        return self._run(self._max_retry)
```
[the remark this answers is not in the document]
oconnoob: good catch, thanks!

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,384 @@
...
+        Run a single websocket connection to AssemblyAI and make sure to reconnect
+        when something went wrong.
+        """
+        try:
+            retry_count = 0
+            while not self._closed:
+                try:
+                    live_config = {
+                        "sample_rate": self._opts.sample_rate,
+                        "word_boost": self._opts.word_boost,
```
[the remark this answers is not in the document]
oconnoob: Added a step to filter all key-value pairs with a value of `None` - thanks!

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
oconnoob: fixed ✅ thanks!

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
oconnoob: fixed, thanks!

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
oconnoob: good point, done!

oconnoob:
@ploeber could you test the current changes when you find some time tomorrow?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,395 @@
...
+        self._queue.put_nowait(frame)
+
+    async def aclose(self, *, wait: bool = True) -> None:
+        self._closed = True
+        self._queue.put_nowait(SpeechStream._CLOSE_MSG)
+
+        if not wait:
+            self._main_task.cancel()
+
+        with suppress(asyncio.CancelledError):
```
ploeber, line 168: must be `contextlib.suppress`, and contextlib must be imported

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,395 @@
...
+    async def _run_ws(self, ws: aiohttp.ClientWebSocketResponse) -> None:
+        """
+        This method can throw ws errors, these are handled inside the _run method
+        """
+
+        closing_ws = False
+        END_UTTERANCE_SILENCE_THRESHOLD_MSG = json.dumps(
+            {"end_utterance_silence_threshold": self._opts.end_utterance_silence_threshold }
+        )
+        self._queue.put_nowait(END_UTTERANCE_SILENCE_THRESHOLD_MSG)
```
ploeber, line 234: Only do this if the threshold is not None. Otherwise we can use the default settings from our API

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,395 @@
...
+                interim_results=True,)
+            )
+        api_key = api_key or os.environ.get("ASSEMBLYAI_API_KEY")
+        if api_key is None:
+            raise ValueError("AssemblyAI API key is required. " \
+                             "Pass one in via the `api_key` parameter, " \
+                             "or set it as the `ASSEMBLYAI_API_KEY` environment variable")
+        self._api_key = api_key
+
+        self._opts = STTOptions(
```
ploeber, line 80: maybe also add a check that encoding is one of the 2 allowed values, because later you try to access the dictionary with those values

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,395 @@
...
+        if not self._speaking:
+            logger.info(
+                "trying to commit final events without being in the speaking state",
+            )
+            return
+
+        if len(self._final_events) == 0:
+            return
+
+        self._speaking = False
```
ploeber, line 315: should we reset this variable before the `if len(self._final_events) == 0:` above?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,395 @@
...
+
+        # combine all final transcripts since the start of the speech
+        sentence = ""
+        confidence = 0.0
+        for f in self._final_events:
+            alt = f.alternatives[0]
+            sentence += f"{alt.text.strip()} "
```
ploeber, line 322: nit pick but I think this can be simplified to
`sentence = " ".join(f.alternatives[0].text.strip() for f in self._final_events)`
and then you don't need the rstrip below.
Even if you need a second loop for the confidence, I think this solution is more efficient. You could even use the `sum` function to have another one-liner for the confidence

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,395 @@
...
+            alts = live_transcription_to_speech_data(ENGLISH, data)
+            if len(alts) > 0 and alts[0].text:
+                final_event = stt.SpeechEvent(
+                    type=stt.SpeechEventType.FINAL_TRANSCRIPT,
+                    alternatives=alts,
+                )
+                self._final_events.append(final_event)
+                self._event_queue.put_nowait(final_event)
+                self._end_speech()
+        elif data["message_type"] == "RealtimeError":
```
ploeber, line 372: handling of session_terminated and session_information events is missing. See https://github.com/AssemblyAI/assemblyai-python-sdk/blob/master/assemblyai/transcriber.py#L1318

ploeber (changes requested):
Additionally:
- package.json is missing
- plugin needs to be added to livekit-plugins/install_plugins_editable.sh

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -0,0 +1,395 @@
...
+        elif data["message_type"] == "FinalTranscript":
+            alts = live_transcription_to_speech_data(ENGLISH, data)
+            if len(alts) > 0 and alts[0].text:
+                final_event = stt.SpeechEvent(
+                    type=stt.SpeechEventType.FINAL_TRANSCRIPT,
+                    alternatives=alts,
+                )
+                self._final_events.append(final_event)
+                self._event_queue.put_nowait(final_event)
+                self._end_speech()
```
ploeber, line 371: I don't think this is correct. We don't want to end the speech for every final transcript. a final transcript can be just a final sentence but not the end of the speech

ploeber (approved):

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
oconnoob: added `post_init` to the dataclass

oconnoob:
@ploeber addressed the rest of your comments 🙏

theomonnom:
Hey, can  ryan-assemblyai sign the CLI?

oconnoob:
@theomonnom signed!

oconnoob:
When I click on the CLA it says `You have agreed to the CLA for livekit/agents` - not sure why it's not updating

davidzhao:
@oconnoob ah I think it picked up on your other handle. this is fine IMO.

davidzhao:
@oconnoob last thing, could you run:

```
ruff format . && ruff check --fix .
```

just to get rid of these formatting warnings

oconnoob:
@davidzhao all set!

davidzhao:
Thanks @oconnoob!

Issue #692, opened 2024-09-02
llamaindex example


ryanleecode:
is there an integration that uses llamaindex vectors and chatengine directly?

theomonnom:
Hey, I added a bunch of examples in this [PR](https://github.com/livekit/agents/pull/696)

ryanleecode:
> Hey, I added a bunch of examples in this [PR](https://github.com/livekit/agents/pull/696)

You're amazing! 🤩

How would i choose between `query engine` versus `retrieval`? they both try to do the same thing right but in differents ways. could they even be combined?

theomonnom:
Hey, I think it depends, I'm not sure at 100% but you can definitely combine both. You need to experiment by trial and error.

rswearin:
You rock. These examples are great. Thanks.

PR #696, opened 2024-09-03
livekit-plugins-llama-index

theomonnom:


theomonnom:
https://github.com/livekit/agents/issues/692

livekit-plugins/livekit-plugins-llama-index/README.md
```diff
@@ -0,0 +1,14 @@
+# LiveKit Plugins Minimal
+
+This is a minimal example of a LiveKit plugin for Agents.
+
+### Developer note
```
davidzhao, line 5: let's also fix this readme

theomonnom:
> the chat engine example is kind of busted because chat engine doesnt allow for concurrency. it's chat memory is shared for all connections.

Interesting, even if the chat_history is always set here (so using it instead of the internal memory?)

https://github.com/livekit/agents/pull/696/files#diff-c9c5aab8143dd7a1c7599a1791904401561e31b50f20b0dc4f836752c80541a2R64

ryanleecode:
I think so because in their implementation its a shared memory buffer

https://github.com/run-llama/llama_index/blob/4e8d3d8cabfea5a2289c01fb9d435a50cdb8c360/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py#L244

it can still work but you'd have to create a new chat engine per room

theomonnom:
> I think so because in their implementation its a shared memory buffer
> 
> https://github.com/run-llama/llama_index/blob/4e8d3d8cabfea5a2289c01fb9d435a50cdb8c360/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py#L244
> 
> it can still work but you'd have to create a new chat engine per room

Oh, creating a new ChatEngine per room is actually the way I was implementing it.
Can you give me more details about what you're doing? Is the MemoryBuffer saved somewhere?

examples/voice-assistant/llamaindex-rag/chat_engine.py
```diff
@@ -0,0 +1,56 @@
...
+    documents = SimpleDirectoryReader("data").load_data()
+    index = VectorStoreIndex.from_documents(documents)
+    # store it for later
+    index.storage_context.persist(persist_dir=PERSIST_DIR)
+else:
+    # load the existing index
+    storage_context = StorageContext.from_defaults(persist_dir=PERSIST_DIR)
+    index = load_index_from_storage(storage_context)
+
+chat_engine = index.as_chat_engine(chat_mode=ChatMode.CONTEXT)
```
ryanleecode, line 30: isnt this outside though?

ryanleecode: it would need to be created in the entrypoint function so its not shared across rooms if my undersranding is correct

keepingitneil (approved):

PR #706, opened 2024-09-04
voiceassistant: add before_tts_cb callback

theomonnom:
- deprecate callbacks with `will_*` naming (now using the before `keyword` instead) — added deprecation warnings
- added `aio.itertools` utilities
- add `before_tts_cb` callback on the voice assistant
- added tokenize.utils to easily replace words

### DIFF
--- a/examples/voice-assistant/custom_pronunciation.py
+++ b/examples/voice-assistant/custom_pronunciation.py
@@ -1,0 +1,52 @@
+from __future__ import annotations
+from typing import AsyncIterable
+import asyncio
+
+from dotenv import load_dotenv
+from livekit import rtc
+from livekit.agents import AutoSubscribe, JobContext, WorkerOptions, cli, llm, utils
+from livekit.agents.voice_assistant import VoiceAssistant
+from livekit.plugins import deepgram, cartesia, openai, silero
+
+load_dotenv()
+
+
+async def entrypoint(ctx: JobContext):
+    initial_ctx = llm.ChatContext().append(
+        role="system",
+        text=(
+            "You are a voice assistant created by LiveKit. Your interface with users will be voice. "
+            "You should use short and concise responses, and avoiding usage of unpronouncable punctuation."
+        ),
+    )
+
+    await ctx.connect(auto_subscribe=AutoSubscribe.AUDIO_ONLY)
+
+    async def _will_synthesize_assistant_speech(
+        assistant: VoiceAssistant, text: str | AsyncIterable[str]
+    ):
+        # Cartesia TTS is incorrectly pronouncing "LiveKit", so we'll replace it with a phonetic
+        # spelling
+        return utils.replace_words(
+            text=text, replacements={"LiveKit": "<<L|ˈaɪ|ve>>Kit"}
+        )
+
+    # also for this example, we also intensify the keyword "LiveKit" to make it more likely to be
+    # recognized with the STT
+    deepgram_stt = deepgram.STT(keywords=[("LiveKit", 2.0)])
+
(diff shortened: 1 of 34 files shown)


examples/voice-assistant/custom_pronunciation.py
```diff
@@ -0,0 +1,52 @@
...
+            "You should use short and concise responses, and avoiding usage of unpronouncable punctuation."
+        ),
+    )
+
+    await ctx.connect(auto_subscribe=AutoSubscribe.AUDIO_ONLY)
+
+    async def _will_synthesize_assistant_speech(
+        assistant: VoiceAssistant, text: str | AsyncIterable[str]
+    ):
+        # Cartesia TTS is incorrectly pronouncing "LiveKit", so we'll replace it with a phonetic
```
davidzhao, line 28: nit: maybe just indicate this in general, versus a Cartesia specific issue.

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -145,6 +160,9 @@ def __init__(
             transcription: Options for assistant transcription.
             will_synthesize_assistant_reply: Callback called when the assistant is about to synthesize a reply.
```
davidzhao, line 161: nit. these events tend to confuse me a bit. since we are adding a new one, I wonder if there's a better way to handle these.
and should we use callbacks rather than events? 

i.e. assistant.llm_preprocess_cb=myfunc and assistant.llm_postprocess_cb

where both functions can manipulate chat context, and postprocess could also determine what goes into TTS

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -17,11 +18,20 @@
...
-WillSynthesizeAssistantReply = Callable[
+
+PreLLMGenerationCallback = Callable[
     ["VoiceAssistant", ChatContext],
     Union[Optional[LLMStream], Awaitable[Optional[LLMStream]]],
 ]
 
+WillSynthesizeAssistantReply = PreLLMGenerationCallback
+
+PreTTSGenerationCallback = Callable[
```
davidzhao, line 29: should this be something like?

```
Union[
  Callable[["VoiceAssistant", str], str],
  Callable[["VoiceAssistant", AsyncIterable[str]], AsyncIterable[str]],
]
```

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -143,21 +163,34 @@ def __init__(
...
+                (e.g: editing the pronunciation of a word).
             plotting: Whether to enable plotting for debugging. matplotlib must be installed.
             loop: Event loop to use. Default to asyncio.get_event_loop().
         """
         super().__init__()
         self._loop = loop or asyncio.get_event_loop()
+
+        if will_synthesize_assistant_reply is not None:
+            warnings.warn(
+                "will_synthesize_assistant_reply is deprecated and will be removed in 1.5.0, use on_generate_llm instead",
```
davidzhao, line 179:
```suggestion
                "will_synthesize_assistant_reply is deprecated and will be removed in 1.5.0, use before_llm_cb instead",
```

livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
theomonnom: It is different, we're OK receiving AsyncIterable[str] and finally returning a str for e.g

davidzhao (approved):
🚀🚀

PR #712, opened 2024-09-05
add JobContext.wait_for_participant

theomonnom:


livekit-agents/livekit/agents/job.py
```diff
@@ -91,6 +91,35 @@ def add_shutdown_callback(
     ) -> None:
         self._shutdown_callbacks.append(callback)
 
+    async def wait_for_participant(
+        self, *, identity: str | None = None
```
keepingitneil, line 95: maybe an optional filter fnc would be useful here (for example to filter in/out based on things other than identity)

davidzhao (approved):
let's land this and can always add lambda filter later on.

PR #721, opened 2024-09-08
Enable Google TTS with application default credentials

rsinnet:
Enable the use of the Google TTS plugin via application default
credentials, where no credential info is provided explicitly,
and instead is default via the default resolution order.

This enables two other common configurations including
deployment on Google Cloud where application
credentials are injected via the VM Metadata Service or on
developer systems where credentials are stored at
`~/.config/gcloud/application_default_credentials.json`
without setting the `GOOGLE_APPLICATION_CREDENTIALS`
environment variable.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
@@ -112,7 +112,11 @@ class STT(stt.STT):
         # recognizers may improve latency https://cloud.google.com/speech-to-text/v2/docs/recognizers#understand_recognizers
 
         # TODO(theomonnom): find a better way to access the project_id
-        project_id = self._ensure_client().transport._credentials.project_id  # type: ignore
+        try:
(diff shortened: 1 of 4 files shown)


theomonnom (approved):
Thanks! TIL

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -112,7 +117,11 @@ def _recognizer(self) -> str:
         # recognizers may improve latency https://cloud.google.com/speech-to-text/v2/docs/recognizers#understand_recognizers
 
         # TODO(theomonnom): find a better way to access the project_id
-        project_id = self._ensure_client().transport._credentials.project_id  # type: ignore
+        try:
+            project_id = self._ensure_client().transport._credentials.project_id  # type: ignore
+        except AttributeError:
+            from google.auth import default as ga_default
+            _, project_id = ga_default()
```
theomonnom, line 124: 🙏🙏

PR #737, opened 2024-09-10
Adding UtteranceEnd support for Deepgram

hackyon:
Adding support for UtteranceEnd to Deepgram agent plugin. The UtteranceEnd is another feature offered by Deepgram to detect end of speech.

The Deepgram documentation is here:
https://developers.deepgram.com/docs/utterance-end

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -45,6 +45,7 @@ class STTOptions:
     smart_format: bool
     no_delay: bool
     endpointing_ms: int
+    utterance_end_ms: int
     filler_words: bool
     sample_rate: int
     num_channels: int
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -63,6 +64,7 @@ def __init__(
         smart_format: bool = True,
         no_delay: bool = True,
         endpointing_ms: int = 25,
+        utterance_end_ms: int | None,
```
theomonnom, line 67: Hey, can you default it to None?

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -391,6 +401,12 @@ def _process_stream_event(self, data: dict) -> None:
 
         elif data["type"] == "Metadata":
             pass  # metadata is too noisy
+        elif data["type"] == "UtteranceEnd":
```
theomonnom, line 404: Is it being sent each time we finished speaking? If so we should remove the previous dispatch of END_OF_SPEECH to avoid duplicated events. (only when end_of_utterance_ms is enabled)

PR #739, opened 2024-09-10
ci: don't run tests on forks

nbsp:


theomonnom (approved):

theomonnom:
Is it possible to use
`if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}`
at the job level? instead of every steps (not sure if there is a way to do it)
otherwise lgtm

keepingitneil (approved):

nbsp:
> Is it possible to use `if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}` at the job level? instead of every steps (not sure if there is a way to do it) otherwise lgtm

doesn't seem like it, no. we can have a noop job that all other jobs depend on, if that's better

theomonnom:
> doesn't seem like it, no. we can have a noop job that all other jobs depend on, if that's better
let's merge, not a big deal

PR #740, opened 2024-09-10
Add support for cartesia voice control

bcherry:
This adds basic support for initializing the Cartesia TTS plugin with their [Speed & Emotion controls](https://docs.cartesia.ai/user-guides/voice-control).

I also improved the docs and added a full Cartesia TTS example.

Not implemented here: changing the voice control parameters on the fly during a session.


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -53,15 +54,25 @@ def __init__(
         language: str = "en",
         encoding: TTSEncoding = "pcm_s16le",
         voice: str | list[float] = TTSDefaultVoiceId,
+        voice_controls: dict[str, any] | None = None,
```
keepingitneil, line 57: Is it possible to be more specific about these types?

bcherry: yes - with a new type definition. it looks like `"speed"` can be a signed number or a string, and `"emotion"` is an array of signed numbers or strings. No other keys currently supported. so we could add `speed: str | float | None` and `emotion: list[str | float] | None` instead of `voice_controls`, or we could define a new container type for them.

davidzhao: I vote for explicit options with types. While it does require us to keep it up to date, it would help to avoid confusion when users put in incorrect values/types

bcherry: Yep makes sense. Updated to break them up. Speed got a proper type since only five strings are valid (or any number -1 to 1) but emotion didn't because each string is both an emotions and a degrees so there are 25 valid strings and I didn't think it made sense to enumerate.

I also added this to the test cases!

theomonnom (approved):

theomonnom:
one annoying thing is that we could also do `positivity:high`
but not a big deal, we can merge in the current state

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -53,15 +55,27 @@ def __init__(
         language: str = "en",
         encoding: TTSEncoding = "pcm_s16le",
         voice: str | list[float] = TTSDefaultVoiceId,
+        speed: TTSVoiceSpeed | float | None = None,
+        emotion: list[TTSVoiceEmotion] | None = None,
```
theomonnom, line 59: Nice, let's still allow str in case there is new ones 
```suggestion
        emotion: list[TTSVoiceEmotion | str] | None = None,
```

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -41,6 +41,8 @@ class _TTSOptions:
     encoding: TTSEncoding
     sample_rate: int
     voice: str | list[float]
+    speed: TTSVoiceSpeed | float | None
+    emotion: list[TTSVoiceEmotion] | None
```
theomonnom, line 45:
```suggestion
    emotion: list[TTSVoiceEmotion | str] | None
```

theomonnom:
To fix the ruff CI you can do:
`ruff format && ruff check --fix`

and to fix the Check Types CI, just need to read the errors lol

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -274,6 +290,15 @@ def _to_cartesia_options(opts: _TTSOptions) -> dict[str, Any]:
     else:
         voice["mode"] = "embedding"
         voice["embedding"] = opts.voice
+        
+    voice_controls = {}
```
theomonnom, line 294: doing that will fix the Check Types CI
```suggestion
    voice_controls: dict = {}
```

bcherry:
thank you @theomonnom!

davidzhao (approved):
nice!

PR #741, opened 2024-09-11
Allow setting LLM temperature with VoiceAssistant

### LINKED ISSUE #718: allow adjusting LLM temperature when using VoiceAssistant
Currently it's not possible to adjust this LLM setting per session. Ideally this is something that a user could set on the LLM instance, as well as being able to change it dynamically mid-session

davidzhao:
Also including LLMOptions for future settings that should tune subsequent LLM chat completions

closes #718

### DIFF
--- a/livekit-agents/livekit/agents/llm/llm.py
+++ b/livekit-agents/livekit/agents/llm/llm.py
@@ -29,6 +29,10 @@ class ChatChunk:
 
 
 class LLM(abc.ABC):
+    def __init__(self, *, temperature: float | None = None):
+        # set default temperature, used by `chat` method when temperature isn't provided
+        self._temperature = temperature
+
     @abc.abstractmethod
     def chat(
         self,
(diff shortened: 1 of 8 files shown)


livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -29,6 +29,10 @@ class ChatChunk:
 
 
 class LLM(abc.ABC):
+    def __init__(self, *, temperature: float | None = None):
+        # set default temperature, used by `chat` method when temperature isn't provided
+        self._temperature = temperature
```
theomonnom, line 34: If possible I would have created a private dataclass called `_LLMOptions`. But feel free to ignore since we only expose temperature for now

theomonnom (approved):
lgtm

livekit-agents/livekit/agents/llm/llm.py
davidzhao: hmm that's true.. we already use LLMOptions for model and user.. I'll add to that

davidzhao: tho the current design they are private to OpenAI/Anthropic.

PR #753, opened 2024-09-13
voiceassistant: allow to cancel llm generation inside before_llm_cb

theomonnom:


davidzhao (commented):
lg, can we document this behavior?

davidzhao (approved):

PR #754, opened 2024-09-14
Frequency and Presence Penalty

PatroxGaurab:


theomonnom (approved):
lgtm, thanks!

theomonnom (changes requested):
It seems like the arguments aren't passed to the openai chat method.

PatroxGaurab:
It has been passed in a different commit within the same PR -> https://github.com/livekit/agents/commit/ad2b1e6d5c16eb66d38dad9102dbb66a12a3e95c

PatroxGaurab (commented):
Already present in this commit. pls review.

davidzhao:
You'd also need to pass it in [here](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py#L404), otherwise setting it to `_opts` is a no-op

PR #756, opened 2024-09-15
improve graceful shutdown 

theomonnom:


rektdeckard (approved):
lg! 🙏

davidzhao (commented):
is the behavior that Ctrl+C wouldn't do anything the second time? is there a way we could make it shutdown immediately?

PR #762, opened 2024-09-17
better dev defaults

theomonnom:
in dev mode:
 - don't prewarm processes 
 - http server defaults to a random available port
 - default log level to debug
 - disable load threshold for capacity (always accept)

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -142,11 +142,11 @@ class WorkerOptions:
     job_executor_type: JobExecutorType = _default_job_executor_type
     """Which executor to use to run jobs. (currently thread or process are supported)"""
     load_threshold: float | _WorkerEnvOption[float] = _WorkerEnvOption(
-        dev_default=math.inf, prod_default=0.8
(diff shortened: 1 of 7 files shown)


livekit-agents/livekit/agents/worker.py
```diff
@@ -125,9 +141,16 @@ class WorkerOptions:
...
     """Which executor to use to run jobs. (currently thread or process are supported)"""
-    load_threshold: float = 0.65
-    """When the load exceeds this threshold, the worker will be marked as unavailable."""
-    num_idle_processes: int = 3
+    load_threshold: float | _WorkerEnvOption[float] = _WorkerEnvOption(
+        dev_default=math.inf, prod_default=0.8
+    )
+    """When the load exceeds this threshold, the worker will be marked as unavailable.
+    
+    Defaults to 0.8 on "production" mode, and is disabled in "development" mode.
```
davidzhao, line 149: `0.75`? we can raise a bit, but IMO still better to not run too hot.

davidzhao (approved):
lg!

PR #776, opened 2024-09-18
Add transcription type to AssistantTranscriptionOptions

lukasIO:
Note: this is still missing an implementation, this PR is in draft mode to discuss how we want to expose the option. 
Once we settled on how it should look, I'll add the implementation and mark the PR as ready for review.


bcherry (commented):
makes sense

PR #790, opened 2024-09-24
Fix the problem of infinite loop when agent speech is interrupted

coderlxn:
When playing the Agent's speech, if the user interrupts, the while loop here will become an infinite loop and needs to be exited when interrupted.


livekit-agents/livekit/agents/voice_assistant/voice_assistant.py
```diff
@@ -640,6 +640,10 @@ def _commit_user_question_if_needed() -> None:
             )
 
             _commit_user_question_if_needed()
+            
+            if speech_handle.interrupted:
+                logger.warning(f"speech interrupted, quit loop")
```
theomonnom, line 645: We're already logging interrupted status at the end of _play_speech 
```suggestion
```

theomonnom (approved):
Nice catch! Thanks 🙏

PR #794, opened 2024-09-25
Add support for Azure TTS custom voices (AGT-445)

mike-r-mclaughlin:
Added an endpoint_id property to the TTS() constructor to allow for passing the Azure endpoint_id.  When combined with the voice property, this allows customers to use a custom voice.

Also updated the exception message to get more details when the speak_text_async() is cancelled.  It was challenging to debug the issue without these details.


davidzhao:
@mike-r-mclaughlin can you run `ruff format . && ruff check --fix .` to get rid of the formatting issues? Thanks!

davidzhao:
> @mike-r-mclaughlincan you run `ruff format . && ruff check --fix .`to get rid of the formatting issues? Thanks!

nvm, this was from a diff PR.. I've fixed it here.

davidzhao (approved):

PR #800, opened 2024-09-26
examples: added trigger-phrase agent example

s-hamdananwar:


examples/trigger-phrase/agent.py
```diff
@@ -0,0 +1,166 @@
...
+load_dotenv()
+
+logger = logging.getLogger("my-worker")
+logger.setLevel(logging.INFO)
+
+word_tokenizer_without_punctuation: tokenize.WordTokenizer = (
+    tokenize.basic.WordTokenizer(ignore_punctuation=True)
+)
+
+trigger_phrase = "Hi Bob!"
```
nbsp, line 28: nit: `TRIGGER_PHRASE` instead to show that this is a changeable constant

examples/trigger-phrase/agent.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+    options = rtc.TrackPublishOptions()
+    options.source = rtc.TrackSource.SOURCE_MICROPHONE
+    publication = await ctx.room.local_participant.publish_track(agent_track, options)
+
+    # setup LLM
+    initial_ctx = llm.ChatContext().append(
+        role="system",
+        text=(
+            f"You are {trigger_phrase}, a voice assistant created by LiveKit. Your interface with users will be voice. "
```
nbsp, line 129: weird misleading use of trigger_phrase here. this implies that it can only be used as a name, i think it's best to drop it

examples/trigger-phrase/agent.py
```diff
@@ -0,0 +1,166 @@
...
+                elif i == len(trigger_phrase_words) - 1:
+                    # trigger phrase is validated
+                    new_chat_context = llm_stream.chat_ctx.append(
+                        text=ev.alternatives[0].text
+                    )
+                    llm_stream = local_llm.chat(chat_ctx=new_chat_context)
+                    llm_reply_stream = _llm_stream_to_str_iterable(llm_stream)
+                    async for seg in llm_reply_stream:
+                        tts_stream.push_text(seg)
+                    tts_stream.flush()
```
nbsp, line 86: what happens if neither of these if statements are true?

nbsp (commented):
a few comments, plus run `ruff check . && ruff format .` for linting

examples/trigger-phrase/agent.py
s-hamdananwar: oh okay, didn't know that - thanks!

examples/trigger-phrase/agent.py
s-hamdananwar: yeah I agree. I was also debating about this but my thought was that it might be helpful to give the LLM a bit more context

examples/trigger-phrase/agent.py
s-hamdananwar: it would go to the next word in the `new_transcribed_text_words` and continues until all the words in `trigger_phrase_words` are validated, at least that's my assumption. please let me know if I missed something in that logic!

examples/trigger-phrase/README.md
```diff
@@ -0,0 +1,23 @@
+# Trigger Phrase Initiated Agent
+
+This example demonstrates an agent that only responds to the user queries if the user provided trigger phrase is validated in the beginning of a speech.
+
+The trigger phrase can be edited by changing this line:
+
+```
+trigger_phrase = "Hi Bob!"
```
nbsp, line 8: update here as well

examples/trigger-phrase/agent.py
```diff
@@ -0,0 +1,165 @@
...
+        yield content
+
+
+async def entrypoint(ctx: JobContext):
+    logger.info("starting trigger-phrase agent example")
+
+    vad = silero.VAD.load(
+        min_speech_duration=0.01,
+        min_silence_duration=0.5,
+    )
```
nbsp, line 104: you should have this be in the prewarm function so it doesn't block the job from starting

nbsp (approved):
looks good! i noticed a few things:
- there's no STT transcriptions in chat, can you add those?
- it's a bit slow. worth looking into
- semantically this should probably be inside the `voice_assistant` examples directory

examples/trigger-phrase/README.md
s-hamdananwar: oops good catch, thanks

examples/trigger-phrase/agent.py
s-hamdananwar: oope I see it in the docs now, should have read it better 🤦‍♂️

s-hamdananwar:
> * it's a bit slow. worth looking into

I think it is mainly due to the 0.5 sec timeout set for the VAD, and maybe partly due to the computation that needs to happen on every END_OF_SPEECH event. I am not sure the best way to address them though. Since the primary goal of this example is to show the users a way to use transcribed words to trigger the LLM, I didn't go down the path of ensuring minimum possible latency like `VoiceAssistant` does. 

> * semantically this should probably be inside the `voice_assistant` examples directory

Even though _technically_ this is a voice assistant, since we are not using the VoiceAssistant class, I feel like it would be confusing and counter intuitive to the users if we placed in that directory and hence resorted to a stand alone example directory. What do you think?

nbsp:
> I think it is mainly due to the 0.5 sec timeout set for the VAD, and maybe partly due to the computation that needs to happen on every END_OF_SPEECH event.

in my testing i encountered closer to three or sometimes four seconds of silence before the response started playing. this doesn't need to be _fully_ optimized as an example, but at this point it is hurting the effectiveness of the demo.

re: directory, disregard; did not notice this doesn't actually use VoicePipelineAgent.

s-hamdananwar:
- STT transcriptions is now added ✅ 
- VAD is removed, both due to issues with adding StreamAdapter to Deepgram and also hopefully to reduce latency
- first_participant constraint removed

dsgolman:
@s-hamdananwar this is how I was able to manage "multiple" participants in a single raise hand queue, check out the PR and let me know if this can help resolve the issue of still only listening to the first participant that joins the room. 

[PR](https://github.com/livekit/agents/pull/971/)

PR #803, opened 2024-09-27
Add Telnyx integration for LLM

jamestwhedbee:
Adds integration with [Telnyx Inference](https://developers.telnyx.com/docs/inference/getting-started) to the OpenAI plugin, at which point we can also update our [LiveKit Voice Assistant Integration guide](https://developers.telnyx.com/docs/inference/livekit)


theomonnom (approved):
awesome! Thanks!

PR #806, opened 2024-09-28
Speech to Video

mrdrprofuroboros:
Hey LiveKit team!

I'm working on an animated agent that is lipsynced with the TTS-generated audio. So I decided it would be good to contribute back to you guys, because LiveKit is awesome and I'm really grateful you're doing it open source.

Here's a draft that I wrote more like a proposal, happy to get your comments and improve it, I really think speech-to-video and animated avatars is something that could be reused in the community

So 10000ft view of what's going on here:
1) I'm using 11labs streaming alignment to syncronize video with the TTS: https://elevenlabs.io/docs/api-reference/websockets#streaming-output-audio
2) now it works only with 11labs streaming API, but can be easily modified to support their chunked stream as well
3) the rest of the TTS unfortunately doesn't provide char alignment, but we can come up with some hacks similar to what you've done with transcript forwarding. might be enough for simple animated faces
4) I've added "alignment" capability to the TTS base class (but not really used it yet lol)
5) Next I've added video_playout_source to AgentPlayout and video buffer to AgentOutput
6) I syncronize audio and video streams in the AgentPlayout. audio stream is a primary source for sync rn (and I don't quite like it because if someone decides to use a more sophisticated model to generate face from audio+transcript, we'd have a delay between first audio chunk and first video frame and video will have to catch up with audio during agent speech, which might look off)
7) so the STV (speech-to-video) class has two streams it operates: SpeechStream (for lipsync frames) and IdleStream (for animations in between utterances. I wasn't sure here about the right approach (comments are very welcome), but the SpeechStream generates timestamped rtc.VideoFrameEvent sequences (that base off timestamp offsets from the start of the audio to match alignment timings) while the IdleStream just generates rtc.VideoFrames at it's own pace (since there's no good start \ sync point, but might be the end of the last utterance)
8) There's a very crude example that blinks random solid squares while idle and spits out syncronized letters on white background one-by-one when agent speaks
9) Some next steps would be to add or modify your WordStreamer\hyphenator to be able to consume timestamped char alignment streams and convert them to a timestamped stream of phonemes/visemes, making an adapter for other TTS engines that doesn't support timestamps, maybe syncronizing tts_forwarder during the playout time and not generation time as it is now and reusing alignment there also with the help of those timestamped text streaming adapters
10) it was a little had to wrap my head around your way of passing handlers into handlers to stream streams from streams lol. I've added a bunch of TODOs here and there as hooks for discussion or places to continue, this whole thing needs a thorough review

ok, I hope you could help me out here to pull it up to par with the rest of your codebase if this feels useful.
P.S. I'll need multimodal input to vision transformers, do you have any plans there? do you need help?

### DIFF
--- a/examples/speech-to-video/README.md
+++ b/examples/speech-to-video/README.md
@@ -1,0 +1,3 @@
+# Speech-to-video
+
+This small exmple publishes a random solid color video frame when idle, and a stream of words when the agent is speaking.
--- a/examples/speech-to-video/agent.py
+++ b/examples/speech-to-video/agent.py
@@ -1,9 +1,6 @@
 import asyncio
 import logging
-import os
 import random
-import string
-import sys
 
 from dotenv import load_dotenv
 from livekit import rtc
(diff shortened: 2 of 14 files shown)


ChenghaoMou:
Commenting for support. This will be a very interesting addition to the framework.

My use case is where I am generating video from audio like the one in [EchoMimic](https://github.com/BadToBest/EchoMimic). So based on the design here, as you pointed out in \#6, it would create a lagging effect considering video generation can be much slower than audio generation.

PR #811, opened 2024-09-30
add(plugin/deepgram): accepts parameter profanity_filter

jebjebs:
Deepgram supports `profanity_filter` which recognizes profanity and converts it to the nearest recognized non-profane word or removes it from the transcript completely.
https://developers.deepgram.com/docs/profanity-filter

Adds support for this option in `STTOptions` and forwards the option in the live_config


theomonnom (approved):
lgtm, thanks!

PR #826, opened 2024-10-03
Fix tool call output type error

KillianLucas:
Solves #827.

In `realtime_model.py`, tool call outputs (`message_content`) are not strings by the time they reach the assertion `assert isinstance(message_content, str)`. 

This PR handles tool calls before other message types, which depend on `message_content` being a list. You can also just replace `message_content` with `message.content`, and keep the original order— both fixes work.

Tested according to the [minimal Multimodel Agent + Function calling example here](https://docs.livekit.io/agents/voice-agent/function-calling/).

Also: extraordinary work, Livekit team! Thank you for everything you do. ❤️ and ofc no need to solve it this way / merge if there's another fix you're working on.

### DIFF
--- a/.changeset/fresh-mayflies-shout.md
+++ b/.changeset/fresh-mayflies-shout.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-openai": patch
+---
+
+oai-realtime: fix function calls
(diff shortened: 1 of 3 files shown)


theomonnom (approved):
Thanks!!

PR #833, opened 2024-10-03
Fix tests

keepingitneil:
event loops were shared across calls function calls because of the test fixture so we switch to a factory pattern


davidzhao (approved):

theomonnom (approved):
🔥

Issue #825, opened 2024-10-03
If an agent calls a function before synthesizing audio, it gets stuck in an infinite loop


martin-purplefish:
If the agent calls a function before it has synthesized any audio, it gets stuck in an infinite loop.

https://github.com/livekit/agents/blob/1802d3466f8b56e1ec16f1bfc9f75f7960ed62cc/livekit-agents/livekit/agents/pipeline/pipeline_agent.py#L638

This future never resolves.

The easiest way to reproduce is to create an agent that can call functions, and open with a question that causes a function call before the agent's response.

A workaround for this is to make the agent play some empty audio before.

martin-purplefish:
It also seems to get stuck if you return False from the before_llm_cb - I worked around that by just creating a fake and empty LLM stream.

martin-purplefish:
It appears that https://github.com/livekit/agents/blob/1802d3466f8b56e1ec16f1bfc9f75f7960ed62cc/livekit-agents/livekit/agents/pipeline/agent_playout.py#L150 never returns if there are no frames

PR #836, opened 2024-10-04
Make sure there is some queued duration before waiting for audio playout

martin-purplefish:
I believe this fixes https://github.com/livekit/agents/issues/825

The problem seems to be we get occasional empty audio frames which causes freezes.


quanglam2807 (approved):
LGTM!

bora-fish (approved):
⚡

theomonnom (approved):
nice catch! Thanks

PR #848, opened 2024-10-05
Support for Realtime API with Azure OpenAI

davidzhao:


bcherry (approved):
looks good - I didn't test this one. I think we should consider just making the example work automatically either way based on env var detection. I made that change in the JS one while testing it and it felt good to me. lmk what you think

davidzhao:
> looks good - I didn't test this one. I think we should consider just making the example work automatically either way based on env var detection. I made that change in the JS one while testing it and it felt good to me. lmk what you think

sure, we can do that! which env var do you think we should switch on?

bcherry:
> > looks good - I didn't test this one. I think we should consider just making the example work automatically either way based on env var detection. I made that change in the JS one while testing it and it felt good to me. lmk what you think
> 
> sure, we can do that! which env var do you think we should switch on?

I went with endpoint in https://github.com/livekit/agents-js/pull/97/files#diff-8a3cd674863fd6f2a24d4dfc3f6b151e66623fb99bbb0738ca45b75a76ddf906R17

theomonnom (approved):
lgtm

theomonnom:
IMO Azure shouldn't be implicitly used. Users must specify `RealtimeModel.with_azure`

examples/multimodal_agent.py
```diff
@@ -44,12 +44,30 @@ async def get_weather(
...
+    participant = await ctx.wait_for_participant()
 
-    assistant = multimodal.MultimodalAgent(
+    # to use Microsoft Azure, uncomment the following lines
+    # agent = multimodal.MultimodalAgent(
+    #     model=openai.realtime.RealtimeModel.with_azure(
+    #         azure_deployment="<model-deployment>",
+    #         azure_endpoint="wss://<endpoint>.openai.azure.com/", # or AZURE_OPENAI_ENDPOINT
+    #         api_key="<api-key>", # or AZURE_OPENAI_API_KEY
+    #         api_version="2024-10-01-preview", # or OPENAI_API_VERSION
```
davidzhao, line 59: I think this is azure specific

theomonnom: Yeah this is weird but this is what the official OpenAI sdk is using

theomonnom: Keeping the same env vars since this is what is being documented on their docs

Issue #835, opened 2024-10-04
Sometimes VAD Freezes/Unable to detect speech end


hari01584:
We are encountering a very weird bug where sometimes in conversation, VAD/Agent cannot detect end of speech.. the usual flow is that user starts speaking and we can get live transcripts for it but when they stop there is no OpenAI call or subsequent agent response (even after waiting for long time), the solution here is to add on something more (user speaks one or two more words and then again stop), then it detects the whole speech segment and now agent replies back.

Very peculiar but annoying bug, Please fix

martin-purplefish:
We're experiencing this too.

martin-purplefish:
it's strange that I still get the user finished speaking events, but the message never comes.

hari01584:
Indeed, my team is creating automated interviewing solutions and we are rigorously using live kit, there is few bugs popping up but overall this is amazing

theomonnom:
Hey, I just took a look at the VAD implementation and I'm quite sure it isn't coming from it.

So I think something froze inside the VoicePipelineAgent?

Can you share the debug logs of your agent when this is happening? It would be useful to track where the pipeline hangs.

martin-purplefish:
@theomonnom I'll get some debug logs, but here are some info level ones:

```
{"message": "Agent state thinking -> speaking", "level": "INFO", "name": "livekit.agents.pipeline", "pid": 23642, "job_id": "AJ_6Z4b4mTfh6eR", "timestamp": "2024-10-04T21:10:32.386400+00:00"}
{"message": "Agent stopped speaking", "level": "INFO", "name": "livekit.agents.pipeline", "pid": 23642, "job_id": "AJ_6Z4b4mTfh6eR", "timestamp": "2024-10-04T21:10:38.647632+00:00"}
{"message": "logging agent message to backend", "level": "INFO", "name": "livekit.agents.pipeline", "pid": 23642, "job_id": "AJ_6Z4b4mTfh6eR", "timestamp": "2024-10-04T21:10:38.648797+00:00"}
{"message": "Participant attributes: {'lk.agent.state': 'listening'}", "level": "INFO", "name": "livekit.agents.pipeline", "pid": 23642, "job_id": "AJ_6Z4b4mTfh6eR", "timestamp": "2024-10-04T21:10:38.977859+00:00"}
[...]

PR #850, opened 2024-10-05
Fix bug where agent would get stuck on uninterruptable speech

martin-purplefish:


bora-fish (approved):
Brilliant

martin-purplefish:
@theomonnom https://github.com/livekit/agents/pull/667 - related to this, just wanna make sure I'm not breaking anything.

theomonnom (approved):
oh wow, nice catch!!

theomonnom:
Ah let me think about the "stacked replies" scenario

theomonnom:
> @theomonnom #667 - it looks like it reverts, just wanna make sure I'm not breaking anything. Maybe you meant a `break` here or something? This definitely seems to fix my stuck agent problem, but I've also experienced the "stacked replies".

Do you know which speech is not allowing interruptions in your cases? Is it coming from the say method?

martin-purplefish:
> > @theomonnom #667 - it looks like it reverts, just wanna make sure I'm not breaking anything. Maybe you meant a `break` here or something? This definitely seems to fix my stuck agent problem, but I've also experienced the "stacked replies".
> 
> Do you know which speech is not allowing interruptions in your cases? Is it coming from the say method?

I sent Mike some audio files and a transcript example. I think in this case where we run into stacked replies, the agent would just get stuck - the agent would ask a question, the user would reply, but the agent wouldn't hear the reply because it wouldn't get committed. In our case we have a timeout that would kick in and the agent would repeat itself, then the user would say something, and the whole log would get committed: https://github.com/livekit/agents/issues/835

For example, from the user's point of view, they'd experience:

Agent: What's your name?
User: Martin.
Agent: <silence timeout 5 seconds later> What's your name?
User: Martin.
Agent: Nice to meet you, Martin!

But from the agent's point of view, it'd see this, because the first time they spoke would never get validated.

Agent: What's your name?
Agent: <silence timeout 5 seconds later> What's your name?
User: Martin. Martin.
Agent: Nice to meet you, Martin!

martin-purplefish:
> > @theomonnom #667 - it looks like it reverts, just wanna make sure I'm not breaking anything. Maybe you meant a `break` here or something? This definitely seems to fix my stuck agent problem, but I've also experienced the "stacked replies".
> 
> Do you know which speech is not allowing interruptions in your cases? Is it coming from the say method?

It's just from the llm.chat - I only have one .say that is the intro when the convo starts. However, I have the agent with allow_interruptions=False (which as an aside, still seems to allow interruptable speech sometimes)

quanglam2807:
> Do you know which speech is not allowing interruptions in your cases? Is it coming from the say method?

I'm from the same team with @martin-purplefish.

Just to clarify, in our implementation, we set `allow_interruptions=False` when creating the `VoicePipelineAgent`, which prevents interruptions on all speeches. Additionally, we explicitly invoke the `say` method at the beginning of the conversation (with `allow_interruptions=False`)

martin-purplefish:
https://github.com/livekit/agents/pull/857 - incorporates this + others.

PR #857, opened 2024-10-07
Fix agent state

martin-purplefish:
Agent state would update to listening when the user started speaking, even if the agent was still speaking.

This also fixes some bugs where uninterruptable speech would actually be interrupted.


martin-purplefish:
@theomonnom - PTAL

theomonnom (approved):
this lgtm, will try locally before merging

theomonnom (approved):

PR #860, opened 2024-10-07
Add wait_for_participant to mma example

bcherry:
This was already in other VPA examples but missing here. It's not always necessary but improves reliability (e.g. SIP) and seems worth having in the default case.

### DIFF
--- a/examples/multimodal_agent.py
+++ b/examples/multimodal_agent.py
@@ -49,6 +49,10 @@ async def entrypoint(ctx: JobContext):
 
     await ctx.connect(auto_subscribe=AutoSubscribe.AUDIO_ONLY)
 
+    print("waiting for participant")
+    await ctx.wait_for_participant()
+    
+    print("participant joined, starting agent")
     assistant = multimodal.MultimodalAgent(
         model=openai.realtime.RealtimeModel(
             voice="alloy",
--- a/examples/multimodal_agent.py
+++ b/examples/multimodal_agent.py
@@ -51,7 +51,7 @@ async def entrypoint(ctx: JobContext):
 
     print("waiting for participant")
     await ctx.wait_for_participant()
-    
+
     print("participant joined, starting agent")
     assistant = multimodal.MultimodalAgent(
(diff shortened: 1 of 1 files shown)


theomonnom:
Hey, this for VPA it is already handled internally. I think we should do the same for MMA

theomonnom:
It is already the case actually:
https://github.com/livekit/agents/blob/5fedcfdfce510ce96d34d85b38807f0d03389bfc/livekit-agents/livekit/agents/multimodal/multimodal_agent.py#L232

bcherry:
@theomonnom yeah but still all the VPA examples have a wfp in them - is it really unnecessary?

this could be indicative of some sort of bug with how it's set up - it does seem like that should basically work (although it doesn't wait for subscription from the linked_participant but rather from any participant).  and someone in community reported that SIP didn't work til they added this. might be worth digging into, I agree it feels unnecessary and it would be good to simplify it away in all of the examples.

examples/multimodal_agent.py
```diff
@@ -49,6 +49,10 @@ async def get_weather(
 
     await ctx.connect(auto_subscribe=AutoSubscribe.AUDIO_ONLY)
 
+    print("waiting for participant")
+    await ctx.wait_for_participant()
```
davidzhao, line 53:
```suggestion
    participant = await ctx.wait_for_participant()
```

theomonnom: Do you think we have to wait for both, the `wait_for_subscription` + `wait_for_participant`?
How is that handled in the server side. Is it possible for `wait_for_subscription` to returns before `wait_for_participant`

davidzhao: I think we should wait for both.
* wait_for_participant - ensures that we know who we are talking to, so we can set up the agent correctly
* wait_for_subscription - ensures the participant can listen to the agent, so we aren't talking into the void

theomonnom:
Seems like wait_for_participant was already added by dz in some previous PRs

PR #862, opened 2024-10-07
Update doc generation workflow

keepingitneil:


livekit-plugins/livekit-plugins-browser/livekit/plugins/browser/__init__.py
```diff
@@ -27,3 +27,12 @@ def __init__(self):
...
 Plugin.register_plugin(BrowserPlugin())
+
+# Cleanup docs of unexported modules
+module = dir()
+NOT_IN_ALL = [m for m in module if m not in __all__]
+
+__pdoc__ = {}
+
+for n in NOT_IN_ALL:
+    __pdoc__[n] = False
```
theomonnom, line 38: I'm curious if this is really needed.
https://pdoc3.github.io/pdoc/doc/pdoc/#what-objects-are-documented

keepingitneil: It seems like pdoc if a directory has an __init__.py, pdoc is considering it a submodule. Even if it's not in __all__. Technically these modules are importable in a normal, non-package setup, so I guess it makes sense as a default behavior.

.changeset/tame-cows-listen.md
```diff
@@ -0,0 +1,17 @@
...
+"livekit-plugins-browser": patch
+"livekit-plugins-cartesia": patch
+"livekit-plugins-deepgram": patch
+"livekit-plugins-elevenlabs": patch
+"livekit-plugins-google": patch
+"livekit-plugins-nltk": patch
+"livekit-plugins-openai": patch
+"livekit-plugins-playht": patch
+"livekit-plugins-rag": patch
+"livekit-plugins-silero": patch
```
theomonnom, line 14: No need to patch

theomonnom: Going to be noisy otherwise

livekit-plugins/livekit-plugins-nltk/livekit/plugins/nltk/__init__.py
```diff
@@ -38,3 +38,12 @@ def download_files(self):
 
 
 Plugin.register_plugin(NltkPlugin())
+
+# Cleanup docs of unexported modules
+module = dir()
```
theomonnom, line 43:
```suggestion
_module = dir()
```

theomonnom (approved):

PR #865, opened 2024-10-08
Handle llm stream returning false

martin-purplefish:
It would get stuck

### DIFF
--- a/.changeset/yellow-eggs-shout.md
+++ b/.changeset/yellow-eggs-shout.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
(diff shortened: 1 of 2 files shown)


martin-purplefish:
btw check this out too (lined up for next) - I think this fixes most of the interruption/stuck agent problems: https://github.com/martin-purplefish/agents/pull/4/files

theomonnom (approved):

theomonnom:
Awesome, thanks for digging into those issues!

martin-purplefish:
@theomonnom this is fun! btw would love to hop on a call at some point and get to know how I can help more :)

PR #867, opened 2024-10-08
disable preemptive_synthesis by default

theomonnom:


davidzhao (approved):

chasemcdo:
Trying to better understand the implications of this parameter. The docs don't give much detail, but it having been disabled by default seems to imply there are negative consequences to its enablement.

Would it be possible to get elaboration on the reasoning behind this?

Issue #884, opened 2024-10-10
worker unable to receive job when load function return value bigger than 1


Max-Thuc:
I found that when load function return value bigger than 1, despite the threshold is bigger than 1, the worker is unable to receive job. I check the server code and found the issue seem to cause by this PR https://github.com/livekit/livekit/pull/3018. You can look at line https://github.com/livekit/livekit/blob/master/pkg/service/agentservice.go#L465. Where the `normalizedLoad` is calculated by 1 - worker_load. This issue is linked to this issue https://github.com/livekit/agents/issues/834

davidzhao:
Load should be a float between 0-1, it represents percentage load. it doesn't make sense for workers that are more than 100% utilized to receive more jobs.

Max-Thuc:
Hi @davidzhao , thank you for your answer. Should we make it clear in the code by throw some exception if the load threshold is declared bigger than 1

Max-Thuc:
Hi @davidzhao , I have one more thing to ask. Do you have a channel where you announce which code version livekit cloud is using?

davidzhao:
> Should we make it clear in the code by throw some exception if the load threshold is declared bigger than 1

I think printing a warning would make sense.. if a developer computes an incorrect value, it seems like their worker should not crash

dmthuc:
> > Should we make it clear in the code by throw some exception if the load threshold is declared bigger than 1
> 
> I think printing a warning would make sense.. if a developer computes an incorrect value, it seems like their worker should not crash

I see, I've created a PR for this idea here https://github.com/livekit/agents/pull/902/files

PR #885, opened 2024-10-10
fix worker report load

Max-Thuc:
Fix issue https://github.com/livekit/agents/issues/884


davidzhao:
`self._opts.load_threshold` is the % threshold that the worker should stop accepting jobs. it should not be used as a denominator for normalization

PR #891, opened 2024-10-10
Allow forcing interruptions

martin-purplefish:


martin-purplefish:
@theomonnom PTAL

theomonnom (approved):
Ah that makes sense. Can you create another function named "`cancel`" instead? 
This fnc will not check if interruptions is allowed. 

We're currently using the purpose of "cancelling" being the same as "interrupting" which can be confusing

martin-purplefish:
@theomonnom done!

PR #892, opened 2024-10-10
Make VAD optional for voice agents that use streaming STT

martin-purplefish:
For agents using streaming STT, the VAD isn't strictly necessary. In particular this reduces CPU usage quite a bit.


martin-purplefish:
@theomonnom ptal

PR #894, opened 2024-10-11
fix: fix speech_buffer missing data in VADStream

longcw:
Originally `speech_buffer` got only a subset of the audio data for each audio frame from mic bc of
```python
to_copy_buffer = min(self._model.window_size_samples, available_space)
```

This makes the subsequent non-stream STT doesn't work (e.g. openai.STT). Fix it by copying all data to the buffer.

### DIFF
--- a/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
+++ b/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
@@ -273,7 +273,7 @@ class VADStream(agents.vad.VADStream):
 
                 # copy the inference window to the speech buffer
                 available_space = len(speech_buffer) - speech_buffer_index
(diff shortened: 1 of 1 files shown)


zhanghx0905:
I've spent a couple of days hoping to solve this problem without any success. Your EXCELLENT job has helped me a lot, thanks!

longcw:
> I've spent a couple of days hoping to solve this problem without any success. Your EXCELLENT job has helped me a lot, thanks!

Yeah it's a weird behavior in the original code that copy only a part of the audio data to the buffer. Finally it works all good on my end after a few hours of debugging. Post it here also want to see if it's a bug or something intended. Glad it helped you as well :)

theomonnom:
Thanks, I needed some changes so I created another PR
https://github.com/livekit/agents/pull/898

PR #895, opened 2024-10-11
Update README.md

kushal34712:
converted example table from bullets to the table


davidzhao (approved):
this is awesome! thanks for cleaning this up.

PR #897, opened 2024-10-11
OpenAI - Include content in function call results

martin-purplefish:
It seems the other providers also exclude content - however I'm not using them so wasn't able to test.

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -717,7 +717,9 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
                 )
 
             if tool_calls:
-                extra_tools_messages.append(ChatMessage.create_tool_calls(tool_calls))
+                extra_tools_messages.append(
+                    ChatMessage.create_tool_calls(tool_calls, content=collected_text)
+                )
                 extra_tools_messages.extend(tool_calls_results_msg)
 
                 chat_ctx = speech_handle.source.chat_ctx.copy()
--- a/.changeset/six-suits-shake.md
+++ b/.changeset/six-suits-shake.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+Include chat context on collected tool calls
--- a/.changeset/strong-lobsters-push.md
+++ b/.changeset/strong-lobsters-push.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-openai": patch
+---
+
+Fix functions to include context
--- a/.changeset/strong-lobsters-push.md
+++ b/.changeset/strong-lobsters-push.md
@@ -2,4 +2,4 @@
 "livekit-plugins-openai": patch
 ---
 
-Fix functions to include context
+Fix functions to include content
--- a/livekit-agents/livekit/agents/llm/chat_context.py
+++ b/livekit-agents/livekit/agents/llm/chat_context.py
@@ -79,11 +79,10 @@ class ChatMessage:
     @staticmethod
     def create_tool_calls(
         called_functions: list[function_context.FunctionCallInfo],
-        content: ChatContent | list[ChatContent] | None,
+        *,
+        text: str = "",
     ) -> "ChatMessage":
-        return ChatMessage(
-            role="assistant", tool_calls=called_functions, content=content
-        )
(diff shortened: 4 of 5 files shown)


martin-purplefish:
ptal @davidzhao  @theomonnom

davidzhao:
do you know any providers that would do this? it'd be great to understand the behavior on our side.

martin-purplefish:
> do you know any providers that would do this? it'd be great to understand the behavior on our side.

I'm only familiar with OpenAI: https://platform.openai.com/playground/chat?models=gpt-4o&preset=preset-yQpTdwXuiyOaanbcYPoObp97 - here's an example.

The JSON looks like:

```
import OpenAI from "openai";

const openai = new OpenAI({
  apiKey: process.env.OPENAI_API_KEY,
});

const response = await openai.chat.completions.create({
  model: "gpt-4o",
  messages: [
    {
      "role": "system",
      "content": [
        {
          "text": "Create a response that includes an interjection before calling the get_weather function to simulate processing time while the backend queries the API.\n\n# Steps\n\n1. Start with a natural-sounding interjection to acknowledge the user's request or to signal that you are processing the request.\n2. Call the `get_weather` function to retrieve the weather data.\n3. Return the final response with the weather details after the interjection.\n\n# Output Format\n\n- A natural interjection followed by the weather data response.\n- The response should be conversational and seamlessly integrate the interjection.\n\n# Examples\n\n**Input Request**: \"What's the weather like today?\"\n\n**Example Output**: \n- \"Hmm, let me check that for you...The weather today is [sunny with a high of 75°F].\" \n(Note: Replace [sunny with a high of 75°F] with actual weather data retrieved from get_weather.)\n\n# Notes\n\n- Ensure that the interjection is friendly and engaging.\n- The transition from interjection to the actual weather data should be smooth.\n- You can vary the interjection for different interactions if needed to enhance the natural feel.",
          "type": "text"
        }
      ]
    },
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "What's the weather in NYC right now?"
        }
      ]
    },
    {
      "role": "assistant",
      "content": [
        {
          "type": "text",
          "text": "Oh, just a moment while I check that for you... \n\nLet me find out the current weather in NYC by querying our weather service."
        }
      ],
      "tool_calls": [
        {
          "id": "call_Gwqkvbb3r0oRJnj3zzrIEL1R",
          "type": "function",
          "function": {
            "name": "get_weather",
            "arguments": "{\"location\":\"New York, NY\",\"unit\":\"f\"}"
          }
        }
      ]
    },
    {
      "role": "tool",
      "content": [
        {
          "text": "",
          "type": "text"
        }
      ],
      "tool_call_id": "call_Gwqkvbb3r0oRJnj3zzrIEL1R"
    }
  ],
  temperature: 1,
  max_tokens: 2048,
  top_p: 1,
  frequency_penalty: 0,
  presence_penalty: 0,
  tools: [
    {
      "type": "function",
      "function": {
        "name": "get_weather",
        "strict": true,
        "parameters": {
          "type": "object",
          "required": [
            "location",
            "unit"
          ],
          "properties": {
            "unit": {
              "enum": [
                "c",
                "f"
              ],
              "type": "string"
            },
            "location": {
              "type": "string",
              "description": "The city and state e.g. San Francisco, CA"
            }
          },
          "additionalProperties": false
        },
        "description": "Determine weather in my location"
      }
    }
  ],
  parallel_tool_calls: true,
  response_format: {
    "type": "text"
  },
});
```

martin-purplefish:
@davidzhao I noticed a few more places to change for this to work, gimme a few

martin-purplefish:
Alright this is working. Not sure how you'd want to version it - it won't break but it'll require the newer openai plugin to actually work (or content will never be populated)

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -717,7 +717,9 @@ def _commit_user_question_if_needed() -> None:
                 )
 
             if tool_calls:
-                extra_tools_messages.append(ChatMessage.create_tool_calls(tool_calls))
+                extra_tools_messages.append(
+                    ChatMessage.create_tool_calls(tool_calls, content=collected_text)
```
martin-purplefish, line 721: note small functional change here - this'll be "" instead of None if there are no messages in the context (can change it but wasn't sure it mattered)

davidzhao:
nice! I'll give this a try today

theomonnom (approved):
Thanks man!!

PR #898, opened 2024-10-11
silero: fix incomplete speech_buffer on END_OF_SPEECH

theomonnom:
from https://github.com/livekit/agents/pull/894


longcw:
Any reason to use `to_copy_int` instead of the actual audio frame size? 
```python
to_copy = (
                    self._model.window_size_samples * resampling_ratio
                    + input_copy_remaining_fract
                )
```
`to_copy` is a fixed value of the window size of VAD, it may still be smaller than the actual frame size from my observation
```
# to_copy_int, len(frame.data)
1536 1920
1536 1920
1536 1824
1536 1824
1536 1728
1536 1728
1536 1632
```
Will this still lose some audio data from the input stream?

keepingitneil (approved):

longcw:
> Any reason to use `to_copy_int` instead of the actual audio frame size?
> 
> ```python
> to_copy = (
>                     self._model.window_size_samples * resampling_ratio
>                     + input_copy_remaining_fract
>                 )
> ```
> 
> `to_copy` is a fixed value of the window size of VAD, it may still be smaller than the actual frame size from my observation
> 
> ```
> # to_copy_int, len(frame.data)
> 1536 1920
> 1536 1920
> 1536 1824
> 1536 1824
> 1536 1728
> 1536 1728
> 1536 1632
> ```
> 
> Will this still lose some audio data from the input stream? My understanding is that `INFERENCE_DONE` only outputs the audio of the window processed by VAD, but `END_OF_SPEECH` should output the full audio from speech start to end that detected by VAD, instead of a combination of the windows used for detection. (Or actually the `frame.data` is redundant with padding which I don't know?)

But I tested the `to_copy_int` and `len(frame.data)`, the first one has a better audio quality so I think you are correct 👍 , guess there is a padding which I don't know.

theomonnom:
> But I tested the `to_copy_int` and `len(frame.data)`, the first one has a better audio quality so I think you are correct 👍 , guess there is a padding which I don't know.


We're re-adding the remaining data at the end of the inference here. https://github.com/livekit/agents/blob/3125a3457b5662f86775baf5ccc236075559af9b/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py#L411

longcw:
> > But I tested the `to_copy_int` and `len(frame.data)`, the first one has a better audio quality so I think you are correct 👍 , guess there is a padding which I don't know.
> 
> We're re-adding the remaining data at the end of the inference here.
> 
> https://github.com/livekit/agents/blob/3125a3457b5662f86775baf5ccc236075559af9b/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py#L411

Okay I see, so the remaining data will be processed in the next iteration.

PR #902, opened 2024-10-12
warning incorrect threshold config

dmthuc:
Give a warning when config for load threshold is invalid.
```
{"message": "load_threshold in prod env must be less than 1, current value: 2", "level": "WARNING", "name": "livekit.agents", "timestamp": "2024-10-12T08:20:02.168613+00:00"}
```
This pr is for issue https://github.com/livekit/agents/issues/884

### DIFF
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -234,6 +234,7 @@ def _run_dev(
 
 def run_worker(args: proto.CliArgs) -> None:
     setup_logging(args.log_level, args.devmode)
+    args.validate_config()
 
     loop = asyncio.get_event_loop()
     worker = Worker(args.opts, devmode=args.devmode, loop=loop)
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -234,7 +234,7 @@ def _run_dev(
 
 def run_worker(args: proto.CliArgs) -> None:
     setup_logging(args.log_level, args.devmode)
-    args.validate_config()
+    args.opts.validate_config(args.devmode)
 
     loop = asyncio.get_event_loop()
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/cli/cli.py
```diff
@@ -234,6 +234,7 @@ async def _run_loop():
 
 def run_worker(args: proto.CliArgs) -> None:
     setup_logging(args.log_level, args.devmode)
+    args.validate_config()
```
davidzhao, line 237: it'd be better to make the call directly vs another abstraction

```suggestion
    args.opts.validate_config()
```

livekit-agents/livekit/agents/cli/proto.py
```diff
@@ -30,6 +30,9 @@ class CliArgs:
     # when reload/dev mode is enabled
     mp_cch: socket.socket | None = None
 
+    def validate_config(self):
+        self.opts.validate_config(self.devmode)
+
```
davidzhao, line 35: this wouldn't be needed

livekit-agents/livekit/agents/worker.py
```diff
@@ -185,6 +185,14 @@ class WorkerOptions:
     The HTTP server is used as a health check endpoint.
     """
 
+    def validate_config(self, devmode: bool):
+        if not devmode:
+            load_threshold = _WorkerEnvOption.getvalue(self.load_threshold, False)
```
davidzhao, line 190:
```suggestion
        load_threshold = _WorkerEnvOption.getvalue(self.load_threshold, devmode)
```

livekit-agents/livekit/agents/cli/cli.py
dmthuc: I've updated according to your advices

davidzhao (approved):

PR #906, opened 2024-10-12
Fix bug with before_tts_cb when speech is string but before_tts_cb returns async iterable

martin-purplefish:
If you said agent.say(), but your before_tts_cb returned an AsyncIterable, it'd throw - because the transcript was a string but the before_tts_cb was AsyncIterable.

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/agent_output.py
+++ b/livekit-agents/livekit/agents/pipeline/agent_output.py
@@ -276,7 +276,6 @@ async def _stream_synthesis_task(
 
     tts_stream = handle._tts.stream()
     read_tts_atask: asyncio.Task | None = None
-
     read_transcript_atask: asyncio.Task | None = None
 
     try:
--- a/livekit-agents/livekit/agents/pipeline/agent_output.py
+++ b/livekit-agents/livekit/agents/pipeline/agent_output.py
@@ -189,6 +189,7 @@ class AgentOutput:
             await utils.aio.gracefully_cancel(synth)
 
 
+@utils.log_exceptions(logger=logger)
 async def _str_synthesis_task(
     tts_text: str, transcript: str, handle: SynthesisHandle
 ) -> None:
--- a/livekit-agents/livekit/agents/pipeline/agent_output.py
+++ b/livekit-agents/livekit/agents/pipeline/agent_output.py
@@ -172,14 +172,6 @@ class AgentOutput:
             tts_source = await tts_source
         if isinstance(transcript_source, Awaitable):
             transcript_source = await transcript_source
-        if not self._tts.capabilities.streaming:
-            logger.debug(
-                "forcing tts source to be a string since TTS does not streaming"
-            )
-            if isinstance(tts_source, AsyncIterable):
-                tts_source = "".join([seg async for seg in tts_source])
-            if isinstance(transcript_source, AsyncIterable):
-                transcript_source = "".join([seg async for seg in transcript_source])
 
         if isinstance(tts_source, str) and isinstance(transcript_source, str):
             co = _str_synthesis_task(tts_source, transcript_source, handle)
--- a/livekit-agents/livekit/agents/pipeline/agent_output.py
+++ b/livekit-agents/livekit/agents/pipeline/agent_output.py
@@ -12,7 +12,7 @@ from .. import tts as text_to_speech
 from .agent_playout import AgentPlayout, PlayoutHandle
 from .log import logger
 
-AgentSpeech = AsyncIterable[str] | str
+AgentSpeech = Union[AsyncIterable[str], str]
 SpeechSource = Union[Awaitable[AgentSpeech], AgentSpeech]
 
 
--- a/livekit-agents/livekit/agents/pipeline/agent_output.py
+++ b/livekit-agents/livekit/agents/pipeline/agent_output.py
@@ -172,9 +172,18 @@ class AgentOutput:
             tts_source = await tts_source
         if isinstance(transcript_source, Awaitable):
             transcript_source = await transcript_source
+        if not self._tts.capabilities.streaming:
+            logger.debug("forcing tts source to be a string since TTS is not streaming")
+            if isinstance(tts_source, AsyncIterable):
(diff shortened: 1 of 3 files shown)


martin-purplefish:
PTAL @davidzhao @theomonnom - I couldn't figure out how to write a nice before_llm_cb  without this.

livekit-agents/livekit/agents/pipeline/agent_output.py
```diff
@@ -170,8 +171,11 @@ async def _synthesize_task(self, handle: SynthesisHandle) -> None:
 
         if isinstance(tts_source, Awaitable):
             tts_source = await tts_source
-            co = _str_synthesis_task(tts_source, transcript_source, handle)
-        elif isinstance(tts_source, str):
+        if isinstance(transcript_source, Awaitable):
+            transcript_source = await transcript_source
+
+        if isinstance(tts_source, str):
+            assert isinstance(transcript_source, str)
```
martin-purplefish, line 178: I don't think this would ever happen (tts_source is str, transcript is stream) but could implement it if we wanted.

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -784,7 +784,7 @@ def _synthesize_agent_speech(
 
         tts_source = self._opts.before_tts_cb(self, og_source)
         if tts_source is None:
-            logger.error("before_tts_cb must return str or AsyncIterable[str]")
+            raise ValueError("before_tts_cb must return str or AsyncIterable[str]")
```
martin-purplefish, line 787: This is probably a bug so i figured it should throw

livekit-agents/livekit/agents/pipeline/agent_output.py
```diff
@@ -186,49 +185,21 @@ async def _synthesize_task(self, handle: SynthesisHandle) -> None:
             await utils.aio.gracefully_cancel(synth)
 
 
-@utils.log_exceptions(logger=logger)
```
martin-purplefish, line 189: I figured it's cleaner to just have one method, and convert the others into async if they're not.

martin-purplefish:
Alright cleaned it up a bunch. Ready for review

quanglam2807 (approved):

martin-purplefish:
@davidzhao @theomonnom

theomonnom:
Hey, I think we still want to use the "synthesize" method instead of the "stream" when the input is a stream?
wdyt?

I'm just not sure about this part:
```
async def _str_to_aiter(s: str) -> AsyncIterable[str]:
    yield s
```

martin-purplefish:
> Hey, I think we still want to use the "synthesize" method instead of the "stream" when the input is a string? wdyt?
> 
> I'm just not sure about this part:
> 
> ```
> async def _str_to_aiter(s: str) -> AsyncIterable[str]:
>     yield s
> ```

That was so we could treat the string as a stream that yields just that token, and just have the `stream_synthesis_task` instead of also `str_synthesis_task`. Otherwise we have to deal with 4 different cases of tts_source and transcript_source being str or iterable (all of the permutations are possible depending on how the user calls the LLM and what types before_tts_cb returns)

theomonnom:
By the synthesize method, I mean the [TTS.synthesize](https://github.com/livekit/agents/blob/cc2fe7782d1cc4156a87dd2fe33b584ec8b6c6e7/livekit-agents/livekit/agents/tts/tts.py#L51)

martin-purplefish:
> By the synthesize method, I mean the [TTS.synthesize](https://github.com/livekit/agents/blob/cc2fe7782d1cc4156a87dd2fe33b584ec8b6c6e7/livekit-agents/livekit/agents/tts/tts.py#L51)

Oh I see, there's another wrinkle :D

How about this - we stream if it supports it, otherwise we throw if both the tts_source and the transcript_source aren't strings?

martin-purplefish:
What do you think of this @theomonnom ?

theomonnom:
Hey, I think this makes sense, what do you think about always following what is the type of tts_source.
And even if the transcript_source is an AsyncIterable, we can still create a task that listen to it inside the str_synthesis_task

martin-purplefish:
Got it, yeah I can do that.

martin-purplefish:
@theomonnom ptal

livekit-agents/livekit/agents/pipeline/agent_output.py
```diff
@@ -187,16 +189,28 @@ async def _synthesize_task(self, handle: SynthesisHandle) -> None:
...
-async def _str_synthesis_task(
-    tts_text: str, transcript: str, handle: SynthesisHandle
+async def _read_transcript_task(
+    transcript_source: AsyncIterable[str] | str, handle: SynthesisHandle
 ) -> None:
-    """synthesize speech from a string"""
+    if isinstance(transcript_source, str):
+        handle._tr_fwd.push_text(transcript_source)
+    else:
+        async for seg in transcript_source:
```
theomonnom, line 198: last nit: we should do that inside its own task otherwise it'll wait to get the whole transcript before synthesizing speech. This may add latency

theomonnom (approved):
lgtm thanks a lot!

livekit-agents/livekit/agents/pipeline/agent_output.py
martin-purplefish: Isn't it in its own task already? Sorry if I'm misunderstanding, I'm not super familiar with asyncio

theomonnom: It is, but I guess you need to create another task. Otherwise the `tts.synthesize` method will never be called until transcript_source is done.

martin-purplefish: (sorry for the double ping - sent a message on slack too). _read_transcript_task is called with

```
  read_transcript_atask = asyncio.create_task(
    _read_transcript_task(transcript_source, handle)
)
```
inside the tts.synthesize loop on the first frame. What would the second task do here?

theomonnom (approved):

Issue #911, opened 2024-10-14
Unable to use SSML with azure


samirsalman:
Right now is not possible to use the SSML (https://learn.microsoft.com/en-us/azure/ai-services/speech-service/speech-synthesis-markup) with the Azure TTS component.
I will open a PR to add this feature.

PR #912, opened 2024-10-14
Azure TTS Prosody SSML support

samirsalman:
In this PR, I have added support for SSML prosody tags in Azure TTS. This will allow users to control the speech rate, pitch, and volume of the generated speech. The prosody tag is a part of the SSML specification and is supported by Azure TTS.

To use the prosody tags in Azure TTS component the user needs to provide a `ProsodyConfig` object to the `TTS` component. 
The `ProsodyConfig` is validated in the `TTS` component.

```python
@dataclass
class ProsodyConfig:
    """
    Prosody configuration for Azure TTS.

    Args:
        rate: Speaking rate. Can be one of "x-slow", "slow", "medium", "fast", "x-fast", or a float. A float value of 1.0 represents normal speed.
        volume: Speaking volume. Can be one of "silent", "x-soft", "soft", "medium", "loud", "x-loud", or a float. A float value of 100 (x-loud) represents the highest volume and it's the default pitch.
        pitch: Speaking pitch. Can be one of "x-low", "low", "medium", "high", "x-high".
    """

    rate: Literal["x-slow", "slow", "medium", "fast", "x-fast"] | float | None = "medium"
    volume: (
        Literal["silent", "x-soft", "soft", "medium", "loud", "x-loud"] | float | None
    ) = "x-loud"
    pitch: Literal["x-low", "low", "medium", "high", "x-high"] | None = "medium"

```

When the user specify the `ProsodyConfig` object, the `TTS` component will generate the SSML with the prosody tags and use the `speak_ssml_async` method to generate the speech. Otherwise, the `TTS` component will use the default `speak_text_async` method to generate the speech.

No changes are required in the `TTS` component to support the prosody tags. The `TTS` component will automatically generate the SSML with the prosody tags if the `ProsodyConfig` object is provided.

I also added the language parameter to the TTS initialization to explicitly specify the language of the text to be synthesized, useful when the model is multilingual.

https://github.com/livekit/agents/issues/911

### DIFF
--- a/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
+++ b/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
@@ -15,14 +15,52 @@ from __future__ import annotations
 import asyncio
 import os
 from dataclasses import dataclass
-
-from livekit.agents import tts, utils
+from typing import Literal
 
 import azure.cognitiveservices.speech as speechsdk  # type: ignore
+from livekit.agents import tts, utils
 
 AZURE_SAMPLE_RATE: int = 16000
 AZURE_BITS_PER_SAMPLE: int = 16
 AZURE_NUM_CHANNELS: int = 1
+
+@dataclass
+class ProsodyConfig:
+    """
+    Prosody configuration for Azure TTS.
+
+    Args:
+        rate: Speaking rate. Can be one of "x-slow", "slow", "medium", "fast", "x-fast", or a float. A float value of 1.0 represents normal speed.
+        volume: Speaking volume. Can be one of "silent", "x-soft", "soft", "medium", "loud", "x-loud", or a float. A float value of 100 (x-loud) represents the highest volume and it's the default pitch.
+        pitch: Speaking pitch. Can be one of "x-low", "low", "medium", "high", "x-high".
+    """
+
+    rate: Literal["x-slow", "slow", "medium", "fast", "x-fast"] | float | None = "medium"
+    volume: (
+        Literal["silent", "x-soft", "soft", "medium", "loud", "x-loud"] | float | None
+    ) = "x-loud"
+    pitch: Literal["x-low", "low", "medium", "high", "x-high"] | None = "medium"
+
+    def validate(self) -> None:
+        if isinstance(self.rate, float) and not 0.5 <= self.rate <= 2:
+            raise ValueError("Prosody rate must be between 0.5 and 2")
+        if isinstance(self.volume, float) and not 0 <= self.volume <= 100:
+            raise ValueError("Prosody volume must be between 0 and 100")
+        if self.pitch not in ["x-low", "low", "medium", "high", "x-high"]:
+            raise ValueError(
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
```diff
@@ -15,16 +15,55 @@
...
+    """
+    Prosody configuration for Azure TTS.
+
+    Args:
+        rate: Speaking rate. Can be one of "x-slow", "slow", "medium", "fast", "x-fast", or a float. A float value of 1.0 represents normal speed.
+        volume: Speaking volume. Can be one of "silent", "x-soft", "soft", "medium", "loud", "x-loud", or a float. A float value of 100 (x-loud) represents the highest volume and it's the default pitch.
+        pitch: Speaking pitch. Can be one of "x-low", "low", "medium", "high", "x-high".
+    """
+
+    rate: Literal["x-slow", "slow", "medium", "fast", "x-fast"] | float | None = "medium"
```
theomonnom, line 39: Should we leave the defaults to None?

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
```diff
@@ -43,6 +86,8 @@ def __init__(
         speech_region: str | None = None,
         voice: str | None = None,
         endpoint_id: str | None = None,
+        language: str | None = None,
+        prosody: ProsodyConfig | None = None,
```
theomonnom, line 90: should we rename this field to "default_prosody"?
I'm just considering the case where users add their own SSML inside the text.

theomonnom (approved):
Hey, thanks for the PR! Just a few nits

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
theomonnom: Maybe it isn't a big deal since it already defaults to None. So no users will be impacted if they're already using their own SSML

samirsalman: @theomonnom also right now it's not possible to use the SSML at all

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
samirsalman: yep

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
samirsalman: no way to provide SSML using the text, because the Azure TTS calls `synthesizer.speak_text_async(self._text).get()`

theomonnom: Ah I see!

theomonnom (approved):
awesome, thanks!!

PR #915, opened 2024-10-15
Fix OpenAI Assistants API example

### LINKED ISSUE #396: Add Support for OpenAI Assistants API to OpenAI plugin
**Describe the bug**
The assistants completion API is slightly different from the base model completions api so the `base_url` cannot be simply swapped to the endpoint of an Assistant. 

**To Reproduce**
1. Swap OpenAI plugin `base_url` to an Assistant `base url` formatted as `https://api.openai.com/v1/assistants/{assistant_id}/messages`
2. Attempt to use the assistant for completions
3. Receive stack trace around argument mismatch in `LLM.chat()`

**Expected behavior**
Assistant completions work as expected (similarly to base model completions)

davidzhao:
removed example from readme and fixed usage

closes #396


theomonnom (approved):
Thanks!

PR #918, opened 2024-10-15
Fixed Prosody Config Validation

samirsalman:
## Summary 
This pull request addresses a validation issue within the ProsodyConfig class for Azure Text-to-Speech (TTS) in the livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py file.

## Changes

- Corrected the default value for volume from "x-loud" to None.
- Fixed the validation logic for rate by adding type checks and ensuring the value is one of the specified literals or within the valid float range.
- Updated the validation for volume to include type checks.
- Ensured volume is validated against the correct list of literals if it's a string type.

### DIFF
--- a/.changeset/lucky-kangaroos-fly.md
+++ b/.changeset/lucky-kangaroos-fly.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-azure": patch
+---
+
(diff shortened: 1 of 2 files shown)


theomonnom (approved):
ah, makes sense, thanks!

PR #924, opened 2024-10-16
Publish llama-index plugin

davidzhao:


livekit-plugins/livekit-plugins-llama-index/README.md
```diff
@@ -1,14 +1,36 @@
...
+Query Engine is primarily used for RAG. See [example voice agent](https://github.com/livekit/agents/blob/main/examples/voice-pipeline-agent/llamaindex-rag/query_engine.py)
+
+## Chat Engine
 
-```json
-{
-  "name": "livekit-plugins-<name>",
-  "private": true
-}
+Chat Engine can be used as an LLM within the frame work.
```
theomonnom, line 17: ChatEngine isn't the recommended way tho. Since this one doesn't support function calls (It is using the LLM from lama_index)

theomonnom: This one: https://github.com/livekit/agents/blob/main/examples/voice-pipeline-agent/llamaindex-rag/query_engine.py
and this one https://github.com/livekit/agents/blob/main/examples/voice-pipeline-agent/llamaindex-rag/retrieval.py
are more appropriate, but it depends on the needs.

The ChatEngine is also doing some "context" management (like keeping the previous rag search of the prior question etc.). They all have pros and cons

theomonnom (approved):
Thanks man (ignore CI fails, it's my fault)

livekit-plugins/livekit-plugins-llama-index/README.md
```diff
@@ -1,14 +1,36 @@
...
+
+```python
+# load the existing index
+storage_context = StorageContext.from_defaults(persist_dir=<mydir>)
+index = load_index_from_storage(storage_context)
+chat_engine = index.as_chat_engine(chat_mode=ChatMode.CONTEXT)
+
+async def entrypoint(ctx: JobContext):
+    ...
```
theomonnom, line 26:
```suggestion

async def entrypoint(ctx: JobContext):

     chat_engine = index.as_chat_engine(chat_mode=ChatMode.CONTEXT)
    ...
```

livekit-plugins/livekit-plugins-llama-index/README.md
davidzhao: I see, you have an example for it, I'm just documenting it.

theomonnom (approved):

Issue #807, opened 2024-09-28
OpenAI timeouts are too low / aren't handled well


jezell:
The LLM client is configured with a 5 second read timeout. If the client times out (which it does very often with a short timeout), the stream is not resumed. 

The error handling needs to be more graceful, but the timeout really shouldn't be 5 seconds by default.

davidzhao:
that's a good point.. we set it on the lower side mostly because it would seem awkward in the flow of the conversation to be stuck for awhile. 

but I understand your point, by default it probably should be higher.

PR #926, opened 2024-10-16
openai: fix low timeouts 

theomonnom:
fix ridiculously low timeouts
future work would be to expose those in every client we have

fix https://github.com/livekit/agents/issues/807

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -75,11 +75,11 @@ class LLM(llm.LLM):
             api_key=api_key,
             base_url=base_url,
             http_client=httpx.AsyncClient(
-                timeout=httpx.Timeout(timeout=30, connect=10, read=5, pool=5),
+                timeout=httpx.Timeout(connect=15.0, read=5.0, write=5.0, pool=5.0),
                 follow_redirects=True,
                 limits=httpx.Limits(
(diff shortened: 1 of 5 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -75,11 +75,11 @@ def __init__(
             api_key=api_key,
             base_url=base_url,
             http_client=httpx.AsyncClient(
-                timeout=httpx.Timeout(timeout=30, connect=10, read=5, pool=5),
+                timeout=httpx.Timeout(connect=15.0, read=5.0, write=5.0, pool=5.0),
```
davidzhao, line 78: the change is from 10s -> 15s for connect? what does removing timeout mean? should we also increase the other timeouts?

davidzhao (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
theomonnom: timeout where setting everything to 30.0

theomonnom: llm was OK, it's more for TTS/STT that the changes is going from 5 to 15

davidzhao: got it, sounds good

PR #932, opened 2024-10-16
Add option for a min delay between playouts

martin-purplefish:
My agent occasionally wants to play audio immediately after it's done playing - this is to allow for a minimum gap between the two to avoid awkward responses. I've found a value around 0.7 is good, but I wanted to leave it at 0 to avoid a behavior change.


martin-purplefish:
PTAL @davidzhao @theomonnom

PR #929, opened 2024-10-16
 Fix SSML Implementation by Adding <voice> Tag

samirsalman:
This pull request fixes the SSML implementation in the Azure Text-to-Speech (TTS) integration by correctly adding the <voice> tag. This ensures that the SSML is properly structured and the specified voice is used during synthesis.

Hope to get soon a document with info about contribution and code test guidelines  🙏

PR #943, opened 2024-10-17
Add voice tag on ssml and add default voice

harmlessman:
Add voice tag on ssml and add default voice.

After code modification, I saw the pull request below....
https://github.com/livekit/agents/pull/929/files

However, there is no default voice, so I upload a pull request.


harmlessman:
I missed branch..

PR #945, opened 2024-10-17
Additional options enabled on Google TTS

hari-truviz:
- Added arguments to Google TTS class to accept and set Pitch and Audio Effect Profiles.


theomonnom (approved):
Thanks!

PR #946, opened 2024-10-17
remove macos-14-large 

theomonnom:
it is failing and it is the same as macos-14
idk why we added it


nbsp:
macos-large should be aarch as opposed to x86_64, no?

theomonnom:
> macos-large should be aarch as opposed to x86_64, no?

oh is it? I'm not sure, if this is the case then we should keep it

theomonnom:
yep, seems like this one is x64

PR #956, opened 2024-10-18
Converted self._speech_q into an asyncio.Queue to safely manage speech tasks 

jayeshp19:
We're manually managing the `self._speech_q` list with `asyncio.Event()`
- **PR :** use `asyncio.Queue` since it's built for handling of asynchronous producer-consumer patterns.

CC: @davidzhao  @theomonnom please review it and let me know there's any change

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -246,9 +246,7 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
             loop=self._loop,
         )
 
-        self._speech_q: list[SpeechHandle] = []
-        self._speech_q_changed = asyncio.Event()
-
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -505,16 +507,11 @@ def _on_playout_stopped(interrupted: bool) -> None:
...
-                await self._play_speech(speech)
-                self._speech_q.pop(0)  # Remove the element only after playing
-                self._playing_speech = None
-
-            self._speech_q_changed.clear()
+            speech_handle = await self._speech_q.get()
+            self._playing_speech = speech_handle
+            await self._play_speech(speech_handle)
+            self._speech_q.task_done()
+            self._playing_speech = None
```
theomonnom, line 514: I think the big difference here is that the speech is now removed from the queue before the playout is done

theomonnom:
I don't think we need that as the behaviour isn't the same

PR #969, opened 2024-10-22
Fix race in default load calc.

martin-purplefish:
the thread could start and then lock would be undefined, and you'd get the occasional

`AttributeError: '_DefaultLoadCalc' object has no attribute '_lock'`


martin-purplefish:
PTAL @davidzhao @theomonnom

theomonnom (approved):
thanks!

PR #970, opened 2024-10-22
voicepipeline: support recursive/chained function calls

### LINKED ISSUE #967: VoicePipelineAgent pauses indefinitely when chaining tool calls
## Overview

The `VoicePipelineAgent` can get stuck in a state where when trying to chain tool calls it pauses and waits for further input from the user, despite the LLMStream returning a chunk to call the function. I know the LLMStream returns it because I've put logging  [here](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py#L515) and have seen the second tool call coming through.

I think this may be related to the LLM going to call one tool and discovering it actually needs to call a different tool first.

I was having trouble tracing exactly why this additional tool call doesn't come through to the agent.

## Example

We can recreate this condition consistently with the following example. It just modifies the existing weather example, to split into two tools, one that gets the ID for the location and the other that gets the weather for the location ID. It ends up only executing the function to get the ID, and then pauses without executing the other function or saying anything.

```python
import logging
from typing import Annotated

import aiohttp
from livekit.agents import (
    AutoSubscribe,
    JobContext,
    JobProcess,
    WorkerOptions,
    cli,
    llm,
)
from livekit.agents.pipeline import VoicePipelineAgent
from livekit.plugins import deepgram, openai, silero

load_dotenv()

logger = logging.getLogger("weather-demo")
logger.setLevel(logging.INFO)


class AssistantFnc(llm.FunctionContext):
    """
    The class defines a set of LLM functions that the assistant can execute.
    """

    @llm.ai_callable()
    async def get_location_id(
        self,
        location: Annotated[
            str, llm.TypeInfo(description="The location to get the id for")
        ],
    ):
        """Called when the user asks about the weather. This function will return the id for the location that can be used with `get_location_id`"""
        logger.info(f"getting id for {location}")
        return "loc_123"

    @llm.ai_callable()
    async def get_weather(
        self,
        location_id: Annotated[
            str,
            llm.TypeInfo(
                description="The id of the location to get the weather for. You must use the `get_location_id` tool to get the ID for this argument."
            ),
        ],
    ):
        """Called when the user asks about the weather. This function will return the weather for the given location_id."""
        logger.info(f"getting weather for {location_id}")
        return "It's sunny and 75 degrees"


def prewarm_process(proc: JobProcess):
    # preload silero VAD in memory to speed up session start


    fnc_ctx = AssistantFnc()  # create our fnc ctx instance
    initial_chat_ctx = llm.ChatContext().append(
        text=(
            "You are a weather assistant created by LiveKit. Your interface with users will be voice. "
            "You will provide weather information for a given location."
        ),
        role="system",
    )
    agent = VoicePipelineAgent(
        vad=ctx.proc.userdata["vad"],
        stt=deepgram.STT(),
        llm=openai.LLM(),
        tts=openai.TTS(),
        fnc_ctx=fnc_ctx,
        chat_ctx=initial_chat_ctx,
    )
    # Start the assistant. This will automatically publish a microphone track and listen to the participant.
    agent.start(ctx.room, participant)
    await agent.say(
        "Hello from the weather station. Would you like to know the weather? If so, tell me your location."
    )


if __name__ == "__main__":
    cli.run_app(
        WorkerOptions(
            entrypoint_fnc=entrypoint,
            prewarm_fnc=prewarm_process,
        ),
    )

```

### Other Info

Running latest agent framework `0.10.2` on Python 3.12.

theomonnom:
fix #967

### DIFF
--- a/.changeset/twenty-poems-whisper.md
+++ b/.changeset/twenty-poems-whisper.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+voicepipeline: support recursive/chained function calls
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -99,6 +99,7 @@ class _ImplOptions:
     int_speech_duration: float
     int_min_words: int
     min_endpointing_delay: float
+    max_recursive_fnc_calls: int
     preemptive_synthesis: bool
     before_llm_cb: BeforeLLMCallback
     before_tts_cb: BeforeTTSCallback
@@ -149,6 +150,7 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
         interrupt_speech_duration: float = 0.5,
         interrupt_min_words: int = 0,
         min_endpointing_delay: float = 0.5,
+        max_recursive_fnc_calls: int = 1,
         preemptive_synthesis: bool = False,
         transcription: AgentTranscriptionOptions = AgentTranscriptionOptions(),
         before_llm_cb: BeforeLLMCallback = _default_before_llm_cb,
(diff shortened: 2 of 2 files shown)


davidzhao (approved):
nice fix!

PR #972, opened 2024-10-22
fix docker cpu load calc

### LINKED ISSUE #966: Incorrect CPU usage detection on shared hosts
The worker currently monitors CPU usage and shuts down if the usage exceeds a certain threshold. It determines CPU usage using the function `psutil.cpu_percent(0.5)`, as shown here: [livekit worker code](https://github.com/livekit/agents/blob/4e92fbf427b1f9c5bca82a832ce48c8d1e41bb16/livekit-agents/livekit/agents/worker.py#L80). This method works well on dedicated hosts, but on shared hosts (e.g., Render.com), it returns the CPU usage of the entire base system rather than the specific instance container. If the base system is shared among multiple containers and is under heavy load, the worker may incorrectly interpret this as high CPU usage for its instance, leading to unnecessary shutdowns.

martin-purplefish:
Fixes #966. In collaboration with @quang-purplefish

Let me know if you want this to be more configurable - we verified this on render.com

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -69,10 +69,23 @@ def _is_docker() -> bool:
     return os.path.exists("/sys/fs/cgroup/cpu.max")
 
 
+def _read_cpu_max() -> tuple[str, int]:
+    with open("/sys/fs/cgroup/cpu.max", "r") as f:
+        data = f.read().strip().split()
+        quota = data[0]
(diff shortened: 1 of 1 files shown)


martin-purplefish:
PTAL @davidzhao @theomonnom

davidzhao:
this is awesome! thanks Martin! I'm going to refactor this slightly and will merge after!

davidzhao (approved):

PR #984, opened 2024-10-24
expose usage metrics

theomonnom:
followup for https://github.com/livekit/agents/pull/957


livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -18,6 +18,7 @@ class TTSMetrics(TypedDict):
     duration: float
     audio_duration: float
     cancelled: bool
+    characters: int
```
theomonnom, line 21: characters or num_characters? no strong opinion

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -21,6 +21,10 @@ class LLMMetrics(TypedDict):
     duration: float
     label: str
     cancelled: bool
+    completion_tokens: int
```
keepingitneil, line 24: not sure if already discussed but time to first token also interesting

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -853,6 +853,9 @@ async def _llm_stream_to_str_generator(
         ) -> AsyncGenerator[str]:
             try:
                 async for chunk in stream:
+                    if not chunk.choices:
```
keepingitneil, line 856: was this happening?

livekit-agents/livekit/agents/tts/tts.py
keepingitneil: character_cnt?

keepingitneil (approved):

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
theomonnom: It can if we receive a chunk that is only for CompletionUsage

livekit-agents/livekit/agents/llm/llm.py
theomonnom: it is already here, it is called ttft

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -76,9 +88,12 @@ async def recognize(
         event = await self._recognize_impl(buffer, language=language)
         duration = time.perf_counter() - start_time
         stt_metrics: STTMetrics = {
+            "request_id": event.request_id,
             "timestamp": time.time(),
             "duration": duration,
             "label": self._label,
+            "audio_duration": calculate_audio_duration(buffer),
```
davidzhao, line 95: any differences between this and what's contained in `ReconitionUsage`?

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -18,6 +18,7 @@ class TTSMetrics(TypedDict):
     duration: float
     audio_duration: float
     cancelled: bool
+    num_characters: int
```
davidzhao, line 21: nit, following the same `<resource>_<unit>` convention

`character_count: int`

davidzhao (commented):
lg! just a few nits

davidzhao:
another question. OpenAI supports token usage reports. wouldn't it make sense to use provider-provided usage when it exists? it seems like it would be more accurate

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -28,6 +32,8 @@ class SpeechEventType(str, Enum):
     FINAL_TRANSCRIPT = "final_transcript"
     """final transcript, emitted when the STT is confident enough that a certain
     portion of speech will not change"""
+    RECOGNITION_USAGE = "recognition_usage"
```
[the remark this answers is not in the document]
theomonnom: INFERENCE_USAGE is used to forward the metrics to the base class. (Similar to what the CompletionUsage dataclass is doing for the LLM). 
We need a "general" way, to automatically emit metrics based on what the subclasses give us. For now it is just audio_duration

livekit-agents/livekit/agents/stt/stt.py
theomonnom: Yeah the only difference is that RecognitionUsage is used by the subclasses, the metrics emitted doesn't use this class

theomonnom:
> another question. OpenAI supports token usage reports. wouldn't it make sense to use provider-provided usage when it exists? it seems like it would be more accurate

I'm already using it

davidzhao:
> > another question. OpenAI supports token usage reports. wouldn't it make sense to use provider-provided usage when it exists? it seems like it would be more accurate
> 
> I'm already using it

oops my bad. missed it. ignore :)

PR #985, opened 2024-10-24
support language code in ElevenLabs TTS

cch41:
While only supported for turbo v2.5, specifying language can be a helpful constraint. We currently overwrite your eleven labs tts to do this, upstreaming it here.

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -87,7 +87,7 @@ class _TTSOptions:
     api_key: str
     voice: Voice
     model: TTSModels | str
-    language_code: str | None
+    language: str | None
     base_url: str
     encoding: TTSEncoding
     sample_rate: int
@@ -115,7 +115,7 @@ class TTS(tts.TTS):
         http_session: aiohttp.ClientSession | None = None,
         # deprecated
         model_id: TTSModels | str | None = None,
-        language_code: str | None = None,
+        language: str | None = None,
     ) -> None:
         """
         Create a new instance of ElevenLabs TTS.
@@ -131,7 +131,7 @@ class TTS(tts.TTS):
             enable_ssml_parsing (bool): Enable SSML parsing for input text. Defaults to False.
             chunk_length_schedule (list[int]): Schedule for chunk lengths, ranging from 50 to 500. Defaults to [80, 120, 200, 260].
             http_session (aiohttp.ClientSession | None): Custom HTTP session for API requests. Optional.
-            language_code (str | None): Language code for the TTS model, as of 10/24/24 only valid for "eleven_turbo_v2_5". Optional.
+            language (str | None): Language code for the TTS model, as of 10/24/24 only valid for "eleven_turbo_v2_5". Optional.
         """
 
         super().__init__(
@@ -165,7 +165,7 @@ class TTS(tts.TTS):
             word_tokenizer=word_tokenizer,
             chunk_length_schedule=chunk_length_schedule,
             enable_ssml_parsing=enable_ssml_parsing,
-            language_code=language_code,
+            language=language,
         )
         self._session = http_session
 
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -141,6 +144,9 @@ def __init__(
         api_key = api_key or os.environ.get("ELEVEN_API_KEY")
         if not api_key:
             raise ValueError("ELEVEN_API_KEY must be set")
+        
+        if language_code is not None and model != "eleven_turbo_v2_5":
+            raise ValueError("language_code is only supported for the 'eleven_turbo_v2_5' model")
```
bcherry, line 149: It would be better to leave this to the ElevenLabs API. What is its behavior right now if you send language_code on a different model? Is it ignored, or is it a 4xx error?

bcherry:
@theomonnom this seems reasonable and pretty much identical to the way this is supported for the Cartesia plugin already https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py#L77

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -80,6 +80,7 @@ class _TTSOptions:
     api_key: str
     voice: Voice
     model: TTSModels | str
+    language_code: str | None
```
theomonnom, line 83: I think we should rename language_code to language for consistency

theomonnom (approved):
otherwise lgtm

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
cch41: It throws a 400 error. Want me to just log a warning message instead? Also, should I disclude language_code if model != eleven_turbo_v2_5 in the api call then?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
cch41: renamed

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
cch41: set it to None in this case and logged a warning

cch41:
Think the check fails are unrelated

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
bcherry: fwiw i disagree - language_code is the underlying name in the elevenlabs API

cch41:
@theomonnom mind re-running the tests so this can be merged?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -141,6 +144,10 @@ def __init__(
         api_key = api_key or os.environ.get("ELEVEN_API_KEY")
         if not api_key:
             raise ValueError("ELEVEN_API_KEY must be set")
+        
+        if language is not None and model != "eleven_turbo_v2_5":
+            language = None
+            logger.warning("Excluding 'language' as it is only supported for the 'eleven_turbo_v2_5' model")
```
theomonnom, line 150: Let's do nothing by default? I think 11labs will throw an error if we use a different language no?
Otherwise lgtm

theomonnom:
Can you run `ruff format` and `ruff check --fix`?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
cch41: Yea it will throw an error, will remove this

cch41:
>Can you run ruff format and ruff check --fix?
Screenshot 2024-11-01 at 1 45 17 PM
rebased, removed the language check, and ran ruff

theomonnom (approved):
Thanks lgtm!

Issue #980, opened 2024-10-24
Integration with FAL's Whizper API


davidzhao:
[Whizper API](https://fal.ai/models/fal-ai/wizper/api#queue-submit) has a fairly low word-error rate: https://artificialanalysis.ai/speech-to-text

We'd like to have a plugin for it (if it doesn't work with OpenAI's STT plugin)

PR #991, opened 2024-10-28
add FAL's whizper support

jayeshp19:
implements https://github.com/livekit/agents/issues/980

PTAL @davidzhao @theomonnom


livekit-plugins/livekit-plugins-fal/livekit/plugins/fal/stt.py
```diff
@@ -0,0 +1,132 @@
...
+        language="en",
+        task="transcribe",
+        chunk_level="segment",
+        version="3",
+        api_key=None,
+    ):
+        super().__init__(
+            capabilities=STTCapabilities(streaming=False, interim_results=True)
+        )
+        self._api_key = api_key or os.getenv("FAL_API_KEY")
```
egoldschmidt, line 36: fwiw it doesn't look like `self._api_key` is used anywhere, and the fal client seems to expect a different env var, `FAL_KEY`, so this isn't working out of the box for me right now. Hopefully helpful!

livekit-plugins/livekit-plugins-fal/livekit/plugins/fal/stt.py
```diff
@@ -0,0 +1,132 @@
...
+
+        with wave.open(io_buffer, "wb") as wav:
+            wav.setnchannels(merged_buffer.num_channels)
+            wav.setsampwidth(2)  # 16-bit
+            wav.setframerate(merged_buffer.sample_rate)
+            wav.writeframes(bytes(merged_buffer.data))
+
+        return io_buffer.getvalue()
+
+    # def _create_data_uri(self, wav_bytes: bytes) -> str:
```
egoldschmidt, line 74: Curious - did you get a sense for the perf tradeoffs between data URI and uploading/submitting like you do below? My assumption is the latter would be worse for realtime use cases but I'd love your thoughts

livekit-plugins/livekit-plugins-fal/livekit/plugins/fal/stt.py
```diff
@@ -0,0 +1,132 @@
...
+    #     """Create a properly formatted data URI from WAV bytes."""
+    #     base64_audio = base64.b64encode(wav_bytes).decode('utf-8')
+    #     return f"data:audio/wav;base64,{base64_audio}"
```
egoldschmidt, line 77: Also if you were having trouble with this approach I think you might need `audio/x-wav` to be the MIME type

Also `fal_client.encode` basically does the same thing too

livekit-plugins/livekit-plugins-fal/livekit/plugins/fal/stt.py
jayeshp19: good catch @egoldschmidt  I will update it next commit.

livekit-plugins/livekit-plugins-fal/livekit/plugins/fal/stt.py
jayeshp19: This is very helpful. I'll test and update the PR

livekit-plugins/livekit-plugins-fal/livekit/plugins/fal/stt.py
theomonnom: +1 Agree with that. The less request we make the better :)

livekit-plugins/livekit-plugins-fal/livekit/plugins/fal/stt.py
```diff
@@ -0,0 +1,110 @@
...
+                    "language": config.language,
+                    "chunk_level": config.chunk_level,
+                    "version": config.version,
+                },
+            )
+
+            try:
+                result = await fal_client.result_async(
+                    "fal-ai/wizper", handler.request_id
+                )
```
theomonnom, line 93: Do we have to do two requests with their API? Or there is a way to do one one?

livekit-plugins/livekit-plugins-fal/livekit/plugins/fal/stt.py
```diff
@@ -0,0 +1,110 @@
...
+                text = result.get("text", "")
+                return self._transcription_to_speech_event(text=text)
+            except fal_client.client.FalClientError as e:
+                logger.error(f"FAL AI API error: {e}")
+                return self._transcription_to_speech_event(text="")
```
theomonnom, line 98: We should make the users aware when there is an exception. Can you raise one of the exceptions [here](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/_exceptions.py)

livekit-plugins/livekit-plugins-fal/livekit/plugins/fal/stt.py
egoldschmidt: Yeah I suggest we call run_async instead of submit & result, that way it's only one network request

theomonnom:
Types seems wrong, can you add this plugin to https://github.com/livekit/agents/blob/0c9eac91938d21ba4c2f2d16cfe05fb8fa237786/.github/workflows/tests.yml#L102

theomonnom:
sry can you also add it here :) https://github.com/livekit/agents/blob/main/.github/workflows/ci.yml

theomonnom:
mmh note usure maybe you need a py.typed file ([e.g](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/py.typed))

jayeshp19:
I've added py.typed and changed initial version to 0.1.0 from 0.0.0,
still facing the same issue @theomonnom

theomonnom:
Pretty sure the types aren't good, but the CI isn't working for whatever reasons. Can you run mypy locally on your machine to do type checks?

davidzhao:
@jayeshp19 `py.typed` is added to the wrong dir

davidzhao (approved):

PR #993, opened 2024-10-28
Add Events Before LLM & TTS Invocation

Toubat:
Add two new events for assistant:

```py
@assistant.on("llm_started_streaming")
def before_calling_llm(chat_ctx: ChatContext):
    pass

@assistant.on("tts_started_streaming")
def before_calling_tts(source: AsyncIterator[str] | str):
    pass
```


martin-purplefish:
Why not just use before_tts_cb and before_llm_cb directly?

Toubat:
cc @davidzhao @theomonnom

PR #998, opened 2024-10-29
Send JobStatus Updates

keepingitneil:


nbsp (approved):
weak lgtm, haven't messed with python in a while now but the logic looks great

PR #999, opened 2024-10-29
pipelineagent: add metrics example

theomonnom:
- also fix metrics being emitted twice when using the StreamAdapter

### DIFF
--- a/examples/voice-pipeline-agent/pipeline_metrics.py
+++ b/examples/voice-pipeline-agent/pipeline_metrics.py
@@ -16,6 +16,12 @@ load_dotenv()
 logger = logging.getLogger("metrics-example")
 
 
+OPENAI_LLM_INPUT_PRICE = 2.50 / (10**6)  # $2.50 per million tokens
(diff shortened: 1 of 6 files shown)


examples/voice-pipeline-agent/pipeline_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @agent.on("metrics_collected")
+    def on_metrics_collected(metrics: PipelineMetrics):
+        if metrics["type"] == "vad_metrics":
+            return  # don't log VAD metrics because it is noisy
+
+        logger.info(f"metrics collected: {metrics}")
+
+        # calculate the cost of a LLM request
+        if metrics["type"] == "llm_metrics":
```
davidzhao, line 51: it seems like it'd be better to aggregate these and compute total cost at the end? ideally each metric is printed out in real time (for duration/speed/etc metrics), and then the total cost is computed during context shutdown

davidzhao (approved):
awesome, lg!

keepingitneil (approved):

Issue #877, opened 2024-10-09
Expose transcription events to the agent (multimodal)


mike-r-mclaughlin:
Users need a way to get the transcript for both the agent and participant(s).  Currently, the expectation (based on [docs](https://docs.livekit.io/agents/voice-agent/transcriptions/)) is that the developer would listen for the `transcription_received` event.  However, this event isn't emitted.  

Could we provide a way for agents to access the transcripts from within the Python (and NodeJS) code?

gauravkesharwani:
This a major issue which is preventing us from upgrading to multimodal agent, quick resolution is highly appreciated....

theomonnom:
Fixed by https://github.com/livekit/agents/pull/1001
It is going to be released on livekit-agents==0.11.0 (tmrw)

kylekaplan:
Hi, I see that `agent_speech_committed` event has been added, but this seems to wait for the entire response before emitting. Is there a way to get intermediate transcriptions from the agent? Thanks!

creightontaylor:
`agent_speech_committed` and `user_speech_committed` are not emitting for my multimodal node.js project (created from your quickstart documentation) and they are not listed as events in the documentation (unlike Python). Will these be added soon? They are fairly crucial for us and the last major component for us to start testing with users.

mike-r-mclaughlin:
@creightontaylor : see https://github.com/livekit/agents-js/issues/164.  We're working on it this week.

creightontaylor:
@mike-r-mclaughlin awesome! I eagerly await this release! I'm confidently we'll figure something out because you already have it for Python. I'm prepping with some functionality around it.

PR #1001, opened 2024-10-29
feat: expose user_speech_committed and agent_speech_committed for multimodal agents

longcw:
- Expose the `user_speech_committed`, `agent_speech_committed`, and `agent_speech_interrupted` for multimodal agents
- Emit `user_started_speaking` and `user_stopped_speaking` for multimodal agents
- Fixed a bug that `handle._tr_fwd` of multimodal agents cannot be closed for some interruption cases

Related to: https://github.com/livekit/agents/issues/877

### DIFF
--- a/.changeset/proud-wombats-perform.md
+++ b/.changeset/proud-wombats-perform.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+expose transcriptions for multimodal agents
--- a/livekit-agents/livekit/agents/multimodal/agent_playout.py
+++ b/livekit-agents/livekit/agents/multimodal/agent_playout.py
@@ -165,6 +165,9 @@ class AgentPlayout(utils.EventEmitter[EventTypes]):
 
             await utils.aio.gracefully_cancel(read_text_task)
 
+            # make sure the text_data.sentence_stream is closed
+            handle._tr_fwd.mark_text_segment_end()
+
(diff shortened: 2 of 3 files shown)


theomonnom:
Can you add a `changesets`?

theomonnom:
Any reason why tr_frwd.aclose isn't closing correctly the sentence streams?
https://github.com/livekit/agents/blob/850f8e07fba64a0b3b7bf8de67d8eeb0f5c037dd/livekit-agents/livekit/agents/transcription/tts_forwarder.py#L226

ideally, it is closing directly, even if mark_text_segment_end isn't called

theomonnom:
Oh I guess we're using pop, so it is being removed from the queue too early
https://github.com/livekit/agents/blob/850f8e07fba64a0b3b7bf8de67d8eeb0f5c037dd/livekit-agents/livekit/agents/transcription/tts_forwarder.py#L288

Ideally, we only pop and the end of the iteration

longcw:
> Oh I guess we're using pop, so it is being removed from the queue too early
> 
> https://github.com/livekit/agents/blob/850f8e07fba64a0b3b7bf8de67d8eeb0f5c037dd/livekit-agents/livekit/agents/transcription/tts_forwarder.py#L288
> 
> 
> 
> Ideally, we only pop and the end of the iteration

Yes, I think so.

Issue #1000, opened 2024-10-29
"/sys/fs/cgroup/cpu.max" not found


longcw:
For some reason there is no "/sys/fs/cgroup/cpu.max" on my linux system. 
```
$ ls /sys/fs/cgroup/cpu*
/sys/fs/cgroup/cpu.pressure           /sys/fs/cgroup/cpuset.mems.effective  /sys/fs/cgroup/cpu.stat.local
/sys/fs/cgroup/cpuset.cpus.effective  /sys/fs/cgroup/cpu.stat
```
Should we add a fallback when this file is missing, for example
```python
    def _read_cpu_max(self) -> tuple[str, int]:
        try:
            with open("/sys/fs/cgroup/cpu.max", "r") as f:
                data = f.read().strip().split()
            quota = data[0]
            period = int(data[1])
        except FileNotFoundError:
            quota = "max"
            period = 100000
        return quota, period
```

davidzhao:
oh good catch.. wanna open a PR?

PR #1002, opened 2024-10-29
fix: fallback when cpu.max not found

longcw:
Fix https://github.com/livekit/agents/issues/1000


davidzhao (approved):
LG!

theomonnom:
Should we fallback to using psutil when this file doesn't exist instead?

davidzhao:
no, it's possible for cgroups to report stats, but have no quotas set. defaulting to max will allow us use actual core count from the machine

PR #1003, opened 2024-10-29
Cancel main task in aclose

martin-purplefish:
Also simplified a few places in the main loop code - no other functional changes.

This wasn't really a big deal, since the job logic would clean it up anyway. But if someone was using it outside of a job, it'd run forever.


martin-purplefish:
PTAL @theomonnom

PR #1006, opened 2024-10-29
Allow kind to be list or single value

keepingitneil:


livekit-agents/livekit/agents/job.py
```diff
@@ -183,7 +190,8 @@ def add_participant_entrypoint(
             [JobContext, rtc.RemoteParticipant], Coroutine[None, None, None]
         ],
         *_,
```
theomonnom, line 192: what is the diff between *_ and *?

keepingitneil: IDE complains about *, maybe when it's not the first arg?

PR #1007, opened 2024-10-30
prepare for release

theomonnom:


.github/workflows/check-types.yml
```diff
@@ -1,4 +1,4 @@
-name: Check Types
+name: CI
```
nbsp, line 1: same name as ruff.yml?

nbsp (approved):
ci nit but otherwise lg

PR #1009, opened 2024-10-30
add metrics utilities & use dataclasses instead of TypedDict

davidzhao:


livekit-agents/livekit/agents/metrics/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    llm_completion_tokens: int = 0
+    tts_characters_count: int = 0
+    stt_audio_duration: float = 0
+
+
+def create_metrics_logger(
+    logger: logging.Logger | None = None,
+) -> tuple[Callable[[PipelineMetrics], PipelineUsageSummary], None]:
+    if logger is None:
+        logger = logging.getLogger("")
```
theomonnom, line 20: let's use the default logger of livekit-agents instead
`from ..log import logger`

theomonnom (approved):

PR #1015, opened 2024-10-31
feat: sync chat_ctx for openai RealtimeModel from and to the remote realtime session

longcw:
- write transcripts to chat context
- update realtime session based on input chat_ctx
- add `sync_chat_ctx` to sync the local chat context to OAI realtime session


longcw:
Some known issues to be solved:
1. The user speaking at the beginning before the sync is finished may cause an error and the agent hangs with `ERROR livekit.plugins.openai.realtime - OpenAI S2S error {'type': 'error', 'event_id': 'event_APLDybDWwpjHeB4rVohDV', 'error': {'type': 'invalid_request_error', 'code': None, 'message': 'Only model output audio messages can be truncated', 'param': None, 'event_id': None}}`
2. The function calls are also stored as conversation items on OAI, for now the local `chat_ctx` only has the chat items that may mismatch with the remote version when there are function calls.

davidzhao:
> 1. The user speaking at the beginning before the sync is finished may cause an error and the agent hangs with `ERROR livekit.plugins.openai.realtime - OpenAI S2S error {'type': 'error', 'event_id': 'event_APLDybDWwpjHeB4rVohDV', 'error': {'type': 'invalid_request_error', 'code': None, 'message': 'Only model output audio messages can be truncated', 'param': None, 'event_id': None}}`

For this issue, could we address by ignoring user input during the sync? how long does that take typically in your tests?

longcw:
> For this issue, could we address by ignoring user input during the sync? how long does that take typically in your tests?

Yes, I think so. The sync takes just 1-2 seconds on my end. I'll try to understand the root cause and address it.

longcw:
> Some known issues to be solved:
> 
> 1. The user speaking at the beginning before the sync is finished may cause an error and the agent hangs with `ERROR livekit.plugins.openai.realtime - OpenAI S2S error {'type': 'error', 'event_id': 'event_APLDybDWwpjHeB4rVohDV', 'error': {'type': 'invalid_request_error', 'code': None, 'message': 'Only model output audio messages can be truncated', 'param': None, 'event_id': None}}`
> 2. The function calls are also stored as conversation items on OAI, for now the local `chat_ctx` only has the chat items that may mismatch with the remote version when there are function calls.

The first problem I found with the agent hanging was because the real-time API responded with text instead of audio in this case. In some cases the API will only output in text if a text chat context is set.

I disabled the audio playout in ([this commit](https://github.com/livekit/agents/pull/1015/commits/5745e30f91e3e062dd8b2606ffd24a41f3a5ccb3)) if the response is a text to avoid forever `await` on the audio buffer, but when the API outputs a text it will always output in text. Do you have any idea about this? @theomonnom @davidzhao

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -767,6 +798,35 @@ def session_update(
             }
         )
 
+    def _sync_chat_ctx_to_session(
```
theomonnom, line 801: Do you think this can be an async function that waits for the OAI answers?
We're currently emitting `conversation_item_deleted` and `conversation_item_created` instead.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -660,6 +685,12 @@ async def aclose(self) -> None:
     def chat_ctx(self) -> llm.ChatContext:
         return self._chat_ctx
 
+    @chat_ctx.setter
+    def chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
+        """Sync the ctx to the session and reset the chat context."""
+        self._sync_chat_ctx_to_session(self._chat_ctx, chat_ctx)
+        self._chat_ctx = chat_ctx
```
theomonnom, line 692: I don't think we should add a setter. Instead we should have a public `sync_ctx` function.
I guess RealtimeModel will needs to have 2 llm.ChatContext (one with the current server states, and one that the user can edit)

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
theomonnom: I guess `conversation.item.create` needs to be async?

livekit-agents/livekit/agents/utils/message_change.py
```diff
@@ -0,0 +1,177 @@
...
+        ],
+    )
+
+
+def _compute_list_changes(old_list: list[T], new_list: list[T]) -> MessageChange[T]:
+    """Compute minimum changes needed to transform old_list into new_list
+
+    Rules:
+    - Delete first, then insert
+    - Can't insert at start if list not empty (must delete all first)
```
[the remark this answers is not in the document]
theomonnom: To add to this comment, I'm mostly mentioning it because you can also send "audio data" as part of the message (for user messages). So it requires more bandwidth

theomonnom:
> 2. The function calls are also stored as conversation items on OAI, for now the local `chat_ctx` only has the chat items that may mismatch with the remote version when there are function calls.

The chat_ctx does support function tools messages. E.g on how we handle it for the [PipelineAgent](https://github.com/livekit/agents/blob/381421e14e6c2235e8c131a91f893b9f67b0d132/livekit-agents/livekit/agents/pipeline/pipeline_agent.py#L787)

theomonnom:
> I disabled the audio playout in ([this commit](https://github.com/livekit/agents/pull/1015/commits/5745e30f91e3e062dd8b2606ffd24a41f3a5ccb3)) if the response is a text to avoid forever await on the audio buffer, but when the API outputs a text it will always output in text. Do you have any idea about this?

Not sure to understand what you mean by this? Is there no audio being generated?

livekit-agents/livekit/agents/utils/message_change.py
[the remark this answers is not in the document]
longcw: It will delete all items only if we want to insert a new item to the first. It seems that the API can only insert "after an item" with the `previous_item_id` so if the session is not empty there is no way to insert to the beginning. Please let me know if there is a way to do that.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: sounds good!

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: Yeah, I am thinking something similar that we should keep a "copy" of server state in the RealtimeModel that user cannot edit. I'll add that.

longcw:
> > 2. The function calls are also stored as conversation items on OAI, for now the local `chat_ctx` only has the chat items that may mismatch with the remote version when there are function calls.
> 
> The chat_ctx does support function tools messages. E.g on how we handle it for the [PipelineAgent](https://github.com/livekit/agents/blob/381421e14e6c2235e8c131a91f893b9f67b0d132/livekit-agents/livekit/agents/pipeline/pipeline_agent.py#L787)

So you mean to append the function tool messages to the chat_ctx, right? That makes sense.

longcw:
> > I disabled the audio playout in ([this commit](https://github.com/livekit/agents/pull/1015/commits/5745e30f91e3e062dd8b2606ffd24a41f3a5ccb3)) if the response is a text to avoid forever await on the audio buffer, but when the API outputs a text it will always output in text. Do you have any idea about this?
> 
> Not sure to understand what you mean by this? Is there no audio being generated?

No, there is no audio generated in that case, the API response in `text.delta` instead. Here are some example messages from the realtime API
```
received WSMessage(type=<WSMsgType.TEXT: 1>, data='{"type":"response.text.delta","event_id":"event_APlzgvMN6oippCHur2YKD","response_id":"resp_APlzfiKfj4UAJHFskFcHd","item_id":"item_APlzfSORFWckGGCUgaNCg","output_index":0,"content_index":0,"delta":" How"}', extra='')
received WSMessage(type=<WSMsgType.TEXT: 1>, data='{"type":"response.text.delta","event_id":"event_APlzgjnpD7buTPItBG2XQ","response_id":"resp_APlzfiKfj4UAJHFskFcHd","item_id":"item_APlzfSORFWckGGCUgaNCg","output_index":0,"content_index":0,"delta":" are"}', extra='')
received WSMessage(type=<WSMsgType.TEXT: 1>, data='{"type":"response.text.delta","event_id":"event_APlzg67kWskfiC2Iijy4Z","response_id":"resp_APlzfiKfj4UAJHFskFcHd","item_id":"item_APlzfSORFWckGGCUgaNCg","output_index":0,"content_index":0,"delta":" you"}', extra='')
received WSMessage(type=<WSMsgType.TEXT: 1>, data='{"type":"response.text.delta","event_id":"event_APlzgLiTHPW6ctfJm7W68","response_id":"resp_APlzfiKfj4UAJHFskFcHd","item_id":"item_APlzfSORFWckGGCUgaNCg","output_index":0,"content_index":0,"delta":"?"}', extra='')
received WSMessage(type=<WSMsgType.TEXT: 1>, data='{"type":"response.text.done","event_id":"event_APlzgNY8o4MYrIRjeHZYk","response_id":"resp_APlzfiKfj4UAJHFskFcHd","item_id":"item_APlzfSORFWckGGCUgaNCg","output_index":0,"content_index":0,"text":"Hello! I\'m doing well, thank you. How are you?"}', extra='')
```
The main branch currently doesn't check the part type of the `response.content part.added` message, which creates a playout instance waiting for the audio, which never comes, so the audio stream doesn't close.

But the main problem is how to make sure the API always returns audio instead of text.

davidzhao:
> But the main problem is how to make sure the API always returns audio instead of text.

if the API doesn't return audio for any reason, can make it so the agent doesn't wait for playout?

livekit-agents/livekit/agents/utils/message_change.py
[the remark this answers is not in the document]
theomonnom: Ah this makes sense, kind of weird their API doesn't support to prepend an item 😅

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: Created a `_conversation_items` (maybe give it another name later) in the realtime session to track the item created and deleted events from the API. The content of the item like user and agent transcriptions and the function call outputs will be updated to the item from the multimodal agent.

The `_session.chat_ctx` will be created from the tracked conversation items and user can sync a `chat_ctx` to the session as well.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: Add a async version of item create.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/converstation_items.py
```diff
@@ -0,0 +1,126 @@
...
+@dataclass
+class ConversationItem:
+    """A node in the conversation linked list"""
+
+    message: llm.ChatMessage
+    _prev: Optional[ConversationItem] = field(default=None, repr=False)
+    _next: Optional[ConversationItem] = field(default=None, repr=False)
+
+
+class ConversationItems:
```
longcw, line 21: This is a double linked list to mimic the item operation in the Realtime API. Let me know if this is necessary or if there is any other tools for easy deleting and inserting items to the list. And this may need a rename to avoid confusion with the Conversation Items in the API.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -767,6 +810,82 @@ def session_update(
...
+        del self._item_created_futs[message.id]
+
+    async def aitem_delete(self, item_id: str) -> None:
+        fut = asyncio.Future[None]()
+        self._item_deleted_futs[item_id] = fut
+        self.conversation.item.delete(item_id)
+        await fut
+        del self._item_deleted_futs[item_id]
+
+    async def async_chat_ctx(self, new_ctx: llm.ChatContext) -> None:
```
theomonnom, line 834: I was more thinking about making the current API async. Feel free to break the API. The RealtimeModel API is still beta :)

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/converstation_items.py
theomonnom: I think this is OK since this is internal. Maybe `RemoteConversationItems`?
Let're prepend '`_`' to the file name and classes to explicitly mark them as internal

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: Do you mean to make the `self.conversation.item.create` and `self.conversation.item.delete` be async? 

Maybe I can move the `acreate` and `adelete` to the `conversation.item` but keep the original version of create and delete, as for some cases like `sync_chat_ctx` we may want to add all the create and delete messages to the event queue at once and wait at the end?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/converstation_items.py
longcw: Sound good.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
theomonnom: > Do you mean to make the self.conversation.item.create and self.conversation.item.delete be async?
Yep.

I don't think we need two functions (sync and async)

theomonnom:
What do you think of an API looking like:

```
model = RealtimeModel()
model.chat_ctx.append(...)
model.chat_ctx.append(role="user", ...)
await model.sync_chat_ctx() # Here sync_chat_ctx doesn't have arguments
```


If you want to replace the whole chat_ctx we can do smthg like

```
model.chat_ctx = my_new_chat_ctx
await model.sync_chat_ctx()

```

theomonnom (approved):
See message on Slack, otherwise lgtm!

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: My concern is that when we call `gather(item.acreate(item1), item.create(item2))`, is there a guarantee for the two items are pushed to the message queue in the correct order?

longcw:
> What do you think of an API looking like:
> 
> ```
> model = RealtimeModel()
> model.chat_ctx.append(...)
> model.chat_ctx.append(role="user", ...)
> await model.sync_chat_ctx() # Here sync_chat_ctx doesn't have arguments
> ```
> 
> If you want to replace the whole chat_ctx we could do smthg like
> 
> ```
> model.chat_ctx = my_new_chat_ctx
> await model.sync_chat_ctx()
> ```

The following is the current API, the user always needs to grab a copy of the current chat_ctx then modify it
```python
updated_ctx = model.chat_ctx  # maybe make this a method to emphasize this returns a copy of the ctx 
updated_ctx.append(...)
await model.sync_chat_ctx(updated_ctx)
```

I think your version is more straightforward, but there might be a problem that the internal managed ctx and the user-modifiable ctx might be out of sync when user modified the latter without a immediate `await sync_chat_ctx`.

For example, if user call `model.chat_ctx.append(...)` or `model.chat_ctx = my_new_chat_ctx` without a following `await model.sync_chat_ctx()` (maybe forget or not in time) before the internal chat ctx is updated with new conversations, the `sync_chat_ctx` will actually remove the new conversations which may not be expected.

davidzhao:
it'd be great to explicitly call out that you are getting a copy of the context.. i.e.

```python
ctx_copy = model.chat_ctx_copy()
...
await model.set_chat_ctx(ctx_copy)
```

if we are simply replacing, instead of modifying in-place, then I think `set_` would be a better API compared to `sync_`

longcw:
@davidzhao  I have updated the API as you suggested.

examples/multimodal_agent.py
```diff
@@ -67,6 +68,23 @@ async def get_weather(
     #     fnc_ctx=fnc_ctx,
     # )
 
+    # create a chat context with chat history
+    chat_ctx = llm.ChatContext()
+
+    # Add some test context to verify if the sync_chat_ctx works
+    # FIXME: OAI realtime API does not support this properly when the chat context is too many
+    # It may answer with the text responses only for some cases
```
theomonnom, line 76: Seems like this comment is now outdated?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -865,6 +851,29 @@ async def set_chat_ctx(self, new_ctx: llm.ChatContext) -> None:
             },
         )
 
+        # append an empty audio message if all messages are text
+        if new_ctx.messages and all(
+            isinstance(msg.content, str) for msg in new_ctx.messages
+        ):
```
theomonnom: Can you check if `audio` is inside the modality

theomonnom:
> it'd be great to explicitly call out that you are getting a copy of the context.. i.e.
> 
> ```python
> ctx_copy = model.chat_ctx_copy()
> ...
> await model.set_chat_ctx(ctx_copy)
> ```
> 
> if we are simply replacing, instead of modifying in-place, then I think `set_` would be a better API compared to `sync_`

This makes sense, but it would be great to still allow in-place edit. Since ppl have access to the model.chat_ctx, they may expect that it is editable.

So adding `sync_chat_ctx` will start the synchronization after the user did some updates. I think it is OK to have both, `set_chat_ctx` when you want to update the whole context. or simply call `sync_chat_ctx` if you directly did some small changes.
Wdyt?

theomonnom:
Awesome! Only small nits are remaining, can you fix the type-check CI?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: Changed to 
```
if new_ctx.messages and not any(
            isinstance(msg.content, llm.ChatAudio) for msg in new_ctx.messages
        ):
```

longcw:
> > it'd be great to explicitly call out that you are getting a copy of the context.. i.e.
> > ```python
> > ctx_copy = model.chat_ctx_copy()
> > ...
> > await model.set_chat_ctx(ctx_copy)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > if we are simply replacing, instead of modifying in-place, then I think `set_` would be a better API compared to `sync_`
> 
> This makes sense, but it would be great to still allow in-place edit. Since ppl have access to the `model.chat_ctx`, they may expect that it is editable.
> 
> So adding `sync_chat_ctx` will start the synchronization after the user did some updates. I think it is OK to have both, `set_chat_ctx` when you want to update the whole context. or simply call `sync_chat_ctx` if you directly did some small changes. Wdyt?

Actually it has only `model.chat_ctx_copy()` now. As `model.chat_ctx` is always a copy of the current ctx, we may don't want user to do something like `model.chat_ctx.append()`.

longcw:
> Awesome! Only small nits related to the API are remaining. Can you also fix the type-check CI?

Fixed.

davidzhao:
> This makes sense, but it would be great to still allow in-place edit. Since ppl have access to the `model.chat_ctx`, they may expect that it is editable.

It'll more likely race when users edit in place while the realtime API may add to it. it can lead to a bit of unpredictability in terms of what is in the chat history.

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
```diff
@@ -134,12 +142,30 @@ def start(
         self._session = self._model.session(
             chat_ctx=self._chat_ctx, fnc_ctx=self._fnc_ctx
         )
-        self._main_atask = asyncio.create_task(self._main_task())
+
+        # Create a task to wait for initialization and start the main task
+        async def _init_and_start():
+            try:
+                await self._session._init_sync_task
```
kkacquah, line 149: Weirdly, I was testing out this branch using the virtual assistant sandbox playground had an issue in which the agent wouldn't start, while waiting for this task to complete. Does that make sense at all? I removed it temporarily from my local `venv` and the agent works fine, albiet after a 2 second waiting period to sync the initial context.

longcw: Can you share the package and the version you installed with the issue? I can take a look and try to reproduce it.

kkacquah: Using the latest commit actually helped fix the issue, apologies! Thanks for the tip that it could've been an issue with the commit I was on.

theomonnom:
> > This makes sense, but it would be great to still allow in-place edit. Since ppl have access to the `model.chat_ctx`, they may expect that it is editable.
> 
> It'll more likely race when users edit in place while the realtime API may add to it. it can lead to a bit of unpredictability in terms of what is in the chat history.

Stuff can't race if the whole "edit code" is sync and the first thing async fnc to be called is model.sync_chat_ctx()

davidzhao:
> Stuff can't race if the whole "edit code" is sync and the first async fnc to be called is model.sync_chat_ctx()

It's more that there may be content added to the context that that's unexpected by the dev. imagine the following:

* user is speaking to agent
* dev calls `.append()` to prompt the LLM to return something else
* user transcript committed, and gets added to the end of context
* dev calls `.sync_chat_ctx()`, which actually syncs the user prompt, instead of dev prompt

the overall point is that if two separate processes could modify chat_context, the order and content of the message history would end up being unpredictable.

theomonnom:
> > Stuff can't race if the whole "edit code" is sync and the first async fnc to be called is model.sync_chat_ctx()
> 
> It's more that there may be content added to the context that that's unexpected by the dev. imagine the following:
> 
> * user is speaking to agent
> * dev calls `.append()` to prompt the LLM to return something else
> * user transcript committed, and gets added to the end of context
> * dev calls `.sync_chat_ctx()`, which actually syncs the user prompt, instead of dev prompt
> 
> the overall point is that if two separate processes could modify chat_context, the order and content of the message history would end up being unpredictable.

Agree it could create more confusion. Let's see if ppl are asking for direct access.

theomonnom (approved):

theomonnom (approved):

theomonnom (approved):

PR #1020, opened 2024-10-31
support for additional deepgram config

VinothRajasekar:
### 1. Addition of `additional_config` Parameter
- The `STTOptions` dataclass now includes an `additional_config` field, which is a dictionary for storing extra configuration parameters for deepgram.
- This allows users to pass parameters like `paragraph` and `diarize` when initializing the `STT` class.

### 2. Integration of `additional_config` in STT Initialization
- The `STT` class constructor now accepts an `additional_config` argument, which is then passed to the `STTOptions` dataclass.
- This dictionary dynamically updates configuration options during the `_sanitize_options` method call.

### 3. Dynamic Configuration Handling
- The `_sanitize_options` method has been updated to incorporate additional configuration options specified in `additional_config`.
- This ensures that any extra parameters are included in requests to the Deepgram API.

## Example Usage

Users can now initialize the `STT` class with additional configuration parameters as follows:

```python

stt_instance = deepgram.STT(
    additional_config={
        "paragraph": True,
        "diarize": True
    }
)

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -98,6 +98,7 @@ class STTOptions:
     num_channels: int
     keywords: list[Tuple[str, float]]
     profanity_filter: bool
+    additional_config: dict = dataclasses.field(default_factory=dict)
 
 
 class STT(stt.STT):
@@ -118,6 +119,7 @@ class STT(stt.STT):
         profanity_filter: bool = False,
         api_key: str | None = None,
         http_session: aiohttp.ClientSession | None = None,
+        additional_config: dict | None = None,
     ) -> None:
         """
         Create a new instance of Deepgram STT.
@@ -168,6 +170,7 @@ class STT(stt.STT):
             num_channels=1,
             keywords=keywords,
             profanity_filter=profanity_filter,
+            additional_config=additional_config or {},
         )
         self._session = http_session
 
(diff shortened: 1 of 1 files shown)


theomonnom:
Hey, I think it would be better to explicitly support those fields inside the constructor?

VinothRajasekar:
> Hey, I think it would be better to explicitly support those fields inside the constructor?

Hey, thanks for the feedback! Sorry for getting back late on this one. I completely understand the idea of making paragraph and diarize explicit in the constructor. However,I would like to highlight why keeping additional_config as the main way to pass deepgram parameters is a more scalable and flexible approach:

**Avoids Frequent PRs & Code Changes**
Deepgram continuously evolves, adding new parameters and features. If we hardcode specific parameters like paragraph and diarize inside our constructor, we’ll need to modify the code and submit a PR every time a new parameter is introduced or needs to be added

**Maximizes Flexibility for Users**
By keeping additional_config, we give users the freedom to pass any Deepgram-supported parameter dynamically without waiting for us to explicitly support it in the livekit deepgram plugin. This makes the integration more future-proof.

**Deepgram Already Handles Validation**
Deepgram itself validates the provided parameters. If a user mistakenly provides an invalid parameter, Deepgram returns appropriate error messages. This eliminates the need for us to maintain separate validation logic and ensures that we are always in sync with Deepgram’s latest updates.

**Explicit Parameters Can Be Restrictive**
While paragraph and diarize are useful today, there might be other parameters (e.g., model, tier, language) that different users might find more important. If we explicitly support some and not others, it provides some inflexibility and have to do a PR everytime to add new deepgram parameters.

please let me know what you think. thanks.

Issue #1010, opened 2024-10-30
Expose metrics from MultimodalAgent and RealtimeModel


davidzhao:

davidzhao:
please feel free to submit a PR :)

davidzhao:
MultimodalAgent metrics are in #1080

PR #1036, opened 2024-11-04
Expose usage metrics for Realtime model

yuyuma:
Handles #1010  to include usage (number of input/output tokens) in the response_done event for realtime model

Conforms to the usage object model as shown at https://platform.openai.com/docs/api-reference/realtime-server-events/response/done at time of submission


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -972,6 +974,7 @@ def _handle_response_created(
             status=response["status"],
             status_details=status_details,
             output=[],
+            usage=response["usage"],
```
davidzhao, line 977: should this be `.get("usage")` when usage isn't enabled?

theomonnom (approved):
Thanks! lgtm

theomonnom:
Hey, ruff checks seems to fails.

Can you run `ruff format && ruff check --fix`?

PR #1040, opened 2024-11-05
Fix type check failures by returning CPU count as float

davidzhao:


theomonnom (approved):

keepingitneil (approved):

theomonnom (changes requested):
mmh do `or 1` instead, float(None) can raise

livekit-agents/livekit/agents/utils/hw/cpu.py
```diff
@@ -21,7 +21,7 @@ def cpu_percent(self, interval: float = 0.5) -> float:
 
 class DefaultCPUMonitor(CPUMonitor):
     def cpu_count(self) -> float:
-        return psutil.cpu_count()
+        return float(psutil.cpu_count())
```
theomonnom, line 24:
```suggestion
        return psutil.cpu_count() or 1.0
```

theomonnom (approved):

PR #1042, opened 2024-11-05
feat: add support for vertex ai

jayeshp19:


examples/voice-pipeline-agent/vertex_ai.py
```diff
@@ -0,0 +1,80 @@
...
+            "You are a weather assistant created by LiveKit. Your interface with users will be voice. "
+            "You will provide weather information for a given location."
+        ),
+        role="system",
+    )
+    participant = await ctx.wait_for_participant()
+    agent = VoicePipelineAgent(
+        vad=ctx.proc.userdata["vad"],
+        stt=openai.STT(),
+        llm=google.LLM(project="tribal-archery-440118-d1", location="us-central1"),
```
davidzhao: if the only difference between the main function calling example and this one is the google LLM usage, could you just add this line to the other example with this line commented out? also you should remove the project name, (location is fine to keep, but could use a comment indicating as such)

davidzhao: perhaps it's better to indicate something like:

```
    # to use Google AI
    agent = VoicePipelineAgent(
        vad=ctx.proc.userdata["vad"],
        stt=google.STT(),
        llm=google.LLM(project="tribal-archery-440118-d1", location="us-central1"),
        tts=google.TTS(),
    )
```

jayeshp19: yes correct, I'll update it

livekit-plugins/livekit-plugins-google/livekit/plugins/google/models.py
```diff
@@ -1,5 +1,9 @@
 from typing import Literal
 
+ChatModels = Literal[ 
+    "gemini-pro"
```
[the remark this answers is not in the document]
jayeshp19: Sure

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,353 @@
...
+        *,
+        chat_ctx: llm.ChatContext,
+        fnc_ctx: Optional[llm.FunctionContext] = None,
+        temperature: Optional[float] = None,
+        n: Optional[int] = 1,
+        **kwargs,
+    ) -> "LLMStream":
+        logger.info(
+            "Starting chat with temperature: %s, candidate count: %s", temperature, n
+        )
```
theomonnom, line 94: Let's remove logs

theomonnom:
Do you think there is a way we reuse https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/llm/_oai_api.py when running/checking function calls. The good think about it is that it verify the types are correct before running any code

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
jayeshp19: on it

jayeshp19:
@theomonnom I think we could reuse `create_ai_function_info` from `_oai_api.py`, but `build_oai_function_description` wouldn't be compatible, since Vertex AI doesn’t accept the same function description format as OpenAI.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
jayeshp19: updated

jayeshp19:
@theomonnom  I've updated code to utilize `create_ai_function_info` from`_oai_api.py`

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+            raise ValueError
+        if not self._location:
+            logger.error("Location is required for Vertex AI initialization")
+            raise ValueError
+        if not self._gac:
+            logger.error(
+                "`GOOGLE_APPLICATION_CREDENTIALS` environment variable is not set. please set it to the path of the service account key file."
+            )
+            raise ValueError
+
```
theomonnom, line 82: Can you add the messages directly inside the Exception instead of logging them.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+            project,
+            location,
+            model,
+        )
+        super().__init__()
+        self._project = project
+        self._location = location
+        self._model_name = model
+        self._kwargs = kwargs
+        self._gac = os.getenv("GOOGLE_APPLICATION_CREDENTIALS")
```
theomonnom, line 69: Can you follow what we did for auth on the google.STT.

https://github.com/livekit/agents/blob/1a99aae891d6bd440da20f88a3f6e29eab9069cc/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py#L105

As well as the extra arguments for the credentials

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+class LLM(llm.LLM):
+    def __init__(
+        self,
+        *,
+        project: Optional[str] = None,
+        location: Optional[str] = None,
+        model: str | GeminiModels = "gemini-1.5-flash-002",
+        **kwargs,
```
theomonnom, line 56: Let's not use **kwargs inside our constructor. This is error-prone

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+                    )
+                ],
+            )
+
+        return None
+
+
+def _build_vertex_context(chat_ctx: llm.ChatContext, cache_key: Any) -> List[dict]:
+    contents = []
+    current_entry: Optional[dict] = None
```
theomonnom, line 244: The typing isn't very useful here

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+
+from __future__ import annotations
+
+import inspect
+import json
+import os
+from typing import Any, AsyncIterator, Awaitable, Dict, List, Optional
+from uuid import uuid4 as uuid
+
+import vertexai  # type: ignore
```
theomonnom, line 23: I don't think we should ignore the types of those packages

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+        vertexai.init(project=project, location=location)
+
+    def chat(
+        self,
+        *,
+        chat_ctx: llm.ChatContext,
+        fnc_ctx: Optional[llm.FunctionContext] = None,
+        temperature: Optional[float] = None,
+        n: Optional[int] = 1,
+        **kwargs,
```
theomonnom, line 92: Same, can we remove kwargs

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+            if current_entry:
+                contents.append(current_entry)
+            current_entry = {"role": role, "parts": parts}
+        else:
+            current_entry["parts"].extend(parts)
+
+    if current_entry:
+        contents.append(current_entry)
+
+    logger.debug("Built vertex context.")
```
theomonnom, line 268:
```suggestion
```

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+            self._vai_stream = await self._awaitable_vai_stream
+        try:
+            async for response in self._vai_stream:
+                for idx, candidate in enumerate(response.candidates):
+                    chat_chunk = self._parse_candidate(candidate, idx)
+                    if chat_chunk:
+                        self._event_ch.send_nowait(chat_chunk)
+        except Exception as e:
+            logger.error("Error during main task: %s", e)
+            raise
```
theomonnom, line 200: We have an utility to directly log errors inside a task. 
Use [@utils.log_exceptions](https://github.com/livekit/agents/blob/1a99aae891d6bd440da20f88a3f6e29eab9069cc/livekit-agents/livekit/agents/utils/log.py#L7)

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+        contents.append(current_entry)
+
+    logger.debug("Built vertex context.")
+    return contents
+
+
+def _build_content(msg: llm.ChatMessage, cache_key: Any) -> Optional[dict]:
+    role = "model" if msg.role in {"system", "assistant"} else msg.role
+
+    parts: List[Dict[str, Any]] = []
```
theomonnom, line 275: Same, typing isn't very useful here

livekit-plugins/livekit-plugins-google/setup.py
```diff
@@ -19,7 +19,7 @@
 import setuptools.command.build_py
 
 here = pathlib.Path(__file__).parent.resolve()
-about = {}
+about = {}  # type: ignore
```
theomonnom, line 22: Do we need type ignore here?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+                function_call_info = create_ai_function_info(
+                    fnc_ctx=self._fnc_ctx,
+                    tool_call_id=f"{function_call.name}_{uuid()}",
+                    fnc_name=function_call.name,
+                    raw_arguments=json.dumps(function_call.args),
+                )
+                self._function_calls_info.append(function_call_info)
+            choice_delta.tool_calls = self._function_calls_info
+            return llm.ChatChunk(
+                request_id=str(uuid()),
```
theomonnom, line 218: Can you use `utils.shortuuid()`
Also it seems like we're generating a new request_id for each chunk?
It should be the same for the entire request

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+            system_instruction=system_instruction,
+            **self._kwargs,
+        )
+
+        contents = _build_vertex_context(chat_ctx, id(self))
+
+        cmp = model.generate_content_async(
+            contents=contents,
+            generation_config=generation_config,
+            tools=tools or None,
```
theomonnom, line 115: This can be done above on line 96

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+        if not self._location:
+            logger.error("Location is required for Vertex AI initialization")
+            raise ValueError
+        if not self._gac:
+            logger.error(
+                "`GOOGLE_APPLICATION_CREDENTIALS` environment variable is not set. please set it to the path of the service account key file."
+            )
+            raise ValueError
+
+        vertexai.init(project=project, location=location)
```
theomonnom, line 83: Seems like it is a singleton, what happens if the user already initialized it himself?
Is this function returning smthg?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+        generation_config = GenerationConfig(
+            temperature=temperature,
+            candidate_count=n,
+            max_output_tokens=1024,
+        )
+
+        model = GenerativeModel(
+            model_name=self._model_name,
+            system_instruction=system_instruction,
+            **self._kwargs,
```
theomonnom, line 107:
```suggestion
```

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -0,0 +1,298 @@
...
+        )
+        for fnc_info in fnc_ctx.ai_functions.values()
+    ]
+    return tools
+
+
+def _build_parameters(arguments: Dict[str, llm.FunctionArgInfo]) -> Dict[str, Any]:
+    properties = {
+        arg_name: {
+            "type": JSON_SCHEMA_TYPE_MAP.get(arg_info.type, "string"),
```
theomonnom, line 153: I don't think we should default to string?

theomonnom: Let's raise an error instead if the type isn't supported

theomonnom (changes requested):

jayeshp19:
Closing this as this feature is already implemented https://github.com/livekit/agents/pull/1084

PR #1043, opened 2024-11-05
fix: support azure speech running in containers

longcw:
Add support for azure speech containers: https://learn.microsoft.com/en-us/azure/ai-services/speech-service/speech-container-stt?tabs=container&pivots=programming-language-python#use-the-container

### DIFF
--- a/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
+++ b/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
@@ -142,7 +142,7 @@ class TTS(tts.TTS):
         speech_key = speech_key or os.environ.get("AZURE_SPEECH_KEY")
         speech_region = speech_region or os.environ.get("AZURE_SPEECH_REGION")
 
-        if not speech_host and (not speech_key or not speech_region):
+        if not speech_host and not (speech_key and speech_region):
             raise ValueError(
                 "AZURE_SPEECH_HOST or AZURE_SPEECH_KEY and AZURE_SPEECH_REGION must be set"
             )
--- a/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
+++ b/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
@@ -99,6 +99,8 @@ class ProsodyConfig:
 class _TTSOptions:
     speech_key: str | None = None
     speech_region: str | None = None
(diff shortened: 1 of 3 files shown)


livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
```diff
@@ -135,13 +138,14 @@ def __init__(
...
+        speech_host = speech_host or os.environ.get("AZURE_SPEECH_HOST")
         speech_key = speech_key or os.environ.get("AZURE_SPEECH_KEY")
-        if not speech_key:
-            raise ValueError("AZURE_SPEECH_KEY must be set")
-
         speech_region = speech_region or os.environ.get("AZURE_SPEECH_REGION")
-        if not speech_region:
-            raise ValueError("AZURE_SPEECH_REGION must be set")
+
+        if not speech_host and (not speech_key or not speech_region):
```
davidzhao, line 145: nit: slightly more readable
 
```suggestion
        if not speech_host and not (speech_key and speech_region):
```

davidzhao (approved):
lg, just one nit

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
```diff
@@ -271,11 +275,14 @@ def close(self) -> None:
...
-        speech_recognition_language=config.language or "en-US",
-    )
+    if config.speech_host:
+        speech_config = speechsdk.SpeechConfig(host=config.speech_host)
+    else:
+        speech_config = speechsdk.SpeechConfig(
+            subscription=config.speech_key,
+            region=config.speech_region,
+            speech_recognition_language=config.language or "en-US",
+        )
```
theomonnom, line 285: When there is a custom host, we don't need to configure the other fields?

theomonnom (approved):

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/tts.py
longcw: Yes, it only needs `speech_config = speechsdk.SpeechConfig(
    host="http://localhost:5000")` (ref https://learn.microsoft.com/en-us/azure/ai-services/speech-service/speech-container-ntts?tabs=container&pivots=programming-language-python#use-the-container)

PR #1046, opened 2024-11-05
Include job count in WorkerStatus and load_fnc

keepingitneil:


theomonnom:
I think we should pass the Worker instance to the load_fnc, not juste the number of jobs bc it seems very specific

keepingitneil:
ahh missed this, ok yeah that makes sense - will update

livekit-agents/livekit/agents/worker.py
```diff
@@ -683,15 +696,29 @@ def _on_process_job_launched(self, proc: ipc.job_executor.JobExecutor) -> None:
...
-            None, self._opts.load_fnc
-        )
+        def load_fnc():
+            signature = inspect.signature(self._opts.load_fnc)
+            parameters = signature.parameters
+            accepts_kwargs = any(
+                param.kind == inspect.Parameter.VAR_KEYWORD
+                for param in parameters.values()
+            )
+            if not accepts_kwargs and "job_cnt" not in parameters:
```
theomonnom, line 715: We should not get the argument by name,  I don't think the argument should be a kwarg, but the worker would be the first positional one. I don't think we will add any other variables in the future.

livekit-agents/livekit/agents/worker.py
```diff
@@ -59,6 +68,10 @@ async def _default_request_fnc(ctx: JobRequest) -> None:
     await ctx.accept()
 
 
+class LoadFunction(Protocol):
+    def __call__(self, worker: Worker) -> float: ...
+
```
theomonnom, line 73: why not `Callable[[Worker], float]`?

livekit-agents/livekit/agents/worker.py
```diff
@@ -690,15 +703,24 @@ def _on_process_job_launched(self, proc: ipc.job_executor.JobExecutor) -> None:
...
             return
 
-        current_load = await asyncio.get_event_loop().run_in_executor(
-            None, self._opts.load_fnc
-        )
+        def load_fnc():
+            signature = inspect.signature(self._opts.load_fnc)
+            parameters = list(signature.parameters.values())
+            if len(parameters) == 0:
+                return self._opts.load_fnc()  # type: ignore
```
theomonnom, line 719: not sure about the `type: ignore` here
What are they used for?

livekit-agents/livekit/agents/ipc/proc_pool.py
```diff
@@ -66,6 +66,9 @@ def get_by_job_id(self, job_id: str) -> JobExecutor | None:
             None,
         )
 
+    def get_running_jobs(self) -> list[RunningJobInfo]:
+        return [x.running_job for x in self._executors if x.running_job]
+
     def start(self) -> None:
```
theomonnom, line 72: You can use https://github.com/livekit/agents/blob/0c9eac91938d21ba4c2f2d16cfe05fb8fa237786/livekit-agents/livekit/agents/worker.py#L303 instead

livekit-agents/livekit/agents/worker.py
keepingitneil: type:ignore is needed because it's not known which load_fnc we have, one with arg or one without?

theomonnom (approved):
+ changesets otherwise lgtm

livekit-agents/livekit/agents/worker.py
keepingitneil: changed it, thx

livekit-agents/livekit/agents/ipc/proc_pool.py
```diff
@@ -66,7 +66,7 @@ def get_by_job_id(self, job_id: str) -> JobExecutor | None:
             None,
         )
 
-    def get_running_jobs(self) -> list[RunningJobInfo]:
```
theomonnom: lol what I mean is that this function is already in the worker

keepingitneil: ohhhh lol removed

theomonnom (approved):

Issue #1044, opened 2024-11-05
Feature: add after_llm_cb to modify llm generated text


achapla:
## Question regarding handling special tokens in conversation transcription
 
First of all, thanks for making this wonderful SDK to easily create voice-enabled applications!
 
I'm currently building a quiz agent that asks questions to users. The user's response is evaluated by an LLM, and if it's correct, the agent congratulates the user and adds a special 'QUESTION_END' token in the response. This token is used to identify that the conversation related to the current question is finished. I then create a new chat context with the next question.
 
### Issue
 
The issue I'm facing is that while I have removed the special 'QUESTION_END' token in the `before_tts_cb` function, so my agent does not speak 'QUESTION_END', it still appears in the conversation text. It seems that the 'QUESTION_END' text was not removed from the LLM's response before transcription.
 
Upon further investigation, I discovered the following code is executed after calling the LLM, which utilizes two different variables, `tts_source` and `transcription_source`, for different purposes:
```python
def _synthesize_agent_speech(
    self,
    speech_id: str,
    source: str | LLMStream | AsyncIterable[str],
) -> SynthesisHandle:
    assert (
        self._agent_output is not None
    ), "agent output should be initialized when ready"
 
    if isinstance(source, LLMStream):
        source = _llm_stream_to_str_iterable(speech_id, source)
 
    og_source = source
    transcript_source = source
    if isinstance(og_source, AsyncIterable):
        og_source, transcript_source = utils.aio.itertools.tee(og_source, 2)
 
    tts_source = self._opts.before_tts_cb(self, og_source)
    if tts_source is None:
        raise ValueError("before_tts_cb must return str or AsyncIterable[str]")
 
    return self._agent_output.synthesize(
        speech_id=speech_id,
        tts_source=tts_source,
        transcript_source=transcript_source,
        transcription=self._opts.transcription.agent_transcription,
        transcription_speed=self._opts.transcription.agent_transcription_speed,
        sentence_tokenizer=self._opts.transcription.sentence_tokenizer,
        word_tokenizer=self._opts.transcription.word_tokenizer,
        hyphenate_word=self._opts.transcription.hyphenate_word,
    )
```
The `tts_source` does not contain the 'QUESTION_END' token, so it won't be played out, but the `transcription_source` contains 'QUESTION_END', causing it to be included in the transcription when committed.
 
### Request for Enhancement
 
[...]

PR #1052, opened 2024-11-06
Add AfterLLMCallback to modify LLM stream text post-generation

achapla:
Introduced an after_llm_cb function to enable developers to modify the LLM-generated text before it is committed to transcription and passed to the before_tts_cb function. This function provides flexibility to adjust or remove specific tokens or phrases from the transcription source, ensuring a cleaner conversation flow without requiring frontend workarounds. The modified text now flows through to before_tts_cb, allowing consistent text handling across the conversation stages.

https://github.com/livekit/agents/issues/1044

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -40,6 +40,20 @@ BeforeLLMCallback = Callable[
 
 WillSynthesizeAssistantReply = BeforeLLMCallback
 
+AfterLLMCallback = Callable[
+    ["VoicePipelineAgent", 
+    Union[
+        AsyncGenerator[str, None],
+        str,
+        AsyncIterable[str]]
+    ], 
+    Union[
+        AsyncGenerator[str, None],
+        str,
+        AsyncIterable[str]
+    ],
+]
+
 BeforeTTSCallback = Callable[
     ["VoicePipelineAgent", Union[str, AsyncIterable[str]]],
     SpeechSource,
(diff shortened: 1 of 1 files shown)


achapla:
@davidzhao I don't have experience with Python, so please review the parameter types for `after_llm_cb`.

Note: Modifying the text in the `after_llm_cb` callback function will pass the modified text to `before_tts_cb`.

theomonnom:
Hey, this makes sense, I'll take a look at this PR asap. I think we should expose the [before_forward_cb](https://github.com/livekit/agents/blob/1a99aae891d6bd440da20f88a3f6e29eab9069cc/livekit-agents/livekit/agents/transcription/tts_forwarder.py#L86) of the TTSForwarder somehow inside the VoicePipelineAgent instead

davidzhao:
> Hey, this makes sense, I think we should expose the [before_forward_cb](https://github.com/livekit/agents/blob/1a99aae891d6bd440da20f88a3f6e29eab9069cc/livekit-agents/livekit/agents/transcription/tts_forwarder.py#L86) argument of the TTSForwarder inside the VoicePipelineAgent instead

I think what's committed and what's forwarded should match.. the TTSForwarder would likely change with the new TextStream APIs as well?

in that sense.. perhaps `after_llm_cb` would make more sense? naming is always hard..

achapla:
@theomonnom @davidzhao Thanks for the feedback! You have a solid understanding of the inner workings of the framework.
Personally, I prefer adding `after_llm_cb` as it is clearer and more intuitive than `before_forward_cb`.
I can also work on making the necessary changes to expose the `before_forward_cb` callback. 

This feature would undoubtedly benefit many developers.

PR #1074, opened 2024-11-12
feat: tts retry & tts.FallbackAdapter

theomonnom:
requires https://github.com/livekit/rust-sdks/pull/486

This also adds:
- NotGiven type (currently only used in Mocks)
- sample_rate options to Azure STT


livekit-agents/livekit/agents/tts/fallback_adapter.py
```diff
@@ -0,0 +1,507 @@
...
+    available: bool
+
+
+class FallbackAdapter(
+    TTS[Literal["tts_availability_changed"]],
+):
+    def __init__(
+        self,
+        tts: list[TTS],
+        *,
```
nbsp, line 53: a syntactic sugar idea i had when i wrote this a few months back:

```suggestion
        *tts: list[TTS],
```

that way you can write `FallbackAdapter(tts1, tts2, tts3, timeout=5.0)`

livekit-agents/livekit/agents/tts/fallback_adapter.py
```diff
@@ -0,0 +1,507 @@
...
+        self._no_fallback_after_audio_duration = no_fallback_after_audio_duration
+
+        self._status: list[_TTSStatus] = []
+        for t in tts:
+            resampler = None
+            if sample_rate != t.sample_rate:
+                logger.info(
+                    f"resampling {t.label} from {t.sample_rate}Hz to {sample_rate}Hz"
+                )
+                resampler = rtc.AudioResampler(
```
nbsp, line 88: 🔥, was missing back then, glad it exists now

livekit-agents/livekit/agents/tts/fallback_adapter.py
theomonnom: it looks better but this issue with this synthetic sugar, is that the parameter is optional

livekit-agents/livekit/agents/tts/fallback_adapter.py
```diff
@@ -0,0 +1,507 @@
...
+            audio_duration = 0.0
+            async with tts.synthesize(
+                self._input_text, conn_options=req_conn_options
+            ) as stream:
+                while True:
+                    try:
+                        audio = await asyncio.wait_for(
+                            stream.__anext__(),
+                            self._tts._timeout if audio_duration == 0.0 else None,
+                        )
```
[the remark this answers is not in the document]
theomonnom: The TTS itself will handle the timeout, the fallback adapter has the choice to fallback early

theomonnom:
I'm going to simplify the "recovery request" with an hardcoded string like. "Hello world, this is a random text".
So I don't have to keep track of the user input. I guess it is fine, wdyt @livekit/agent-devs

davidzhao:
> I'm going to simplify the "recovery request" with an hardcoded string like. "Hello world, this is a random text".

Can you elaborate when this is used? I'm not a fan of hardcoded, language-specific strings in general.

theomonnom:
> > I'm going to simplify the "recovery request" with an hardcoded string like. "Hello world, this is a random text".
> 
> Can you elaborate when this is used? I'm not a fan of hardcoded, language-specific strings in general.

Please ignore, I didn't think about languages, true that hardcoded strings isn't good in that case.

davidzhao (approved):

PR #1075, opened 2024-11-13
Clean up Cartesia audio frame send

martin-purplefish:
We're running into an issue where our audio gets glitchy every so often (I sent a sample on Slack) - I think it's because we're either reordering or dropping frames.

### DIFF
--- a/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
+++ b/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
@@ -317,13 +317,9 @@ class SynthesizeStream(tts.SynthesizeStream):
                     for frame in audio_bstream.write(b64data):
                         _send_audio_frame(frame, segment_id, False)
                 elif data.get("done"):
-                    last_frame: rtc.AudioFrame | None = None
-                    for frame in audio_bstream.flush():
-                        if last_frame is not None:
-                            _send_audio_frame(frame, segment_id, False)
-                        last_frame = frame
-                    if last_frame is not None:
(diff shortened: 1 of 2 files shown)


martin-purplefish:
PTAL @davidzhao @theomonnom - not sure why Cartesia wasn't sending the first audio frame, but it seems to work. I'm not sure if this fixes the glitches but it seems less wrong.

martin-purplefish:
actually I guess this was sending everything just in a confusing way.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -319,14 +315,11 @@ def _send_last_frame(*, segment_id: str, is_final: bool) -> None:
...
+                        _send_audio_frame(frame, segment_id, False)
                 elif data.get("done"):
-                    for frame in audio_bstream.flush():
-                        _send_last_frame(segment_id=segment_id, is_final=False)
-                        last_frame = frame
-
-                    _send_last_frame(segment_id=segment_id, is_final=True)
+                    frames = list(audio_bstream.flush())
+                    for i, frame in enumerate(frames):
+                        _send_audio_frame(frame, segment_id, i == len(frames) - 1)
```
theomonnom, line 322: This doesn't guarantee that at least one frame with is_final=True is sent.
Because `audio_bstream.flush()` could return an empty list in some scenarios

martin-purplefish: ah you're right

PR #1078, opened 2024-11-13
pipelineagent: fix speech_committed never called 

### LINKED ISSUE #1069: When tool call includes content, it is not committed or used in the follow up response
OpenAI can return content along side a tool call:

```
    {
      "role": "assistant",
      "content": [
        {
          "text": "Oh, just a moment while I check that for you... \n\nLet me find out the current weather in NYC by querying our weather service.",
          "type": "text"
        }
      ],
      "tool_calls": [
        {
          "id": "call_Gwqkvbb3r0oRJnj3zzrIEL1R",
          "type": "function",
          "function": {
            "name": "get_weather",
            "arguments": "{\"location\":\"New York, NY\",\"unit\":\"f\"}"
          }
        }
      ]
    }
```

When this happens, there are a few problems:

1. The agent's audio plays, BUT the audio is not committed to the log.
2. There's no `agent_speech_committed` event.
3. (I think) the audio gets played AFTER the function call, rather than before it.

theomonnom:
fix #1069

### DIFF
--- a/.changeset/breezy-rockets-camp.md
+++ b/.changeset/breezy-rockets-camp.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+pipelineagent: fix speech_committed never called 
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -742,6 +742,13 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
 
             llm_stream = speech_handle.source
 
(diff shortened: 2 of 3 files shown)


theomonnom:
PTAL @martin-purplefish

martin-purplefish:
Oh sweet! Looks great

davidzhao (approved):
lgtm

martin-purplefish:
@davidzhao Hold up on merging though - we're trying to figure out if we are duplicating something here, with the tool messages.

theomonnom:
I think this is the only scenario where there is some "playable text" alongside fnc calls

PR #1082, opened 2024-11-13
Updated assembly.ai integration

davidzhao:
Leveraging common utils including streaming resampler

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -16,16 +16,14 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
 import dataclasses
 import json
 import os
 from dataclasses import dataclass
-from typing import List, Literal, Optional, Union
+from typing import List, Literal, Optional
 from urllib.parse import urlencode
 
 import aiohttp
-from livekit import rtc
 from livekit.agents import stt, utils
 from livekit.agents.utils import AudioBuffer
 
@@ -160,7 +158,6 @@ class SpeechStream(stt.SpeechStream):
         # keep a list of final transcripts to combine them inside the END_OF_SPEECH event
         self._final_events: List[stt.SpeechEvent] = []
 
-
     async def aclose(self) -> None:
         await super().aclose()
         if self._session:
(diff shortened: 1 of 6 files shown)


theomonnom (approved):

theomonnom:
Can you make the http_session optional;

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -66,9 +64,9 @@ def __init__(
         encoding: Optional[str] = "pcm_s16le",
         disable_partial_transcripts: bool = False,
         enable_extra_session_information: bool = False,
-        end_utterance_silence_threshold: int = 1000,
+        end_utterance_silence_threshold: Optional[int] = 100,
         http_session: Optional[aiohttp.ClientSession] = None,
-        buffer_size_seconds: float = 0.2,
+        buffer_size_seconds: float = 0.05,
```
theomonnom, line 69: Not sure if this option is really relevant for the users. Let's pick the same one as DG?

davidzhao: i didn't want to introduce additional latency, it seems that 50ms chunks is a good number for everything, wdyt?

theomonnom: yeah 50ms seems good

theomonnom: what I mean is that it is fine to hardcode it, but no big deal to leave it there

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -360,17 +320,17 @@ def _end_speech(self) -> None:
             ],
         )
         self._final_events = []
-        self._event_queue.put_nowait(end_event)
+        self._event_ch.send_nowait(end_event)
         # break async iteration if speech has ended
-        self._event_queue.put_nowait(None)
+        self._event_ch.send_nowait(None)
```
theomonnom, line 325: We should never send None

theomonnom: if the intent is to close, just do nothing, the STT baseclass will close the event_ch when the main task closes

PR #1084, opened 2024-11-14
Vertex with openai

jayeshp19:


livekit-plugins/livekit-plugins-openai/setup.py
```diff
@@ -47,7 +47,11 @@
     license="Apache-2.0",
     packages=setuptools.find_namespace_packages(include=["livekit.*"]),
     python_requires=">=3.9.0",
-    install_requires=["livekit-agents[codecs, images]>=0.11", "openai>=1.50"],
+    install_requires=[
+        "livekit-agents[codecs, images]>=0.11",
+        "openai>=1.50",
+        "google-auth",
```
theomonnom, line 53: we should put google-auth on an extra_requires.

theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -161,6 +162,54 @@ def with_cerebras(
...
+        from google.auth import default
+        from google.auth.transport import requests
+
+        credentials, _ = default(
+            scopes=["https://www.googleapis.com/auth/cloud-platform"]
+        )
+        auth_request = requests.Request()
+        credentials.refresh(auth_request)
+        base_url = f"https://{location}-aiplatform.googleapis.com/v1beta1/projects/{project_id}/locations/{location}/endpoints/openapi"
+        api_key = credentials.token
```
theomonnom, line 202: What is the TTL of this token?

theomonnom: We should automatically refresh it, otherwise the LLM instance could start to fail after this TTL

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -492,6 +572,7 @@ def chat(
             temperature = self._opts.temperature
 
         messages = _build_oai_context(chat_ctx, id(self))
+        logger.info(f"messages: {messages}")
```
theomonnom, line 575: Remove this line

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -161,6 +162,85 @@ def with_cerebras(
...
+            def __init__(self, **kwargs: Any) -> None:
+                self.client = openai.AsyncClient(**kwargs, api_key="DUMMY")
+                self.creds, self.project = google.auth.default(
+                    scopes=["https://www.googleapis.com/auth/cloud-platform"]
+                )
+
+            def __getattr__(self, name: str) -> Any:
+                if not self.creds.valid:
+                    auth_req = google.auth.transport.requests.Request()
+                    self.creds.refresh(auth_req)
```
theomonnom, line 216: Do you know if refresh is doing a network call?

jayeshp19: it probably should but I will check their implementation

theomonnom: The issue if it is doing one is they it'll block the main thread since the request is sync

examples/voice-pipeline-agent/function_calling_weather.py
```diff
@@ -67,6 +67,8 @@ async def entrypoint(ctx: JobContext):
         vad=ctx.proc.userdata["vad"],
         stt=deepgram.STT(),
         llm=openai.LLM(),
+        # To use vertex AI LLM
+        # llm=openai.LLM.with_vertexai(project_id="your-project-id", location="us-central1"),
```
nbsp, line 71: is there a specific reason vertex would be used here? feels odd adding this example when we have a good few other adapters

nbsp (approved):

examples/voice-pipeline-agent/function_calling_weather.py
jayeshp19: I just added for example, I should probably add another file specific to vertex with google stt and tts

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -161,6 +163,113 @@ def with_cerebras(
...
+                    self.stop_event.wait(timeout=self.refresh_interval)
+                    if self.stop_event.is_set():
+                        break
+                    self._refresh_credentials()
+
+            def _refresh_credentials(self) -> None:
+                with self.refresh_lock:
+                    try:
+                        auth_req = google.auth.transport.requests.Request()
+                        self.creds.refresh(auth_req)
```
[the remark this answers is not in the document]
jayeshp19: then we would also need to make chat request async??

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
davidzhao: google seems to support asyncio in their library: https://github.com/googleapis/google-auth-library-python/issues/1417

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -161,6 +164,95 @@ def with_cerebras(
             temperature=temperature,
         )
 
+    @staticmethod
+    def with_vertexai(
+        *,
+        model: str | VertexModels = "google/gemini-1.5-pro",
+        project_id: str | None = None,
+        location: str | None = "us-central1",
+        client: Any | None = None,
```
theomonnom, line 173:
```suggestion
```

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -161,6 +164,95 @@ def with_cerebras(
...
+
+            async def _refresh_credentials(self) -> None:
+                if self.creds and self.creds.valid and not self._token_needs_refresh():
+                    return
+                async with aiohttp.ClientSession(auto_decompress=False) as session:
+                    auth_req = Request(session=session)
+                    await self.creds.refresh(auth_req)
+
+                if not self.creds.valid:
+                    raise RuntimeError("Unable to refresh auth")
```
theomonnom, line 233:
```suggestion
```

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -161,6 +164,95 @@ def with_cerebras(
             temperature=temperature,
         )
 
+    @staticmethod
+    def with_vertexai(
```
theomonnom, line 168: nit: We took a decision a while ago about not adding ai at the end of every brand (only for those who aren't obvious)
```suggestion
    def with_vertex(
```

livekit-plugins/livekit-plugins-openai/setup.py
```diff
@@ -47,7 +47,13 @@
     license="Apache-2.0",
     packages=setuptools.find_namespace_packages(include=["livekit.*"]),
     python_requires=">=3.9.0",
-    install_requires=["livekit-agents[codecs, images]>=0.11", "openai>=1.50"],
+    install_requires=[
+        "livekit-agents[codecs, images]>=0.11",
+        "openai>=1.50",
+    ],
+    extras_require={
+        "vertex": ["google-auth"],
```
theomonnom, line 55: Can you specify a minimum version

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -524,8 +616,11 @@ def __init__(
         self._tool_call_id: str | None = None
         self._fnc_name: str | None = None
         self._fnc_raw_arguments: str | None = None
+        self._tool_index: int | None = None
 
     async def _main_task(self) -> None:
+        if hasattr(self._llm._client, "_refresh_credentials"):
```
theomonnom, line 622: Ideally we would have used a subclass to add this new function. But it is fine to merge so feel free to ignore

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/_oai_api.py
```diff
@@ -115,6 +116,16 @@ def type2str(t: type) -> str:
             if arg_info.choices:
                 p["enum"] = arg_info.choices
 
+            if arg_info.type is int and arg_info.choices and model is not None:
```
theomonnom, line 119: model is not None doesn't make a ton of sense here? U could just check if it is Vertex or not

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/_oai_api.py
```diff
@@ -115,6 +116,16 @@ def type2str(t: type) -> str:
...
+                from .models import VertexModels
+
+                vertex_models_set = set(get_args(VertexModels))
+
+                if model in vertex_models_set:
+                    raise ValueError(
+                        f"Parameter '{arg_info.name}' uses 'choices' with 'int', which is not supported by Vertex AI."
+                    )
+
```
theomonnom, line 128: One issue I see with this approach is that we have to keep a up to date list of the supported models Google has. Maybe we can simply create a dataclass with supported stuff?

like:

@dataclass
class SupportedFunctionArguments:
 ....

theomonnom:
Idk if this is the right approach: but our anthropic inject a user message if the chatcontext doesn't start with one
https://github.com/livekit/agents/blob/a5ca1a58041dff7281b811efc72fc7c6cb0b6b05/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py#L300

theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/_oai_api.py
jayeshp19: That's a good idea, I'll update pr

jayeshp19: @theomonnom  I've updated code with LLMCapabilities class

jayeshp19:
> Idk if this is the right approach: but our anthropic inject a user message if the chatcontext doesn't start with one
> 
> https://github.com/livekit/agents/blob/a5ca1a58041dff7281b811efc72fc7c6cb0b6b05/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py#L300

Are you saying we should add for openai aswell?

theomonnom (approved):

PR #1085, opened 2024-11-14
fix: conversation item create and delete return a Future

longcw:


theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -541,7 +544,8 @@ def create(
                         message,
                         extra=self._sess.logging_extra(),
                     )
-                    return
+                    fut.set_result(False)
+                    return fut
```
theomonnom, line 548: I know it's been like that before, but just thinking out loud, should we raise an Exception instead of logging a warning? Since the creation failed anyway

longcw: What do you think if we let the user of this method to decide the agent to keep running or be failed based on the value of the Future.

theomonnom: I think Python users don't tend to check the output of functions for errors. Most of the time they except seeing Exception and catching them

longcw: Ok that makes sense, I can submit another PR for this.

PR #1089, opened 2024-11-15
enh: Extending api to support chirp_2 model, and support passing location

brightsparc:
Add support for `chirp_2` [transcription model](https://cloud.google.com/speech-to-text/v2/docs/transcription-model), that is supported in both batch and streaming transcription see [release notes](https://cloud.google.com/speech-to-text/docs/release-notes#October_07_2024)

Update tests to include both default SST, and SST with new model passing specific [location and language](https://cloud.google.com/speech-to-text/v2/docs/speech-to-text-supported-languages).  Note that spoken punctuation is not supported with this model.

### DIFF
--- a/.changeset/chatty-grapes-scream.md
+++ b/.changeset/chatty-grapes-scream.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-google": minor
+---
+
+Add support for google STT chirp_2 model.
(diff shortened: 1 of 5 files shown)


theomonnom (approved):
Thanks!

theomonnom:
Can you add a changesets by clicking on the above comment 🙏

PR #1090, opened 2024-11-15
fix Deepgram missing first word, disabled energy filter by default

davidzhao:
- energy filter should run on smaller chunks of audio
- it's disabled by default, as it's an optimization
- buffers audio frame immediately prior to energy detection to improve accuracy

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -279,13 +279,12 @@ class SpeechStream(stt.SpeechStream):
         self._speaking = False
         self._max_retry = max_retry
 
+        self._audio_energy_filter: Optional[AudioEnergyFilter] = None
         if opts.energy_filter:
             if isinstance(opts.energy_filter, AudioEnergyFilter):
                 self._audio_energy_filter = opts.energy_filter
             else:
                 self._audio_energy_filter = AudioEnergyFilter()
-        else:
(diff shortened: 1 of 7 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -361,32 +376,52 @@ async def send_task():
...
-                            recognition_usage=stt.RecognitionUsage(
-                                audio_duration=self._pushed_audio_duration
-                            ),
-                        )
-                        self._pushed_audio_duration = 0.0
-                        self._event_ch.send_nowait(usage_event)
+                    self._pushed_audio_duration += frame.duration
+                    await ws.send_bytes(frame.data.tobytes())
+
+                    if has_ended:
```
theomonnom, line 411: IIUC, has_ended is only true when Flushing when the energy filter is disabled, so the metrics will never be emitted on the VoicePipelineAgent because it is only flushed at shutdown.

I think we should still emit metrics even when energy_filter is False (by always checking for the SpeechEnded from the EnergyFilter avg)

davidzhao: it's also set when flush sentinel is seen.

davidzhao: I see what you mean, let me see if I can tweak this quickly

theomonnom (approved):

Issue #1095, opened 2024-11-16
Error in Google STT doesn't reconnect


martin-purplefish:
```
{"message": "Error in _recognize_task\nTraceback (most recent call last):\n File \"/app/.venv/lib/python3.12/site-packages/livekit/agents/utils/log.py\", line 16, in async_fn_logs\n return await fn(*args, **kwargs)\n ^^^^^^^^^^^^^^^^^^^^^^^^^\n File \"/app/.venv/lib/python3.12/site-packages/livekit/agents/pipeline/human_input.py\", line 150, in _recognize_task\n await asyncio.gather(*tasks)\n File \"/app/.venv/lib/python3.12/site-packages/livekit/agents/pipeline/human_input.py\", line 120, in _audio_stream_co\n stt_stream.push_frame(ev.frame)\n File \"/app/.venv/lib/python3.12/site-packages/livekit/agents/stt/stt.py\", line 167, in push_frame\n self._check_not_closed()\n File \"/app/.venv/lib/python3.12/site-packages/livekit/agents/stt/stt.py\", line 229, in _check_not_closed\n raise RuntimeError(f\"{cls.__module__}.{cls.__name__} is closed\")\nRuntimeError: livekit.plugins.google.stt.SpeechStream is closed", "level": "ERROR", "name": "livekit.agents.pipeline", "pid": 20222, "job_id": "AJ_ATBa5D7tNRDP", "timestamp": "2024-11-16T15:34:04.426177+00:00"}
```

At this point, the agent couldn't hear the other person and we got no more messages.

davidzhao:
which version are you seeing this?

does it only happen with Google? how could one reproduce this?

martin-purplefish:
We're using 0.7.3 - it just happened, I think a random service error? the effect was the agent stopped getting text.

davidzhao:
were there other errors before this? for example:

```
google stt connection aborted
```

martin-purplefish:
Oh yes, we do get the "google stt connection aborted" right before. I don't see any other errors in the logs though.

davidzhao:
yeah, ok, I think I know what's going on

Issue #852, opened 2024-10-07
google STT connect doesn't gracefully shutdown when agent disconnected


brightsparc:
When using the google STT, unlike deepgram STT the run loop doesn't gracefully terminate when agent disconnected:

### Error

```
2024-10-07 10:55:10,339 - WARNING livekit.plugins.google - google stt connection failed, retrying in 0s
Traceback (most recent call last):
  File "/Users/julian/miniforge3/envs/livekit-py311/lib/python3.11/site-packages/google/api_core/grpc_helpers_async.py", line 106, in _wrapped_aiter
    async for response in self._call:  # pragma: no branch
  File "/Users/julian/miniforge3/envs/livekit-py311/lib/python3.11/site-packages/grpc/aio/_call.py", line 365, in _fetch_stream_responses
    await self._raise_for_status()
  File "/Users/julian/miniforge3/envs/livekit-py311/lib/python3.11/site-packages/grpc/aio/_call.py", line 272, in _raise_for_status
    raise _create_rpc_error(
grpc.aio._call.AioRpcError: <AioRpcError of RPC that terminated with:
        status = StatusCode.ABORTED
        details = "Stream timed out after receiving no more client requests."
        debug_error_string = "UNKNOWN:Error received from peer ipv6:%5B2404:6800:4006:809::200a%5D:443 {grpc_message:"Stream timed out after receiving no more client requests.", grpc_status:10, created_time:"2024-10-07T10:55:10.333325+11:00"}"
>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/julian/miniforge3/envs/livekit-py311/lib/python3.11/site-packages/livekit/plugins/google/stt.py", line 259, in _run
    await self._run_stream(stream)
  File "/Users/julian/miniforge3/envs/livekit-py311/lib/python3.11/site-packages/livekit/plugins/google/stt.py", line 279, in _run_stream
    async for resp in stream:
  File "/Users/julian/miniforge3/envs/livekit-py311/lib/python3.11/site-packages/google/api_core/grpc_helpers_async.py", line 109, in _wrapped_aiter
    raise exceptions.from_grpc_error(rpc_error) from rpc_error
google.api_core.exceptions.Aborted: 409 Stream timed out after receiving no more client requests. [type_url: "type.googleapis.com/util.StatusProto"
[...]

PR #1100, opened 2024-11-17
Fix: add retry logic for google stt abort

jayeshp19:
https://github.com/livekit/agents/issues/1095


.changeset/moody-snails-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+"livekit-plugins-google": patch
+---
+
+fix:add retry logic for google stt abort exception
```
davidzhao, line 5:
```suggestion
fix: add retry logic for google stt abort exception
```

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -303,10 +303,12 @@ async def input_generator():
                 retry_count = 0  # connection successful, reset retry count
 
                 await self._run_stream(stream)
-            except Aborted:
-                logger.error("google stt connection aborted")
-                break
-            except Exception as e:
+            except (Aborted, Exception) as e:
```
davidzhao, line 306: I think we need to not retry or log errors if the input queue is closed.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -303,10 +303,12 @@ async def input_generator():
...
+                logger.error(
```
theomonnom, line 307: We're already logging bellow. Is the Aborted error only fired when the maximum stream duration is reached? If so, we shouldn't log the error to make it more seamless to the user

jayeshp19: I think that's a diffrent issue, @davidzhao answered here https://github.com/livekit/agents/issues/852#issuecomment-2463792401
Aborted will fired when someone closes the agent session or some other concurrency issues 
image
https://cloud.google.com/speech-to-text/v2/docs/reference/rest/v2/Code

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
jayeshp19: This will not be fired because `self._input_ch.closed` will stay false until agent session closes

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -303,22 +303,19 @@ async def input_generator():
...
                     )
                     break
 
                 retry_delay = min(retry_count * 2, 5)  # max 5s
                 retry_count += 1
                 logger.warning(
-                    f"google stt connection failed, retrying in {retry_delay}s",
-                    exc_info=e,
+                    f"google stt connection {error_type.lower()}, retrying in {retry_delay}s",
                 )
```
theomonnom, line 319: exc_info is an important parameter to keep

theomonnom (approved):

theomonnom (approved):

PR #1102, opened 2024-11-17
feat: support for custom toolchoice in LLMs

jayeshp19:
### What
This PR implements custom tool choice and parallel tool call functionality for LLMs.

### Why
Feature Improvement

### Example Usage
```python
agent = VoicePipelineAgent(
    vad=ctx.proc.userdata["vad"],
    stt=deepgram.STT(),
    llm=openai.LLM(
        parallel_tool_calls=True, 
        tool_choice="auto",
    ),
    tts=openai.TTS(),
    fnc_ctx=fnc_ctx,
    chat_ctx=initial_chat_ctx,
)
```

### Relevant Links
- https://docs.anthropic.com/en/docs/build-with-claude/tool-use#controlling-claudes-output
- https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice


theomonnom:
I think we should introduce `tool_choice` inside the chat method in the baseclass for all LLMs

jayeshp19:
@theomonnom Shouldn't these parameters be in the `__init__` method of baseclass? End users should be able to configure them, right?

livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
```diff
@@ -40,11 +50,18 @@
 )
 
 
+class ToolChoice(TypedDict, total=False):
+    type: Literal["auto", "any", "tool", "none", "required"]
```
theomonnom, line 54: what is the difference between `tool` and `required`?

What is `any` used for?

jayeshp19: `required` (OpenAI specific param) - the model must call one or more tools. 
`any` (Anthropic specific param) - the model must call one or more tools. 
`tool` (Anthropic specific param) - the model must call specific tool

theomonnom: can we merge `any` and `required`?

jayeshp19: Yes, we can, but I think we may need to document this otherwise, it might confuse users.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -578,6 +680,14 @@ def chat(
 
             if fnc_ctx and parallel_tool_calls is not None:
                 opts["parallel_tool_calls"] = parallel_tool_calls
+            if tool_choice is not None:
+                if isinstance(tool_choice, dict):
```
theomonnom, line 684:
```suggestion
                if isinstance(tool_choice, FunctionToolChoice):
```

theomonnom (commented):
Can you also update this class: https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-llama-index/livekit/plugins/llama_index/llm.py

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -51,6 +60,11 @@ class ChatChunk:
     usage: CompletionUsage | None = None
 
 
+class FunctionToolChoice(TypedDict):
```
theomonnom, line 63:
```suggestion
class ToolChoice(TypedDict):
```

theomonnom (approved):

.changeset/real-phones-cheat.md
```diff
@@ -0,0 +1,7 @@
+---
+"livekit-plugins-anthropic": patch
+"livekit-plugins-openai": patch
+"livekit-agents": patch
```
theomonnom, line 4: add llama-index here

PR #1106, opened 2024-11-18
google STT - use the baseclass resampler

jayeshp19:


theomonnom (approved):
Nice, awesome, in addition, can you default the Google STT to a lower sample rate like ~~24000~~ 16Khz?

theomonnom:
From https://cloud.google.com/speech-to-text/docs/speech-to-text-requests#sample-rates:

> If you have a choice when encoding the source material, capture audio using a sample rate of 16000 Hz. Values lower than this may impair speech recognition accuracy, and higher levels have no appreciable effect on speech recognition quality.
> 
> However, if your audio data has already been recorded at an existing sample rate other than 16000 Hz, do not resample your audio to 16000 Hz. Most legacy telephony audio, for example, use sample rates of 8000 Hz, which may give less accurate results. If you must use such audio, provide the audio to the Speech API at its native sample rate.


16Khz may be the right choice

theomonnom (approved):

PR #1108, opened 2024-11-19
Add mulaw encoding option to ElevenLabs plugin

chrisackermann:
Add support for ElevenLabs' [mulaw encoding format ](https://elevenlabs.io/docs/api-reference/websockets#query-parameters)(`ulaw_8000`) since services such as Twilio recommend using this encoding for outbound calls.


chrisackermann:
Closing since this is by design

sbichenko:
@chrisackermann Can you share why this is by design? This seems like a significant downside to choosing LiveKit for voice orchestration, compared to other solutions for telephony. (But maybe I'm missing something)

chrisackermann:
@sbichenko Sorry for the confusion around this!  Our agents framework supports PCM and the conversion from PCM to ulaw happens on the SIP server automatically.

Issue #1054, opened 2024-11-06
Allow queuing custom responses from function call


davidzhao:
During a function call, sometimes it's desirable to let the user know that this is going to take awhile. While it is possible to prompt the LLM to return a text response while this is happening, it would offer more control to be able to do `agent.say("....")`.

Currently when that is done, the speech gets queued *after* the LLM's response. We'd want the ability to inject a response as the immediate thing that's communicated.

Don-Chad:
It would be huge if we could even pre-buffer the audio for this message...  Thanks guys!

theomonnom:
> It would be huge if we could even pre-buffer the audio for this message...  Thanks guys! 

Hey, what do you mean by pre-buffer, do you want to synthesize the agent speech ahead of time?

Don-Chad:
@theomonnom hi Theo, yes exactly. Then we could get to very nice and low latencies with an initial prebuffered message, as we are streaming the full answer.

Hume for example is also offering this feature.

longcw:
@davidzhao Where do you think the "quick response" before the function call done should come from:
prompt the LLM to say something after function_calls_collected or allow the function to early return a customized text before it done?

davidzhao:
some folks are already prompting LLM to return the right text. I think others are choosing to use `agent.say` to queue up a custom placeholder response.

currently when trying the latter path, the speech doesn't get enqueued right away

longcw:
> some folks are already prompting LLM to return the right text. I think others are choosing to use `agent.say` to queue up a custom placeholder response.
> 
> currently when trying the latter path, the speech doesn't get enqueued right away

I see. I can take a look if it's an issue of `agent.say` doesn't work as expected.

longcw:
The current function call logic is like the following pseudo-code
```python
fncs = speech_handle.synthesis_handle.play().join()
chat_ctx = speech_handle.source.chat_ctx
nest = 0
while fncs and nest < max_nested:
  tools_messages = [fnc.exectue() for fnc in fncs]

  answer_llm_stream = llm.chat(chat_ctx.copy() + tools_messages)
  speech_handle.synthesis_handle = _synthesize_agent_speech(answer_llm_stream)

  fncs = speech_handle.synthesis_handle.play().join()
  nest += 1
```
[...]

PR #1112, opened 2024-11-20
Fix/async function calls

longcw:
https://github.com/livekit/agents/issues/1054


livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -660,6 +754,19 @@ async def _synthesize_answer_task(
             SpeechDataContextVar.reset(tk)
 
     async def _play_speech(self, speech_handle: SpeechHandle) -> None:
+        if speech_handle.extra_tools_messages:
+            # synthesize the tool speech with the latest chat context and the tools messages
```
longcw, line 758: I moved the llm call for the tool outputs here to be able to use the latest chat context. It's intended to avoid repeating the same sentence from `say` before the function call completed. Let me know if this makes sense.
```
user: What is the weather in New York?
agent (say): Let me check that for you.
(function call...)
agent: The weather in New York is ...
```

martin-purplefish: Could you optimize and check if the speech is interruptable? The way I'm thinking of this - it'll just be a short interjection ("Thanks for sharing!") so I'm fine making that uninterruptable.

longcw: Since we are now running the function call asynchronously, there are three options when human input occurs while the function call is still running:
1. ignore the human input and while the function calls finishing (the existing way)
2. interrupt the function calls
3. answer the human input meanwhile keep function calls and insert the function call answer later

I have discussed with @theomonnom that I'll make all function calls uninterruptable to have the same behavior with the existing version (option 1). And maybe in the future we can make this configurable and the developer can choose which behavior to have.

(Note that the above is only for human input, the speech from `say` will always be send out when the function call is running.)

livekit-agents/livekit/agents/llm/function_context.py
```diff
@@ -100,6 +102,52 @@ class CalledFunction:
...
+    """Represents a set of function calls to be executed, along with their context"""
+
+    function_calls: Sequence[FunctionCallInfo]
+    llm_stream: LLMStream
+    nest_depth: int
+    speech_id: str
+    _interrupted: bool = False
+
+    async def execute(self) -> list[CalledFunction]:
+        called_fncs = []
```
theomonnom, line 116: I don't think this should be part of the llm module. It is very specific to the PipelineAgent

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -276,6 +284,9 @@ def __init__(
         self._speech_q: list[SpeechHandle] = []
         self._speech_q_changed = asyncio.Event()
 
+        self._fnc_q: list[FunctionCallsHandle] = []
+        self._fnc_q_changed = asyncio.Event()
+
         self._update_state_task: asyncio.Task | None = None
```
theomonnom, line 291: I'm not sure if we should use a different queue.
I was more thinking about nesting all speeches together. 

Imagine this scenario:
1. User speaks and request a fnc call
2. The agent generates an answer containing this specific call
3. The fnc call contains multiple calls to the say method.
4. If the user speaks it may interrupt this specific "say" speech but not the "parent" 

With the current approach, there is also no straightforward way to map the function calls to a speech/answer?

theomonnom: The advantages with nested calls is that we can make it recursive:

e.g:
Function call is called but the say method takes an `LLMStream` in input that may call other functions.
In this case we have to wait for this whole recursive process before ending the root speech

longcw: I got two questions:

How do you think the nested_speeches should work, like a speech triggers a function call, then we add the speeches from this function call (including `say` or any nested function calls) to the `nested_speeches`?

Where should these nested_speeches be handled, a separate task inside the `_play_speech` or reusing the `_play_speech` somehow?

longcw: > Imagine this scenario:
> User speaks and request a fnc call
The agent generates an answer containing this specific call
The fnc call contains multiple calls to the say method.
If the user speaks it may interrupt this specific "say" speech but not the "parent"

From here https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/pipeline/pipeline_agent.py#L820 the `interrupted` flag is not used to exit the nested function call loop, so in the existing version the nested function calls are already not interruptable from my understanding. Is that correct?

longcw: > The advantages with nested calls is that we can make it recursive

I agree with this one. If I understand correctly, what you mean is that the function calls triggered by `say` should be handled before the parent function calls while the current implementation cannot do that. I think this makes sense.

theomonnom: > How do you think the nested_speeches should work, like a speech triggers a function call, then we add the speeches from this function call (including `say` or any nested function calls) to the `nested_speeches`?

Yep exactly

> Where should these nested_speeches be handled, a separate task inside the `_play_speech` or reusing the `_play_speech` somehow?

Reusing the _play_speech would be the best, so we have the same code.



I'm imagining this scenario for interruption involving nested speeches.


```
    @llm.ai_callable()
    async def get_weather(
        self,
        location: Annotated[
            str, llm.TypeInfo(description="The location to get the weather for")
        ],
    ):
          speech_handle = await agent.say("This is a random speech")
          if speech_handle.interrupted: # user can make this decision for e.g
               return
```
Don't implement this feature on this PR since speech handles are currently private but this is just to give you an order of idea that interruptions should be "per" nested speeches.

longcw: @theomonnom I created a new PR based on your suggestion: https://github.com/livekit/agents/pull/1130

PR #1130, opened 2024-11-23
feat: nested speech handles

longcw:
Replace https://github.com/livekit/agents/pull/1112 for https://github.com/livekit/agents/issues/1054

### DIFF
--- a/examples/voice-pipeline-agent/function_calling_weather.py
+++ b/examples/voice-pipeline-agent/function_calling_weather.py
@@ -40,10 +40,14 @@ class AssistantFnc(llm.FunctionContext):
         # interaction, consider varying the responses based on context or user input.
         call_ctx = AgentCallContext.get_current()
         message = f"Let me check the weather in {location} for you."
-        await call_ctx.agent.say(message)
+        speech_handle = await call_ctx.agent.say(message)  # noqa: F841
 
         # (optional) add the filler message to the chat context for synthesis the tool call speech
         call_ctx.chat_ctx.append(text=message, role="assistant")
+
+        # Or wait for the speech to finish, the said message will be added to the chat context
+        # automatically when the `add_to_chat_ctx` is True (default)
+        # await speech_handle.join()
 
         logger.info(f"getting weather for {location}")
         url = f"https://wttr.in/{location}?format=%C+%t"
--- a/examples/voice-pipeline-agent/function_calling_weather.py
+++ b/examples/voice-pipeline-agent/function_calling_weather.py
@@ -11,7 +11,7 @@ from livekit.agents import (
     cli,
     llm,
 )
-from livekit.agents.pipeline import VoicePipelineAgent
+from livekit.agents.pipeline import AgentCallContext, VoicePipelineAgent
 from livekit.plugins import deepgram, openai, silero
 
 load_dotenv()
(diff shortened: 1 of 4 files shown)


examples/voice-pipeline-agent/function_calling_weather.py
```diff
@@ -33,6 +33,18 @@ async def get_weather(
...
+        # Example of a filler message while waiting for the function call to complete.
+        # NOTE: This message illustrates how the agent can engage users by using the `say()` method
+        # while awaiting the completion of the function call. To create a more dynamic and engaging
+        # interaction, consider varying the responses based on context or user input.
+        call_ctx = AgentCallContext.get_current()
+        message = f"Let me check the weather in {location} for you."
+        await call_ctx.agent.say(message)
+
+        # (optional) add the filler message to the chat context for synthesis the tool call speech
+        call_ctx.chat_ctx.append(text=message, role="assistant")
```
longcw, line 46: Can we allow users to manage the chat ctx used for tool call answer synthesis? For example, if a user inserts some messages using "say" in a function call, they can choose to attach that messages to the chat ctx used to synthesize the answer later.

This would allow us to trigger the llm at the end of the function call instead of waiting for the speech to finish and using the latest global chat ctx. See the difference in this commit https://github.com/livekit/agents/pull/1130/commits/d24c43766b9f460b7638b869e0493801932de94d.

@theomonnom please let me know what do you think.

theomonnom: Yes, I think it is a great idea! Users should be able to edit the chat_ctx inside the call_ctx

theomonnom: `await call_ctx.agent.say(message, add_to_chat_ctx=True)` should be automatic.

There should be no need to explicitly call `call_ctx.chat_ctx.append(text=message, role="assistant")` if you used the `say` method


But for other usecases like injecting RAG for the current call, I think `call_ctx.chat_ctx.append` makes a ton of sense.

longcw: `await call_ctx.agent.say(message, add_to_chat_ctx=True)` only updates the chat ctx after speech playing so the answer synthesis of the function call needs to wait for the speech finished. By manually updating `call_ctx.chat_ctx`, we can trigger a llm call once the function completes, and the user can decide what to put into the chat ctx.

theomonnom: Right I see, I think we should maybe add a new argument to `agent.say`? To declare if we want to wait or not?
Maybe we should make `agent.say` sync but it returns an `asyncio.Future` that users can optionally choose to wait?

wdyt?

longcw: Let `say` return a `Future` do makes sense in case some one want to wait for the speech, but I don't think waiting for `say` in a function call can solve the problem. It's still decided by the pipeline agent where to synthesize the tool call answer,
1. right after the function call returned 
2. or until this tool call speech handle being processed in `play_speech`

Perhaps we can make this an option to the user from the function call?

theomonnom: We have to wait for the function call to returns. Because we use the return value as the "answer"

longcw: Yeah, both of the options will wait for the function call returned. I'll send you a message in slack.

theomonnom (approved):

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -754,83 +781,104 @@ def _commit_user_question_if_needed() -> None:
...
+                )
+                try:
+                    await called_fnc.task
+                except Exception as e:
+                    logger.exception(
+                        "error executing ai function",
                         extra={
                             "function": fnc.function_info.name,
                             "speech_id": speech_handle.id,
                         },
```
theomonnom, line 805: Do we still allow nested function calls (For the same SpeechHandle). Seems like this changed quite a bit. Just making sure

longcw: Yes, if there is any nested function calls they will be processed in the nested speech handle recursively.

PR #1136, opened 2024-11-26
fix: add max_job_memory_usage for worker

longcw:


livekit-agents/livekit/agents/worker.py
```diff
@@ -781,3 +800,54 @@ async def _update_job_status(self, proc: ipc.job_executor.JobExecutor) -> None:
         )
         msg = agent.WorkerMessage(update_job=update)
         await self._queue_msg(msg)
+
+    async def _monitor_memory(self) -> None:
```
theomonnom, line 804: I think we should move that to the PorcJobExecutor class. Otherwise looks good!

longcw: Moved to the ProcJobExecutor

livekit-agents/livekit/agents/worker.py
```diff
@@ -158,6 +158,11 @@ class WorkerOptions:
 
     Defaults to 0.75 on "production" mode, and is disabled in "development" mode.
     """
+    max_job_memory_usage: float = 0
```
theomonnom, line 161: Let's put something like 300MB by default?

livekit-agents/livekit/agents/ipc/proc_job_executor.py
```diff
@@ -403,6 +417,48 @@ async def _pong_timeout_co():
...
+                        },
+                    )
+                    self._exception = JobExecutorError_MemoryLimitExceeded()
+                    self._send_kill_signal()
+
+            except (psutil.NoSuchProcess, psutil.AccessDenied) as e:
+                logger.warning(
+                    "Failed to get memory info for process",
+                    extra={"error": str(e), **self.logging_extra()},
+                )
```
theomonnom, line 450:
```suggestion
                logger.warning(
                    "Failed to get memory info for process",
                    extra=self.logging_extra(). exc_info=e,
                )
```

livekit-agents/livekit/agents/ipc/proc_job_executor.py
```diff
@@ -403,6 +417,48 @@ async def _pong_timeout_co():
...
+            except Exception as e:
+                if self._closing or self._kill_sent:
+                    return
+
+                logger.exception(
+                    "Error in memory monitoring task",
+                    extra={"error": str(e), **self.logging_extra()},
+                )
```
theomonnom, line 458:
```suggestion
                logger.exception(
                    "Error in memory monitoring task",
                    extra=self.logging_extra(),
                )
```

theomonnom (approved):

theomonnom:
Lgtm just a few nits!

livekit-agents/livekit/agents/worker.py
longcw: I would prefer to leave it empty and let the users to decide.

This might be a breaking change if there is a default limit. I have tested that the minimal assistant agent the RAM usage is around 150M, but it could be higher if users load some other data or publishing/receiving videos. Those agents will be killed unexpectedly after upgrading.

davidzhao: it makes sense to not break existing apps. Is it possible to have a "warn" mode where we initially let folks know about the high memory usage without killing the requests? Since most users don't pay attention to memory, making memory issues clearly visible would be beneficial.

maybe we should have two separate options?
 - `job_memory_warn`
 - `job_memory_limit`

longcw: `job_memory_warn_mb` (defaults to 300) and `job_memory_limit_mb` are added.

theomonnom (approved):

PR #1148, opened 2024-11-28
fix: empty agent transcript

jayeshp19:
This PR fixes the bug where a message was being added to the chat context even when `msg.content` was empty.

image
It's the same speech id which has tool calls with no content from model's response
image

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -882,6 +882,7 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
 
             if interrupted:
                 collected_text += "..."
+
             if collected_text:
                 msg = ChatMessage.create(text=collected_text, role="assistant")
                 self._chat_ctx.messages.append(msg)
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -882,25 +882,25 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
 
             if interrupted:
                 collected_text += "..."
+            if collected_text:
+                msg = ChatMessage.create(text=collected_text, role="assistant")
+                self._chat_ctx.messages.append(msg)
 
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -883,24 +883,25 @@ async def _execute_function_calls() -> None:
             if interrupted:
                 collected_text += "..."
 
-            msg = ChatMessage.create(text=collected_text, role="assistant")
-            self._chat_ctx.messages.append(msg)
+            if collected_text:
```
davidzhao, line 886: this would skip `speech_handle.mark_speech_committed()`, do we want to mark it as committed before skipping the other operations?

theomonnom, line 886: When can this happens? I'm more worried about the cause than fixing the scenario where it is indeed empty

jayeshp19: This happens when agent finish speaking  tool call's response. we would also get an extra empty transcript. [screenshot](https://github.com/livekit/agents/pull/1148#issue-2702435425)

Can there be a possiblity when llm response is just empty?

jayeshp19: model's response contains empty content when it responds with a tool call, should we still mark it as committed? because after executing function model will respond with string that will be marked as committed

theomonnom: Ah right, I think we should add new arguments to the agent_speech_committed event. I think it is a valid usecase to have an empty content but having a function calls

davidzhao: > Ah right, I think we should add new arguments to the agent_speech_committed event.

@theomonnom wouldn't that break existing callbacks? I'm not sure if there's value in it because function calls are supposed to be handled by the framework

theomonnom: I will think more about that alongside the new IO proposal

theomonnom: Let's put this PR in standby for now

PR #1160, opened 2024-12-02
add FallbackAdapter example

theomonnom:


examples/voice-pipeline-agent/fallback_adapter.py
```diff
@@ -0,0 +1,77 @@
...
+        [
+            cartesia.TTS(),
+            openai.TTS(),
+        ]
+    )
+
+    agent = VoicePipelineAgent(
+        vad=vad,
+        stt=fallback_stt,
+        llm=openai.LLM(),
```
davidzhao, line 61: should we have an LLM fallback as well?

examples/voice-pipeline-agent/fallback_adapter.py
```diff
@@ -0,0 +1,77 @@
...
+            deepgram.STT(),
+            stt.StreamAdapter(stt=openai.STT(), vad=vad),
+        ]
+    )
+
+    # fallback to OpenAI if Cartesia goes down
+    fallback_tts = tts.FallbackAdapter(
+        [
+            cartesia.TTS(),
+            openai.TTS(),
```
davidzhao, line 54: nit: use elevenlabs since it supports voice cloning.. OAI doesn't.

```suggestion
            elevenlabs.TTS(),
```

davidzhao (approved):
nice. approving with a few nits.

let's merge after the latest version is released

PR #1161, opened 2024-12-02
Flexible IO proposal

theomonnom:


proposal.md
```diff
@@ -0,0 +1,59 @@
...
+        return None
+
+    def after_llm_node(self, source: AsyncIterator[ChatChunk]) -> AsyncIterator[ChatChunk]:
+        return source
+
+    def before_tts_node(self, source: AsyncIterator[str]) -> AsyncIterator[rtc.AudioFrame] | None:
+        return source
+
+    def after_tts_node(self, source: AsyncIterator[rtc.AudioFrame]) -> AsyncIterator[rtc.AudioFrame]:
+        return source
```
rektdeckard, line 40: Not my wheelhouse, but is this in any way more convenient or idiomatic than if each pipeline stage managed its own pre/post transform callbacks? E.G.:

```python
def passthrough_audio(source: AsyncIterator[rtc.AudioFrame]) -> AsyncIterator[rtc.AudioFrame]:
    return source

def filter_swearwords(source: AsyncIterator[SpeechEvent]) -> AsyncIterator[SpeechEvent]:
    return source

agent = PipelineAgent(
    stt=STT(pre=passthrough_audio, post=filter_swearwords)
    ...
)
```

theomonnom: The scope is different, ideally, we can add new parameters like speech_id inside each step like before_tts_node, ...

proposal.md
```diff
@@ -0,0 +1,137 @@
...
+    def clear_queue(self) -> None: ...
+
+
+class TextOutput(Protocol):
+    async def write(self, text: str) -> None: ...
+
+    def flush(self) -> None: ...
+
+
+class PipelineIO(ABC):
```
bcherry, line 26: this name feels a little odd, given that in addition to `PipelineIO` we also have `PipelineOutput` (and maybe some forthcoming `Input` protocol too), but the "IO" in `PipelineIO` is not related to the pipeline's input nor output itself... might be better as `PipelineHooks`

davidzhao: I like `PipelineHooks` as a name

Issue #1162, opened 2024-12-02
Append agent.say after function_call output


longcw:
After test https://github.com/livekit/agents/pull/1130 for some cases, I found it better to append the contents of the `say` called in the function after the `tool_output` so that LLM could complete the full answer exactly following the said content.

Can we add such a flexibility for user to set a `extra_chat_messages` in `AgentCallContext` that will be appended after the `extra_tools_messages`?

PR #1165, opened 2024-12-03
fix: add extra_chat_message to the end of the function call outputs

longcw:
For https://github.com/livekit/agents/issues/1162

When adding the said message in function call to the end of the function call outputs, the LLM can complete the answer following the added text, for example
```python
agent.say("The current weather in New York is")

# The answer LLM after function call will be
"overcast with a temperature of 37°F."
```
If the said message is added before the function call outputs, the LLM tend to repeat it some times
```python
The current weather in New York is  # agent.say
The current weather in New York is overcast with a temperature of 37°F.  # answer LLM output
```

### DIFF
--- a/examples/voice-pipeline-agent/function_calling_weather.py
+++ b/examples/voice-pipeline-agent/function_calling_weather.py
@@ -50,12 +50,11 @@ class AssistantFnc(llm.FunctionContext):
         ]
         message = random.choice(filler_messages).format(location=location)
 
-        # NOTE: set add_to_fnc_call_ctx=True will add the message to the end
+        # NOTE: set add_to_chat_ctx=True will add the message to the end
         #   of the chat context of the function call for answer synthesis
-        speech_handle = await call_ctx.agent.say(message, add_to_fnc_call_ctx=True)  # noqa: F841
+        speech_handle = await call_ctx.agent.say(message, add_to_chat_ctx=True)  # noqa: F841
 
-        # Or wait for the speech to finish, the said message will be added to the chat context
-        # automatically when the `add_to_chat_ctx` is True (default)
+        # To wait for the speech to finish
         # await speech_handle.join()
 
         logger.info(f"getting weather for {location}")
--- a/examples/voice-pipeline-agent/function_calling_weather.py
+++ b/examples/voice-pipeline-agent/function_calling_weather.py
@@ -49,16 +49,10 @@ class AssistantFnc(llm.FunctionContext):
             "The current weather in {location} is ",
         ]
         message = random.choice(filler_messages).format(location=location)
-        speech_handle = await call_ctx.agent.say(message)  # noqa: F841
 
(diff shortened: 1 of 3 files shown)


theomonnom (approved):
Makes sense!

examples/voice-pipeline-agent/function_calling_weather.py
```diff
@@ -39,11 +40,25 @@ async def get_weather(
...
+        # message = f"Let me check the weather in {location} for you."
+        message = f"Here is the weather in {location}: "
+        filler_messages = [
+            "Let me check the weather in {location} for you.",
+            "Let me see what the weather is like in {location} right now.",
+            # LLM will complete this sentence if it is added to the end of the chat context
+            "The current weather in {location} is ",
+        ]
+        message = random.choice(filler_messages).format(location=location)
         speech_handle = await call_ctx.agent.say(message)  # noqa: F841
```
theomonnom, line 52: Ideally, the say method should automatically do it, so the users don't have to do anything

longcw: There might be other sources of `say` like from a callback of an event, we can't know if it's from a function call even there is a function call in progress right?

theomonnom: AgentCallContext is using a ContextVar that is only available inside function calls

longcw: I am thinking to get the call_ctx using `AgentCallContext.get_current` inside `say` method, but wondering is there a risk that there is a function call running while the user call `say` from other places like a callback of a event. In that case the `AgentCallContext.get_current` still returns a context?

longcw: If that's possible, we can add an explicit arg like `add_to_call_ctx` (default False) for `say` so at least users can do that easier. What do you think?

theomonnom: I don't think it is possible (at least very unlikely)

theomonnom: Tho we can still add this new parameter

longcw: @theomonnom Added a arg `add_to_fnc_call_ctx` for say, please take a look.

longcw: I don't think so. `add_to_ctx` can be True for most of cases no matter it's inside or outside a function call.

theomonnom: I think it's fine, you almost always want your new speech to be inside the context. And making it False will do what users expect to do

longcw: Ok, I'll update to share the `add_to_ctx`

longcw: Updated.

theomonnom (approved):

PR #1168, opened 2024-12-03
feat: add AVSynchronizer

longcw:


livekit-agents/livekit/agents/utils/av_sync.py
```diff
@@ -0,0 +1,173 @@
...
+            av_sync.push(video_frame)
+            av_sync.push(audio_frame)
+    """
+
+    def __init__(
+        self,
+        *,
+        audio_source: Optional[rtc.AudioSource],
+        video_source: Optional[rtc.VideoSource],
+        video_sample_rate: Optional[float],
```
davidzhao, line 33: nit: better to call it fps

```suggestion
        video_fps: Optional[float],
```

livekit-agents/livekit/agents/utils/av_sync.py
```diff
@@ -0,0 +1,173 @@
...
+
+    Usage:
+        av_sync = AVSynchronizer(
+            audio_source=audio_source,
+            video_source=video_source,
+            video_sample_rate=video_sample_rate,
+        )
+
+        async for video_frame, audio_frame in video_generator:
+            av_sync.push(video_frame)
```
davidzhao, line 24:
```suggestion
            await av_sync.push(video_frame)
```

davidzhao: for docs, maybe also clarify that it could be pushed from different tasks?

livekit-agents/livekit/agents/utils/av_sync.py
```diff
@@ -0,0 +1,173 @@
...
+            av_sync.push(audio_frame)
+    """
+
+    def __init__(
+        self,
+        *,
+        audio_source: Optional[rtc.AudioSource],
+        video_source: Optional[rtc.VideoSource],
```
davidzhao, line 32: should these be required?

livekit-agents/livekit/agents/utils/av_sync.py
longcw: Yeah, I'll create an example to test and demonstrate the use case.

davidzhao: this example is awesome! it'll be super useful. It might be useful to include an example file too.. I'll see about moving some of these examples to its own organized repo in livekit-examples org

longcw: Yes, or I was thinking of another example of reading video and audio from the client and sending it back, or an audio wave with a graph so we can easily see the synchronization while not needing an additional mp4 file.

livekit-agents/livekit/agents/utils/av_sync.py
```diff
@@ -0,0 +1,158 @@
...
+
+        self._video_queue_max_size = int(
+            self._video_fps * self._video_queue_size_ms / 1000
+        )
+        self._video_queue = asyncio.Queue[rtc.VideoFrame](
+            maxsize=self._video_queue_max_size
+        )
+        self._capture_video_task = asyncio.create_task(self._capture_video())
+
+    async def push(self, frame: Union[rtc.VideoFrame, rtc.AudioFrame]) -> None:
```
theomonnom, line 53: Should this wait to have at least one AudioFrame and one VideoFrame to start the capture? So we synchronize the start of the stream as well

livekit-agents/livekit/agents/utils/av_sync.py
```diff
@@ -0,0 +1,158 @@
...
+        if isinstance(frame, rtc.AudioFrame):
+            # TODO: test if frame duration is too long
```
theomonnom, line 55: Yes, I think we need to keep track of the pushed audio.

theomonnom:
Any idea on how we could handle interruptions with it?

livekit-agents/livekit/agents/utils/av_sync.py
longcw: I think no, the agent can stream the video frames only if there is no audio at the beginning. The user only need to make sure when there is a audio for playing, the first audio frame is pushed at the same time with the corresponding video frame.

longcw: we need to make the queue_size smaller enough for that case, e.g. 100ms or even less.

longcw:
> Any idea on how we could handle interruptions with it?

First if queue_size_ms is small, for interruption we just need to stop the audio pushing.
I'll also add a `clear_queue` method for the AVSync as discussed.

davidzhao:
@longcw I'm also wondering, perhaps it'd be better for the core sync utils to go into our Python SDK? https://github.com/livekit/python-sdks/

since it doesn't require agents here.

longcw:
> @longcw I'm also wondering, perhaps it'd be better for the core sync utils to go into our Python SDK? https://github.com/livekit/python-sdks/
> 
> since it doesn't require agents here.

I agree. I'll finish this with some more testing and documents, then maybe create a pr for the sdk?

theomonnom (approved):

longcw:
@davidzhao The PR for the python sdk https://github.com/livekit/python-sdks/pull/324

PR #1174, opened 2024-12-04
Support PlayHT/PlayAI TTS

jayeshp19:
### What
This PR implements PlayHT support for TTS

### Why
Feature Improvement

### Docs
- https://docs.play.ht/reference/python-sdk


theomonnom:
I think the plugin should still be named PlayHT? https://play.ht/

jayeshp19:
It was a suggestion from @davidzhao to rename as playai, they renamed the company from playht to [play.ai](http://play.ai/) recently

davidzhao:
> interesting, it seems like only their VoiceAgent product is under the play.ai name

I think they've changed it officially. Here's the docs link: https://docs.play.ai/tts-api-reference/endpoints/v1/tts/stream/post-playdialog

davidzhao:
ok, so this is a bit confusing but looks like there are two APIs:

1. PlayHT API/SDK v3: https://docs.play.ht/reference/playht-tts-websocket-api
2. PlayAI API/SDK v1: https://docs.play.ai/tts-api-reference/endpoints/v1/tts/post-async-play30mini

PlayAI seems like their real API play, so I think we should integrate that with the [Websocket API](https://docs.play.ai/tts-api-reference/endpoints/v1/tts/post-async-play30mini) (without their SDK)

PlayHT now states they don't have API access within their paid offerings: https://play.ht/pricing/

jayeshp19:
As per the discussion with @davidzhao, we're keeping the name `playht` as it seems like their tts product.

livekit-agents/livekit/agents/utils/codecs/mp3.py
```diff
@@ -39,6 +39,20 @@ def __init__(self):
...
+            # The size is encoded in the next 4 bytes (bytes 6-9)
+            # Each byte only uses 7 bits (most significant bit is always 0)
+            if len(chunk) >= 10:
+                size = (
+                    ((chunk[6] & 0x7F) << 21)
+                    | ((chunk[7] & 0x7F) << 14)
+                    | ((chunk[8] & 0x7F) << 7)
+                    | (chunk[9] & 0x7F)
+                )
+                chunk = chunk[10 + size :]
```
theomonnom, line 54: Weird that avcodec doesn't ignore it

theomonnom (approved):

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
```diff
@@ -0,0 +1,313 @@
...
+            raise ValueError(
+                "PlayAI API key and user ID are required, either as arguments or set PLAYAI_API_KEY and PLAYAI_USER_ID environment variables"
+            )
+
+        self._client = PlayHTAsyncClient(
+            user_id=user_id,
+            api_key=api_key,
+        )
+        # suppress verbose websocket logs
+        logging.getLogger("websockets.client").setLevel(logging.INFO)
```
theomonnom, line 97: This should be done inside the log.py in the cli package

theomonnom:
The Stream implementation doesn't seems to support multiple segments (We don't have a test covering it yet — discussed about it already in Slack with @jayeshp19).

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
```diff
@@ -0,0 +1,313 @@
...
+from livekit.agents import (
+    DEFAULT_API_CONNECT_OPTIONS,
+    APIConnectionError,
+    APIConnectOptions,
+    tokenize,
+    tts,
+    utils,
+)
+from pyht import AsyncClient as PlayHTAsyncClient  # type: ignore
+from pyht.client import Format, Language, TTSOptions  # type: ignore
```
theomonnom, line 18: I don't think we want to ignore type checks for the whole pyht package?

examples/voice-pipeline-agent/minimal_assistant.py
```diff
@@ -48,7 +48,7 @@ async def entrypoint(ctx: JobContext):
         vad=ctx.proc.userdata["vad"],
         stt=deepgram.STT(model=dg_model),
         llm=openai.LLM(),
-        tts=openai.TTS(),
```
theomonnom, line 51: Let's not change the minimal_assistant code

.github/workflows/tests.yml
```diff
@@ -80,6 +80,12 @@ jobs:
           sudo dpkg -i libssl1.1_1.1.1-1ubuntu2.1_arm64.deb
           sudo dpkg -i libssl-dev_1.1.1-1ubuntu2.1_arm64.deb
 
+        # pytorch doesn't support 3.13 yet, we have to explicitly install 3.12 to prevent
+        # ffmpeg from pulling in 3.13
+      - name: Install python 3.12 (macOS)
+        if: ${{ startsWith(matrix.os, 'macos') }}
+        run: brew install python@3.12
+
```
theomonnom, line 88: Why not use the setup-python action above?

jayeshp19:
Do they support playai via playht sdk? I think we might have to update code according to https://docs.play.ai/tts-api-reference/websocket

.github/workflows/tests.yml
davidzhao: brew will still try to install its version, and complain about a conflicted file

examples/voice-pipeline-agent/minimal_assistant.py
davidzhao: will not be merged.. this is for the Play.ai team to test

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
davidzhao: their package is not typed

livekit-agents/livekit/agents/utils/codecs/mp3.py
davidzhao: this is part of the container, usually decoders don't handle container-specific logic. libavformat handles it.. but currently the mp3 abstraction doesn't work well with it.

the "right" way of using it is to construct a stream each time we receive a new file, and then read the rest of the decoded stream.

since the abstraction is incorrect and would be a bigger change, I opted for a simpler path since mp3 headers are fairly simple to handle anyways.

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
davidzhao: yeah good point about doing it here.. this might not be an issue anymore, the team suggested using their websockets directly, since that supports play.ai keys. I'll update this later on

davidzhao:
> Do they support playai via playht sdk? I think we might have to update code according to https://docs.play.ai/tts-api-reference/endpoints/v1/tts/post-async-play30mini

Asking them about that, they suggested using the new websocket directly, which supports playai. I'll update this to remove the SDK

> 
> also only websocket may not be ideal as it can be timed out in 10 seconds https://docs.play.ai/tts-api-reference/websocket#connection-timeout

They indicated the new websocket does not timeout like this, so that'll be better anyways

.github/workflows/tests.yml
```diff
@@ -109,6 +115,8 @@ jobs:
           GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
           ASSEMBLYAI_API_KEY: ${{ secrets.ASSEMBLYAI_API_KEY }}
           FAL_KEY: ${{ secrets.FAL_KEY }}
+          PLAYAI_API_KEY: ${{ secrets.PLAYAI_API_KEY }}
```
bryananderson, line 118: You'll need to use play.ht keys, not play.ai keys, with this SDK. This may be the reason for the failing auth in the tests.

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
```diff
@@ -0,0 +1,313 @@
...
+
+from .models import TTSModel
+
+
+@dataclass
+class _TTSOptions:
```
bryananderson, line 24: Is there a reason to reimplement this class instead of using the imported `TTSOptions` from `pyht` on [this line](https://github.com/livekit/agents/pull/1174/files#diff-dd57d2ed46c52ac07ce07d9ec376392e537c01aef81b937938991438d355d18bR18)?

A number of new options have been added to the class in `pyht` which aren't here, and this may continue to happen.

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
```diff
@@ -0,0 +1,313 @@
...
+        """
+        super().__init__(
+            capabilities=tts.TTSCapabilities(
+                streaming=True,
+            ),
+            sample_rate=sample_rate,
+            num_channels=1,
+        )
+
+        api_key = api_key or os.environ.get("PLAYAI_API_KEY")
```
bryananderson, line 84: See my other comment on `tests.yml`; you want to use play.ht credentials here, not play.ai

bryananderson: I know this distinction is confusing and I'm sorry! In the future we hope to smooth all of this out.

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
```diff
@@ -0,0 +1,313 @@
...
+            language=Language(language),
+            temperature=temperature,
+            top_p=top_p,
+            text_guidance=text_guidance,
+            voice_guidance=voice_guidance,
+            style_guidance=style_guidance,
+            repetition_penalty=repetition_penalty,
+        )
+
+    def update_options(
```
bryananderson, line 113: See my previous comment about new options that have been added; maybe simpler to just use `**kwargs` here instead?

bryananderson:
Added a few comments. How has your testing been going? Any issues besides the auth key thing?

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/models.py
```diff
@@ -0,0 +1,15 @@
+from typing import Literal
+
+from pyht.client import Format  # type: ignore
+
+TTSModel = Literal["Play3.0-mini-ws", "PlayDialog-ws", "Play3.0-mini", "PlayDialog"]
+FORMAT = Literal["raw", "mp3", "wav", "ogg", "flac", "mulaw", "pcm"]
```
davidzhao, line 6: let's include only the formats that the framework could decode. I'd recommend sticking with MP3 for now

davidzhao (commented):
nice! just one comment on format

davidzhao (approved):

gianpaj:
@jayeshp19 could you please publish the plugin on pypi? 🙏 

https://pypi.org/project/livekit-plugins-playai - 404

jayeshp19:
Hi @gianpaj, we'll be publishing it soon! In the meantime, you can install it directly from the agents directory using:  
```
pip install livekit-plugins\livekit-plugins-playai
```  

Hope that helps!

PR #1181, opened 2024-12-05
Add docs to ChatImage

bcherry:
Very basic usage doc. I also updated the error message from the anthropic plugin to be easier to understand (the old one wasn't clear it came from LiveKit plugins rather than from Anthropic itself)

### DIFF
--- a/livekit-agents/livekit/agents/llm/chat_context.py
+++ b/livekit-agents/livekit/agents/llm/chat_context.py
@@ -26,11 +26,40 @@ ChatRole = Literal["system", "user", "assistant", "tool"]
 
 @dataclass
 class ChatImage:
+    """
+    ChatImage is used to input images into the ChatContext on supported LLM providers / plugins.
+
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -26,11 +26,40 @@
 
 @dataclass
 class ChatImage:
+    """
+    ChatImage is used to input images into the ChatContext on supported LLM providers / plugins.
+
+    You may need to consult your LLM provider's documentation on supported URL types.
+
+    ```python
+    # With a VideoFrame, which will be automatically converted a data URL internally
```
kath0la, line 35: Is this missing a word?

```suggestion
    # With a VideoFrame, which will be automatically converted to a data URL internally
```

kath0la (approved):
One minor comment, but this is fantastic!

nbsp (approved):

PR #1197, opened 2024-12-09
chore: Refactor chat completion creation to a separate method

mykola-mokhnach-parloa:
This allows to:
- simplify the `chat` method 
- customize OpenAI completion arguments in subclasses.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -630,13 +630,15 @@ class LLM(llm.LLM):
 
     @wraps(AsyncCompletions.create)
     def create_chat_completions_stream(self, **kwargs):
-        return self._client.chat.completions.create(**{
-            "model": self._opts.model,
-            "stream_options": {"include_usage": True},
-            "stream": True,
-            "user": self._opts.user or openai.NOT_GIVEN,
-            **kwargs,
-        })
+        return self._client.chat.completions.create(
+            **{
+                "model": self._opts.model,
+                "stream_options": {"include_usage": True},
+                "stream": True,
+                "user": self._opts.user or openai.NOT_GIVEN,
+                **kwargs,
+            }
+        )
 
     def chat(
         self,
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -626,6 +628,16 @@ def create_azure_client(
             tool_choice=tool_choice,
         )
 
+    @wraps(AsyncCompletions.create)
+    def create_chat_completions_stream(self, **kwargs):
```
theomonnom, line 632: Is the goal to be able to edit the parameters of the LLM inside the VoicePipelineAgent?

mykola-mokhnach-parloa: in general, yes
we'd like to inject custom values to the corresponding openai API in the descendant class, which are either not implemented in the agent for now or that contain customized logic, that would probably not make much sense for this particular (generic) agent implementation

theomonnom: Could the before_llm_cb callback inside PipelineAgent do the work? You can return an LLMStream from there (so you can use chat with any parameter you want)

mykola-mokhnach-parloa: I assume with the recent changes in main this PR becomes obsolete.
The original intention was to customize parameters passed to the OpenAI completion creation API rather than these that the `chat` one exposes. 

For now I think it would make sense to close this PR since we anyway override `_main_task` (now called `_run`) of the `llm.LLMStream` class

theomonnom: Interesting, I think this [PR](https://github.com/livekit/agents/pull/1210) may be interesting for it

mykola-mokhnach-parloa: most likely it won't because  `_extra_options` cannot override existing arguments

Issue #1199, opened 2024-12-10
How do I handle errors 


nbrustein:
I'm trying to figure out how to respond when an error occurs. For instance, I've experienced [this error](https://github.com/livekit/agents/issues/1056).

I would expect that I would be able to listen for errors. Something like:

```
model = openai.realtime.RealtimeModel(
    instructions=instructions,
    modalities=["audio", "text"],
)
assistant = MultimodalAgent(
    model=model,
    fnc_ctx=VoiceCoursePracticeFunctionContext(),
)

session = model.sessions[0]
session.on('error', __on_error)
```

But, reading through the code for `RealtimeSession`, it looks like nothing happens in response to an error from the RealtimeApi other that that it is logged. (I'm looking in the handling of the `'error'` event in `_recv_task`).

Am I thinking about this correctly? Is there some way for me to listen to errors and respond?

yepher:
+1 for me too.

Until the root issue is resolved, it would be super helpful to detect when the problem occurs so we can programmatically kick the agent to continue.

I have a use case that uses a lot of function calls, so it gets stuck quite often in the flow. A user has to ask the agent to continue, or, I like to say, "Say it out loud."

longcw:
@nbrustein Is this PR https://github.com/livekit/agents/pull/1200 what you needed?

nbrustein:
> @nbrustein Is this PR #1200 what you needed?

yup. That's what I had in mind.

Issue #1183, opened 2024-12-06
System prompt is ignored after fatal errors with realtime api


FiddlersCode:
**How to replicate**
trigger an OpenAI error (e.g. send a tool call description that's too long which triggers the error below)

**Expected behaviour**

- the error is logged and the conversation continues
- our system prompt **is still used**

**Actual behaviour**
- the error is logged but the conversation then continues
- crucially, the system prompt and config that we send to OpenAI are now ignored - e.g. the voice reverts to the default voice and all the safeguards from the system prompt are ignored

```
ERROR livekit.plugins.openai.realtime - OpenAI S2S error {'type': 'error', 'event_id': 'event_AXSNgEBBhqwMQmz2iOy5l', 'error': {'type': 'invalid_request_error', 'code': 'string_above_max_length', 'message': "Invalid 'session.tools[1].description': string too long. Expected a string with maximum length 1024, but got a string with length 1198 instead.", 'param': 'session.tools[1].description', 'event_id': None}}
 ```

Versions
livekit = "0.18.1"
livekit-agents = "0.11.3"
livekit-plugins-openai = "0.10.7"
livekit-protocol = "0.7.0"
 
We’re using a realtime-api model hosted on Azure.
 
I suspect this is the line causing the error - https://github.com/livekit/agents/blob/d67130223e3c00ecdac88a337e6f5b1236fd469b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py#L997
 
While we can of course fix this specific error, we’re concerned that errors are being swallowed by the system and the user is allowed to continue without safeguards. I was also able to raise uncaught exceptions elsewhere in the code to recreate the error.

longcw:
I have created a PR https://github.com/livekit/agents/pull/1200 to emit the error from the realtime API, so that agent can decide to reset or close the session in the callback. Does that make sense?

FiddlersCode:
> I have created a PR #1200 to emit the error from the realtime API, so that agent can decide to reset or close the session in the callback. Does that make sense?

Hi @longcw - many thanks for this! We're now able to read the event emitted (RealtimeError) so that's good.

My only followup question - is it well understood why we can keep talking to OpenAI with the system prompt being ignored? I'm wondering if a new session is being created with default opts but I haven't been able to dig around enough yet so it's only an idea.

This is still a concern for us because there may be other back doors whereby the system prompt is ignored so would be good to understand the fundamental issue.

[...]

PR #1200, opened 2024-12-10
fix: emit error from realtime api

longcw:
Emit the error message from the `RealtimeSession`
```python
    def _handle_error(self, error: api_proto.ServerEvent.Error):
        logger.error(
            "OpenAI S2S error %s",
            error,
            extra=self.logging_extra(),
        )
        error_content = error["error"]
        self.emit(
            "error",
            RealtimeError(
                event_id=error["event_id"],
                type=error_content["type"],
                message=error_content["message"],
                code=error_content.get("code"),
                param=error_content.get("param"),
            ),
        )
```


longcw:
Related to https://github.com/livekit/agents/issues/1183, https://github.com/livekit/agents/issues/1199

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -141,6 +141,15 @@ class InputTranscriptionOptions:
     model: api_proto.InputTranscriptionModel | str
 
 
+@dataclass
+class RealtimeError:
```
davidzhao, line 145: we should export this class too, for those that want to keep strong types

davidzhao (approved):
lg, just comment around export

theomonnom (approved):

PR #1201, opened 2024-12-10
Support Deepgram TTS

jayeshp19:
### What
- This PR implements Deegram support for TTS
### Why
Feature Improvement

### Docs
https://developers.deepgram.com/docs/text-to-speech
https://developers.deepgram.com/reference/text-to-speech-api

https://developers.deepgram.com/docs/streaming-text-to-speech
https://developers.deepgram.com/reference/transform-text-to-speech-websocket

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/_utils.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/_utils.py
@@ -1,6 +1,5 @@
 import time
 from typing import Callable, Generic, Optional, TypeVar
-from urllib.parse import urlencode
 
 T = TypeVar("T")
 
@@ -35,27 +34,3 @@ class PeriodicCollector(Generic[T]):
             self._callback(self._total)
             self._total = None
         self._last_flush_time = time.monotonic()
-
-
-def _to_deepgram_url(
-    opts: dict,
-    base_url: str,
-    *,
-    websocket: bool,
-) -> str:
-    if opts.get("keywords"):
-        # convert keywords to a list of "keyword:intensifier"
-        opts["keywords"] = [
-            f"{keyword}:{intensifier}" for (keyword, intensifier) in opts["keywords"]
-        ]
-
-    # lowercase bools
-    opts = {k: str(v).lower() if isinstance(v, bool) else v for k, v in opts.items()}
-
-    if websocket and base_url.startswith("http"):
-        base_url = base_url.replace("http", "ws", 1)
-
-    elif not websocket and base_url.startswith("ws"):
-        base_url = base_url.replace("ws", "http", 1)
-
-    return f"{base_url}?{urlencode(opts, doseq=True)}"
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/_utils.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/_utils.py
@@ -1,5 +1,6 @@
 import time
 from typing import Callable, Generic, Optional, TypeVar
+from urllib.parse import urlencode
 
 T = TypeVar("T")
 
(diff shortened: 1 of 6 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/_utils.py
```diff
@@ -34,3 +35,27 @@ def flush(self) -> None:
...
+    # lowercase bools
+    opts = {k: str(v).lower() if isinstance(v, bool) else v for k, v in opts.items()}
+
+    if websocket and base_url.startswith("http"):
+        base_url = base_url.replace("http", "ws", 1)
+
+    elif not websocket and base_url.startswith("ws"):
+        base_url = base_url.replace("ws", "http", 1)
+
+    return f"{base_url}?{urlencode(opts, doseq=True)}"
```
theomonnom, line 61: I don't think we should share this function because some stuff are specific to stt here

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+class TTS(tts.TTS):
+    def __init__(
+        self,
+        *,
+        model: str = "aura-asteria-en",
+        encoding: str = "linear16",
+        sample_rate: int = 24000,
+        container: str = "none",
```
theomonnom, line 44: will not be used since rtc.AudioFrame doesn't support it 
```suggestion
```

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
```diff
@@ -0,0 +1,341 @@
...
+
+            while True:
+                msg = await ws.receive()
+                if msg.type in (
+                    aiohttp.WSMsgType.CLOSE,
+                    aiohttp.WSMsgType.CLOSED,
+                    aiohttp.WSMsgType.CLOSING,
+                ):
+                    return
+
```
theomonnom, line 283: Errors are not propagated to the _run_task. So no retry will happen on failure

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
jayeshp19: image
https://developers.deepgram.com/docs/tts-media-output-settings
I think it won't affect rtc.AudioFrame because it's just a way to get specific Content-Type

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/_utils.py
jayeshp19: maybe we can have param to check if it's coming stt or tts, should we keep it separate?

theomonnom: separate is better

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
```diff
@@ -0,0 +1,413 @@
...
+                )
+                closing_ws = False
+
+                tasks = [
+                    asyncio.create_task(_tokenize_input()),
+                    asyncio.create_task(_run_segments(ws)),
+                    asyncio.create_task(recv_task(ws)),
+                ]
+                wait_reconnect_task = asyncio.create_task(self._reconnect_event.wait())
+                connection_timeout_task = asyncio.create_task(_connection_timeout())
```
theomonnom, line 371: This task is never cancelled

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
```diff
@@ -0,0 +1,413 @@
...
+                    elif mtype == "Warning":
+                        logger.warning("Deepgram warning: %s", resp.get("warn_msg"))
+                    elif mtype == "Metadata":
+                        pass
+                    else:
+                        logger.debug("Unknown message type: %s", resp)
+
+        async def _connection_timeout():
+            # Deepgram has a 60-minute timeout period for websocket connections
+            await asyncio.sleep(3599)
```
theomonnom, line 344: 1 second isn't enough, let's do 60min - 5min

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
```diff
@@ -0,0 +1,413 @@
...
+            logger.warning("Deepgram TTS connection timed out. reconnecting...")
```
theomonnom, line 345: Change the log to say that this timeout is expected (like maximum connection time reached..)

theomonnom (approved):
Otherwise lgtm, nice work!

Issue #1056, opened 2024-11-08
"Conversation already has an active response" error with MultimodalAgent


davidzhao:
We've gotten reports in Slack that when certain function calls are used with MultimodalAgent, it would sometimes error out with:

```
2024-11-07 17:14:37,898 - ERROR livekit.plugins.openai.realtime - OpenAI S2S error {'type': 'error', 'event_id': 'event_AR7s53avu18NBq2Lk5O4r', 'error': {'type': 'invalid_request_error', 'code': None, 'message': 'Conversation already has an active response', 'param': None, 'event_id': None}} {"session_id": "sess_AR7r1ss6nsPVlPETCgFxi"}
```

It'd be great if anyone could produce a path to reproducing the error.

yepher:
I've been getting this error quite a bit recently. I feel like it happens more when I give a pretty long-winded response to the LLM.

rpiacquadioryz:
Hey Guys! just got the same error

I have implemented a function to switch the prompt. In my use case the prompt is so long that I'm trying to split it into several files. So at some point of the first prompt I call the action to switch to the following one, and then I do the same thing in the second prompt.

Here's the snippet:


`
class PromptManager:
    def __init__(self):
        self.current_agent = 1
        self.max_agents = 5
        self.prompts = {}
        self._load_prompts()
    
    def _load_prompts(self):
        """Load all prompt files"""
        for i in range(1, self.max_agents + 1):
            try:
                with open(f'prompts/prompt_{i}.txt', 'r', encoding='utf-8') as f:
                    self.prompts[f'prompt_{i}'] = f.read().strip()
            except Exception as e:
                print(f"Error loading prompt_{i}: {e}")
    
    def get_current_prompt(self) -> str:
        """Get the current prompt text"""
        logger.info(f"Getting current prompt: {self.current_agent}")
        prompt_key = f'prompt_{self.current_agent}'
        return self.prompts.get(prompt_key, '')
    
    def move_to_next_agent(self) -> bool:
        """Move to next agent if possible"""
        logger.info(f"Moving to next agent: {self.current_agent}")
        if self.current_agent < self.max_agents:
            self.current_agent += 1
            return True
        return False
    
    def get_current_agent_number(self) -> int:
        return self.current_agent
    
    def reset(self):
        """Reset to first agent"""
        self.current_agent = 1


prompt_manager = PromptManager()

# first define a class that inherits from llm.FunctionContext
class AssistantFnc(llm.FunctionContext):
[...]

PR #1204, opened 2024-12-10
fix: cancel active response for tool call in multimodal agent

longcw:
For https://github.com/livekit/agents/issues/1056

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -689,6 +689,41 @@ class RealtimeSession(utils.EventEmitter[EventTypes]):
         def create(self) -> None:
             self._sess._queue_msg({"type": "response.create"})
 
+        async def acreate(self, *, cancel_existing: bool = True) -> None:
+            # FIXME: should we merge this with `create`?
+            pending_create_fut = self._sess._response_create_fut
+            if pending_create_fut is not None:
+                if not cancel_existing:
+                    return
+                await pending_create_fut
+
+            pending_resp_id = self._sess._active_response_id
+            if pending_resp_id:
+                if not cancel_existing:
+                    logger.warning(
+                        "active response exists, skipping creation",
+                        extra={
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -689,6 +689,41 @@ def __init__(self, sess: RealtimeSession) -> None:
...
+            pending_create_fut = self._sess._response_create_fut
+            if pending_create_fut is not None:
+                if not cancel_existing:
+                    return
+                await pending_create_fut
+
+            pending_resp_id = self._sess._active_response_id
+            if pending_resp_id:
+                if not cancel_existing:
+                    logger.warning(
```
davidzhao, line 703: I think there are three behaviors right?
1. canceling existing response, enqueuing new response
2. not canceling existing response, aborting the new response
3. not canceling existing response, enqueuing new response

it seems 1 or 2 are the most common path, but for the sake of API design, should it be:
* on_duplicate: "cancel_existing" | "cancel_new" | "keep_both" ?

davidzhao: not loving the name `acreate` either.. `create` seems so good.. but.. maybe `async request`, like `await response.request()` ?

just an idea. .not sure if I love it either

longcw: Yes, there are three options. I'll update the API. I think `create` is good but it's originally a sync method, not sure if it's ok to change it to an async method which might be a breaking changes as I saw many users are calling `response.create` in the entrypoint.

davidzhao: right, we do not want to break the existing API for that reason.

longcw: Modified the `create` to return a `Future` or a `Task`, how do you think?

PR #1209, opened 2024-12-11
added ConversationPersistor() to document events/transcriptions in external file

tinalenguyen:


theomonnom:
Thanks, this is useful, but I believe it should be placed in the examples folder. Could you use [aiofiles](https://github.com/Tinche/aiofiles) when writing to a file to ensure non-blocking I/O on the main thread?

tinalenguyen:
@theomonnom Thank you for the suggestion, I moved it and used aiofiles. Let me know if there are any other issues/suggestions!

theomonnom (commented):
I think a better pattern would be to create a Queue of every log we want to write inside the file and write it inside a single task. It makes it easier to "close" it. Here is an e.g 

```
class ConversationPersistor:
	def __init__(self) -> None:
	    self._log_q = asyncio.Queue[EventLog | None]()
	    self._main_task = asyncio.create_task(self._main_atask())

	async def _main_atask(self) -> None:
		while True:
			log = await self._log_q.get()
			if log is None:
				break

			# write the log to the file here


		self._log_q.put_nowait(None)
		await self._main_task


	@self._model.on("agent_started_speaking")
	def _agent_started_speaking():
		event = EventLog(eventname="agent_started_speaking")
		self._log_q.put_nowait(event) 
```


For this example to be complete, can you also add an agent entrypoint. (e.g https://github.com/livekit/agents/blob/52880aa9876df77e55aefae7947bbd7a4a5867f3/examples/multimodal_agent.py#L84)

Otherwise lgtm, thanks!

tinalenguyen:
@theomonnom Everything should be all good now, let me know if there's anything else!

theomonnom (approved):
Thanks! lgtm!

Can you run `ruff format && ruff check --fix` before we merge it :)

tinalenguyen:
@theomonnom Just ran and adjusted it! Thanks for helping with my first pull request :)

examples/conversation_persistor.py
```diff
@@ -0,0 +1,212 @@
...
+            voice="alloy",
+            temperature=0.8,
+            instructions="You are a helpful assistant.",
+            turn_detection=openai.realtime.ServerVadOptions(
+                threshold=0.6, prefix_padding_ms=200, silence_duration_ms=500
+            ),
+        ),
+    )
+
+    cp = ConversationPersistor(model=agent, log="log.txt")
```
theomonnom, line 204: ruff CI is failing due to this line not being used.
I think what we could do is to rename main_task() to start()
then call start() here?

tinalenguyen: @theomonnom I updated per your suggestions! Just curious, is it usually better practice to not start the object upon instantiation or is it situational? I figured users would call upon cp in their own code so it wouldn't be a problem.

theomonnom: It's generally better to avoid creating a new `asyncio.Task` during object instantiation. Moving the task creation into the start method offers a more flexible API. This allows for scenarios where you might want to configure or initialize an object without immediately starting it. 

By making the code more explicit, users of the API can confidently assume that no actions such as writing to a file will occur until the `start` method is explicitly called. There are no hidden or implicit behaviors.

For e.g, in our [SupervisedProcess](https://github.com/livekit/agents/blob/b6542abe5f42f5a49bc845301b57e91f049b7cd8/livekit-agents/livekit/agents/ipc/supervised_proc.py#L94) class, the object can be instantiated and initialized, but the actual OS process is only created when the `start` method is called.

theomonnom:
One last thing I promise, can you move the [task creation](https://github.com/tinalenguyen/agents/blob/3718115eef4776456938a4d20b3b7346e0138563/examples/conversation_persistor.py#L73) to the start method? :)

examples/conversation_persistor.py
tinalenguyen: I thought it would be more efficient to have it all done upon instantiating but I definitely see how it's more flexible and user-friendly this way. Thank you for the feedback!

davidzhao:
could you do a `ruff format . && ruff check --fix .`? it's just formatting now and I think it's ready to be merged

PR #1210, opened 2024-12-11
Added extra_options parameter to OpenAI and Anthropic plugins

zahariadaniel16:
This should have been added by default from the beginning, as we want to have access to all the parameters from the LLM APIs to be able to customize them.

One use case would be observability and monitoring. I tried to integrate LiveKit with Langfuse, and I couldn't pass the additional parameters needed to the completions call.

This PR fixes that and lets other developers customize the LLM settings how they want, instead of waiting for each parameter to be added by the LK team (like the new parallel_tool_calls or tool_choice).

### DIFF
--- a/.changeset/green-phones-lick.md
+++ b/.changeset/green-phones-lick.md
@@ -1,0 +1,6 @@
+---
+"livekit-plugins-anthropic": patch
+"livekit-plugins-openai": patch
+---
+
+Added extra_options parameter to OpenAI and Anthropic plugins


theomonnom (approved):

AndrazP:
Hey @theomonnom, can this be merged?

jayeshp19:
> Hey @theomonnom, can this be merged?

Looks like there're some merge conflicts, can you pull latest main in your branch and resolve conflicts?

Issue #1208, opened 2024-12-11
OpenAI agent cannot properly parse various complex function argument types


mykola-mokhnach-parloa:
Steps to reproduce:
1. Define a new function similar to the example in https://github.com/livekit/agents/blob/c5897c1b79f6d57a72cbf4c92c87a5bd85ad3e5d/examples/multimodal_agent.py#L29
2. Set the function argument type to `Optional[str]` or  `Union[int, type(None)]` or `float | None`
3. Run the pipeline

Actual result:
An exception stack trace is raised because the [build_oai_function_description](https://github.com/livekit/agents/blob/6b4e9039a70b0e6ef050b1db964ef21f39e774c0/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/_oai_api.py#L86) type support is limited to only (str, int, float, bool), even though the [FunctionContext](https://github.com/livekit/agents/blob/6b4e9039a70b0e6ef050b1db964ef21f39e774c0/livekit-agents/livekit/agents/llm/function_context.py#L116) class is capable of recognizing the above nullable types.

Expected result:
The `build_oai_function_description` mentioned above should be able to properly parse nullable types like ones above.  
For example, we expect `Optional[str]` to be transformed into `["string", "null"]` OpenAI type definition.

davidzhao:
fixed in #1211

mykola-mokhnach-parloa:
Unfortunately the above change does not completely fix the original issue. The build_oai_function_description function remains the same and is still [unable](https://github.com/livekit/agents/blob/57869fcbba3a480963fc1497bc47d8e1f263ecbb/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/_oai_api.py#L105) to parse optional type definitions. It would be great to have an actual unit test that verifies parsing of above cases and confirms the actual fix.

mykola-mokhnach-parloa:
thanks for your reply @jayeshp19 

I expect the below unit tests to pass (currently some of them fail):

```python
import pytest
from typing import Union, Optional
from inspect import _empty

from livekit.agents.llm import FunctionInfo, FunctionArgInfo
from livekit.agents.llm.function_context import _is_optional_type
from livekit.agents import llm


def test_typing():
    assert _is_optional_type(Optional[int]) == (True, int)
    assert _is_optional_type(Union[str, None]) == (True, str)
    assert _is_optional_type(None | float) == (True, float)
    assert _is_optional_type(Union[str, int]) == (False, None)


[...]

Issue #1185, opened 2024-12-06
Optional values in @llm.ai_callable cause ValueError


FiddlersCode:
**How to replicate**
1. Set tool call params as below:
2. LLM sets this parameter to None

```
@llm.ai_callable(
    description="""
    "Returns a list of transactions during a specified period and filters...blah blah
    """
)
async def a_tool_call(
    self,
    start_date: Annotated[
        Optional[str],
        llm.TypeInfo(
            description="""
                The start date in the format %Y-%m-%d, 
                for example, 2020-09-01 for 1st September 2020.
                If you don't think this field is relevant for the question then set this parameter to None.
                """
        ),
    ] = None,
...more parameters
``` 

**Expected behaviour**
Tool call proceeds as usual with `None` set as a parameter

**Actual behaviour**
We receive the following error message and the tool call is now able to proceed
```
2024-11-29 17:34:13,600 - ERROR livekit.plugins.openai.realtime - failed to handle OpenAI S2S message
Traceback (most recent call last):
  File "/Users/paula.muldoon/dev/ai/ai-livekit-agent/.venv/lib/python3.11/site-packages/livekit/plugins/openai/realtime/realtime_model.py", line 1038, in _recv_task
    self._handle_response_output_item_done(data)
  File "/Users/paula.muldoon/dev/ai/ai-livekit-agent/.venv/lib/python3.11/site-packages/livekit/plugins/openai/realtime/realtime_model.py", line 1341, in _handle_response_output_item_done
    fnc_call_info = _oai_api.create_ai_function_info(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paula.muldoon/dev/ai/ai-livekit-agent/.venv/lib/python3.11/site-packages/livekit/agents/llm/_oai_api.py", line 72, in create_ai_function_info
    sanitized_value = _sanitize_primitive(
                      ^^^^^^^^^^^^^^^^^^^^
  File "/Users/paula.muldoon/dev/ai/ai-livekit-agent/.venv/lib/python3.11/site-packages/livekit/agents/llm/_oai_api.py", line 148, in _sanitize_primitive
    raise ValueError(f"expected str, got {type(value)}")

ValueError: expected str, got <class 'NoneType'>
```

I've managed to print out these lines from `livekit/agents/llm/_oai_api.py`
line 50:
```

         arg_value = parsed_arguments[arg_info.name]
        print(f"arg info: {arg_info.name}")
[...]

PR #1211, opened 2024-12-11
Handle optional func args in tool calls when set to `None`

jayeshp19:
This PR fixes ValueError when optional function args are set to `None`

Related issues: 
https://github.com/livekit/agents/issues/1185
https://github.com/livekit/agents/issues/1208

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/_oai_api.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/_oai_api.py
@@ -64,13 +64,19 @@ def create_ai_function_info(
             inner_type = typing.get_args(arg_info.type)[0]
             sanitized_value = [
                 _sanitize_primitive(
-                    value=v, expected_type=inner_type, choices=arg_info.choices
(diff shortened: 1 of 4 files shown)


davidzhao (approved):
nice

davidzhao:
@jayeshp19 can you add a changeset?

jayeshp19:
It's already added,  not sure why It's not detecting..

jayeshp19:
retried and it detected

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/_oai_api.py
```diff
@@ -64,13 +64,19 @@ def create_ai_function_info(
             inner_type = typing.get_args(arg_info.type)[0]
```
mykola-mokhnach-parloa, line 64: fetching the first argument does not work for type definitions like `Union[None, str]` or `None | str`, where the None is passed as the first argument. The validation should verify if arguments lengths is between 1 and 2 and then take the first matched non-None-type argument.

PR #1212, opened 2024-12-11
improvements to endpointing latency

davidzhao:
it can take some time for final transcriptions to arrive. we would use end-of-speech timing to compute end of utterance delay, instead of from the moment final transcripts arrive.

also fixed the case where agents would be stuck when final transcripts do not end up coming in.

driveby, added a turn detector example

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -1109,11 +1109,10 @@ class _DeferredReplyValidation:
     PUNCTUATION = ".!?"
     PUNCTUATION_REDUCE_FACTOR = 0.75
 
-    LATE_TRANSCRIPT_TOLERANCE = 1.5  # late compared to end of speech
-
     # Long delay to use when the model thinks the user is still speaking
-    # TODO: make this configurable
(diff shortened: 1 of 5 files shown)


theomonnom (approved):

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -1109,12 +1109,11 @@ class _DeferredReplyValidation:
     PUNCTUATION = ".!?"
     PUNCTUATION_REDUCE_FACTOR = 0.75
 
-    LATE_TRANSCRIPT_TOLERANCE = 1.5  # late compared to end of speech
-
     # Long delay to use when the model thinks the user is still speaking
-    # TODO: make this configurable
     UNLIKELY_ENDPOINT_DELAY = 6
 
+    FINAL_TRANSCRIPT_TIMEOUT = 5
```
theomonnom, line 1115: 5 seems like a lot, what about 2?

davidzhao: STT could be slow sometimes, and not sending a final one is rare.. so trying to make this a last resort kind of thing

PR #1213, opened 2024-12-11
Add support for OpenAI image detail parameter, and add support for Anthropic image data URLs

bcherry:
Two vision-related improvements:

1. Allow the user to override the OpenAI "detail" parameter. While their docs suggest that "auto" is responsive to image sizing, in my testing it appears to not default into "low detail" mode for 512x512 images. Maybe it will for much smaller images? The only way to ensure this uses low detail mode (fixed budget of 85 tokens) is by specifying it.
2. Fill in the missing Anthropic plugin image url support. This one was trickier because Anthropic doesn't support image urls, it only supports base64 images in a special format. Since our ChatImage class doesn't have direct base64 image support, but does support URLs, the straightforward approach solution is to just use data URL anyways and transform it into Anthropic's desired format since it has all the necessary information already. The plugin still throws an error if you pass a non-data-URL (as OpenAI supports remote URLs but Anthropic does not).

Also I improved error messages to be much clearer they're coming from the LiveKit stack and not from the openai or anthropic APIs


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py
```diff
@@ -67,7 +67,7 @@ def _build_oai_image_content(image: llm.ChatImage, cache_key: Any):
     if isinstance(image.image, str):  # image url
         return {
             "type": "image_url",
-            "image_url": {"url": image.image, "detail": "auto"},
+            "image_url": {"url": image.image, "detail": image.inference_detail},
```
theomonnom, line 70: This should be done automatically depending on the inference_width & inference_height. 

Tho we don't have tests covering this today so it may not work as expected

bcherry: I'm not so sure...  there are legitimate use-cases for high detail on small images, which on a 512x512 image will use 255 tokens vs 85 tokens for low detail, and gives better results.  Low detail on large images is less useful of course..., and you should probably resize it to 512x512 yourself first to improve inference speed.

this class as it is is a little messy, since inference_width/height are only used for video frames and ignored for URLs, and the actual API capabilities in the underlying models vary quite a bit (I'm not aware of any other provider with the "detail" parameter, for instance).

bcherry:
@theomonnom updated to expand the scope, but let me know if you think we should keep iterating on the detail param...

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py
theomonnom: From my understanding, putting low basically resize the image to 512x512. So having an input image of this size is the same as detail "low"
https://platform.openai.com/docs/guides/vision#low-or-high-fidelity-image-understanding

theomonnom: > high will enable "high res" mode, which first allows the model to first see the low res image (using 85 tokens) and then creates detailed crops using 170 tokens for each 512px x 512px tile.

Ah ok I guess the behaviour is different, then it lgtm

theomonnom (approved):
Feel free to ignore but if you have time it would be great to add tests for images.
Otherwise lgtm (just fix CI)

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py
bcherry: yeah i assumed the same based on the docs but it isn't what i observe in testing. while it is true that passing "low" will resize your image to 512x512, the docs don't explicitly say the reverse is true. it only says 

> By default, the model will use the auto setting which will look at the image input size and decide if it should use the low or high setting.

That doesn't specify where the threshold to flip to "low" is, and my testing suggests this threshold is much lower than 512x512. in "high" mode each 512x512 tile gets a 170tk budget on top of the whole image's 85tk budget for the 512x512 thumbnail, which does imply that using "high" on a 512x512 image has value and provides deeper understanding than using "low".

tests/hearts.rgba
theomonnom, line 1: nice! make sure to put this file to git LFS

tests/test_llm.py
```diff
@@ -369,3 +372,82 @@ async def _request_fnc_call(
...
+
+_HEARTS_JPEG_PATH = Path(__file__).parent / "hearts.jpg"
+with open(_HEARTS_JPEG_PATH, "rb") as f:
+    _HEARTS_IMAGE_DATA_URL = (
+        f"data:image/jpeg;base64,{base64.b64encode(f.read()).decode()}"
+    )
+
+
+@pytest.mark.parametrize("llm_factory", LLMS)
+async def test_chat_with_image_data_url(llm_factory: Callable[[], llm.LLM]):
```
theomonnom, line 393: Let's limit this test to OAI? I think they're the only one to support it

bcherry: i modified the anthropic plugin to support these (by parsing the URL and reconstructing it as an object) and it passes the tests. not sure what will happen with vertex, didn't check that one locally so will watch in CI

theomonnom: TIL Anthropic supports images

theomonnom (approved):
Nice awesome, definitely a great improvements to image handling

tests/hearts.jpg
nbsp, line 1: it's a small file but still bumps up the git repo size by quite a bit, maybe lfs this file as well?

nbsp (approved):

tests/hearts.jpg
bcherry: I made this change. I wonder if I should try to clean up the copies that went into repo history when I originally pushed them? it's involved but is easier now than once it hits main

nbsp: i can take a crack at rebasing them out, as the resident git witch

nbsp: though, since we're squashing this and deleting the workbranch right after, i don't think it actually reaches anyone's tree unless they pulled already, in which case they have it in there already. so it doesn't matter

bcherry: thank you! i think there may be tests/hearts.jpg, tests/hearts.rgba, and a short-lived tests/hearts.bmp in there.

bcherry: 🤷🏻 will trust you on this one

nbsp: will go through this just in case.

nbsp: should be good now. ignore the extra commits not from you, they're there from the merge you did. they'll get squashed out when this PR is merged.

davidzhao, line 1: I don't think this file took? might have to delete it, commit.. then re-add, then commit again

davidzhao (approved):
lgtm! nit on git lfs image

tests/hearts.jpg
bcherry: I think it is in LFS, it's just so small that it's rendered anyways. If you click ... -> View File you'll see it 
image

davidzhao: ahh i see. coo

bcherry:
@theomonnom @nbsp this branch keeps failing CI on seemingly unrelated failures. I've gotten it to pass a couple times by rerunning. is this known in the repo or should I dig into to whether it really is coming from here?

theomonnom:
yes this is ffmpeg installation on macos failing, feel free to ignore

PR #1221, opened 2024-12-12
fix: correctly parse function argument types

jayeshp19:


tests/test_build_func_desc.py
```diff
@@ -0,0 +1,51 @@
...
+        name="foo",
+        description="foo",
+        auto_retry=False,
+        callable=lambda: None,
+        arguments={
+            "arg": FunctionArgInfo(
+                name="foo",
+                description="foo",
+                type=inner_type,
+                is_optional=is_optional,
```
theomonnom, line 40: type should be arg_typ directly for the test to be accurate?

theomonnom:
let's make sure this work on python 3.9 before merging. I know the behaviours of Optional changed between python versions (With the new A | B syntax)

livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
```diff
@@ -560,24 +562,26 @@ def type2str(t: type) -> str:
             raise ValueError(f"unsupported type {t} for ai_property")
 
         p: dict[str, Any] = {}
-        if arg_info.default is inspect.Parameter.empty:
+        if not _is_optional_type(arg_info.type)[0]:
```
theomonnom, line 565: I think the logic is different?
Does it still accept default values?  `my_var: int = 5 `

theomonnom: Can we add a test for it as well? (for default values)

jayeshp19: we should count it as optional right?

theomonnom: Yes, when sending it to OAI, they should not be inside the "`required`" field. Tho the type itself inside the Python code isn't really "optional"

tests/test_create_func.py
```diff
@@ -0,0 +1,209 @@
...
+
+    assert arg_info.name == "param2"
+    assert arg_info.description == ""
+    assert arg_info.type is Optional[List[str]]
+    assert arg_info.default is None
+    assert arg_info.choices == ()
+    assert build_arg_info["type"] == "array"
+    assert build_arg_info["items"]["type"] == "string"
+
+    # check 'param3'
```
jayeshp19, line 119: @theomonnom Added test case with default param

theomonnom (approved):

PR #1236, opened 2024-12-15
streaming audio decoder, enables receiving compressed audio from TTS services

davidzhao:
also moved it to an executor, to avoid decoding while blocking async runloop

### DIFF
--- a/livekit-agents/livekit/agents/utils/codecs/decoder.py
+++ b/livekit-agents/livekit/agents/utils/codecs/decoder.py
@@ -1,0 +1,159 @@
+# Copyright 2024 LiveKit, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


theomonnom (approved):
awesome!

livekit-agents/livekit/agents/utils/codecs/decoder.py
```diff
@@ -0,0 +1,159 @@
...
+        self._started = False
+        self._output_finished = False
+        self._input_buf = StreamBuffer()
+        self._loop = asyncio.get_event_loop()
+
+    async def push(self, chunk: bytes):
+        self._input_buf.write(chunk)
+        if not self._started:
+            self._started = True
+            self._loop.run_in_executor(None, self._decode_loop)
```
theomonnom, line 111: maybe push shouldn't be async?

theomonnom:
Thanks for fixing the CI!

Issue #1238, opened 2024-12-16
Choppy/Duplicate audio on LLM timeout


martin-purplefish:
When the LLM times out after having streamed some tokens already, we end up with two audio streams playing at the same time, and all the text getting added to the same message.

It's basically:
* The LLM starts streaming some tokens
* Audio starts playing
* The LLM times out so retry kicks off
* Original audio keeps playing, and the new audio stream from retry also starts playing

The end result is a single agent message that's very confused (since it's coming from the same LLMStream), and two audio streams playing at the same time, leading to choppy audio that skips around.

In terms of how to fix this.

1. We should cancel the speech if the LLM times out after some tokens have been streamed instead of retrying.
2. If the LLM times out after some tokens have been played, treat it as an interruption, and retry the entire call to `llm.chat()` - including the call to `before_llm_cb` and `before_tts_cb`

This is what was really happening in #1077  - see that ticket for a detailed example.

Here's another person encountering it in the slack: https://livekit-users.slack.com/archives/C07FY8WHGPM/p1734179180491739

davidzhao:
#1238 addresses part of this. keeping this issue open as the pipeline agent needs to restart the inference on these unrecoverable errors

PR #1239, opened 2024-12-16
Do not retry on already streamed tokens

martin-purplefish:
Part 1 of #1238

This should address timeouts causing duplicate audio, however I believe the behavior will be the LLM gets stuck and just goes silent.

### DIFF
--- a/.changeset/dirty-jeans-burn.md
+++ b/.changeset/dirty-jeans-burn.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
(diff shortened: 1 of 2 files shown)


martin-purplefish:
PTAL @theomonnom @davidzhao

this is a hard one to reproduce, but I think this fixes the duplicate issue (and the agent will just stop talking on a timeout).

The next step is to have the pipeline agent catch this somehow, and retry the llm.chat call.

davidzhao:
will handle this in #1298

PR #1240, opened 2024-12-16
Support Gemini Live API

jayeshp19:
- Remove openai dependencies from Openai base class
- Gemini Live API support


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1262,7 +1254,7 @@ def _handle_conversation_item_input_audio_transcription_completed(
         transcript = transcription_completed["transcript"]
         self.emit(
             "input_speech_transcription_completed",
-            InputTranscriptionCompleted(
+            Transcription(
```
davidzhao, line 1257: `Transcription` might be too generic, should we name it `InputTranscription`?

davidzhao (commented):
nice work!

does Gemini return user's input transcription? I didn't see we had any handling for it.

it'd be great to include docstrings and comments 🙏

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
```diff
@@ -265,11 +340,14 @@ def _input_speech_started():
             if self._playing_handle is not None and not self._playing_handle.done():
                 self._playing_handle.interrupt()
 
-                self._session.conversation.item.truncate(
-                    item_id=self._playing_handle.item_id,
-                    content_index=self._playing_handle.content_index,
-                    audio_end_ms=int(self._playing_handle.audio_samples / 24000 * 1000),
-                )
+                if self._session.supports_conversation_manipulation():
+                    self._session._truncate_conversation_item(
```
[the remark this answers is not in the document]
jayeshp19: It notifies with server event

livekit-plugins/livekit-plugins-google/livekit/plugins/google/multimodal/live_api.py
```diff
@@ -0,0 +1,353 @@
...
+            project=self._opts.project,
+            location=self._opts.location,
+        )
+        self._main_atask = asyncio.create_task(
+            self._main_task(), name="gemini-realtime-session"
+        )
+        # dummy task to wait for the session to be initialized # TODO: sync chat ctx
+        self._init_sync_task = asyncio.create_task(
+            asyncio.sleep(0), name="gemini-realtime-session-init"
+        )
```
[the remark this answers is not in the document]
jayeshp19: Yes, it supports text inputs aswell, we can send chat_ctx as initial input

livekit-plugins/livekit-plugins-google/livekit/plugins/google/multimodal/live_api.py
```diff
@@ -0,0 +1,353 @@
...
+                                        data=part.inline_data.data,
+                                        sample_rate=24000,
+                                        num_channels=1,
+                                        samples_per_channel=len(part.inline_data.data)
+                                        // 2,
+                                    )
+                                    content.audio.append(frame)
+                                    content.audio_stream.send_nowait(frame)
+                        if response.server_content.interrupted:
+                            self.emit("input_speech_started")
```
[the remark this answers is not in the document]
jayeshp19: Yes VAD is handled by Gemini itself, there's no clear way to detect that user has started the speech. probably when response turn is completed from server , it goes into listening state.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/multimodal/live_api.py
```diff
@@ -0,0 +1,353 @@
...
+        chat_ctx: llm.ChatContext | None = None,
+        fnc_ctx: llm.FunctionContext | None = None,
+    ) -> MultimodalSession:
+        session = GeminiRealtimeSession(
+            opts=self._opts,
+            chat_ctx=chat_ctx or llm.ChatContext(),
+            fnc_ctx=fnc_ctx,
+            loop=self._loop,
+        )
+        self._rt_sessions.append(session)
```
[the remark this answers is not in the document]
jayeshp19: as per my understanding, we are maintaining list of sessions because we could've multiple sessions running, they will get closed at once whenever `aclose` is called. @theomonnom correct me if I'm wrong

jayeshp19:
> nice work!
> 
> does Gemini return user's input transcription? I didn't see we had any handling for it.
> 
> it'd be great to include docstrings and comments 🙏

Gemin does not returns transcription as of now, but I was thinking maybe we could add our stt service to maintain chat_ctx

livekit-plugins/livekit-plugins-google/livekit/plugins/google/multimodal/live_api.py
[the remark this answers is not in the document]
theomonnom: it is a WeakSet

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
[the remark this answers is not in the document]
davidzhao: right, but when interrupted, it needs to know how much of the speech was played back? how do we notify them that the speech was truncated

livekit-plugins/livekit-plugins-google/livekit/plugins/google/multimodal/live_api.py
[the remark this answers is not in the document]
davidzhao: got it. what about modifying the context after it's started? i.e. synchronizing context?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/multimodal/live_api.py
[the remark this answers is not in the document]
davidzhao: if a user is relying on `user_speech_started` event, this means that they won't be able to accurately detect it right? worth noting that difference in the comments.. this is something that we can address in the future (i.e. by still running VAD ourselves)

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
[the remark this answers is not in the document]
jayeshp19: I don't think gemini provides any way to do this.

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
```diff
@@ -25,9 +25,66 @@
...
+    text: str
+    audio: list[rtc.AudioFrame]
+    text_stream: AsyncIterable[str]
+    audio_stream: AsyncIterable[rtc.AudioFrame]
+    content_type: Literal["text", "audio"]
+
+
+@dataclass
+class Capabilities:
+    supports_chat_ctx_manipulation: bool = False
```
theomonnom, line 57: Can we also use Protocol for the fields?
(E.g, ideally OpenAI code should change)

jayeshp19: Updated code as per suggestions
- now there're only few changes are in openai subclass, I've merged InputTranscriptionCompleted and InputTranscriptionFailed into one class, 
- removed toolcalls from RealtimeContent, I coundn't find we  use that anywhere, I also couldn't find use of accumalted audio or text, should I remove them??
- moved  methods _handle_response_content_done and _truncate_conversation_item to openai, because it's relavent there

examples/realtime/gemini_agent.py
davidzhao, line 1: could we change this folder to `multimodal_agent` for now ?

sorry, this is because currently all of our docs still reference MultimodalAgent.. with v1 we could change it

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -105,8 +90,21 @@ class RealtimeToolCall:
     """id of the tool call"""
 
 
-# TODO(theomonnom): add the content type directly inside RealtimeContent?
-# text/audio/transcript?
+@dataclass
+class Capabilities:
+    supports_chat_ctx_manipulation: bool
```
davidzhao, line 95: looking at what this is used for, it seems like it's only used for messages truncation? should this just be: `supports_truncate`?

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
```diff
@@ -248,9 +267,11 @@ def _input_speech_transcription_completed(
             user_msg = ChatMessage.create(
                 text=ev.transcript, role="user", id=ev.item_id
             )
-            self._session._update_conversation_item_content(
-                ev.item_id, user_msg.content
-            )
+
+            if self._model.capabilities.supports_chat_ctx_manipulation:
+                self._session._update_conversation_item_content(
```
davidzhao, line 272: `_update_conversation_item_content` isn't making any network calls, why does it need to be guarded with `supports_chat_ctx_manipulation`

jayeshp19: I thought it's unrelated to gemini, because we are not maintaining transcript or chatctx, maybe we should move `_RemoteConversationItems` class to baseclass?

wdyt @theomonnom

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
theomonnom, line 95: I thinks since this is temporary, we shouldn't care about implementing Capabilities today? wdyt?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -41,22 +42,6 @@
...
-
-
-@dataclass
-class InputTranscriptionFailed:
-    item_id: str
-    """id of the item"""
-    message: str
-    """error message"""
-
-
```
theomonnom, line 59: Let's keep tne 1 on 1  mapping with the OAI API

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -125,8 +123,6 @@ class RealtimeContent:
     """stream of text content"""
     audio_stream: AsyncIterable[rtc.AudioFrame]
     """stream of audio content"""
-    tool_calls: list[RealtimeToolCall]
-    """pending tool calls"""
```
theomonnom, line 129: The reason why it's there is because RealtimeAPI is a 1/1 mapping to the OAI ws API

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1023,6 +1031,38 @@ def _recover_from_text_response(self, item_id: str | None = None) -> None:
...
+                extra={
+                    "item_id": message.item_id,
+                    "text": message.text,
+                    "retries": self._text_response_retries,
+                },
+            )
+            self._recover_from_text_response(message.item_id)
+        else:
+            self._text_response_retries = 0
+
```
theomonnom, line 1065: I'm not sure about this, should we create a separate PR to discuss about it?

theomonnom (approved):

livekit-plugins/livekit-plugins-google/livekit/plugins/google/log.py
```diff
@@ -1,3 +1,6 @@
 import logging
 
 logger = logging.getLogger("livekit.plugins.google")
+
+
+logging.getLogger("websockets.client").setLevel(logging.INFO)
```
theomonnom, line 6:
```suggestion
```

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
jayeshp19: Should I kepp `_truncate_conversation_item` it's similar to _push_audio

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
jayeshp19: How do we handle truncation in baseclass then?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/log.py
jayeshp19: This is to suppress verbose websocket logs

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1662,13 +1685,12 @@ async def _run_fnc_task(self, fnc_call_info: llm.FunctionCallInfo, item_id: str)
...
-                previous_item_id=item_id,
-            )
-            await self.response.create(on_duplicate="keep_both")
-            await create_fut
+        create_fut = self.conversation.item.create(
+            tool_call,
+            previous_item_id=item_id,
+        )
+        await self.response.create(on_duplicate="keep_both")
+        await create_fut
```
theomonnom, line 1693: any reason to change this logic, this is actually an expected thing where we don't want to send results to the LLM if the aifunction returns None

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
theomonnom: both OpenAI and Google doesn't use it right now?
`supports_chat_ctx_manipulation=False`

The motivation here is that I want to avoid adding new APIs that may be hard to remove in the future as it'll be a breaking change

livekit-plugins/livekit-plugins-google/livekit/plugins/google/log.py
theomonnom: This is too opiniated to be inside this place, you can add it here instead https://github.com/livekit/agents/blob/32383939025453616d23c82ccc8c725ee87a9870/livekit-agents/livekit/agents/cli/log.py#L15

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
```diff
@@ -28,6 +28,57 @@
...
+        chat_ctx: llm.ChatContext | None = None,
+        fnc_ctx: llm.FunctionContext | None = None,
+    ) -> RealtimeAPISession:
+        """
+        Create a new realtime session with the given chat and function contexts.
+        """
+        pass
+
+
+class RealtimeAPISession(Protocol):
```
theomonnom, line 72: Let's put all the protocol "private" (prefix by _)

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
jayeshp19: When Function returns exception, `called_fnc.result` is `None`. and we are adding exception in content. maybe we should check `if tool_call.content is not None:` I'll update it

theomonnom: I don't think we want to do that, if there is an exception, the LLM should know

jayeshp19: Yeah we will send it through content, like we do in pipeline

theomonnom: ah ok makes sense

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -284,6 +288,9 @@ def __init__(
             ValueError: If the API key is not provided and cannot be found in environment variables.
         """
         super().__init__()
+        self._capabilities = Capabilities(
+            supports_chat_ctx_manipulation=False,
```
theomonnom, line 292: re (from Slack): this isn't used on OpenAI

jayeshp19: oh wait, It should be True for Openai, because we need to few operations such as truncate, update conversations items etc.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
theomonnom: @jayeshp19

theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
gianpaj: i hit this issue as well. See the 2nd error here https://github.com/livekit/agents/issues/1329

Issue #1244, opened 2024-12-17
Chat message with None content breaks chat history sync to realtime model


longcw:
In `RealtimeSession.set_chat_ctx` the local messages are inserted to the realtime model based on the previous item id. If there is any empty messages in the list, `ConversationItem.create` will return without actually create the item then the previous id is missing in the realtime model.

The empty message can from the mock empty user audio from `_create_empty_user_audio_message` that without a transcript.

PR #1245, opened 2024-12-17
fix: filter out empty message for set chat ctx in realtime model

longcw:
For https://github.com/livekit/agents/issues/1244


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -506,10 +506,6 @@ def create(
 
             message_content = message.content
             tool_call_id = message.tool_call_id
-            if not tool_call_id and message_content is None:
```
longcw, line 509: There is a check below with proper warning log.

theomonnom (approved):

PR #1249, opened 2024-12-17
Add JPEG quality param to image encoder

bcherry:
The default through PIL appears to be 75 but it makes sense to give users explicit control.

While I was at it I added docstrings.


davidzhao (approved):

davidzhao:
some type issues.. need to use `Optional[type]` instead

theomonnom (approved):
thanks!

PR #1250, opened 2024-12-17
updated default realtime model to gpt-4o-realtime-preview-2024-12-17

davidzhao:
also improved weather example to highlight two methods of feedback

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/api_proto.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/api_proto.py
@@ -155,6 +155,12 @@ class InputTokenDetails(TypedDict):
     cached_tokens: int
     text_tokens: int
     audio_tokens: int
+    cached_tokens_details: CachedTokenDetails
+
+
+class CachedTokenDetails(TypedDict):
+    text_tokens: int
+    audio_tokens: int
 
 
 class OutputTokenDetails(TypedDict):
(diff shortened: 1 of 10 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -38,6 +38,11 @@
     "metrics_collected",
 ]
 
+OpenAIModel = Literal[
+    "gpt-4o-realtime-preview-2024-12-17", "gpt-4o-mini-realtime-preview-2024-12-17"
```
bcherry, line 42: this should also allow the unversioned ones. there's actually six valid values

image

davidzhao: good point. the default should point to unversioned

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/api_proto.py
```diff
@@ -27,6 +27,16 @@ class FunctionToolChoice(TypedDict):
     "in_progress", "completed", "incomplete", "cancelled", "failed"
 ]
 
+# https://platform.openai.com/docs/models/gp#gpt-4o-realtime
+OpenAIModel = Literal[
```
bcherry, line 31: should probably be called `Model` to match the other types here

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -197,7 +192,7 @@ def __init__(
         *,
         instructions: str = "",
         modalities: list[api_proto.Modality] = ["text", "audio"],
-        model: OpenAIModel | str = DefaultOpenAIModel,
+        model: api_proto.OpenAIModel | str = api_proto.DefaultOpenAIModel,
```
bcherry, line 195: the other defaults are in this file as UPPER_CAMEL_CASE, rather than in api_proto

bcherry (approved):

theomonnom (approved):

PR #1253, opened 2024-12-18
fix: add session_updated event for realtime model

longcw:
Emit `session_updated` event from the `RealtimeSession` for anyone want to verify the session is correctly updated. Related to https://github.com/livekit/agents/issues/1183


theomonnom (approved):
lg!

PR #1254, opened 2024-12-18
fix: avoid duplicated chat ctx for function call with messages

longcw:
When there is a message along with the function calls, the message will be added to the chat ctx twice: after it's played, and after the function calls finished.
```python
[
...
 ChatMessage(role='assistant',
             id='item_1158d15105e5',
             name=None,
             content=' Hello from the weather station. Would you like to know '
                     'the weather? If so, tell me your location.',
             tool_calls=None,
             tool_call_id=None,
             tool_exception=None),
 ChatMessage(role='user',
             id='item_bf51c7ec1c80',
             name=None,
             content='New York.',
             tool_calls=None,
             tool_call_id=None,
             tool_exception=None),
 ChatMessage(role='assistant',
             id='item_d41fa35f7251',
             name=None,
             content=' Let me check the weather in New York for you.',
             tool_calls=None,
             tool_call_id=None,
             tool_exception=None),
 ChatMessage(role='assistant',
             id='item_cb041bd44d6c',
             name=None,
             content=' Let me check the weather in New York for you.',
             tool_calls=[FunctionCallInfo(tool_call_id='call_nuu9ElDEKiU3FgShMJCshaly',
      ...
             tool_exception=None),
 ChatMessage(role='tool',
             id='item_6dabd488c4a1',
             name='get_weather',
             content='The weather in New York is Clear +33°F.',
             tool_calls=None,
             tool_call_id='call_nuu9ElDEKiU3FgShMJCshaly',
             tool_exception=None),
...
```
I have tested that remove the text for `ChatMessage.create_tool_calls(tool_calls_info, text=collected_text)` won't affect the follow up answers.

### DIFF
--- a/.changeset/curvy-knives-promise.md
+++ b/.changeset/curvy-knives-promise.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+avoid duplicated chat ctx for function calls with messages
--- a/examples/voice-pipeline-agent/function_calling_weather.py
+++ b/examples/voice-pipeline-agent/function_calling_weather.py
@@ -67,8 +67,8 @@ class AssistantFnc(llm.FunctionContext):
                         f"Failed to get weather data, status code: {response.status}"
                     )
 
-        # To wait for the speech to finish before giving results of the function call
-        await speech_handle.join()
+        # (optional) To wait for the speech to finish before giving results of the function call
+        # await speech_handle.join()
         return weather_data
 
 
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -801,12 +801,20 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
             speech_handle.source.function_calls
         )
 
+        message_id_committed: str | None = None
         if (
             collected_text
             and speech_handle.add_to_chat_ctx
             and (not user_question or speech_handle.user_committed)
         ):
             if speech_handle.extra_tools_messages:
+                msgs = self._chat_ctx.messages
+                if msgs and msgs[-1].id == speech_handle.fnc_text_message_id:
+                    # remove text message alongside function calls if it's the last in the ctx
+                    msgs.pop()
+                elif speech_handle.extra_tools_messages[0].tool_calls:
+                    # remove the content of the tool call message
+                    speech_handle.extra_tools_messages[0].content = ""
                 self._chat_ctx.messages.extend(speech_handle.extra_tools_messages)
 
             if interrupted:
@@ -814,7 +822,7 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
 
             msg = ChatMessage.create(text=collected_text, role="assistant")
             self._chat_ctx.messages.append(msg)
-
+            message_id_committed = msg.id
             speech_handle.mark_speech_committed()
 
             if interrupted:
@@ -904,7 +912,9 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
                 return
 
             # create a nested speech handle
-            extra_tools_messages = [ChatMessage.create_tool_calls(tool_calls_info)]
+            extra_tools_messages = [
+                ChatMessage.create_tool_calls(tool_calls_info, text=collected_text)
+            ]
             extra_tools_messages.extend(tool_calls_results)
 
             new_speech_handle = SpeechHandle.create_tool_speech(
@@ -912,6 +922,7 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
                 add_to_chat_ctx=speech_handle.add_to_chat_ctx,
                 extra_tools_messages=extra_tools_messages,
                 fnc_nested_depth=speech_handle.fnc_nested_depth + 1,
+                fnc_text_message_id=message_id_committed,
             )
 
             # synthesize the tool speech with the chat ctx from llm_stream
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -904,9 +904,7 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
                 return
 
             # create a nested speech handle
-            extra_tools_messages = [
-                ChatMessage.create_tool_calls(tool_calls_info)
-            ]
+            extra_tools_messages = [ChatMessage.create_tool_calls(tool_calls_info)]
             extra_tools_messages.extend(tool_calls_results)
 
             new_speech_handle = SpeechHandle.create_tool_speech(
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -905,7 +905,7 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
 
             # create a nested speech handle
             extra_tools_messages = [
-                ChatMessage.create_tool_calls(tool_calls_info, text=collected_text)
+                ChatMessage.create_tool_calls(tool_calls_info)
             ]
             extra_tools_messages.extend(tool_calls_results)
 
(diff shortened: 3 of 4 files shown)


davidzhao (approved):
lgtm!

jmugicagonz:
I think there is an issue with this solution. Now the text won't be duplicated, but what originally the LLM sent in one call (text + function call) will be moved to two messages. Me, as a user, want the LLM to both send some text and call the function in the same message. If the chat context is modified, the LLM ends just sending the text and waiting for user response to call the function (undesired behavior).
I'm not sure what the fix is as I'm pretty new to this code, but I'd say we should bring the text and the function call into the same message.

jmugicagonz:
`            # Remove any existing message with the same text as collected_text
            self._chat_ctx.messages = [
                msg for msg in self._chat_ctx.messages if msg.content != collected_text
            ]

            extra_tools_messages = [
                ChatMessage.create_tool_calls(tool_calls_info, text=collected_text)
            ]
            extra_tools_messages.extend(tool_calls_results)`

I tried this code on my end and it respects better the chat history. Is there a cleaner way of doing it?

longcw:
I had the same concern that this may teach LLM to not put a message with the function calls, during my testing if I put `when performing function calls, let user know that you are checking the weather` in prompt, it can always response a message after the change (in this example https://github.com/livekit/agents/blob/main/examples/voice-pipeline-agent/function_calling_weather.py#L90). But this may be different to different prompts.

For the option of only add the first chat ctx if interrupted, now we can say anything inside a function call, so if we only add the message after function call the order is different to what user heard:
```
actual: message from LLM, something said in fnc, fnc answer
chat_ctx: something said in fnc, message from LLM, fnc answer
```

@jmugicagonz Could you give me an example of system prompt of this change affects the LLM behavior. I'll test with that and mark this pr as a draft. We may need more discussion on how to fix it or if it's a real issue.

jmugicagonz:
Thanks for your answer Long. I'd say that you don't always want the LLM to tell the user 'i'm checking the weather' as it can feel repetitive for longer conversations and with other functions. Moreover, a longer prompt may make that information to be missed. You could maybe try with this prompt:

"You are a weather assistant created by LiveKit. Your interface with users will be voice. "
"You will provide weather information for a given location. "
"When a user asks for the weather information in a given location, you will say 'thanks for asking! I'll get you the weather information for that location' and then you will inmediately call the function 'get_weather' to retrieve the weather information."

Anyways, I can confirm you that the behavior 'this may teach LLM to not put a message with the function calls' definetly happens, I have extensively tested it on my end.

I'm not sure I understand what you mean with the problem of 'only add the first chat ctx if interrupted'. Now, a message with content and a function call will first append the content to the chat context and then later on the function call, when this finishes. If you just wait the llm to end the function call, and you append both the content and the function call together, how does that change the order? (it's totally possible, I just don't see why)

longcw:
> I'm not sure I understand what you mean with the problem of 'only add the first chat ctx if interrupted'. Now, a message with content and a function call will first append the content to the chat context and then later on the function call, when this finishes. If you just wait the llm to end the function call, and you append both the content and the function call together, how does that change the order? (it's totally possible, I just don't see why)

The problem is that it's possible to use `agent.say` to say something in a function call. The message from the fnc_call is played first, then the `say` inside the fnc_call, and last the answer of the fnc_call.

The order user heard is `1. message from LLM, 2. something said in fnc, 3. fnc answer`.

If the agent don't append the message from LLM first, or remove it after the function call, the order in the chat ctx will be
`1. something said in fnc, 2. message from LLM,  3. fnc answer`.

longcw:
If the content for a chat message with function call is required, I would suggest to keep it as is. also this behavior to have the message saved twice is actually there a long time ago. It should be ok if it works in the past.

jmugicagonz:
> > I'm not sure I understand what you mean with the problem of 'only add the first chat ctx if interrupted'. Now, a message with content and a function call will first append the content to the chat context and then later on the function call, when this finishes. If you just wait the llm to end the function call, and you append both the content and the function call together, how does that change the order? (it's totally possible, I just don't see why)
> 
> The problem is that it's possible to use `agent.say` to say something in a function call. The message from the fnc_call is played first, then the `say` inside the fnc_call, and last the answer of the fnc_call.
> 
> The order user heard is `1. message from LLM, 2. something said in fnc, 3. fnc answer`.
> 
> If the agent don't append the message from LLM first, or remove it after the function call, the order in the chat ctx will be `1. something said in fnc, 2. message from LLM, 3. fnc answer`.

I get now what you mean, thanks for the explanation @longcw .
In that case, I would create a single message, where the content would be "message from LLM" + "something said in fnc" and the function call. And then another message with the "fnc answer". Is that possible?

jmugicagonz:
I also think that saying something inside a function, even if it helps the user understand what is going on, and smooth the waiting time, will confuse the LLM. The LLM will see in its history messages that weren't in the instructions, changing its behavior. This is difficult to see in short interactions with small basic prompts. Our testing shows it in longer interactions with more complex prompts.

jmugicagonz:
I have added an alternative proposal in this [PR](https://github.com/livekit/agents/pull/1262)

longcw:
Implemented the proposal in https://github.com/livekit/agents/pull/1262#issuecomment-2554514428:
1. we add it to the chat ctx before the fnc call, and record the message id
2. after function finished, if the last item in the chat ctx is still the message recorded in step 1, we replace it with the message containing the content and the function call. If the last item in the chat ctx has changed, i.e. the user called .say in the function call, we remove the content of the function call and add it to the ctx.

cc @jmugicagonz  @davidzhao 

The example of two cases:
1. prompt LLM to say something with fnc
```python
[
...
 ChatMessage(role='assistant',
             id='item_f4a0fc5a4a02',
             name=None,
             content=' Hello from the weather station. Would you like to know the weather? If so, tell me your location.',
             tool_calls=None,
             tool_call_id=None,
             tool_exception=None),
 ChatMessage(role='user',
             id='item_c4102c5a3861',
             name=None,
             content='New York.',
             tool_calls=None,
             tool_call_id=None,
             tool_exception=None),
 ChatMessage(role='assistant',
             id='item_9b39bcb41ff5',
             name=None,
             content=' Let me check the weather for New York. Please hold on for a moment.',
             tool_calls=[FunctionCallInfo(tool_call_id='call_KCAuyRssGQpRMFFuc2seyEXL',
...
                                          arguments={'location': 'New York'})],
             tool_call_id=None,
             tool_exception=None),
 ChatMessage(role='tool',
             id='item_10aae8cc82af',
             name='get_weather',
             content='The weather in New York is Sunny +39°F.',
             tool_calls=None,
             tool_call_id='call_KCAuyRssGQpRMFFuc2seyEXL',
             tool_exception=None),
 ChatMessage(role='assistant',
             id='item_0b0604788ec6',
             name=None,
             content=' The weather in New York is currently sunny with a '
                     'temperature of 39°F.',
             tool_calls=None,
             tool_call_id=None,
             tool_exception=None)]
```

2. prompt LLM to **not say** anything but call `.say` inside the function call
```python
[
...
 ChatMessage(role='user',
             id='item_68e0bce3f5bb',
             name=None,
             content='New York.',
             tool_calls=None,
             tool_call_id=None,
             tool_exception=None),
 ChatMessage(role='assistant',
             id='item_4c3f981dd320',
             name=None,
             content=' The current weather in New York is',  # from `say`
             tool_calls=None,
             tool_call_id=None,
             tool_exception=None),
 ChatMessage(role='assistant',
             id='item_006f08b148d7',
             name=None,
             content='',
             tool_calls=[FunctionCallInfo(tool_call_id='call_Rs5RsUZnfTBb5OVSYdoDzv7o',
 ...
             tool_call_id=None,
             tool_exception=None),
 ChatMessage(role='tool',
             id='item_875357725db7',
             name='get_weather',
             content='The weather in New York is Sunny +39°F.',
             tool_calls=None,
             tool_call_id='call_Rs5RsUZnfTBb5OVSYdoDzv7o',
             tool_exception=None),
 ChatMessage(role='assistant',
             id='item_41910771d609',
             name=None,
             content=' Sunny with a temperature of 39°F.',
             tool_calls=None,
             tool_call_id=None,
             tool_exception=None)]
```

Indeed, I also noticed that if we do both: prompt the LLM to return a message with the function call, and call `say` inside the function without adding the LLM message to the function call ctx, the LLM tends to just return the message without the function call and call the function in the next response. But this issue can be fixed by simply adding an explicit requirement for calling the function like `Call the function to get the weather.` 

The prompt I used for testing
```
            "You are a weather assistant created by LiveKit. Your interface with users will be voice. "
            "You will provide weather information for a given location. "
            # when using option 1, you can suppress from the agent with prompt
            # "do not say anything while waiting for the function call to complete."
            # uncomment this to use option 2
            "when performing function calls, let user know that you are checking the weather."
            "Call the function to get the weather."
```

The behaviors of LLM can vary depends on the prompt and length of the context. Users can choose different options based on needs

davidzhao (approved):
this approach makes sense to me.

PR #1257, opened 2024-12-19
use plain onnxruntime for turn detector, remove pytorch

jeradf:


livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/eou.py
```diff
@@ -13,10 +13,21 @@
 from .log import logger
 
 HG_MODEL = "livekit/turn-detector"
+ONNX_FILENAME = "model_quantized.onnx"
 PUNCS = string.punctuation.replace("'", "")
 MAX_HISTORY = 4
 
 
+def _download_from_hf_hub(repo_id, filename):
```
davidzhao, line 21: shouldn't this be in download_files? we don't want to download during runtime

jeradf: should be fixed now

theomonnom (commented):
Awesome!

I'm wondering if we need similar settings to what we do for silero.VAD
https://github.com/livekit/agents/blob/e32278b9c31f62f56225382ad636b508f5948d52/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/onnx_model.py#L24

This disallow CPU spinning, when the inference isn't running, do you still see high CPU usage on your machine, may be due to spinning

PR #1260, opened 2024-12-19
Move `create_ai_function_info` to function_context.py

jayeshp19:
- Moved `create_ai_function_info` to function_context.py for better reusability and reduce repetation
- Fixes `UnboundLocalError` in `test_stt.py::test_stream[24000-assemblyai]` by ensuring `ws` is initialized before access.

### DIFF
--- a/livekit-agents/livekit/agents/llm/function_context.py
+++ b/livekit-agents/livekit/agents/llm/function_context.py
@@ -18,6 +18,7 @@ import asyncio
 import enum
 import functools
 import inspect
+import json
 import types
 import typing
 from dataclasses import dataclass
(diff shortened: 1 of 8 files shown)


livekit-agents/livekit/agents/llm/function_context.py
```diff
@@ -303,3 +304,96 @@ def _is_optional_type(typ) -> Tuple[bool, Any]:
...
     return False, None
+
+
+def create_ai_function_info(
+    fnc_ctx: FunctionContext,
+    tool_call_id: str,
+    fnc_name: str,
+    raw_arguments: str,  # JSON string
+) -> FunctionCallInfo:
+    if fnc_name not in fnc_ctx.ai_functions:
```
theomonnom, line 315: This function was initially for OpenAI but if they all support the same kind of arguments (same json format).
Then it makes sense to move it.

Tho I'm not sure about what the implementation will looks like when we add support for structured output

theomonnom: I would keep this function prefixed with _ tho. It's not really relevant to users

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -289,6 +289,8 @@ async def recv_task(ws: aiohttp.ClientWebSocketResponse):
                 except Exception:
                     logger.exception("failed to process AssemblyAI message")
 
+        ws: aiohttp.ClientWebSocketResponse | None = None
```
theomonnom, line 292: We didn't add assemblyai to the CI?

theomonnom (approved):

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
jayeshp19: we have, but test case was failing because it wasn't able to access `ws` in `finally` https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py#L313

PR #1262, opened 2024-12-19
removed duplications in function messages.

jmugicagonz:
@longcw @davidzhao this is my proposal for the duplicated messages problem.
When the LLM is given a prompt, the LLM will follow those instructions of the prompt. It can generate messages and call functions. The LLM has no visibility on what happens inside the functions, except for the description, parameters, what is returned from those functions. If we generate assistant messages inside the functions and we append them to the overall context, this ends up confusing the LLM. 

I have extensively tested it, and it ends up changing the behavior.

Is it as if i'm talking to you following some instructions and I see messages in the chat, coming from me, that I didn't write.
I understand the benefits of calling `agent.say` inside a function. However, IMHO, what we include in a context as said by the assistant, should come from the assistant. At least inside functions, where I've found this to be problematic. Outside of functions, using `agent.say` with `add_to_chat_ctx` is still not ideal, but up to the livekit user. Doing it inside the functions, forces every livekit user to have one original message split into 2 messages and the agent get confused.

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -69,7 +69,6 @@ class AgentCallContext:
         self._assistant = assistant
         self._metadata = dict[str, Any]()
         self._llm_stream = llm_stream
-        self._extra_chat_messages: list[ChatMessage] = []
 
     @staticmethod
     def get_current() -> "AgentCallContext":
@@ -91,15 +90,6 @@ class AgentCallContext:
 
     def llm_stream(self) -> LLMStream:
         return self._llm_stream
-
-    def add_extra_chat_message(self, message: ChatMessage) -> None:
-        """Append chat message to the end of function outputs for the answer LLM call"""
-        self._extra_chat_messages.append(message)
-
-    @property
-    def extra_chat_messages(self) -> list[ChatMessage]:
-        return self._extra_chat_messages
-
 
 def _default_before_llm_cb(
     agent: VoicePipelineAgent, chat_ctx: ChatContext
@@ -446,7 +436,6 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
         await self._track_published_fut
 
         call_ctx = None
-        fnc_source: str | AsyncIterable[str] | None = None
         if add_to_chat_ctx:
             try:
                 call_ctx = AgentCallContext.get_current()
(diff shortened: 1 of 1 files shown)


longcw:
I don't really agree with "what we include in a context as said by the assistant, should come from the assistant.". LLM is a complementation model even it's in the chat mode, it should have the ability to complement the content following the input user gives it.

Meanwhile, users can set `add_to_chat_ctx` for `say` to avoid it being added to the ctx. From my testing in https://github.com/livekit/agents/pull/1165, adding the said message to the end of the chat ctx makes the generated answer more natural.


> I have extensively tested it, and it ends up changing the behavior.

I agree that adding messages to the chat ctx will change the behavior of the LLM, what I don't agree is this is a problem. It's highly depends on what messages added to the ctx, and users should have the flexibility to decide if this is the behavior they expected.

longcw:
It will be very helpful if you could share some cases this causes the problem.

I have tested the prompt below and it worked as expected that function call always returns with a message. In my tests, `say` is disabled in function calls, and only one of these two options should be used at a time (say something or prompt LLM to return a message).
> "You are a weather assistant created by LiveKit. Your interface with users will be voice. "
"You will provide weather information for a given location. "
"When a user asks for the weather information in a given location, you will say 'thanks for asking! I'll get you the weather information for that location' and then you will inmediately call the function 'get_weather' to retrieve the weather information."

jmugicagonz:
Sorry @longcw but I can tell you that the current setup, splitting one message with content + function into two messages, one with content and another with function ends up confusing the agent which will no longer do both things in the same message, and sends the content without calling the function.

> I agree that adding messages to the chat ctx will change the behavior of the LLM, what I don't agree is this is a problem. It's highly depends on what messages added to the ctx, and users should have the flexibility to decide if this is the behavior they expected.

I agree that users should keep the flexibility. And if they want to add a message in the middle, let it be that way. However, for those users that want to respect the original chat context, the moment 1 message goes into 2 messages, the user no longer have flexibility to keep them as one, and the behavior of the LLM changes.

jmugicagonz:
> It will be very helpful if you could share some cases this causes the problem.
> 
> I have tested the prompt below and it worked as expected that function call always returns with a message. In my tests, `say` is disabled in function calls, and only one of these two options should be used at a time (say something or prompt LLM to return a message).
> 
> > "You are a weather assistant created by LiveKit. Your interface with users will be voice. "
> > "You will provide weather information for a given location. "
> > "When a user asks for the weather information in a given location, you will say 'thanks for asking! I'll get you the weather information for that location' and then you will inmediately call the function 'get_weather' to retrieve the weather information."

I just tried with this prompt and I can't replicate it... it might be because it's "too simple". But definitely happens when both the prompt and the conversation get longer

jmugicagonz:
> #1078

This LGTM. Thanks for proposing it @longcw !

jmugicagonz:
Changes done in https://github.com/livekit/agents/pull/1254

PR #1266, opened 2024-12-21
fix: set USE_DOCSTRING as default for ai_callable

longcw:
`@llm.ai_callable` need a explicit `description=llm.USE_DOCSTRING` to use the docstring as the description.

### DIFF
--- a/livekit-agents/livekit/agents/llm/function_context.py
+++ b/livekit-agents/livekit/agents/llm/function_context.py
@@ -105,7 +105,7 @@ class CalledFunction:
 def ai_callable(
     *,
     name: str | None = None,
-    description: str | _UseDocMarker | None = USE_DOCSTRING,
+    description: str | _UseDocMarker = USE_DOCSTRING,
     auto_retry: bool = False,
 ) -> Callable:
     def deco(f):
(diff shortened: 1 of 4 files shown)


davidzhao (commented):
this is rather annoying tho as a requirement.. wdyt about just making that the default behavior?

if no explicit description is provided, it'll use docstring. if docstring isn't set, it raises

longcw:
I think that makes sense, most people will explicitly set a description or write that in the docstring.

At least the current example is very misleading, I've always assumed that I just need to set the description at docstring, but it actually also needs `description=llm.USE DOCSTRING` to take effect.

davidzhao:
> I think that makes sense, most people will explicitly set a description or write that in the docstring.
> 
> At least the current example is very misleading, I've always assumed that I just need to set the description at docstring, but it actually also needs `description=llm.USE_DOCSTRING` to take effect.

honestly I did too.. 🤦  I think it should just "do the obvious thing" here..

davidzhao:
@longcw do you wanna take a stab at making it the default behavior?

longcw:
@davidzhao sure, updated.

livekit-agents/livekit/agents/llm/function_context.py
```diff
@@ -105,7 +105,7 @@ class CalledFunction:
 def ai_callable(
     *,
     name: str | None = None,
-    description: str | _UseDocMarker | None = None,
+    description: str | _UseDocMarker | None = USE_DOCSTRING,
```
davidzhao, line 108: we don't need the `None` value anymore? it seems that we need either a description, or USE_DOCSTRING

davidzhao (approved):

Issue #1274, opened 2024-12-22
cerebras llm - 422 - property metadata is unsupported


sam-trost:
The openai llm component is including a `metadata` and `store` fields that are not supported by all openai compliant providers (ex. cerebras, groq)

```bash
raise self._make_status_error_from_response(err.response) from None
openai.UnprocessableEntityError: Error code: 422 - {'message': "metadata: property 'metadata' is unsupported", 'type': 'invalid_request_error', 'param': 'validation_error', 'code': 'wrong_api_format'}
```

Example agent

```python
import os
import asyncio

from livekit.agents import JobContext, WorkerOptions, cli, JobProcess
from livekit.agents.llm import (
    ChatContext,
    ChatMessage,
)
from livekit.agents.voice_assistant import VoiceAssistant
from livekit.plugins import deepgram, silero, elevenlabs, openai

from dotenv import load_dotenv

load_dotenv()


def prewarm(proc: JobProcess):
    proc.userdata["vad"] = silero.VAD.load()


async def entrypoint(ctx: JobContext):
    initial_ctx = ChatContext(
        messages=[
            ChatMessage(
                role="system",
                content="You are a voice assistant. Pretend we're having a human conversation, no special formatting or headings, just natural speech.",
            )
        ]
    )

    assistant = VoiceAssistant(
        vad=ctx.proc.userdata["vad"],
        stt=deepgram.STT(),
        llm=openai.LLM.with_cerebras((
            base_url="https://api.cerebras.ai/v1",
            api_key=os.environ.get("CEREBRAS_API_KEY"),
            model="llama-3.3-70b",
        ),
        tts=elevenlabs.TTS(),
        chat_ctx=initial_ctx,
    )

    await ctx.connect()
    assistant.start(ctx.room)
    await asyncio.sleep(1)
    await assistant.say("Hi there, how are you doing today?", allow_interruptions=True)


if __name__ == "__main__":
    cli.run_app(WorkerOptions(entrypoint_fnc=entrypoint, prewarm_fnc=prewarm))

```

SyedAddanACH:
Having the same issue with groq. Hopefully your PR gets merged soon! ✌🏽

robertvy:
having this issue on azure:

[...]

PR #1276, opened 2024-12-22
fix unknown `metadata` & `store` fields on OpenAI-like API

theomonnom:
adding metadata & store break Groq & Cerebras API


theomonnom:
fix #1274

davidzhao (approved):

PR #1279, opened 2024-12-23
fix context when functions have been called

jmugicagonz:
If a function has been called but the extra tools haven't been committed to the chat context, the response will be generated with a copied chat ctx that doesn't include those calls, and thus the LLM will 're-call' them.
This is a fix so the copied chat ctx contains those calls as well.


longcw (approved):
lgtm!

PR #1280, opened 2024-12-23
Substitute google error for warning

jmugicagonz:
Google Cloud offers other authentication methods apart from the env variable: see [here](https://cloud.google.com/docs/authentication/application-default-credentials)

Thus, if this variable is not set, we should show a warning, but not raise an error, as the user might be using other methods


davidzhao (approved):
makes sense. thanks!

PR #1284, opened 2024-12-23
encode boost words

jmugicagonz:
Boost words do not work with the Assembly AI plugin as they are not encoded in the URL. This fix encodes them.


davidzhao (approved):
lg!

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -318,7 +318,7 @@ async def recv_task(ws: aiohttp.ClientWebSocketResponse):
     async def _connect_ws(self) -> aiohttp.ClientWebSocketResponse:
         live_config = {
             "sample_rate": self._opts.sample_rate,
-            "word_boost": self._opts.word_boost,
+            "word_boost": json.dumps(self._opts.word_boost) if self._opts.word_boost is not None else None,
```
davidzhao, line 321:
```suggestion
            "word_boost": json.dumps(self._opts.word_boost)
            if self._opts.word_boost is not None
            else None,
```

PR #1288, opened 2024-12-23
Support Multisegments for TTS input

jayeshp19:
This PR implements support for multiple segments in tts input


livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -251,12 +260,52 @@ def __init__(self, *, tts: TTS, conn_options: APIConnectOptions) -> None:
...
-                return await self._run()
+                tasks = [
+                    asyncio.create_task(_tokenize_input()),
+                    asyncio.create_task(_run_segments()),
+                ]
+                try:
+                    await asyncio.gather(*tasks)
+                finally:
+                    await utils.aio.gracefully_cancel(*tasks)
+                return
```
theomonnom, line 308: I don't think we need any tokenizer in the base class?
This logic is handled separately (and differently) on each subclass

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -251,12 +260,52 @@ def __init__(self, *, tts: TTS, conn_options: APIConnectOptions) -> None:
...
+                        input_stream.end_input()
+
+                    input_stream = None
+
+            self._segments_ch.close()
+
+        @utils.log_exceptions(logger=logger)
+        async def _run_segments():
+            async for input_stream in self._segments_ch:
+                await self._run(input_stream)
```
theomonnom, line 296: This will not reuse the existing connection? (I know Cartesia supports having multiple segments without disconnecting, 11labs may too but I'm not sure)

jayeshp19: I don't think 11labs supports it, I've updated code for plugins to reuse ws connection

tests/test_tts.py
```diff
@@ -109,27 +109,36 @@ async def test_stream(tts_factory):
...
 
-    stream.flush()
-    # if i == 1:
-    stream.end_input()
+        stream.flush()
+        if i == 1:
+            stream.end_input()
 
     frames = []
     is_final = False
```
theomonnom, line 144: each segment should have its own `is_final`

PR #1131, opened 2024-11-23
feat: add update_options to STTs

jayeshp19:
### What
This PR implements update_options functionality for STTs.

### Why
Feature Improvement

### How
Example Usage
```python
stt = openai.STT()
agent = VoicePipelineAgent(…)

stt.update_options(…)
```

PR #1291, opened 2024-12-24
Ensure STT exceptions are being propagated

davidzhao:
when #1131 changed plugins to use `asyncio.await` instead of `asyncio.gather`, it broke the retry mechanisms that we had.

`.gather` automatically propagates exceptions, while `.await` does not. this means we'd have to propagate them explicitly in order for retries to work.

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -25,7 +25,13 @@ from typing import List, Literal, Optional
 from urllib.parse import urlencode
 
 import aiohttp
-from livekit.agents import DEFAULT_API_CONNECT_OPTIONS, APIConnectOptions, stt, utils
+from livekit.agents import (
+    DEFAULT_API_CONNECT_OPTIONS,
+    APIConnectOptions,
+    APIStatusError,
+    stt,
+    utils,
+)
 from livekit.agents.stt import SpeechEvent
 from livekit.agents.utils import AudioBuffer
 
@@ -274,8 +280,11 @@ class SpeechStream(stt.SpeechStream):
                     if closing_ws:  # close is expected, see SpeechStream.aclose
                         return
 
-                    raise Exception(
+                    raise APIStatusError(
                         "AssemblyAI connection closed unexpectedly",
+                        status_code=-1,
+                        request_id=None,
+                        body=None,
                     )  # this will trigger a reconnection, see the _run loop
 
                 if msg.type != aiohttp.WSMsgType.TEXT:
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -282,9 +282,6 @@ class SpeechStream(stt.SpeechStream):
 
                     raise APIStatusError(
                         "AssemblyAI connection closed unexpectedly",
-                        status_code=-1,
-                        request_id=None,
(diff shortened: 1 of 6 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -471,7 +471,7 @@ async def recv_task(ws: aiohttp.ClientWebSocketResponse):
                         return
 
                     # this will trigger a reconnection, see the _run loop
-                    raise Exception("deepgram connection closed unexpectedly")
+                    raise APIStatusError("deepgram connection closed unexpectedly")
```
davidzhao, line 474: generic exceptions are not retried, so we'd want to raise `APIError`s

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -498,6 +498,14 @@ async def recv_task(ws: aiohttp.ClientWebSocketResponse):
                         [asyncio.gather(*tasks), wait_reconnect_task],
                         return_when=asyncio.FIRST_COMPLETED,
                     )  # type: ignore
+
+                    # need to propagate exceptions from completed tasks
+                    for task in done:
+                        if task != wait_reconnect_task:
+                            exc = task.exception()
+                            if exc is not None:
+                                raise exc
```
theomonnom, line 507: can be simplified to:
```suggestion
                            task.result()
```

jayeshp19:
I've updated for remaining stts aswell

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -274,8 +280,11 @@ async def recv_task(ws: aiohttp.ClientWebSocketResponse):
                     if closing_ws:  # close is expected, see SpeechStream.aclose
                         return
 
-                    raise Exception(
+                    raise APIStatusError(
                         "AssemblyAI connection closed unexpectedly",
+                        status_code=-1,
+                        request_id=None,
+                        body=None,
```
davidzhao, line 287: just style nit: probably leave these out? instead of setting explicitly to None

jayeshp19: its for typecheck

davidzhao: ah I see. should we fix the constructor to default to None for them? and maybe we can pass in a `request_id` when available

jayeshp19: image
it expects those params, or we can adjust defination to allow raising APIStatusError without explicitly setting each param

davidzhao: yes.. that's what I meant.. can't we just change the class so it's not as annoying to use?

davidzhao: anyways, just a nit. the rest of the change looks good. let's get a fix out as this would improve recoverability

jayeshp19: updated, Let me know if looks good

davidzhao: lgtm

jayeshp19 (approved):

Issue #1179, opened 2024-12-05
Add method to manually interrupt the agent.


martin-purplefish:
It would be nice to have more control over the agent from the outside.

One use case I have: The user hangs up while the agent is speaking. I would like to be able to manually interrupt the agent, so that I can log what the agent had already said on my backend.
 
Basically, in my participant_disconnected handler, I'd like to be able to say `agent.interrupt()` - which should interrupt uninterruptable speech - and then flush the final events so I can send them to my backend.

longcw:
@martin-purplefish Is https://github.com/livekit/agents/pull/1294 works for you? we can close this one if so.

martin-purplefish:
Yep! Though I need something for the human side as well - basically flush when I hang up. But I'll close this one - THANK YOU!

Jeandcc:
Hi folks. 

What do you suggest in terms of approach to "resume" the agent? Say that we call `agent.interrupt` because the user started to draw something on screen. How could I "resume" the agent once I determine it's time for the agent to go back speaking?

I don't love the idea of using `agent.say` because it requires adding an extra interaction that is not needed. Ideally, the agent would just pick up from where it left off.

@martin-purplefish , @longcw

longcw:
@Jeandcc you can use `session.generate_reply()` in the agents 1.0

Jeandcc:
Hi @longcw ! Forgot to send this to you, but thank you for the reply! Will use it after I rework my codebase to work with the v1.

PR #1294, opened 2024-12-24
fix: add manual interrupt for pipeline agent

longcw:
Add `interrupt` method for pipeline agent. 

Related to https://github.com/livekit/agents/issues/1179

### DIFF
--- a/.changeset/slow-walls-bake.md
+++ b/.changeset/slow-walls-bake.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+add manual interrupt method for pipeline agent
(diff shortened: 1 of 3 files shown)


martin-purplefish:
What do you think about also interrupting human speech (basically forcing a flush of it - maybe a separate method)? For our use case- if we end the room while the human is talking, we don't get a final commit event. It's rare in practice but it happens.

longcw:
> What do you think about also interrupting human speech (basically forcing a flush of it - maybe a separate method)? For our use case- if we end the room while the human is talking, we don't get a final commit event. It's rare in practice but it happens.

Yeah this should be another method. For STT in stream mode it should be a force flush, for non-stream mode, it needs to trigger a VAD event to trigger the STT recognize.

theomonnom (approved):

PR #1301, opened 2024-12-26
Forward events for partial and final transcript from human input

martin-purplefish:
This will allow the agent to act on updating user speech, as needed.


martin-purplefish:
@davidzhao @theomonnom lmk what you think about this. I'm trying to make my agent dynamically adjust min_endpointing_delay based on how long the user has been talking.

PR #1306, opened 2024-12-27
Do not pass function context if at max depth

martin-purplefish:
As it stands, the agent just stops when it hits the maximum depth - this way it'll be forced to synthesize a reply.

### DIFF
--- a/.changeset/khaki-candles-rest.md
+++ b/.changeset/khaki-candles-rest.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+Do not pass function context if at max depth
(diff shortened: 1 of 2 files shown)


martin-purplefish:
PTAL @longcw @davidzhao  - I was running into the occasional "max function calls nested depth reached" - I think just not letting it make function calls at max depth is a better user experience? lmk what you think.

longcw:
This makes sense to me. I also noticed that when the max depth reached the agent may stop answering even user speech to the agent again, it will reach the max depth quickly again.
But it's better to add a warning there. @davidzhao @theomonnom wdyt?

davidzhao:
agreed.. it'd be good to log a warning so users are aware that tools are not being used here. we should prioritize not having the agent being stuck

martin-purplefish:
good call - fixed

davidzhao (approved):
lgtm!

PR #1309, opened 2024-12-29
added the ability to attach fields to Agent logger

davidzhao:
also logging worker_id to improve debugging


livekit-agents/livekit/agents/log.py
```diff
@@ -1,6 +1,27 @@
...
+        # Merge global fields with any per-message extras
+        extras = self.extra
+        if "extra" in kwargs:
+            extras = self.extra.copy()
+            extras.update(kwargs["extra"])
+        kwargs["extra"] = extras
+        return msg, kwargs
+
+    def with_fields(self, fields: dict[str, Any]) -> "FieldsLogger":
+        return FieldsLogger(self.logger, {**self.extra, **fields})
```
theomonnom, line 24: I don't think this is needed if we allow people to use their own loggers.
The relevant thing to do would be to create an example that use a custom logger for everything (even plugins)

davidzhao: I'm not sure that I agree.. the loggers in python are pretty primitive. this is just porting over the pattern that has worked well for us on the Go side here.

if folks are using their own loggers, there still isn't a simple way to attach fields to it so that they are logged each time.. even our own implementation of adding pid feels hacky.

PR #1311, opened 2024-12-30
Support Silero TTS

yorick-ml:
Add support [Silero TTS models ](https://github.com/snakers4/silero-models) for local usage


davidzhao:
hi @yorick-ml, thank you for your contribution.. we'd love to merge this.. however, I think the current model will be problematic.

each job runs in their own process, so we wouldn't want to load models for each job.  I think the right thing to do is to start an inference process like we do with the [turn detection model](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/eou.py), and running inference in a subclass of `InferenceRunner`

PR #1316, opened 2024-12-31
avoid warnings when function depth matches limit

davidzhao:
currently we'd get warnings when synthesizing the response to a function call. due to a current limitation, we aren't able to only include the function that we are processing.

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -963,15 +963,17 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
             fnc_ctx = self.fnc_ctx
             if (
                 fnc_ctx
-                and new_speech_handle.fnc_nested_depth > self._opts.max_nested_fnc_calls
+                and new_speech_handle.fnc_nested_depth
+                >= self._opts.max_nested_fnc_calls
             ):
-                logger.warning(
-                    "max function calls nested depth reached, not propagating fnc ctx",
-                    extra={
-                        "speech_id": speech_handle.id,
-                        "fnc_nested_depth": speech_handle.fnc_nested_depth,
-                    },
-                )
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -963,8 +963,7 @@ async def _execute_function_calls() -> None:
             fnc_ctx = self.fnc_ctx
             if (
                 fnc_ctx
-                and new_speech_handle.fnc_nested_depth
-                >= self._opts.max_nested_fnc_calls
+                and new_speech_handle.fnc_nested_depth > self._opts.max_nested_fnc_calls
```
longcw, line 966: it's a little bit tricky, it should be `>=` to avoid a new function call from this round of chat but we don't want a warning everytime there is a function call when the `max_nested` is 1.

davidzhao: ideally we'd want the current function to be passed in.. to give the LLM the full context right?

longcw: There are two options:
1. remove the fnc_ctx when `nested_depth >= max_nested` so the next answer won't have any function calls, one issue is that this would be a "silent failure" (or log a warning every time a function is called when `max_nested` is 1).
2. change the `>=` to `>` and if there are function calls have `nested_depth == max_nested + 1`, we don't actually execute the functions but return a message like "max nested reached" to LLM to generate an answer to users. (The current behaviour is skip the function calls silently so sometimes the agent is stuck after max_nested reached)

wdyt @davidzhao @martin-purplefish

davidzhao: ended up doing option1, but only logging the info if there are other functions involved. otherwise there are no change to functionality

longcw (approved):

PR #1319, opened 2024-12-31
ensure job status updates contain the correct status

davidzhao:
we were marking every job as failed to the server, due to an order of op error. _main_task is running within the subprocess, and would not be able to retrieve its own exit code.

it was always None, and was evaluated to JobStatus.FAILED

### DIFF
--- a/livekit-agents/livekit/agents/ipc/job_proc_executor.py
+++ b/livekit-agents/livekit/agents/ipc/job_proc_executor.py
@@ -102,7 +102,6 @@ class ProcJobExecutor(SupervisedProc):
     async def _supervise_task(self) -> None:
         await super()._supervise_task()
 
-        logger.info(f"job exit code: {self.exitcode}")
         self._job_status = JobStatus.SUCCESS if self.exitcode == 0 else JobStatus.FAILED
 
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/ipc/job_proc_executor.py
```diff
@@ -98,9 +98,11 @@ async def _main_task(self, ipc_ch: aio.ChanReceiver[channel.Message]) -> None:
...
             await aio.gracefully_cancel(*self._inference_tasks)
 
-            self._job_status = (
-                JobStatus.SUCCESS if self.exitcode == 0 else JobStatus.FAILED
-            )
+    @log_exceptions(logger=logger)
+    async def _supervise_task(self) -> None:
+        await super()._supervise_task()
+
+        self._job_status = JobStatus.SUCCESS if self.exitcode == 0 else JobStatus.FAILED
```
theomonnom, line 105:
```suggestion
        try:
                await super()._supervise_task()
        finally:
                self._job_status = JobStatus.SUCCESS if self.exitcode == 0 else JobStatus.FAILED
```

theomonnom (approved):

PR #1320, opened 2024-12-31
Add out-of-band responses + controlled response timing 

tinalenguyen:
I recently came across new additions that OpenAI developed and would love for them to be included in Livekit!

I worked on [out-of-band responses](https://platform.openai.com/docs/guides/realtime-model-capabilities#create-responses-outside-the-default-conversation) and an [extra VAD option](https://platform.openai.com/docs/guides/realtime-model-capabilities#voice-activity-detection-vad) where responses are not automatic. I also added an option to turn server VAD off by `turn_detection="None"`  in `session.update()`.

Please let me know your thoughts! :-)

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -76,6 +76,8 @@ class RealtimeResponse:
     """timestamp when the response was created"""
     _first_token_timestamp: float | None = None
     """timestamp when the first token was received"""
+    metadata: map | None = None
+    """developer-provided string key-value pairs"""
 
 
 @dataclass
@@ -140,6 +142,7 @@ class ServerVadOptions:
     threshold: float
     prefix_padding_ms: int
     silence_duration_ms: int
+    create_response: bool
 
 
 @dataclass
@@ -191,6 +194,7 @@ DEFAULT_SERVER_VAD_OPTIONS = ServerVadOptions(
     threshold=0.5,
     prefix_padding_ms=300,
     silence_duration_ms=500,
+    create_response=True,
 )
 
 DEFAULT_INPUT_AUDIO_TRANSCRIPTION = InputTranscriptionOptions(model="whisper-1")
@@ -717,6 +721,10 @@ class RealtimeSession(utils.EventEmitter[EventTypes]):
             on_duplicate: Literal[
                 "cancel_existing", "cancel_new", "keep_both"
             ] = "keep_both",
+            instructions: str | None = None,
+            modalities: list[api_proto.Modality] | None = None,
+            conversation: Literal["auto", "none"] = "auto",
+            metadata: map | None = None,
         ) -> asyncio.Future[bool]:
             """Creates a new response.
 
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -287,9 +288,6 @@ def __init__(
             ValueError: If the API key is not provided and cannot be found in environment variables.
         """
         super().__init__()
-        self._capabilities = Capabilities(
```
theomonnom, line 290: We should keep the `capabilities` property

tinalenguyen: Was that a new addition? I just updated with the differences

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -880,7 +938,7 @@ def session_update(
         input_audio_format: api_proto.AudioFormat | None = None,
         output_audio_format: api_proto.AudioFormat | None = None,
         input_audio_transcription: InputTranscriptionOptions | None = None,
-        turn_detection: ServerVadOptions | None = None,
+        turn_detection: ServerVadOptions | Literal["None"] | None = None,
```
theomonnom, line 941: I didn't see this PR was also adding None to the turn_detection.
It has been done inside this PR: https://github.com/livekit/agents/pull/1347

We're still interested to have this out-of-band responses tho! :)

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -755,7 +769,29 @@ def create(
...
+                    self._sess._queue_msg(
+                        {
+                            "type": "response.create",
+                            "response": {
+                                "conversation": conversation,
+                                "metadata": metadata,
+                                "modalities": modalities,
+                            },
+                        }
+                    )
```
theomonnom, line 794: Is the reason for this if-else condition the fact that OAI will returns an error if instructions is set to None/null

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
theomonnom: Yes it is a new one

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
tinalenguyen: My bad, I should've specified but I integrated the changes! It should all work now :D

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
tinalenguyen: I forgot but that sounds about right, it works without it now though

theomonnom (approved):
Thanks!

PR #1321, opened 2024-12-31
Support Transcriber session with Google realtime API

jayeshp19:


livekit-agents/livekit/agents/multimodal/multimodal_agent.py
```diff
@@ -311,19 +311,32 @@ def _input_speech_transcription_completed(ev: _InputTranscriptionProto):
                     alternatives=[stt.SpeechData(language="", text=ev.transcript)],
                 )
             )
-            user_msg = ChatMessage.create(
-                text=ev.transcript, role="user", id=ev.item_id
-            )
+            if self._model.capabilities.supports_truncate:
+                user_msg = ChatMessage.create(
```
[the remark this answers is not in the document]
jayeshp19: Some methods are not implemented in Gemini. We maintain remoteconversations in OpenAI, but not in Gemini. We should prevent invoking those methods when using Gemini. The purpose of supports_truncate is to differentiate between that

examples/multimodal-agent/gemini_agent.py
```diff
@@ -50,13 +50,25 @@ async def get_weather(
...
     participant = await ctx.wait_for_participant()
 
+    # chat_ctx is used to serve as initial context, Agent will start the conversation first if chat_ctx is provided
     chat_ctx = llm.ChatContext()
+    chat_ctx.append(text="What is LiveKit?", role="user")
+    chat_ctx.append(
+        text="LiveKit is the platform for building realtime AI. The main use cases are to build AI voice agents. LiveKit also powers livestreaming apps, robotics, and video conferencing.",
+        role="assistant",
+    )
+    chat_ctx.append(text="What is the LiveKit Agents framework?", role="user")
```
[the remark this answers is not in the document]
jayeshp19: No, it can be either `assistant` or `user`.

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
```diff
@@ -311,19 +311,34 @@ def _input_speech_transcription_completed(ev: _InputTranscriptionProto):
...
             )
+            self._emit_speech_committed("agent", ev.transcript)
+
+        # Similar to _input_speech_started, this handles updating the state to "listening" when the agent's speech is complete.
+        # However, since Gemini doesn't support VAD events, we are not emitting the `user_started_speaking` event here.
+        @self._session.on("agent_speech_completed")
+        def _agent_speech_completed():
+            self._update_state("listening")
+            if self._playing_handle is not None and not self._playing_handle.done():
+                self._playing_handle.interrupt()
```
theomonnom, line 341: Why we should interrupt here?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/api_proto.py
```diff
@@ -1,10 +1,22 @@
 from __future__ import annotations
 
 import inspect
+import json
 from typing import Any, Dict, List, Literal, Sequence, Union
 
+from livekit.agents import llm
+
 from google.genai import types  # type: ignore
```
theomonnom, line 9: The code here is hard to follow, really sad we don't have types

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -246,12 +271,21 @@ def __init__(
...
+            self._transcriber = TranscriberSession(
+                client=self._client, model=self._opts.model
+            )
+            self._agent_transcriber = TranscriberSession(
+                client=self._client, model=self._opts.model
+            )
+            self._transcriber.on("input_speech_done", self._on_input_speech_done)
+            self._agent_transcriber.on("input_speech_done", self._on_agent_speech_done)
+        # init dummy task
+        self._init_sync_task = asyncio.create_task(asyncio.sleep(0))
```
theomonnom, line 284: ?

theomonnom: This isn't really doing anything?

theomonnom:
Where do we make sure the transcribed user speech is inside the chat_ctx and always before the generated agent speech?

jayeshp19:
> This isn't really doing anything?

Yes, it is being called directly from the base class. We need to keep the dummy task unless we wrap it with capabilities.support_truncate in the base class.
> Where do we make sure the transcribed user speech is inside the chat_ctx and always before the generated agent speech?

I don't think we need it, as the transcriber and LLM are independent of each other.

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
jayeshp19: Because we call this function when speech is interrupted as well. They likely made some changes, but now Gemini returns `server.turn_complete` instead of `server.interrupted` when interrupted. It's confusing. In both cases, we are calling this function.

theomonnom:
> Yes, it is being called directly from the base class. We need to keep the dummy task unless we wrap it with capabilities.support_truncate in the base class.
I'm not sure to follow, the baseclass is utils.EventEmitter[EventTypes]


> I don't think we need it, as the transcriber and LLM are independent of each other.
How do we get the user messages inside the ChatContext?

jayeshp19:
> I'm not sure to follow, the baseclass is utils.EventEmitter[EventTypes]
> 

I mean `multimodal.py` 

> How do we get the user messages inside the ChatContext?
> 

from here when audio transcription is done- https://github.com/livekit/agents/pull/1321/files#diff-4b3e6842c9b1bf3130541b6b2fd18dcc7d1b0051285496eca0355e62938d13fbR351

theomonnom:
> > How do we get the user messages inside the ChatContext?
> 
> from here when audio transcription is done- [#1321 (files)](https://github.com/livekit/agents/pull/1321/files#diff-4b3e6842c9b1bf3130541b6b2fd18dcc7d1b0051285496eca0355e62938d13fbR351)

What I mean [here](https://github.com/livekit/agents/pull/1321#issuecomment-2586980465) is that on some bad timings or if the VAD events were different, the data inside the chat context will not be "stable".

E.g; 
- You could have multiple user messages for only one assistant messages
- The user messages could be appended *after* the assistant message (The order is wrong) 
- etc..

jayeshp19:
> Where do we make sure the transcribed user speech is inside the chat_ctx and always before the generated agent speech?
> 
> What I mean [here](https://github.com/livekit/agents/pull/1321#issuecomment-2586980465) is that on some bad timings or if the VAD events are different, the data inside the chat context will not be "stable".
> 
> E.g;
> 
> * You could have multiple user messages for only one assistant messages
> * The user messages could be appended _after_ the assistant message (The order is wrong)
> * etc..

User audio is usually processed in real-time, and we receive transcriptions quickly. However, you're right that these scenarios can occur.
Do you have any suggestions on how we can ensure the chat context remains stable and maintains the correct sequence?

theomonnom:
> User audio is usually processed in real-time, and we receive transcriptions quickly. However, you're right that these scenarios can occur. Do you have any suggestions on how we can ensure the chat context remains stable and maintains the correct sequence?

Maybe we can just ignore for now and don't add then to the ChatContext.

theomonnom (approved):

PR #1322, opened 2024-12-31
Fix env var typo

CatalinVoss:
Fix a minor typo in realtime Gemini live docstring…


davidzhao (approved):
oops, thanks for catching this!

PR #1326, opened 2025-01-02
fix: revert from weakset to list in multimodal for maintaining sessions

jayeshp19:


theomonnom:
self._rt_sessions.add(new_session) -> self._rt_sessions.append(new_session)

davidzhao (approved):

PR #1336, opened 2025-01-04
Add memory to load calc

martin-purplefish:
With the new turn detector, our processes are now memory bound - so incorporate that in the check.

### DIFF
--- a/livekit-agents/livekit/agents/utils/hw/memory.py
+++ b/livekit-agents/livekit/agents/utils/hw/memory.py
@@ -1,0 +1,63 @@
+import os
+from abc import ABC, abstractmethod
+
+import psutil
+
+
+class MemoryMonitor(ABC):
+    @abstractmethod
+    def memory_total(self) -> int:
+        """Total memory available in bytes."""
+        pass
+
+    @abstractmethod
+    def memory_used(self) -> int:
+        """Memory currently in use in bytes."""
+        pass
+
+    @abstractmethod
+    def memory_percent(self) -> float:
+        """Memory usage percentage between 0 and 1"""
+        pass
+
+
+class DefaultMemoryMonitor(MemoryMonitor):
+    def memory_total(self) -> int:
+        return psutil.virtual_memory().total
+
+    def memory_used(self) -> int:
+        return psutil.virtual_memory().used
+
+    def memory_percent(self) -> float:
+        return psutil.virtual_memory().percent / 100.0
+
+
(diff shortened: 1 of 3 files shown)


martin-purplefish:
PTAL @davidzhao @theomonnom

livekit-agents/livekit/agents/worker.py
```diff
@@ -88,19 +91,23 @@ def __init__(self) -> None:
...
-            return self._m_avg.get_avg()
+            return self._m_avg_cpu.get_avg(), self._m_avg_mem.get_avg()
 
     @classmethod
     def get_load(cls, worker: Worker) -> float:
         if cls._instance is None:
             cls._instance = _DefaultLoadCalc()
 
-        return cls._instance._m_avg.get_avg()
+        return max(
```
martin-purplefish, line 108: we could be more clever here (e.g. if we wanted different thresholds for memory and CPU usage)

livekit-agents/livekit/agents/utils/hw/memory.py
```diff
@@ -0,0 +1,63 @@
...
+        return psutil.virtual_memory().percent / 100.0
+
+
+class CGroupV2MemoryMonitor(MemoryMonitor):
+    def memory_total(self) -> int:
+        try:
+            with open("/sys/fs/cgroup/memory.max", "r") as f:
+                max_memory = f.read().strip()
+                if max_memory == "max":
+                    return psutil.virtual_memory().total
```
martin-purplefish, line 41: I _think_ this makes sense, alternatively we can just treat it as unlimited.

davidzhao (commented):
thanks for adding this!

I think the memory monitor should work a bit differently compared to CPU due to the type of resources.

for memory, I think it's rather binary: as long as we have x% or xMB of memory free, we should be able to take jobs. for CPU, we'd like to stop earlier because jobs could have bursty usage patterns, so leaving plenty of headroom (stopping to take jobs at 65% utilization) would reduce the likelihood that we'll get throttled by CPU

martin-purplefish:
> thanks for adding this!
> 
> I think the memory monitor should work a bit differently compared to CPU due to the type of resources.
> 
> for memory, I think it's rather binary: as long as we have x% or xMB of memory free, we should be able to take jobs. for CPU, we'd like to stop earlier because jobs could have bursty usage patterns, so leaving plenty of headroom (stopping to take jobs at 65% utilization) would reduce the likelihood that we'll get throttled by CPU

Got it, I'll take a pass at that

PR #1337, opened 2025-01-06
fix Google STT handling of session timeouts

### LINKED ISSUE #1258: Google STT times out at 5 minutes and crashes agent
Google STT seems to consistently fail at 5 minutes into the call with the following error:

```
2024-12-18 15:59:58,786 - ERROR livekit.agents.pipeline - Error in _recognize_task
  File "/Users/nick/repos/hellopatient/lk_agent/.venv/lib/python3.12/site-packages/livekit/agents/utils/log.py", line 16, in async_fn_logs
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nick/repos/hellopatient/lk_agent/.venv/lib/python3.12/site-packages/livekit/agents/pipeline/human_input.py", line 150, in _recognize_task
    await asyncio.gather(*tasks)
  File "/Users/nick/repos/hellopatient/lk_agent/.venv/lib/python3.12/site-packages/livekit/agents/pipeline/human_input.py", line 120, in _audio_stream_co
    stt_stream.push_frame(ev.frame)
  File "/Users/nick/repos/hellopatient/lk_agent/.venv/lib/python3.12/site-packages/livekit/agents/stt/stt.py", line 265, in push_frame
    self._check_not_closed()
  File "/Users/nick/repos/hellopatient/lk_agent/.venv/lib/python3.12/site-packages/livekit/agents/stt/stt.py", line 327, in _check_not_closed
    raise RuntimeError(f"{cls.__module__}.{cls.__name__} is closed")
RuntimeError: livekit.plugins.google.stt.SpeechStream is closed {"pid": 54793, "job_id": "AJ_zKEGnh4r4ttx"}
``` 

Unfortunately it's an unhandled error and so even though we were using it with in the `FallbackAdapter` our agent stops responding completely at 5 minutes into the call. 

I have worked with the Google API before and I know it [times out the streaming connection at 5 minutes](https://cloud.google.com/speech-to-text/v2/quotas#streaming_requests) so I imagine that is the issue. There's [an official tutorial on endless streaming](https://cloud.google.com/speech-to-text/docs/transcribe-streaming-audio#endless-streaming) that shows how to switch to a new request and stitch them together in real-time.

davidzhao:
also reduces initial retry delay with STT/LLM/TTS connection errors

fixes #1258

### DIFF
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -119,6 +119,7 @@ class STT(
                 return event
 
             except APIError as e:
+                retry_interval = conn_options._interval_for_retry(i)
                 if conn_options.max_retry == 0:
                     raise
                 elif i == conn_options.max_retry:
@@ -127,7 +128,7 @@ class STT(
                     ) from e
                 else:
                     logger.warning(
-                        f"failed to recognize speech, retrying in {conn_options.retry_interval}s",
+                        f"failed to recognize speech, retrying in {retry_interval}s",
                         exc_info=e,
                         extra={
                             "tts": self._label,
@@ -136,7 +137,7 @@ class STT(
                         },
                     )
 
-                await asyncio.sleep(conn_options.retry_interval)
+                await asyncio.sleep(retry_interval)
 
         raise RuntimeError("unreachable")
 
@@ -215,8 +216,7 @@ class RecognizeStream(ABC):
 
         while num_retries <= max_retries:
             try:
-                await self._run()
-                num_retries = 0
+                return await self._run()
             except APIError as e:
                 if max_retries == 0:
                     raise
@@ -225,8 +225,9 @@ class RecognizeStream(ABC):
                         f"failed to recognize speech after {num_retries} attempts",
                     ) from e
                 else:
+                    retry_interval = self._conn_options._interval_for_retry(num_retries)
                     logger.warning(
-                        f"failed to recognize speech, retrying in {self._conn_options.retry_interval}s",
+                        f"failed to recognize speech, retrying in {retry_interval}s",
                         exc_info=e,
                         extra={
                             "tts": self._stt._label,
@@ -234,8 +235,8 @@ class RecognizeStream(ABC):
                             "streamed": True,
                         },
                     )
-                if num_retries > 0:
-                    await asyncio.sleep(self._conn_options.retry_interval)
+                    await asyncio.sleep(retry_interval)
+
                 num_retries += 1
 
     async def _metrics_monitor_task(
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -210,15 +210,19 @@ class RecognizeStream(ABC):
     async def _run(self) -> None: ...
 
     async def _main_task(self) -> None:
-        for i in range(self._conn_options.max_retry + 1):
+        max_retries = self._conn_options.max_retry
+        num_retries = 0
+
+        while num_retries <= max_retries:
             try:
-                return await self._run()
+                await self._run()
+                num_retries = 0
             except APIError as e:
-                if self._conn_options.max_retry == 0:
+                if max_retries == 0:
                     raise
-                elif i == self._conn_options.max_retry:
+                elif num_retries == max_retries:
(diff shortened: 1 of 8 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -449,9 +469,17 @@ async def process_stream(stream):
...
+            except GoogleAPICallError as e:
+                raise APIStatusError(
+                    e.message,
+                    status_code=e.code or -1,
+                )
+            except Exception as e:
+                raise APIConnectionError() from e
             finally:
-                if not self._reconnect_event.is_set():
-                    break
```
davidzhao, line 454: @jayeshp19 `break` here has the unexpected behavior of swallowing the exception. when this is set, the `except` in the parent class does not catch exceptions raised here.

jayeshp19: good catch! let's move this inside try block

jayeshp19 (changes requested):

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -399,6 +409,13 @@ async def process_stream(stream):
                                 alternatives=[speech_data],
                             )
                         )
+                        if (
```
jayeshp19, line 412: Might be easier to create a separate task to handle timeout, like [here](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py#L343)

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
davidzhao: why do we need this break?

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -210,28 +210,33 @@ def __init__(
...
+                        f"failed to recognize speech after {num_retries} attempts",
                     ) from e
                 else:
                     logger.warning(
                         f"failed to recognize speech, retrying in {self._conn_options.retry_interval}s",
                         exc_info=e,
                         extra={
                             "tts": self._stt._label,
-                            "attempt": i + 1,
+                            "attempt": num_retries,
```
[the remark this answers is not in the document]
davidzhao: this is intentional, attempt starts with 0

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
jayeshp19: This is because we have an infinite `while` loop, so we need to break if `wait_reconnect_task` is not  in done. This means other tasks are either finished or had exceptions, and we don't want to retry in that case.

jayeshp19 (approved):

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -44,6 +45,10 @@
 LgType = Union[SpeechLanguages, str]
 LanguageCode = Union[LgType, List[LgType]]
 
+# Google STT has a timeout of 5 mins, we'll attempt to restart the session
+# before that timeout is reached
+_max_session_duration = 4
```
theomonnom, line 50: This seems like 4 seconds to me

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -399,6 +411,20 @@ async def process_stream(stream):
                                 alternatives=[speech_data],
                             )
                         )
+                        if (
+                            time.time() - self._session_connected_at
+                            > _max_session_duration
+                        ):
+                            logger.debug("restarting session due to timeout")
```
theomonnom, line 418: Unnecessary log? Should be handled seamlessly IMO

anishhguptaa:
Hey, I am still getting this issue. Is the fix pushed in the latest version of 1.x or in the 0.12 as well?

`RuntimeError: livekit.plugins.google.stt.SpeechStream is closed`

PR #1340, opened 2025-01-06
playai: enable streaming TTS

davidzhao:


jayeshp19:
- added retry mechanism for ocassionally websocket connection drop

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
```diff
@@ -231,8 +235,13 @@ def _send_last_frame(*, segment_id: str, is_final: bool) -> None:
                 last_frame = None
 
         input_task = asyncio.create_task(self._tokenize_input())
+
         try:
             text_stream = await self._create_text_stream()
+            self._client = PlayHTAsyncClient(
+                user_id=self._user_id,
+                api_key=self._api_key,
+            )
```
theomonnom, line 244: Do we want to recreate a new client for every request?

jayeshp19: Yeah, it's a workaround for now. WebSocket connection drops from playht server sometimes and it's not handled in `pyht` sdk.

jayeshp19 (approved):

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
bryananderson: Hey guys, this is causing latency problems for customers--there is a significant overhead around creating the connection on every request.

jayeshp19: Hi, The workaround was introduced to mitigate occasional connection drops from the PlayHT server, as the current pyht SDK doesn't seem to handle these drops robustly. 

Could the SDK be updated to include automatic reconnection? This would greatly help reduce overhead

Let me know your thoughts!

bryananderson: Can you verify for me the pyht version that you have installed (even if a newer one is in your requirements.txt, maybe it has not actually been upgraded)?

Currently the ws.send() call is in a try-except that should re-establish the connection if there is an exception. Could be it is still dropping this somehow! But curious if upgrading to the latest pyht will help.

bryananderson: I think I figured out how this can still occur even in the latest pyht.

It is a little hard to test it as I have not been able to reproduce the disconnect, but I have a PR [here](https://github.com/playht/pyht/pull/87) and will try to have it tested and released tomorrow so you can remove the per-request reconnection.

PR #1346, opened 2025-01-07
update Cartesia plugin default model and voice id

noahlt:
This change updates the default model to `sonic` (see https://docs.cartesia.ai/build-with-sonic/models for details about our model aliases).

It also updates the default voice ID to one that we like a bit better.

(Hope you all are doing well :)


theomonnom (approved):

davidzhao (approved):
thanks for the PR Noah!

PR #1347, opened 2025-01-08
Support manual vad for realtime api

longcw:
To disable the server VAD for OAI realtime api, set `turn_detection=None`
```python
model=openai.realtime.RealtimeModel(
            voice="alloy",
            temperature=0.8,
            instructions="You are a helpful assistant",
            turn_detection=None,
        )
```

To manually commit user speech and trigger response:
```python
# at the end of the user's turn
session: RealtimeSession = agent._session
session.input_audio_buffer.commit() 

# listen to the input_speech_committed event and create a new response after that
@agent.on("input_speech_committed")
def _on_input_speech_committed():
    agent.interrupt()  # can also be called when the local VAD detects the start of the user speech
    session: RealtimeSession = agent._session
    session.response.create(on_duplicate="cancel_existing")
```

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -195,6 +195,13 @@ DEFAULT_SERVER_VAD_OPTIONS = ServerVadOptions(
 DEFAULT_INPUT_AUDIO_TRANSCRIPTION = InputTranscriptionOptions(model="whisper-1")
 
 
+class _NOTSET:
+    pass
+
+
+NOTSET = _NOTSET()
+
+
 class RealtimeModel:
     @overload
     def __init__(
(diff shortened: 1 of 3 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -195,6 +195,13 @@ class _ContentPtr(TypedDict):
 DEFAULT_INPUT_AUDIO_TRANSCRIPTION = InputTranscriptionOptions(model="whisper-1")
 
 
+class _NOTSET:
```
theomonnom, line 198: Let's use the NOT_GIVEN we already have here:
https://github.com/livekit/agents/blob/4a9b845117e0569ff8b221dc2c83e72e728ab853/livekit-agents/livekit/agents/types.py#L30

theomonnom (approved):

PR #1349, opened 2025-01-08
fix: function calls without text response are not added to chat ctx

longcw:
Someone in Slack has encountered this bug, where function calls without a text response are not added to the chat ctx, e.g. the first fnc in a nested function call returns another fnc but no text.

Fixed this by moving the collected text check to after adding the function call in the chat ctx.

### DIFF
--- a/.changeset/hip-bottles-joke.md
+++ b/.changeset/hip-bottles-joke.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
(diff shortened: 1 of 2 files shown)


martin-purplefish:
Oh nice, I think we ran into this a few times too.

davidzhao (approved):
lg

theomonnom (approved):

PR #1355, opened 2025-01-10
fix nested speech done future

longcw:
Add nested_speech_done_fut only if nested speech loop starts


longcw:
The llm test failed for some cases but it shouldn't be related to the changes in this pr. @jayeshp19 @davidzhao Any ideas?

jayeshp19:
Vertex fails very often. I’ll debug later if there have been any internal changes in how they handle tool_choice logic.

PR #1359, opened 2025-01-10
create_response and cancel_response APIs for MultimodalAgent

longcw:
Created a push-to-talk example for the manual VAD of OAI realtime api.

A follow up of https://github.com/livekit/agents/pull/1347


lukasIO:
just my 2c, but would be nice if this was using @bcherry's RPC feature via LiveKit, instead of a custom WebSocket ?

longcw:
> just my 2c, but would be nice if this was using @bcherry's RPC feature via LiveKit, instead of a custom WebSocket ?

The frontend in the example is just a simple button without a full livekit client. What is the simplest way to incorporate the RPC in a single web page or do we want to add a full client in the example. Or mention to use RPC for a real production in readme and comments?

longcw:
@lukasIO @davidzhao updated to use RPC with frontend example https://github.com/livekit-examples/voice-assistant-frontend/pull/23

examples/multimodal-agent/openai_manual_vad/push_to_talk.py
```diff
@@ -0,0 +1,75 @@
...
+    agent.start(ctx.room, participant)
+
+    @ctx.room.local_participant.register_rpc_method("ptt")
+    async def handle_ptt(data):
+        logger.info(f"Received PTT action: {data.payload}")
+        if data.payload == "push":
+            agent.interrupt()
+        elif data.payload == "release":
+            agent.generate_reply(on_duplicate="cancel_existing")
+        return "ok"
```
bcherry, line 45: FWIW i think it would make more sense for these to just be separate rpc methods e.g. `ptt.start` and `ptt.end` with an empty payload

longcw: Yeah, this is more clear. updated.

davidzhao (commented):
@longcw since the frontend will take awhile, can we push the multimodal_agent changes in first, without the PTT example?

examples/multimodal-agent/openai_manual_vad/push_to_talk.py
bcherry: thinking about this some more... you might consider having the frontend send PTT heartbeats every few seconds while the user continues to hold. that way if something goes wrong in the remote client the agent can time out its PTT handler and generate a response? otherwise it may be possible to get the state out of sync and have the agent hang for a significant period of time?

longcw:
> @longcw since the frontend will take awhile, can we push the multimodal_agent changes in first, without the PTT example?

sounds good. how about we keep the example but just mention how should it work in readme without a link to the frontend.

examples/multimodal-agent/openai_manual_vad/push_to_talk.py
longcw: This makes sense. but since it's just an example about how to interrupt and manually trigger an agent response, I prefer to keep it simple and mention the risk of state out of sync and solution in the comment.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1017,6 +1017,25 @@ def _validate_message(msg: llm.ChatMessage) -> bool:
...
+        ] = "keep_both",
+    ) -> None:
+        self.response.create(on_duplicate=on_duplicate)
+
+    def commit_audio_buffer(self) -> None:
+        self.input_audio_buffer.commit()
+
+    @property
+    def server_vad_enabled(self) -> bool:
+        return self._opts.turn_detection is not None
```
theomonnom, line 1037: I think we should put that inside the Capabilities dataclass

longcw: I was about to add a `support_manual_vad` in Capabilities dataclass. But it still need to read the `self._opts.turn_detection` check if the server vad enabled. Should we have both `support_manual_vad` and `session.vad_mode`?

davidzhao (approved):

PR #1362, opened 2025-01-12
fix azure speech for all types of configs

### LINKED ISSUE #1361: Azure plugin doesn't support Azure Speech service over Private endpoints
`livekit-plugins-azure` works fine with standard Azure Speech endpoints. However,  for connecting to Azure Speech over private endpoint using Speech SDK, need authentication support using `speech_host` and `speech_api_key`, and no `speech_region`.

Microsoft docs: https://learn.microsoft.com/en-us/azure/ai-services/speech-service/speech-services-private-link?tabs=portal#speech-resource-with-a-custom-domain-name-and-a-private-endpoint-usage-with-the-speech-sdk

Following lines of code need support for both `speech_host` + `speech_api_key` .
https://github.com/livekit/agents/blob/9ae16b01acb234b747874e5d6711816c46a3e5cc/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py#L82
https://github.com/livekit/agents/blob/9ae16b01acb234b747874e5d6711816c46a3e5cc/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py#L308

This affects both `stt.py` and `tts.py` classes.

longcw:
Pass all options to the `SpeechConfig` as it already has the sanity check (https://learn.microsoft.com/en-us/python/api/azure-cognitiveservices-speech/azure.cognitiveservices.speech.speechconfig?view=azure-python)
```python
        if endpoint is not None or host is not None:
            if region is not None:
                raise ValueError('cannot construct SpeechConfig with both region and endpoint or host information')
            if auth_token is not None:
                raise ValueError('cannot specify both auth_token and endpoint or host when constructing SpeechConfig. '
                                 'Set authorization token separately after creating SpeechConfig.')

        if region is not None and subscription is None and auth_token is None:
            raise ValueError('either subscription key or authorization token must be given along with a region')

        if subscription is not None and endpoint is None and host is None and region is None:
            raise ValueError('either endpoint, host, or region must be given along with a subscription key')

```

fixes #1361

### DIFF
--- a/.changeset/blue-boats-sin.md
+++ b/.changeset/blue-boats-sin.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-azure": patch
+---
+
+azure speech support all different configs
--- a/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
+++ b/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
@@ -41,6 +41,7 @@ class STTOptions:
     languages: list[
         str
     ]  # see https://learn.microsoft.com/en-us/azure/ai-services/speech-service/language-support?tabs=stt
+    speech_endpoint: str | None = None
 
 
 class STT(stt.STT):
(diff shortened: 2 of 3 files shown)


davidzhao (approved):
lg!

PR #1363, opened 2025-01-12
remove duplicated message

jmugicagonz:
This is the same fix as for the main chat context when speech is committed but to make sure that when we call the LLM with a copied chat_ctx, just after a tool call, we don't have the tool text message duplicated
@longcw


longcw:
lgtm!

longcw (approved):

PR #1364, opened 2025-01-12
🚀 livekit-agents v1.0

theomonnom:


livekit-agents/livekit/agents/pipeline/generation.py
```diff
@@ -0,0 +1,109 @@
...
+        if isinstance(llm_node, AsyncIterable):
+            # forward llm stream to output channels
+            try:
+                async for chunk in llm_node:
+                    # io.LLMNode can either return a string or a ChatChunk
+                    if isinstance(chunk, str):
+                        data.generated_text += chunk
+                        text_ch.send_nowait(chunk)
+
+                    elif isinstance(chunk, ChatChunk):
```
martin-purplefish, line 54: Maybe an else with a log or exception?

theomonnom: Sure, good idea in case users give us random stuff

livekit-agents/livekit/agents/multimodal/realtime.py
nbsp, line 1: so this namespace would remain `agents.multimodal`? why not rename it to realtime if we're renaming the classes?

theomonnom: Tho my 2c is that realtime isn't really the new category of those models. They're "multimodal"

livekit-agents/livekit/agents/pipeline/generation.py
```diff
@@ -0,0 +1,109 @@
...
+
+@dataclass
+class _LLMGenerationData:
+    text_ch: aio.Chan[str]
+    tools_ch: aio.Chan[FunctionCallInfo]
+    generated_text: str = ""
+    generated_tools: list[FunctionCallInfo] = field(default_factory=list)
+
+
+def do_llm_inference(
```
davidzhao, line 27: nit: `do_` -> `perform_` ? just wanted to standardize on a set of action verbs

livekit-agents/livekit/agents/pipeline/audio_recognition.py
```diff
@@ -0,0 +1,279 @@
...
+
+    This class is also responsible for the end of turn detection.
+    """
+
+    UNLIKELY_END_OF_TURN_EXTRA_DELAY = 6.0
+
+    def __init__(
+        self,
+        *,
+        agent: PipelineAgent,
```
davidzhao, line 49: would be better to take in a protocol that is a provider of chat context. would make it more easily testable.

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -0,0 +1,211 @@
...
+    Union,
+)
+
+from livekit import rtc
+
+from .. import llm, stt
+
+STTNode = Callable[
+    [AsyncIterable[rtc.AudioFrame]],
+    Union[Awaitable[Optional[AsyncIterable[stt.SpeechEvent]]]],
```
davidzhao, line 21: the Union here isn't doing anything right?

it'd be great to provide example of what are valid output types for these

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -0,0 +1,211 @@
...
+# TODO(theomonnom): Add documentation to VideoSink
+class VideoSink(ABC):
+    @abstractmethod
+    async def capture_frame(self, text: rtc.VideoFrame) -> None: ...
+
+    @abstractmethod
+    def flush(self) -> None: ...
+
+
+class AgentInput:
```
davidzhao, line 149: 💯

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -0,0 +1,211 @@
...
+    def __init__(self, video_changed: Callable, audio_changed: Callable) -> None:
+        self._video_stream: VideoStream | None = None
+        self._audio_stream: AudioStream | None = None
+        self._video_changed = video_changed
```
davidzhao, line 153: nit: should we keep our _cb suffix to indicate it's a callback?

livekit-agents/livekit/agents/pipeline/io.py
davidzhao: what about text input? ideally there's an interface to hook into the DataStreams API in the same way

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -0,0 +1,211 @@
...
+
+        return self.__last_playback_ev
+
+    @property
+    def sample_rate(self) -> int | None:
+        """The sample rate required by the audio sink, if None, any sample rate is accepted"""
+        return self._sample_rate
+
+    @abstractmethod
+    async def capture_frame(self, frame: rtc.AudioFrame) -> None:
```
davidzhao, line 110: do we want to keep our `capture_frame` terminology? or start switching to `push_frame` like we use elsewhere?

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -0,0 +1,211 @@
...
+
+    @abstractmethod
+    def clear_buffer(self) -> None:
+        """Clear the buffer, stopping playback immediately"""
+        ...
+
+
+class TextSink(ABC):
+    @abstractmethod
+    async def capture_text(self, text: str) -> None:
```
davidzhao, line 130: is there a way to update existing text? thinking in the case we are pushing interim transcripts out

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -1,1330 +1,664 @@
...
-) -> str | AsyncIterable[str]:
-    return text
+EventTypes = Literal[
+    "user_started_speaking",
+    "user_stopped_speaking",
+    "agent_started_speaking",
+    "agent_stopped_speaking",
+    "user_message_committed",
+    "agent_message_committed",
+    "agent_message_interrupted",
```
davidzhao, line 104: we need to add metrics still?

what about these?
    "function_calls_collected",
    "function_calls_finished",

livekit-agents/livekit/agents/utils/_message_change.py
davidzhao, line 1: we don't need to compute delta to sync anymore?

livekit-agents/livekit/agents/pipeline/io.py
theomonnom: no we don't, this isn't the transcript, this is the llm output

livekit-agents/livekit/agents/pipeline/io.py
theomonnom: It's internal but we can yes

livekit-agents/livekit/agents/pipeline/io.py
theomonnom: TextInput can be used with agent.generate_reply("my text")

theomonnom: We will need to check for the DataStream API but there is no reason for the text input to be streamed

livekit-agents/livekit/agents/utils/_message_change.py
theomonnom: it does?

theomonnom: https://github.com/livekit/agents/blob/dev-1.0/livekit-agents/livekit/agents/llm/utils.py

examples/minimal_worker.py
```diff
@@ -1,18 +1,27 @@
...
 
-    logger.info("connected to the room")
-    # add your agent logic here!
+    agent = PipelineAgent(
+        llm=openai.LLM(), stt=deepgram.STT(), tts=cartesia.TTS(), vad=silero.VAD.load()
+    )
+    agent.start()
+
+    # start a chat inside the CLI
+    chat_cli = ChatCLI(agent)
```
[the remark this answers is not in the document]
theomonnom: right, this isn't done, there is no Room implementation yet

livekit-agents/livekit/agents/multimodal/realtime.py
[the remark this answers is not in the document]
theomonnom: `realtime.RealtimeModel`?

livekit-agents/livekit/agents/http_server.py
```diff
@@ -18,9 +13,12 @@ def __init__(
         self._host = host
         self._port = port
         self._app = web.Application(loop=self._loop)
-        self._app.add_routes([web.get("/", health_check)])
```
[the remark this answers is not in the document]
theomonnom: it is moved

livekit-agents/livekit/agents/multimodal/realtime.py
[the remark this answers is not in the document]
davidzhao: sure.. I think the main difference between the models is:
* HTTP based vs Websocket
* Stateless vs stateful

so realtime seems like as good of a descriptor as any?

livekit-agents/livekit/agents/pipeline/io.py
davidzhao: agreed.. the same API is designed for chat as well. but just looking for a similar abstraction.. so it:
- by default, hooks up to a default topic/etc for the DataStreams API 
- allow users to override it too, so that they could hook it up to whatever source

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -0,0 +1,211 @@
...
+
+        self.__nb_playback_finished_needed = 0
+        self.__playback_finished_count = 0
+
+    def on_playback_finished(
+        self, *, playback_position: float, interrupted: bool
+    ) -> None:
+        """
+        Developers building audio sinks must call this method when a playback/segment is finished.
+        Segments are segmented by calls to flush() or clear_buffer()
```
longcw, line 72: Do we still need to call playback finished multiple times after `clear_buffer()` if there are multiple segments in the queue.

theomonnom: The PipelineAgent should guarantee we only process one segment at a time (If one segment was pushed, it'll call wait_for_playout and then push the next one). So clear_queue is always flushing the "last" segment in this case, so you call it once

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -0,0 +1,211 @@
...
+            self.__nb_playback_finished_needed += 1
+
+    @abstractmethod
+    def flush(self) -> None:
+        """Flush any buffered audio, marking the current playback/segment as complete"""
+        if self.__capturing:
+            self.__capturing = False
+
+    @abstractmethod
+    def clear_buffer(self) -> None:
```
longcw, line 123: should `flush` and `clear_buffer` be also async?
for the avatar use case, it needs to call `response = await room.local_participant.perform_rpc` to notify the remote participant.

theomonnom: in the current implementation, we assume it happens instantly (so we're not waiting). 
If we change those functions to async, we may have to add some more synchronization primitives where the agent uses it.

We don't have to be consistent, but as a note, TTS and STT doesn't have async methods for them too

longcw: so is that ok for the case `await room.local_participant.perform_rpc("interrupt_playback")` in `clear_buffer` we use a `create_task`?

theomonnom: IMO It is OK, but we can discuss further in case we think the devex for ppl developing custom Sinks is bad

livekit-agents/livekit/agents/pipeline/speech_handle.py
```diff
@@ -1,235 +1,84 @@
...
+            # will raise InvalidStateError if the future is already done (interrupted)
+            self._playout_done_fut.set_result(None)
 
-    def mark_nested_speech_done(self) -> None:
-        if self._nested_speech_done_fut.done():
-            return
-        self._nested_speech_done_fut.set_result(None)
+    async def wait_until_interrupted(self, aw: list[Awaitable]) -> None:
+        await asyncio.wait(
+            [*aw, self._interrupt_fut], return_when=asyncio.FIRST_COMPLETED
```
longcw, line 83: When I testing in py3.11 it raises `TypeError: Passing coroutines is forbidden, use tasks explicitly.` (https://docs.python.org/3/library/asyncio-task.html#waiting-primitives)

perhaps the following is needed
```python
    async def wait_until_interrupted(self, aw: list[Awaitable]) -> None:
        aw = [asyncio.create_task(task) for task in aw]
        await asyncio.wait(
            [*aw, self._interrupt_fut], return_when=asyncio.FIRST_COMPLETED
        )
        for task in aw:
            if not task.done():
                task.cancel()
```

longcw: or more complicated if there are tasks in `aw`
```python
    async def wait_until_interrupted(self, aw: list[Awaitable]) -> None:
        temp_tasks = []
        tasks = []
        for task in aw:
            if not isinstance(task, asyncio.Task):
                task = asyncio.create_task(task)
                temp_tasks.append(task)
            tasks.append(task)

        await asyncio.wait([*tasks, self._interrupt_fut], return_when=asyncio.FIRST_COMPLETED)
        for task in temp_tasks:
            if not task.done():
                task.cancel()
```

livekit-agents/livekit/agents/avatar/avatar_worker.py
```diff
@@ -0,0 +1,173 @@
...
+@dataclass
+class MediaOptions:
+    video_width: int
+    video_height: int
+    video_fps: float
+    audio_sample_rate: int
+    audio_channels: int
+
+
+# TODO(long): move to a plugin like livekit-plugins-avatar?
```
davidzhao, line 38: I think it should be part of the framework, vs a plugin. the logic is rather simple here.

livekit-agents/livekit/agents/avatar/avatar_worker.py
```diff
@@ -0,0 +1,173 @@
...
+class AvatarWorker:
```
davidzhao, line 39: nit: perhaps don't call it a "worker" to avoid confusion with the "agent worker".

how about calling it `AvatarRunner` or `AvatarHandler`, etc

livekit-agents/livekit/agents/pipeline/task.py
```diff
@@ -0,0 +1,181 @@
...
+    from .task_activity import TaskActivity
+
+
+class AgentTask:
+    def __init__(
+        self,
+        *,
+        instructions: str,
+        chat_ctx: NotGivenOr[llm.ChatContext] = NOT_GIVEN,
+        fnc_ctx: list[llm.AIFunction] = [],
```
davidzhao, line 34:
```suggestion
        functions: list[llm.AIFunction] = [],
```

livekit-agents/livekit/agents/pipeline/task.py
```diff
@@ -0,0 +1,181 @@
...
+        self._tts = tts
+        self._vad = vad
+        self._activity: TaskActivity | None = None
+
+    @property
+    def instructions(self) -> str:
+        return self._instructions
+
+    @property
+    def ai_functions(self) -> list[llm.AIFunction]:
```
davidzhao, line 76: nit: not loving the `ai_function` naming, but it should be consistent with what they are passing in.

```suggestion
    def functions(self) -> list[llm.AIFunction]:
```

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -1,1330 +1,302 @@
...
-                call_ctx = AgentCallContext.get_current()
-            except LookupError:
-                # no active call context, ignore
-                pass
-            else:
-                if isinstance(source, LLMStream):
-                    logger.warning(
-                        "LLMStream will be ignored for function call chat context"
+        async with self._lock:
+            if is_given(room):
```
longcw, line 153: should we add a warning if the agent starts without any input or output? Otherwise, the agent will be unresponsive if the user forgets to initiate it with `agent.start(room)` as the room is optional now.

davidzhao: yeah we should.

theomonnom: (Sry I removed it)

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -369,13 +369,13 @@ def __init__(
         participant: rtc.Participant | str,
         *,
         track: rtc.Track | rtc.TrackPublication | str | None = None,
-        capture_delta: bool = True,
+        is_stream: bool = True,
```
theomonnom, line 372: maybe name it `streamed`?

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -1,1379 +1,378 @@
...
-            # make sure at least some speech was played before committing the user message
-            # since we try to validate as fast as possible it is possible the agent gets interrupted
-            # really quickly (barely audible), we don't want to mark this question as "answered".
-            if (
-                speech_handle.allow_interruptions
-                and not is_using_tools
-                and (
-                    play_handle.time_played < self.MIN_TIME_PLAYED_FOR_COMMIT
-                    and not join_fut.done()
+            if self.input.audio is not None:
```
longcw: I do think we need a warning to tell user that they may forget the room here if no outputs set.

theomonnom: maybe we should wait 2 seconds, and warn if there is still not IO? wdyt

mrdrprofuroboros:
@theomonnom @davidzhao 
I have a question: I'm looking now at AudioRecognition and StreamAdapterWrapper and as far as I understand for the STTs that doesn't support streaming you effectively run 2 VADs in parallel - why so?

Moreover why have you decided to synchronize VAD and STT stream on event level in AudioRecognition and why do you even have VAD there in the first place? only to trigger the turn_detector model? couldn't it be done by END_OF_SPEECH event produced by STT?

I look at deepgram implementation and it has its own sort of VAD with all that AudioEnergyFilter, isn't it? So even when the STT supports streaming you don't send all audio (even silent) there blindly, what was the motivation here to write a custom VAD rather than using standard plugin approach?

livekit-agents/livekit/agents/voice/task_activity.py
```diff
@@ -0,0 +1,921 @@
...
+        pass
+
+    async def resume(self) -> None:
+        pass
+
+    async def start(self) -> None:
+        async with self._lock:
+            self._agent_task._activity = self
+            self._main_atask = asyncio.create_task(self._main_task(), name="_main_task")
+            self._audio_recognition = AudioRecognition(
```
mrdrprofuroboros, line 126: Why do you start AudioRecognition for RealtimeModels? Is it even used?

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -284,83 +282,48 @@ async def _execute_tools_task(
...
-                        fnc_out.exception = task.exception()
-                        out.append(_sanitize_function_output(fnc_call, fnc_out))
-                        return
-
-                    fnc_out.output = task.result()
-                    out.append(_sanitize_function_output(fnc_call, fnc_out))
-                    tasks.remove(task)
-
-                task.add_done_callback(_log_exceptions)
-            else:
```
jayeshp19: Does this mean function must be coroutine? what happens when it's not coroutine?

livekit-agents/livekit/agents/voice/task_activity.py
```diff
@@ -372,6 +377,9 @@ def _on_metrics_collected(self, ev: AgentMetrics) -> None:
 
     def _on_input_speech_started(self, _: llm.InputSpeechStartedEvent) -> None:
         log_event("input_speech_started")
+        if self._current_speech and not self._current_speech.allow_interruptions:
+            logger.warning("input_speech_started, but current speech is not interruptable")
+            return
```
theomonnom, line 382: This is too late, OpenAI always interrupt, so we can't do much. self.interrupt() is used here to sync the states

livekit-agents/livekit/agents/metrics/base.py
```diff
@@ -1,143 +1,129 @@
...
-    PipelineLLMMetrics,
-    PipelineTTSMetrics,
-    PipelineVADMetrics,
-    MultimodalLLMMetrics,
+    EOUMetrics,
 ]
+
+# @dataclass
+# class MultimodalLLMError(Error):
+#     type: str | None
```
niqodea, line 89: FYI this line in particular breaks [mypy](https://github.com/python/mypy) when installing `livekit-agents` in the virtual environment. This is because it's in the form `# type: ...` which is interpreted as a typing comment by mypy. What follows is `str | None`, which is not a valid directive, and the `mypy` command crashes.

In general, this could also be avoided by renaming `type` to `type_`. This would also adhere to the best practice of not shadowing the builtin `type` inside this class scope.

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -337,194 +359,177 @@ async def _tokenize_input():
...
+                        continue
 
+                    data = json.loads(msg.data)
+                    if data.get("audio"):
+                        b64data = base64.b64decode(data["audio"])
+                        decoder.push(b64data)
+
+                        if alignment := data.get("normalizedAlignment"):
+                            received_text += "".join(alignment.get("chars", [])).replace(" ", "")
+                            if received_text == expected_text:
```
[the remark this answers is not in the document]
jayeshp19: @fa9r  This is fixed in main branch, try with latest version :)

livekit-agents/livekit/agents/metrics/base.py
theomonnom: Thanks, for reporting this, will fix

PR #1371, opened 2025-01-14
playai: create new client only upon exception

jayeshp19:


livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
```diff
@@ -162,18 +162,20 @@ def __init__(
         self._opts = opts
         self._config = self._opts.tts_options
         self._mp3_decoder = utils.codecs.Mp3StreamDecoder()
+        self._client: PlayHTAsyncClient | None = None
```
davidzhao, line 165: should we create it during init as well?something like `ensure_client`

livekit-plugins/livekit-plugins-playai/livekit/plugins/playai/tts.py
```diff
@@ -194,6 +196,10 @@ async def _run(self) -> None:
                 )
         except Exception as e:
             raise APIConnectionError() from e
+        finally:
+            if self._client is not None:
```
davidzhao, line 200: this would effectively be the same, no? if the issue is with retries, then we should be resetting only when an exception is caught

btw, should we be closing the client each time? if that creates the websocket, we might end up in the same place.

jayeshp19: I think we would need to change the logic for all tts streaming providers as we're closing ws connection similarly for each request

theomonnom: The PipelineAgent is creating a new stream for every synthesis, so this isn't the right fix.
Let's create a thread on Slack?

PR #1374, opened 2025-01-15
update Cartesia example to use streaming

davidzhao:
according to Cartesia, their /bytes endpoint doesn't support continuations. streaming endpoint produces a more natural speech.

### DIFF
--- a/examples/text-to-speech/cartesia_tts.py
+++ b/examples/text-to-speech/cartesia_tts.py
@@ -35,7 +35,7 @@ async def entrypoint(job: JobContext):
         async for audio in stream:
             await source.capture_frame(audio.frame)
 
-    asyncio.create_task(_playback_task())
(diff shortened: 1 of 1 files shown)


examples/text-to-speech/cartesia_tts.py
```diff
@@ -29,14 +29,27 @@ async def entrypoint(job: JobContext):
...
 
-    await asyncio.sleep(4)
-    logger.info('Saying "Goodbye."')
-    async for output in tts.synthesize("Goodbye I hope to see you again soon."):
-        await source.capture_frame(output.frame)
+    async def _playback_task():
+        async for audio in stream:
+            await source.capture_frame(audio.frame)
+
+    asyncio.create_task(_playback_task())
```
theomonnom, line 38: keep a reference to the task otherwise lgtm

theomonnom (approved):

PR #1379, opened 2025-01-16
Plugin azure add voice styles

jmugicagonz:
added possibility to add styles to voices (see [documentation](https://learn.microsoft.com/en-us/azure/ai-services/speech-service/speech-synthesis-markup-voice#style-and-degree-example))


davidzhao (approved):
nice!

davidzhao:
@jmugicagonz could you also run `ruff format . && ruff check --fix .` ? we are running ruff 0.9.0

PR #1380, opened 2025-01-16
added echo agent example

davidzhao:


dsa (approved):
LG

bcherry (commented):
nice - this works flawlessly for me!

PR #1382, opened 2025-01-16
feat: support gemini llm in google plugin

jayeshp19:


livekit-plugins/livekit-plugins-google/livekit/plugins/google/_utils.py
```diff
@@ -0,0 +1,191 @@
...
+import base64
+import inspect
+import json
+from typing import Any, Dict, List, Optional, get_args, get_origin
+
+from livekit import rtc
+from livekit.agents import llm, utils
+from livekit.agents.llm.function_context import _is_optional_type
+
+from google.genai import types  # type: ignore
```
theomonnom, line 12: Let's do the workaround I sent on slack for the types

jayeshp19:
```
[mypy.google-genai]
follow_imports = normal
disallow_untyped_calls = True
```
didn't fix type issues, so added  `ignore_missing_imports = True` instead

mypy.ini
```diff
@@ -0,0 +1,5 @@
+[mypy]
+
+[mypy-google.genai.*]
+ignore_missing_imports = True
```
theomonnom, line 4: I guess this is just ignoring too?

jayeshp19: Yeah, we might just have to wait till their sdk add `py.typed`

davidzhao: I don't think their package is fully typed according to the last CI run: https://github.com/livekit/agents/actions/runs/12853743478/job/35837266341?pr=1382

While it's worth trying.. my 2c is to focus our typing exercise in our own code.. vs being concerned about how third party libs are being typed

davidzhao (approved):
lg

mypy.ini
theomonnom: Checking the errors inside the above CI, the types errors seems related to our code, not their code

davidzhao: ah my bad.. I misread the package as coming from them.. agreed we should fix our types.

jayeshp19: @theomonnom I've updated code to fix types, is this ready to merge?

livekit-agents/livekit/agents/cli/log.py
```diff
@@ -19,6 +19,7 @@
     "watchfiles",
     "anthropic",
     "websockets.client",
+    "root",
```
theomonnom, line 22:
```suggestion
```
we can't do that

theomonnom: root is the parent of all loggers..

jayeshp19: ah my bad. I will remove in next pr (gemini transcriber)

PR #1383, opened 2025-01-16
fix: address breaking change from google-genai>= 0.3.0

jayeshp19:


jayeshp19:
> should we lock the dependency versions?

Good Idea, just updated PR

PR #1384, opened 2025-01-16
PlayAI plugin: don't reconnect the websocket for each request (and update PlayAI Python SDK)

bryananderson:
This should resolve the latency that customers were experiencing due to reestablishing the WS before every requests.

This was done because of server-side WebSocket disconnect problems, which are now handled gracefully in the PlayAI Python SDK (version updated here)


jayeshp19:
Hi, thanks for raising the PR! I’ve incorporated your changes into this PR with some added cleanup and a changeset: https://github.com/livekit/agents/pull/1387.

PR #1391, opened 2025-01-20
avatar plugin based on v1.0

longcw:
1. AudioSink based on DataStream (https://github.com/livekit/python-sdks/pull/347)
2. Avatar worker example with video generation and av sync


livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
```diff
@@ -0,0 +1,223 @@
...
+    async def capture_frame(self, frame: rtc.AudioFrame) -> None:
+        """Capture and stream audio frame to remote worker"""
+        if not self._remote_participant:
+            raise RuntimeError("Worker participant not found")
+
+        if not self._stream_writer:
+            # start new stream
+            # TODO: any better option to send the metadata?
+            name = f"audio_{frame.sample_rate}_{frame.num_channels}"
+            self._stream_writer = await self._room.local_participant.stream_file(
```
davidzhao, line 61: this is a good use of stream extensions:

```
writer = await room.local_participant.stream_file("audio",
    extensions={"sample_rate": "48000", "channels": "1"})
```
or
```
writer = await room.local_participant.stream_file("audio",
    extensions={"audio_settings": json.dumps({"sample_rate": 48000, channels: 1})})
```

livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
```diff
@@ -0,0 +1,223 @@
...
+                    # segment completed
+                    pass
+                else:
+                    # process frame
+                    pass
+        """
+        # mark self as receiver
+        await self._room.local_participant.set_attributes({AUDIO_RECEIVER_ATTR: "true"})
+
+        self._remote_participant = await wait_for_participant(
```
davidzhao, line 135: it seems here we can just wait for participant.kind == agent?

if we wanted to handle multiple avatars in the room, then the integration should take in the controller's identity.

davidzhao (commented):
this looks great! just a few comments.

we'll also need some error handling in various parts.. how do both sides handle cases where the other side is disconnected. if the avatar participant is gone for longer than a reasonable timeout, then the agent would likely need to report that error and shutdown itself.

similarly.. if the controller is gone, the service on the other side might want to avoid consuming resources and also exit

livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
```diff
@@ -0,0 +1,223 @@
...
+
+        self._room.local_participant.register_rpc_method(
+            RPC_INTERRUPT_PLAYBACK, _handle_interrupt
+        )
+
+        def _handle_stream_received(
+            reader: rtc.FileStreamReader, remote_participant_id: str
+        ) -> None:
+            if remote_participant_id != self._remote_participant.identity:
+                logger.warning(
```
lukasIO, line 150: would we really want to warn on any other incoming file stream? that seems like a rather narrow use case for this plugin

livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
```diff
@@ -0,0 +1,223 @@
...
+
+    async def stream(self) -> AsyncIterator[Union[rtc.AudioFrame, AudioFlushSentinel]]:
+        while True:
+            await self._stream_received.wait()
+
+            while self._stream_readers:
+                reader = self._stream_readers.pop(0)
+                async for data in reader.stream_reader:
+                    yield rtc.AudioFrame(
+                        data=data,
```
lukasIO, line 190: this pattern would suggest that we're sure a single audio frame never exceeds `STREAM_CHUNK_SIZE` (~15kb)

livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
longcw: oh I see, I'll filter for the audio stream first, so other data streams can still be processed by other handlers.

Btw, what is the use case of the `file_name` in data stream, can I pass a tag and the metadata like `sample_rate` and `num_channels` using file name, or is there any better option for this.

lukasIO: see @davidzhao's comment above, the best option is the extensions map on the stream.

livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
longcw: Oh I see. the `extensions` is some kind of `metadata`? then what is the reason it is named `extensions`?

livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
longcw: For audio bytes, splitting large chunks into smaller chunks before sending is fine, and should be less than 15kb. but for other use cases, receiving a different number of chunks than it send may not be good behavior. Maybe add a size limit at the send side?

livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
```diff
@@ -0,0 +1,223 @@
...
+        self._room = room
+        self._remote_participant: Optional[rtc.RemoteParticipant] = None
+
+        self._stream_writer: Optional[rtc.FileStreamWriter] = None
+
+    async def start(self) -> None:
+        """Wait for worker participant to join and start streaming"""
+        # mark self as sender
+        await self._room.local_participant.set_attributes({AUDIO_SENDER_ATTR: "true"})
+        self._remote_participant = await wait_for_participant(
```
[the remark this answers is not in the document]
longcw: sounds good! on the avatar side, it wait for the audio stream with a particular name from the participant with `kind=='agent'`.

livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
lukasIO: I originally had a size limit in there, @theomonnom's wish was that we wouldn't enforce such a limit, but I agree, it might make things trickier if we don't have a sender side size limit

livekit-plugins/livekit-plugins-avatar/livekit/plugins/avatar/io.py
davidzhao: yeah.. I think `attributes` is probably a better name

examples/avatar/agent_worker.py
```diff
@@ -0,0 +1,110 @@
...
+    url: str  # LiveKit server URL
+    token: str  # Token for avatar worker to join
+
+
+async def launch_avatar_worker(
+    ctx: JobContext, avatar_dispatcher_url: str, avatar_identity: str
+) -> None:
+    """Wait for worker participant to join and start streaming"""
+    # create a token for the avatar worker
+    # TODO(long): do we need to set agent=True here? in playground if not the video track is not automatically displayed
```
bcherry, line 38: we might want a new kind to distinguish but we probably shouldn't join with the standard participant kind

examples/avatar/agent_worker.py
```diff
@@ -0,0 +1,110 @@
...
+    token = (
+        api.AccessToken()
+        .with_identity(avatar_identity)
+        .with_name("Avatar Worker")
+        .with_grants(api.VideoGrants(room_join=True, room=ctx.room.name, agent=True))
+        .with_metadata("avatar_worker")
+        .to_jwt()
```
bcherry, line 45: we should specify the identity of the "original" agent as well so that the integrator can add a guard on their RPC handlers methods (`if data.caller_identity !== agent_identity:  print("RPC call from unexpected participant"); return`)

examples/avatar/agent_worker.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+
+    logger.info(f"Sending connection info to avatar dispatcher {avatar_dispatcher_url}")
+    connection_info = AvatarConnectionInfo(
+        room_name=ctx.room.name, url=ctx._info.url, token=token
+    )
```
bcherry, line 51: we should specify the identity of the "original" agent as well so that the integrator can add a guard on their RPC handlers methods (`if data.caller_identity !== agent_identity:  print("RPC call from unexpected participant"); return`)

longcw: This makes sense! I added the check in the rpc hanlders.

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -82,6 +82,7 @@ def audio(self) -> AsyncIterator[rtc.AudioFrame] | None:
         async def _read_stream():
             async for event in self._audio_stream:
                 yield event.frame
+                await asyncio.sleep(0)
```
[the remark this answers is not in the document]
longcw: It was added for debugging an event loop issue. Will remove it.

PR #1398, opened 2025-01-21
gemini improvements: exception handling, transcription and ensure `content.parts` is non-empty in gemini context

jayeshp19:


livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -315,15 +316,38 @@ async def _run(self) -> None:
...
+        except ServerError as e:
+            raise APIStatusError(
+                "gemini llm: server error",
+                status_code=e.code,
+                body=e.message,
+                request_id=request_id,
+                retryable=retryable,
+            ) from e
+        except APIError as e:
+            raise APIStatusError(
```
theomonnom, line 344: You can use the Python syntax "raise e from exc" instead

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/transcriber.py
```diff
@@ -163,3 +165,9 @@ async def _recv_task():
             finally:
                 await utils.aio.gracefully_cancel(*tasks)
                 await self._session.close()
+
+
+def clean_transcription(text: str) -> str:
+    text = text.replace("\n", " ")
```
davidzhao, line 171: for trailing spaces, should we leave a space there? or remove entirely?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -315,15 +316,38 @@ async def _run(self) -> None:
...
                         )
                     )
-
+        except ClientError as e:
+            raise APIStatusError(
+                "gemini llm: client error",
+                status_code=e.code,
+                body=e.message,
+                request_id=request_id,
+                retryable=False if e.code != 429 else True,
```
davidzhao, line 333: I'm not sure if this is right. I think we do not want to retry 4xx errors, as it's typically indicating issue with the input. it's debatable if 429 should be retried.. but I think the answer is No, if rate limited.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/transcriber.py
jayeshp19: Trailing spaces will be removed automatically with `text.strip()`.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
davidzhao: got it.. i guess in that case it probably fell outside of the rate limit.. considering it's an edge case.. I think that makes sense.

however, the point remains. other errors like 5xx are definitely retryable.. so I think it's an incorrect behavior to only retry 429s

jayeshp19: should we keep false by default for client errors?

davidzhao: ah yeah you are right.. they've separated out client vs server errors.. I'm going to blame the fever for not seeing it.

PR #1404, opened 2025-01-23
Room IO for agent v1.0

longcw:


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -81,12 +81,18 @@ class ImageContent(BaseModel):
...
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
 
 class AudioContent(BaseModel):
     type: Literal["audio_content"] = Field(default="audio_content")
     frame: list[rtc.AudioFrame]
     transcript: Optional[str] = None
 
+    # temporary fix for pydantic before rtc.AudioFrame is supported
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
theomonnom, line 94: sounds good, will fix asap

livekit-agents/livekit/agents/pipeline/speech_handle.py
```diff
@@ -79,6 +79,17 @@ def _mark_playout_done(self) -> None:
...
+                temp_tasks.append(task)
+            tasks.append(task)
+
         await asyncio.wait(
-            [*aw, self._interrupt_fut], return_when=asyncio.FIRST_COMPLETED
+            [*tasks, self._interrupt_fut], return_when=asyncio.FIRST_COMPLETED
         )
+        for task in temp_tasks:
+            if not task.done():
+                task.cancel()
```
theomonnom, line 95: that's pretty annoying, will check if I can handle that differently (I'm on python 3.9 so it worked)

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -0,0 +1,261 @@
...
+from functools import partial
+from typing import Callable, Optional
+
+from livekit import rtc
+
+from ..utils import aio
+from .io import AudioSink, AudioStream, VideoStream
+
+
+class RoomAudioSink(AudioSink):
```
theomonnom, line 13: I think we should create another class that does both directly at the same time (like ChatCLI).
We can either create another class or merge both. Not sure about the naming but I was thinking about RoomIO

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -0,0 +1,261 @@
...
+            ),
+            enabled=audio_enabled,
+        )
+        self._video_streamer = self._RemoteTrackHandler(
+            track_to_stream=rtc.VideoStream,
+            enabled=video_enabled,
+        )
+
+        self._participant_ready = asyncio.Event()
+        self._room.on("participant_connected", self._on_participant_connected)
```
theomonnom, line 175: Let's handle room reconnection. If you get a "reconnected" event, you have to republish the local track.
Also we will have to "refind" the track we want to subscribe to

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -0,0 +1,261 @@
...
+            self, stream: rtc.AudioStream | rtc.VideoStream
+        ) -> None:
+            assert self._data_ch is not None
+            async for event in stream:
+                self._data_ch.send_nowait(event.frame)
+
+    def __init__(
+        self,
+        room: rtc.Room,
+        participant_identity: Optional[str] = None,
```
theomonnom, line 138: Let's add TrackPublishOption

livekit-agents/livekit/agents/pipeline/room_io.py
longcw: did you mean to set the input and output of agent when calling `RoomIO(agent)`. But I think this actually make it less flexible if user want to use some different input and output sources, for example for avatar, the input is `RoomInput`, the output could be `DataStreamAudioSink`. wdyt?

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -0,0 +1,261 @@
...
+                if self._participant:
+                    break
+
+    async def wait_for_participant(self) -> rtc.RemoteParticipant:
+        await self._participant_ready.wait()
+        assert self._participant is not None
+        return self._participant
+
+    @property
+    def audio(self) -> AudioStream | None:
```
davidzhao, line 196: neat.. lets us swap sources without disrupting downstream

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -0,0 +1,261 @@
...
+            self.on_playback_finished(
+                playback_position=self._pushed_duration, interrupted=True
+            )
+            self._pushed_duration = None
+
+
+class RoomInput:
+    """Creates video and audio streams from a remote participant in a LiveKit room"""
+
+    class _RemoteTrackHandler:
```
theomonnom, line 114: I'm not sure if we really need this utility class, directly creating a rtc.AudioStream and rtc.VideoStream seems clearer? Those streams are already compatible with io.*Stream

livekit-agents/livekit/agents/pipeline/room_io.py
theomonnom: I was thinking that if you don't want a Room as the Input. The users could simply only change the `agent.output.audio` and leave `agent.input.audio` empty (But they create only one class which is RoomIO)

theomonnom: Maybe this is handled directly inside a run method (like the ChatCLI).

Something like

`await room_io.run(input=False)`

longcw: I still think `RoomInput` and `RoomOutput` could be two classes for the reasons
1. they both have the audio and video sinks, if in a single RoomIO then there will be four options `room_io.run(input_audio=True, input_video=False, output_audio=True, output_video=False)`
2. the input and output can be different as the avatar use case: agent uses `RoomInput` and datastream output, avatar worker uses data stream input and `RoomOutput`
3. We can make the Room input and output as the default for the agent, then if user want to change one of them it's just `agent.output.audio = xxx`

livekit-agents/livekit/agents/pipeline/room_io.py
longcw: oh yeah, I was thinking the audio and video tracks need some different processing steps so moved the methods and track states into a class. I removed the handler class since video and audio actually share the same methods now.

To use `rtc.*Stream` directly as `agent.input.audio` or `agent.input.video`, we must wait for the tracks to be subscribed before setting `agent.input.audio = stream`, as they start as None. so the agent won't start before the track published for the case like camera or microphone is disabled when joining the room. do you have any ideas to solve this issue?

longcw: I saw there is a `AudioStream.from_participant`, maybe that's the answer?

longcw: ok, much cleaner now

livekit-agents/livekit/agents/pipeline/room_io.py
theomonnom: ah yep, that makes sense!

longcw: one question: regarding the output, do we want a `RoomOutput` class including both audio and video sinks, and optionally can enable the av sync? For now I think just a `RoomAudioSink` is okay as most of the use cases of agent is to publish audios to the room. @davidzhao  @theomonnom

davidzhao: I think we can just do a single audio sink for now.. in spirit of unblocking current functionality

theomonnom: Agree with dz, tho I think it would be cleaner to have only 2 class: RoomInput and RoomOutput. 
It may be harder for the users to understand if they have to initialize the RoomAudioSink

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+
+
+@dataclass
+class RoomInputOptions:
+    audio_enabled: bool = True
+    """Whether to subscribe to audio"""
+    video_enabled: bool = False
+    """Whether to subscribe to video"""
+    audio_sample_rate: int = 16000
```
theomonnom, line 97: Ideally, we automatically detect the sample_rate.

livekit-agents/livekit/agents/pipeline/room_io.py
longcw: I can make a RoomOutput without `RoomOutput().audio` that is an AudioSink.

examples/roomio_worker.py
```diff
@@ -0,0 +1,62 @@
...
+        ctx.room,
+        options=RoomInputOptions(
+            subscribe_audio=True,
+            subscribe_video=False,
+        ),
+    )
+    audio_output = RoomAudioSink(ctx.room, sample_rate=24000, num_channels=1)
+
+    agent.input.audio = room_input.audio
+    agent.output.audio = audio_output
```
theomonnom, line 42: Maybe something like this to be consistent:
```
    room_output = RoomOutput(ctx.room, sample_rate=24000, num_channels=1)

    agent.input.audio = room_input.audio
    agent.output.audio = room_output.audio
```

longcw: Yes, this is what I am think.

livekit-agents/livekit/agents/pipeline/room_io.py
theomonnom: The rtc.AudioStream isn't resampling, when setting a sample_rate + num_channels parameters, it expects to receive frames matching those options when calling `capture_frame`

longcw: how can we know what is the frame rate and num channels read from a remote track? or it's always the same value like 16000 and 1?

longcw: I think my question is whether the sample rate and number of channels provided to `rtc.AudioStream.from_participant` should match the track's sample rate, or if the audio frames will be automatically resampled to these values before putting to the stream?
```
@classmethod
    def from_participant(
        cls,
        *,
        participant: Participant,
        track_source: TrackSource.ValueType,
        loop: Optional[asyncio.AbstractEventLoop] = None,
        capacity: int = 0,
        sample_rate: int = 48000,
        num_channels: int = 1,
    ) -> AudioStream:
```

theomonnom: Oh sorry, I got confused, so the rtc.AudioStream does indeed resample (the default is always 48000 & 2 channels).
But the rtc.AudioSource doesn't (and this is where we should introduce a new sample_rate field on the baseclass)

theomonnom: You can directly use the frames coming from the rtc.AudioStream and inject them to the PipelineAgent. We already automatically resample on the STT for this usecase

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -0,0 +1,222 @@
...
+        self._publish_task: asyncio.Task | None = None
+        self._pushed_duration: float | None = None
+
+    async def start(self) -> None:
+        """Start publishing the audio track to the room"""
+        if self._publication:
+            return
+
+        # TODO: handle reconnected, do we need to cancel and re-publish? seems no
+        self._publication = await self._room.local_participant.publish_track(
```
theomonnom, line 45: After reconnection, we indeed need to republish the track

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -0,0 +1,289 @@
...
+            room: The LiveKit room to publish audio to
+            sample_rate: Sample rate of the audio in Hz
+            num_channels: Number of audio channels
+            capacity: Capacity of the internal audio queue, 0 means unlimited
+        """
+        super().__init__(sample_rate=sample_rate)
+        self._room = room
+
+        # buffer the audio frames as soon as they are captured
+        self._audio_buffer: aio.Chan[rtc.AudioFrame | None] = aio.Chan(maxsize=capacity)
```
longcw, line 196: I've added a buffer to ensure `caputre_frame` reads audio as quickly as possible. Initially this was for the tts transcription forwarder to read audio frames from the sink.

But now considering the need to integrate transcription at the agent level (https://github.com/livekit/agents/pull/1412#discussion_r1929716345), I may revert it to using `AudioSource.capture_frame` directly for simplicity, is there any use case to make the audio sink have an internal buffer?

theomonnom: When using an rtc.AudioSource we don't need a buffer, since there is already a buffer for it internally on the Rust side

longcw: Removed the buffer and using the internal queue of `AudioSource` now.

theomonnom (approved):
lgtm, I think there is a few parameters we shouldn't expose yet, but I'll push a commit on the v1.0 branch.
I'll go ahead and merge this PR!

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -0,0 +1,265 @@
...
+            self._audio_stream = None
+        if self._video_stream is not None:
+            await self._video_stream.aclose()
+            self._video_stream = None
+
+
+class RoomOutput:
+    """Manages audio output to a LiveKit room"""
+
+    def __init__(
```
davidzhao, line 148: TrackPublishOptions should be allowed here on the output.

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -80,10 +82,33 @@ def __init__(
     # They can all be overriden by subclasses, by default they use the STT/LLM/TTS specified in the
     # constructor of the PipelineAgent
 
-    def start(self) -> None:
+    async def start(
+        self,
+        room: Optional[rtc.Room] = None,
+        room_input_options: Optional[RoomInputOptions] = None,
```
davidzhao, line 88: ideally we should have an option of passing in a participant, so it's possible to target the participant. similarly, for PTT type of interactions, I would like to explicitly state that it's not bound to participants by default. (but then to call `link_participant(participant)` manually

davidzhao (commented):
🚀

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
longcw: makes sense. I'll update the input args.

PR #1410, opened 2025-01-24
fix false positive interruption tripping up certain LLMs

davidzhao:
LLMs like Anthropic and Gemini do not accept empty content as part of the user question. Using a continue marker to ensure output continues from point of interruption

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -737,12 +737,14 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
                     )
                 )
 
-        # we want to add this question even if it's empty. during false positive interruptions,
-        # adding an empty user message gives the LLM context so it could continue from where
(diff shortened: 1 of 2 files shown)


jayeshp19 (approved):

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -737,12 +737,14 @@ async def _synthesize_answer_task(
...
-        copied_ctx.messages.append(
-            ChatMessage.create(text=handle.user_question, role="user")
-        )
+        # when user_question is empty, it's due to a false positive interruption
+        # when this happens, we'd want to add a continue marker to the chat context.
+        # while some LLMs could deal with empty content during an inference request
+        # others would fail.
+        user_input = handle.user_question
+        if not user_input.strip():
+            user_input = "<continue>"
```
theomonnom, line 746: Is it opiniated, but I think it's OK as a temporary fix before v1.0

theomonnom (approved):

PR #1411, opened 2025-01-24
gemini-realtime: fix input audio sample rate

jayeshp19:


livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/transcriber.py
```diff
@@ -77,13 +84,28 @@ def __init__(
     def _push_audio(self, frame: rtc.AudioFrame) -> None:
         if self._closed:
             return
-        self._queue_msg(
-            types.LiveClientRealtimeInput(
-                media_chunks=[
-                    types.Blob(data=frame.data.tobytes(), mime_type="audio/pcm")
-                ]
+        if self._sample_rate != 16000:
+            bstream = utils.audio.AudioByteStream(
```
theomonnom, line 88: Is the intent to resample here? If so AudioByteStream isn't resampling. You should use the rtc.SoxrResampler

jayeshp19: Thankss

theomonnom (approved):

PR #1412, opened 2025-01-24
TTS/STT transcription sync for agent 1.0

longcw:
To have different transcription destinations, I extracted the TTS segment synchronization logic into a base class. Then created two forwarders derived from this base class: one sends the `TranscriptionSegment` to a room, and the other to `sys.stdout`. This could apply to a data stream easily as well. The usage is the same as before
```
# Start playout
tts_forwarder.segment_playout_started()

# async task
for chunk in text_stream:
    tts_forwarder.push_text(text)
tts_forwarder.mark_text_segment_end()

# async task
for frame in audio_stream:
    tts_forwarder.push_audio(frame)
tts_forwarder.mark_audio_segment_end()

# Playout finished
tts_forwarder.segment_playout_finished()
```


longcw:
How about integrate the forwarder like the following, where we can create any number of forwarders as tasks for the agent output. It needs to add callbacks for `capture_frame`, `flush`, and `clear_buffer` for the sinks. @theomonnom what do you think?

```python
asyncio.create_task(
    TTSRoomForwarder.run(audio_sink=agent.output.audio, text_sink=agent.output.text)
)

asyncio.create_task(
    TTSDataStreamForwarder.run(audio_sink=agent.output.audio, text_sink=agent.output.text)
)
```

examples/roomio_worker.py
```diff
@@ -53,6 +65,13 @@ async def entrypoint(ctx: JobContext):
     # await room_input.wait_for_participant()
     # await room_output.start()
 
+    # TTS transcription forward
+    agent.output.text = MockTextSink()
+    tts_forwarder = TTSRoomForwarder(ctx.room, participant=ctx.room.local_participant)
+    asyncio.create_task(
+        tts_forwarder.run(audio_sink=agent.output.audio, text_sink=agent.output.text)
```
longcw, line 72: I tested the idea `tts_forwarder.run(audio_sink=agent.output.audio, text_sink=agent.output.text)` here, it works but has an issue when interruption. Originally the tts forwarder is designed to be created for each speech handle and should be closed after the playout done or interrupted. The current implementation didn't handle the cleanup for the new speech handles so the interruptions doesn't work.

It's feasible to add the cleanup for each speech handles but I am thinking we may still need to add the a base `TTSSegmentsSync` into the `SpeechHandle` as the agent also need the `played_text` for the chat history. Then we could add an event like `tts_transcription_segment`, to pass the `rtc.TranscriptionSegment` to functions for forwarding to different destinations. @theomonnom @davidzhao wdyt?

livekit-agents/livekit/agents/transcription/tts_segments_sync.py
theomonnom, line 1: Do I need to review this file? Or it didn't change much from the main branch?

examples/roomio_worker.py
theomonnom: I think we should decouple the logic from the PipelineAgent/SpeechHandle.
Let's add another method that isn't streamed instead (something like `_truncate_message(full_generation: str, playout_duration: float) -> str`)

theomonnom:
I'm not sure if we need to keep the usage as before. It's very unlikely that people will use TTSSegmentsForwarder and STTSegmentsForwarder directly

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -113,30 +118,37 @@ def sample_rate(self) -> int | None:
     @abstractmethod
     async def capture_frame(self, frame: rtc.AudioFrame) -> None:
         """Capture an audio frame for playback, frames can be pushed faster than real-time"""
+        self.emit("capture_frame", frame, self.__capturing)
```
theomonnom, line 121: I don't think we should emit events for that

livekit-agents/livekit/agents/transcription/tts_segments_sync.py
longcw: Not too much right now.

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -113,30 +118,37 @@ def sample_rate(self) -> int | None:
...
     def flush(self) -> None:
         """Flush any buffered audio, marking the current playback/segment as complete"""
+        self.emit("flush")
         self.__capturing = False
 
     @abstractmethod
     def clear_buffer(self) -> None:
         """Clear the buffer, stopping playback immediately"""
+        self.emit("clear_buffer")
+        self.__capturing = False
```
theomonnom, line 136: This logic is tricky, but I don't think we should reset the capturing flag to False here.
The only difference is that it means we're creating  a new segment after each clear. I initially do it only on flush so it is easier to reason about, wdyt?

examples/roomio_worker.py
longcw: ok, this makes sense. the `_truncate_message` can have a simpler logic in that case.

livekit-agents/livekit/agents/pipeline/io.py
theomonnom: Instead I was thinking about a Chainable pattern? Something like 
```
room_output = RoomOutput()
transcription_sync = TranscriptionSyncIO(room_output) 
agent.output.text = transcription_sync.text
```

livekit-agents/livekit/agents/pipeline/io.py
longcw: Ok, so the `flush` is always called even the speech is interrupted? if so yeah the we shouldn't set the capturing to False.

theomonnom: Yes, flush is also called on interruption

livekit-agents/livekit/agents/pipeline/io.py
longcw: How about 
```python
transcription_sync = TranscriptionSyncIO(audio_sink)  # can be different audio sink
agent.output.text = transcription_sync.text
```
There could be some subclasses of `TranscriptionSyncIO` for different destinations.

And do we really need `TranscriptionSyncIO`, why not just a TextSink?

longcw: If we want to read the audio from an `AudioSink` we may still need a callback for the `AudioSink.capture_frame`?

theomonnom: Ideally, the TranscriptionSyncIO itself can also redirect its own output to another io.AudioSink/io.TextSink (making them chainable).

Yes, sorry, forgot audio but same idea here, we can also wrap the audio.

Let me know wdyt

longcw: Makes sense!
How about `TranscriptionSyncIO(audio_sink: AudioSink, text_sink: Optional[TextSink] = None)`, the text sink is optional, audio and text sinks can be from different IOs.

longcw:
> I'm not sure if we need to keep the usage as before. It's very unlikely that people will use TTSSegmentsForwarder and STTSegmentsForwarder directly.
> 
> Can you give an example of how we can access the synchronized transcription for the PipelineAgent on the python side? (Not published on the room yet) How can we combine it with the RoomIO?

It could be 
```python
room_output = RoomOutput()
transcription_sync = TranscriptionSyncIO(room_output.audio) 
agent.output.audio = transcription_sync.audio
agent.output.text = transcription_sync.text

segment_stream: AsyncIterator[rtc.TranscriptionSegment] = transcription_sync.segment_stream(self)
async for segment in segment_stream:
    ...
```

Then I can create some subclass of TranscriptionSyncIO for sending the segments to datastream or publish transcription.

Btw, if we have `_truncate_message` in the agent, we don't need to care about `played_text` in TranscriptionSyncIO anymore right? it will just take text in, generate `TranscriptionSegment`.

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -144,30 +157,57 @@ async def aclose(self) -> None:
             self._video_stream = None
 
 
-class RoomOutput:
+class RoomOutput(rtc.EventEmitter[Literal["transcription_segment"]]):
```
theomonnom, line 160: I would say to keep it simple and not emit any events for now.

theomonnom (approved):
otherwise lgtm

livekit-agents/livekit/agents/transcription/forwarder.py
```diff
@@ -0,0 +1,241 @@
...
+        """Forward transcription segment to the stream with real-time display."""
+        self._stream.write(segment.text)
+        self._stream.flush()
+
+        if segment.final:
+            self._stream.write("\n")
+            self._stream.flush()
+
+
+DEFAULT_TRANSCRIPTION_TOPIC = "lk.transcription"
```
lukasIO, line 180: should we change this to "lk.chat" instead or do we want an extra topic just for transcriptions?

PR #1420, opened 2025-01-28
Gemini realtime for 1.0

jayeshp19:


livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -246,118 +237,174 @@ def __init__(
...
-            async for msg in self._send_ch:
-                await self._session.send(msg)
+            async for msg in self._msg_ch:
+                await self._session.send(input=msg)
 
-            await self._session.send(".", end_of_turn=True)
+            await self._session.send(input=".", end_of_turn=True)
 
         @utils.log_exceptions(logger=logger)
         async def _recv_task():
```
theomonnom, line 336: Can you divide this task into multiple method?:

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -373,52 +420,41 @@ async def _recv_task():
...
-            await self._session.send(tool_response)
-
-            self.emit("function_calls_finished", [called_fnc])
+            # self._chat_ctx.append(text=content.text, role="user")
+            # TODO: implement sync mechanism to make sure the transcribed user speech is inside the chat_ctx and always before the generated agent speech
+
+    def _on_agent_speech_done(self, content: TranscriptionContent) -> None:
+        if not self._is_interrupted and content.response_id and content.text:
+            self.emit(
+                "agent_speech_transcription_completed",
```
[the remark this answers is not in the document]
theomonnom: I need to think more about it but I think we should add `conversation_item_added` and `conversation_item_removed` events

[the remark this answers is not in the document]
jayeshp19: That makes sense

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1094,31 +1094,33 @@ async def _read_messages(
...
-                        speech_handle=handle, user_initiated=False, source="tool_response"
-                    ),
-                )
-                self._create_speech_task(
-                    self._realtime_reply_task(
-                        speech_handle=handle,
-                        model_settings=ModelSettings(
-                            tool_choice=model_settings.tool_choice if not draining else "none",
+                # should we have separate flag for this?
+                if self.llm.capabilities.message_truncation:
```
theomonnom, line 1098: I don't think we're truncating here, why do we need this condition?
The following is just answering a tool response

jayeshp19: `message_truncation` isn't the correct flag here. With Gemini realtime, we don't need to explicitly call `generate_reply` to invoke generation. For tool responses, we're already sending `types.LiveClientToolResponse(function_responses=function_responses)` through `update_chat_ctx`, which triggers the reply.

examples/voice_agents/weather_agent.py
```diff
@@ -0,0 +1,75 @@
...
+        logger.info(f"getting weather for {latitude}, {longitude}")
+        url = f"https://api.open-meteo.com/v1/forecast?latitude={latitude}&longitude={longitude}&current=temperature_2m"
+        weather_data = {}
+        async with aiohttp.ClientSession() as session:
+            async with session.get(url) as response:
+                if response.status == 200:
+                    data = await response.json()
+                    # response from the function call is returned to the LLM
+                    weather_data = {
+                        "temperature": data["current"]["temperature_2m"],
```
theomonnom, line 52: This example is quite outdated (We don't use Annotated anymore), maybe we can use the extended-logger one (it is pretty small, with the Gemini real-time?)

Let's reserve the weather-example for a standalone demo that doesn't use Gemini. I see we don't have it on v1.0 yet, could you add it :) ?

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -49,6 +49,7 @@ class RealtimeCapabilities:
     message_truncation: bool
     turn_detection: bool
     user_transcription: bool
+    explicit_reply_after_tool_response: bool
```
theomonnom, line 52: Not a fan of this name, to be honest, I'm not even sure if this should be a capability. Can you maybe just check if the real-time session is Google on the agent_activity.py? 

Reason is that I'm pretty sure we will remove this field in the near-future once there is a better Gemini Live API

theomonnom (approved):

livekit-agents/livekit/agents/llm/realtime.py
jayeshp19: This would mean we've a dependency of google plugin in livekit agents

Issue #1422, opened 2025-01-28
User transcribed text is not cleared when before_llm_cb is returned Flase


s-hamdananwar:
Currently, `self._transcribed_text` is not cleared when `before_llm_cb` is returned false. This means that any new transcript will be appended to the "cancelled" user text as part of `self._transcribed_text`.

PR #1423, opened 2025-01-28
fix(pipeline_agent): clear user transcript when before_llm_cb returns false

s-hamdananwar:
fixes https://github.com/livekit/agents/issues/1422

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -753,6 +753,7 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
                 llm_stream = await llm_stream
 
             if llm_stream is False:
+                self._transcribed_text = ""
                 handle.cancel()
                 return
 
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -905,9 +905,9 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
                 return
 
             assert isinstance(speech_handle.source, LLMStream)
-            assert (
-                not user_question or speech_handle.user_committed
-            ), "user speech should have been committed before using tools"
+            assert not user_question or speech_handle.user_committed, (
+                "user speech should have been committed before using tools"
+            )
 
             llm_stream = speech_handle.source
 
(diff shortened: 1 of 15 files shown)


livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -753,6 +753,7 @@ async def _synthesize_answer_task(
                 llm_stream = await llm_stream
 
             if llm_stream is False:
+                self._transcribed_text = ""
```
davidzhao, line 756:
```suggestion
                # user chose not to synthesize an answer, so we do not want to
                # leave the same question in chat context. otherwise it would be
                # unintentionally committed when the next set of speech comes in.
                if len(self._transcribed_text) >= len(handle.user_question):
                    self._transcribed_text = self._transcribed_text[len(handle.user_question) :]
```

s-hamdananwar: @davidzhao just to confirm, i don't think we expect to have a scenario where len(self._transcribed_text) **>** len(handle.user_question) right? because in that case we would still have some text saved when the next speech comes.

davidzhao: it's not typical, but still possible that when running the before_llm_cb, new speech from the user has come in. so this is just trying to not lose any incoming text.

davidzhao (approved):

davidzhao:
@s-hamdananwar can you run `ruff format . && ruff check --fix .`?

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
s-hamdananwar: that makes sense, didn't think of that!

s-hamdananwar:
@davidzhao no idea why ruff check is failing on GH actions, running it locally on `livekit-agents/livekit/agents/pipeline/` (where the check is failing) gives me "All checks passed!"

SiyuanQi:
Hello, I am not sure this is the right approach. Just because we’re not responding immediately doesn’t mean we won’t need the full context later. If we clear it, we might lose important info that’s necessary for generating a correct response when we do decide to answer. I’d suggest keeping the transcribed text.

davidzhao:
it's definitely a tricky balance to strike.

in your use case, are you trying to delay synthesis?

SiyuanQi:
In my case, the agent would detect the right time to respond, instead of replying to every individual sentence from the user.

PR #1409, opened 2025-01-24
fix: ensure llm.FallbackAdapter executes function calls correctly

davidzhao:
fixes #1207

PR #1429, opened 2025-01-30
fix: fallbackadapter to correctly handle function calls

davidzhao:
when an LLM is in a failing state, sometimes that failure isn't discovered right away, but only when we attempt to iterate through the chunks. #1409 executes function correctly only when failure mode was not encountered

### DIFF
--- a/livekit-agents/livekit/agents/llm/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/fallback_adapter.py
@@ -131,8 +131,17 @@ class FallbackLLMStream(LLMStream):
         return self._current_stream.execute_functions()
 
     async def _try_generate(
-        self, *, llm: LLM, recovering: bool = False
+        self, *, llm: LLM, check_recovery: bool = False
     ) -> AsyncIterable[ChatChunk]:
+        """
+        Try to generate with the given LLM.
+
+        Args:
+            llm: The LLM instance to generate with
+            check_recovery: When True, indicates this is a background recovery check and the
+                          result will not be used. Recovery checks verify if a previously
+                          failed LLM has become available again.
+        """
         try:
             async with llm.chat(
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/llm/fallback_adapter.py
```diff
@@ -147,8 +148,11 @@ async def _try_generate(
                     retry_interval=self._fallback_adapter._retry_interval,
                 ),
             ) as stream:
-                self._current_stream = stream
+                first_chunk = True
                 async for chunk in stream:
+                    if first_chunk:
```
theomonnom, line 153: Let's also check recovering is False

theomonnom (approved):

livekit-agents/livekit/agents/llm/fallback_adapter.py
```diff
@@ -125,6 +125,7 @@ def fnc_ctx(self) -> FunctionContext | None:
         return self._current_stream.fnc_ctx
 
     def execute_functions(self) -> list[CalledFunction]:
+        # this function is unused, but putting it in place for completeness
```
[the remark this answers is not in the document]
theomonnom: Right I see tho it's a public API so it could still be used.

livekit-agents/livekit/agents/llm/fallback_adapter.py
davidzhao: hmm, why would that impact if this is set as the current stream?

theomonnom: When recovering is True, we're checking in the background to see if an LLM recovered but the generated output isn't used

davidzhao: got it.. I didn't realize that.. will make the change

PR #1431, opened 2025-01-30
improved TTFB metrics for streaming TTS

davidzhao:
for streaming TTS that were sentence based, it was counting the time to generate the first sentence in the TTFB metrics.


theomonnom (approved):
Makes sense, ideally I wanted subclass to not worry about metrics at all but I don't see another solution.
Agree that it doesn't make sense for the LLM generation to be included inside this delay.

PR #1432, opened 2025-01-30
Update examples

davidzhao:
using a new weather API, and using LLM as a geocoder


examples/voice-pipeline-agent/function_calling_weather.py
```diff
@@ -34,10 +32,19 @@ async def get_weather(
         location: Annotated[
```
bcherry, line 32: might be better to call this location_name

bcherry (approved):
nice, this is better

PR #1434, opened 2025-01-31
openai realtime: default create_response to True

davidzhao:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -143,7 +143,7 @@ class ServerVadOptions:
     threshold: float
     prefix_padding_ms: int
     silence_duration_ms: int
-    create_response: bool
```
jayeshp19, line 146: What's the reason behind this? we already have defalt servervad options [here](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py#L194)

davidzhao: when this wasn't set.. it was breaking our examples that customize VAD options. in general we should strive to reduce breakage to user ccode

jayeshp19: Oh I see, Makes sense,

jayeshp19: In that case we should also set default values for other arguments too?

jayeshp19: IMO we should remove defaults from [here](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py#L194) and keep it directly in dataclass

PR #1441, opened 2025-02-03
fix: Do not set fnc_ctx to None for anthropic at max depth

jayeshp19:
- Added an additional field in LLM capabilities class to check if model providers support function call history within chat context without needing function definitions.


livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -50,6 +50,7 @@ class Choice:
 @dataclass
 class LLMCapabilities:
     supports_choices_on_int: bool = True
+    supports_function_history_without_fnc_ctx: bool = True
```
theomonnom, line 53: Can we find a better name?

jayeshp19: how about `requires_explicit_tool_context`? feel free to suggest :)

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -50,7 +50,9 @@ class Choice:
 @dataclass
 class LLMCapabilities:
     supports_choices_on_int: bool = True
-    supports_function_history_without_fnc_ctx: bool = True
+    """check whether the LLM supports integer enums choices as function arguments"""
+    requires_persistent_functions: bool = True
+    """check whether the LLM requires a function definitions for past function calls in chat context"""
```
davidzhao:
```suggestion
    """if the LLM requires function definition when previous function calls exist in chat context"""
```

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -73,9 +76,9 @@ class LLM(
     rtc.EventEmitter[Union[Literal["metrics_collected"], TEvent]],
     Generic[TEvent],
 ):
-    def __init__(self) -> None:
+    def __init__(self, *, capabilities: LLMCapabilities) -> None:
```
davidzhao, line 79: we should default this to None, and init if none

jayeshp19: Nice, we should correct this for TTS and STT as well.

davidzhao (approved):
lg!

PR #1446, opened 2025-02-04
Gemini Realtime: Transcribe model audio via gemini api 

jayeshp19:
- add latest models from google
- updated testcase for array arguments

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
@@ -104,11 +104,11 @@ class RealtimeModel:
         self,
         *,
         instructions: str | None = None,
-        model: LiveAPIModels | str = "gemini-2.0-flash-exp",
+        model: LiveAPIModels | str = "gemini-2.0-flash",
         api_key: str | None = None,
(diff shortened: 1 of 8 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -382,7 +382,7 @@ def _on_input_speech_done(self, content: TranscriptionContent) -> None:
         # TODO: implement sync mechanism to make sure the transcribed user speech is inside the chat_ctx and always before the generated agent speech
 
     def _on_agent_speech_done(self, content: TranscriptionContent) -> None:
```
davidzhao, line 384: when interrupted, are we only transcribing until the moment of interruption?

davidzhao (approved):

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
davidzhao: got it. I think that's fine.. in the v1 branch, the synchronization/truncation logic will be downstream from the model.. model should just produce the entire thing.

PR #1447, opened 2025-02-04
Add prompt to STT options

ShayneP:
We should be able to set the prompt for STT providers to be give hints for transcriptions, this can help spelling.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
@@ -39,7 +39,7 @@ class _STTOptions:
     language: str
     detect_language: bool
     model: WhisperModels | str
-
+    prompt: str | None = None
 
 class STT(stt.STT):
     def __init__(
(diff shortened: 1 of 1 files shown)


davidzhao:
wanna run `ruff format .` on this? otherwise it lgtm

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -91,9 +93,12 @@ def update_options(
         *,
         model: WhisperModels | GroqAudioModels | None = None,
         language: str | None = None,
+        prompt: str | None = None,
     ) -> None:
         self._opts.model = model or self._opts.model
         self._opts.language = language or self._opts.language
+        if prompt is not None:
```
davidzhao, line 100: I think we need to update it back to None if a user wants to clear the previous prompt too.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -104,6 +109,7 @@ def with_groq(
         client: openai.AsyncClient | None = None,
         language: str = "en",
         detect_language: bool = False,
+        prompt: str = "Always spell 'Grok' as 'Groq'",
```
davidzhao, line 112: IMO this doesn't belong as a default prompt

ShayneP:
Closing as stale

PR #1448, opened 2025-02-05
fix agent transcription could not be disabled

### LINKED ISSUE #1444: agent transcriptions cannot be disabled in the frontend - is it a bug?
I'm trying to get transcriptions in the frontend using a component from the example https://docs.livekit.io/agents/voice-agent/transcriptions/

It works, but I can't disable forwarding the agent transcription to the client in the Agent settings. However, user transcriptions can be disabled

```
from livekit.agents.pipeline import VoicePipelineAgent, AgentTranscriptionOptions

assistant = VoiceAssistant(
        vad=silero.VAD.load(),
        stt=deepgram.STT(),
        llm=openai.LLM(model="gpt-4o-mini"),
        tts=openai.TTS(),
        chat_ctx=initial_ctx,
        transcription=AgentTranscriptionOptions(user_transcription=False, agent_transcription=False)
    )
```

With these settings I don't get user transcription, but I still get agent transcription

davidzhao:
fixes #1444

### DIFF
--- a/.changeset/five-rats-impress.md
+++ b/.changeset/five-rats-impress.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+fix agent transcription could not be disabled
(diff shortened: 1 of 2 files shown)


dsa (approved):
LG

PR #1457, opened 2025-02-06
[Cartesia] Adding base url option to Cartesia TTS plugin

chongzluong:
Per request from some of our customers, we want to provide the optionality to specify the base URL for the Cartesia API when using the Cartesia TTS plugin, including localhost for debugging purposes.

### DIFF
--- a/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
+++ b/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
@@ -64,12 +64,10 @@ class _TTSOptions:
     base_url: str
 
     def get_http_url(self, path: str) -> str:
-        protocol = "http" if "localhost" in self.base_url else "https"
-        return f"{protocol}://{self.base_url}{path}"
+        return f"{self.base_url}{path}"
 
     def get_ws_url(self, path: str) -> str:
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -92,6 +102,7 @@ def __init__(
             sample_rate (int, optional): The audio sample rate in Hz. Defaults to 24000.
             api_key (str, optional): The Cartesia API key. If not provided, it will be read from the CARTESIA_API_KEY environment variable.
             http_session (aiohttp.ClientSession | None, optional): An existing aiohttp ClientSession to use. If not provided, a new session will be created.
+            base_url (str, optional): The base URL for the Cartesia API. Defaults to "api.cartesia.ai".
```
davidzhao, line 105: seems that this is being provided as a host rather than URL.. since we are standardizing on `base_url`, I would recommend:
* keeping default to `https://api.cartesia.ai`
* in`get_ws_url`, replaces `^http` to `ws`

chongzluong: Ah mb the naming might be a lil off there (it's just we were previously using in our Python client). Sure I'll make that change

davidzhao (approved):
lgtm

PR #1460, opened 2025-02-07
support agent.say in before_llm_cb

longcw:
This pr supports calling `agent.say` inside a `before_llm_cb`, it will playout the content of say first then the returned llm stream
```python
async def before_llm_cb(
    agent: VoicePipelineAgent, chat_ctx: llm.ChatContext
) -> llm.LLMStream:
    speech_handle = await agent.say("say something")

    llm_stream = agent.llm.chat(chat_ctx=chat_ctx, fnc_ctx=agent.fnc_ctx)
    # await speech_handle.join()  # optionally wait for the playout finished
    return llm_stream
```

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
+++ b/livekit-agents/livekit/agents/pipeline/pipeline_agent.py
@@ -475,6 +475,8 @@ class VoicePipelineAgent(utils.EventEmitter[EventTypes]):
 
         if self._playing_speech and not self._playing_speech.nested_speech_done:
             self._playing_speech.add_nested_speech(new_handle)
+        elif self._speech_q:
+            self._speech_q[0].add_nested_speech(new_handle)
         else:
             self._add_speech_for_playout(new_handle)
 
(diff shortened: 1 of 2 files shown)


theomonnom (approved):

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
```diff
@@ -884,6 +932,7 @@ def _commit_user_question_if_needed() -> None:
                         "speech_id": speech_handle.id,
                     },
                 )
+        playing_lock.release()
```
jayeshp19, line 935: Should we wrap `playing_lock.release()` in `finally` block? 
something like 
```
await playing_lock.acquire()
try:
    ...(existing logic)
finally:
    playing_lock.release()

```

taqiycarbon:
Nice mate ! Took me a while to figure this out.

I went a different route and since before_llm_cb is always called after user_stopped_speaking I create the handle there.

```python
   async def delayed_starter_words_generator(event, agent):
        await event.wait()

        starter_words = None
        if agent.starter_words:
            starter_words = agent.starter_words
            agent.starter_words = None

        yield starter_words  # Then yield the response

    async def say_starter_words(agent):
        start_time = time.perf_counter()
        event = asyncio.Event()
        gen = delayed_starter_words_generator(event, agent)
        # Store the event on the agent to access it in before_llm_cb
        agent.starter_words_event = event
        await agent.say(gen)
```

and then all you need to do is 

```python
async def before_llm_cb(agent, chat_ctx):
    logger.debug("Entering before_llm_cb.")


    if hasattr(agent, 'starter_words_event'):
        filler = random.choice(FILLER_OPTIONS)
        agent.starter_words = filler
        agent.starter_words_event.set()
        agent.starter_words_event = None
```

I put this there in case the PR is not approved for people that would face this problem and are considering writing money patches 🐒 


Otherwise this is an important feature and it would be great to have it as part of the official SDK 👽

livekit-agents/livekit/agents/pipeline/pipeline_agent.py
longcw: it might be okay as if the main speech failed we should stop the whole speech handle and its nested speeches.

PR #1461, opened 2025-02-07
feat: Reuse synthesized stream for TTS

jayeshp19:


jayeshp19:
using connection_pool approach to maintain persistant websocket connections, hence closing this

PR #1464, opened 2025-02-07
chore(Deepgram STT): add nova-3 model to type literal

chasemcdo:
With the early access release of [Nova 3 from Deepgram](https://developers.deepgram.com/docs/models-languages-overview#nova-3) add the general model to the model type literal for STT


davidzhao (approved):
lgtm

PR #1469, opened 2025-02-10
fix openai realtime response create

longcw:
Fixed a bug since https://github.com/livekit/agents/pull/1320 that overwrites the instructions with empty unexpectedly during response creation in the OpenAI Realtime model.


davidzhao (approved):
nice catch!

Issue #1470, opened 2025-02-10
Cannot use o3-mini model


nuttyroll:
When we use the o3-mini model of open ai we get an error 

livekit.agents._exceptions.APIStatusError: Error code: 400 - {'error': {'message': "Unsupported parameter: 'temperature' is not supported with this model.", 'type': 'invalid_request_error', 'param': 'temperature', 'code': 'unsupported_parameter'}}

We will have to update LLM plugin to exclude temperature in the case of o3 and other models which does not support temperature parameter.

PR #1474, opened 2025-02-10
fix: [openai] only send params when set

jayeshp19:
fixes: https://github.com/livekit/agents/issues/1470
and minor improvements


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -188,15 +188,8 @@ def with_cerebras(
 
         ``api_key`` must be set to your Cerebras API key, either using the argument or by setting
         the ``CEREBRAS_API_KEY`` environmental variable.
-        @integrations:cerebras:llm
         """
-
-        api_key = api_key or os.environ.get("CEREBRAS_API_KEY")
```
davidzhao, line 194: 👍

davidzhao (approved):
lgtm

PR #1475, opened 2025-02-10
Allow shutdown to take reason

martin-purplefish:
I made this backward compatible, which I think is a nicer interface anyway since people don't always care about the shutdown reason.

### DIFF
--- a/livekit-agents/livekit/agents/job.py
+++ b/livekit-agents/livekit/agents/job.py
@@ -159,7 +159,7 @@ class JobContext:
             self._shutdown_callbacks.append(callback)  # type: ignore
         else:
 
-            async def wrapper(reason: str) -> None:
+            async def wrapper(_: str) -> None:
                 await callback()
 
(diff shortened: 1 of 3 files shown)


quang-purplefish (approved):

theomonnom (approved):

livekit-agents/livekit/agents/job.py
```diff
@@ -143,9 +145,24 @@ def agent(self) -> rtc.LocalParticipant:
...
+            Callable[[], Coroutine[None, None, None]],
+            Callable[[str], Coroutine[None, None, None]],
+        ],
     ) -> None:
-        self._shutdown_callbacks.append(callback)
+        """
+        Add a callback to be called when the job is shutting down.
+        Optionally the callback can take a single argument, the shutdown reason.
+        """
+        if callback.__code__.co_argcount > 0:
```
egoldschmidt, line 158: FYI I think this misses the case where you pass in a reference to a bound function w/ `self` first arg

PR #1476, opened 2025-02-10
Adding "str" check on before_llm_cb response

vmsobral:
The reason for that PR is that someone could want to do LLM processing manually before sending the answer to the agent to be synthesized. The only way to do this is with `before_llm_cb`.

For example, I have two tasks that need LLM processing and are self-excluding. I want to do them in parallel so that I do not increase the latency. The choice between those 2 tasks is done by another LLM task, like this:
image

By the end, LLM 1 and LLM 2 are already processed so I should have a final text or string and I do not want to create another LLM chat context and make the processing all over again. Since `_synthesize_agent_speech` already supports `llm_stream` as str, this could be accomplished by prevent the fallback from creating a default cb when `llm_stream` is `str`.


davidzhao:
please rebase the PR against 0.x first

PR #1482, opened 2025-02-12
Update azure/stt.py - added speech_endpoint in init of azure STT class so that it's value can be initialized and passed to azure speech sdk properly

sarthakgoyal23:
When these changes were merged
https://github.com/livekit/agents/pull/1362

speech_endpoint property was added but not used in init and also not set in the config and thereby it was not passed to azure SpeechConfig class.

In this PR added the parameter in init and then later initialized in _config variables which is later used while calling SpeechConfig


mbadala:
Hi any reason why this pr was closed? we faced the same issue when using endpoint instead of using region and key

PR #1486, opened 2025-02-13
Add handlers for supported synthesis events for Azure TTS

anishnag:
Adding handlers for each of the `SpeechSynthesizer` subscribable events according to the [docs](https://learn.microsoft.com/en-us/azure/ai-services/speech-service/how-to-speech-synthesis?tabs=browserjs%2Cterminal&pivots=programming-language-python):
- `BookmarkReached`
- `SynthesisCanceled`
- `SynthesisCompleted`
- `SynthesisStarted`
- `Synthesizing`
- `VisemeReceived`
- `WordBoundary`

### DIFF
--- a/.changeset/weak-weeks-shave.md
+++ b/.changeset/weak-weeks-shave.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-azure": patch
+---
+
+Add handlers for supported synthesis events for Azure TTS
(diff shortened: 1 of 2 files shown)


theomonnom:
Hey, can you fix the Ruff CI. 

Also curious, why do you need those callbacks? It seems hard to correlate them with the ChunkedStream (Since the frames will be read asynchronously)

anishnag:
> Hey, can you fix the Ruff CI.
> 
> Also curious, why do you need those callbacks? It seems hard to correlate them with the ChunkedStream output (Since the frames will be read asynchronously)

@theomonnom For now, we do client-side syncing given these events. Eventually we'll want to make it work on the agent side which would probably look like monitoring the audio frames and relaying the events based on running duration.

theomonnom (approved):

PR #1487, opened 2025-02-13
fix(cli): update main_file path to use current directory

imsakg:
Use `os.getcwd()` to set the `main_file` path, ensuring it points to the current working directory instead of the script directory. This change solves problems on script execution context handling in managed virtual enviroments such as rye/uv/poetry.

edit: This causes problem when `--watch` enabled. It's looking for changes on `.venv/bin` dir.

### DIFF
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -1,7 +1,7 @@
 import asyncio
+import os
 import pathlib
 import signal
-import sys
 
 import click
 from livekit.protocol import models
@@ -216,7 +216,7 @@ def _run_dev(
         from .watcher import WatchServer
 
         setup_logging(log_level, args.devmode)
-        main_file = pathlib.Path(sys.argv[0]).parent
+        main_file = pathlib.Path(os.getcwd())
 
         async def _run_loop():
             server = WatchServer(


davidzhao:
I think script parent directory is where we should be watching.. how are you running the agent?

this PR would cause issues if my directory is structured as:

```
myagent/
    agent.py
    ...
```

and I ran `python myagent/agent.py`

imsakg:
Hey @davidzhao,

What type of issue are you encountering with this PR? My project structure is exactly the same, yet I'm facing issues while this PR has none.

Here’s my project structure:

-  my_fancy_project
  - .venv
  - src
    - agent.py
    - ...
  - pyproject.toml

I'm using a managed environment with [Rye](https://rye.astral.sh), so I run my project like this: `rye run agent dev`.

When I run my project, I get `/Users/my_user/Projects/my_fancy_project/.venv/bin` as the watch directory. The agent is looking in that directory for changes, but my app does not update on-live whenever I make changes to the source code.

However, it works fine without Rye, like in your example. I tried the exact setup by running `python myagent/agent.py`, as you mentioned, and I got this:

```2025-02-14 12:14:55,705 - DEV  livekit.agents - Watching /Users/my_user/Projects/myagent/```

**Comment by davidzhao** [comment]
> However, this PR works fine without Rye, **like in your example**.

Here's what I get, running in `~/`

```
% python myagent/agent.py dev
2025-02-15 16:17:33,641 - DEBUG asyncio - Using selector: KqueueSelector 
2025-02-15 16:17:33,642 - DEV  livekit.agents - Watching /Users/davidzhao
```

imsakg:
I completely understand your concern, but this is a common side effect of [chdir](https://man7.org/linux/man-pages/man2/chdir.2.html) in many projects. The path should be where you execute the Python interpreter. In our case, we should run it from the root of our project folder.

Perhaps we could add an optional `watch_dir` parameter. What do you think?

theomonnom (approved):

theomonnom:
Ah I merged too fast, true that it could be an issue if you start your python script from a completely different location

PR #1488, opened 2025-02-13
fix aio create task references

longcw:


livekit-agents/livekit/agents/avatar/avatar_worker.py
```diff
@@ -56,6 +56,7 @@ def __init__(
         self._audio_stream_received: asyncio.Event = asyncio.Event()
         self._playback_position = 0.0
         self._audio_playing = False
+        self._clear_buffer_tasks: set[asyncio.Task] = set()
```
theomonnom, line 59: feel free to just name it _tasks

theomonnom (approved):

livekit-agents/livekit/agents/avatar/avatar_worker.py
longcw: updated

PR #1493, opened 2025-02-14
rename to avatar runner and cleanup

longcw:


examples/avatar/agent_worker.py
```diff
@@ -26,22 +30,24 @@
...
 ) -> None:
     """Wait for worker participant to join and start streaming"""
     # create a token for the avatar worker
-    # TODO(long): do we need to set agent=True here? in playground if not the video track is not automatically displayed
+    agent_identity = ctx.room.local_participant.identity
     token = (
         api.AccessToken()
         .with_identity(avatar_identity)
         .with_name("Avatar Worker")
         .with_grants(api.VideoGrants(room_join=True, room=ctx.room.name, agent=True))
```
davidzhao, line 49: I think we don't want agent=True, but rather setting the participant Kind to Agent: `with_kind("agent")`

examples/avatar/avatar_runner.py
```diff
@@ -188,8 +198,8 @@ async def main(room: rtc.Room):
             audio_sample_rate=24000,
             audio_channels=1,
         )
-        video_generator = VideoGenerator(media_options)
-        worker = AvatarWorker(
+        video_generator = MyVideoGenerator(media_options)
+        worker = AvatarRunner(
```
davidzhao, line 202: runner

davidzhao (approved):
a couple of nits.. otherwise lgtm

examples/avatar/agent_worker.py
longcw: updated

longcw: also add transcription sync for the avatar example

PR #1497, opened 2025-02-14
Add text stream sink and multi text sink

lukasIO:
depends on https://github.com/livekit/python-sdks/pull/362

### DIFF
--- a/livekit-agents/livekit/agents/pipeline/io.py
+++ b/livekit-agents/livekit/agents/pipeline/io.py
@@ -141,6 +141,18 @@ class TextSink(ABC):
         ...
 
 
+class MultiTextSink(TextSink):
+    def __init__(self, sinks: list[TextSink]) -> None:
+        self._sinks = sinks
+
+    async def capture_text(self, text: str) -> None:
+        await asyncio.gather(*[sink.capture_text(text) for sink in self._sinks])
+
+    def flush(self) -> None:
+        for sink in self._sinks:
+            sink.flush()
+
+
 # TODO(theomonnom): Add documentation to VideoSink
 class VideoSink(ABC):
     @abstractmethod
--- a/livekit-agents/livekit/agents/pipeline/io.py
+++ b/livekit-agents/livekit/agents/pipeline/io.py
@@ -131,7 +131,7 @@ class AudioSink(ABC, rtc.EventEmitter[Literal["playback_finished"]]):
 
 class TextSink(ABC):
     @abstractmethod
-    async def capture_text(self, text: str, *, segment_id: str | None = None) -> None:
+    async def capture_text(self, text: str) -> None:
         """Capture a text segment (Used by the output of LLM nodes)"""
         ...
 
@@ -141,14 +141,12 @@ class TextSink(ABC):
         ...
 
 
-class MultiTextSink(TextSink):
-    def __init__(self, sinks: list[TextSink]) -> None:
+class ParallelTextSink(TextSink):
+    def __init__(self, *sinks: TextSink) -> None:
         self._sinks = sinks
 
-    async def capture_text(self, text: str, *, segment_id: str | None = None) -> None:
-        await asyncio.gather(
-            *[sink.capture_text(text, segment_id=segment_id) for sink in self._sinks]
-        )
+    async def capture_text(self, text: str) -> None:
+        await asyncio.gather(*[sink.capture_text(text) for sink in self._sinks])
 
     def flush(self) -> None:
         for sink in self._sinks:
--- a/livekit-agents/livekit/agents/pipeline/io.py
+++ b/livekit-agents/livekit/agents/pipeline/io.py
@@ -131,7 +131,7 @@ class AudioSink(ABC, rtc.EventEmitter[Literal["playback_finished"]]):
 
 class TextSink(ABC):
     @abstractmethod
-    async def capture_text(self, text: str) -> None:
+    async def capture_text(self, text: str, *, segment_id: str | None = None) -> None:
         """Capture a text segment (Used by the output of LLM nodes)"""
         ...
 
@@ -145,8 +145,10 @@ class MultiTextSink(TextSink):
     def __init__(self, sinks: list[TextSink]) -> None:
         self._sinks = sinks
 
-    async def capture_text(self, text: str) -> None:
-        await asyncio.gather(*[sink.capture_text(text) for sink in self._sinks])
+    async def capture_text(self, text: str, *, segment_id: str | None = None) -> None:
+        await asyncio.gather(
+            *[sink.capture_text(text, segment_id=segment_id) for sink in self._sinks]
+        )
 
     def flush(self) -> None:
         for sink in self._sinks:
--- a/livekit-agents/livekit/agents/pipeline/room_io.py
+++ b/livekit-agents/livekit/agents/pipeline/room_io.py
@@ -660,11 +660,11 @@ class RoomTranscriptEventSink(TextSink):
         self._pushed_text = ""
         self._current_id = utils.shortuuid("SG_")
 
-    async def capture_text(self, text: str) -> None:
+    async def capture_text(self, text: str, *, segment_id: str | None = None) -> None:
         if not self._capturing:
             self._capturing = True
             self._pushed_text = ""
-            self._current_id = utils.shortuuid("SG_")
+            self._current_id = segment_id or utils.shortuuid("SG_")
 
         if self._is_stream:
             self._pushed_text += text
@@ -758,10 +758,13 @@ class DataStreamSink(TextSink):
         self._participant_identity = identity
         self._latest_text = ""
 
-    async def capture_text(self, text: str) -> None:
+    async def capture_text(self, text: str, *, segment_id: str | None = None) -> None:
+        if segment_id is not None and segment_id != self._current_id:
+            self.flush()
+
         self._latest_text = text
         if not self._is_capturing:
-            self._current_id = utils.shortuuid("SG_")
+            self._current_id = segment_id or utils.shortuuid("SG_")
             self._is_capturing = True
 
         try:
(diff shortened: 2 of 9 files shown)


livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -105,8 +105,19 @@ async def start(self, agent: Optional["PipelineAgent"] = None) -> None:
         agent.input.video = self.video
         if self._options.forward_user_transcript:
             # TODO: support multiple participants
-            self._text_sink = RoomTranscriptEventSink(
-                room=self._room, participant=self._participant, is_stream=False
+            self._text_sink = MultiTextSink(
```
longcw, line 108: should we make this configurable that enables one or both of the sinks?

lukasIO: Not sure what other decisions were made for the io API, but in my opinion it would be nice for users to be able to set a TextSink themselves

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -713,3 +733,89 @@ def _on_track_published(
 
         if track.source == rtc.TrackSource.SOURCE_MICROPHONE:
             self._track_id = track.sid
+
+
+class DataStreamSink(TextSink):
```
longcw, line 738: perhaps move this to the datastream_io

livekit-agents/livekit/agents/pipeline/transcription/synchronizer.py
```diff
@@ -358,7 +358,7 @@ def _on_text_updated(self, segment: rtc.TranscriptionSegment) -> None:
             self._current_segment_id = segment.id
 
         async def _capture_text():
-            await self._base_text_sink.capture_text(segment.text)
+            await self._base_text_sink.capture_text(segment.text, segment_id=segment.id)
```
longcw, line 361: this won't work for other text sinks

longcw: There is already a check for the segment id before capture
```
        if self._current_segment_id != segment.id:
            self._base_text_sink.flush()
            self._current_segment_id = segment.id
```

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -713,3 +733,89 @@ def _on_track_published(
...
+    def set_participant(
+        self,
+        participant: rtc.Participant | str,
+        track: rtc.Track | rtc.TrackPublication | str | None = None,
+    ) -> None:
+        identity = participant if isinstance(participant, str) else participant.identity
+        self._participant_identity = identity
+        self._latest_text = ""
+
+    async def capture_text(self, text: str, *, segment_id: str | None = None) -> None:
```
longcw, line 767: I prefer to not add `segment_id` here to keep the same api for all text sinks, user should use `flush()` to mark the end of the segment ideally.

Though actually I also want a segment id here in case the text with different ids comes not in order like `id1, id2, id1, ...`. wdyt @theomonnom ?

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -131,7 +131,7 @@ def clear_buffer(self) -> None:
 
 class TextSink(ABC):
     @abstractmethod
-    async def capture_text(self, text: str) -> None:
+    async def capture_text(self, text: str, *, segment_id: str | None = None) -> None:
```
longcw, line 134: ok, it was added here.

lukasIO: yeah, I think having it is better than not having it, guessing your ok with the change as long as it's part of the base class here?

livekit-agents/livekit/agents/pipeline/room_io.py
lukasIO: just wanted to keep all the text sinks together, but don't have a strong opinion if you think it makes more sense to move it to datastream_io. 

maybe a text_sinks.py makes sense where we have all of them?

livekit-agents/livekit/agents/pipeline/io.py
longcw: I still have a concern that the `segment_id` is conflicted with `flush`. If calling `flush` means the end of the segment, what is the "correct" behavior if the segment_id is the same as before after flush?

lukasIO: I don't think that's a problem that's caused by the segment_id being part of the params.

Is you concern about replacing `current_id` with `segment_id`?
I just wanted to avoid any instances where there are multiple places trying to call `capture_text` with different segment ids. In my head the place to handle this is the sink itself, but we can also leave it as is if we can be certain that every caller checks for the segment_id and flushes if necessary

longcw: not about the current_id or segment_id, but just not clear what is the expect behavior if the same segment_id captured after flush. Sounds like the flush and segment_id are overlapped in functionality.

longcw: I'd like to keep flush without segment_id, and let the caller decide when to call the flush. Actually it seems that only user transcription has the custom segment id, the agent transcription is handled by the agent pipeline fully based on flush.

theomonnom: I don't think we want to add segment_id here. Why couldn't we generate it downstream?

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -141,6 +141,20 @@ def flush(self) -> None:
         ...
 
 
+class MultiTextSink(TextSink):
+    def __init__(self, sinks: list[TextSink]) -> None:
+        self._sinks = sinks
+
+    async def capture_text(self, text: str, *, segment_id: str | None = None) -> None:
+        await asyncio.gather(
```
theomonnom, line 149: Let's keep the public API minimal for V1.0, I wouldn't expose it for now and keep this class private.

livekit-agents/livekit/agents/pipeline/room_io.py
theomonnom: Should this be merged inside the RoomOutput? I don't think we need to have it in a separate class. Also we have an ongoing discussion about merging RoomInput and RoomOutput together in Slack

livekit-agents/livekit/agents/pipeline/io.py
longcw: I removed the segment id here https://github.com/livekit/agents/pull/1497/commits/446e3a9e930ec25c13ae587f14e34083b0a9ddb1, let's merge this one if everyone feels ok, and I'll do a refactor to merge the RoomInput and Output and also the topic things for the user text inputs.

lukasIO: hm, I don't think this will work. 
Did you test this and what it looks like for data streams?

lukasIO: You need a stable Id across multiple captures for STT in the case of `is_delta_stream: false`. you want to send the corrected/appended transcription with the same id as before

longcw: @lukasIO what did you mean it won't work. The segment id is the unchanged when delta is false until `flush()` called. The `flush` is called only received the `final` flag or the segment id from the text synchronizer is changed.

longcw: I didn't see a logic change in the DatastreamSink from this commit https://github.com/livekit/agents/commit/446e3a9e930ec25c13ae587f14e34083b0a9ddb1.

For agent transcription with delta=True, the change of the `segment_id` is checked before calling the `sink.capture_text()` here https://github.com/livekit/agents/blob/lukas/ds-sink/livekit-agents/livekit/agents/pipeline/transcription/synchronizer.py#L356-L363,
```python
        if self._current_segment_id != segment.id:
            self._base_text_sink.flush()
            self._current_segment_id = segment.id

        async def _capture_text():
            await self._base_text_sink.capture_text(segment.text)
            if segment.final:
                self._base_text_sink.flush()

        task = asyncio.create_task(_capture_text())
```

and for user transcription with delta=False, the capture is called here https://github.com/livekit/agents/blob/lukas/ds-sink/livekit-agents/livekit/agents/pipeline/room_io.py#L420-L426 so it only flush after the final transcript received. 
```python
        async def _capture_text():
            if ev.alternatives:
                data = ev.alternatives[0]
                await self._text_sink.capture_text(data.text)

            if ev.type == stt.SpeechEventType.FINAL_TRANSCRIPT:
                self._text_sink.flush()
```

lukasIO: that means it needs to be guaranteed that `captureText()` is not called for the same segment_id again after it has been flushed once with that segment_id. 
Is that guaranteed?

longcw: For user transcription, yes, there is no segment id naturally in that case.
For agent transcription, it's not guaranteed but it streams delta so it should be fine.

longcw: could you point out what is the logic difference introduced by the commit if that's the concern?

lukasIO: the difference is that previously if a `segment_id` was provided to `capture_text` it would re-use that id for sending the text stream. 

if we create a new id after a `flush` it will result in streams getting a new id, even though it might be originating from the same `segment_id`. 
On the receiving side a new id would show up as a new message, while having the same id for the same segments would prompt the receiving side to replace the previous message instead of adding a new one. 
Does that make it clearer? 
Happy to jump on a call and discuss

lukasIO (commented):
lgtm! thanks for fixing the outstanding issues 🙏

theomonnom (approved):

PR #1498, opened 2025-02-14
support link and unlink participant in RoomInput

longcw:


livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -62,111 +62,165 @@ def __init__(
...
+        self._streams: dict[
+            Literal["audio", "video"], Optional[rtc.VideoStream | rtc.AudioStream]
+        ] = {
+            "audio": None,
+            "video": None,
+        }
+        self._stream_ready: dict[Literal["audio", "video"], asyncio.Event] = {
+            "audio": asyncio.Event(),
+            "video": asyncio.Event(),
+        }
```
theomonnom, line 89: IMO it is actually harder to read when using this pattern of dicts

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -62,111 +62,165 @@ def __init__(
...
-                yield event.frame
+        stream: AsyncIterator[rtc.VideoFrame] = self._read_stream("video")
+        return stream
+
+    @property
+    def active_participant(self) -> rtc.RemoteParticipant | None:
+        """Get currently active participant"""
+        return self._active_participant
+
+    def link_participant(self, participant: rtc.RemoteParticipant | str) -> None:
```
theomonnom, line 135: When setting a string and the participant isn't found, we should listen to participant_connected events. (This is useful if the participant disconnect/reconnect).


Also nit: wdyt of `update_participant` instead?

longcw: To handle participant disconnect/reconnect, maybe we keep track the activate participant id and if the id disconnected and reconnected we still link to it.

We also provide a method to set the target participant id, if the participant is connected then switch to it, if not we deactivate the current participant and wait for it?

how about `update_participant(participant, on_participant_missing: Literal["wait", "raise"] = "wait")`

livekit-agents/livekit/agents/pipeline/room_io.py
longcw: There will be a `text_stream` later, with text, audio, and video sharing similar management functions, like start, close, and ready events. Do you have any suggestions for the code clarity?

davidzhao: wdyt about a intermediate type: `StreamHolder` with
- stream: rtc.VideoStream | rtc.AudioStream
- ready: asyncio.Event

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -62,111 +62,165 @@ def __init__(
...
-                self._link_participant(participant)
-        else:
-            for participant in self._room.remote_participants.values():
-                self._link_participant(participant)
-                if self._participant:
-                    break
+        # link existing participants
+        for participant in self._room.remote_participants.values():
+            self._on_participant_connected(participant)
+        self._room.on("participant_connected", self._on_participant_connected)
```
davidzhao, line 96: we should listen to connected before iterating.. there could be a small chance that a participant is connected after iteration began but before an event handler is registered

livekit-agents/livekit/agents/pipeline/room_io.py
davidzhao: I like the ability to wait.. it would help handle the timing differences for things like PTT. instead of an enum, what about:

```
def link_participant(self, participant: rtc.RemoteParticipant | str, *, wait_for_connection=False)
```

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -184,29 +238,92 @@ async def _capture_text():
...
         self._tasks.clear()
 
-        if self._audio_stream is not None:
-            await self._audio_stream.aclose()
-            self._audio_stream = None
-        if self._video_stream is not None:
-            await self._video_stream.aclose()
-            self._video_stream = None
+    def _start_stream(self, *, audio: bool = True, video: bool = True) -> None:
+        self._close_stream(audio=audio, video=video)
```
davidzhao, line 260: I think we need to clear out both audio and video here. it wouldn't make sense to handle audio from one user, but video from another

livekit-agents/livekit/agents/pipeline/room_io.py
theomonnom: I think we should do it by default and not even expose an option for it. I don't see a case where it would be an issue to have this behaviour

livekit-agents/livekit/agents/pipeline/room_io.py
longcw: Added `StreamHandle` and moved the stream creation, closing, reading into it.

livekit-agents/livekit/agents/pipeline/room_io.py
longcw: Yeah, I agree that we don't need a `wait_for_connection` actually. I added a `set_participant` (open to other names) to just set the target participant identity, then the stream handle listens to the `track_published` event to link to the stream belong to the target participant. We don't really care if the participant is connected or not when setting the target participant.

livekit-agents/livekit/agents/pipeline/room_io.py
longcw: Fixed.

examples/roomio_worker.py
```diff
@@ -61,6 +62,31 @@ def on_playback_finished(ev: PlaybackFinishedEvent) -> None:
     if agent.output.audio is not None:
         agent.output.audio.on("playback_finished", on_playback_finished)
 
+    @ctx.room.local_participant.register_rpc_method("set_participant")
+    async def on_set_participant(data: rtc.RpcInvocationData) -> None:
```
davidzhao, line 66: this is great! should this be an option to enable RPCs? If the user's frontend uses these functions, it seems like they would not want these exposed

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -50,123 +50,338 @@ class RoomOutputOptions:
...
+        """Switch audio and video streams to specified participant"""
+        if participant_identity is None:
+            self.unset_participant()
+            return
 
-        async def _read_stream():
-            async for event in self._audio_stream:
-                yield event.frame
+        # reset future if switching to a different participant
+        if (
```
davidzhao, line 308: when a new participant is set.. should we automatically trigger a interrupt event? for PTT functionality, it seems like you'd want that?

davidzhao (commented):
looking good!

examples/roomio_worker.py
longcw: Yes, this is for the example. users may also want a verification on the caller id, it depends..

davidzhao: oh oops.. i misread..

livekit-agents/livekit/agents/pipeline/room_io.py
davidzhao: makes sense.. I think we should set up native RPCs.. but not for this PR.

davidzhao (approved):

PR #1506, opened 2025-02-16
chore(turn-detector): log from job process instead of inference

davidzhao:
this will preserve any logging context that we are adding


mike-r-mclaughlin (approved):
looks good.  thanks for doing this!

theomonnom (approved):

PR #1508, opened 2025-02-16
feat: connection pooling & reuse to improve latency

davidzhao:
With SpeechStream, we'd like to logically create a new connection instance each time. This tends to introduce a bit of latency if a new connection needs to be made to TTS each time.

This PR introduces a generic connection pool, so the underlying connections can be reused without breaking the abstractions.

### DIFF
--- a/livekit-agents/livekit/agents/tts/tts.py
+++ b/livekit-agents/livekit/agents/tts/tts.py
@@ -5,7 +5,15 @@ import time
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from types import TracebackType
-from typing import AsyncIterable, AsyncIterator, Generic, Literal, TypeVar, Union
+from typing import (
+    AsyncIterable,
+    AsyncIterator,
+    Generic,
+    Literal,
+    Optional,
+    TypeVar,
+    Union,
+)
 
 from livekit import rtc
 
@@ -45,13 +53,19 @@ class TTS(
     Generic[TEvent],
 ):
     def __init__(
-        self, *, capabilities: TTSCapabilities, sample_rate: int, num_channels: int
+        self,
+        *,
+        capabilities: TTSCapabilities,
+        sample_rate: int,
+        num_channels: int,
+        conn_options: Optional[APIConnectOptions] = None,
     ) -> None:
         super().__init__()
         self._capabilities = capabilities
         self._sample_rate = sample_rate
         self._num_channels = num_channels
         self._label = f"{type(self).__module__}.{type(self).__name__}"
+        self._conn_options = conn_options or DEFAULT_API_CONNECT_OPTIONS
 
     @property
     def label(self) -> str:
@@ -74,11 +88,11 @@ class TTS(
         self,
         text: str,
         *,
-        conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
+        conn_options: Optional[APIConnectOptions] = None,
     ) -> ChunkedStream: ...
 
     def stream(
-        self, *, conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS
+        self, *, conn_options: Optional[APIConnectOptions] = None
     ) -> SynthesizeStream:
         raise NotImplementedError(
             "streaming is not supported by this TTS, please use a different TTS or use a StreamAdapter"
@@ -102,11 +116,15 @@ class ChunkedStream(ABC):
     """Used by the non-streamed synthesize API, some providers support chunked http responses"""
 
     def __init__(
-        self, *, tts: TTS, input_text: str, conn_options: APIConnectOptions
+        self,
+        *,
+        tts: TTS,
+        input_text: str,
+        conn_options: Optional[APIConnectOptions] = None,
     ) -> None:
         self._input_text = input_text
         self._tts = tts
-        self._conn_options = conn_options
+        self._conn_options = conn_options or DEFAULT_API_CONNECT_OPTIONS
         self._event_ch = aio.Chan[SynthesizedAudio]()
 
         self._event_aiter, monitor_aiter = aio.itertools.tee(self._event_ch, 2)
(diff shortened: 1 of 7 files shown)


livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -74,11 +88,11 @@ def synthesize(
         self,
         text: str,
         *,
-        conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
+        conn_options: Optional[APIConnectOptions] = None,
```
theomonnom, line 91: DEFAULT_API_CONNECT_OPTIONS is frozen, so it shouldn't be needed to change to Optional

livekit-agents/livekit/agents/utils/pool.py
```diff
@@ -0,0 +1,147 @@
...
+                or now - self._connections[conn] <= self._max_session_duration
+            ):
+                return conn
+            # Connection expired; mark it for resetting.
+            self.reset(conn)
+
+        # Create new connection
+        logger.info("creating new connection")
+        conn = await self.connect()
+        logger.info("created new connection")
```
theomonnom, line 83: Remove logs

livekit-agents/livekit/agents/utils/pool.py
```diff
@@ -0,0 +1,147 @@
...
+            raise NotImplementedError("Must provide connect_cb or implement connect()")
+        return await self._connect_cb()
+
+    async def _drain_to_close(self) -> None:
+        """Drain and close all the connections queued for closing."""
+        for conn in list(self._to_close):
+            await self._maybe_close_connection(conn)
+        self._to_close.clear()
+
+    async def get(self) -> T:
```
theomonnom, line 58: Instead of get/put, what about using a ContextManager?

livekit-agents/livekit/agents/utils/pool.py
theomonnom, line 1: rename the file to connection_pool

livekit-agents/livekit/agents/utils/pool.py
```diff
@@ -0,0 +1,147 @@
...
+        """Get an available connection or create a new one if needed.
+
+        Returns:
+            An active connection object
+        """
+        # First, close out any connections that have previously been marked.
+        await self._drain_to_close()
```
theomonnom, line 65: should this be handled inside get()?

livekit-agents/livekit/agents/utils/pool.py
```diff
@@ -0,0 +1,147 @@
+import logging
+import time
+from typing import Awaitable, Callable, Generic, Optional, Set, TypeVar
+
+T = TypeVar("T")
+
+logger = logging.getLogger("livekit.agents.pool")
+
+
+class ConnectionPool(Generic[T]):
```
theomonnom, line 10: I think lot of the exposed API could be made private? Like connect()

livekit-agents/livekit/agents/utils/pool.py
```diff
@@ -0,0 +1,147 @@
...
+    async def _maybe_close_connection(self, conn: T) -> None:
+        """Close a connection if close_cb is provided.
+
+        Args:
+            conn: The connection to close
+        """
+        if self._close_cb is not None:
+            await self._close_cb(conn)
+
+    def reset(self, conn: T) -> None:
```
theomonnom, line 107:
```suggestion
    def remove(self, conn: T) -> None:
```

livekit-agents/livekit/agents/tts/tts.py
davidzhao: yes, but making it required with default forces the pattern of repeating this everywhere.. Plugins should not have to worry about these defaults

it's much cleaner to have a single place that picks the defaults (base class of ChunkedStream).

livekit-agents/livekit/agents/utils/pool.py
davidzhao: I didn't want to make remove/reset/etc worry about being async. it will slow down get a bit.. but if any previous connections have been removed, the next time we interact with get would clean it up

livekit-agents/livekit/agents/utils/pool.py
davidzhao: renamed connect. what else do you see that should be private?

davidzhao:
@theomonnom PTAL

theomonnom (approved):

livekit-agents/livekit/agents/utils/connection_pool.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Args:
+            conn: The connection to reset
+        """
+        self._available.discard(conn)
+        if conn in self._connections:
+            self._to_close.add(conn)
+            self._connections.pop(conn, None)
+
+    def maybe_remove(self, conn: T) -> None:
```
theomonnom, line 139: I'm not a fan of this "maybe_remove". Hard to grasp what it is doing initially. Would be great if you don't have to think about it

livekit-agents/livekit/agents/utils/connection_pool.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        if conn in self._connections:
+            now = time.time()
+            if (
+                self._max_session_duration is None
+                or now - self._connections[conn] > self._max_session_duration
+            ):
+                self.remove(conn)
+
+    def remove_all(self) -> None:
```
theomonnom, line 153: maybe?
```suggestion
    def invalidate(self) -> None:
```

theomonnom: I see only Google uses it

davidzhao: I like it

livekit-agents/livekit/agents/utils/connection_pool.py
davidzhao: right. it's basically opportunistically handling removing the connection at the right moment. def a bit harder to understand

davidzhao: perhaps we don't need this.. and relying on the timeout is sufficient..

theomonnom: yes, relying on timeout should be sufficient IMO. I see only Google uses it once again, and double checking the code, the Google code could even just use the `remove` function since it is already checking for expiration:
```python
                        if (
                            time.time() - self._session_connected_at
                            > _max_session_duration
                        ):
                            logger.debug(
                                "Google STT maximum connection time reached. Reconnecting..."
                            )
                            self._pool.maybe_remove(client)
```

PR #1521, opened 2025-02-19
text input from datastream for v1.0

longcw:


livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -399,6 +411,28 @@ async def _capture_text():
...
+            if not self._agent:
+                return
+
+            text = await reader.read_all()
+            logger.debug(
+                "received text input",
+                extra={"text": text, "participant": self._participant_identity},
+            )
+            self._agent.interrupt()
+            self._agent.generate_reply(user_input=text)
```
longcw, line 430: we are calling `agent` inside the RoomInput, is that okay? should the `agent.input` has a `text` input, or we make `agent` required for `RoomInput`?

theomonnom (approved):

livekit-agents/livekit/agents/pipeline/room_io.py
theomonnom: Let's make agent required

longcw: Ok, I'll do that when merging the RoomInput and Output.

livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -48,6 +50,7 @@ class RoomOutputOptions:
 DEFAULT_ROOM_INPUT_OPTIONS = RoomInputOptions()
 DEFAULT_ROOM_OUTPUT_OPTIONS = RoomOutputOptions()
 LK_PUBLISH_FOR_ATTR = "lk.publish_for"
+LK_TEXT_INPUT_TOPIC = "lk.room_text_input"
```
lukasIO, line 53: we should be clear about what topics we want to support. If the goal is to have this work ootb with chat components, then choosing a custom topic here might not be the best choice

longcw:
Do we wan to keep the ChatManager in python sdk, related PR https://github.com/livekit/python-sdks/pull/360?

livekit-agents/livekit/agents/pipeline/room_io.py
longcw: I think the question is are we going to keep the chat components in python/js sdk, and having both the original and the datastream or only the datastream for it? If so I can adjust here accordingly. cc @davidzhao

davidzhao: the chat components are for client-side. but Python/Node agents should agree on the same topic so that it works with the client-side components

@lukasIO what do you recommend we should use? is client-side component listening to both the transcription topic and chat?

davidzhao:
> Do we wan to keep the ChatManager in python sdk, related PR [livekit/python-sdks#360](https://github.com/livekit/python-sdks/pull/360)?

IMO we should deprecate and remove ChatManager.. agents should not use it any longer

livekit-agents/livekit/agents/pipeline/room_io.py
lukasIO: currently chat components only listen to the chat topic and also send their messages only on the chat topic

davidzhao: how would this work with how we are sending transcriptions? do you suggest also sending transcriptions to chat topic?

lukasIO: that was my understanding, yes. But maybe I misunderstood or am forgetting something. 
Why wouldn't you want it on the chat topic?

PR #1523, opened 2025-02-19
use connection pool for deepgram tts

jayeshp19:
Reuse websocket connection using connection pooling for deepgram and elevanlabs

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
@@ -5,12 +5,12 @@ import json
 import os
 import weakref
 from dataclasses import dataclass
+from typing import Optional
 from urllib.parse import urlencode
 
 import aiohttp
 from livekit import rtc
 from livekit.agents import (
-    DEFAULT_API_CONNECT_OPTIONS,
     APIConnectionError,
     APIConnectOptions,
     APIStatusError,
@@ -114,7 +114,7 @@ class TTS(tts.TTS):
         self,
         text: str,
         *,
-        conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
+        conn_options: Optional[APIConnectOptions] = None,
     ) -> "ChunkedStream":
         return ChunkedStream(
             tts=self,
@@ -127,7 +127,7 @@ class TTS(tts.TTS):
         )
 
     def stream(
-        self, *, conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS
+        self, *, conn_options: Optional[APIConnectOptions] = None
     ) -> "SynthesizeStream":
         stream = SynthesizeStream(
             tts=self,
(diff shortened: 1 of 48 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
```diff
@@ -391,21 +394,6 @@ async def _connection_timeout():
                         *tasks, wait_reconnect_task, connection_timeout_task
                     )
 
-            except asyncio.TimeoutError as e:
```
davidzhao, line 394: we still need to catch & convert these errors.

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -366,161 +382,133 @@ async def _run_ws(
         word_stream: tokenize.WordStream,
         max_retry: int = 3,
     ) -> None:
-        ws_conn: aiohttp.ClientWebSocketResponse | None = None
```
davidzhao, line 369: I think we should update this plugin to also use the retry mechanism in TTS superclass. otherwise removing the retry code here would break failure handling

jayeshp19: This plugin already uses retry mechanism from superclass, this was extra for establishing ws connection

davidzhao: got it. I misinterpreted the `_run` local function as the instance method. my bad

davidzhao (approved):

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
```diff
@@ -314,12 +327,10 @@ def _send_last_frame(*, segment_id: str, is_final: bool) -> None:
                     aiohttp.WSMsgType.CLOSED,
                     aiohttp.WSMsgType.CLOSING,
                 ):
-                    if not closing_ws:
-                        raise APIStatusError(
-                            "Deepgram websocket connection closed unexpectedly",
-                            request_id=request_id,
-                        )
-                    return
```
theomonnom, line 322: I don't think this is right, it raise this error anytime the connection is closed. Which could be expected

theomonnom: Nvm this is deleted

jayeshp19: We've removed `closing_ws` flag, since we're keeping ws connection open, it will raise status error everytime

theomonnom: Mmh why we would want the error after gracefully closing the connection?

davidzhao: while the connection is checked out of the pool, there shouldn't be anything that would attempt to close it.

though we do need to close the pool eventually.. currently it'll be closed only when the process exits.. I think we need a `tts.aclose` that correctly cleans up after it

PR #1525, opened 2025-02-19
acoustic echo cancellation

theomonnom:
Using libwebrtc AEC3
Everything happens in `livekit-plugins/livekit-plugins-aec/copy_aec3.py` (mostly) 

The webrtc files take a lot of LOC, but having them inside the repo will make builds easier (instead of having to clone the libwebrtc repo) 

We're copying without changes the AEC3 module, this way it's easy to keep it in sync with upstream.
This is currently using the latest commit in the libwebrtc google repo |

Everything inside `src/` is the webrtc code


theomonnom:
skipped SIMD optimizations, as it will take some time to improve the builds. Also only the `console` command is going to use it

PR #1528, opened 2025-02-20
add agent state for v1.0

longcw:


livekit-agents/livekit/agents/pipeline/generation.py
```diff
@@ -153,21 +153,29 @@ class _TextOutput:
 
 def perform_text_forwarding(
     *, text_output: io.TextSink | None, llm_output: AsyncIterable[str]
-) -> tuple[asyncio.Task, _TextOutput]:
+) -> tuple[asyncio.Task, _TextOutput, asyncio.Future]:
```
theomonnom, line 156: Do you think there is a way where we don't add this future? 
I think the right strategy would be to add agent_started_speaking and agent_stopped_speaking event, and listen to them accordingly to update the agent state, wdyt?

theomonnom: Otherwise this lgtm

longcw: I agree with the `agent_started_speaking` and `agent_stopped_speaking` events, but to emit the start speaking after the first frame captured, we still need this future? or emit them from the audio sink? I prefer to emit it from the task activity if there is no other options. wdyt?

theomonnom (approved):
This lg, I'll add agent_speech_started and agent_speech_stopped soon. (will do at the same time as metrics)

PR #1529, opened 2025-02-20
fix: perplexity requires initial user message in chat context

jayeshp19:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -696,6 +700,15 @@ async def _run(self) -> None:
             opts = _strip_nones(opts)
 
             messages = _build_oai_context(self._chat_ctx, id(self))
+            if (
+                self._llm._capabilities.requires_initial_user_message
+                and len(messages) > 0
+            ):
+                if messages[0]["role"] == "system" and messages[1]["role"] != "user":
+                    messages.insert(1, {"role": "user", "content": ""})
+                elif messages[0]["role"] != "user":
```
davidzhao, line 709: I'm not sure if this is sufficient.. I think Perplexity (and others) require interleaved user / assistant turns.. 

IMO we need to scan through the messages and enforce interleaved messages. maybe also renaming the capability: `single_message_per_turn`?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -491,6 +491,10 @@ def with_perplexity(
             parallel_tool_calls=parallel_tool_calls,
             tool_choice=tool_choice,
         )
+        perplexity_llm._capabilities = llm.LLMCapabilities(
```
davidzhao, line 494: do not override defaults, instead just do:

```
perplexity_llm._capabilities.requires_initial_user_message = False
```

theomonnom:
I'm not sure about the strategy to handle this. Should this be a new adapter (like we do for StreamAdapter).

Anthropic also has this issue.

PR #1532, opened 2025-02-20
examples for 1.0

tinalenguyen:
to be reviewed:
- dentist scheduler: a multi-agent example offering different functionalities integrated via Cal.com and Supabase APIs
- conversation persistor (realtime and pipeline): an updated version for 1.0 events
- conversation recorder: example of grabbing input/output audioframes for a wav recording via stt and tts nodes


examples/dentist_scheduler.py
```diff
@@ -0,0 +1,135 @@
...
+            instructions="You are an assistant scheduling dentist appointments for users. Always speak in English.",
+            llm=openai.realtime.RealtimeModel(),
+        )
+
+    async def on_enter(self) -> None:
+        speech_handle = await self.agent.generate_reply(
+            instructions="Welcome the user to the LiveKit Dental Office and ask how you could assist."
+        )
+
+    @ai_function(description="This functions allows the user to leave a message for the dentist.")
```
theomonnom, line 34: We moved the description inside the docstrings.

examples/dentist-scheduler/dentist_scheduler.py
```diff
@@ -0,0 +1,383 @@
...
+        await self.agent.generate_reply(
+            instructions=f"""Welcome the user to the LiveKit Dental Office and ask how you can assist. The user's name is {self._userinfo.name}. 
+            If the user wants to manage an appointment or leave a message and their name is not given, ask for it before proceeding."""
+        )
+
+    @ai_function()
+    async def hours_inquiry(self):
+        """Answers user inquiries about the LiveKit dental office's hours of operation."""
+        await self.agent.current_speech.wait_for_playout()
+        await self.agent.generate_reply(
```
theomonnom, line 55: Hey, generate_reply may not necessarily be needed, you could just return the string here

tinalenguyen: when returning a string, is it included in generate_reply() or used directly in say()?

theomonnom:
for ruff, you can automatically fix the imports using `ruff check --fix`

examples/other/conversation-history/conversation_recorder.py
```diff
@@ -0,0 +1,70 @@
...
+            self._file.setframerate(24000)
+            self._file.setsampwidth(2)
+
+    async def stt_node(
+        self, audio: AsyncIterable[rtc.AudioFrame], model_settings: ModelSettings
+    ) -> AsyncIterable[stt.SpeechEvent]:
+        async def record_audio():
+            async for frame in audio:
+                wavframe = frame.data
+                self._file.writeframes(wavframe)
```
tinalenguyen, line 33: the input audio sounds high-pitched, could it be that its frame rate is different?

examples/other/conversation-history/conversation_persistor.py
```diff
@@ -0,0 +1,210 @@
...
+        session: AgentSession | None,
+        log: str | None,
+        transcriptions_only: bool = False,
+    ):
+        """
+        Initializes a ConversationPersistor instance which records the events and
+        transcriptions of a VoiceAgent.
+
+        Args:
+            model (AgentSession): an instance of an AgentSession
```
rinarakaki, line 53: `model` → `session`

examples/other/conversation-history/conversation_persistor.py
```diff
@@ -0,0 +1,210 @@
...
+        # Exits
+        self._log_q.put_nowait(None)
+        await self._main_task
+
+    def start(self) -> None:
+        # Listens for emitted events
+        self._main_task = asyncio.create_task(self._main_atask())
+
+        @self.session.on("user_started_speaking")
+        def _user_started_speaking(ev: AgentEvent):
```
rinarakaki, line 130: It might be better to use specific non-union event types, otherwise typechecker (e.g. pyright) emits errors:

- `error: Cannot access attribute "is_final" for class "UserStartedSpeakingEvent"`
- `error: Cannot access attribute "transcript" for class "UserStartedSpeakingEvent"`
- etc.

examples/other/conversation-history/conversation_persistor.py
```diff
@@ -0,0 +1,210 @@
...
+                transcription = TranscriptionLog(role="assistant", transcription=ev.message.content)
+                self._log_q.put_nowait(transcription)
+            event = EventLog(eventname=ev.type)
+            self._log_q.put_nowait(event)
+
+        @self.session.on("agent_state_changed")
+        def _agent_state_changed(ev: AgentEvent):
+            """The state is either "initializing", "listening", or "speaking" """
+            name = "agent_state_changed: " + ev.state.value
+            event = EventLog(name)
```
rinarakaki, line 159: `EventLog(eventname=ev.type)`

examples/other/conversation-history/conversation_persistor.py
```diff
@@ -0,0 +1,210 @@
...
+            if ev.is_final:
+                event = EventLog(eventname=ev.type)
+                self._log_q.put_nowait(event)
+                transcription = TranscriptionLog(role="user", transcription=ev.transcript)
+                self._log_q.put_nowait(transcription)
+
+        @self.session.on("conversation_item_added")
+        def _conversation_item_added(ev: AgentEvent):
+            if ev.message.role == "assistant":
+                transcription = TranscriptionLog(role="assistant", transcription=ev.message.content)
```
rinarakaki, line 150: `error: Argument of type "list[ChatContent]" cannot be assigned to parameter "transcription" of type "str" in function "__init__"`

examples/other/conversation-history/conversation_persistor.py
tinalenguyen: Thank you for your feedback! I'm working on it right now, I also think I'll make separate versions (one for OpenAI's Realtime model and another for regular pipeline)

examples/other/conversation-history/conversation_recorder.py
tinalenguyen: seems like the data needs to be resampled depending on the tts

Bilal-io:
pydub depends on audioop which was deprecated in Python 3.11 and removed in Python 3.13
There is a [PR](https://github.com/jiaaro/pydub/pull/816) but not merged yet.

tinalenguyen:
@Bilal-io thanks for the heads up, i decided to opt for LK's audio resampler which uses SoX

tinalenguyen:
@Bilal-io I can't seem to replicate that problem, are you using the same pipeline setup as the `conversation_recorder.py` example?

tinalenguyen:
@Bilal-io Thank you for the gist, I was able to replicate the issue and fix it! Let me know if it works now :)

Bilal-io:
Hey @tinalenguyen thank you for the quick fix.
I am seeing two different issues:
1- First call works fine, second call causes an error `...return stt(self, record_audio(), model_settings)...` But I fixed this by using the same pattern as the `tts_recorder` instead of returning the `stt(self, record_audio(), model_settings)` directly, I did this:
```
async for result in stt(self, record_audio(), model_settings):
            yield result
```

2- The audio of the agent sounds great when speaking but comes out choppy in the saved file. This is the case even without the change mentioned above. I've attached an audio sample (converted to mp4 to be able to attach here). Not sure if this is related to Livekit itself or your implementation.

I appreciate your input

https://github.com/user-attachments/assets/ba35bbd2-b63b-47ea-814a-d414290303e3

tinalenguyen:
@Bilal-io Good catch, that approach makes more sense! 

As for the agent audio, I suspect it's from STT audio cutting in during the agent's speech and not mixing well. I've resolved it by changing the quality of the resampler to very high:

`self._audio_resampler = AudioResampler(input_rate=frame.sample_rate, output_rate=FRAMERATE, quality="very_high")`

If the audio still isn't consistent, let me know and I'll look into crossfading/transitioning the audio streams. Thank you again for trying out my work!!

tinalenguyen:
@Bilal-io Thank you for the feedback, I ended up rewriting most of it and I think it works way better now. Sorry about the bugs/delay, this must be a sign for me to stop coding at 4 AM..

Let me know what you think, and thanks again!!

Issue #1533, opened 2025-02-20
Deepgram's keyterms should be keyterm, not the plural


wdhwg001:
livekit-plugins-deepgram = "0.6.19"

See: https://developers.deepgram.com/docs/keyterm

It's in `SpeechStream._connect_ws`:

```python3
    async def _connect_ws(self) -> aiohttp.ClientWebSocketResponse:
        live_config: dict[str, Any] = {
            "model": self._opts.model,
            "punctuate": self._opts.punctuate,
            "smart_format": self._opts.smart_format,
            "no_delay": self._opts.no_delay,
            "interim_results": self._opts.interim_results,
            "encoding": "linear16",
            "vad_events": True,
            "sample_rate": self._opts.sample_rate,
            "channels": self._opts.num_channels,
            "endpointing": False
            if self._opts.endpointing_ms == 0
            else self._opts.endpointing_ms,
            "filler_words": self._opts.filler_words,
            "profanity_filter": self._opts.profanity_filter,
        }
        if self._opts.keywords:
            live_config["keywords"] = self._opts.keywords
        if self._opts.keyterms:
            live_config["keyterms"] = self._opts.keyterms  # <-- should be `keyterm`, not `keyterms`

        if self._opts.language:
            live_config["language"] = self._opts.language
```

PR #1535, opened 2025-02-21
fix(deepgram): fix STT keyterm parameter

wdhwg001:
closes https://github.com/livekit/agents/issues/1533


davidzhao (approved):
lg! thanks for catching this

PR #1536, opened 2025-02-21
Proposal: Add simple video input support for gemini live

bcherry:
There is active developer interest in this feature and its pretty easy to at least provide a hook on the RealtimeSession to push video frames through and we can give a little guidance on how to use it. The interface will of course change with 1.0 and we can perfect it then.

I've tested this and it works pretty well.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
@@ -9,6 +9,7 @@ from typing import AsyncIterable, Literal
 from livekit import rtc
 from livekit.agents import llm, utils
 from livekit.agents.llm.function_context import _create_ai_function_info
+from livekit.agents.utils import images
 
 from google import genai
 from google.genai._api_client import HttpOptions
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
lg

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -325,14 +326,48 @@ def fnc_ctx(self) -> llm.FunctionContext | None:
     def fnc_ctx(self, value: llm.FunctionContext | None) -> None:
         self._fnc_ctx = value
 
-    def _push_audio(self, frame: rtc.AudioFrame) -> None:
-        if self._opts.enable_user_audio_transcription:
-            self._transcriber._push_audio(frame)
+    def push_media_chunk(self, data: bytes, mime_type: str) -> None:
```
theomonnom, line 329:
```suggestion
    def _push_media_chunk(self, data: bytes, mime_type: str) -> None:
```

bcherry: @theomonnom the reason to keep this public is to give developers an ability to adjust the encoding method used (e.g. use PNG instead of JPEG, or a different size).

bcherry: just going to make it private for now and update the other comment so we can move this along

davidzhao:
test is my fault.. will be fixed in my PR

PR #1549, opened 2025-02-25
fix transcription sync for datastream sink

longcw:


livekit-agents/livekit/agents/pipeline/room_io.py
```diff
@@ -683,8 +687,11 @@ def set_participant(
...
-        if identity != self._participant_identity and self._text_writer:
-            # close the previous stream if the participant has changed
+
+        if identity == self._participant_identity:
+            return
+
+        if self._text_writer:
             current_writer = self._text_writer
             self._text_writer = None
             task = asyncio.create_task(current_writer.aclose())
```
lukasIO, line 697: wondering if this still might create a race here due to `set_participant` not being async and not waiting for aclose to complete 👀

livekit-agents/livekit/agents/pipeline/transcription/synchronizer.py
```diff
@@ -334,42 +333,34 @@ def text_sink(self) -> "_TextSink":
         """Get the text sink wrapper"""
         return self._text_sink
 
-    def _segment_playout_started(self) -> None:
-        self._synchronizer.segment_playout_started()
+    async def _forward_event(self) -> None:
+        while not self._closed:
+            async for segment in self._synchronizer:
+                if not self._base_text_sink:
```
lukasIO, line 339: under which circumstances would base_text_sink not be defined here? 

would we want to drop the segments in that case or buffer them until the text sink is available?

livekit-agents/livekit/agents/pipeline/transcription/synchronizer.py
```diff
@@ -334,42 +333,34 @@ def text_sink(self) -> "_TextSink":
...
+                    continue
+                if self._current_segment_id != segment.id:
+                    self._base_text_sink.flush()
```
lukasIO, line 342: flush is currently not async, would it be easier to handle some of these edge cases if it was?

livekit-agents/livekit/agents/pipeline/transcription/synchronizer.py
```diff
@@ -334,42 +333,34 @@ def text_sink(self) -> "_TextSink":
...
+        current_synchronizer = self._synchronizer
+        self._synchronizer = _TextAudioSynchronizer(options=self._sync_options)
+        # close the old synchronizer
+        task = asyncio.create_task(current_synchronizer.aclose())
         self._tasks.add(task)
         task.add_done_callback(self._tasks.discard)
 
     async def aclose(self) -> None:
         """Close the forwarder and cleanup resources"""
+        self._closed = True
```
lukasIO, line 360: should we flush the text sink when the synchronizer gets closed?

livekit-agents/livekit/agents/pipeline/transcription/synchronizer.py
longcw: Yes, actually the mi case like `id1, id1, id2, id1, id2, ...` shouldn't happen now, the new segments are always send after the old one got closed.

livekit-agents/livekit/agents/pipeline/room_io.py
longcw: replaced this with `flush` and added a lock to avoid writer closed in flush when it's still writing the data

livekit-agents/livekit/agents/pipeline/transcription/synchronizer.py
lukasIO: got it, just a general thought, would it be easier to handle (maybe also outside of synchronizer) if `flush` was async?

livekit-agents/livekit/agents/pipeline/room_io.py
lukasIO: sounds good, what scope should the lock have?

i.e. should closing it also be within the lock scope?

livekit-agents/livekit/agents/pipeline/transcription/synchronizer.py
lukasIO: thanks for adding it!

livekit-agents/livekit/agents/pipeline/transcription/synchronizer.py
longcw: I asked @theomonnom before but I think for consistence and people using the flush from sync function we still want it to be sync.

PR #1551, opened 2025-02-25
noise cancellation plugin support in VPA & MMA

bcherry:
This is tested and works great. The interface to adopt in 1.0 may be different but this seems like the straightforward approach for 0.12.


livekit-agents/livekit/agents/pipeline/human_input.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         stt: speech_to_text.STT,
         participant: rtc.RemoteParticipant,
         transcription: bool,
+        noise_cancellation: tuple[str, dict[str, Any]] | None = None,
```
typester, line 31: `rtc.NoiseCancellationOptions` is exported to represent this type. So once [my p-r](https://github.com/livekit/python-sdks/pull/350) was merged, you might want to use it instead of this

theomonnom (approved):

typester (approved):

PR #1552, opened 2025-02-26
change default model for google stt and add aws llm test case

jayeshp19:
- added test case for consecutive messages in chat context

### DIFF
--- a/livekit-plugins/install_plugins_editable.sh
+++ b/livekit-plugins/install_plugins_editable.sh
@@ -5,7 +5,8 @@ if [[ -z "$VIRTUAL_ENV" ]]; then
     echo "You are not in a virtual environment."
     exit 1
 fi
-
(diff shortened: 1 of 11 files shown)


livekit-plugins/install_plugins_editable.sh
```diff
@@ -5,7 +5,8 @@ if [[ -z "$VIRTUAL_ENV" ]]; then
     echo "You are not in a virtual environment."
     exit 1
 fi
-
+pip install -e ./livekit-plugins-anthropic --config-settings editable_mode=strict
```
davidzhao, line 8: also add to ./install_local ? mypy checks would use that version

davidzhao (approved):

livekit-plugins/install_plugins_editable.sh
jayeshp19: it's already there

PR #1554, opened 2025-02-26
toggle agent input/output dynamically for agent 1.0 

longcw:
+ Added the API for `AgentInput` and `AgentOutput` to toggle the audio/video/text input/output
+ If disabled, input audio/video will be consumed in a background task
+ The transcription sync will be  disabled if any of the audio/text output disabled.


livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -179,27 +228,91 @@ def audio(self) -> AudioStream | None:
...
+            # reset the inactive stream
+            self._inactive_audio_stream = None
+            if self._consume_audio_task:
+                self._consume_audio_task.cancel()
+                self._consume_audio_task = None
         self._audio_changed()
 
+    async def _consume_stream(self, stream: AudioStream | VideoStream) -> None:
+        async for frame in stream:
+            pass
```
theomonnom, line 241: Let's add a clear_queue method to https://github.com/livekit/python-sdks/blob/d28f54c14675f26b3bdea889c79ed49d44b6579c/livekit-rtc/livekit/rtc/audio_stream.py#L42 instead?

livekit-agents/livekit/agents/pipeline/io.py
```diff
@@ -179,27 +228,91 @@ def audio(self) -> AudioStream | None:
...
+            self._consume_audio_task.cancel()
+        if self._consume_video_task:
+            self._consume_video_task.cancel()
+
+        self._consume_audio_task = None
+        self._consume_video_task = None
+
 
-class AgentOutput:
+class AgentOutput(rtc.EventEmitter[Literal["video_changed", "audio_changed", "text_changed"]]):
```
theomonnom, line 253: Do we really need to expose those events? If we're not sure about it I would say to remove them for now

livekit-agents/livekit/agents/pipeline/io.py
longcw: But the input stream is not always AudioStream or VideoStream. Actually now from RoomIO it's a `aio.Chan` to support switching participant.

livekit-agents/livekit/agents/pipeline/io.py
longcw: The RoomIO need them to enable/disable the transcription sync. Otherwise we need to pass the `AgentOutput` to the transcription sync.

livekit-agents/livekit/agents/pipeline/io.py
theomonnom: Ah ok, but I think we wanted to only support this for RoomIO. I think it's kind of unexpected for the users to have their video/audio streams being pulled. But everything is dropped

longcw: I think toggle the input/output can be an API of the agent, but yeah dropping the frames is only for realtime stream.

longcw: maybe user can just set `agent.input.audio = None` if the don't want the stream be consumed? We should always assume the input stream is a realtime stream?

livekit-agents/livekit/agents/pipeline/io.py
theomonnom: Ah makes sense!

theomonnom: I'm wondering how this is working actually, can't you just decide to create the TranscriptionSync based on whether the AgentOutput.audio is None?

longcw: Right now it's the TranscriptionSync listen to the event and enable/disable the sync inside it. The check is not only for creation at start, the `agent.output.audio` may change when we enable/disable the audio, the sync should be disabled if the audio or text output is disabled in agent.

PR #1556, opened 2025-02-26
handle process initialization failure 

theomonnom:
catch user initialization cb failures & allow concurrent proc initializations

### DIFF
--- a/livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
+++ b/livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
@@ -70,7 +70,11 @@ def proc_main(args: ProcStartArgs) -> None:
 
     pid = current_process().pid
     logger.info("initializing job process", extra={"pid": pid})
-    client.initialize()
+    try:
+        client.initialize()
+    except Exception:
+        return  # initialization failed, exit
+
     logger.info("job process initialized", extra={"pid": pid})
 
     client.run()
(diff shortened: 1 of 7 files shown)


davidzhao (approved):

livekit-agents/livekit/agents/ipc/proc_pool.py
```diff
@@ -19,7 +19,7 @@
     "process_job_launched",
 ]
 
-MAX_CONCURRENT_INITIALIZATIONS = 1
+MAX_CONCURRENT_INITIALIZATIONS = 5
```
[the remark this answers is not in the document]
theomonnom: The caused effect was spiky CPU pattern when new processes load. I don't see how we can avoid that, I think it is OK as long as it doesn't overload.

We can scale with the number of CPU cores, tho prewarm functions aren't necessarily CPU bound. So it is OK to have more, perhaps it could be a new option to the worker.

livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
```diff
@@ -70,7 +70,11 @@ def proc_main(args: ProcStartArgs) -> None:
 
     pid = current_process().pid
     logger.info("initializing job process", extra={"pid": pid})
-    client.initialize()
+    try:
+        client.initialize()
+    except Exception:
+        return  # initialization failed, exit
```
martin-purplefish, line 76: Log something?

theomonnom: The log is on the worker side

PR #1563, opened 2025-02-27
retain the llm format for datastream transcription sink

longcw:


longcw:
replaced by https://github.com/livekit/agents/pull/1595

PR #1573, opened 2025-02-28
update testcase: consecutive messages

jayeshp19:


davidzhao:
what is the goal of this change? is the idea that the last message before `.chat` should be a user message?

jayeshp19:
> what is the goal of this change? is the idea that the last message before `.chat` should be a user message?

Yes

davidzhao (approved):

PR #1574, opened 2025-02-28
Add elevenlabs speed parameter

MatthiasGruba:
Elevenlabs added a speed parameter in the last release: https://elevenlabs.io/docs/changelog/2025/2/25#text-to-speech

### DIFF
--- a/.changeset/mighty-pandas-dress.md
+++ b/.changeset/mighty-pandas-dress.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-elevenlabs": patch
+---
+
+Add elevenlabs speed parameter
--- a/.changeset/mighty-pandas-dress.md
+++ b/.changeset/mighty-pandas-dress.md
@@ -1,5 +1,0 @@
----
-"livekit-plugins-elevenlabs": patch
----
-
-Add elevenlabs speed parameter
--- a/.changeset/tender-beds-chew.md
+++ b/.changeset/tender-beds-chew.md
@@ -1,0 +1,23 @@
+---
+"livekit-plugins-elevenlabs": patch
+---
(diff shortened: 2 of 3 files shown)


jayeshp19:
Thanks for contribution @MatthiasGruba , can you run `ruff format .` and also add changeset?

davidzhao:
we should also expose this in `update_settings` [here](https://github.com/livekit/agents/pull/1574/files#diff-3000320415d13830aaea00279bd5c5e31b9a6e8ced1cb51c079bbb131c99ffbcR206)

MatthiasGruba:
> here

I'm not sure if I understood that correctly - the speed is part of the voice?

MatthiasGruba:
> we should also expose this in `update_settings` [here](https://github.com/livekit/agents/pull/1574/files#diff-3000320415d13830aaea00279bd5c5e31b9a6e8ced1cb51c079bbb131c99ffbcR206)

I'm not sure if I understood this correctly - the speed is part of the voice which can be updated with this method?

MatthiasGruba:
> Thanks for contribution @MatthiasGruba , can you run `ruff format .` and also add changeset?

Done and added! First contribution here so let me know if that was correct!

davidzhao:
> > we should also expose this in `update_settings` [here](https://github.com/livekit/agents/pull/1574/files#diff-3000320415d13830aaea00279bd5c5e31b9a6e8ced1cb51c079bbb131c99ffbcR206)
> 
> I'm not sure if I understood this correctly - the speed is part of the voice which can be updated with this method?

my apologies. it's already part of the Voice parameter. ignore me :)

davidzhao (approved):

Issue #1578, opened 2025-03-01
Dockerfile-example Fails to Build voice-pipeline-agent Due to Missing g++


shwetd19:
### Description
When building a Docker image using the provided `examples/Dockerfile-example` with the `requirements.txt` from `examples/voice-pipeline-agent/`, the build fails due to a missing C++ compiler (`g++`). The error occurs while compiling the `annoy` package, which is a transitive dependency of one of the required packages (likely `livekit-plugins-silero` or similar).

### The build fails with the following error:
```
gcc: fatal error: cannot execute ‘cc1plus’: execvp: No such file or directory
compilation terminated.
error: command '/usr/bin/gcc' failed with exit code 1
ERROR: Failed building wheel for annoy
```

### Root Cause
- The `annoy` package requires C++ compilation (evident from `.cc` files in its build output).
- The `Dockerfile-example` installs `gcc` but not `g++`, which is needed to compile C++ code.

### Suggested Fix
Add `g++` to the `apt-get install` command in `examples/Dockerfile-example`:
```dockerfile
RUN apt-get update && \
    apt-get install -y \
    gcc \
    g++ \
    python3-dev \
    && rm -rf /var/lib/apt/lists/*
```
### Additional Context
requirements.txt used: https://github.com/livekit/agents/blob/main/examples/voice-pipeline-agent/requirements.txt
Dockerfile: https://github.com/livekit/agents/blob/main/examples/Dockerfile-example
Tested with python:3.11.6-slim base image.

This seems like a small but impactful fix to make the example more robust. Happy to submit a PR if this aligns with the project’s goals!

davidzhao:
fixed in #1579

PR #1579, opened 2025-03-01
Added g++ to Dockerfile-example to Support Building annoy Dependency

shwetd19:
### Summary
This PR updates `examples/Dockerfile-example` to include `g++` as a build dependency, fixing a build failure when using the `requirements.txt` from `examples/voice-pipeline-agent/`. The `annoy` package, a transitive dependency of one of the listed packages (likely `livekit-plugins-silero`), requires a C++ compiler, but the original Dockerfile only included `gcc`.

### Problem
When building the Docker image with:
- Dockerfile: `examples/Dockerfile-example`
- `requirements.txt`: `examples/voice-pipeline-agent/requirements.txt`

The process fails with:
```
gcc: fatal error: cannot execute ‘cc1plus’: execvp: No such file or directory
compilation terminated.
error: command '/usr/bin/gcc' failed with exit code 1
ERROR: Failed building wheel for annoy
```

This happens because `annoy` contains C++ code (e.g., `annoymodule.cc`), and `gcc` alone cannot compile it without `g++`.

### Testing
- Built the image locally using the updated Dockerfile and `requirements.txt` from `examples/voice-pipeline-agent/`.
- Confirmed the build completes successfully without errors.
- Tested with base image `python:3.11.6-slim`.

### Impact
- Fixes the build process for users following the `voice-pipeline-agent` example.
- Adds a minimal dependency (`g++`) with a small size increase, preserving the example’s “minimal” intent while ensuring functionality.

### Related Issue
#1578 

This change ensures the example Dockerfile works out-of-the-box with the provided `voice-pipeline-agent` example. Let me know if any adjustments are needed!

### DIFF
--- a/examples/Dockerfile-example
+++ b/examples/Dockerfile-example
@@ -21,11 +21,11 @@ RUN adduser \
     --uid "${UID}" \
     appuser
 
-
-# Install gcc and other build dependencies.
+# Install gcc, g++ and other build dependencies.
 RUN apt-get update && \
     apt-get install -y \
     gcc \
+    g++ \
     python3-dev \
     && rm -rf /var/lib/apt/lists/*
 


davidzhao (approved):
good fix

PR #1582, opened 2025-03-02
fix: increase elevenlabs websocket timeout

jayeshp19:


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -102,6 +103,7 @@ class _TTSOptions:
     word_tokenizer: tokenize.WordTokenizer
     chunk_length_schedule: list[int]
     enable_ssml_parsing: bool
+    inactivity_timeout: float
```
davidzhao, line 106: let's use a int, since this is seconds

davidzhao (approved):
lg

PR #1584, opened 2025-03-02
use streaming AudioDecoder to handle compressed formats

davidzhao:
previous instream Mp3Decoder had a couple of flaws:
- it could block main thread for decoding
- it shared a decoder across sessions

both of these could be be responsible for glitches during synthesis.

also changed:
- support for idle session limits with connection pool
- removed pcm formats from TTS
- increased bitrate/quality for 11labs


livekit-agents/livekit/agents/utils/connection_pool.py
```diff
@@ -24,17 +24,20 @@ def __init__(
         self,
         *,
         max_session_duration: Optional[float] = None,
+        touch_on_get: bool = False,
```
theomonnom, line 27: nit: (not sure about this one — got confused about the "touch" naming) 
```suggestion
        refresh_duration_on_get: bool = False,
```

theomonnom (approved):
Thanks!

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -292,6 +279,12 @@ async def _run(self) -> None:
             "voice_settings": voice_settings,
         }
 
+        decoder = utils.codecs.AudioStreamDecoder(
```
theomonnom, line 282: very cool that you don't even need to specify the codec

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -430,3 +430,48 @@ async def __aexit__(
         exc_tb: TracebackType | None,
     ) -> None:
         await self.aclose()
+
+
+class AudioFrameEmitter:
```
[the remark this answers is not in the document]
theomonnom: can u add a ticket to use this class to other TTSs as well?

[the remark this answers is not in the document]
davidzhao: i'll just update it together

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
davidzhao: yeah, libav handles everything

davidzhao:
OAI test is failing, but it's passing when I run from my machine.. it might be related to CI environment

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/log.py
```diff
@@ -1,3 +1,5 @@
 import logging
 
 logger = logging.getLogger("livekit.plugins.aws")
+for logger_name in ["botocore", "aiobotocore"]:
+    logging.getLogger(logger_name).setLevel(logging.INFO)
```
theomonnom, line 5: We have a place in livekit-agents for those noisy loggers:

https://github.com/livekit/agents/blob/c6291d513527a1fdc34c20b4e5cb4be14d387333/livekit-agents/livekit/agents/cli/log.py#L15

davidzhao: ah ok, i'll move over there

PR #1585, opened 2025-03-03
fix multimodal agent interrupts itself during function call

### LINKED ISSUE #1178: MultimodalAgent interrupts itself when response is created from function result
During the function call, the model could return audio letting the user know that it's performing an operation.

While that speech is being played back, if the function call finishes with the result, we would have created an additional response that would interrupt that speech:

For example:

```
Let's say I have a tool called turn_on_the_light.
The agent will say "I'll turn on the {cut-in}} the light is now on".
```

This is happening because of the `response.create()` line
```
if called_fnc.result is not None:
            self.conversation.item.create(tool_call, item_id)
            self.response.create()
```

We should wait until current speech handle is finished before queuing additional speech

longcw:
fixes #1178

### DIFF
--- a/livekit-agents/livekit/agents/multimodal/multimodal_agent.py
+++ b/livekit-agents/livekit/agents/multimodal/multimodal_agent.py
@@ -121,6 +121,11 @@ class _RealtimeAPISession(Protocol):
         self, item_id: str, content_index: int, audio_end_ms: int
     ) -> None: ...
 
+    @property
+    def agent_silent(self) -> asyncio.Event | None:
+        """Event that is set when the playout is done"""
+        pass
+
 
 @dataclass(frozen=True)
 class AgentTranscriptionOptions:
@@ -282,6 +287,15 @@ class MultimodalAgent(utils.EventEmitter[EventTypes]):
                 text_stream=message.text_stream,
                 audio_stream=message.audio_stream,
             )
+
+            if self._session.agent_silent is not None:
+                self._session.agent_silent.clear()
+                self._playing_handle._done_fut.add_done_callback(
+                    lambda _: self._session.agent_silent.set()
+                )
+                self._playing_handle._int_fut.add_done_callback(
+                    lambda _: self._session.agent_silent.set()
+                )
 
         @self._session.on("response_content_done")
         def _response_content_done(message: _ContentProto):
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/multimodal/multimodal_agent.py
```diff
@@ -121,6 +121,11 @@ def _truncate_conversation_item(
         self, item_id: str, content_index: int, audio_end_ms: int
     ) -> None: ...
 
+    @property
+    def agent_silent(self) -> asyncio.Event | None:
```
davidzhao, line 125: should we call this `playout_complete`?

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
```diff
@@ -275,14 +280,24 @@ def _on_content_added(message: _ContentProto):
...
+            self._playing_handle = new_handle = self._agent_playout.play(
                 item_id=message.item_id,
                 content_index=message.content_index,
                 transcription_fwd=tr_fwd,
                 text_stream=message.text_stream,
                 audio_stream=message.audio_stream,
             )
 
+            if self._session.agent_silent is not None:
+                self._session.agent_silent.clear()
```
davidzhao, line 292: would this work in the following case, where two speech are queued back to back.. is it possible for the first event to be fired after the second speech goes in here?

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
longcw: sounds good

livekit-agents/livekit/agents/multimodal/multimodal_agent.py
longcw: There shouldn't be two speeches as the realtime api doesn't allow two responses at a time.

longcw: but maybe we can listen to the `playout_started` and `playout_stopped` of the `_agent_playout` so it more robust.

davidzhao (approved):
lgtm

davidzhao:
the oai tts issue is my fault. i'll fix in my other PR

PR #1587, opened 2025-03-03
handle decoder errors and tts prewarm

davidzhao:
- added a tts.prewarm method to start the connection pool early. we should do this automatically in v1. currently just the stubs
- in agent_output, combined both streaming and non-streaming synthesis paths
- fixed a bug in AudioStreamDecoder where it could fail on close
- deprecated elevenlabs' `optimize_stream_latency` option

### DIFF
--- a/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
+++ b/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
@@ -148,10 +148,8 @@ class TTS(tts.TTS):
 
         return self._session
 
-    async def prewarm(self) -> None:
-        async with self._pool.connection():
-            # open the connection and return it back to the pool
-            pass
+    def prewarm(self) -> None:
+        self._pool.prewarm()
 
     def update_options(
         self,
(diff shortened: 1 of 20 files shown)


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -148,6 +148,11 @@ def _ensure_session(self) -> aiohttp.ClientSession:
 
         return self._session
 
+    async def prewarm(self) -> None:
+        async with self._pool.connection():
+            # open the connection and return it back to the pool
+            pass
```
theomonnom, line 154: I would say to remove it from this PR if this is a TODO

theomonnom (approved):

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
davidzhao: it's not.. this prewarms the connection (by grabbing a connection and returning it to the pool)

alternatively I could add a `.prewarm()` function on the pool to make this less confusing

theomonnom: ahhh, nevermind then

It's because you said

> added a tts.prewarm method to start the connection pool early. we should do this automatically in v1. currently just the stubs

theomonnom: looks good, VoiceAgent.start should call it automatically. but feel free to ignore for now

PR #1588, opened 2025-03-03
V1: LLM plugin changes

jayeshp19:


livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -102,49 +104,53 @@ def is_context_type(ty: type) -> bool:
     return is_call_context
 
 
-def build_legacy_openai_schema(
-    ai_function: AIFunction, *, internally_tagged: bool = False
```
theomonnom, line 106: Can we keep this here? We moved this to the openai plugin, but no idea why we did this

livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -102,49 +104,53 @@ def is_context_type(ty: type) -> bool:
...
-    info = get_function_info(ai_function)
-    schema = _strict.to_strict_json_schema(model)
-
+def serialize_fnc_item(fnc: llm.AIFunction, strict: bool = False) -> dict[str, Any]:
+    model = utils.function_arguments_to_pydantic_model(fnc)
+    if strict:
+        schema = _strict.to_strict_json_schema(model)
+    else:
+        schema = model.model_json_schema()
+    info = utils.get_function_info(fnc)
```
theomonnom, line 149: This should normally be for OAI only (strict mode)

theomonnom: Not fan of sharing this method. Let's just use the two functions accordingly on the plugins..

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py
```diff
@@ -22,22 +21,64 @@ def get_base_url(base_url: Optional[str]) -> str:
...
+
+    return {
+        "type": "function",
+        "function": {
+            "name": fnc["name"],
+            "strict": True,
+            "description": fnc["description"],
+            "parameters": fnc["schema"],
+        },
+    }
```
theomonnom, line 71: this is harder than before, you have to keep track of whats inside the fnc dict

livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -102,49 +104,53 @@ def is_context_type(ty: type) -> bool:
...
 
-def build_legacy_openai_schema(
-    ai_function: AIFunction, *, internally_tagged: bool = False
-) -> dict[str, Any]:
-    """non-strict mode tool description
-    see https://serde.rs/enum-representations.html for the internally tagged representation"""
-    model = function_arguments_to_pydantic_model(ai_function)
-    info = get_function_info(ai_function)
-    schema = model.model_json_schema()
+def serialize_image(image: llm.ImageContent, cache_key: Any) -> dict[str, Any]:
```
theomonnom, line 107: Makes sense! Tho returning the dict may also introduce type unsafety,

theomonnom: Let's return a dataclass

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py
jayeshp19: what if we return data class from `serialize_fnc_item`, because we'd need to modify schema for google and anthropic after serializing acording to their api

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -116,6 +116,7 @@ class FunctionCall(BaseModel):
 
 class FunctionCallOutput(BaseModel):
     id: str = Field(default_factory=lambda: utils.shortuuid("item_"))
+    name: str = Field(default="")
```
theomonnom, line 119: Thinking about this, I think we have a way to always get the name here. Let's remove the default value

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -272,7 +272,10 @@ async def _execute_tools_task(
 
             try:
                 function_model = llm_utils.function_arguments_to_pydantic_model(ai_function)
-                parsed_args = function_model.model_validate_json(fnc_call.arguments)
+                if fnc_call.arguments:
+                    parsed_args = function_model.model_validate_json(fnc_call.arguments)
+                else:
+                    parsed_args = function_model.model_validate({})
```
theomonnom, line 278: When does it happen?

livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
theomonnom, line 1: Nice I like to see lines being removed :)

livekit-plugins/livekit-plugins-google/livekit/plugins/google/utils.py
```diff
@@ -0,0 +1,208 @@
...
+
+class _GeminiJsonSchema:
+    """
+    Transforms the JSON Schema from Pydantic to be suitable for Gemini.
+    based on pydantic-ai implementation
+    https://github.com/pydantic/pydantic-ai/blob/085a9542a7360b7e388ce575323ce189b397d7ad/pydantic_ai_slim/pydantic_ai/models/gemini.py#L809
+    """
+
+    # Type mapping from JSON Schema to Gemini Schema
+    TYPE_MAPPING = {
```
theomonnom, line 106: to make it clear, maybe you can type it

TYPE_MAPPING: dict[str, ...]

livekit-plugins/livekit-plugins-google/livekit/plugins/google/utils.py
```diff
@@ -0,0 +1,208 @@
...
+    fnc = llm.utils.build_legacy_openai_schema(ai_function, internally_tagged=True)
+    json_schema = _GeminiJsonSchema(fnc["parameters"]).simplify()
+    return types.FunctionDeclaration(
+        name=fnc["name"],
+        description=fnc["description"],
+        parameters=json_schema,
+    )
+
+
+class _GeminiJsonSchema:
```
theomonnom, line 98: Nice, would love to see tests for the conversion!

livekit-agents/livekit/agents/llm/chat_context.py
jayeshp19: we don't get name from openai realtime function call output

theomonnom: but we could guess it tho?

jayeshp19: Not sure what you mean, How do we guess it [here](https://github.com/livekit/agents/blob/dev-1.0/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py#L622) ?

PR #1591, opened 2025-03-03
Restrict v0 examples package requirements to <1.0

chrisackermann:
Updating the package requirements for pre-1.0 examples to ensure that we only pull versions prior to v1.0.0 in order to maintain compatibility.


davidzhao (approved):
thank you!

PR #1593, opened 2025-03-04
fix: STT fallbackAdapter use streaming when single frame recognition is unsupported

jayeshp19:


livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -58,8 +58,9 @@ class SpeechEvent:
 
 @dataclass
 class STTCapabilities:
-    streaming: bool
-    interim_results: bool
+    streaming: bool = False
+    interim_results: bool = False
+    single_frame_recognition: bool = True
```
davidzhao, line 63: I think single-frame is the wrong choice of words.. what about calling it `recognize = False` or `only_streaming = True`?

livekit-agents/livekit/agents/stt/fallback_adapter.py
```diff
@@ -54,6 +54,9 @@ def __init__(
             capabilities=STTCapabilities(
                 streaming=all(t.capabilities.streaming for t in stt),
                 interim_results=all(t.capabilities.interim_results for t in stt),
+                single_frame_recognition=all(
```
davidzhao, line 57: I think we should handle this a bit differently.

IMO streaming and recognize should both be supported by fallback adapter. and if one of the plugins don't support a capability, we can use our workaround path.
* if streaming isn't supported, then we should throw an error letting them know they should wrap it with a fallback adapter
* recognize should be supported, you already have a workaround to use streaming to handle recognition

livekit-agents/livekit/agents/stt/stt.py
jayeshp19: `recognize` feels too generic, maybe `batch_recognition` ?

davidzhao: works for me

livekit-agents/livekit/agents/stt/fallback_adapter.py
jayeshp19: I think only one approach is needed, and we shouldn't introduce runtime errors for fallback adapters since users likely have no way to test them.

- if we want to give prefrence to streaming, we can pre-check if any plugin supports it. If so, we can notify users to wrap non-streaming plugins with a streaming adapter.
- If none of the plugins support streaming, we can use recognize method.

In that case, we don't need workaround for recognize method.

theomonnom:
What are the STTs not supporting the recognize method?

jayeshp19:
> What are the STTs not supporting the recognize method?

Azure, AWS, Speechmatics, Assemblyai

theomonnom:
I think we should do this in the v1.0 branch. the FallbackAdapter is currently broken there and we should fix it at the same time (mostly types issues)

livekit-agents/livekit/agents/stt/fallback_adapter.py
davidzhao: within agents, we *only* use the streaming method for STT. so this would cause problems if you are creating a fallback adapter that is reporting that it doesn't streaming.. we'd have to wrap everything with a StreamAdapter.. this is not a good idea.

the reason why recognize is used is to recover from a streaming failure.. I don't see why we shouldn't also declare it support it since you are handling it internally.

livekit-agents/livekit/agents/stt/stt.py
theomonnom: what about calling it just non_streaming?

theomonnom: it isn't really batching

jayeshp19: this PR needs to be updated, I think these changes are not needed. we should notify users to wrap non-streaming plugins inside stream-adapters if one of the tts plugin supports streaming. wdyt?

more context: https://github.com/livekit/agents/pull/1593#discussion_r1978984153

PR #1595, opened 2025-03-04
add transcription node for agent 1.0

longcw:


livekit-agents/livekit/agents/pipeline/task.py
```diff
@@ -34,6 +34,7 @@ def __init__(
         vad: NotGivenOr[vad.VAD | None] = NOT_GIVEN,
         llm: NotGivenOr[llm.LLM | llm.RealtimeModel | None] = NOT_GIVEN,
         tts: NotGivenOr[tts.TTS | None] = NOT_GIVEN,
+        transcriber: NotGivenOr[transcription.TextTranscriber | None] = NOT_GIVEN,
```
theomonnom, line 37: What is this for? Ideally we don't need it and there is just the transcription_node

longcw: so only a method `transcription_node` in the task?

theomonnom: yeah, this is what I've been thinking about

longcw: sg. I thought we want a class. But a inheritable method might be enough for transcription processing

theomonnom: yeah, the idea of the AgentTask, is that if you want to customize any node, you just have to extend it. 
(Users can even get rid of the stt.STT, llm.LLM if they want to just make http requests inside the llm_node for e.g)

longcw: Ok, updated

PR #1598, opened 2025-03-04
integrate krisp to 1.0

davidzhao:
requires `livekit-plugins-noise-cancellation` package.

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io.py
+++ b/livekit-agents/livekit/agents/voice/room_io.py
@@ -36,6 +36,8 @@ class RoomInputOptions:
     """Number of audio channels"""
     text_input_topic: str | None = TOPIC_CHAT
     """Topic for text input"""
+    noise_cancellation: rtc.NoiseCancellationOptions | None = None
+    """Noise cancellation options"""
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/voice/room_io.py
```diff
@@ -71,12 +71,13 @@ def __init__(
         link_to_participant: Optional[rtc.RemoteParticipant | str] = None,
         input_options: RoomInputOptions = DEFAULT_ROOM_INPUT_OPTIONS,
         output_options: RoomOutputOptions = DEFAULT_ROOM_OUTPUT_OPTIONS,
+        noise_cancellation: rtc.NoiseCancellationOptions | None = None,
```
theomonnom, line 74: Can you add this to the RoomInputOptions, this way we can directly use it inside the VoiceAgent without having to manually create the RoomIO. wdyt?

livekit-agents/livekit/agents/voice/voice_agent.py
```diff
@@ -44,6 +44,7 @@ def __init__(
         llm: NotGivenOr[llm.LLM | llm.RealtimeModel] = NOT_GIVEN,
         tts: NotGivenOr[tts.TTS] = NOT_GIVEN,
         userdata: NotGivenOr[Userdata_T] = NOT_GIVEN,
+        noise_cancellation: NotGivenOr[rtc.NoiseCancellationOptions] = NOT_GIVEN,
```
theomonnom, line 47: Oh nvm u added it there too

theomonnom: Not sure if we should, since this is for RoomIO only

livekit-agents/livekit/agents/voice/room_io.py
theomonnom: the start method of VoiceAgent has a RoomInputOptions

PR #1614, opened 2025-03-07
[draft] Avatar integration example for agent 1.0

longcw:


examples/avatar/simli/agent_worker.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @agent.output.audio.on("playback_finished")
+    def on_playback_finished(ev: PlaybackFinishedEvent) -> None:
+        logger.info(
+            "playback_finished",
+            extra={
+                "playback_position": ev.playback_position,
+                "interrupted": ev.interrupted,
+            },
+        )
```
niqodea, line 102: Hi! Just wondering, shouldn't this callback be positioned _before_ awaiting `agent.start`, similarly to how we do it for `room.connect` [here](https://github.com/livekit/python-sdks/blob/e0e0b5a675a6d595384c547428942582cc8d9f9d/livekit-rtc/livekit/rtc/room.py#L329-L338)?

longcw: It could be positioned before `agent.start()` if the `agent.output` is set before start.

`agent.start` is not blocking for run forever, it will return after the tasks created, if you are using the input and output from LK room, it will also wait for the participant joining the room.

niqodea: Ah never mind, it returns upon successfully waiting for another participant!

niqodea: Yes, thank you for the clarification!

examples/avatar/simli/agent_worker.py
```diff
@@ -65,30 +66,26 @@ async def launch_avatar_worker(
...
-        room_output_options=RoomOutputOptions(
-            audio_enabled=False,
-            transcription_enabled=True,
-            agent_transcription_identity=AVATAR_IDENTITY,
+        agent=Agent(
+            instructions="Talk to me!",
         ),
+        room=ctx.room,
+        room_output_options=RoomOutputOptions(audio_enabled=False, transcription_enabled=True),
     )
```
niqodea, line 89: Hi! I am trying to understand why we (used to) have `agent_transcription_identity=AVATAR_IDENTITY` here.

After thinking a bit about it, I thought this was necessary for the local agent to impersonate the avatar agent when writing messages to chat. That is, the local agent would generate text and immediately post it to the room chat, but it would then show up as if the avatar agent wrote that.

However, now I see that we no longer pass this parameter here. Is it because simli in particular does not support transcription, and thus this is not required? Or maybe I didn't really understand what this does/used to do?

Thank you for taking the time!

nicodea: Just realized it should not matter whether simli supports transcriptions since this example only uses an audio-to-video service

niqodea: I see, thanks! I think I had it backwards: I thought the local agent was the one to be hidden. Did you flip the logic recently during development or has it always been like that?

longcw:
@ChenghaoMou That's a different issue, the agent-playground is using legacy transcription event that can only show transcription from the participant streaming out the audio (which is the avatar runner in the example), while the agent transcription is sent under the identity of the agent.

`lk.transcription` is the new text stream for transcription stream, it raised the warning bc their is no handler registered in the avatar runner. We should remove this warning or make it only display once. 

Using [this example](https://github.com/livekit/agents/blob/main/examples/other/datastream-chat-listener.py) to join the same room and listen to the transcription topic you can see the transcription sending using the text stream. I think we are going to update the agent-playground to support.

ChenghaoMou:
> @ChenghaoMou That's a different issue, the agent-playground is using legacy transcription event that can only show transcription from the participant streaming out the audio (which is the avatar runner in the example), while the agent transcription is sent under the identity of the agent.
> 
> `lk.transcription` is the new text stream for transcription stream, it raised the warning bc their is no handler registered in the avatar runner. We should remove this warning or make it only display once.
> 
> Using [this example](https://github.com/livekit/agents/blob/main/examples/other/datastream-chat-listener.py?rgh-link-date=2025-04-07T13%3A26%3A23Z) to join the same room and listen to the transcription topic you can see the transcription sending using the text stream. I think we are going to update the agent-playground to support.

That makes total sense. As soon as I added a handler, I can see the transcription and the warning is gone. Thanks a ton for the detailed response!

PR #1617, opened 2025-03-07
increase idle processes & wait for it before registering 

theomonnom:


davidzhao (approved):

davidzhao:
looks like need to update tests

PR #1618, opened 2025-03-09
Forward stt start_time & end_time

mdelbo:
start_time and end_time were available in ev.alternatives[0] but not getting forwarded - this PR forwards them.

Note: TranscriptionSegment.start_time is an int and ev.alternatives[0].start_time is a float so I converted to milliseconds before type casting. It might be better to change the type of TranscriptSegment.start_time to avoid the conversion and typecast.


davidzhao:
how do you plan to use the timestamps on the client side? since this is the server-side time, it will not correspond to the client's clock.

mdelbo:
Deepgram documentation:

> words: an object containing each word in the transcript, along with its start time and end time (in seconds) from the beginning of the audio stream, and a confidence value.

So, on the client, it enables things like navigating the audio via the transcript segment start_times.

PR #1624, opened 2025-03-10
Add end-of-utterance detection to speechmatics STT plugin

dumitrugutu:


livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/types.py
```diff
@@ -76,7 +76,7 @@ def asdict(self):
 class ConnectionSettings:
     """Defines connection parameters."""
 
-    url: str
+    url: str = "wss://eu2.rt.speechmatics.com/v2"
```
davidzhao, line 79: should this default to the EU server? do you happened to have a global endpoint where it would connect to the closest instance from where the agent is running?

davidzhao (commented):
this is awesome! thanks for the PR

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/types.py
dumitrugutu: we don't have a global endpoint. here's the the link to our docs listing all the endpoints https://docs.speechmatics.com/introduction/authentication#supported-endpoints

PR #1630, opened 2025-03-11
fix fnc calls and add parallel fnc calls test

longcw:
* fix the fnc call output
* fix fnc call speech for realtime model
* TBD: fix the parallel function calls in pipeline agent

### DIFF
--- a/livekit-agents/livekit/agents/voice/task_activity.py
+++ b/livekit-agents/livekit/agents/voice/task_activity.py
@@ -932,7 +932,7 @@ class TaskActivity(RecognitionHooks):
 
                 self._rt_session.interrupt()
                 try:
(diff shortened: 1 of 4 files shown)


theomonnom (approved):

livekit-agents/livekit/agents/voice/task_activity.py
```diff
@@ -940,7 +940,7 @@ async def _read_messages(
 
                 self._rt_session.interrupt()
                 try:
-                    await self._rt_session.generate_reply()
+                    await self.generate_reply()
```
theomonnom, line 943: makes a ton of sense, can u add a comment, so we avoid to do the same mistake in the future

theomonnom (approved):
thx!

PR #1633, opened 2025-03-12
add text input callback

longcw:


livekit-agents/livekit/agents/voice/room_io.py
```diff
@@ -22,6 +31,21 @@
...
 
 
+@dataclass
+class TextInputEvent:
+    text: str
+    info: rtc.TextStreamInfo
+    participant: rtc.RemoteParticipant
+
+
+TextInputCallback = Callable[[TextInputEvent, "VoiceAgent"], Coroutine[None, None, None] | None]
```
theomonnom, line 41: Should we put the VoiceAgent inside the TextInputEvent? Or maybe VoiceAgent being the first argument?

longcw: ok, I'll make the VoiceAgent the first augment.

PR #1634, opened 2025-03-12
add transcribed_track_id and rename to TextStreamSink

longcw:


examples/ds_chat_listener.py
```diff
@@ -96,15 +98,16 @@ async def main(room: rtc.Room, room_name: str):
     def on_text_received(reader: rtc.TextStreamReader, participant_identity: str):
         async def _on_text_received():
             stream_id = reader.info.stream_id
+            track_id = reader.info.attributes.get("lk.transcribed_track_id", None)
```
lukasIO, line 101: maybe use `ATTRIBUTE_TRACK_ID` here?

lukasIO (approved):

examples/ds_chat_listener.py
longcw: oh this is just a example/test script for the text stream, I guess we may remove this file later.

longcw: In RoomIO, it's indeed `ATTRIBUTE_TRACK_ID`

livekit-agents/livekit/agents/voice/avatar/_byte_stream_io.py
```diff
@@ -103,9 +103,9 @@ def clear_buffer(self) -> None:
         task.add_done_callback(self._tasks.discard)
 
 
-class DataStreamAudioReceiver(AudioReceiver):
+class ByteStreamAudioReceiver(AudioReceiver):
```
theomonnom, line 106: I'm still in favor of AudioStreamReceiver, users don't need to know the underlying transport. Datastream is hidden

lukasIO: but it also makes sense that `ByteStreamAudioReceiver` is a specific implementation of `AudioReceiver` ? 
`AudioStreamReceiver` and `AudioReceiver` sound arbitrary/interchangeable

PR #1635, opened 2025-03-12
interrupt agent when user disconnected

longcw:


livekit-agents/livekit/agents/voice/room_io.py
```diff
@@ -299,6 +299,22 @@ def _on_participant_disconnected(self, participant: rtc.RemoteParticipant) -> No
...
+        remote_participants = [
+            p
+            for p in self._room.remote_participants.values()
+            if p.attributes.get(ATTRIBUTE_PUBLISH_FOR) != self._room.local_participant.identity
+        ]
+        if not remote_participants:
+            # no other participants in the room, interrupt the agent
+            try:
+                logger.debug("no other participants in the room, interrupting agent")
+                self._agent.interrupt()
```
longcw, line 312: do we want to force interrupt the agent in this case even `allow_interruptions` is False?

theomonnom: I think it will be tricky to introduce a "force interruption" rn. We can see that after releasing v1. This will happens rarely

theomonnom (approved):

theomonnom:
Let's also listen for room_disconnected?

PR #1643, opened 2025-03-13
resample audio for realtime model and audio sink

longcw:


livekit-agents/livekit/agents/voice/io.py
```diff
@@ -50,13 +50,15 @@ class PlaybackFinishedEvent:
 
 
 class AudioSink(ABC, rtc.EventEmitter[Literal["playback_finished"]]):
-    def __init__(self, *, sample_rate: int | None = None) -> None:
+    def __init__(self, *, sample_rate: int | None = None, num_channels: int | None = None) -> None:
```
theomonnom, line 53: We can't add num_channels yet, we don't have audio remixing

theomonnom: Feel free to ignore the channels for now

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -323,12 +324,13 @@ async def update_instructions(self, instructions: str) -> None:
         )
 
     def push_audio(self, frame: rtc.AudioFrame) -> None:
-        self._msg_ch.send_nowait(
-            InputAudioBufferAppendEvent(
-                type="input_audio_buffer.append",
-                audio=base64.b64encode(frame.data).decode("utf-8"),
+        for f in self._resample_audio(frame):
```
theomonnom, line 327: do we have to use a generator?
 For reference this is how it's done in STT https://github.com/livekit/agents/blob/648c850af6df1c60016751cd5eca596b6c985624/livekit-agents/livekit/agents/stt/stt.py#L278

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -204,16 +204,52 @@ async def _audio_forwarding_task(
...
+            return True
+        return False
+
     try:
+        resampler: rtc.AudioResampler | None = None
         async for frame in tts_output:
             out.audio.append(frame)
-            await audio_output.capture_frame(frame)
+
+            if not out.first_frame_fut.done() and _needs_resampling(frame) and resampler is None:
```
theomonnom, line 222: Do we need to check first_frame_fut, seems like it isn't needed? We can't just check if a resampler is there

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: imaging that the input audio may change the sample rate bc of the change of `agent.input`, we maye need to check the sample rate even there is already a `resampler` so moved the logic to a method to be clear.

livekit-agents/livekit/agents/voice/generation.py
longcw: yes

livekit-agents/livekit/agents/voice/io.py
longcw: Do all TTS always output with one channel? otherwise if user misconfigured the num_channel of TTS and RoomIO.AudioSource it will raise the error.

Maybe that's too much, just let it fail seems ok?

longcw: updated

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
theomonnom: ah I see, I think it is also a OK assumption to say the input must always use the same samplerate

livekit-agents/livekit/agents/voice/io.py
theomonnom: Generally, all TTS outputs one channel (at least by default).
Let's fail for now it's easier

longcw: Yeah, removed for now. The `AudioSource` will fail if the number of channel mismatch.

theomonnom (approved):

theomonnom:
merging to avoid future conflicts since a lot is happening due to the refactor

PR #1645, opened 2025-03-13
Add Azure TTS callbacks as updatable options

anishnag:
We make the callbacks from #1486 updatable options.


theomonnom (approved):
awesome thanks!

PR #1651, opened 2025-03-14
Fix: multimodal - audio_samples should never be negative

DaivdYuan:
Fix agent_playout.py: audio_samples should be non-negative

See: https://github.com/livekit/agents-js/pull/319

Similar issues: https://github.com/livekit/agents-js/issues/304


theomonnom:
is it happening too for python-agents?

DaivdYuan:
Yes, I got errors in the similar fashion, and after I applied fix in the source code, these errors are gone.

PR #1659, opened 2025-03-15
flush transcriptions when segment is received

davidzhao:
we should flush transcriptions in sync even if it's not at sentence boundary

### DIFF
--- a/livekit-agents/livekit/agents/voice/transcription/synchronizer.py
+++ b/livekit-agents/livekit/agents/voice/transcription/synchronizer.py
@@ -108,6 +108,8 @@ class _TextAudioSynchronizer:
 
         self._text_data.pushed_text += text
         self._text_data.sentence_stream.push_text(text)
+        # when delta is received not at the sentence boundary, flush sentence stream
+        self._text_data.sentence_stream.flush()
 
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -232,6 +233,8 @@ async def _forward_input() -> None:
             SessionUpdateEvent(
                 type="session.update",
                 session=session_update_event.Session(
+                    # restore previous instructions on reconnect
```
theomonnom, line 236: This isn't really a reconnect right?

livekit-agents/livekit/agents/voice/transcription/synchronizer.py
```diff
@@ -108,6 +108,8 @@ def push_text(self, text: str) -> None:
 
         self._text_data.pushed_text += text
         self._text_data.sentence_stream.push_text(text)
+        # when delta is received not at the sentence boundary, flush sentence stream
+        self._text_data.sentence_stream.flush()
```
theomonnom, line 112: Discussed in Slack

PR #1664, opened 2025-03-16
update plugins according to new changes

jayeshp19:


livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
```diff
@@ -178,7 +178,7 @@ def __init__(
         region_name: str,
         chat_ctx: ChatContext,
         conn_options: APIConnectOptions,
-        fnc_ctx: list[AIFunction] | None,
+        fnc_ctx: list[FunctionTool] | None,
```
theomonnom, line 181: fnc_ctx has been renamed to tools

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -280,8 +280,9 @@ async def _execute_tools_task(
                 continue
 
             try:
-                function_model = llm_utils.function_arguments_to_pydantic_model(ai_function)
-                parsed_args = function_model.model_validate_json(fnc_call.arguments)
+                function_model = llm_utils.function_arguments_to_pydantic_model(function_tool)
+                json_args = fnc_call.arguments or "{}"
+                parsed_args = function_model.model_validate_json(json_args)
```
jayeshp19, line 285: @theomonnom  How come empty string is converted into json for openai but not for other plugins?

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -280,8 +280,9 @@ async def _execute_tools_task(
                 continue
 
             try:
-                function_model = llm_utils.function_arguments_to_pydantic_model(ai_function)
-                parsed_args = function_model.model_validate_json(fnc_call.arguments)
+                function_model = llm_utils.function_arguments_to_pydantic_model(function_tool)
+                json_args = fnc_call.arguments or "{}"
```
jayeshp19, line 284: we need to cover edge case when function call sends empty arguments. cc: @theomonnom

PR #1665, opened 2025-03-16
Removes ChatManager which should not be used.

yepher:


davidzhao (approved):

matigumma:
what should i use instead?
can any provide an example?

cuongpham-1001:
can you provide an example of alternative solution?

mpadillatabuenca:
what should be used then?

PR #1672, opened 2025-03-18
agent.chat_ctx returns a readonly chat ctx

longcw:


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -190,3 +190,37 @@ def from_dict(cls, data: dict) -> ChatContext:
...
+
+    def add_message(self, **kwargs) -> ChatMessage:
+        raise RuntimeError(_ReadOnlyChatContext.error_msg)
+
+    @property
+    def items(self) -> list[ChatItem]:
+        return self._ImmutableList(self._items)
+
+    def copy(self) -> ChatContext:
+        return ChatContext(self._items.copy())
```
theomonnom, line 226: You don't even need to add those methods, It'll already raise? since items is _ImmutableList? We should maybe remove from_chat_context and use _ImmutableList inside the constructor?

PR #1677, opened 2025-03-19
 add guide: How to create a LiveKit Plugin 

jayeshp19:


nbsp:
would this make more sense on the Wiki tab as opposed to in the root of the repository? or at least a `doc/` subdir

bcherry:
@jayeshp19 what's the status of this? would love to point to it from docs

jayeshp19:
> I'd be very happy to add this to docs, we have a stub section on it already. if you're cool with that, I'll pull this in today and give you the url you can add in to the readme in its place

That works with me

jayeshp19:
> I'd be very happy to add this to docs, we have a stub section on it already. if you're cool with that, I'll pull this in today and give you the url you can add in to the readme in its place

sidenote: this guide is relevant for v0, for v1 our tests are not ready yet + we're not using `pnpm changeset`

bcherry:
oh interesting in that case it is probably better in the repo since it will be version scoped more clearly?

jayeshp19:
> oh interesting in that case it is probably better in the repo since it will be version scoped more clearly?

I don’t have a strong preference. I think having it in the docs is fine too, we’ll have tests in place soon anyway. We can just remove the specific steps about creating a changeset, since we still use changesets, but now they can be generated directly from the pull request.

PR #1678, opened 2025-03-19
add audio energy and syllable for transcription sync

longcw:


livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -19,7 +19,7 @@ def __init__(
         sample_rate: int,
         num_channels: int,
         track_publish_options: rtc.TrackPublishOptions,
-        queue_size_ms: int = 100_000,  # TODO(long): move buffer to python
+        queue_size_ms: int = 300_000,  # TODO(long): move buffer to python
```
longcw, line 22: I have tested to move the buffer to python using `aio.Chan`, actually it uses more ~20M RAM than 300s rust buffer, and changing the rust buffer from 100s to 300s doesn't make many difference.

longcw: actually that makes sense, 1s 16kHz 16bit audio is 32KB, 300s only takes 9.6MB

theomonnom: let's move to python in another ticket, our issue here is that our code doesn't handle when we reach the end of the buffer (and capture_frame starts blocking)
Tho I'm surprised you really see a difference? The rtc.AudioSource is 24000hz no?

theomonnom: the only difference is that the rust buffer is static where Python doesn't. So python buffer may just needs 20s responses max while Rust allocated for way more

livekit-agents/livekit/agents/voice/transcription/_speaking_rate.py
```diff
@@ -0,0 +1,234 @@
...
+        """
+        Calculate speaking rate based on spectral flux.
+        Higher spectral flux correlates with more rapid speech articulation.
+        """
+        # Parameters
+        frame_length = int(sample_rate * 0.025)  # 25ms
+        hop_length = frame_length // 2  # 50% overlap
+
+        # calculate stft
+        _, _, Zxx = signal.stft(
```
theomonnom, line 195: If we only use scipy for this function. I think this could be easily reproducible with numpy?

PR #1682, opened 2025-03-20
Remove unnecessary version pins

hauntsaninja:
These are behind latest

See also https://iscinumpy.dev/post/bound-version-constraints/


theomonnom (approved):
Makes sense! Thanks!

PR #1690, opened 2025-03-20
Add support for different search mode of perplexity.

SiyuanQi:
Perplexity is migrating to new models with search modes: 
https://docs.perplexity.ai/guides/models/sonar#detailed-pricing-breakdown-for-sonar.

Adding support for users to customize.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -108,6 +108,7 @@ class LLM(llm.LLM):
             metadata=metadata,
             max_tokens=max_tokens,
         )
+        self.extra_body: dict[str, Any] = {}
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -478,15 +479,19 @@ def with_perplexity(
         temperature: float | None = None,
         parallel_tool_calls: bool | None = None,
         tool_choice: Union[ToolChoice, Literal["auto", "required", "none"]] = "auto",
+        web_search_options: dict[str, str] | None = {"search_context_size": "low"},
```
jayeshp19, line 482: This will enable search by default for everyone, let keep it `None` by default.

SiyuanQi: Thanks, done.

PR #1694, opened 2025-03-21
fix: double transcript issue  for google stt

jayeshp19:
remove `enable_voice_activity_events`  param from google stt

### DIFF
--- a/tests/test_stt.py
+++ b/tests/test_stt.py
@@ -15,6 +15,7 @@ from livekit.plugins import (
     azure,
     deepgram,
     fal,
+    google,
     openai,
     silero,
     speechmatics,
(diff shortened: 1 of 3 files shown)


tests/test_stt.py
```diff
@@ -72,15 +73,15 @@ async def test_recognize(stt_factory, sample_rate):
...
-    #         languages=["en-AU"],
-    #         model="chirp_2",
-    #         spoken_punctuation=False,
-    #         location="us-central1",
-    #     ),
-    #     id="google.chirp_2",
-    # ),
+    pytest.param(
+        lambda: google.STT(
+            languages=["en-AU"],
```
davidzhao, line 78: why AU?

davidzhao (approved):

tests/test_stt.py
jayeshp19: no particular reason, It was there already, just uncommented, let me update it

PR #1695, opened 2025-03-21
add turn detection options for agent 1.0

longcw:
+ turn detection options
+ realtime API with local VAD example
+ push to talk example
+ support use custom STT with realtime model for user transcription
+ support EOU model with realtime model


livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -70,15 +70,17 @@ def tools(self) -> list[llm.FunctionTool]:
...
+            LLM is a RealtimeModel. Otherwise, it returns the internal mutable chat context.
 
         See Also:
             update_chat_ctx: Method to update the internal chat context.
         """
-        return _ReadOnlyChatContext(self._chat_ctx.items)
+        if isinstance(self.llm, llm.RealtimeModel):
+            return _ReadOnlyChatContext(self._chat_ctx.items)
+        else:
+            return self._chat_ctx
```
longcw, line 83: @theomonnom wdyt about make `agent.chat_ctx` mutable for regular LLM?

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -63,6 +63,10 @@ def __init__(self, *, capabilities: RealtimeCapabilities) -> None:
     def capabilities(self) -> RealtimeCapabilities:
         return self._capabilities
 
+    @property
+    @abstractmethod
+    def server_side_turn_detection(self) -> bool: ...
```
theomonnom, line 68: No need, we already expose it via the capabilities, if turn_detection is True, it means it's being used

livekit-agents/livekit/agents/voice/agent.py
theomonnom: I understand why we want to do it, but I would not create major differences between both today. This may be very confusing when switching an agent using RealtimeLLM to a stateless LLM. The code will break

livekit-agents/livekit/agents/llm/realtime.py
longcw: So should we change the `capabilities` based on the args of Realtime model? I though it was fixed for a type of model.

livekit-agents/livekit/agents/voice/agent.py
longcw: I found it's a little bit annoying and inefficient if we make `chat_ctx` of regular LLM readonly, like in this example https://github.com/livekit/agents/pull/1695/files#diff-88ec3d07b4e65630ae21331f8576f5f677e9d5ad4a52e0e8d15c33b573b81846R35-R36

user should now what LLM is using in an agent for most of the case, or they can use `chat_ctx.readonly` flag to use different methods if it's for both RealtimeModel and LLM.

```python
        chat_ctx = chat_ctx.copy()
        chat_ctx.items.append(new_message)
        self.update_chat_ctx(chat_ctx)
        # we copied the chat_ctx 3 times here
 
        # or 
        assert chat_ctx.mutable  # asset not chat_ctx.readonly
        chat_ctx.items.append(new_message)
```

But I don't have a strong opinion here, maybe we wait for the future. But I think there will be a lot questions about that, @lukasIO just met this issue that modifying the readonly chat_ctx failed silently because it was in a async method... So maybe at least it's better to show the error message in a log then raise the error.

longcw: but anyway, raise the error is better than allow it silently, which is much hard to find lol

lukasIO: how are users expected to use `conversation_item_added` event effectively when they need to reconstruct the whole chat ctx each time they update it 🤔

livekit-agents/livekit/agents/llm/realtime.py
theomonnom: Since you can't access the capabilities statically, I took the decision to make them dynamic. So it is indeed stateful

livekit-agents/livekit/agents/voice/agent.py
theomonnom: Where was the async method that was failing silently? We should absolutely log all errors from tasks we create. 

I think I would say to see that later, my opinion is that a code you see on our docs/github should directly be compatible with what you have locally, no matter what model you're using.

I think as a first step, we should maybe allow update_chat_ctx to not be necessarily `awaited`. But I think it's fine for v1.0? 

I wouldn't worry about the efficiency here, we only do shallow copies, it's very fast

lukasIO: I think a "blessed" way to hook into the chat ctx for a persistence layer should not be an afterthought but rather an integral part of 1.0. 

If it's not the `conversation_item_added` event, it's fine for it to be a custom text sink, but if that's the case, then we need all necessary bits exposed publicly

lukasIO: As I understand the main concern is a difference in behaviour between realtime model and regular llm. 

I guess we could just abstract that away for realtime models? where for realtime models a call to `add_message` would update the chat context with the copy? 
That seems more obvious than requiring to copy + update the chat ctx for all models in user land

longcw: It may failed in user defined async function without log. But anyway I'll revert it for now to unblock this pr.

livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -72,25 +75,35 @@ def flush(self) -> None:
...
+        self._flush_task = asyncio.create_task(self._wait_for_playout())
 
     def clear_buffer(self) -> None:
         super().clear_buffer()
-
         if not self._pushed_duration:
             return
+        self._interrupted_event.set()
+
+    async def _wait_for_playout(self) -> None:
```
longcw, line 86: fixed the clear_buffer sometimes doesn't work

theomonnom: Is it related to the issue you pinged me on Linear? https://linear.app/livekit/issue/AGT-1043/audio-source-clear-queue-only-clears-the-first-second-of-audio#comment-9705354d

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -48,6 +48,7 @@ class ErrorEvent:
 class RealtimeCapabilities:
     message_truncation: bool
     turn_detection: bool
+    speech_to_text: bool
```
theomonnom, line 51: `user_transcription`?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -519,10 +536,15 @@ def push_audio(self, frame: rtc.AudioFrame) -> None:
                         audio=base64.b64encode(f.data).decode("utf-8"),
                     )
                 )
+                self._pushed_duration_s += f.duration
 
     def generate_reply(
         self, *, instructions: NotGivenOr[str] = NOT_GIVEN
     ) -> asyncio.Future[llm.GenerationCreatedEvent]:
+        if not self._realtime_model.capabilities.turn_detection and self._pushed_duration_s > 0:
```
theomonnom, line 544: Could this break the `tool_response`, where we add a new FunctionToolOutput item and call `generate_reply`.  Since push_audio is continuous, commit_audio will always be called, are there any side effects?

livekit-agents/livekit/agents/voice/room_io/_output.py
longcw: not sure if it's related, but I do encountered an issue when testing VAD with realtime model that after `clear_queue` the audio was still playing for tens of seconds. I though it was some race conditions that after `clear_queue` there is new audio frames comes in so `wait_for_playout` didn't return.

But that's also weird, since before `clear_queue` we already interrupted the realtime API so there is no new audio came in, and the buffer of AudioSource is bigger enough we shouldn't have a lot of audio frames buffered somewhere before the `capture_frame` when calling the `clear_queue`.

The current solution fixed it and should be better than before.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -249,7 +267,7 @@ async def _send_task() -> None:
                 try:
                     if isinstance(msg, BaseModel):
                         msg = msg.model_dump(
-                            by_alias=True, exclude_unset=True, exclude_defaults=True
+                            by_alias=True, exclude_unset=True, exclude_defaults=False
```
theomonnom, line 270: The exclude_defaults=True is actually from their SDK

longcw: I guess it's a bug, with that the disable turn detection won't be added to update as the default value is None will they create the session with turn detection enabled lol

theomonnom: lol I see

theomonnom: seems like this is https://github.com/openai/openai-python/issues/2199

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: It's a good point, I'll do some tests

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: actually I think it should never use `exclude_defaults=True`, otherwise you won't be able to update the model back to default config if you updated configs to some other values before. I guess they introduce `exclude_defaults=True` only for init so you only need to send a few configs at the begginning.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: added a `rt_session.commit_input_audio()` so we call the commit input audio buffer in `on_end_of_turn`

davidzhao (approved):
lgtm, only a few nits

examples/push_to_talk.py
```diff
@@ -0,0 +1,77 @@
...
+            instructions="You are Echo.",
+            stt=deepgram.STT(),
+            llm=openai.LLM(model="gpt-4o-mini"),
+            tts=cartesia.TTS(),
+        )
+
+    @function_tool
+    async def talk_to_alloy(self, context: RunContext):
+        """Called when want to talk to Alloy."""
+        return AlloyAgent(), "Transferring you to Alloy."
```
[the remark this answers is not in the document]
longcw: yeah, it was to test if the manual turn detection works for both pipeline and realtime agents.

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -188,7 +189,7 @@ async def _bounce_eou_task() -> None:
                     {"probability": end_of_turn_probability},
                 )
                 unlikely_threshold = turn_detector.unlikely_threshold()
-                if end_of_turn_probability > unlikely_threshold:
+                if end_of_turn_probability < unlikely_threshold:
                     await asyncio.sleep(self.UNLIKELY_END_OF_TURN_EXTRA_DELAY)
```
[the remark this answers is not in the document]
longcw: probably `sleep(current_time - (user_stop_speaking_time + delay))`?

`user_stop_speaking` time could be either `vad_end_speech_time` or `final_transcript_time` (if vad disabled), `delay` is `_min_endpointing_delay` or `UNLIKELY_END_OF_TURN_EXTRA_DELAY`.

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -130,6 +131,10 @@ def generate_reply(
         instructions: NotGivenOr[str] = NOT_GIVEN,
     ) -> asyncio.Future[GenerationCreatedEvent]: ...  # can raise RealtimeError on Timeout
 
+    # commit the input audio buffer to the server
+    @abstractmethod
+    def commit_input_audio(self) -> None: ...
```
theomonnom, line 136: Do you think we should rename to just `commit_audio`? since we already have `push_audio` that doesn't imply the "input"

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -101,6 +122,10 @@ def userdata(self) -> Userdata_T:
     def userdata(self, value: Userdata_T) -> None:
         self._userdata = value
 
+    @property
+    def turn_detection_mode(self) -> TurnDetectionMode | None:
```
theomonnom, line 126: Can we remove this in favor of turn_detection?

theomonnom: returning _TurnDetector as well if needed

livekit-agents/livekit/agents/llm/realtime.py
longcw: Sounds good to me.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -74,14 +75,26 @@ class _InputAudioTranscription:
     model: Literal["whisper-1"] = "whisper-1"
 
 
+@dataclass
+class _TurnDetection:
```
theomonnom, line 79: Should we expose directly the OAI class?

theomonnom: https://github.com/openai/openai-python/blob/f66d2e6fdc51c4528c99bb25a8fbca6f9b9b872d/src/openai/types/beta/realtime/session_update_event_param.py#L69

It is also a TypeDict, so maybe easier for users?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -74,14 +75,26 @@ class _InputAudioTranscription:
...
+    type: Literal["server_vad"] = "server_vad"
+    create_response: bool | None = True
+    interrupt_response: bool | None = True
+    prefix_padding_ms: int | None = 300
+    silence_duration_ms: int | None = 500
+    threshold: float | None = 0.65
+
+
 DEFAULT_INPUT_AUDIO_TRANSCRIPTION = _InputAudioTranscription()
```
theomonnom, line 88: Maybe we can do the same for 
https://github.com/openai/openai-python/blob/f66d2e6fdc51c4528c99bb25a8fbca6f9b9b872d/src/openai/types/beta/realtime/session_update_event_param.py#L27

longcw: sounds good, will update it.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -542,37 +608,67 @@ def on_vad_inference_done(self, ev: vad.VADEvent) -> None:
...
+
         if self._current_speech is not None:
             if not self._current_speech.allow_interruptions:
                 logger.warning(
-                    "skipping user input, current speech generation cannot be interrupted",
+                    "skipping reply to user input, current speech generation cannot be interrupted",
                     extra={"user_input": new_transcript},
                 )
+                # in agent.on_end_of_turn user can ignore the message or save it for later
+                await self._agent.on_end_of_turn(
```
theomonnom, line 660: I don't think we should emit the event in this case?

livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -253,7 +253,9 @@ async def on_exit(self) -> None:
         """Called when the task is exited"""
         pass
 
-    async def on_end_of_turn(self, chat_ctx: llm.ChatContext, new_message: llm.ChatMessage) -> None:
+    async def on_end_of_turn(
+        self, chat_ctx: llm.ChatContext, new_message: llm.ChatMessage, generating_reply: bool
```
theomonnom, line 257: I'm not a fan of having the generating_reply field here, is it needed? 
If turn_detection is set to manual, you already expect that no generation will be triggered in the current Agent

theomonnom: I see, but I think it should be handled separately? This should be part of this event? 
But more like a flag saying "Continue to push audio even if allow_interruptions is False" ?

PR #1699, opened 2025-03-21
support openai new voice model for stt and tts

jayeshp19:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -160,16 +160,16 @@ async def _recognize_impl(
...
             )
 
             return stt.SpeechEvent(
                 type=stt.SpeechEventType.FINAL_TRANSCRIPT,
                 alternatives=[
                     stt.SpeechData(
-                        text=resp.text or "",
-                        language=resp.language or config.language or "",
+                        text=resp or "",
+                        language=config.language or "",
```
theomonnom, line 172: I think we supported automatic language recognition with Whisper too?

jayeshp19: hmm the new models don’t support `verbose_json`, perhaps we can add a conditional check instead.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py
```diff
@@ -95,3 +97,9 @@ def _build_oai_image_content(image: llm.ChatImage, cache_key: Any):
     raise ValueError(
         "LiveKit OpenAI Plugin: ChatImage must be an rtc.VideoFrame or a URL"
     )
+
+
+def _get_response_format(model: STTModels | str):
```
theomonnom, line 102: Not need for an utility if it is used once, u can do it inline

theomonnom (approved):

jayeshp19:
we have better stt with realtime streaming supported, https://github.com/livekit/agents/pull/1701
closing this

PR #1701, opened 2025-03-21
Support for OpenAI's new TTS and STT models

davidzhao:
also introducing streaming STT support

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
@@ -300,7 +300,7 @@ class STT(stt.STT):
                     data,
                     "audio/wav",
                 ),
-                model=self._opts.model, # type: ignore
+                model=self._opts.model,  # type: ignore
                 language=config.language,
                 prompt=prompt,
                 response_format=format,
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
@@ -180,6 +180,7 @@ class STT(stt.STT):
         self,
         *,
         language: str | None = None,
+        conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
     ) -> "SpeechStream":
         config = self._sanitize_options(language=language)
         stream = SpeechStream(
@@ -299,7 +300,7 @@ class STT(stt.STT):
                     data,
                     "audio/wav",
                 ),
-                model=self._opts.model,
+                model=self._opts.model, # type: ignore
                 language=config.language,
                 prompt=prompt,
                 response_format=format,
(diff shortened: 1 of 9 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -132,6 +176,100 @@ def with_groq(
             prompt=prompt,
         )
 
+    def stream(
```
jayeshp19, line 179: This is great 🥳

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -132,6 +176,100 @@ def with_groq(
             prompt=prompt,
         )
```
jayeshp19, line 177: we should update capabilities and make streaming false when using with groq

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -48,21 +72,36 @@ def __init__(
...
+            prompt: Optional text prompt to guide the transcription. Only supported for whisper-1.
+            vad_threshold: Voice activity detection threshold (0.0 to 1.0).
+            prefix_padding_ms: Padding to add before speech in milliseconds.
+            silence_duration_ms: Duration of silence to consider end of speech in milliseconds.
+            noise_reduction_type: Type of noise reduction to apply. "near_field" or "far_field"
+                This isn't needed when using LiveKit's noise cancellation.
+            base_url: Custom base URL for OpenAI API.
+            api_key: Your OpenAI API key. If not provided, will use the OPENAI_API_KEY environment variable.
+            client: Optional pre-configured OpenAI AsyncClient instance.
         """
```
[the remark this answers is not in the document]
davidzhao: the realtime endpoint supports whisper as well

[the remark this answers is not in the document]
jayeshp19: Oh interesting, it's not mentioned here: https://platform.openai.com/docs/models/whisper-1

jayeshp19 (approved):
Lgtm!

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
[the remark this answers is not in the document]
davidzhao: https://platform.openai.com/docs/guides/realtime-transcription#realtime-transcription-sessions

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -130,7 +177,102 @@ def with_groq(
...
+            "type": "transcription_session.update",
+            "session": {
+                "input_audio_format": "pcm16",
+                "input_audio_transcription": {
+                    "model": self._opts.model,
+                    "prompt": self._opts.prompt or "",
+                    "language": self._opts.language or "",
+                },
+                "turn_detection": {
+                    "type": "server_vad",
```
theomonnom, line 235: Do you think we can use semantic_vad? or it is the same label now?

theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
davidzhao: we are not relying on this for turn detection, so focusing on a simple, but predictable behavior here.

davidzhao: but agreed we should test it and see how it performs

PR #1706, opened 2025-03-22
update plugin versions, uv update for new plugins

davidzhao:


theomonnom (approved):
will delete the package.json tmrw

PR #1707, opened 2025-03-23
feat(OpenAI STT): add support for semantic_vad

chasemcdo:
Current implementation only allows for using `server_vad`, but many use cases would benefit from the addition of `semantic_vad` support

### DIFF
--- a/.changeset/dirty-countries-rescue.md
+++ b/.changeset/dirty-countries-rescue.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-openai": patch
+---
+
+feat(OpenAI STT): add support for semantic_vad
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
@@ -64,6 +64,8 @@ class _STTOptions:
     silence_duration_ms: int = 350
     prompt: str | None = None
     noise_reduction_type: str | None = None
+    vad_type: str = "server_vad"
+    vad_eagerness: str = "auto"
 
 
 class STT(stt.STT):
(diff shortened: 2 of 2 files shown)


davidzhao (approved):
lg!

I'm going to merge it and then update the params, it's getting confusing with the diff paths

chasemcdo:
Sounds good! Definitely agree that it’s getting a bit non-trivial. What’s the release schedule usually look like? Would love to get this into prod tomorrow

MajorTal:
Awesome work!
Did you forget to update livekit/plugins/openai/realtime/realtime_model.py ?

chasemcdo:
> Awesome work! Did you forget to update livekit/plugins/openai/realtime/realtime_model.py ?

Would appear so. Though actual usage of this is blocked by the V1 migration allowing for the exclusion of VAD during pipeline instantiation.

davidzhao:
v1 would support using STT final transcripts to determine end of turn. (implemented [here](https://github.com/livekit/agents/pull/1695/files#diff-ae28db7b71d42ec453a1b46d74382b9c06c4678645e46f08a26a4452a83ad1ccR36))

with that said, in our testing `semantic_vad` did not perform as well as our turn detector model.

chasemcdo:
Interesting to hear that you found the LK model to outperform the OpenAI one. We are particularly interested in the OAI performance when it comes to isolating intended speech from background noise.

we’ve found that the current model can end up interpreting background conversation as incomplete thoughts leading to extended delays in response.

the hope was that the near_field noise cancellation on OAI’s model might alleviate some of this as well as broaden language support.

davidzhao:
noise cancellation is different from turn detection. have you checked out our Krisp integration on the agent side?

chasemcdo:
We’ve got krisp enabled on our SIP Trunk, but haven’t noticed a huge difference with that. Does the agent based krisp do anything separate from that?

davidzhao:
yeah, SIP trunk krisp is noise cancellation, where as agents are running a bigger model that also cancels background voice. give it a shot. I would recommend it over the SIP version

chasemcdo:
Did not know they were different. Will turn that on thanks! Is it self hosted and will need additional resources or running through an API?

PR #1710, opened 2025-03-23
add BackgroundAudio helper

theomonnom:


livekit-agents/livekit/agents/voice/background_audio.py
```diff
@@ -0,0 +1,88 @@
...
+
+class BackgroundAudio:
+    def __init__(
+        self,
+        room: rtc.Room,
+        *,
+        ambiant_sound: NotGivenOr[str] = NOT_GIVEN,
+        thinking_sound: NotGivenOr[str] = NOT_GIVEN,
+        agent_session: NotGivenOr[AgentSession] = NOT_GIVEN,
+        track_publish_options: NotGivenOr[rtc.TrackPublishOptions] = NOT_GIVEN,
```
theomonnom, line 22: I don't like the name & API, any suggestions?

pyproject.toml
```diff
@@ -34,6 +34,7 @@ dev = [
     "ruff",
     "pytest-asyncio>=0.25.3",
     "jiwer>=3.1.0",
+    "pyav>=14.1.0",
```
davidzhao, line 37: if we are pulling this in, we can ignore the `codecs` optional dep?

livekit-agents/livekit/agents/voice/background_audio.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if self._agent_session:
+            self._agent_session.on("agent_state_changed", self._agent_state_changed)
+
+    async def _agent_state_changed(self, ev: AgentStateChangedEvent) -> None:
+        if ev.state == "thinking" and self._thinking_sound:
+            self._audio_mixer.add_stream(audio_frames_from_file(self._thinking_sound))
+
+    @log_exceptions(logger=logger)
+    async def _mixer(self) -> None:
```
davidzhao, line 68: nit, for clarity
```suggestion
    async def _run_mixer(self) -> None:
```

livekit-agents/livekit/agents/voice/background_audio.py
```diff
@@ -0,0 +1,88 @@
...
+from ..types import NotGiven, NOT_GIVEN, NotGivenOr
+from ..utils.audio import audio_frames_from_file
+from ..utils.aio import cancel_and_wait
+from ..utils import log_exceptions
+from ..log import logger
+
+from .events import AgentStateChangedEvent
+
+
+class BackgroundAudio:
```
davidzhao, line 14: we should also offer a `play(path_to_file | audio_frames)` API.. that'll be more flexible because users might want to only play a specific audio triggered by certain function calls

livekit-agents/livekit/agents/resources/keyboard-typing.ogg
davidzhao, line 1: if we are including them by default, could we export a constant as well? i.e.

```python
from livekit.agents import BackgroundAudio

ba = BackgroundAudio(
    ambient_sound=BackgroundAudio.PRESET_OFFICE_AMBIENCE,
    ...
)
```

davidzhao (approved):

pyproject.toml
theomonnom: this is in the dev dependencies

livekit-agents/livekit/agents/voice/background_audio.py
theomonnom: yeah 100%

PR #1720, opened 2025-03-24
emit conversation item added event

longcw:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1171,9 +1186,30 @@ async def _read_messages(
                 self._session._update_agent_state(AgentState.LISTENING)
 
             # TODO(theomonnom): truncate message (+ OAI serverside mesage)
+
+            for msg_id, text_out, _ in message_outputs:
```
theomonnom, line 1190: It is OK to assume there is only one generation, we don't support more then 1

theomonnom:
Do we handle the conversation items created by the Realtime API? When we use their turn_detection/receiving the generation_created_event

longcw:
> Do we handle the conversation items created by the Realtime API? When we use their turn_detection/receiving the generation_created_event

I think so, I added the event after every `perform_audio_forwarding` and audio played out. Is that sufficient?

PR #1721, opened 2025-03-24
Add option to enable Claude Thinking

bryanhoulton:
This is my first contribution and I'm not sure how to test, so I'm opening as a draft. Any pointers are appreciated!


bryanhoulton:
Struggling a lot with the local dev here. Running into issues running checks.

PR #1726, opened 2025-03-25
expose turn_detection options with openai STT

davidzhao:
exposing vad options when in realtime mode: https://platform.openai.com/docs/guides/realtime-vad

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
@@ -41,6 +41,9 @@ from livekit.agents.utils import AudioBuffer
 
 import openai
 from openai.types.audio import TranscriptionVerbose
+from openai.types.beta.realtime.transcription_session_update_param import (
+    SessionTurnDetection,
+)
 
 from .log import logger
 from .models import GroqAudioModels, STTModels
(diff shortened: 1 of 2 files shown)


jayeshp19 (approved):
LGTM

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -59,13 +62,9 @@ class _STTOptions:
     model: STTModels | str
     language: str
     detect_language: bool
-    vad_threshold: float = 0.5
-    prefix_padding_ms: int = 600
-    silence_duration_ms: int = 350
+    turn_detection: SessionTurnDetection
```
theomonnom, line 65: Much better thx 🙏

longcw: one question here: `SessionTurnDetection` is a TypedDict, what if we only set part of the params like `{prefix_padding_ms: 200}`. It should work but the they type is incorrect?

davidzhao: I think if they choose to override, they should pass in the params correctly. since this is a provider-specific plugihn, it'll be on the user to read their platform docs and pass in the correct values

PR #1728, opened 2025-03-25
Support for function calls that does not wait for the response

msaelices:
The proposal is to be able to define a function like this one:
```python
import time
class TestFunctionContext(llm.FunctionContext):
    @llm.ai_callable(wait_for_response=False)
    def long_function(self):
        # simulate a long running task
        time.sleep(10)
```


martin-purplefish:
It seems like you could achieve this by tracking the task yourself; e.g.

```
    def __init__(self):
      self._tasks = set()
    @llm.ai_callable(wait_for_response=False)
    def long_function(self):
        # simulate a long running task
        task = asyncio.create_task(time.sleep(10))  #wrap your long function in a task.
        self._tasks.add(task)
 ```

PR #1730, opened 2025-03-25
Fix STT/TTS plugins types for v1

jayeshp19:


livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
```diff
@@ -304,11 +304,11 @@ def _create_speech_recognizer(
 ) -> speechsdk.SpeechRecognizer:
     # let the SpeechConfig constructor to validate the arguments
     speech_config = speechsdk.SpeechConfig(
-        subscription=config.speech_key,
-        region=config.speech_region,
-        endpoint=config.speech_endpoint,
-        host=config.speech_host,
-        auth_token=config.speech_auth_token,
+        subscription=config.speech_key if is_given(config.speech_key) else None,
```
theomonnom, line 307: u can do `subscription=config.speech_key or None` instead

jayeshp19: config.speech_key can be NOT_GIVEN aswell? in that case it won't work as expected.

theomonnom: NOT_GIVEN is the same as False

theomonnom: will merge, I need this

PR #1733, opened 2025-03-25
support external urls for image input

jayeshp19:


livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -109,27 +110,44 @@ def is_context_type(ty: type) -> bool:
...
+                    f"Unsupported media type {media_type}. Must be jpeg, png, webp, or gif"
+                )
+
+            return SerializedImage(
+                data_bytes=encoded_data,
+                media_type=media_type,
+                inference_detail=image.inference_detail,
+            )
+        else:
+            return SerializedImage(
```
bcherry, line 145: its probably worth actually validating that it starts with `http:` or `https:` raising a value error otherwise. I don't think there's any scenario where e.g. a `file:` url could work

livekit-plugins/livekit-plugins-google/livekit/plugins/google/utils.py
```diff
@@ -81,6 +83,16 @@ def to_chat_ctx(
 
 def _to_image_part(image: llm.ImageContent, cache_key: Any) -> types.Part:
     img = llm.utils.serialize_image(image)
+    if img.external_url:
+        if img.media_type:
+            mime_type = img.media_type
+        else:
+            logger.warning(
+                "No media type provided for image, using default image/jpeg. "
+                "Please provide a media type for better performance."
```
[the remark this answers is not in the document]
jayeshp19: Yes, it works. or should we raise value error instead?

livekit-agents/livekit/agents/llm/utils.py
jayeshp19: t could be a cloud bucket URL, such as: [gs://generativeai-downloads/images/scones.jpg](https://github.com/googleapis/python-genai?tab=readme-ov-file#provide-a-non-function-call-part)

bcherry: TIL!

livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -109,27 +110,44 @@ def is_context_type(ty: type) -> bool:
 
 @dataclass
 class SerializedImage:
-    data_bytes: bytes
-    media_type: str
     inference_detail: str
+    media_type: str | None
```
lukasIO, line 114: how does media_type differ from a mime_type? 

If it doesn't, let's just call it mime_type here as well for consistency?

lukasIO (approved):

livekit-plugins/livekit-plugins-google/livekit/plugins/google/utils.py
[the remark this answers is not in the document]
davidzhao: should we warn if there's a sane default? let's do this:
* make it a debug message
* remove the line about better performance (we are just picking a default, which is a lossy format)

PR #1740, opened 2025-03-26
Example for Livekit with mem0's enhanced memory capabilities

parshvadaftari:
Added an example of a voice assistant using [mem0](https://mem0.ai/)'s enhanced memory capabilities. This example is a simple travel guide voice assistant that uses mem0 to remember conversations and user's preferences based on the `USER_ID` parameter.

### DIFF
--- a/examples/voice-assistant/memory-assistant/.env.example
+++ b/examples/voice-assistant/memory-assistant/.env.example
@@ -1,0 +1,6 @@
+LIVEKIT_URL=your_livekit_url
+LIVEKIT_API_KEY=your_livekit_api_key
+LIVEKIT_API_SECRET=your_livekit_api_secret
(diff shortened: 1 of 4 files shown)


parshvadaftari:
@theomonnom Hey can you please review this PR and let me know if this requires changes or will it be merger?

theomonnom:
Hey, sorry for the delay, this is a good example but the v1.0 API changed quite a lot. Seems like it is only compatible for 0.x versions

parshvadaftari:
> Hey, sorry for the delay, this is a good example but the v1.0 API changed quite a lot. Seems like it is only compatible for 0.x versions

I raised this PR  before the v1.0 was made public. I can have an example for the v1.0 API as well. Will make changes according to the v1.0 api of the livekit.

PR #1742, opened 2025-03-26
Added a Semantic VAD configuration option

MajorTal:
There is a new VAD mechanism available to the OpenAI realtime model: semantic VAD.
https://platform.openai.com/docs/guides/realtime-vad#semantic-vad
I've added this option to the OpenAI realtime plugin.


davidzhao (approved):
lgtm. thanks!

PR #1748, opened 2025-03-27
Expose error handler for agent session

Shubhrakanti:
Close the `AgentSession` When there is a fatal error in the LLM/TTS/STT. Emit a `session_closed` event which the used can act on.


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -423,3 +452,77 @@ def _on_text_output_changed(self) -> None:
...
+                    asyncio.create_task(self.aclose())
+                elif result == ErrorCallbackResult.CONTINUE:
+                    logger.info(f"Continuing session despite unrecoverable error in {component}")
+                else:
+                    logger.warning(f"Unknown error callback result: {result}")
+                    asyncio.create_task(self.aclose())
+            except Exception as e:
+                logger.exception(f"Error in unrecoverable error callback: {e}")
+                # If the callback raises an exception, end the session as a precaution
+                asyncio.create_task(self.aclose())
```
Shubhrakanti, line 515: Is this the right thing to do? If we just pass the session to the user then they can handle closing it and all other errors that happen in the call back function itself.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -54,6 +56,26 @@ class VoiceOptions:
 """
 
 
+@dataclass
+class UnrecoverableErrorInfo:
+    """Information about an unrecoverable error that occurred in a component."""
+    component: str
+    component_instance: Any
```
Shubhrakanti, line 63: should be a Union[LLM | SST | TTS]

theomonnom:
I think the right path would be to reuse the `metrics_collected` event.
There is a unused [Error](https://github.com/livekit/agents/blob/5790e2084e820a7a667b684b4dd82dd0c8b54763/livekit-agents/livekit/agents/metrics/base.py#L9) field today. Ideally a new metric is sent every time we get an error (for every try). If every attempt failed, we could maybe add a field like: `recoverable` inside the metrics.

This way we know when something isn't recoverable and we can emit a new event inside the AgentSession.
Also let's make it an event (use the rtc.EventEmitter), I don't think a callback is needed here, we don't need the return value. This event is only fired when it's too late, no continue is possible.

examples/errorcall_back_agent.py
```diff
@@ -0,0 +1,54 @@
...
+        llm=openai.LLM(),
+        tts=cartesia.TTS(),
+        vad=silero.VAD.load(),
+    )
+
+    close_task = None
+
+    @session.on("metrics_collected")
+    def on_metrics_collected(ev: MetricsCollectedEvent):
+        nonlocal close_task
```
Shubhrakanti, line 39: We need to keep a handle to the task in a broader scope so GC doesn't mess with the `sessions.aclose` task. We can't do 

```python
@session.on("metrics_collected")
async def on_metrics_collected(ev: MetricsCollectedEvent):
    ...
    await session.aclose()
```
since the `EventEmitter` can't take in coroutines.

This still feels really complicated especially  since it's user code. Is there a better way to do this?

Shubhrakanti (commented):
@theomonnom is this what you were envisioning?

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -126,13 +127,44 @@ async def _main_task(self) -> None:
...
+                    request_id="",
+                    ttft=0.0,
+                    duration=0.0,
+                    cancelled=self._task.cancelled(),
+                    label=self._llm._label,
+                    completion_tokens=0,
+                    prompt_tokens=0,
+                    total_tokens=0,
+                    tokens_per_second=0.0,
+                )
```
theomonnom, line 141: Can we add default values to the metrics instead?

theomonnom:
Let's do the same for TTS/STT

RealtimeAPI can be another PR

examples/errorcall_back_agent.py
theomonnom: We should still emit another event, I meant to keep metrics_collected for llm/tts/stt. But the AgentSession can have another one?

+1 for the async usage, I'm not sure if we should really make it a callback

examples/errorcall_back_agent.py
```diff
@@ -0,0 +1,54 @@
...
+                close_task = asyncio.create_task(session.aclose())
+                close_task.add_done_callback(lambda _: logger.info("Session closed"))
+        return
+
+    await session.start(agent=MyTask(), room=ctx.room)
+
+
+if __name__ == "__main__":
+    # set the OPENAI_API_KEY to a invalid key to simulate an error
+    os.environ["OPENAI_API_KEY"] = "invalid"
```
theomonnom, line 53: For the example, we should find a better example to simulate failures, maybe editing the llm_node

examples/errorcall_back_agent.py
```diff
@@ -0,0 +1,82 @@
...
+        super().__init__(*args, **kwargs)
+
+    async def _run(self):
+        raise APIStatusError(  # noqa: B904
+            message="This is a test error",
+            status_code=500,
+            request_id="123",
+            body="test",
+            retryable=False,
+        )
```
Shubhrakanti, line 49: @theomonnom is this what you meant?

examples/errorcall_back_agent.py
```diff
@@ -0,0 +1,82 @@
...
+
+async def entrypoint(ctx: JobContext):
+    await ctx.connect()
+
+    # Define an async error handler
+    async def on_agent_error(session: AgentSession, ev: AgentErrorEvent):
+        if not ev.error.retryable or ev.error.attempts_remaining == 0:
+            logger.info("Ran into an unrecoverable LLM error, ending session.")
+            await session.aclose()
+            logger.info("Session closed")
```
Shubhrakanti, line 67: I like this a lot better. Let the `AgentSession` handle the complexity of keeping a reference to the coroutine.

examples/errorcall_back_agent.py
```diff
@@ -0,0 +1,82 @@
...
+    # Create session with the error handler
+    session = AgentSession(
+        stt=deepgram.STT(),
+        llm=ErrorLLM(),  # pass in a custom LLM that raises an error
+        tts=cartesia.TTS(),
+        vad=silero.VAD.load(),
+        error_handler=on_agent_error,  # Pass the async error handler
+    )
+
+    await session.start(agent=MyTask(), room=ctx.room)
```
Shubhrakanti, line 78: @theomonnom do you think the `error_handler` should be passed into the constructor of `AgentSession` or as a parameter to `AgentSession.start()`?

livekit-agents/livekit/agents/metrics/base.py
```diff
@@ -5,24 +5,26 @@
...
 
 class LLMMetrics(BaseModel):
     type: Literal["llm_metrics"] = "llm_metrics"
-    label: str
-    request_id: str
     timestamp: float
-    duration: float
-    ttft: float
+    label: str
     cancelled: bool
```
Shubhrakanti, line 18: Didn't set defaults on these since the defaults are never used.

livekit-agents/livekit/agents/metrics/base.py
```diff
@@ -5,24 +5,26 @@
 from pydantic import BaseModel
 
 
-class Error(BaseModel):
+class AgentComponentError(BaseModel):
```
Shubhrakanti, line 8: Through it's a better name. Open to feedback

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -115,6 +116,11 @@ class SpeechCreatedEvent(BaseModel):
     """The speech handle that was created"""
 
 
+class AgentErrorEvent(BaseModel):
+    type: Literal["agent_error"] = "agent_error"
+    error: AgentComponentError
+
+
```
Shubhrakanti, line 123: @theomonnom is this what you meany by making a new event?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -547,6 +548,8 @@ def _on_metrics_collected(self, ev: AgentMetrics) -> None:
         if speech_handle := _SpeechHandleContextVar.get(None):
             ev.speech_id = speech_handle.id
         self._session.emit("metrics_collected", MetricsCollectedEvent(metrics=ev))
+        if ev.error:
+            self._session.emit("agent_error", AgentErrorEvent(error=ev.error))
```
Shubhrakanti, line 552: Should this be done here or in `AgentSession`? Since there is already a callback registered to `metrics_collected` in `AgentActivity` I chose to do it here.

Shubhrakanti (commented):
@theomonnom let me know if this structurally looks okay, then I will add in the TTS and STT.

examples/errorcall_back_agent.py
```diff
@@ -0,0 +1,81 @@
...
+            instructions="You are a helpful assistant that can answer questions and help with tasks.",  # noqa: E501
+        )
+
+
+async def entrypoint(ctx: JobContext):
+    await ctx.connect()
+
+    # Define an async error handler
+    async def on_agent_error(session: AgentSession, ev: AgentErrorEvent):
+        if not ev.error.retryable or ev.error.attempts_remaining == 0:
```
Shubhrakanti, line 63: Is this confusing? Should we only be emitting an error when it's unrecoverable?

What's a use case in which a user would want a callback defined on an error with retries left?

livekit-agents/livekit/agents/metrics/base.py
theomonnom: I'm not sure about this name, we've never named this pipeline "component" 

It isn't the end of the world to name this Error since users access. the type by doing metrics.Error

examples/errorcall_back_agent.py
theomonnom: No, we have a llm_node inside the Agent class

examples/errorcall_back_agent.py
Shubhrakanti: Discussed offline we will move to a `session_closed` event being emitted.

examples/errorcall_back_agent.py
Shubhrakanti: Discussed offline moving away from async callback

livekit-agents/livekit/agents/voice/events.py
Shubhrakanti: Discussed offline moving to a "session_closed" event

livekit-agents/livekit/agents/voice/agent_activity.py
Shubhrakanti: Discussed offline moving to a "session_closed" event

examples/errorcall_back_agent.py
Shubhrakanti: Not sure what you mean here.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -547,6 +550,18 @@ def _on_metrics_collected(self, ev: AgentMetrics) -> None:
...
+                component = self._get_component_from_metrics(ev)
+                self._session._create_session_close_task(component, ev.error)
+
+    def _get_component_from_metrics(self, metrics: AgentMetrics) -> LLM | STT | TTS:
+        if isinstance(metrics, LLMMetrics):
+            return self.llm
+        elif isinstance(metrics, STTMetrics):
+            return self.stt
+        elif isinstance(metrics, TTSMetrics):
+            return self.tts
```
Shubhrakanti, line 564: Is there a better way to do this or is this okay?

Shubhrakanti (commented):
Update based on [slack discussion](https://live-kit.slack.com/archives/C07TWVC0W4A/p1743626761816229)

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -126,13 +126,40 @@ async def _main_task(self) -> None:
             try:
                 return await self._run()
             except APIError as e:
+                error_metrics = LLMMetrics(
+                    timestamp=time.time(),
+                    label=self._llm._label,
+                    cancelled=self._task.cancelled(),
```
theomonnom, line 132: cancelled would always be False here?

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -126,13 +126,40 @@ async def _main_task(self) -> None:
...
+                    self._llm.emit("metrics_collected", error_metrics)
                     raise
                 elif i == self._conn_options.max_retry:
+                    error_metrics.error = Error(
+                        error=e.message,
+                        retryable=e.retryable,
+                        attempts_remaining=0,
+                        component=self._llm,
+                    )
+                    self._llm.emit("metrics_collected", error_metrics)
```
theomonnom, line 150: I think we can reduce code duplication here

theomonnom:
I'm not sure if we emit the metrics_collected events correctly?
Seems like we're going to have duplicates since we also emit metrics inside _metrics_monitor_task

examples/errorcall_back_agent.py
theomonnom: I was thinking about https://github.com/livekit/agents/blob/781faca7dd0184b0202f7215b615604af7ac2993/livekit-agents/livekit/agents/voice/agent.py#L314
But I guess metrics_collected only cover the LLM class itself. So it may not be possible

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -363,6 +367,21 @@ def update_agent(self, agent: Agent) -> None:
...
+
+    async def _drain_and_close_session(self) -> None:
+        """Close the session after waiting for any current activity to drain."""
+        # Drain the current activity if it exists
+        if self._activity is not None:
+            await self._activity.drain()
+
+        # Finally close the session
+        await self.aclose()
+
```
theomonnom, line 384: I was thinking about adding the drain() method to AgentSession directly, it can be a public method.

Impl is just:
```
        if self._activity is not None:
            await self._activity.drain()
```

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -115,6 +119,13 @@ class SpeechCreatedEvent(BaseModel):
     """The speech handle that was created"""
 
 
+class SessionCloseEvent(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    type: Literal["session_close"] = "session_close"
+    error: Error
+    component: LLM | STT | TTS = Field(..., exclude=True)
```
theomonnom, line 126: This should also be called on normal close. 
error should be optional, for component maybe it can be moved into a new SessionError class? Where the source field is the component?

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -57,6 +59,8 @@ def userdata(self) -> Userdata_T:
     "function_tools_executed",
     "metrics_collected",
     "speech_created",
+    "agent_error",
```
theomonnom, line 62:
```suggestion
```

livekit-agents/livekit/agents/voice/events.py
Shubhrakanti: Yeah that makes sense. Will update.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -563,6 +563,14 @@ def _on_metrics_collected(self, ev: AgentMetrics) -> None:
         if speech_handle := _SpeechHandleContextVar.get(None):
             ev.speech_id = speech_handle.id
         self._session.emit("metrics_collected", MetricsCollectedEvent(metrics=ev))
+        if ev.error:
+            if not ev.error.retryable or ev.error.attempts_remaining == 0:
+                if isinstance(ev, LLMMetrics):
+                    self._session._create_session_close_task(self.llm, ev.error)
+                elif isinstance(ev, STTMetrics):
```
Shubhrakanti, line 570: @theomonnom this is what I mean by  `.on("metrics_collected", change_application_flow())`

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -58,7 +59,7 @@ class ChatChunk(BaseModel):
 
 class LLM(
     ABC,
-    rtc.EventEmitter[Union[Literal["metrics_collected"], TEvent]],
+    rtc.EventEmitter[Union[Literal["metrics_collected"], Literal["error"], TEvent]],
     Generic[TEvent],
```
Shubhrakanti, line 63: Why is this a generic

livekit-agents/livekit/agents/metrics/base.py
```diff
@@ -5,51 +5,43 @@
...
-    characters_count: int
-    streamed: bool
+    label: str
+    request_id: str = ""
+    ttfb: float = 0.0
+    duration: float = 0.0
+    audio_duration: float = 0.0
+    cancelled: bool = False
+    characters_count: int = 0
+    streamed: bool = False
```
Shubhrakanti, line 43: should remove defaults now that we don't need them

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -564,6 +569,17 @@ def _on_metrics_collected(self, ev: AgentMetrics) -> None:
...
 
+    def _on_error(self, ev: Error) -> None:
+        if isinstance(ev, LLMError):
+            self._session.emit("error", ErrorEvent(error=ev, component=self.llm))
+        elif isinstance(ev, STTError):
+            self._session.emit("error", ErrorEvent(error=ev, component=self.stt))
+        elif isinstance(ev, TTSError):
+            self._session.emit("error", ErrorEvent(error=ev, component=self.tts))
+
+        if not ev.retryable or ev.attempts_remaining == 0:
```
Shubhrakanti, line 580: should the error just have like a `error.fatal` field or something?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -363,6 +366,17 @@ def update_agent(self, agent: Agent) -> None:
                 self._update_activity_task(self._agent), name="_update_activity_task"
             )
 
+    async def drain(self) -> None:
+        if self._activity is not None:
+            await self._activity.drain()
```
Shubhrakanti, line 371: When do we call this?

examples/errorcall_back_agent.py
```diff
@@ -0,0 +1,81 @@
...
+
+load_dotenv()
+
+
+"""
+Simulate an LLM provider being down.
+"""
+
+
+class ErrorLLM(LLM):
```
davidzhao, line 24: can we actually wrap an actual LLM, and let the first request go through? so that it's more realistic, versus dying immediately

examples/errorcall_back_agent.py
```diff
@@ -0,0 +1,81 @@
...
+        vad=silero.VAD.load(),
+    )
+
+    @session.on("error")
+    def on_error(ev: ErrorEvent):
+        logger.info(f"Session is closing due to error in {ev.source.__class__.__name__}")
+
+    @session.on("session_close")
+    def on_session_close(ev: SessionCloseEvent):
+        logger.info("Play custom audio before session close")
```
davidzhao, line 75: actually use session.say() to play a message?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -363,6 +367,17 @@ def update_agent(self, agent: Agent) -> None:
...
+
+    def _create_session_close_task(self, cause: ErrorEvent | None = None) -> None:
+        self.emit(
+            "session_close",
+            SessionCloseEvent(cause=cause),
+        )
+        self._closing_task = asyncio.create_task(self.aclose())
```
davidzhao, line 379: this doesn't actually close session for both sides.. it only closes it for the participant. the other side (phone caller or app user) would still be in the room by themselves. To end the session for everyone, we'd have to call `ctx.api.room.delete_room(...)`. I don't think we should do that automatically though.. because there are a lot of use cases that would warrant the user being able to continue without kicking everyone out.

so it raises the question if this should be an event or an async callback. There are a few scenarios where the event based approach doesn't work:
1. I'd like to make an HTTP call to grab the right end message. that would break in this case because while we are making a call to retrieve, closing_task would have ran.
2. I'd like to delete the room (disconnecting the participant) after playing a message: that wouldn't work either, because we don't know when it's fully played back.

It seems much simpler to have a async callback as an error handler.. and the default behavior is to close the agent

Shubhrakanti: I agree this should be an async call back.

examples/errorcall_back_agent.py
Shubhrakanti: Updated.

examples/errorcall_back_agent.py
Shubhrakanti: My commit for this got lost for some reason. Updated!

examples/error_handling/errorcall_back_agent.py
```diff
@@ -0,0 +1,110 @@
...
+            # If you define a custom audio file, it will play out even if the TTS provider is down.
+            audio=audio_frames_from_file(error_wav_path),
+        )
+
+    @session.on("session_close")
+    def on_session_close(_: SessionCloseEvent):
+        global room_close_task
+        # Delete the room and disconnect all participants
+        room_close_task = asyncio.create_task(ctx.delete_room())
+        room_close_task.add_done_callback(lambda _: logger.info("Room closed!"))
```
Shubhrakanti, line 101: @theomonnom should this be default behavior to close?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -273,17 +276,26 @@ async def start(
             self._started = True
             self._update_agent_state(AgentState.LISTENING)
 
-    async def aclose(self) -> None:
+    async def aclose(self, cause: ErrorEvent | None = None) -> None:
         async with self._lock:
             if not self._started:
                 return
 
+            await self._aclose_current_activity()
```
Shubhrakanti, line 284: Is there ever a case where we want to close the session without calling `drain()` on the current activity?

Shubhrakanti (commented):
Updated based on slack discussion

examples/error_handling/error_message.wav
theomonnom, line 1: Can this file be converted to ogg/opus. 4MB is too big for this repo

examples/error_handling/errorcall_back_agent.py
theomonnom, line 1: I think we should show how to use the new callbacks, but let's not add this ErrorLLM, this example is for users and they assume it works. ErrorLLM would be used for testing for e.g

livekit-agents/livekit/agents/errors/__init__.py
theomonnom, line 1: I don't think we want a new error module, can you add move all error to their respective LLM/STT/TTS class?

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -58,7 +59,7 @@ class ChatChunk(BaseModel):
 
 class LLM(
     ABC,
-    rtc.EventEmitter[Union[Literal["metrics_collected"], TEvent]],
+    rtc.EventEmitter[Union[Literal["metrics_collected"], Literal["error"], TEvent]],
```
theomonnom, line 62: I think this should work
```suggestion
    rtc.EventEmitter[Union[Literal["metrics_collected", "error"], TEvent]],
```

livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +291,10 @@ async def connect(
 
         _apply_auto_subscribe_opts(self._room, auto_subscribe)
 
+    async def delete_room(self) -> None:
```
theomonnom, line 294: Are you going to do this in a separate PR?

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -143,6 +149,20 @@ async def _main_task(self) -> None:
                     )
 
                 await asyncio.sleep(self._conn_options.retry_interval)
+                # Reset the flag when retrying
+                self._current_attempt_has_error = False
+
+    def _emit_error(self, api_error: APIError, recoverable: bool):
```
theomonnom, line 155: it would be great to get the original BaseException

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -317,19 +319,22 @@ async def start(self) -> None:
...
+                self.stt.on("error", self._on_error)
 
             if isinstance(self.tts, tts.TTS):
                 self.tts.on("metrics_collected", self._on_metrics_collected)
+                self.tts.on("error", self._on_error)
 
             if isinstance(self.vad, vad.VAD):
                 self.vad.on("metrics_collected", self._on_metrics_collected)
-
+                self.vad.on("error", self._on_error)
```
theomonnom, line 337:
```suggestion
```
I don't think we have `error` event for vad

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -273,17 +276,26 @@ async def start(
             self._started = True
             self._update_agent_state(AgentState.LISTENING)
 
-    async def aclose(self) -> None:
+    async def aclose(self, cause: ErrorEvent | None = None) -> None:
```
theomonnom, line 279: Since this is a user facing API, I doin't think it should have any parameter, let's create an internal aclose_impl instead?

livekit-agents/livekit/agents/voice/agent_session.py
theomonnom: Yes! aclose should never call drain, calls to aclose should terminates instantly, without waiting for any speech

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -387,19 +399,25 @@ def update_agent(self, agent: Agent) -> None:
...
-            if self._activity is not None:
-                await self._activity.drain()
-                await self._activity.aclose()
+            await self._aclose_current_activity()
 
             self._activity = self._next_activity
             self._next_activity = None
             await self._activity.start()
 
+    async def _aclose_current_activity(self) -> None:
```
theomonnom, line 416: I don't think this function is really useful, just reuse the code

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -115,6 +120,18 @@ class SpeechCreatedEvent(BaseModel):
     """The speech handle that was created"""
 
 
+class ErrorEvent(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    type: Literal["error"] = "error"
+    error: Error
+    source: LLM | STT | TTS
```
theomonnom, line 127: source could also be `| Any` to be future proof

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -115,6 +120,18 @@ class SpeechCreatedEvent(BaseModel):
     """The speech handle that was created"""
 
 
+class ErrorEvent(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    type: Literal["error"] = "error"
+    error: Error
```
theomonnom, line 126: `error` would then be stt.Error, tts.Error, llm.Error?

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -115,6 +120,18 @@ class SpeechCreatedEvent(BaseModel):
...
+    source: LLM | STT | TTS
+
+
+class SessionCloseEvent(BaseModel):
+    type: Literal["session_close"] = "session_close"
```
theomonnom, line 131: Should we name our event `close` or `session_close`?

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -115,6 +120,18 @@ class SpeechCreatedEvent(BaseModel):
...
+    cause: ErrorEvent | None = None
```
theomonnom, line 132: let's not reuse the ErrorEvent here, it's ok to copy the fields

theomonnom (changes requested):

examples/error_handling/errorcall_back_agent.py
theomonnom: Once we have your next PR, it'll just be a single call to ctx.delete_room(), we will not even need a task

livekit-agents/livekit/agents/job.py
Shubhrakanti: Yeah.

livekit-agents/livekit/agents/voice/events.py
Shubhrakanti: `close` since we call it as `session.on("close")`

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -273,17 +277,42 @@ async def start(
...
+        await self._activity.drain()
+
+    async def _aclose_impl(self, cause: ErrorEvent | None = None) -> None:
         async with self._lock:
             if not self._started:
                 return
 
+            if self._draining_task is not None:
+                await self._draining_task
+
```
Shubhrakanti, line 299: We have to do this since we can't do 

```python
@session.on("error")
    async def on_error(ev: ErrorEvent):
        session.say(...)
        await session.drain()
```

Otherwise we get this warning 

```
025-04-07 19:17:54,406 - WARNING livekit.agents - _SegmentSynchronizerImpl.push_audio called after close {"pid": 67184, "job_id": "AJ_UzafNMZhdXtv"}
2025-04-07 19:17:54,406 - WARNING livekit.agents - _SegmentSynchronizerImpl.end_audio_input called after close {"pid": 67184, "job_id": "AJ_UzafNMZhdXtv"}
2025-04-07 19:18:00,130 - WARNING livekit.agents - _SegmentSynchronizerImpl.playback_finished called after close {"pid": 67184, "job_id": "AJ_UzafNMZhdXtv"}
```

Shubhrakanti: This happens because we need to make sure we call close _after_ [we acquire the lock on self._activity](https://github.com/livekit/agents/blob/f7869bc9da13add8e41d00b44022abd53dda2bf9/livekit-agents/livekit/agents/voice/agent_activity.py#L358) to drain it

livekit-agents/livekit/agents/voice/agent_activity.py
Shubhrakanti: Ahh cursor going rogue

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -53,12 +53,21 @@ class ChatChunk(BaseModel):
     usage: CompletionUsage | None = None
 
 
+class LLMError(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    type: Literal["llm_error"] = "llm_error"
+    timestamp: float
+    label: str
+    error: APIError = Field(..., exclude=True)
+    recoverable: bool
```
Shubhrakanti, line 62: I feel like LLMError, TTSError and STTError should inherit from a base error class?

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -115,6 +119,18 @@ class SpeechCreatedEvent(BaseModel):
...
+class ErrorEvent(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    type: Literal["error"] = "error"
+    error: LLMError | STTError | TTSError | Any
+    source: LLM | STT | TTS
+
+
+class CloseEvent(BaseModel):
+    type: Literal["close"] = "close"
+    cause: LLMError | STTError | TTSError | None = None
```
theomonnom, line 131:
```suggestion
    error: LLMError | STTError | TTSError | None = None
```

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -115,6 +119,18 @@ class SpeechCreatedEvent(BaseModel):
     """The speech handle that was created"""
 
 
+class ErrorEvent(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    type: Literal["error"] = "error"
+    error: LLMError | STTError | TTSError | Any
+    source: LLM | STT | TTS
```
theomonnom, line 126:
```suggestion
    source: LLM | STT | TTS | Any
```

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -387,6 +421,13 @@ def update_agent(self, agent: Agent) -> None:
                 self._update_activity_task(self._agent), name="_update_activity_task"
             )
 
+    def _create_close_task(
+        self,
+        *,
+        cause: llm.LLMError | stt.STTError | tts.TTSError | None = None,
+    ) -> None:
+        self._closing_task = asyncio.create_task(self._aclose_impl(cause=cause))
```
theomonnom, line 429: This doesn't protect against a closing_task already existing

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -273,17 +276,48 @@ async def start(
...
+    async def _drain_impl(self) -> None:
+        if self._activity is None:
+            raise RuntimeError("AgentSession isn't running")
+
+        await self._activity.drain()
+
+    async def _aclose_impl(
+        self,
+        *,
+        cause: llm.LLMError | stt.STTError | tts.TTSError | None = None,
```
theomonnom, line 296:
```suggestion
        error: llm.LLMError | stt.STTError | tts.TTSError | None = None,
```

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -273,17 +276,48 @@ async def start(
             self._started = True
             self._update_agent_state(AgentState.LISTENING)
 
-    async def aclose(self) -> None:
+    def drain(self) -> None:
```
theomonnom, line 279: drain should be an async method, we should not create any task

livekit-agents/livekit/agents/voice/agent_session.py
theomonnom: The users don't need to call drain on error, we will do it automatically before closing the session if it was an unrecoverable error

theomonnom (changes requested):

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -595,6 +599,20 @@ def _on_metrics_collected(self, ev: AgentMetrics) -> None:
...
+            self._session.emit("error", error_event)
+        elif isinstance(error, stt.STTError):
+            error_event = ErrorEvent(error=error, source=self.stt)
+            self._session.emit("error", error_event)
+        elif isinstance(error, tts.TTSError):
+            error_event = ErrorEvent(error=error, source=self.tts)
+            self._session.emit("error", error_event)
+
+        if not error.recoverable:
+            self._session._create_close_task(cause=error)
```
Shubhrakanti, line 614: @theomonnom

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -401,6 +420,27 @@ def update_agent(self, agent: Agent) -> None:
...
+
+        async def drain_and_close() -> None:
+            logger.info("Closing task started")
+            await self.drain()
+            if self._activity is not None:
+                await self._activity.aclose()
+            await self._aclose_impl(error=error)
+
+        def on_close_done(_: asyncio.Task) -> None:
+            logger.info("Closing task done")
```
theomonnom, line 438:
```suggestion
```

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -401,6 +420,27 @@ def update_agent(self, agent: Agent) -> None:
...
 
+    def _on_error(
+        self,
+        error: llm.LLMError | stt.STTError | tts.TTSError,
+    ) -> None:
+        if self._closing_task or error.recoverable:
+            return
+
+        async def drain_and_close() -> None:
+            logger.info("Closing task started")
```
theomonnom, line 431:
```suggestion
```

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -273,17 +275,34 @@ async def start(
...
 
+            self.emit("close", CloseEvent(error=error))
+
             if self._forward_audio_atask is not None:
                 await utils.aio.cancel_and_wait(self._forward_audio_atask)
 
             if self._room_io:
                 await self._room_io.aclose()
 
+        logger.debug("AgentSession closed")
```
theomonnom, line 301:
```suggestion
```

theomonnom (approved):
Otherwise lgtm! 🎉

examples/error_handling/errorcall_back_agent.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Create session
+    session = AgentSession(
+        stt=deepgram.STT(),
+        llm=openai.LLM(),
+        tts=cartesia.TTS(),
+        vad=silero.VAD.load(),
+    )
+
+    custor_error_audio = os.path.join(pathlib.Path(__file__).parent.absolute(), "error_message.ogg")
```
theomonnom, line 37:
```suggestion
    custom_error_audio = os.path.join(pathlib.Path(__file__).parent.absolute(), "error_message.ogg")
```

examples/error_handling/errorcall_back_agent.py
theomonnom, line 1: can you rename it to error_callback.py

PR #1751, opened 2025-03-27
add agent session history and function_tools_executed event

longcw:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -995,20 +988,30 @@ def _on_first_frame(_: asyncio.Future) -> None:
...
+                # add the function call and output to the event, including the None outputs
+                fnc_executed_ev.function_calls.append(sanitized_out.fnc_call)
+                fnc_executed_ev.function_call_outputs.append(sanitized_out.fnc_call_out)
+
                 if new_agent_task is not None and sanitized_out.agent_task is not None:
                     logger.error(
                         "expected to receive only one AgentTask from the tool executions",
                     )
                     ignore_task_switch = True
+                    # TODO(long): should we mark the function call as failed to notify the LLM?
```
theomonnom, line 1011: Nope it's OK, it's a big failure from the user, we can't really recover

theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: I got this error sometimes the LLM call the same function twice (tho not sure why), but when this happens the transition failed while the LLM still responses "your are transferred to xxx". So I am wondering if it's better to notify the LLM there is an error the transition failed.

theomonnom: I see, it's very unclear why it calls it two times, is the description of the tool clear? Is it with the RealtimeLLM?

PR #1752, opened 2025-03-27
[agent 1.0] Fix ruff 

jayeshp19:


davidzhao:
fixed in my earlier PR.

PR #1762, opened 2025-03-27
support dynamic tool creation

theomonnom:
ideally pydantic adds support for "`choices`" in their `Field` dataclass for better devex

otherwise any BaseModel/common python types work


davidzhao (approved):
lgtm

PR #1767, opened 2025-03-27
fix: publish transcription only when still connected

davidzhao:


theomonnom (approved):

livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -200,7 +200,8 @@ async def _publish_transcription(self, id: str, text: str, final: bool) -> None:
             ],
         )
         try:
-            await self._room.local_participant.publish_transcription(transcription)
+            if self._room.isconnected():
+                await self._room.local_participant.publish_transcription(transcription)
         except Exception as e:
             logger.warning("failed to publish transcription", exc_info=e)
```
theomonnom, line 206: fwiw the warning is still possible when not connected, but it'll definitely reduce it

PR #1769, opened 2025-03-27
chore(docs): update publish-docs workflow to use uv

rektdeckard:
Pydocs cannot currently be deployed in `dev-1.0` or any branch after migrating to `uv`. This attempts to fix up the deploy workflow.

https://www.loom.com/share/d08dc82ad5af498ca63e8914a9ae8c5b

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -21,30 +21,35 @@ jobs:
         with:
           python-version: "3.12"
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
+
       - name: Create and activate virtual environment
         run: |
-          python -m venv venv
-          source venv/bin/activate
+          uv venv 
+          source .venv/bin/activate
+
+      - name: Install package and all plugins
+        run: |
+          source .venv/bin/activate
+          uv sync --active
 
       - name: Install pdoc and other dependencies
         run: |
-          source venv/bin/activate
-          python -m pip install pdoc3 setuptools
-
-      - name: Install package
-        run: |
-          source venv/bin/activate
-          pip install ./livekit-agents
-          ./livekit-plugins/install_local.sh
+          source .venv/bin/activate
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -21,26 +21,31 @@ jobs:
...
         run: |
           source venv/bin/activate
-          python -m pip install pdoc3 setuptools
+          uv install pdoc3 setuptools
 
       - name: Install package
         run: |
           source venv/bin/activate
-          pip install ./livekit-agents
-          ./livekit-plugins/install_local.sh
```
rektdeckard, line 38: NOTE: I can't find this script in the git history, unclear what it was for?

davidzhao: this has been removed from v1.. 

actually for this docs workflow.. we'd need to think about v1 vs pre-v1.. do we want to keep separate reference docs?

.github/workflows/publish-docs.yml
```diff
@@ -21,26 +21,31 @@ jobs:
...
+          uv pip install ./livekit-agents
```
davidzhao, line 43: i think we might want `uv sync`, because we'd want all the plugins to be installed too in order to generate their docs

.github/workflows/publish-docs.yml
rektdeckard: There's probably something clever we can do using the branch name itself in the deploy script to upload to two different S3 buckets. Since we only run this manually, we could say `dev-1.0` (eventually promoted to `main`) pushes to the current bucket, and `0.X` or whatever `main` is demoted to is deployed to a second bucket?

.github/workflows/publish-docs.yml
rektdeckard: Right :)

.github/workflows/publish-docs.yml
davidzhao: maybe it's fine.. let's move forward then.

davidzhao (approved):

theomonnom (approved):

PR #1770, opened 2025-03-28
Support OpenAI Realtime audio transcription parameters

anishnag:
Expose `language` and `prompt` parameters for `_InputAudioTranscription`.


longcw:
I think we are going to replace the `_InputAudioTranscription` and `_TurnDetection` with the types from OpenAI, so it will be more consistent.

theomonnom:
I see that there is also a TypedDict version of InputAudioTranscription, I think we should use it instead, for better devex

theomonnom:
https://github.com/openai/openai-python/blob/a8fa0def5cd999044dae39b6cdff7a54db25c627/src/openai/types/beta/realtime/session_update_event_param.py#L27

Inside src/openai/types/beta/realtime/session_update_event_param.py

anishnag:
@theomonnom `TypedDict` seems to not support dot notation, which seems inconsistent with a bunch of other code in the `realtime_model.py` file.

theomonnom:
I see, yeah I guess it lgtm! Thanks a lot!

PR #1778, opened 2025-03-28
cleanup AudioStreamDecoder resources

theomonnom:
- not sure if this was the cause of some mem leak reported in the last 0.x version


davidzhao (approved):
THANK YOU!

PR #1779, opened 2025-03-28
merge 0.x changes to main

jayeshp19:


theomonnom (approved):
if this is multiple PRs, cherry picking would have been easier to follow, otherwise lgtm

davidzhao (commented):
thank you for doing this. could you include more specific change set than "merge from 0.x"? it wouldn't be clear to those that are following.

also agree individual PRs would've been better

jayeshp19:
updated here - https://github.com/livekit/agents/pull/1784

PR #1786, opened 2025-03-28
fixed order of content and tool calls when building gemini chat ctx

jmugicagonz:
The order of the messages with both content and function calls is wrong. Content should go first. Otherwise, the model gets confused.

@yepher


jmugicagonz:
there are code errors in this PR. I'll open a new one

PR #1795, opened 2025-03-28
fix eou according to v1.0 and ignore fnc calls and outputs in message history

### LINKED ISSUE #1793: AttributeError: 'FunctionCall' object has no attribute 'role' [V1.03]
When: Occurs within the livekit-agents framework, specifically in the livekit.plugins.turn_detector.eou.predict_end_of_turn function called via livekit.agents.voice.audio_recognition.py's _bounce_eou_task. This happens after a user speaks, seemingly following an interaction involving a **tool call**

Indication: The internal chat context (chat_ctx) being processed by the End-of-Utterance turn detector contains a raw FunctionCall object where a standard message object (with a .role) is expected. The turn detector plugin cannot handle this object type in the message history.

jayeshp19:
fixes: https://github.com/livekit/agents/issues/1793

### DIFF
--- a/livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/eou.py
+++ b/livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/eou.py
@@ -141,26 +141,30 @@ class EOUModel:
         messages = []
 
         for msg in chat_ctx.items:
-            if msg.role not in ("user", "assistant"):
-                continue
-
-            if isinstance(msg.content, str):
-                messages.append(
-                    {
-                        "role": msg.role,
-                        "content": msg.content,
-                    }
-                )
-            elif isinstance(msg.content, list):
-                for cnt in msg.content:
-                    if isinstance(cnt, str):
+            # ignore function calls and function call outputs
+            if msg.type == "message":
+                for content in msg.content:
+                    if isinstance(content, str):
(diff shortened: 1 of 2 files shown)


davidzhao (approved):

theomonnom:
U beat me at this lol

livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/eou.py
```diff
@@ -141,26 +141,30 @@ async def predict_end_of_turn(
...
+                            }
+                        )
+                    elif isinstance(content, dict):
+                        messages.append(
+                            {
+                                "role": msg.role,
+                                "content": json.dumps(content),
+                            }
+                        )
+                    elif isinstance(content, llm.ImageContent):
```
davidzhao, line 161: we should exclude images and dict too actually

jayeshp19:
fixed in https://github.com/livekit/agents/pull/1796

livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/eou.py
jayeshp19: Yes it was by mistake, anyways theo has another pr so I closed this one

PR #1768, opened 2025-03-27
Support `temperature` configuration for OpenAI Realtime model

anishnag:
Expose the `temperature` parameter for OpenAI Realtime model.

PR #1802, opened 2025-03-29
Add default temperature for OpenAI Realtime model

anishnag:
An additional step to avoid error from in supporting temperature for Realtime model (extension of #1768). We default to `0.8` temperature.


theomonnom:
It seems like we now have the same issue with the turn_detection/input_audio_transcription, since there is no default, they get disabled

theomonnom:
done here https://github.com/livekit/agents/pull/1809

PR #1803, opened 2025-03-29
PlayAI plugin: bump Python SDK version (add Groq support via PlayDialog-turbo voice engine)

bryananderson:


theomonnom (approved):
Thanks!

PR #1809, opened 2025-03-29
fix OpenAI realtime LLM types 

theomonnom:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -356,39 +372,48 @@ async def _recv_task() -> None:
...
-        session = session_update_event.Session.model_construct(
-            model=self._realtime_model._opts.model,
-            voice=self._realtime_model._opts.voice,
-            input_audio_format="pcm16",
-            output_audio_format="pcm16",
-            modalities=["text", "audio"],
-            turn_detection=turn_detection,
-            input_audio_transcription=input_audio_transcription,
+            if turn_detection
+            else None
```
theomonnom, line 398: Their server has different structs depending on semantic_vad/server_vad, so you get an error even if the "None" is there. I'm really wondering how they're going to fix their mistake without breaking everyone code.

We exclude the defaults here and not inside the update_session to still allow to disable turn_detection by using None

PR #1810, opened 2025-03-29
better idle capacity management

theomonnom:


livekit-agents/livekit/agents/ipc/proc_pool.py
```diff
@@ -20,7 +21,7 @@
     "process_job_launched",
 ]
 
-MAX_CONCURRENT_INITIALIZATIONS = 5
+MAX_CONCURRENT_INITIALIZATIONS = os.cpu_count() or 1
```
davidzhao, line 24: this isn't container aware, we need to use our capacity util here: https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/utils/hw/cpu.py

livekit-agents/livekit/agents/worker.py
```diff
@@ -166,7 +166,7 @@ class WorkerOptions:
 
     """Number of idle processes to keep warm."""
     num_idle_processes: int | _WorkerEnvOption[int] = _WorkerEnvOption(
-        dev_default=0, prod_default=3
+        dev_default=0, prod_default=(os.cpu_count() or 1)
```
davidzhao, line 169: same here

livekit-agents/livekit/agents/worker.py
```diff
@@ -373,6 +394,27 @@ def load_fnc():
...
                 self._worker_load = await asyncio.get_event_loop().run_in_executor(None, load_fnc)
+
+                load_threshold = _WorkerEnvOption.getvalue(self._opts.load_threshold, self._devmode)
+                default_num_idle_processes = _WorkerEnvOption.getvalue(
+                    self._opts.num_idle_processes, self._devmode
+                )
+
+                if not math.isinf(load_threshold):
+                    active_jobs = len(self.active_jobs)
+                    if active_jobs > 0:
```
davidzhao, line 405: should we raise the limit? when it only has one job, the avg is likely skewed

davidzhao (approved):

PR #1816, opened 2025-03-30
add: ChatMessage.text_content() to retrieve text from message

### LINKED ISSUE #1792: Difficulty Capturing Agent Transcript via Events [V1.03]
Goal:
To capture the REALTIME agent's spoken transcript in real-time or upon completion.

Problem:
There doesn't appear to be a direct, simple event analogous to user_input_transcribed for capturing the agent's final spoken output. In 0.x we used

        # self.session.on("agent_speech_committed", self.on_agent_speech_committed)
        # self.session.on("agent_speech_interrupted", self.on_agent_speech_interrupted)
        # self.session.on("user_speech_committed", self.on_user_speech_committed)

Which worked well for RealTime agents. But they are removed in v1

Attempts & Findings:

Using conversation_item_added Event:

The most intuitive event-based approach seems to be listening to conversation_item_added and filtering for messages where message.role == "assistant".
Issue: Directly accessing message.content.text within this event handler often fails with an AttributeError: 'list' object has no attribute 'text'. This is because message.content can be a list (e.g., [{'type': 'text', 'text': '...'}]) rather than a simple object or string, requiring developers to implement parsing logic to extract the text.
Using Custom TextOutput Sink:

Attaching a custom TextOutput class to self.session.output.transcription allows capturing the transcript stream directly as it's generated.
Drawback: This requires defining a custom class, which is less straightforward than using a simple event listener.
Suggestion:
It would significantly improve the developer experience if the SDK provided a dedicated event for agent speech output, similar to user_input_transcribed. This could be:

An event like agent_output_transcribed that fires with interim and final transcript segments.
Or, ensure that the message.content within the conversation_item_added event consistently provides a simple way to access the final text transcript without needing complex type checking and iteration.
This would simplify the common task of logging or displaying agent responses alongside user input

davidzhao:
closes #1792

### DIFF
--- a/livekit-agents/livekit/agents/llm/chat_context.py
+++ b/livekit-agents/livekit/agents/llm/chat_context.py
@@ -110,6 +110,17 @@ class ChatMessage(BaseModel):
     interrupted: bool = False
     hash: bytes | None = None
 
+    def text_content(self) -> str | None:
+        """
+        Returns a string of all text content in the message.
+
+        Multiple text content items will be joined by a newline.
+        """
+        text_parts = [c for c in self.content if isinstance(c, str)]
+        if not text_parts:
+            return None
+        return "\n".join(text_parts)
+
 
 ChatContent: TypeAlias = Union[ImageContent, AudioContent, str]
 
--- a/livekit-agents/livekit/agents/llm/chat_context.py
+++ b/livekit-agents/livekit/agents/llm/chat_context.py
@@ -110,6 +110,7 @@ class ChatMessage(BaseModel):
     interrupted: bool = False
     hash: bytes | None = None
 
+    @property
     def text_content(self) -> str | None:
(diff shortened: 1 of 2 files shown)


theomonnom (approved):

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -110,6 +110,17 @@ class ChatMessage(BaseModel):
     interrupted: bool = False
     hash: bytes | None = None
 
+    def text_content(self) -> str | None:
```
longcw, line 113: should this be a property?

theomonnom: I generally don't like property that does more than just returning a value. But no strong opinion

PR #1822, opened 2025-03-31
fix update_instructions on copied chat ctx

longcw:
`update_instructions` on copied chat ctx still modifies the original `chat_ctx` bc the it's a shadow copy.


davidzhao (approved):
lgtm

PR #1827, opened 2025-03-31
Use NOT_GIVEN in STT Fallback Adapter

ChenghaoMou:
Currently, using Deepgram STT (expecting str or NOT_GIVEN) with a FallbackAdapter (str | None = None) would make the `_opts.language` `None` during streaming. This changes it to `NOT_GIVEN` to be consistent with STT base implementation.


jayeshp19 (approved):
LGTM

jayeshp19:
Thanks. Can you please also add it for stream adapter?

theomonnom:
actually the issue is Deepgram code
It should be NotGivenOr[str | None]

ChenghaoMou:
> actually the issue is Deepgram code [this](https://github.com/livekit/agents/blob/9aef151420b413658b46ccaa3fb11db34621452e/livekit-agents/livekit/agents/stt/stt.py#L148) should be NotGivenOr[str | None]

If I understand it correctly, allowing None is permitted for language detection, right? In this case, should I still use NOT_GIVEN as default?

theomonnom:
> If I understand it correctly, allowing None is permitted for language detection, right? In this case, should I still use NOT_GIVEN as default?

Yes, that's correct, the default is still NOT_GIVEN

livekit-agents/livekit/agents/stt/fallback_adapter.py
```diff
@@ -215,15 +215,15 @@ async def recognize(
         self,
         buffer: AudioBuffer,
         *,
-        language: str | None = None,
```
theomonnom, line 218: let's do the same here. NotGivenOr[str | None].
Otherwise lgtm!

theomonnom (approved):

PR #1833, opened 2025-04-01
chore(resemble): clean up types and removed unused function

davidzhao:


theomonnom (approved):
nit, but quite a lot of inconsistencies with other plugins (e.g using websockets vs aiohttp)

livekit-plugins/livekit-plugins-resemble/livekit/plugins/resemble/tts.py
```diff
@@ -18,147 +18,176 @@
...
 class TTS(tts.TTS):
     def __init__(
         self,
         *,
         api_key: str | None = None,
-        voice_uuid: str | None = DEFAULT_VOICE_UUID,
+        voice_uuid: str | None = None,
+        tokenizer: tokenize.SentenceTokenizer | None = None,
         sample_rate: int = 44100,
+        precision: str | None = None,
```
theomonnom, line 62: There is no point at exposing precision?

theomonnom (approved):

PR #1839, opened 2025-04-01
fix .gitattributes filename

longcw:


toddsor (approved):
Renaming is what is needed.

theomonnom (approved):

PR #1841, opened 2025-04-01
remove instructions and reset tool_choice for realtime model

longcw:
remove instructions for update_chat_ctx and add update_options for realtime model


examples/full_examples/restaurant_agent/restaurant_agent.py
```diff
@@ -127,7 +127,9 @@ async def on_enter(self) -> None:
             content=f"You are {agent_name} agent. Current user data is {userdata.summarize()}",
         )
         await self.update_chat_ctx(chat_ctx)
-        self.session.generate_reply(tool_choice="none")
+        speech_handle = self.session.generate_reply(tool_choice="none")
+        # reset tool choice for realtime model
+        speech_handle.add_done_callback(lambda _: self.session.update_options(tool_choice="auto"))
```
theomonnom, line 132: This shouldn't be done by the user, after the generate_reply is started, ideally our framework is automatically putting back the old value.

`speech_handle = self.session.generate_reply(tool_choice="none")` is temporary for this message


Only update_options is definitive

longcw: yeah, it's better to make it automated for `generate_reply`. For regular LLM it's fine as it's stateless, where can we read the old value before update it for `generate_reply`?

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -107,6 +107,10 @@ def chat_ctx(self) -> ChatContext: ...
     @abstractmethod
     def tools(self) -> ToolContext: ...
 
+    @property
+    @abstractmethod
+    def tool_choice(self) -> ToolChoice: ...
```
theomonnom, line 112: i'm not sure we should make it a state? Can this be tracked inside the TaskActivity instead?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -288,8 +288,16 @@ def emit(self, event: EventTypes, ev: AgentEvent) -> None:  # type: ignore
...
+    def update_options(self, *, tool_choice: NotGivenOr[llm.ToolChoice] = NOT_GIVEN) -> None:
+        if self._activity is None:
+            raise RuntimeError("AgentSession isn't running")
+
+        if self._activity.draining:
+            if self._next_activity is None:
+                raise RuntimeError("AgentSession is closing, cannot use update_options()")
+            self._next_activity.update_options(tool_choice=tool_choice)
+        else:
+            self._activity.update_options(tool_choice=tool_choice)
```
theomonnom, line 300: I think this state should only bet set during on Agent, and we shouldn't automatically update it for the whole AgentSession.
So it's OK to delete this method IMO?

longcw: sounds good to me

livekit-agents/livekit/agents/llm/realtime.py
longcw: sg, but then we need to make sure the `update_options` will only be called from the TaskActivity

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -486,7 +498,11 @@ def generate_reply(
                     tools=self._agent.tools,
                     user_input=user_input or None,
                     instructions=instructions or None,
-                    model_settings=ModelSettings(tool_choice=tool_choice),
+                    model_settings=ModelSettings(
+                        tool_choice=tool_choice
+                        if utils.is_given(tool_choice) or self._tool_choice is None
+                        else self._tool_choice
+                    ),
```
longcw, line 505: use the `self._tool_choice` for pipeline reply if set for consistency

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -220,10 +221,21 @@ async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
...
 
         if self._rt_session is not None:
+            remove_instructions(chat_ctx)
             await self._rt_session.update_chat_ctx(chat_ctx)
+        else:
+            update_instructions(
+                chat_ctx, instructions=self._agent.instructions, add_if_missing=True
+            )
+
+    def update_options(self, *, tool_choice: NotGivenOr[llm.ToolChoice | None] = NOT_GIVEN) -> None:
```
theomonnom, line 233: the `| None` isn't needed

theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -459,6 +459,8 @@ def update_options(
             oai_tool_choice = tool_choice
             if isinstance(tool_choice, dict) and tool_choice["type"] == "function":
                 oai_tool_choice = tool_choice["function"]
+            if oai_tool_choice is None:
```
theomonnom, line 462: I don't think we should allow none

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: `None` is for reset the tool_choice to default whatever the LLM has. Maybe replace it with NOT_GIVEN for default

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: The original `update_options` api has the `None` option `tool_choice: NotGivenOr[llm.ToolChoice | None] = NOT_GIVEN,`

I think it's reasonable to have `None` to reset the tool_choice to default. For realtime model the default is `auto`.

longcw:
Merge it for now, we can change the API `| None` later. This pr is following the original API but fixed the issues.

PR #1837, opened 2025-04-01
fix(aws): improve credential handling and client lifecycle (#1719) - Version 0.x

andormarkus:
This PR addresses issue https://github.com/livekit/agents/issues/1719 by improving AWS credential handling and client lifecycle management across the AWS plugin.

Key changes:
- Store AWS credentials in instance instead of global env vars to prevent credential leakage
- Use async context managers for AWS clients to ensure proper resource cleanup
- Add proper credential validation to fail fast if credentials are invalid
- Fix session token handling for temporary credentials
- Ensure proper resource cleanup across all AWS services

The changes make the code more robust, safer, and consistent across all AWS services while properly handling async operations.

PR #1849, opened 2025-04-01
feat: Refactor AWS plugin with pool connection and async session

jayeshp19:
clean up over pr https://github.com/livekit/agents/pull/1837


jayeshp19:
> The implemented `get_aws_credentials` will not work in AWS environment and should be removed. You should simple get a session. Depending the execution environment `session.get_credentials()` will return optionally `aws_session_token` as well.
> 
> You can move the region check into utils to make the code DRY

the purpose of `get_aws_credentials` is to just validate credentials. we're getting  session separately from `get_aws_async_session` 

are you getting any errors, if so can you share it?

jayeshp19:
> You can move the region check into utils to make the code DRY

it won't make any diffrence, we're already passing region in `get_aws_async_session `

andormarkus:
The `get_aws_credentials` implementation is problematic for AWS environments using temporary credentials (EC2 instance profiles, EKS pod identity, or explicit STS credentials) for two major reasons:

1. **Credential freezing**: By calling `session.get_credentials()` at initialization time, you're extracting temporary credentials at a single point in time. These will eventually expire (typically after 1 hour), with no mechanism to refresh them.
2. **Missing session token**: When temporary credentials are used in AWS, they include a session token alongside the access key and secret key. This implementation drops the session token, which will cause authentication failures.

A better approach for credential validation would be to use `get_caller_identity()` from the STS service:
```python
def validate_aws_credentials(api_key=None, api_secret=None, session_token=None, region=None):
    """Validate AWS credentials by attempting to call STS get_caller_identity."""
    try:
        session = boto3.Session(
            aws_access_key_id=api_key,
            aws_secret_access_key=api_secret,
            aws_session_token=session_token,
            region_name=region or DEFAULT_REGION
        )
        sts = session.client('sts')
        identity = sts.get_caller_identity()
        return True, identity
    except Exception as e:
        return False, str(e)
 ```
 
This approach:
 - Validates credentials without extracting and freezing them
 - Returns the identity details which can be useful for debugging
 - Allows boto3's credential provider chain to handle credential refreshing naturally

**Comment by andormarkus** [comment]
While reviewing the code, I noticed that our implementation for AWS credential handling might not work correctly with long-term AWS credentials (direct access/secret key pairs without session tokens).

In the [STT class](https://github.com/livekit/agents/blob/40c788da08589e4f16998c78b079c1d59f3fd0eb/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py#L54)'s `_create_client` method, we're unconditionally passing `frozen_credentials.token` to the `StaticCredentialResolver`:

```python
return TranscribeStreamingClient(
    region=self._config.region,
    credential_resolver=StaticCredentialResolver(
        access_key_id=frozen_credentials.access_key,
        secret_access_key=frozen_credentials.secret_key,
        session_token=frozen_credentials.token,  # This could be None with long-term credentials
    ),
)
```

Long-term AWS credentials don't include a token component, which is only present when using temporary credentials from environment variables, IAM roles, or AWS STS.

**Suggested Fix**
Since I don't have the ability to test with long-term credentials right now, could we make the code more robust by conditionally including the session token only when it exists?

```python
resolver_args = {
    "access_key_id": frozen_credentials.access_key,
    "secret_access_key": frozen_credentials.secret_key,
}
if frozen_credentials.token:
    resolver_args["session_token"] = frozen_credentials.token

return TranscribeStreamingClient(
    region=self._config.region,
    credential_resolver=StaticCredentialResolver(**resolver_args),
)
```

This approach would ensure compatibility with both temporary and long-term credentials without requiring us to test every credential type directly.

andormarkus:
LGTM for Livekit version `1.x`

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
```diff
@@ -98,6 +105,35 @@ def __init__(
             top_p=top_p,
             additional_request_fields=additional_request_fields,
         )
+        self._client_cm = None
+        self._pool = utils.ConnectionPool[aioboto3.Session.client](
```
davidzhao, line 109: we need a `close_cb` as well, to close the client correctly

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
```diff
@@ -98,6 +105,35 @@ def __init__(
...
+            await self._client_cm.__aexit__(None, None, None)
+            self._client_cm = None
+
+        session = self._session or await get_aws_async_session(
+            region=self._region,
+            api_key=self._api_key,
+            api_secret=self._api_secret,
+        )
+        # context manager for the client
+        self._client_cm = session.client("bedrock-runtime")
```
davidzhao, line 128: this pattern isn't good.. storing the session as a global reference.. 

reading into this, it seems totally fine to reuse the same async session for multiple concurrent requests. if that's true, then it'd be much simpler to get rid of all the connection pooling logic here

jayeshp19: > this pattern isn't good.. storing the session as a global reference..

do you mean storing a context manager? we can store a session instead

> reading into this, it seems totally fine to reuse the same async session for multiple concurrent requests. if that's true, then it'd be much simpler to get rid of all the connection pooling logic here

we'd need to reconnect at some point if we're not using aws long-term credentials

davidzhao: I see, so it's primarily for credentials? then I would suggest keeping the session itself in the connection pool, versus the client.

the resource that we are managing should be in the pool, nothing should be stored as an instance variable

jayeshp19: Yeah, that’s the main reason, to refresh credentials. I agree, it makes more sense to just refresh the credentials on the session instead of creating a new session every time in the pool

jayeshp19: I guess we don't even need a pool if we just refresh at certain interval

davidzhao: creating a new session is totally fine IMO.. but the current code is:
* storing a single session in as instance var
* putting the client into the pool
so if a concurrent stream comes in.. you would end up destroying the previous session when it's creating a new connection

andormarkus: If you are using current `aioboto3` or `boto3` you don't need to manually refresh credentials because these SDK are doing for you out of the box. Here implemented refresh by `ConnectionPool`  for `llm.py` and `tts.py` is not necessary if you don't want to pool connection. 

The only exception is the `amazon_transcribe` SDK in `stt.py` . This SDK is an alpha stage for years and not getting updates. It is based on outdated `awscrt` and does not support current STS credentials providers. For this reason you need to "manually" rotate credentials 
See the related issues:
 - https://github.com/awslabs/amazon-transcribe-streaming-sdk/issues/7
 - https://github.com/awslabs/amazon-transcribe-streaming-sdk/issues/49
 - https://github.com/awslabs/amazon-transcribe-streaming-sdk/issues/110

jayeshp19: > If you are using current `aioboto3` or `boto3` you don't need to manually refresh credentials because these SDK are doing for you out of the box. Here implemented refresh by `ConnectionPool` for `llm.py` and `tts.py` is not necessary if you don't want to pool connection.

Oh nice, yeah we don't really need a pool connection for it then

jayeshp19: cool I just confirmed from [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#:~:text=When%20you%20specify%20a%20profile,then%20automatically%20refresh%20the%20credentials.)

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
```diff
@@ -87,6 +94,28 @@ def __init__(
...
+            connect_cb=self._create_client,
+            max_session_duration=refresh_interval
+            if is_given(refresh_interval)
+            else REFRESH_INTERVAL,
+        )
+
+    async def _create_client(self) -> TranscribeStreamingClient:
+        creds = await self._session.get_credentials()
+        frozen_credentials = await creds.get_frozen_credentials()
+        return TranscribeStreamingClient(
```
jayeshp19, line 107: There's no close method for transcribe streaming client

davidzhao: clients don't need to be closed, only the session

livekit-plugins/livekit-plugins-aws/pyproject.toml
```diff
@@ -24,9 +24,9 @@ classifiers = [
 ]
 dependencies = [
     "livekit-agents>=1.0.0.rc5",
-    "aiobotocore==2.19.0",
-    "boto3==1.36.3",
-    "amazon-transcribe>=0.6.2",
+    "aioboto3==14.1.0",
```
theomonnom, line 27: do we need both sync and async versions?

jayeshp19: we're using boto3 for validating credentials

jayeshp19: https://github.com/livekit/agents/pull/1849/files#diff-af9ede2e08c25f595f6021977acd76482740b941c032730e1000232c79d02cfcR31

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
jayeshp19: we don't need to close session, we're just refreshing credentials from session

PR #1857, opened 2025-04-02
remove llama index plugin and update examples

longcw:


longcw:
> I'm wondering if we should delete livekit-plugins-llama-index, but convert our example to directly use their API in our examples

I have no preference. Only the `chat_engine` example is using the llama index plugin, the other two are already using the llama index directly. But for chat_engine, it seems okay to have the llama index as a `LLM` plugin?

theomonnom:
> I have no preference. Only the `chat_engine` example is using the llama index plugin, the other two are already using the llama index directly. But for chat_engine, it seems okay to have the llama index as a `LLM` plugin, otherwise how to specify the `llm` for the agent?

Yes, but the tradeoffs are too big, with llama-index, if you use the ChatEngine, we don't support function calls. So IMO, we should maybe just do an example using llama-index inside the llm_node

longcw:
> Yes, but the tradeoffs are too big, with llama-index, if you use the ChatEngine, we don't support function calls. So IMO, we should maybe just do an example using llama-index inside the llm_node

Sounds good to me.

longcw:
@theomonnom updated all examples to use original llama-index, feel free to remove the pulgin.

examples/voice_agents/llamaindex-rag/chat_engine.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, index: VectorStoreIndex):
+        super().__init__(
+            instructions=(
+                "You are a voice assistant created by LiveKit. Your interface "
+                "with users will be voice. You should use short and concise "
+                "responses, and avoiding usage of unpronouncable punctuation."
+            ),
+            vad=silero.VAD.load(),
+            stt=deepgram.STT(),
+            llm=DummyLLM(),  # use a dummy LLM to enable the pipeline reply
```
longcw, line 50: should we add a mock LLM in the api like `llm=llm.NOT_REQUIRE` to enable the pipeline reply task for ppl want to use a fully customized `llm_node`, for example for multiple LLMs inside the node, or other chat completion libraries like llama index chat engine or langchain.

longcw:
@biouser-abiomix does [this example](https://github.com/livekit/agents/blob/4040d4c92bfb3e3e2e47486f714451c2a85b8e64/examples/voice_agents/llamaindex-rag/chat_engine.py#L40) work for you, without a llamaindex plugin you can still use it in the llm_node of the agent.

theomonnom (approved):
this lg! feel free to remove the llama-index plugin directly inside this PR

Issue #1858, opened 2025-04-02
Agents 1.0 Azure STT language parameter broken


Andhs-eff:
We can't set language in Azure STT like this: 
`azure.STT(language="he-IL")`

In https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py it is blocked by
```
        if not is_given(languages):
            languages = ["en-US"]

        if is_given(language) and not is_given(languages):
            languages = [language]
```

The checks should be swapped over.

longcw:
cc @jayeshp19

jayeshp19:
ah right, that makes sense

praveenpnj94:
Hi All,

We are seeing some delay in generating the initial response while using azure stt? Are you not seeing the same?

praveenpnj94:
Whats your azure stt configuration? Which region and language you are using?

PR #1861, opened 2025-04-02
fix azure language

jayeshp19:
fixes: https://github.com/livekit/agents/issues/1858

### DIFF
--- a/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
+++ b/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
@@ -80,11 +80,11 @@ class STT(stt.STT):
 
         super().__init__(capabilities=stt.STTCapabilities(streaming=True, interim_results=True))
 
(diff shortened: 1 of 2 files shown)


theomonnom:
why do we have two language fields lol?

theomonnom:
let's keep the language field, without s

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
```diff
@@ -42,7 +42,7 @@ class STTOptions:
     segmentation_silence_timeout_ms: NotGivenOr[int]
     segmentation_max_time_ms: NotGivenOr[int]
     segmentation_strategy: NotGivenOr[str]
-    languages: list[
```
theomonnom, line 45: make sure language also accept str

jayeshp19: It does, we're just converting to list before storing into options

PR #1864, opened 2025-04-02
add worker endpoint for hosted agents

paulwe:


livekit-agents/livekit/agents/worker.py
```diff
@@ -305,7 +306,16 @@ def __init__(
         async def health_check(_: Any):
             return web.Response(text="OK")
 
+        async def worker(_: Any):
+            body = json.dumps({
+                "agent_name": self._opts.agent_name,
+                "worker_type": self._opts.worker_type,
```
theomonnom, line 312: That's kind of dumb that our WorkerType is a redefinition of the protobuf JobType object (instead of a str..)

Let's do this instead:
```suggestion
                "worker_type": agent.JobType.Name(self._opts.worker_type.value),
```

theomonnom: This will give JT_ROOM, JT_PUBLISHER

theomonnom (approved):

theomonnom:
Note that this port is defined by the user, it defaults to 8081. So HA may fail to retrieve it when changed?

livekit-agents/livekit/agents/worker.py
paulwe: ngl i didn't test this. i was hoping you would be so offended by my implementation you would rewrite it correctly 😆

paulwe:
> Note that this port is defined by the user, it defaults to 8081. So HA may fail to retrieve it when changed?
we can probably surface this to the user in error logs...

livekit-agents/livekit/agents/worker.py
theomonnom: lol no, I'm offended by my own implementation lol

theomonnom:
yep agree, feel free to merge nbd for now I can add a ticket to detect HA environments

PR #1871, opened 2025-04-03
structured output example

theomonnom:


livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -212,6 +210,54 @@ def build_strict_openai_schema(
     }
 
 
+ResponseFormatT = TypeVar("ResponseFormatT", default=None)
```
longcw, line 213: In python 3.12 this got an error `TypeError: 'default' is an invalid keyword argument for typevar()`

theomonnom: fixed by using typing_extensions

davidzhao (approved):

PR #1875, opened 2025-04-03
fix decoder: if no data was pushed, close the output channel

jayeshp19:


davidzhao (approved):
lgtm!

can we backport it to 0.x as well?

livekit-agents/livekit/agents/utils/codecs/decoder.py
```diff
@@ -118,6 +118,9 @@ def push(self, chunk: bytes):
 
     def end_input(self):
         self._input_buf.end_input()
+        if not self._started:
+            # if no data was pushed, close the output channel
+            self._loop.call_soon_threadsafe(self._output_ch.close)
```
theomonnom, line 123: This isn't needed, you can call close directly

Issue #1670, opened 2025-03-17
FEATURE REQUEST: Add timed transcript (start,end,words in chat history)


akshitdudeja125:
Hi team,
I had a feature request for Python 1.0 release
Is it possible to add start_time, end_time , words(if supported by providers) when the agent's and user's speech is commited?
I had made this change for the voice pipeline agent but It would be great to have some feature for this upstream for that in Voice agent.

Our platform displays transcripts alongside recordings, allowing users to click on a specific message to jump to that exact moment in the recording.
Timestamps in words enable the creation of clips from any selected point in the transcript.
Since these transcripts are used for usability research, we require timestamps at multiple points to serve as reference markers that support key research insights.

[
 
    "role": "assistant",
    "content": "Hey, how can I help you today?",
    "start": 1742134707.016703,
    "end": 1742134708.875511,
    "words": [
      {
        "text": "Hey,",
        "start": 1742134707.016781,
        "end": 1742134707.279758
      },
      {
        "text": "how",
        "start": 1742134707.279854,
        "end": 1742134707.543107
      },
      {
        "text": "can",
        "start": 1742134707.543137,
        "end": 1742134707.806495
      },
      {
        "text": "I",
        "start": 1742134707.806515,
        "end": 1742134708.069907
      },
      {
        "text": "help",
        "start": 1742134708.0699232,
        "end": 1742134708.332927
      },
      {
        "text": "you",
        "start": 1742134708.332947,
        "end": 1742134708.5982761
      },
      {
        "text": "today?",
        "start": 1742134708.598859,
        "end": 1742134708.875474
      }
    ]
  }
]

akshitdudeja125:
Hi @davidzhao @theomonnom @longcw ,
Is it possible to expose them atleast for the providers who support this data in agents v1?

LGR3VE:
This would be extremely useful. Please add it 👍

PR #1882, opened 2025-04-04
Added timestamp to ChatMessage so it gets added to session history

samudranb:
For each message, the current timestamp is added at the point of instantiation


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -109,6 +109,7 @@ class ChatMessage(BaseModel):
     content: list[ChatContent]
     interrupted: bool = False
     hash: bytes | None = None
+    timestamp: str = Field(default_factory=lambda: datetime.now().isoformat())
```
longcw, line 112: should this be a timestamp of `time.time()`?

samudranb:
Personally, I’d prefer it to be human readable. IMHO the performance
considerations do not outweigh the usefulness of having it be human
readable.

livekit-agents/livekit/agents/llm/chat_context.py
samudranb: Personally, I’d prefer it to be human readable. IMHO the performance considerations do not outweigh the usefulness of having it be human readable. I will defer to your judgement.

longcw: two concerns about a datetime string:
1. the timezone is not specified
2. the format of the string can vary, tho here is using isoformat as default but ppl can pass a different formatted string, it's hard for further processing

samudranb: Changed it to use time.time

akshitdudeja125:
Hi @samudranb @longcw 
Can something like https://github.com/livekit/agents/issues/1670 be added in the session history?

theomonnom (approved):
lgtm! thanks! Sorry for the delay

PR #1883, opened 2025-04-04
add audio echo for avatar example

longcw:
read audio from a participant and forward it to the avatar runner


longcw:
> `from livekit.agents.voice.room_io._input import _ParticipantAudioInputStream` is normally internal, do we need it for avatars?

simplified with `rtc.AudioStream.from_participant`

PR #1884, opened 2025-04-04
fix: add speech_endpoint to Azure STT and TTS class init for custom endpoint

Trivedi-grv:
We faced issues when attempting to pass a custom speech endpoint to the STT and TTS classes. Upon investigation, we discovered that the custom endpoint was not being passed to the __init__ method of these classes.

When the changes in PR #1362  were merged, the speech_endpoint property was introduced but was not integrated into the following areas:
- It was not added as a parameter in the __init__ method of the Azure STT and TTS classes.
- It was not included in the configuration (_config variables), which is required to pass it to the Azure SpeechConfig class.

This resulted in speech_endpoint being unavailable for usage in custom speech service configurations.

### DIFF
--- a/.github/next-release/changeset-79b2e16a.md
+++ b/.github/next-release/changeset-79b2e16a.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-azure": patch
+---
+
+fix: add speech_endpoint to Azure STT and TTS class init for custom endpoint (#1884)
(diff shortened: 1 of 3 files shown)


Trivedi-grv:
@longcw can you please review these changes?

longcw (approved):
lgtm!

Issue #1887, opened 2025-04-04
Improvement: deepgram "multi" language support and keyterms check


ChenghaoMou:
1. nova-3/general and nova-2 supports a "multi" language where speakers can do code switch.
2. keyterms is only supported by nova-3, a warning or error should be raised since we downgrade nova-3 to nova-2 automatically with a check. Having keyterms in nova-2 requests will cause some issues in my testing.

Opening this so I don't forget to create PRs. Happy Friday!

davidzhao:
yes! I ran into this yesterday and we have a pending PR for #1 against 0.x. If you could like to tackle #2, that'd be amazing!

schedawg74:
Also nova-3 added multiple new languages besides english. Is this part of the PR as well?

davidzhao:
@schedawg74 that'll be merged soon as well

PR #1898, opened 2025-04-05
Add Deepgram keyterms validataion

ChenghaoMou:
This address the second issue mentioned in #1887 

Reference docs: 
- https://developers.deepgram.com/docs/keywords
- https://developers.deepgram.com/docs/keyterm

TL;DR
keyterms only work with Nova-3 in English;
keywords work with other models;

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -793,20 +793,18 @@ def _validate_keyterms(
     See: https://developers.deepgram.com/docs/keyterm and https://developers.deepgram.com/docs/keywords
     """
     if model.startswith("nova-3") and is_given(keywords):
-        logger.warning(
+        raise ValueError(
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -776,3 +777,36 @@ def _validate_model(
...
+) -> tuple[
+    NotGivenOr[list[str]],
+    NotGivenOr[list[tuple[str, float]]],
+]:
+    """
+    Validating keyterms and keywords for model compatibility.
+    See: https://developers.deepgram.com/docs/keyterm and https://developers.deepgram.com/docs/keywords
+    """
+    if model.startswith("nova-3") and is_given(keywords):
+        logger.warning(
```
davidzhao, line 796: should it throw an error instead?

this seems like a critical error for those that rely on this behavior.

ChenghaoMou: That makes sense. I will change it.

davidzhao (approved):
lg!

PR #1900, opened 2025-04-06
updated a few examples and comments

davidzhao:


examples/voice_agents/realtime_turn_detector.py
```diff
@@ -6,18 +6,27 @@
...
+## In this example, speech is being processed in parallel by both the STT and the realtime API
+
 
 class MyAgent(Agent):
     def __init__(self) -> None:
         super().__init__(
             instructions="You are Alloy.",
             llm=openai.realtime.RealtimeModel(
-                voice="alloy", turn_detection=None, input_audio_transcription=None
+                # it's necessary to turn off turn detection in the Realtime API in order to use
```
jayeshp19, line 25: Should we initiate llm inside session constructor or add an comment there to make it less consfusing for users?

davidzhao: yeah great call!

longcw (approved):

Issue #1906, opened 2025-04-06
Push to talk on rc6


aniche:
Hello , we are using livekit-agents==1.0.0rc6. We used the code from [here](https://github.com/livekit/agents/blob/dev-1.0/examples/voice_agents/push_to_talk.py)
However, we are seeing something very funky, when pushing on the push2talk button, only the third time it actually send the voice message. "on_end_of_turn" code is reached only on the third time the button is clicked, also it seems something is off with the order of messages in the chat context.

on the frontend the push2talk button does is:

await room.localParticipant.performRpc({
destinationIdentity: agentIdentity,
method: 'start_turn',
payload: JSON.stringify({}),
});
await room.localParticipant.setMicrophoneEnabled(true);

what is going on

longcw:
`Agent.on_end_of_turn` is called when STT recognized a text message from user speech but not when the `start_turn` or `end_turn` method called. 

For now the `turn_detection="manual"` means the Agent won't trigger a reply automatically, the user transcription will be send to the `Agent.on_end_of_turn` callback when STT recognized (it may be called multiple times if STT recognized multiple sentences), it only generates a reply when  `session.generate_reply()` is called.

One issue is that when `end_turn` is called right after the speech the STT may not finalize the user transcription so there is no text send to the callback. I can create a fix to buffer the user transcription internally and make sure `on_end_of_turn` is  called only once (even with empty text or interim text from STT) when `session.generate_reply()` is called.

longcw:
Fixed in #1908 with `session.clear_user_turn` and `session.commit_user_turn` methods, the later will trigger a reply just like the VAD turn detection, but manually.

Issue #1901, opened 2025-04-06
Cancel turn feature in manual turn


well-balanced:
I’m developing with the [push_to_talk example](https://github.com/livekit/agents/blob/07f03f92fee1d975e0058ce25fbfa7ff59091cfb/examples/voice_agents/push_to_talk.py) as a reference, and I’m using a custom cancel_turn RPC in addition to the existing ones. The issue I’m encountering is that even when I call `cancel_turn` to disable the input audio, the `on_end_of_turn` method still receives the audio input that was spoken before the cancel.

For example:
```
USER: "My name is Wynn"  
(cancel_turn RPC is called)  
USER: "Do you remember my name?"  
(end_turn RPC is called)  
on_end_of_turn is invoked  
new_message argument: { role: "user", content: [" My name is Wynn\nDo you remember my name?"] }
```

As you can see, the first part (“My name is Wynn”) was meant to be canceled, but it still appears in the final transcript. Is there any way to distinguish or discard the portion of the transcript that was spoken before the cancel_turn?

longcw:
So you want to drop the buffered text when you cancel a turn? You can write your own `on_end_of_turn` https://github.com/livekit/agents/blob/main/examples/voice_agents/push_to_talk.py#L34-L37 to buffer the text until it the `end_turn` called, if it was canceled you can drop the buffer.

well-balanced:
@longcw 

I’m already implementing on_end_of_turn. As shown in the example I shared earlier, when on_end_of_turn is triggered, I have no clear way to decide whether the previous buffered text like “My name is Wynn” should be updated to chat_ctx or not.

Here’s the flow:

Start turn → Text A → Cancel turn (no on_end_of_turn) → Start turn → Text B → End turn → on_end_of_turn (receives Text A + B)

longcw:
One possible problem is that the stt may buffer the user transcription if the first turn canceled too quick. Will check how to ignore the stt buffer as well.

davidzhao:
@longcw theo had proposed raising a specific exception inside `on_end_of_turn`. it's currently not implemented but I've [stubbed out](https://github.com/livekit/agents/blob/9c913c7617f77da8a80bf0a69385ddd2ee33aa7a/examples/voice_agents/push_to_talk.py#L34) how it might look in an example.

well-balanced:
Thanks @longcw and @davidzhao — that all makes sense.
I’ll wait for the upcoming PR to see how the buffer handling changes.
[...]

PR #1908, opened 2025-04-07
add method to manual end the user turn

longcw:
added `session.end_user_turn` to manually trigger the end of turn callback, `Agent.on_end_of_turn` is called only once when the `session.end_user_turn` called, the user transcriptions are buffered in the AudioRecognition when the turn detection is manual. This makes the pipeline agent has the same behavior with the realtime model.

related https://github.com/livekit/agents/issues/1901 and https://github.com/livekit/agents/issues/1906,


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -528,6 +529,19 @@ def interrupt(self) -> None:
         if self._rt_session is not None:
             self._rt_session.interrupt()
 
+    def start_user_turn(self) -> None:
+        if self._audio_recognition:
+            self._audio_recognition.start_user_turn()
+
+    async def end_user_turn(self, *, cancelled: bool) -> None:
```
longcw, line 536: add the `cancelled` option mainly for realtime model, for pipeline agent user can manually skip the message in  `on_end_of_turn` but for realtime model it has to call the `rt_session.clear_audio()`

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -368,6 +368,30 @@ def interrupt(self) -> None:
 
         self._activity.interrupt()
 
+    def start_user_turn(self) -> None:
```
davidzhao, line 371: maybe just `start_turn` and `end_turn`?

since turn detection as a concept applies only to the user.. agent turn is controlled by `generate_reply` and `interrupt`

davidzhao (commented):
this lg!

livekit-agents/livekit/agents/voice/agent_session.py
longcw: how about `mark_turn_start` and `mark_turn_end`?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -528,6 +529,19 @@ def interrupt(self) -> None:
...
 
+    def mark_turn_start(self) -> None:
+        if self._audio_recognition:
+            self._audio_recognition.mark_turn_start()
+
+    async def mark_turn_end(self, *, cancelled: bool) -> None:
+        if self._audio_recognition:
+            await self._audio_recognition.mark_turn_end(cancelled=cancelled)
+        elif not cancelled:
+            await self.on_end_of_turn(new_transcript="")
```
theomonnom, line 540: the _audio_recognition is always there, seems like the `elif not cancelled` will never be used.

theomonnom (commented):
I'm wondering if exposing a "clear_turn" method that clears the transcript and the audio would be sufficient?

Then you use it with the combination of the `generate_reply` but without passing any arguments to it?

Maybe generate_reply by default uses the buffered audio for realtime and the buffered transcript for the stateless llm?

So when you want to cancel a turn you call `session.clear_turn()` and you call `session.generate_reply()` when the turn is done

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: Yes, but from type annotation it could be None, maybe a `assert self._audio_recognition`?

longcw:
> I'm wondering if exposing a "clear_turn" method that clears the transcript and the audio would be sufficient?
> 
> Then you use it with the combination of the `generate_reply` but without passing any arguments to it?
> 
> Maybe generate_reply by default uses the buffered audio for realtime and the buffered transcript for the stateless llm?
> 
> So when you want to cancel a turn you call `session.clear_turn()` and you call `session.generate_reply()` when the turn is done

I had the same idea at the beginning, but what if we want to end the user turn to trigger the `on_end_of_turn` without a response?

theomonnom (approved):

PR #1910, opened 2025-04-07
add realtime audio node

longcw:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1192,7 +1194,8 @@ async def _read_messages(
                     audio_out = None
                     if audio_output is not None:
                         forward_task, audio_out = perform_audio_forwarding(
```
davidzhao, line 1196: wouldn't be cleaner to apply this within `perform_audio_forwarding` ?

examples/voice_agents/process_output_audio.py
```diff
@@ -0,0 +1,119 @@
...
+from livekit.plugins import deepgram, openai, silero
+
+try:
+    import librosa
+except ImportError:
+    raise ImportError(
+        "librosa is required to run this example, install it with `pip install librosa`"
+    ) from None
+
+# uncomment to enable Krisp background voice/noise cancellation
```
davidzhao, line 28: I think we can remove krisp related stuff from here, and replace with a paragraph about what this example is demonstrating.

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: I think this makes `perform_audio_forwarding` cleaner, it take a audio stream and forward it to the AudioOutput but don't care where the audio stream comes from.

examples/voice_agents/process_output_audio.py
longcw: updated

theomonnom (approved):
lgtm!

theomonnom:
Can you change the example name to smtgh like "speed up" audio

PR #1922, opened 2025-04-08
fix stt fallback adapter for openai azure and stream adapter

longcw:


livekit-agents/livekit/agents/stt/stream_adapter.py
```diff
@@ -8,6 +8,11 @@
 from ..vad import VAD, VADEventType
 from .stt import STT, RecognizeStream, SpeechEvent, SpeechEventType, STTCapabilities
 
+# already a retry mechanism in STT.recognize, don't retry in stream adapter
+DEFAULT_STREAM_ADAPTER_API_CONNECT_OPTIONS = APIConnectOptions(
```
theomonnom, line 12: You can keep this var private

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -468,6 +470,7 @@ async def recv_task(ws: aiohttp.ClientWebSocketResponse):
 
                 except Exception:
                     logger.exception("failed to process OpenAI message")
+                    raise
```
theomonnom, line 473: Ideally we only raise the exceptions [here](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/_exceptions.py), like https://github.com/livekit/agents/blob/be54bfacb3e60cb3cc1f69ac3799965382a237e4/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py#L596

theomonnom (approved):

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
```diff
@@ -198,15 +198,20 @@ async def process_input():
 
                 process_input_task = asyncio.create_task(process_input())
                 wait_reconnect_task = asyncio.create_task(self._reconnect_event.wait())
+                wait_stopped_task = asyncio.create_task(self._session_stopped_event.wait())
```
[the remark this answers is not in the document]
longcw: why this would be a deadlock? added it here to stop the `process_input_task` when the session closed.

```python
done, _ = await asyncio.wait(
                        [process_input_task, wait_reconnect_task, wait_stopped_task],
                        return_when=asyncio.FIRST_COMPLETED,
                    )
```

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
longcw: it's to propagate the exception to the fallback adapter. Actually we don't need a try except here, we have retry in`RecognizeStream._main_task` and the fallback adapter

longcw:
> Is this also fixing the FallbackAdapter for TTS?

I have tested the TTS FallbackAdapter and it works well, STT one also works, it was the openai STT plugin didn't raise the error when failed so the fallback adapter cannot switch to the next.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
theomonnom: it isn't necessarily related to the fallback adapter. What I mean is that ideally, a user using livekit-agents should only worry about catching the errors inside https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/_exceptions.py

longcw: It's already covered in https://github.com/livekit/agents/blob/longc/fix-stt-fallback/livekit-agents/livekit/agents/stt/stt.py#L223

longcw: or you mean we should only catch all other errors and raise the same predefined one?

longcw: Reverted the raise here since it's for the message processing, we also catch such errors in deepgram stt.

PR #1923, opened 2025-04-08
Deepgram Nova-3 multi-language support

msaelices:
## Context

Nova-3 already supports multi-language, as you can see here: https://deepgram.com/learn/introducing-nova-3-speech-to-text-api

> Nova-3 is the first voice AI model to offer real-time multilingual transcription.

But, if you tried with a configuration like this one:
```python
    deepgram_stt = deepgram.STT(
        model='nova-3',
        language='multi',
    )
```

You will receive this error in the logs and it will switch back to `nova-2`:

image

## JBTD

So remove the `nova-3` model from the English-only Deepgram models.

### DIFF
--- a/.github/next-release/changeset-65b3292c.md
+++ b/.github/next-release/changeset-65b3292c.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-deepgram": patch
+---
+
+Deepgram Nova-3 multi-language support (#1923)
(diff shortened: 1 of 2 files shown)


theomonnom (approved):
awesome! thanks

msaelices:
I've just seen this PR which has a similar fix: https://github.com/livekit/agents/pull/1736/ but with more features.

The target branch was not `main` so it could conflict with this or maybe it's cleanly merged.

PR #1935, opened 2025-04-08
add an example of RPC method that uses awaitable interrupt

jayeshp19:


examples/voice_agents/push_to_talk.py
```diff
@@ -51,7 +51,7 @@ async def entrypoint(ctx: JobContext):
 
     @ctx.room.local_participant.register_rpc_method("start_turn")
     async def start_turn(data: rtc.RpcInvocationData):
-        session.interrupt()
+        await session.interrupt()
```
longcw, line 54: Do we really need await here?

examples/voice_agents/interrupt_example.py
```diff
@@ -0,0 +1,92 @@
...
+    # example of RPC method that uses awaitable interrupt
+    @ctx.room.local_participant.register_rpc_method("emergency_interrupt")
+    async def on_emergency_interrupt(data: rtc.RpcInvocationData) -> None:
+        logger.info("Emergency interrupt requested via RPC")
+
+        # await interrupt to ensure chat context is fully updated
+        await session.interrupt()
+
+        # now safely access chat context
+        chat_history = session.history
```
longcw, line 78: so the await is required when user want to access the chat context right after the `interrupt` right? maybe rename the example to awaitable interrupt and mention the use case on the top of the example.

examples/voice_agents/push_to_talk.py
jayeshp19: we don't but I think it's a good practice since we now return a future

longcw: I would prefer to await only if it's necessary. For this case, it's better to start the user turn asap to pickup the input audio.

examples/voice_agents/awaitable_interrupt.py
```diff
@@ -0,0 +1,66 @@
...
+
+    session = AgentSession(
+        vad=silero.VAD.load(),
+        stt=openai.STT(),
+        llm=openai.LLM(model="gpt-4o-mini"),
+        tts=openai.TTS(voice="alloy"),
+    )
+
+    @ctx.room.local_participant.register_rpc_method("emergency_interrupt")
+    async def on_emergency_interrupt(data: rtc.RpcInvocationData) -> None:
```
theomonnom, line 45: It's kind of misleading to use a function tool to interrupt? Taking this example from a user perspective, it's hard to understand what is on_emergency_interrupt really doing.

`await session.interrupt()` isn't very useful inside a function_call

PR #1941, opened 2025-04-09
fix nodes types

theomonnom:
The implementation is separated now into Agent.default now (inner class) 

We should recommend people to use `Agent.default.tts_node` instead of `super().tts_node`.

The only difference is the type checker being confused by the latter but both are compatible at runtime. 

The reason why type checkers are confused is because llm_node/tts_node/... could either be an AsyncGenerator or a Coroutine.

### DIFF
--- a/examples/voice_agents/llamaindex-rag/retrieval.py
+++ b/examples/voice_agents/llamaindex-rag/retrieval.py
@@ -76,7 +76,7 @@ class RetrievalAgent(Agent):
         print(f"update instructions: {instructions[:100].replace('\n', '\\n')}...")
         await self.update_instructions(instructions)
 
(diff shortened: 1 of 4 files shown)


longcw (approved):
lgtm!

longcw:
I got an error when running examples
```bash
  File "/home/longc/data/code/agents/livekit-agents/livekit/agents/voice/agent.py", line 397, in stt_node
    activity = agent.__get_activity_or_raise()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'MyAgent' object has no attribute '_default__get_activity_or_raise'. Did you mean: '_Agent__get_activity_or_raise'?
```

davidzhao (approved):

PR #1943, opened 2025-04-09
allow to edit the chat_ctx inside on_user_turn_completed

theomonnom:
- This is currently only implemented for the stateless LLM, this will require more work for the Realtime models (We will need to hash each message, and find the differences to update to the OpenAI server)
- The changes made to the chat_ctx are only for the current turn, they're not preserved inside the Agent.chat_ctx (Only the new user message will be kept)

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -460,7 +460,8 @@ class AgentActivity(RecognitionHooks):
     def generate_reply(
         self,
         *,
-        user_input: NotGivenOr[str] = NOT_GIVEN,
+        user_message: NotGivenOr[llm.ChatMessage] = NOT_GIVEN,
+        chat_ctx: NotGivenOr[llm.ChatContext] = NOT_GIVEN,
         instructions: NotGivenOr[str] = NOT_GIVEN,
         tool_choice: NotGivenOr[llm.ToolChoice] = NOT_GIVEN,
         allow_interruptions: NotGivenOr[bool] = NOT_GIVEN,
@@ -477,9 +478,7 @@ class AgentActivity(RecognitionHooks):
             allow_interruptions = NOT_GIVEN
 
         log_event(
-            "generate_reply",
-            user_input=user_input or None,
-            instructions=instructions or None,
+            "generate_reply", new_message=user_message or None, instructions=instructions or None
         )
 
         from .agent import _get_inline_task_info
@@ -505,7 +504,8 @@ class AgentActivity(RecognitionHooks):
             self._create_speech_task(
                 self._realtime_reply_task(
                     speech_handle=handle,
-                    user_input=user_input or None,
+                    # TODO(theomonnom): support llm.ChatMessage for the realtime model
+                    user_input=user_message.text_content if user_message else None,
                     instructions=instructions or None,
                     model_settings=ModelSettings(tool_choice=tool_choice),
                 ),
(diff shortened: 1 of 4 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -460,7 +460,8 @@ def say(
     def generate_reply(
         self,
         *,
-        user_input: NotGivenOr[str] = NOT_GIVEN,
+        user_message: NotGivenOr[llm.ChatMessage] = NOT_GIVEN,
```
longcw, line 463: not a fan of `user_message`... I'd prefer `user_input: str` for easier usage.

If people want to add a image content, they can use `chat_ctx` directly.

theomonnom: This is internal, the generate_reply of AgentSession is unchanged.

The generate_reply inside the AgentActivity is private

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -505,7 +504,8 @@ def generate_reply(
             self._create_speech_task(
                 self._realtime_reply_task(
                     speech_handle=handle,
-                    user_input=user_input or None,
+                    # TODO(theomonnom): support llm.ChatMessage for the realtime model
+                    user_input=user_message.text_content if user_message else None,
```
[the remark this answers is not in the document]
theomonnom: I think it's fine since you have to use our own turn detector for the on_end_of_turn to be called. It's still better than before

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -460,7 +460,8 @@ def say(
     def generate_reply(
         self,
         *,
-        user_input: NotGivenOr[str] = NOT_GIVEN,
+        user_message: NotGivenOr[llm.ChatMessage] = NOT_GIVEN,
+        chat_ctx: NotGivenOr[llm.ChatContext] = NOT_GIVEN,
```
longcw, line 464: What if both `user_message` and `chat_ctx` provided, we add the user message to the chat ctx?

theomonnom: yes, user_message is the new one, chat_ctx is the context of the llm. It could be combined, but this is how our current approach of the user facing "user_input: str" works

theomonnom:
This is for https://linear.app/livekit/issue/AGT-929/the-ability-to-manipulate-whats-persisted-into-chat-context
It's like another way to do RAG

livekit-agents/livekit/agents/voice/agent_activity.py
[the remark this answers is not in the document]
longcw: But it still doesn't work for the realtime model, the behavior will be different for pipeline agent and realtime agent in `on_end_of_turn`, if we make it implicit it's hard to debug..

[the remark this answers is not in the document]
longcw: If the only benefit is to replace the `self.session.update_chat_ctx()` in the end of turn callback, I would prefer to not do it until we can make the realtime model works like pipeline model.

davidzhao (approved):

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -804,7 +808,12 @@ async def on_end_of_turn(self, info: _EndOfTurnInfo) -> None:
         if isinstance(self.llm, llm.RealtimeModel):
             # ignore stt transcription for realtime model
             new_transcript = ""
```
longcw, line 810: Update the user message to None for realtime model

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -804,7 +808,12 @@ async def on_end_of_turn(self, info: _EndOfTurnInfo) -> None:
...
-        speech_handle = self.generate_reply(user_input=new_transcript)
+
+        # Ensure the new message is passed to generate_reply
+        # This preserves the original message_id, making it easier for users to track responses
+        speech_handle = self.generate_reply(
```
longcw, line 814: How to raise the warning when ppl edited the turn ctx and user message for realtime model? Any realtime use case with local VAD or EOU model goes to here as well.

PR #1946, opened 2025-04-09
bump gemini sdk version & add video support for gemini multimodal 

jayeshp19:


examples/voice_agents/gemini_video_agent.py
```diff
@@ -0,0 +1,56 @@
...
+logger = logging.getLogger("gemini-video-agent")
+
+load_dotenv()
+
+
+class GeminiAgent(Agent):
+    def __init__(self) -> None:
+        super().__init__(
+            instructions="you are gemini, a helpful assistant",
+            llm=google.beta.realtime.RealtimeModel(
```
jayeshp19, line 25: should we remove `beta` now?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -99,6 +109,8 @@ def __init__(
         top_k: NotGivenOr[int] = NOT_GIVEN,
         presence_penalty: NotGivenOr[float] = NOT_GIVEN,
         frequency_penalty: NotGivenOr[float] = NOT_GIVEN,
+        input_audio_transcription: NotGivenOr[bool] = NOT_GIVEN,
+        output_audio_transcription: NotGivenOr[bool] = NOT_GIVEN,
```
theomonnom, line 113: I don't think it is a good idea to do a bool here. What if they add fields in the future

bcherry (approved):
awesome this is great

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
jayeshp19: users currently need to pass AudioTranscriptionConfig() with no arguments to get a transcription. I think that might be confusing. Maybe we can add this later if users request changes?

jayeshp19: rn it's just a placeholder
https://github.com/googleapis/python-genai/blob/main/google/genai/types.py#L10118

theomonnom: It's not a big deal to create a dataclass for it. Let's be future-proof here. Knowing it's enabled by default, most users will just do input_audio_transcription=None when disabling it

jayeshp19: I've set input_transcription to None as default because it's only supported in vertex

Issue #1932, opened 2025-04-08
Deepgram STT doesn't work with FallbackAdapter


zizhong:
Deepgram STT doesn't work with FallbackAdapter.

Sometimes the forward input task exits.

PR #1947, opened 2025-04-09
fix: make STT fallback more robust

zizhong:
to fix https://github.com/livekit/agents/issues/1932

### DIFF
--- a/livekit-agents/livekit/agents/stt/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/stt/fallback_adapter.py
@@ -261,20 +261,22 @@ class FallbackRecognizeStream(RecognizeStream):
         forward_input_task: asyncio.Task | None = None
 
         async def _forward_input_task() -> None:
-            with contextlib.suppress(RuntimeError):  # stream might be closed
-                async for data in self._input_ch:
-                    for stream in self._recovering_streams:
-                        if isinstance(data, rtc.AudioFrame):
-                            stream.push_frame(data)
-                        elif isinstance(data, self._FlushSentinel):
-                            stream.flush()
+            try:
+                while True:
+                    with contextlib.suppress(RuntimeError):  # stream might be closed
+                        async for data in self._input_ch:
+                            for stream in self._recovering_streams:
+                                if isinstance(data, rtc.AudioFrame):
+                                    stream.push_frame(data)
+                                elif isinstance(data, self._FlushSentinel):
+                                    stream.flush()
 
-                    if main_stream is not None:
-                        if isinstance(data, rtc.AudioFrame):
-                            main_stream.push_frame(data)
-                        elif isinstance(data, self._FlushSentinel):
-                            main_stream.flush()
-
+                            if main_stream is not None:
+                                if isinstance(data, rtc.AudioFrame):
+                                    main_stream.push_frame(data)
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/stt/fallback_adapter.py
```diff
@@ -261,20 +261,22 @@ async def _run(self) -> None:
...
-                    if main_stream is not None:
-                        if isinstance(data, rtc.AudioFrame):
-                            main_stream.push_frame(data)
-                        elif isinstance(data, self._FlushSentinel):
-                            main_stream.flush()
-
+            try:
+                while True:
+                    with contextlib.suppress(RuntimeError):  # stream might be closed
+                        async for data in self._input_ch:
```
longcw, line 267: For each retry, it will recreate the forward input task if it's done in https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/stt/fallback_adapter.py#L295. 

It should be fine to let the `_forward_input_task` stop with the recreation. Do you have an example to reproduce the problem, I'd like to understand the root cause.

zizhong: Yes, my setup is I have two deepgram STTs in fallback adapter in my agent session. On top of that, I have a call application. The steps to reproduce the issues are
* start a call
* talk to the agent and confirm STT working
* cut down the connection to the first STT (on MacOS with LittleSniff)

There could be 
* a successful fallback that when I talk again, it continues to work.
* a stuck situation, that no more STT activities.

With the fix, it always works.

livekit-agents/livekit/agents/stt/fallback_adapter.py
```diff
@@ -261,20 +261,22 @@ async def _run(self) -> None:
...
+                                    stream.flush()
+
+                            if main_stream is not None:
+                                if isinstance(data, rtc.AudioFrame):
+                                    main_stream.push_frame(data)
+                                elif isinstance(data, self._FlushSentinel):
+                                    main_stream.flush()
+            finally:
                 if main_stream is not None:
                     main_stream.end_input()
```
longcw, line 281:
```suggestion
            async for data in self._input_ch:
                try:
                    for stream in self._recovering_streams:
                        if isinstance(data, rtc.AudioFrame):
                            stream.push_frame(data)
                        elif isinstance(data, self._FlushSentinel):
                            stream.flush()

                    if main_stream is not None:
                        if isinstance(data, rtc.AudioFrame):
                            main_stream.push_frame(data)
                        elif isinstance(data, self._FlushSentinel):
                            main_stream.flush()
                except RuntimeError:
                    pass
                except Exception:
                    logger.exception("error happened in forwarding input", extra={"streamed": True})

            if main_stream is not None:
                main_stream.end_input()
```

livekit-agents/livekit/agents/stt/fallback_adapter.py
longcw: Thanks @zizhong, I can reproduce the issue. Can you change the fix to like the following, we may still want the forward task to stop when the `_input_ch` closed.

zizhong: Thanks! Makes sense.

longcw (approved):

PR #1950, opened 2025-04-09
Simpler api to end the session and handling sip users

Shubhrakanti:
Wrap the LiveKit API functions to manage sip users and close sessions for better UX


livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +292,101 @@ async def connect(
...
+        def _on_task_done(t: asyncio.Task[Any]):
+            self._pending_tasks.pop(task_id)
+            if t.exception():
+                fut.set_exception(t.exception())
+            else:
+                fut.set_result(t.result())
+
+        task.add_done_callback(_on_task_done)
+        self._pending_tasks[task_id] = fut
+        return fut
```
Shubhrakanti, line 314: Should we move this to a utility? We have something that's not used right now - I could modify and repurpose it as task registry or something. 

https://github.com/livekit/agents/blob/5a17e0eaea5d44ea3c4035fa760f26bc950c2ea3/livekit-agents/livekit/agents/utils/aio/task_set.py#L10-L31

livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +292,101 @@ async def connect(
...
+    def disconnect(self) -> asyncio.Future[None]:
+        """Disconnects the agent from the room, but does not delete the room."""
+        return self._create_task_with_future("disconnect", self._room.disconnect())
+
+    def delete_room(self) -> asyncio.Future[None]:
+        """Deletes the room and disconnects all participants."""
+        return self._create_task_with_future(
+            "delete_room",
+            self.api.room.delete_room(api.DeleteRoomRequest(room=self._room.name))
+        )
```
Shubhrakanti, line 326: @davidzhao I know you mentioned something like `ctx.end_session(all=True) # where all=True will issue a delete_room to end it for everyone`

I think this is better because there are times where we want to do one but not the other. Ex. 
- Content Moderation agent fails - just disconnect for the room
- Transfer a SIP call - close the room for everyone

Let me know if this is on the right path.

bcherry (commented):
this is a great add, love it

livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +292,106 @@ async def connect(
...
+
+        task.add_done_callback(_on_task_done)
+        self._pending_tasks[task_id] = fut
+        return fut
+
+    def disconnect(self) -> asyncio.Future[None]:
+        """Disconnects the agent from the room, but does not delete the room."""
+        return self._create_task_with_future("disconnect", self._room.disconnect())
+
+    def delete_room(self) -> asyncio.Future[None]:
```
davidzhao, line 321: would exceptions be propagated correctly if someone did `await ctx.delete_room()`? if they didn't await it, should we automatically handle this somehow?

in particular `404` would be a common error if the room has already ended

livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +292,106 @@ async def connect(
...
+        self,
+        participant: rtc.RemoteParticipant,
+        transfer_to: str,
+        play_dialtone: bool = False,
+    ) -> asyncio.Future[api.SIPParticipantInfo]:
+        """Transfer a SIP participant to another number.
+
+        Args:
+            participant: The participant to transfer
+            number: The number to transfer the participant to
```
davidzhao, line 370: we can further simplify this.. the protocol takes in a `sip:user@host` or `tel:+1223-445-6678`

```suggestion
            transfer_to: The number or SIP destination to transfer the participant to
```

livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +292,106 @@ async def connect(
...
+        )
+
+    def get_sip_participants(self) -> list[rtc.RemoteParticipant]:
+        return [
+            p
+            for p in self._room.remote_participants.values()
+            if p.kind == rtc.ParticipantKind.PARTICIPANT_KIND_SIP
+        ]
+
+    def add_sip_participant(
```
davidzhao, line 334: I'm not convinced that we should be wrapping add_sip_participant and transfer_sip_participant.. both require careful error handling here

livekit-agents/livekit/agents/job.py
Shubhrakanti: Whoops I updated the variable name but not the comments.

livekit-agents/livekit/agents/job.py
Shubhrakanti: Going off of @theomonnom 's [suggestion](https://live-kit.slack.com/archives/C07TWVC0W4A/p1744023996116479?thread_ts=1743626761.816229&cid=C07TWVC0W4A). 

Couldn't we handle errors with something like?

```python
fut = ctx.transfer_sip_participant(participant, "tel:+18003310500")
fut.add_done_callback(
    lambda f: some_funciton_to_handle_failed_transfer()
    if f.exception() else logger.info("Transfer SIP participant succeeded")
)
```

livekit-agents/livekit/agents/job.py
Shubhrakanti: https://github.com/livekit/agents/pull/1950#discussion_r2036852278

theomonnom: I think it's OK, we can forward the error to the asyncio.Future

livekit-agents/livekit/agents/job.py
theomonnom: Blah agree it's kind of ugly, let's return the asyncio.Task but keep the type to asyncio.Future?

livekit-agents/livekit/agents/job.py
```diff
@@ -108,6 +108,7 @@ def __init__(
                 list[rtc.ParticipantKind.ValueType] | rtc.ParticipantKind.ValueType,
             ]
         ] = []
+        self._pending_tasks: dict[str, asyncio.Future[None]] = {}
```
theomonnom, line 111: I don't think we should have a dict here, you will leak tasks if you call delete_room twice for e.g

livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +292,106 @@ async def connect(
...
+            for p in self._room.remote_participants.values()
+            if p.kind == rtc.ParticipantKind.PARTICIPANT_KIND_SIP
+        ]
+
+    def add_sip_participant(
+        self,
+        sip_call_to: str,
+        sip_trunk_id: str,
+        participant_identity: str,
+        participant_name: str,
```
davidzhao, line 339: let's not require name (i.e. `NotGivenOr[str]`).. if it's not given, just create a default.

livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +292,106 @@ async def connect(
...
+            p
+            for p in self._room.remote_participants.values()
+            if p.kind == rtc.ParticipantKind.PARTICIPANT_KIND_SIP
+        ]
+
+    def add_sip_participant(
+        self,
+        sip_call_to: str,
+        sip_trunk_id: str,
+        participant_identity: str,
```
davidzhao, line 338: use explicit kwargs
```suggestion
        *,
        call_to: str,
        trunk_id: str,
        participant_identity: str,
```

and let's be sure to add docstring for these, along with examples: i.e. call_to example: `+12345555555`

livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +292,106 @@ async def connect(
...
+                    sip_trunk_id=sip_trunk_id,
+                    sip_call_to=sip_call_to,
+                    participant_name=participant_name,
+                )
+            ),
+        )
+
+    def transfer_sip_participant(
+        self,
+        participant: rtc.RemoteParticipant,
```
davidzhao, line 362: we can accept either identity or object

```suggestion
        participant: rtc.RemoteParticipant | str,
```

livekit-agents/livekit/agents/job.py
```diff
@@ -291,6 +292,106 @@ async def connect(
...
+        transfer_to: str,
+        play_dialtone: bool = False,
+    ) -> asyncio.Future[api.SIPParticipantInfo]:
+        """Transfer a SIP participant to another number.
+
+        Args:
+            participant: The participant to transfer
+            transfer_to: The number or SIP destination to transfer the participant to
```
davidzhao, line 370: the transfer format is a bit weird, so we should document:

`sip:<user>@<host>` transfer to another sip host
`tel:+12345...` transfer to a phone number

examples/voice_agents/error_callback.py
```diff
@@ -39,10 +41,36 @@ def on_error(ev: ErrorEvent):
...
+                logger.error(f"Error transferring SIP participant: {f.exception()}")
+            else:
+                logger.info("SIP participant transferred")
+            ctx.delete_room()
+
+        # See https://docs.livekit.io/sip/ on how to set up SIP participants
+        if participant.kind == ParticipantKind.PARTICIPANT_KIND_SIP:
+            ctx.transfer_sip_participant(participant, "tel:+18003310500").add_done_callback(
+                on_sip_transfer_done
+            )
```
Shubhrakanti, line 69: once event emitter can handle asynchronous functions. This would be 
```python 
await ctx.transfer_sip_participant(...)
await ctx.delete_room()
```

examples/voice_agents/error_callback.py
```diff
@@ -39,10 +41,36 @@ def on_error(ev: ErrorEvent):
...
+            for p in ctx.room.remote_participants.values()
+            if p.kind == ParticipantKind.PARTICIPANT_KIND_SIP
+        ][0]
+
+        def on_sip_transfer_done(f: asyncio.Future):
+            if f.exception():
+                logger.error(f"Error transferring SIP participant: {f.exception()}")
+            else:
+                logger.info("SIP participant transferred")
+            ctx.delete_room()
```
Shubhrakanti, line 63: @theomonnom I'm not sure this code is correct. Doesn't the user need to a keep a handle to this task? https://stackoverflow.com/questions/71938799/python-asyncio-create-task-really-need-to-keep-a-reference

Shubhrakanti: I agree that keep track of all the tasks in `job.py` is ugly, but now we dump complexity on the user? 

I guess it's okay since we're going to allow async on event emitter soon so the user can just do `await` on these callbacks/

Shubhrakanti (commented):
Removed task tracking in job.py and added better comments for SIP methods.

examples/voice_agents/error_callback.py
theomonnom: We should still store the tasks ourselves, what I meant is that converting to Future is ugly.

livekit-agents/livekit/agents/job.py
```diff
@@ -266,6 +268,92 @@ async def connect(
...
+            self.api.sip.transfer_sip_participant(
+                api.TransferSIPParticipantRequest(
+                    room_name=self._room.name,
+                    participant_identity=participant.identity,
+                    transfer_to=transfer_to,
+                    play_dialtone=play_dialtone,
+                )
+            ),
+        )
+        self._pending_tasks.append(task)
```
theomonnom, line 354: let's make sure to also delete the task from the list when they're completed

theomonnom (approved):
otherwise lgtm

PR #1952, opened 2025-04-10
[draft] add is_end_of_turn for user transcription

longcw:
This is a draft to see how to implement the `is_end_of_turn` flag for the transcription.
The current implementation only helps the case when using STT with EOU model.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -654,14 +654,15 @@ class AgentActivity(RecognitionHooks):
         if ev.user_transcription_enabled:
             self._session.emit(
                 "user_input_transcribed",
-                UserInputTranscribedEvent(transcript="", is_final=False),
+                UserInputTranscribedEvent(transcript="", is_final=False, is_end_of_turn=False),
             )
 
     def _on_input_audio_transcription_completed(self, ev: llm.InputTranscriptionCompleted) -> None:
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -654,14 +654,15 @@ def _on_input_speech_stopped(self, ev: llm.InputSpeechStoppedEvent) -> None:
...
+                UserInputTranscribedEvent(transcript="", is_final=False, is_end_of_turn=False),
             )
 
     def _on_input_audio_transcription_completed(self, ev: llm.InputTranscriptionCompleted) -> None:
         log_event("input_audio_transcription_completed")
+        # TODO(longc): do we have an "end_of_turn" event synced with user transcription for realtime model?  # noqa: E501
         self._session.emit(
             "user_input_transcribed",
-            UserInputTranscribedEvent(transcript=ev.transcript, is_final=True),
+            UserInputTranscribedEvent(transcript=ev.transcript, is_final=True, is_end_of_turn=True),
```
theomonnom, line 665: for the realtime API, I don't think _on_input_audio_transcription_completed is a reliable signal to trigger is_end_of_turn.

longcw: Yes, but I didn't find a better signal, actually this is the only signal for OAI realtime API? The user transcription generation is async with the conversation items.

PR #1954, opened 2025-04-10
update elevenlabs TTS ctor args

longcw:


davidzhao (approved):
lg!

davidzhao:
> lmk when this ships, we'll want to update https://docs.livekit.io/agents/v1/integrations/tts/elevenlabs/ too

this will make it. feel free to change. wanted to do this now instead of breaking after v1

bcherry:
https://github.com/livekit/web/pull/1117

PR #1968, opened 2025-04-11
Emit errors from open ai realtime model

Shubhrakanti:


livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -38,10 +40,13 @@ class GenerationCreatedEvent:
...
-@dataclass
-class ErrorEvent:
-    type: str
-    message: str
+class RealtimeModelError(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    type: Literal["realtime_model_error"] = "realtime_model_error"
+    timestamp: float
+    label: str
+    error: any = Field(..., exclude=True)
```
Shubhrakanti, line 48: I would like to wrap this with our own Error class (similar to [APIError](https://github.com/livekit/agents/blob/4c3d980d287b93d1fb4417f35098e33f178d2128/livekit-agents/livekit/agents/_exceptions.py#L14). Just not a lot of structure in the errors from the different web socket events in the realtime models.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -888,7 +899,13 @@ def _handle_error(self, event: ErrorEvent) -> None:
         )
         self.emit(
             "error",
-            llm.ErrorEvent(type=event.error.type, message=event.error.message),
+            llm.RealtimeModelError(
+                type="realtime_model_error",
```
theomonnom, line 903:
```suggestion
```

livekit-agents/livekit/agents/llm/realtime.py
theomonnom: I think we can still use APIError. There is the message field, I think it's OK

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -318,7 +319,17 @@ async def _recv_task() -> None:
                 msg = await ws_conn.receive()
                 if msg.type == aiohttp.WSMsgType.CLOSED:
                     if not closing:
-                        raise Exception("OpenAI S2S connection closed unexpectedly")
+                        error = Exception("OpenAI S2S connection closed unexpectedly")
```
theomonnom, line 322: It would be APIConnectionError here?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -888,7 +899,13 @@ def _handle_error(self, event: ErrorEvent) -> None:
         )
         self.emit(
             "error",
-            llm.ErrorEvent(type=event.error.type, message=event.error.message),
+            llm.RealtimeModelError(
+                type="realtime_model_error",
+                timestamp=time.time(),
+                label=self._realtime_model._label,
+                error=event.error,
```
theomonnom, line 906: let's be consistent and give an Exception here, let's use APIError? 
we may not have all the details, but I think it's fine for now

theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -16,6 +17,7 @@
 
 from livekit import rtc
 from livekit.agents import io, llm, utils
+from livekit.agents._exceptions import APIConnectionError, APIError
```
theomonnom, line 20:
```suggestion
from livekit.agents import APIConnectionError, APIError
```

theomonnom (approved):

PR #1974, opened 2025-04-11
add chat_ctx.truncate

longcw:


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -222,6 +226,29 @@ def copy(
 
         return ChatContext(items)
 
+    def truncate(self, *, max_items: int, preserve_instructions: bool = True) -> ChatContext:
```
davidzhao, line 229: should we have a way to truncate tools from context as well?

davidzhao (approved):

livekit-agents/livekit/agents/llm/chat_context.py
theomonnom: We should keep the API consistent with the other nested functions here

discard_instructions = False

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -222,6 +226,29 @@ def copy(
...
+        instructions: ChatMessage | None = None
+        if preserve_instructions:
+            from ..voice.generation import INSTRUCTIONS_MESSAGE_ID
+
+            idx = self.index_by_id(INSTRUCTIONS_MESSAGE_ID)
+            if idx is not None:
+                instructions = self._items.pop(idx)
```
theomonnom, line 236: I'm not sure if INSTRUCTIONS_MESSAGE_ID makes sense, because it is only for our AgentActivity thing, there could still be other system/developers messages inside the chat_ctx

livekit-agents/livekit/agents/llm/chat_context.py
theomonnom: I'm wondering if we need this argument here?

It could be achieved by doing `chat_ctx.copy(discard_instructions=True).truncate()`

davidzhao: that's a good point.. it is more flexible to include filters in `copy`

longcw: `chat_ctx.copy(exclude_instructions=True).truncate(max_items=6)` won't work as truncate will remove the chat items from start, the instructions will be removed if it's not preserved specifically.

`exclude_instructions` for `truncate` seems a little bit confusing, exclude it from truncating (so keep it in the result) or exclude it from the result? wdyt

livekit-agents/livekit/agents/llm/chat_context.py
longcw: I am not sure if we want to keep all the system messages in the middle of the chat ctx during truncating, maybe we save the first message if it's a `system` message instead of checking the `INSTRUCTIONS_MESSAGE_ID`?

livekit-agents/livekit/agents/llm/chat_context.py
longcw: added the instruction filter for `copy`.

but do you think we still need an option for `truncate` to preserve the first system message? I think for most of use cases ppl want to truncate the chat ctx like the follows
```
from
[system, assistant_1, user_2, ... x20, assistant_21, user_22]

to
[system, assistant_20, user_20, assistant_21, user_21]
```

davidzhao: I think by default, truncate should keep system message in the beginning.. users can filter out system message in `copy` if they don't want it to be there.

livekit-agents/livekit/agents/llm/chat_context.py
longcw: updated

livekit-agents/livekit/agents/llm/chat_context.py
longcw: yes, made the keep system message as default

theomonnom (approved):

PR #1978, opened 2025-04-11
Remove english-only disclaimer from turn detector readme

bcherry:


davidzhao (approved):
thanks! I missed it when updating the page

PR #1986, opened 2025-04-13
Update _run.py

MajorTal:
Save and restore terminal setting when the app quits. Fixes the bug where the terminal does not display the user input.

### DIFF
--- a/livekit-agents/livekit/agents/cli/_run.py
+++ b/livekit-agents/livekit/agents/cli/_run.py
@@ -46,91 +46,114 @@ def run_worker(args: proto.CliArgs, *, jupyter: bool = False) -> None:
     loop = asyncio.new_event_loop()
     asyncio.set_event_loop(loop)
 
-    if args.console:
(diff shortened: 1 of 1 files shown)


theomonnom (approved):
That lgtm!

livekit-agents/livekit/agents/cli/_run.py
```diff
@@ -46,91 +46,114 @@ def run_worker(args: proto.CliArgs, *, jupyter: bool = False) -> None:
...
+            finally:
+                loop.close()
+    finally:
+        # Restore terminal settings if needed
+        if original_terminal_settings is not None:
+            try:
+                import termios
+                fd = sys.stdin.fileno()
+                termios.tcsetattr(fd, termios.TCSADRAIN, original_terminal_settings)
+                logger.debug("Terminal settings restored")
```
theomonnom, line 155: I would just be less verbose
```suggestion
```

PR #1989, opened 2025-04-14
support dynamic tools from llm_node

longcw:


livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -57,14 +57,17 @@ def perform_llm_inference(
...
         llm_node = node(
             chat_ctx,
-            list(tool_ctx.function_tools.values()) if tool_ctx is not None else [],
+            tools,
             model_settings,
         )
         if asyncio.iscoroutine(llm_node):
             llm_node = await llm_node
 
+        tool_ctx.update_tools(tools)
```
theomonnom, line 69: tool_ctx could be None

This is definitely something I didn't plan to support. It's great! Since the ToolContext can be None here. Should we return a new ToolContext as part of the _LLMGenerationData

theomonnom: yes, it's ok to remove None.

I'm just wondering if we could do this line:
` tool_ctx.update_tools(tools)`

inside TaskActivity instead, so it's obvious when reading the code there that the tools could change.

otherwise lgtm

theomonnom (approved):

livekit-agents/livekit/agents/voice/generation.py
longcw: sounds good

longcw: to move the `tool_ctx.update_tools(tools)` to task activity it needs a future to track when llm_node returned. let's keep that in generation for now...

PR #1990, opened 2025-04-14
keep exception chaining 

theomonnom:


livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -223,7 +223,7 @@ async def __anext__(self) -> ChatChunk:
             val = await self._event_aiter.__anext__()
         except StopAsyncIteration:
             if not self._task.cancelled() and (exc := self._task.exception()):
-                raise exc from None
+                raise exc
```
longcw, line 226: how to solve the B904 from ruff, maybe `raise ... from err`?

theomonnom:
adding noqa, seems to provide the most accurate stacktrace

PR #1991, opened 2025-04-14
🛠 Fix: Properly retrieve exceptions after graceful task cancellation

gigaverse-oz:
### Summary

This PR improves the `gracefully_cancel` (now called `cancel_and_wait`) function by ensuring that exceptions from cancelled or completed tasks are explicitly retrieved. This prevents `asyncio` from logging warnings like:

```
_GatheringFuture exception was never retrieved
future: <_GatheringFuture finished exception=CancelledError()>
```

### Root Cause

Previously, `gracefully_cancel`:
- Cancelled tasks
- Waited for their completion via internal waiters
- **Did not retrieve `.exception()` from the tasks**

When `asyncio.gather()` is used, it returns a `_GatheringFuture`, which wraps multiple tasks. If the gathered task is cancelled or raises an exception and that exception is never accessed, `asyncio` logs a warning — even if the tasks are awaited indirectly.

### Fix

After all futures are awaited, we now iterate over each future:
- Call `fut.exception()` to mark the exception as retrieved

### Example

Previously:
```python
await gracefully_cancel(*tasks)
# Logs: _GatheringFuture exception was never retrieved
```

Now:
```python
await gracefully_cancel(*tasks)
# No warning – all exceptions are properly retrieved
```

### Impact

- Silences unwanted `_GatheringFuture` warnings in normal cancellation flows
- Improves robustness and clarity during shutdown
- No behavioral change to existing application logic

### DIFF
--- a/.github/next-release/changeset-23d9c87a.md
+++ b/.github/next-release/changeset-23d9c87a.md
@@ -2,4 +2,4 @@
 "livekit-agents": patch
 ---
 
-🛠 Fix: Properly retrieve exceptions after graceful task cancellation (#1991)
+Fix: Properly retrieve exceptions after graceful task cancellation (#1991)
--- a/.github/next-release/changeset-23d9c87a.md
+++ b/.github/next-release/changeset-23d9c87a.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+🛠 Fix: Properly retrieve exceptions after graceful task cancellation (#1991)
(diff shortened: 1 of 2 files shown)


gigaverse-oz:
Hi @theomonnom,

I used deepgram, stt. This is the place that have the issue:
https://github.com/livekit/agents/blob/699eb31b9d5f199d1dad50fa7f1d1f38c3838424/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py#L571

You create 3 tasks, have only try-finally and use `asyncio.wait` with `asyncio.FIRST_COMPLETED`.
When I close the STT, the tasks get asyncio.TaskCancelled exception that is not catched (you exit the "wait" when the first one raises).
You "clean" the function by calling `await utils.aio.gracefully_cancel(*tasks, wait_reconnect_task)`
Simple solution will be also to do `asyncio.gethar(*tasks)` in the finally, but I assumed the issue might repeat in more places/plugins.

I'd be happy with either solutions as long it can be soon. It litters our logs :-(

PR #1995, opened 2025-04-14
delete livekit-plugins-rag

theomonnom:


davidzhao (approved):

bcherry:
@ShayneP FYI we'll need to update this example https://github.com/livekit-examples/python-agents-examples/blob/main/rag/rag_enriched_agent.py

PR #2002, opened 2025-04-15
fix: resolve type mismatch between _TurnDetector protocol and plugins

kechako:
The _TurnDetector protocol definition was not consistent with existing plugin implementations. Updated the protocol to match the actual plugin usage.


theomonnom (approved):
lgtm!

theomonnom:
Thanks!

PR #1975, opened 2025-04-11
add with_azure in openai realtime

jayeshp19:
closes #1973

PR #2005, opened 2025-04-15
fix OAI realtime model tool choice opt

longcw:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -131,6 +132,7 @@ def __init__(
         input_audio_transcription: NotGivenOr[InputAudioTranscription | None] = NOT_GIVEN,
         turn_detection: NotGivenOr[TurnDetection | None] = NOT_GIVEN,
         temperature: NotGivenOr[float] = NOT_GIVEN,
+        tool_choice: NotGivenOr[llm.ToolChoice | None] = NOT_GIVEN,
```
theomonnom, line 135: Is it really relevant to put tool_choice on the constructor? 
The TaskActivity is going to be aware of it and it seems pretty limiting vs doing it inside the Agent?

longcw: I don't think it's needed, but the current version is broken bc the `tool_choice` is added to the options while we didn't set it.

longcw: It's broken from here https://github.com/livekit/agents/pull/1975/files#diff-ff18b8060d7fec5a4f3840b596c32f3c8d4a4331794ce9909e0e685dd76ada0cR83

theomonnom: oh wtf

theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: Set it to None in constructor sounds good. But when we actually send the option to OAI it doesn't accept None so we convert None to `auto` in https://github.com/livekit/agents/pull/2005/files#diff-ff18b8060d7fec5a4f3840b596c32f3c8d4a4331794ce9909e0e685dd76ada0cL640. just FYI

PR #2008, opened 2025-04-15
azure speech sdk version upgrade (>=1.41.0 to >=1.43.0)

jayeshp19:
fixes azure error when using with endpoint


livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
```diff
@@ -111,6 +113,10 @@ def __init__(
                 "AZURE_SPEECH_HOST or AZURE_SPEECH_KEY and AZURE_SPEECH_REGION or speech_auth_token and AZURE_SPEECH_REGION or AZURE_SPEECH_KEY and speech_endpoint must be set"  # noqa: E501
             )
 
+        if speech_region and speech_endpoint:
+            logger.warning("speech_region and speech_endpoint both are set, using speech_endpoint")
```
theomonnom, line 117:
```suggestion
```
Do we need this warning, it seems like both of those values are NOT_GIVEN by default

theomonnom (approved):

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
jayeshp19: I added this because Azure’s SDK raises a ValueError if both are set. It’s a bit messy to unset one via env vars, so this handles it more gracefully

theomonnom: But if I just do `azure.STT()` when I want to use Azure, it'll always warn ?

jayeshp19: It shouldn't cuz `NOT_GIVEN` will return False

theomonnom: ah ok my bad, feel free to merge :)

PR #2014, opened 2025-04-16
Added Support for Rest-API based Deepgram Aura-2 - Version 0.x.

pranav-zc:
### Add Support for Deepgram Aura-2 Models (REST Only)

Deepgram has launched the **Aura-2** series of Text-to-Speech models. Currently, these models are **only supported via the REST API** — WebSocket support is expected to follow soon.

🔗 [Official Documentation](https://developers.deepgram.com/docs/streaming-text-to-speech)

#### Changes in this PR:
- Implemented conditional handling to support both **Aura (legacy)** and **Aura-2** models.
- **Aura models** use **streaming mode (WebSocket)**.
- **Aura-2 models** use **non-streaming mode (REST API)**.
- Mode selection is determined by inspecting the model tag prefix (`aura` vs `aura-2`).

Once WebSocket support for Aura-2 is available, we can enable it through a future patch or hotfix.

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
@@ -76,7 +76,7 @@ class TTS(tts.TTS):
             )
 
         if model.startswith("aura-2") and use_streaming:
-            raise ValueError(
+            logger.warning(
                 "Deepgram Aura-2 models do not support streaming yet, set use_streaming=False to use Aura-2 models. For more information see: https://developers.deepgram.com/docs/streaming-text-to-speech."
             )
 
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
@@ -76,7 +76,7 @@ class TTS(tts.TTS):
             )
 
         if model.startswith("aura-2") and use_streaming:
-            logger.warning(
+            raise ValueError(
                 "Deepgram Aura-2 models do not support streaming yet, set use_streaming=False to use Aura-2 models. For more information see: https://developers.deepgram.com/docs/streaming-text-to-speech."
             )
 
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
@@ -42,6 +42,7 @@ class TTS(tts.TTS):
         sample_rate: int = 24000,
         api_key: str | None = None,
         base_url: str = BASE_URL,
+        use_streaming: bool = True,
         word_tokenizer: tokenize.WordTokenizer = tokenize.basic.WordTokenizer(
             ignore_punctuation=False
         ),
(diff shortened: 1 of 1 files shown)


vnandan (approved):
These look to me

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
```diff
@@ -61,7 +61,10 @@ def __init__(
 
         """
         super().__init__(
-            capabilities=tts.TTSCapabilities(streaming=True),
+            # checking if the model is from aura-2 family (Aura-2 is currently available for the TTS REST API only.)
```
jayeshp19, line 64: Can we set `use_streaming=True` as the default in the constructor instead?

pranav-zc: okay will do the changes

davidzhao: should it be `NotGivenOr[bool]`? if it's not given, it'll default to streaming if model supports it.

pranav-zc: So, we would have to mentioned `use_streaming`, while defining the VoicePipelineAgent, like this?

```
tts=deepgram.TTS(
    model="aura-2-amalthea-en",
    api_key=settings.DEEPGRAM_KEY,
    use_streaming=False, # For Aura-2 models
)
```

davidzhao: I think the behavior should be:
- when it's not given, pick the right default, preferring streaming for aura 1
- when user specifies True, and using aura-2, it should log an error (but allowing it to continue, since DG could enable support for it before we are able to update the plugin)
- when user specifies False, always use non-streaming mode

pranav-zc:
@davidzhao @jayeshp19 I've made the changes according to the above discussion. The new flow is: 
* If no model name is specified (or a legacy model is used), default to the Aura (legacy) model with streaming enabled (original approach).
* If the Aura-2 model is specified and the use_streaming parameter is not provided, log an error suggesting the user set use_streaming=False.
* If the user explicitly sets use_streaming=False, always use the non-streaming mode, regardless of the model specified.

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/tts.py
```diff
@@ -72,6 +75,11 @@ def __init__(
                 "Deepgram API key required. Set DEEPGRAM_API_KEY or provide api_key."
             )
 
+        if model.startswith("aura-2") and use_streaming:
+            logger.error(
```
jayeshp19, line 79: Let's use warning here

pranav-zc: Okay!

pranav-zc:
Updated `logger.error` to `logger.warning`

jayeshp19 (approved):
LGTM

theomonnom:
Can this be ported to v1.0?

pranav-zc:
I'm still in the process of fully setting up v1.0 within our organization. I think having a separate PR for v1.0 would be better.

pranav-zc:
> I'm still in the process of fully setting up v1.0 within our organization. I think having a separate PR for v1.0 would be better.

Hey any updates on the merge?

theomonnom (approved):
lgtm thanks, would be great if we can port that over to v1!

PR #2016, opened 2025-04-16
fix: add missing FunctionToolsExecutedEvent export

mateuszkulpa:
Adds missing FunctionToolsExecutedEvent export in voice agents module


theomonnom (approved):
Thanks!

PR #2024, opened 2025-04-17
fix prompt_cached_tokens in llm metrics

longcw:


davidzhao (approved):
run a ruff pass before merging? to get rid of some of these errors

PR #2032, opened 2025-04-17
feat: fps sampler for pushing video

jayeshp19:


bcherry (commented):
yeah this looks good to me

livekit-agents/livekit/agents/utils/sampler.py
```diff
@@ -0,0 +1,25 @@
+import asyncio
+from abc import ABC, abstractmethod
+
+
+class FrameSampler(ABC):
+    @abstractmethod
+    def allow(self, frame) -> bool:
```
longcw, line 7: what is the type of `frame`?

jayeshp19: it should be `videoframe`, let me update it

theomonnom (commented):
That's a great idea, I'm just not sure of this API, wdyt about making the FrameSampler a Protocol and moving it inside the Google package so we have more flexibility to do changes in the future?

livekit-agents/livekit/agents/utils/sampler.py
```diff
@@ -0,0 +1,8 @@
+from typing import Protocol
+
+from livekit import rtc
+
+
+class FrameSampler(Protocol):
+    def allow(self, frame: rtc.VideoFrame) -> bool:
+        """Return True if this frame should be sent."""
```
theomonnom, line 8: Let's move to the google plugin

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -111,6 +113,7 @@ def __init__(
         frequency_penalty: NotGivenOr[float] = NOT_GIVEN,
         input_audio_transcription: NotGivenOr[AudioTranscriptionConfig | None] = NOT_GIVEN,
         output_audio_transcription: NotGivenOr[AudioTranscriptionConfig | None] = NOT_GIVEN,
+        video_fps: NotGivenOr[float] = NOT_GIVEN,
```
theomonnom, line 116: rename to `video_frame_sampler`?

theomonnom: This seems more flexible than directly limiting FPS

jayeshp19: Makes sense

jayeshp19: One issue I see is `video_frame_sampler` will be shared across sessions if passed from `RealtimeModel`.

jayeshp19: video_frame_sampler is stateful (e.g. FpsSampler tracks _last_time). If we reuse the same instance across multiple sessions, one session can affect another’s sampling behavior.

PR #2035, opened 2025-04-17
allow overriding eou threshold

jeradf:


livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/base.py
```diff
@@ -121,6 +121,7 @@ def __init__(
         self,
         model_type: EOUModelType = "en",  # default to smaller, english-only model
         inference_executor: InferenceExecutor | None = None,
+        custom_threshold: float | None = None,
```
davidzhao, line 124:
```suggestion
        unlikely_threshold: float | None = None,
```
also some docs about the use? i.e. it'll default to what's trained

livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/base.py
```diff
@@ -134,21 +135,32 @@ def __init__(
...
+        # if not found and it has a region code (e.g., 'en-us'), try the base language ('en')
+        if lang_data is None and "-" in lang:
+            base_lang = lang.split("-")[0]
+            lang_data = self._languages.get(base_lang)
+
+        if not lang_data:
+            logger.warning(f"Language {language} not supported by EOU model")
+            return None
+
+        if self._custom_threshold is not None:
```
davidzhao, line 160: this should be moved up.. we don't need to check the thresholds in lang_data if it's custom?

livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/english.py
```diff
@@ -11,8 +11,8 @@ def __init__(self):
 
 
 class EnglishModel(EOUModelBase):
-    def __init__(self):
-        super().__init__(model_type="en")
+    def __init__(self, custom_threshold: float | None = None):
```
davidzhao, line 14:
```suggestion
    def __init__(self, *, unlilkely_threshold: float | None = None):
```

livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/base.py
jeradf: well it needs to return `None` when the language is not found. see the `supports_language()` function below

davidzhao: oh i see.. ignore me then :)

theomonnom (approved):

Issue #2034, opened 2025-04-17
RoomIO.start() throws “text stream handler for topic ‘lk.chat’ already set” if called twice


well-balanced:
Hi team, first of all—thank you for the fantastic work on LiveKit! I’ve been loving the flexibility of the text‐stream API.

I ran into a “text stream handler for topic ‘lk.chat’ already set” error whenever I call RoomIO.start() a second time.

### Steps to Reproduce:
1. Create a RoomIO instance and call await roomIO.start().
2. Tear it down (e.g. call await roomIO.aclose(), but note that this does not unregister the text handler).
3. Call await roomIO.start() again.

You’ll immediately see:
```sh
ValueError: text stream handler for topic 'lk.chat' already set
```

I wondered if we could guard against any leftover handlers by clearing them up front. For example:

```py
async def start(self) -> None:
    self._room._text_stream_handlers.clear()
    # … rest of start() …
```

Thank you again for all your hard work!

well-balanced:
Hey! 😅
I just realized that `room_io.start()` was being called explicitly, and then again internally via the `session.start()` call with `ctx.room`:
```py
await session.start(
    agent=ElseAgent(
        room_metadata=room_metadata,
        ctx=ctx,
    ),
    room=ctx.room,
)
```

So the text stream handler for lk.chat was getting registered twice. That’s totally on me — thanks for the help and sorry for the noise! I’ll close this issue.

longcw:
I think it's better to ignore the room in `session.start` when a RoomIO is already set up for the session to avoid this. I'll add a fix for this.

Issue #2018, opened 2025-04-16
Breaking change in LiveKit 1.X – Unable to pass participant to agent


narendra-bluebash:
After upgrading from LiveKit Python SDK 0.X to 1.X, the logic that previously worked to start an agent with a specific participant (for outbound calls) is now broken or unsupported. I’m unsure how to pass the outbound_participant in the new version.
 Works in LiveKit 0.X
```
agent = VoicePipelineAgent(
    vad=ctx.proc.userdata["vad"],
    stt=STTService(stt_service_name, stt_language).stt_service,
    llm=LLMService(llm_service_name).llm_service,
    tts=TTSService(tts_service_name, tts_voice_id).tts_service,
    chat_ctx=initial_ctx,
    fnc_ctx=fnc_ctx,
    before_llm_cb=_enrich_with_rag,
    preemptive_synthesis=True,
    before_tts_cb=_predicted_text,
    noise_cancellation=noise_cancellation.BVC()
)

if source == 3:
    agent.start(ctx.room, outbound_participant)  # This worked in 0.X
else:
    agent.start(ctx.room)
```

In LiveKit 1.X, the new recommended way appears to be:
```
if source == 3:
    my_agent = OutboundCaller()
else:
    my_agent = Assistant()

session = AgentSession(
    stt=STTService(stt_service_name).stt_service,
    llm=LLMService(llm_service_name).llm_service,
    tts=TTSService(tts_service_name, tts_voice_id).tts_service,
    vad=silero.VAD.load(),
    max_tool_steps=1,
)

await session.start(
    room=ctx.room,
    agent=my_agent,
    room_input_options=RoomInputOptions(
        noise_cancellation=noise_cancellation.BVC()
    )
)
```

But I cannot figure out how to specify or pass outbound_participant to the agent in this new API.

⚙️ Expected Behavior
A way to attach a specific participant (like outbound_participant) when starting the agent, similar to what was possible in LiveKit 0.X.

longcw:
By default when `await session.start(agent=..., room=ctx.room)` it will create a RoomIO instance listen to the first participant joins the room, you can specify the participant listen in `RoomIO` explicitly
```python
    from livekit.agents import RoomIO

    room_io = RoomIO(
        agent_session=session,
        room=ctx.room,
        participant=target_participant,
        output_options=RoomOutputOptions(transcription_enabled=True),
    )
    await room_io.start()

    await session.start(agent=MyAgent())
```

[...]

PR #2037, opened 2025-04-18
avoid duplicate room io initialization

longcw:
to avoid error when RoomIO already created explicitly
```python
    room_io = RoomIO(session, room=ctx.room)
    await room_io.start()
    await session.start(
        agent=..., room=ctx.room
    )
```

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/room_io.py
+++ b/livekit-agents/livekit/agents/voice/room_io/room_io.py
@@ -112,7 +112,12 @@ class RoomIO:
             self._on_participant_connected(participant)
 
         if self._input_options.text_enabled:
-            self._room.register_text_stream_handler(TOPIC_CHAT, self._on_user_text_input)
+            try:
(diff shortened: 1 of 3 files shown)


longcw:
related to https://github.com/livekit/agents/issues/2034 and https://github.com/livekit/agents/issues/2018#issuecomment-2810223805

livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -112,7 +112,12 @@ async def start(self) -> None:
             self._on_participant_connected(participant)
 
         if self._input_options.text_enabled:
-            self._room.register_text_stream_handler(TOPIC_CHAT, self._on_user_text_input)
+            try:
```
davidzhao, line 115: warn on this is fine.. but can we check to skip room_io.start() if it's already started?

longcw: I added that in `session.start()` to skip the room io if there is already one.

PR #2052, opened 2025-04-21
tavus avatar plugin

longcw:


theomonnom (approved):

examples/avatar_agents/tavus/agent_worker.py
```diff
@@ -0,0 +1,44 @@
...
+    persona_id = os.getenv("TAVUS_PERSONA_ID")
+    replica_id = os.getenv("TAVUS_REPLICA_ID")
+    tavus_avatar = tavus.AvatarSession(persona_id=persona_id, replica_id=replica_id)
+    await tavus_avatar.start(session, room=ctx.room)
+
+    await session.start(
+        agent=Agent(instructions="Talk to me!"),
+        room=ctx.room,
+        # TODO: audio is forwarded to the avatar, so we disable room audio output
+        room_output_options=RoomOutputOptions(audio_enabled=True),
```
theomonnom, line 39: Should we ask Tavus to use our Datastream instead?

theomonnom: It also appears to be less flexible, as the audio can't be transmitted faster than real-time.

Issue #1966, opened 2025-04-10
Feature Request: Support for tool calling in llm adapters


jezell:
Language models use JSON Schema, MCP uses JSON Schema, OpenAPI uses JSON Schema, but livekit uses python functions. This creates a mismatch between the way tools actually interact with LLMs and the livekit API that makes it challenging to support things like MCP, OpenAPI, etc. without mapping the schema to a python function and back again as is done in the MCP sample here:

https://github.com/livekit-examples/basic-mcp

While this sample attempts to get the job done, it's full of all sorts of complicated code that shouldn't need to be written in the first place. It only exists because of the constraint that tools must be python functions, which is not an LLM native requirement, it's a LiveKit imposed requirement.

With the release of the Responses API, OpenAI has also added an alternative tool calling model for built in tools, and it's not clear how to plug those tools into a Livekit voice agent due to the way tool calling is implemented.

In a perfect world, the tool calling process would be something overridable / customizable by the llm adapter itself or a tool calling adapter, so someone could create an adapter that better integrates with native LLM capabilities or standards like MCP, instead of having to manage a lossy conversion to python functions and back again. At the moment, tool calling is implemented in a central set of functions that the llm adapter is not even involved in, making it very hard to customize tool calling to use functionality LLMs already support without forking the entire SDK:

https://github.com/livekit/agents/blob/4c3d980d287b93d1fb4417f35098e33f178d2128/livekit-agents/livekit/agents/voice/agent_activity.py#L1315

mnbbrown:
@theomonnom saw your PR 🙌  
Any chance raw function tools could support context - similar to normal function tools? Happy to put up a quick PR iterating on
https://github.com/elyosenergy/agents/blob/c839ac2fcbdf1b80f159095e4ea22f676aa1f05e/livekit-agents/livekit/agents/voice/generation.py#L341-L350

mnbbrown:
Couldn't help myself https://github.com/livekit/agents/pull/2073

theomonnom:
> [@theomonnom](https://github.com/theomonnom) saw your PR 🙌 Any chance raw function tools could support context - similar to normal function tools? Happy to put up a quick PR iterating on https://github.com/elyosenergy/agents/blob/c839ac2fcbdf1b80f159095e4ea22f676aa1f05e/livekit-agents/livekit/agents/voice/generation.py#L341-L350

Thanks, I moved your PR [here](https://github.com/livekit/agents/pull/2077)

[...]

PR #2055, opened 2025-04-21
support raw function descriptions 

theomonnom:
First step toward a more flexible tool abstraction for https://github.com/livekit/agents/issues/1966

Next steps are: support built-in OpenAI tools like web_search & add MCP support.

### DIFF
--- a/examples/voice_agents/raw_function_description.py
+++ b/examples/voice_agents/raw_function_description.py
@@ -1,0 +1,61 @@
+import logging
+
+from dotenv import load_dotenv
+
+from livekit.agents import Agent, AgentSession, JobContext, WorkerOptions, cli, function_tool
+from livekit.plugins import openai
+
+# This demo defines an agent using a raw function tool to open predefined gates via enum input.
+# When using raw function tools, compatibility across LLM providers is not guaranteed,
+# as different models may interpret or format raw schemas differently.
+
+logger = logging.getLogger("raw-function-description")
+
+load_dotenv()
+
+
+class RawFunctionAgent(Agent):
+    def __init__(self):
+        super().__init__(instructions="You are a helpful assistant")
+
+    @function_tool(
+        raw_schema={
+            "name": "open_gate",
+            "description": "Opens a specified gate from a predefined set of access points.",
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    "gate_id": {
+                        "type": "string",
+                        "description": (
+                            "Identifier of the gate to open. Must be one of the "
+                            "system's predefined access points."
+                        ),
+                        "enum": [
+                            "main_entrance",
+                            "north_parking",
+                            "loading_dock",
+                            "side_gate",
+                            "service_entry",
+                        ],
+                    }
+                },
+                "required": ["gate_id"],
+            },
+        }
+    )
+    async def open_gate(self, raw_arguments: dict[str, object]):
+        gate_id = raw_arguments["gate_id"]
+        logger.info(f"Opening gate: {gate_id}")
+        return f"Gate {gate_id} opened successfully"
+
+
+async def entrypoint(ctx: JobContext):
+    await ctx.connect()
+
+    session = AgentSession(llm=openai.realtime.RealtimeModel())
+    await session.start(RawFunctionAgent(), room=ctx.room)
+
+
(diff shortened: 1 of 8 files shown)


longcw (approved):

jezell:
Thanks @theomonnom, going to take this for a spin

examples/voice_agents/raw_function_description.py
```diff
@@ -0,0 +1,72 @@
...
+
+class RawFunctionAgent(Agent):
+    def __init__(self):
+        super().__init__(instructions="You are a helpful assistant")
+
+    @function_tool(
+        raw_schema={
+            "name": "open_gate",
+            "description": "Opens a specified gate from a predefined set of access points.",
+            "parameters": {
```
jezell, line 30: So this matches the OpenAI tool schema, but I think name / description / parameters here is a mistake. The schema is only really the part inside parameters, so while the parameter is called raw_schema, the actual schema is now in:

raw_schema.parameters rather than raw_schema.

It's a minor bone to pick, but compare to an MCP tool definition:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "tools": [
      {
        "name": "get_weather",
        "description": "Get current weather information for a location",
        "inputSchema": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "City name or zip code"
            }
          },
          "required": ["location"]
        }
      }
    ],
    "nextCursor": "next-page-cursor"
  }
}
```
https://modelcontextprotocol.io/specification/2025-03-26/server/tools

Maybe would be a little cleaner for name + description to be promoted to top level properties and raw_schema to be the actual json schema to make it less confusing to always have to add a wrapper around the json schema dict if you have a json schema.

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -104,18 +144,30 @@ def function_tool(
...
-        return cast(FunctionTool, func)
+    f: F | Raw_F | None = None,
+    *,
+    name: str | None = None,
+    description: str | None = None,
+    raw_schema: RawFunctionDescription | dict | None = None,
+) -> FunctionTool | RawFunctionTool | Callable[[F | Raw_F], FunctionTool | RawFunctionTool]:
+    def deco(func: F | Raw_F) -> RawFunctionTool | FunctionTool:
+        if raw_schema is not None:
+            if not raw_schema.get("name") or not raw_schema.get("parameters"):
```
jezell, line 155: See comments above. I think this code here illustrates the overlap, name + description is here in raw_schema despite being passed as params. So now we have two name + descriptions due to the raw_schema having a wrapper around the actual schema.

examples/voice_agents/raw_function_description.py
theomonnom: Yeah, that makes sense. My motivation for including the name and description inside raw_schema was mainly to stay 1:1 with the OpenAI docs.

But you're right — for other cases like MCP where the schema lives inside inputSchema, it adds confusion.

For MCP servers, we'll natively support them through the tools parameter of the Agent.

livekit-agents/livekit/agents/llm/tool_context.py
theomonnom: Since function_tool has 4 overloads, the name and description parameters shouldn't be considered valid when using raw_schema in the type checking.

I'm wondering if OpenAI might add more keys alongside name/description down the line?

There is already a `strict` key today, so it would need to be added to the @function_tool decorator if we go down this path.

jezell: @theomonnom  It's possible they could, but since different LLMs have slightly different ways of declaring and calling tools, internally we've always just had name, description, and schema and then compiled that into the form that's required by the LLM. I believe the OpenAI Agents SDK determines whether strict mode is supported on the fly for example by examining whether the schema itself supports strict mode, since strict mode enforces more constraints on the calls.

Sometimes we also want to do things like return the data used to call a tool as structured output from a LLM call in which case we need direct access to the schema, but structured output format is different than tool calling format, so it's nice to have the schema itself as a unique property and the other props as higher level props. Then you can adapt each of them as needed to the kind of use you require.

PR #2057, opened 2025-04-21
fix gemini live update chatcontext

jayeshp19:


livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -418,6 +434,15 @@ async def _main_task(self):
                 async with self._session_lock:
                     self._session = session
 
+                # for reconnect, send the updated chat context
+                if self._chat_ctx:
```
jayeshp19, line 438: if we're reconnecting while  calling `update_chat_ctx ` we need to send updated context.
this will resend chat context once for normal cases.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -155,19 +156,14 @@ def __init__(
...
+            gemini_api_key = None
         else:
+            if not gemini_api_key:
+                raise ValueError("API key required for Google API")
             gcp_project = None
             gcp_location = None
-            if not gemini_api_key:
-                raise ValueError(
-                    "API key is required for Google API either via api_key or GOOGLE_API_KEY environment variable"  # noqa: E501
-                )
```
theomonnom, line 170: I think the messages where better before?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -271,25 +273,37 @@ def update_options(
...
         async with self._update_chat_ctx_lock:
-            self._chat_ctx = chat_ctx
-            turns, _ = to_chat_ctx(self._chat_ctx, id(self), ignore_functions=True)
-            tool_results = get_tool_results_for_realtime(self._chat_ctx)
-            if turns:
-                self._msg_ch.send_nowait(LiveClientContent(turns=turns, turn_complete=False))
+            diff = llm.utils.compute_chat_ctx_diff(self._chat_ctx, chat_ctx)
+            if diff.to_remove:
+                self._chat_ctx = chat_ctx
+                self.reset_gemini_session()
```
theomonnom, line 287: I think it's OK to reconnect, but ideally, we use aiohttp with a connection pool

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
theomonnom: is there any synchronization if we're calling update_chat_ctx & the reconnection is happening? How do we know it'll not race and send two times?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -602,10 +625,14 @@ def _handle_go_away(self, go_away: LiveServerGoAway):
         )
 
     def commit_audio(self) -> None:
-        raise NotImplementedError("commit_audio_buffer is not supported yet")
+        raise NotImplementedError("commit_audio is not supported yet")
 
     def clear_audio(self) -> None:
         raise NotImplementedError("clear_audio is not supported yet")
 
     def server_vad_enabled(self) -> bool:
```
theomonnom, line 633: it was there before, but I believe this fnc is unused

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
```diff
@@ -258,6 +255,11 @@ async def _close_gemini_session(self) -> None:
                 finally:
                     self._session = None
 
+    def reset_gemini_session(self) -> None:
```
theomonnom, line 258: make it private

davidzhao:
superseded by #2401

PR #2063, opened 2025-04-21
Add Hume TTS Plugin

Saatvik07:
Add plugin for Hume AI's latest TTS model (Octave)
https://www.hume.ai/


livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
```diff
@@ -0,0 +1,422 @@
...
+class ChunkedStream(tts.ChunkedStream):
+    """Stream for Hume TTS JSON streaming API."""
+
+    def __init__(
+        self,
+        *,
+        tts: TTS,
+        input_text: str,
+        opts: _TTSOptions,
+        conn_options: APIConnectOptions,
```
theomonnom, line 255: Can we make sure the timeout inside conn_options is respected, otherwise the tests will not pass

theomonnom (commented):
Otherwise lgtm! This is awesome!

theomonnom (approved):
Awesome lgtm!

Issue #2040, opened 2025-04-18
Inconsistent transcript buffer behavior with start_turn / cancel_turn / end_turn flow (push-to-talk example)


well-balanced:
First of all, thank you for the amazing work on livekit-agents — we’ve been building a new product using it and have found the project incredibly useful and thoughtfully designed.

We’re currently using the `push_to_talk.md` example as the basis for our client integration. However, we’re encountering an issue with synchronizing the audio buffer across `start_turn`, `cancel_turn`, and `end_turn`. In particular, when users interact quickly (e.g., speaking immediately and cancelling early), the buffer state seems to get into an inconsistent state, which leads to unexpected transcripts being sent to the LLM.

## Example scenario
Here’s a simplified flow that demonstrates the issue:

1. `start_turn()` is called
2. User begins speaking: **“How are you today?”**
3. `cancel_turn()` is called immediately (this calls `session.clear_user_turn()` internally)
4. In the logs, we see:
```sh
livekit.agents - received user transcript {"user_transcript": "How are.", ...}
```
5. `start_turn()` is called again (again calls `session.clear_user_turn()`)
6. User says: **“My name is Wynn”**
7. `end_turn()` is called
8. In `on_user_turn_completed` method, the new_message is: **"you today my"**

So despite clearing the user turn twice, the transcript still contains parts of the previous utterance. This seems to happen especially when STT final transcripts arrive slightly after the cancel, leading to fragments from earlier input leaking into the final turn.

### Additional questions
Is there a best practice for handling this kind of flow?
We’d love to hear how others are dealing with this — whether there’s a canonical way to fully flush/reset the STT buffer or if additional manual logic is recommended on the app side.

Thanks again for the fantastic work on the agents — we really appreciate it :)

longcw:
Yeah I noticed this as well, if the user turn closed too fast right after the end of user speech, the STT may not return the final user transcription, causing two issues: 1) the transcription of this turn is not completed, 2) the rest text may be added to the next turn.

Basically these [two methods](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/voice/audio_recognition.py#L136-L147) `clear_user_turn`  and `commit_user_turn` need a fix, I am also looking for if there is any way to flush or reset the STT buffer.

well-balanced:
So for now, is adding a short await `asyncio.sleep(...)` after `end_turn` the best workaround to make sure STT has enough time to flush? Just wanna make sure I’m not missing a better way.

[...]

PR #2070, opened 2025-04-22
reset stt when clear user turn

longcw:


longcw:
related to https://github.com/livekit/agents/issues/2040

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -136,6 +136,7 @@ def update_vad(self, vad: vad.VAD | None) -> None:
     def clear_user_turn(self) -> None:
         self._audio_transcript = ""
         self._audio_interim_transcript = ""
+        self.update_stt(self._stt)  # reset stt
```
theomonnom, line 139: lgtm, can you add a comment on why this is needed
But I'm thinking that we might inadvertently break this in the future if we assume that no further operations should be processed when the same STT is set. To make this more explicit, can we call self.update_stt(None) before this line?

theomonnom (approved):

livekit-agents/livekit/agents/voice/audio_recognition.py
longcw: When cancel the stt task with `self.update_stt(None)` it may log an error
```bash
2025-04-22 16:42:23,125 - ERROR asyncio - _GatheringFuture exception was never retrieved
future: <_GatheringFuture finished exception=CancelledError()>
  File "/home/longc/data/code/agents/livekit-agents/livekit/agents/utils/log.py", line 16, in async_fn_logs
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/longc/data/code/agents/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py", line 536, in recv_task
    msg = await ws.receive()
          ^^^^^^^^^^^^^^^^^^
  File "/home/longc/data/code/agents/.venv/lib/python3.12/site-packages/aiohttp/client_ws.py", line 336, in receive
    msg = await self._reader.read()
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "aiohttp/_websocket/reader_c.py", line 113, in read
  File "aiohttp/_websocket/reader_c.py", line 110, in aiohttp._websocket.reader_c.WebSocketDataQueue.read
asyncio.exceptions.CancelledError {"pid": 2370362, "job_id": "AJ_VfJj5PVApi73"}
```

it's from the `@utils.log_exceptions(logger=logger)`, I think we can ignore it, is there a way to suppress that? I tried the following but doesn't work
```python
        with contextlib.suppress(asyncio.CancelledError):
            self.update_stt(None)
```

theomonnom: ah yeah that's bad, we should make sure to call the aclose cleanly

longcw: updated

PR #2071, opened 2025-04-22
store audio-synchronized transcript in chat_ctx when interrupted

theomonnom:
https://linear.app/livekit/issue/AGT-1138/add-the-audio-synced-transcript-to-the-playbackfinishedevent

### DIFF
--- a/livekit-agents/livekit/agents/voice/generation.py
+++ b/livekit-agents/livekit/agents/voice/generation.py
@@ -635,24 +635,3 @@ def remove_instructions(chat_ctx: ChatContext) -> None:
             chat_ctx.items.remove(msg)
         else:
             break
-
-
-STANDARD_SPEECH_RATE = 0.5  # words per second
-
-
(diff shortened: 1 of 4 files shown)


livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -635,24 +635,3 @@ def remove_instructions(chat_ctx: ChatContext) -> None:
             chat_ctx.items.remove(msg)
         else:
             break
-
-
-STANDARD_SPEECH_RATE = 0.5  # words per second
-
-
-def truncate_message(*, message: str, played_duration: float) -> str:
```
longcw, line 643: should we still have a simple truncate_message in case ppl is not using the RoomIO or the transcription sync is disabled?

Or even should we make the transcription sync a part of the agent instead of RoomIO? just thinking loudly

theomonnom: Yeah ideally the TranscriptionSync can be used everywhere (with the new IO API)

PR #2074, opened 2025-04-22
publish legacy transcription event for avatar

longcw:


livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -137,13 +141,24 @@ def set_participant(
...
+                if self._track_id is not None:
+                    break
+        if not self._track_id:
+            for p in self._room.remote_participants.values():
+                if not self._valid_participant(p):
+                    continue
+                for track in p.track_publications.values():
+                    self._on_track_published(track, p)
+                    if self._track_id is not None:
+                        break
```
theomonnom, line 161: early break is cool, but more LOCs tho :)

PR #2075, opened 2025-04-22
Bug Fix - wrong transition to SPEAKING when the agent is in fact listening

MajorTal:
The agent-state was falsely returned to SPEAKING right after switching to LISTENING.
Tested with the openai realtime model after patching the code locally.

I have timers in my app relying on agent-state so this is a critical bug for me.


theomonnom:
lgtm!

MajorTal:
Wow that was fast! Kudos!

PR #2078, opened 2025-04-22
Bug fix in agent_activity.py: user_state events were not fired

MajorTal:
For the realtime model, the user state events:
self._session._update_user_state("listening")
and
self._session._update_user_state("speaking")
were not fired at all.
They were correctly fired in response to VAD events in on_start_of_speech and on_end_of_speech

This is a 2-liner fix: just firing the event at the right time.

Critical bug for me because my app relies on these events.


theomonnom (approved):
thanks lgtm!

PR #2081, opened 2025-04-22
Revert "make semantic vad default for openai realtime and stt"

theomonnom:
Reverts livekit/agents#2028

### DIFF
--- a/.github/next-release/changeset-c349be13.md
+++ b/.github/next-release/changeset-c349be13.md
@@ -1,5 +1,0 @@
----
-"livekit-plugins-openai": patch
----
-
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
is this due to semantic_vad not being supported on azure?

longcw (approved):

davidzhao:
@yuyuma this is for compatibility with Azure, which does not support semantic vad yet

longcw:
@yuyuma another reason is that `server_vad` is still the default in OpenAI realtime API https://platform.openai.com/docs/api-reference/realtime-sessions/create#realtime-sessions-create-turn_detection

PR #2090, opened 2025-04-23
add bithuman deps for the plugin

longcw:


bcherry (approved):

bcherry:
@longcw something is also wrong with this plugin in generated docs https://docs.livekit.run/reference/python/v1/livekit/plugins/bithuman.html

PR #2095, opened 2025-04-23
fix interruption context when the speech hasn't started yet

theomonnom:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1092,7 +1092,11 @@ def _on_first_frame(_: asyncio.Future) -> None:
 
                 msg = chat_ctx.add_message(
                     role="assistant",
-                    content=playback_ev.synchronized_transcript or text_out.text,
+                    content=(
```
davidzhao, line 1095: should it add an empty message in this case?

davidzhao (approved):

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1092,7 +1092,11 @@ def _on_first_frame(_: asyncio.Future) -> None:
...
+                        playback_ev.synchronized_transcript
+                        if playback_ev.synchronized_transcript is not None
```
longcw, line 1097: This didn't fix the issue, when the response is interrupted before any audio playout, `playback_ev.synchronized_transcript` may contain the message from the previous response.

longcw (approved):

PR #2096, opened 2025-04-23
fix using raw tools with ChatContext copies

zwily:
Previously raw function tools definitions only worked when specified as class members. When specifying them directly when instantiating an Agent, it would fail because ChatContext#copy didn't know how to handle them.


theomonnom (approved):
lgtm! thanks!

PR #2101, opened 2025-04-24
feat: video input sampling

davidzhao:
allows the user to specify a target FPS to sample incoming frames.

### DIFF
--- a/livekit-agents/livekit/agents/utils/__init__.py
+++ b/livekit-agents/livekit/agents/utils/__init__.py
@@ -8,6 +8,7 @@ from .log import log_exceptions
 from .misc import is_given, shortuuid, time_ms
 from .moving_average import MovingAverage
 from .participant import wait_for_participant
+from .video import SamplingVideoStream, VideoFPSSampler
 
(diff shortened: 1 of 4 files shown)


theomonnom:
Shouldn't this be an option with Gemini vs RoomIO?

I'm wondering because you may want to do some other stuff with the video frames before they're dropped

longcw:
> Shouldn't this be an option with Gemini vs RoomIO?
> 
> I'm wondering because you may want to do some other stuff with the video frames before they're dropped

agreed

davidzhao:
chatted with Theo a bit.. and I think this should be more general instead of at the model layer. when a different model is used (say a stateless multimodal LLM), we would want the sampling logic to work the same way.

I also agree RoomIO probably isn't the best spot.. because it should be input agnostic too, and ideally user could update it depending on the state change (i.e. 1fps while idle, 3fps while user is speaking).

given that we've decided to move this into AgentSession, and theo will be taking over

theomonnom:
https://github.com/livekit/agents/pull/2113

PR #2102, opened 2025-04-24
fix transcription truncate and apply for realtime model

longcw:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1080,35 +1080,33 @@ def _on_first_frame(_: asyncio.Future) -> None:
         if speech_handle.interrupted:
             await utils.aio.cancel_and_wait(*tasks)
 
+            forwarded_text = text_out.text
```
theomonnom, line 1083: should we simplify by starting the state to `""`

This would simplify the branch below, (just moving the else branch to ` if playback_ev.synchronized_transcript is not None:`

nit tho

theomonnom (approved):
lgtm! thanks!

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1368,52 +1368,56 @@ async def _read_messages(
...
 
         speech_handle._mark_playout_done()  # mark the playout done before waiting for the tool execution  # noqa: E501
 
-        msg_id, text_out, _ = message_outputs[0]  # there should be only one message
-        msg = llm.ChatMessage(
-            role="assistant", content=[text_out.text], id=msg_id, interrupted=False
-        )
-        self._agent._chat_ctx.items.append(msg)
-        self._session._conversation_item_added(msg)
+        if len(message_outputs) > 0:
```
theomonnom, line 1413: Should we warn if there is more than one message, so in case it happens one day if OAI changes their API, we're quickly aware

longcw: there is already a warning above and it will break from the loop
```python
                   if len(forward_tasks) > 0:
                        logger.warning(
                            "expected to receive only one message generation from the realtime API"
                        )
                        break
```

longcw: so there should never have more than one message in the list

PR #2110, opened 2025-04-24
add an injected silence filler short response example 

theomonnom:


examples/voice_agents/fast-preresponse.py
```diff
@@ -45,6 +45,16 @@ async def on_user_turn_completed(self, turn_ctx: ChatContext, new_message: ChatM
             add_to_chat_ctx=False,
         )
 
+        # Alternatively, if you want the reply to be aware of this "silence filler" response,
+        # you can await the say method—but the tradeoff is that it may be slower since it won’t
+        # execute concurrently with the main response generation:
+        #
+        # speech_handle = await self.session.say(
```
longcw, line 52: The next llm is supposed to be starting after the first llm done but not after the audio played out, otherwise the silence before the second llm is still there.

theomonnom: That's a good point, we could listen to the llm.chat() separately from the say method.

Maybe we could add another method to the SpeechHandle that just wait for the generation to complete.

theomonnom:
Merging rn, the above can be a future improvement, it isn't a blocker

PR #2121, opened 2025-04-25
Set Hume SDK version to 0.8.3 or greater

bcherry:
This resolves a dependency conflict on websockets when used with the google or openai plugins.


davidzhao (approved):

bcherry:
@davidzhao unfortunately this can't be merged at all, since the CI installs all the packages and therefore triggers the underlying conflict

davidzhao:
> @davidzhao unfortunately this can't be merged at all, since the CI installs all the packages and therefore triggers the underlying conflict

I see your edit now.. we'll hold off and wait for the Hume team then

theomonnom (approved):

bcherry:
They released 0.8.3 which now requires websockets 13.1 instead and that has resolved the issues. I bumped this to the same.

PR #2127, opened 2025-04-25
support video input from screenshare

longcw:
1. always use the latest published video track (camera or screenshare), close the screenshare will switch back to camera if any
2. close and reopen (mute or unmute) the camera when it reads from the screenshare it won’t switch to the camera until the screen share stopped


bcherry (commented):
makes sense to me, thanks for adding this!

livekit-agents/livekit/agents/voice/room_io/_input.py
```diff
@@ -50,7 +58,7 @@ def on_attached(self) -> None:
             "input stream attached",
             extra={
                 "participant": self._participant_identity,
-                "source": rtc.TrackSource.Name(self._track_source),
```
davidzhao, line 53: I think it's fine to keep the singular `source`. for consistency when using Datadog or other search tools, it's far easier to search for `source` always, instead of trying to figure out to specify sources or source

davidzhao (approved):
lg

PR #2128, opened 2025-04-25
fix conversion of chat context to dict

jmugicagonz:
If `exclude_defaults` is `True`, the type won't be included in the dict and so you won't be able to convert the dict back to chat context


theomonnom:
Thanks!

PR #2129, opened 2025-04-25
support session resumption and language in gemini live

jayeshp19:
- https://ai.google.dev/gemini-api/docs/live#session-resumption

- https://ai.google.dev/gemini-api/docs/live#supported-languages


jayeshp19:
@davidzhao @theomonnom  PTAL

davidzhao (approved):
lgtm! thanks

aatreya:
It looks like this session resumption change is incomplete - warnings appear that the session will end, but it no resumption actually occurs and the Gemini agent drops.

PR #2138, opened 2025-04-26
Use OpenAI request ID for TTS emitter

anishnag:
We use the OpenAI request ID for TTS instead of random UUID.


theomonnom:
Hey, I've been handling it inside this [PR](https://github.com/livekit/agents/pull/2152).

We changed the way we emit the synthesized audio, so it's much easier to do it (no need for asyncio primitives)
https://github.com/livekit/agents/blob/285883ab07c0d5cc02d0e95212e46b04aadb0f24/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/tts.py#L200

PR #2139, opened 2025-04-26
Use PCM instead of OGG_OPUS for Google TTS

chrisackermann:
Per [Google's guidance](https://github.com/GoogleCloudPlatform/generative-ai/issues/1924#issuecomment-2830994746), use `PCM` instead of `OGG_OPUS` for audio encoding when using streaming for TTS.


davidzhao (approved):
lgtm! def appears to reduce TTFB. great find

manishkjs1:
@davidzhao - How much latency improvement you saw? Can you please share few numbers?

I just saw through this merge and it seems the encoding type has been updated (only). 

The [PR](https://github.com/GoogleCloudPlatform/generative-ai/issues/1924#issuecomment-2830994746) @jayeshp19 is working through is to introduce streaming into TTS. That is supposed to bring down latency into 100-200ms range. 
That is still pending. Can that be expedited please?

davidzhao:
the latency went from 1+s to 300ms. we can definitely accelerate efforts on streaming. do you have any examples on using that API?

manishkjs1:
Great to hear latency numbers @davidzhao , on streaming @jayeshp19 is wrapping up work and shortly going to put a PR.

jayeshp19:
I've created this PR https://github.com/livekit/agents/pull/2143,  this improves latency a bit but to fully support streaming,  i think we need to persist SynthesisStream

danishshaik:
@chrisackermann @davidzhao what voices were you guys using for this?  Seems like i'm getting 

livekit.agents._exceptions.APIStatusError: This voice currently only supports ALAW, LINEAR16, MULAW, MP3 and OGG_OPUS output. (status_code=400, request_id=None, body=None) {"tts": "livekit.plugins.google.tts.TTS", "attempt": 2, "streamed": false, "pid": 31836, "job_id": "AJ_xTbpV4gK3xgb"}

using Chirp3 Aoede?

Also this example: https://github.com/GoogleCloudPlatform/generative-ai/blob/main/audio/speech/getting-started/get_started_with_chirp3_instant_custom_voice.ipynb

uses LINEAR16

PR #2141, opened 2025-04-27
Improve decoding speed for wav

davidzhao:
Certain TTS (like Rime) would stream data back close to realtime. In those cases, it could increase the time it takes to receive the first frame due to libav decoder being conservative about buffering content. Switching to a streaming wav decoder to improve TTFB (Rime went from 1s to 300ms)

### DIFF
--- a/livekit-agents/livekit/agents/utils/codecs/decoder.py
+++ b/livekit-agents/livekit/agents/utils/codecs/decoder.py
@@ -182,7 +182,7 @@ class AudioStreamDecoder:
 
     def _decode_wav_loop(self):
         """Decode wav data from the buffer without ffmpeg, parse header and emit PCM frames.
-        
+
         This can be much faster than using ffmpeg, as we are emitting frames as quickly as possible.
         """
 
@@ -220,7 +220,6 @@ class AudioStreamDecoder:
                     )
                     if audio_format != 1:
                         raise ValueError(f"Unsupported WAV audio format: {audio_format}")
-                    sampwidth = bits_per_sample // 8
                     fmt_chunk_found = True
 
             # parse data chunk
--- a/livekit-agents/livekit/agents/utils/codecs/decoder.py
+++ b/livekit-agents/livekit/agents/utils/codecs/decoder.py
@@ -197,7 +197,7 @@ class AudioStreamDecoder:
                     raise ValueError("Invalid WAV file: incomplete header")
                 header += chunk
             if header[:4] != b"RIFF" or header[8:12] != b"WAVE":
-                raise ValueError("Invalid WAV file: missing RIFF/WAVE")
+                raise ValueError(f"Invalid WAV file: missing RIFF/WAVE: {header}")
 
             # parse fmt chunk
             fmt_chunk_found = False
--- a/livekit-agents/livekit/agents/utils/codecs/decoder.py
+++ b/livekit-agents/livekit/agents/utils/codecs/decoder.py
@@ -94,7 +94,7 @@ class AudioStreamDecoder:
     _executor: Optional[ThreadPoolExecutor] = None
 
     def __init__(
-        self, *, sample_rate: int = 48000, num_channels: int = 1, format: str | None = None
+        self, *, sample_rate: int = 48000, num_channels: int = 1, format: Optional[str] = None
     ):
         self._sample_rate = sample_rate
         self._layout = "mono"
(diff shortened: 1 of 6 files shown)


theomonnom:
Why do we need to use libav for wav?

livekit-agents/livekit/agents/utils/codecs/decoder.py
```diff
@@ -92,13 +93,16 @@ class AudioStreamDecoder:
     _max_workers: int = 10
     _executor: Optional[ThreadPoolExecutor] = None
 
-    def __init__(self, *, sample_rate: int = 48000, num_channels: int = 1):
+    def __init__(
+        self, *, sample_rate: int = 48000, num_channels: int = 1, format: str | None = None
```
theomonnom, line 97: Can we remove the format? and auto-detect?

davidzhao: i don't believe we can very easily, it'll make it pretty messy to feed it to ffmpeg containers after already consuming part of the input

livekit-agents/livekit/agents/utils/codecs/decoder.py
```diff
@@ -126,7 +135,22 @@ def _decode_loop(self):
...
+                    "fflags": "nobuffer+flush_packets",
+                    "probesize": "32",
+                    "analyzeduration": "0",
+                    "max_delay": "0",
+                },
+            )
+            # explicitly disable internal buffering flags on the FFmpeg container
+            container.flags |= (
+                av.container.Flags.no_buffer.value | av.container.Flags.flush_packets.value
+            )
```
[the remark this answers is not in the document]
davidzhao: no, this does not solve the issue for wav

livekit-agents/livekit/agents/utils/codecs/decoder.py
theomonnom: It's unclear when you need to provide a format vs not?

davidzhao:
> I have a local branch, where I don't use the AudioStreamDecoder, but I just passthrough on raw PCM & wav

the idea is to have an abstraction for how to convert anything -> PCM. I would rather not leave code like that in the plugins

livekit-agents/livekit/agents/utils/codecs/decoder.py
davidzhao: it's a good practice to provide the format because it's known. and plugins definitely know the format because they have to ask the API to return content as such

theomonnom: can we use the format for all plugins? the format isn't passed into av.open

davidzhao: it doesn't add immediate value.. that's why it's optional. this only matters when it's wav.. that's the point of this PR.

livekit-agents/livekit/agents/utils/codecs/decoder.py
```diff
@@ -156,6 +180,96 @@ def _decode_loop(self):
...
+                        raise ValueError("Invalid WAV file: incomplete fmt chunk data")
+                    data += part
+                    remaining -= len(part)
+                if chunk_id == b"fmt ":
+                    audio_format, wave_channels, wave_rate, _, _, bits_per_sample = struct.unpack(
+                        "<HHIIHH", data[:16]
+                    )
+                    if audio_format != 1:
+                        raise ValueError(f"Unsupported WAV audio format: {audio_format}")
+                    fmt_chunk_found = True
```
theomonnom, line 223: break instead

livekit-agents/livekit/agents/utils/codecs/decoder.py
```diff
@@ -156,6 +180,96 @@ def _decode_loop(self):
...
+
+            # parse data chunk
+            data_chunk_found = False
+            while not data_chunk_found:
+                sub_header = self._input_buf.read(8)
+                if len(sub_header) < 8:
+                    raise ValueError("Invalid WAV file: incomplete data chunk header")
+                chunk_id, chunk_size = struct.unpack("<4sI", sub_header)
+                if chunk_id == b"data":
+                    data_chunk_found = True
```
theomonnom, line 233: break instead

theomonnom (approved):
I see, it's still not great on how we handle those differences.. raw pcm is also being passed to av.open when it isn't wav

PR #2151, opened 2025-04-28
rewrite Azure TTS

theomonnom:
the Azure SDK is shitty


A-n-a-s-A-b-r-a-r:
why the azure TTS event like on_viseme_event used for lyp sync and some others have been removed ?

PR #2152, opened 2025-04-28
TTS improvements & tests 

theomonnom:
- This fixes all synthesize tests 
- Reduced TTS chunked stream LOC
    * improved the SynthesizedOutputEmitter
- use resp.raise_for_status() ....
- make sure to raise an exception when the TTS succeeded but no frame has been pushed
- print metrics on the tests
- add retry tests & EventCollector
- the tests now make sure the returned frame size are appropriate
- always copy the TTSOptions inside the ChunkedStream
- Fixed the timeout for all TTS (Azure on https://github.com/livekit/agents/pull/2151, as it required a complete rewrite) 
- `LK_DUMP_TTS` will save tts output to cwd/lk_dump/
- rewrite HumeTTS, PlayAI TTS & Neuphonic TTS
- removed ws streaming from PlayAI & Neuphonic (temporary while we fix the tests) 

And a lot of bad code & inconsistencies (NOT_GIVEN..., different parameters from baseclass, ...)


theomonnom:
@claude What is this PR doing

PR #2155, opened 2025-04-29
cleanup examples + e2ee & stats examples

theomonnom:


theomonnom:
> Note that this example is the only one describing how to transcribe all the participants within a room. The usual approach of subscribing to "lk.transcriptions" works only transcribes the first user who joins the room. Defining Agent() with stt also emits transcription events for the first user who joins the room. Ideally, we would like to have enable_multi_transcription=true instead of just enable_transcription=True on the room.

That's a good point, our current example is quite outdated, I'll work on a new one before merging this PR

PR #2156, opened 2025-04-29
read frames from pre-connect audio buffer

longcw:


livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
```diff
@@ -0,0 +1,77 @@
...
+            self._main_atask = asyncio.create_task(self._read_audio_task(reader, participant_id))
+
+        self._room.register_byte_stream_handler(PRE_CONNECT_AUDIO_BUFFER_STREAM, _on_audio_buffer)
+        self._data.received_fut.add_done_callback(
+            lambda _: self._room.unregister_byte_stream_handler(PRE_CONNECT_AUDIO_BUFFER_STREAM)
+        )
+        return self._data
+
+    @utils.log_exceptions(logger=logger)
+    async def _read_audio_task(self, reader: rtc.ByteStreamReader, participant_id: str):
```
theomonnom, line 48: Let's make sure we're reading from the right participant_id? The one that is registered inside the RoomIO

examples/voice_agents/precompute_audio_buffer.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def prewarm(proc: JobProcess):
+    proc.userdata["vad"] = silero.VAD.load()
+
+
+async def entrypoint(ctx: JobContext):
+    # register a pre-connect audio handler before connecting to the room so that
+    # it won't miss the audio buffer (TODO (long): does this makes sense?)
+    pre_connect_audio = PreConnectAudioHandler(ctx.room).register()
```
longcw, line 35: @theomonnom we can do that automatically if the line can run after the `ctx.connect()`, basically register the byte stream handler inside the room io after the `ctx.connect()`. I am not sure if we are going to miss the stream reader if the ctx connect will auto subscribe the audio track.

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
longcw: we don't know which participant listening to at this moment, so maybe we buffer all the pre-connect audios for participants and decide if we want to use them based on the participant id and the received timestamp.

Imagine in a multi-user room, the second user joined the room with the pre-connect audio but the agent is not listening to that user until the `set_participant` is called, then that pre-connect buffer was actually out of date we should ignore.

pblazej:
Works fine for me (in a single participant scenario) 👍

bcherry:
yeah i think this needs to be automatic from the agent side so the developer only has to "turn it on" in one spot (which would be on the client).  right now the client sets a participant attribute indicating this feature is enabled but I think we're going to move that to be a track feature instead. ideally the agents framework can read that and automatically do the right thing

examples/voice_agents/precompute_audio_buffer.py
longcw: @bcherry to enable it automatically, is that okay to register the handler after `ctx.connect`, what will happen if the handler is registered after the byte stream sent?

longcw: My concern is that if we move the registering to somewhere like room io, user may do anything between the `ctx.connect` and `session.start()`, the handler may not be there when the audio is sent, also there might be a gap between the audio track subscribed (done in ctx.connect) and audio stream created from the track (in `room_io.start()`).

(If the pre-connect buffer is sent until the audio track got subscribed.)

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
longcw: fixed for multiple participants

pblazej:
Token detection seems to work fine 👍 I'm leaving the final decision here to @bcherry (more moving parts)

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
```diff
@@ -0,0 +1,116 @@
...
+from collections.abc import Coroutine
+from dataclasses import dataclass, field
+from typing import Any, Callable
+
+from livekit import rtc
+
+from ..agent import logger, utils
+
+PRE_CONNECT_AUDIO_BUFFER_STREAM = "lk.agent.pre-connect-audio-buffer"
+PRE_CONNECT_AUDIO_ATTRIBUTE = "lk.agent.pre-connect-audio"
```
lukasIO, line 13: we discussed this in the client team and the most reliable solution would be to use the `audioTrackFeature` enum on the publication itself to figure out if the preconnect buffer should be handled for that track

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
```diff
@@ -0,0 +1,116 @@
...
+_WaitPreConnectAudio = Callable[[rtc.RemoteParticipant], Coroutine[Any, Any, list[rtc.AudioFrame]]]
+
+
+class PreConnectAudioHandler:
+    def __init__(self, room: rtc.Room, *, timeout: float, max_delta_s: float = 1.0):
+        self._room = room
+        self._timeout = timeout
+        self._max_delta_s = max_delta_s
+
+        self._buffers: dict[str, asyncio.Future[_PreConnectAudioBuffer]] = {}
```
lukasIO, line 31: instead of a dictionary with participant identity keys, it would make more sense to use the track id. We don't know the publication id yet on the client side, but the mediastreamtrack id should work, I think

lukasIO (changes requested):
Please hold this for a moment, until we can get protocol level support for the pre connect buffer as part of the audio track features

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
pblazej: @longcw are you fine with this alternative?

longcw: I think it's okay, how to read that?

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
longcw: I don't think it's needed to distinguish the tracks from the same participant, we always read the first audio track from the participant.

longcw: The track id in transcription makes it's hard to use IMO, I don't know if it's worth to add it here

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
pblazej:
```python
class TrackInfo(_message.Message):
...
    audio_features: _containers.RepeatedScalarFieldContainer[AudioTrackFeature]
```
but the exact case hasn't been added yet, am I right @lukasIO?

lukasIO: that's correct, here's the [PR](https://github.com/livekit/protocol/pull/1057) to add it

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
lukasIO: in what way does it make it harder to use?

longcw:
separate the room io updates to https://github.com/livekit/agents/pull/2167

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
pblazej: I think it is mapped, the snippet above is copied from `models.pyi`

longcw:
separated the change in RoomIO into https://github.com/livekit/agents/pull/2167, created a new one for this feature https://github.com/livekit/agents/pull/2171 (basically reverted a few commits)

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
lukasIO: just so that this doesn't get lost after switching PRs, what's your thinking here @longcw ?

longcw: I don't against it, but let's see what is the new protocol... for example if the buffer is bound with a track id, then it's fine to use track id as the key

bcherry: protocol PR  was merged btw https://github.com/livekit/protocol/pull/1057

PR #2158, opened 2025-04-29
Added ability to set custom Rime URL

milo157:
This PR allows a user to override the default Rime URL in the case of using self-hosted deployments

### DIFF
--- a/livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
+++ b/livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
@@ -66,9 +66,6 @@ class _Mistv2Options:
     phonemize_between_brackets: NotGivenOr[bool] = NOT_GIVEN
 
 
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
```diff
@@ -66,16 +66,14 @@ class _Mistv2Options:
     phonemize_between_brackets: NotGivenOr[bool] = NOT_GIVEN
 
 
-DEFAULT_API_URL = "https://users.rime.ai/v1/rime-tts"
```
MaCaki, line 69: wdyt of keeping this as a const at the top of file?

milo157: I made it a default value in the init - so it keeps the same format as setting the default model

MaCaki (approved):

davidzhao (approved):
lg

PR #2159, opened 2025-04-29
Add mip_opt_out option for Deepgram TTS

MartinGotelli:
Add support for the mip_opt_out boolean of Deepgram API (https://developers.deepgram.com/docs/the-deepgram-model-improvement-partnership-program). Similar to the supported STT value

### DIFF
--- a/.github/next-release/changeset-4ae759b0.md
+++ b/.github/next-release/changeset-4ae759b0.md
@@ -2,4 +2,4 @@
 "livekit-plugins-deepgram": patch
 ---
 
-Add min_opt_out option for Deepgram TTS (#2159)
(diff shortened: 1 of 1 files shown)


theomonnom:
Is it expected that the name is different from the STT?

On the STT it is: `mip_opt_out`, not `min_opt_out`

MartinGotelli:
> Is it expected that the name is different from the STT?
> 
> On the STT it is: `mip_opt_out`, not `min_opt_out`

Oh shoot, that's what happens when you code without copy-pasting property names. Will fix it

theomonnom (approved):
Thanks!

Issue #1655, opened 2025-03-14
ElevenLabs Scribe support


rinarakaki:
I request ElevenLabs Scribe support as `elevenlabs.stt.STT` plugin.
 
- https://elevenlabs.io/blog/meet-scribe
- https://elevenlabs.io/speech-to-text

jayeshp19:
Hey Thanks for creating issue, we will support it soon

firattamur:
Hello,

Is there any updates for this ?

Thanks,

PR #2161, opened 2025-04-29
ElevenLabs Scribe support #1655

nishadmusthafa:
Adding support for elevenlabs scribe. Used documentation on https://elevenlabs.io/docs/api-reference/speech-to-text/convert. They unfortunately don't support streaming yet so went with a similar approach as fal.


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -0,0 +1,127 @@
...
+                "set ELEVEN_API_KEY environmental variable"
+            )
+        self._opts = _STTOptions(
+            api_key=elevenlabs_api_key,
+            base_url=base_url if is_given(base_url) else API_BASE_URL_V1,
+            language_code=language_code,
+        )
+        self._session = http_session
+
+    def _ensure_session(self) -> aiohttp.ClientSession:
```
nishadmusthafa, line 79: This function seems to be repeated across various plugins. I was wondering if it made sense to make a BasePlugin class where this and any other similar methods could reside.

nishadmusthafa:
Still looking at adding tests

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
theomonnom: agree 100%, we should cleanup this!

theomonnom (approved):
Thanks! This is awesome!

No worries about tests for now, we're currently reworking them, this is our p0

PR #2162, opened 2025-04-30
Zg/hume plugin tweaks

zgreathouse:
**Update defaults and accepted parameters for Hume Plugin**
- Removes `sample_rate` parameter and updates default sample rate to `48000` (Hz). 
    - The Hume API does not currently support setting the sample rate, and only outputs audio with a sample rate of `48000` Hz.
- Removes the `num_generations` parameter, allowing for the API to default to `1`. 
    - When more than one generation is specified all generations are streamed at once (e.g., `gen1_chunk1`, `gen2_chunk1`, `gen1_chunk2`, `gen2_chunk2`...etc.)
- Removes `split_utterances` parameter, allowing for the API to default to `true`. 
    - Since the audio is being fed directly to the emitter, and only a single utterance is specified per request, this argument will make no difference to the end user.
- Simplifies the API surface area of the plugin to reflect only a single `utterance` is specified on each request. 
    - Removes `utterance_options` parameter in favor of making `voice`, `description`, and `speed` each individual parameters.
    - Removes `trailing_silence` parameter since `trailing_silence` is an API field for injecting pauses between utterances when multiple utterances are provided.

### DIFF
--- a/livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/__init__.py
+++ b/livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/__init__.py
@@ -19,10 +19,10 @@ __version__ = "1.0.0"
 # make imports available
 from hume.tts import (
     Format,
+    FormatWav,
     PostedContext,
     PostedUtterance,
-    PostedUtteranceVoiceWithId,
-    PostedUtteranceVoiceWithName,
+    PostedUtteranceVoice,
 )
 from livekit.agents import Plugin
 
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/__init__.py
```diff
@@ -32,10 +32,10 @@
 __all__ = [
     "TTS",
     "Format",
-    "PostedUtterance",
+    "FormatWav",
     "PostedContext",
-    "PostedUtteranceVoiceWithName",
-    "PostedUtteranceVoiceWithId",
+    "PostedUtterance",
```
bcherry, line 37: looks like we can probably skip exporting this one now

bcherry (approved):
thank you for this improvement!

livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/__init__.py
zgreathouse: @bcherry Updated the exports to only include types used for `_TTSOptions`.

theomonnom (approved):
Hey, this lgtm! thanks!

PR #2163, opened 2025-04-30
add hedra plugin

longcw:


brandon-halim:
This is the full url for the production endpoint @longcw 

`https://api.hedra.com/public/livekit/v1/session`

PR #2171, opened 2025-05-01
add pre-connected audio buffer

longcw:
replace https://github.com/livekit/agents/pull/2156


livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -123,6 +128,15 @@ def __init__(
         self._tasks: set[asyncio.Task] = set()
         self._update_state_task: asyncio.Task | None = None
 
+        self._pre_connect_audio_handler: PreConnectAudioHandler | None = None
+        if input_options.pre_connect_audio:
+            self._pre_connect_audio_handler = PreConnectAudioHandler(
+                room=self._room,
+                timeout=input_options.pre_connect_audio_timeout,
+            )
+            # TODO: move this to start()
```
theomonnom, line 137: can you do it in this PR?

add a warning if the room is already connected

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
```diff
@@ -0,0 +1,117 @@
...
+            sample_rate = int(reader.info.attributes["sampleRate"])
+            num_channels = int(reader.info.attributes["channels"])
+            logger.debug(
+                "pre-connect audio connected",
+                extra={
+                    "sample_rate": sample_rate,
+                    "num_channels": num_channels,
+                    "participant": participant_id,
+                },
+            )
```
theomonnom, line 95: I would remove this log, we're not very verbose for other stuff, it doesn't seems important

theomonnom: I have a PR where I have a LK_DEBUG=1 or LK_DEBUG=2, ... env var

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
```diff
@@ -0,0 +1,117 @@
...
+                        "pre-connect audio buffer is too old",
+                        extra={"participant": participant_identity, "delta_time": delta},
+                    )
+                    return []
+                return buf.frames
+
+            buf = await asyncio.wait_for(fut, self._timeout)
+            return buf.frames
+        finally:
+            async with self._lock:
```
theomonnom, line 72: The lock isn't needed as far as I can tell

livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -70,6 +71,10 @@ class RoomInputOptions:
     participant_identity: NotGivenOr[str] = NOT_GIVEN
     """The participant to link to. If not provided, link to the first participant.
     Can be overridden by the `participant` argument of RoomIO constructor or `set_participant`."""
+    pre_connect_audio: bool = False
+    """Pre-connect audio enabled or not."""
+    pre_connect_audio_timeout: float = 5.0
+    """The pre-connect audio will be ignored if it doesn't arrive within this time."""
```
theomonnom, line 77: Can we remove the timeout? ideally we know if we should wait for it based on the attributes 
```suggestion
```

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
```diff
@@ -0,0 +1,117 @@
...
+        self._tasks: set[asyncio.Task] = set()
+
+    def register(self):
+        def _handler(reader: rtc.ByteStreamReader, participant_id: str):
+            task = asyncio.create_task(self._read_audio_task(reader, participant_id))
+            self._tasks.add(task)
+            task.add_done_callback(self._tasks.discard)
+
+        if self._room.isconnected():
+            logger.warning("pre-connect audio handler registered after room is connected")
```
theomonnom, line 42: let's move to the start method of room_io, and explain this is due to the pre_connect_audio=True

livekit-agents/livekit/agents/voice/room_io/_input.py
```diff
@@ -202,13 +204,15 @@ def __init__(
         sample_rate: int,
         num_channels: int,
         noise_cancellation: rtc.NoiseCancellationOptions | None,
+        pre_connect_audio_cb: _WaitPreConnectAudio | None,
```
theomonnom, line 207: can't we just directly give the PreConnectAudioHandle? This is harder to read knowing this will always be the wait_for_data function anyway

livekit-agents/livekit/agents/voice/room_io/room_io.py
longcw: the byte stream is not stable in my test, a short timeout is needed IMO. otherwise user will need to wait for tens of seconds to get the response from the agent. I think it's better expose this to user if they want to decrease or increase the value based on needs.

livekit-agents/livekit/agents/voice/room_io/_input.py
longcw: It could be, I was think maybe this can be a different method

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
longcw: It's hard to know the buffer is used or not without any logs? I also saw the case there was the byte stream received but never closed. We will never know that happened without log...

livekit-agents/livekit/agents/voice/room_io/room_io.py
longcw: yeah I was about to move it to room io after https://github.com/livekit/agents/pull/2167, to avoid a conflict

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
longcw: not needed here or for the whole buffers? and why, the byte stream handler will also write to the buffers

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
theomonnom: we can add warnings for the timeout?

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
theomonnom: There is no async ops inside all locks, so it isn't needed everywhere

livekit-agents/livekit/agents/voice/room_io/room_io.py
theomonnom: Right, my intuition is that it should just work OOTB, and this value is just internal

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
longcw: sg, replaced this one with a timeout warning

livekit-agents/livekit/agents/voice/room_io/room_io.py
longcw: updated

livekit-agents/livekit/agents/voice/room_io/_input.py
theomonnom: When will it be a different method?

livekit-agents/livekit/agents/voice/room_io/room_io.py
bcherry: (but also a short timeout of just a few seconds is reasonable. if it's been more than one second since the agent subscribed to the track and the buffer still hasn't come back then the better UX is to discard it and just start listening to the track itself)

longcw: some times I cannot receive it when using the cloud server, probably bc the network issue

livekit-agents/livekit/agents/voice/room_io/_input.py
longcw: ok removed it

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
```diff
@@ -0,0 +1,114 @@
...
+                return buf.frames
+
+            buf = await asyncio.wait_for(fut, self._timeout)
+            return buf.frames
+        finally:
+            self._buffers.pop(participant_identity)
+
+    @utils.log_exceptions(logger=logger)
+    async def _read_audio_task(self, reader: rtc.ByteStreamReader, participant_id: str):
+        if (fut := self._buffers.get(participant_id)) and fut.done():
```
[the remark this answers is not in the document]
pblazej: I don't see any technical limitations here, ofc @lukasIO open for feedback as always 😸 

image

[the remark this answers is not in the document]
pblazej: Just FYI, I'm mostly struggling with managing the `Track` lifecycle here reliably, as previously it was just created during `connect()`, now we need to maintain its identity without introducing any surprises (e.g. record some other track, publish on the newly created one).

pblazej:
iOS PR (WIP) https://github.com/livekit/client-sdk-swift/pull/685

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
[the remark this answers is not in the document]
lukasIO: @bcherry yeah, that's what I meant by https://github.com/livekit/agents/pull/2156#discussion_r2068456223
the problem is that we don't have a track sid before publication time on the publishing client, so we'd need to use e.g. the mediastreamtrack id instead.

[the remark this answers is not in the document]
pblazej: > don't have a track sid before publication time on the publishing client

@lukasIO not sure if I get it - the stream attributes are set just before sending (on iOS) vs track attribute will be set while publishing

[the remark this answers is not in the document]
lukasIO: Basically I wanted to decouple track buffer (depends on `LocalTrack`) from depending on `LocalTrackPublication`'s state. 

It would be a bit faster if we don't have to wait for the track publish response and instead start sending the byte stream already in parallel to the track publication request.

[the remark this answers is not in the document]
lukasIO: thinking about it a potential downside of my idea is that we might have a short gap in audio between audio buffer that's sent until the publication is sending valid data :/ 
might be worse than the additional time waiting for the response, but would need some comparison testing to make sure

livekit-agents/livekit/agents/voice/room_io/room_io.py
pblazej: @longcw in some cases I noticed there's no `pre-connect audio received` on the agent side, whereas the client is still sending (with the same logic). Applies mostly to the 1st connection after app relaunch. Maybe there's still a race condition somewhere.

longcw: I noticed something similar and I though it was a bug in the app. or probably the byte stream header got lost in transition.

pblazej: I updated the example with my last changes (already handles track props via protobuf) https://github.com/livekit-examples/voice-assistant-swift/pull/16 - feel free to break it 💣

longcw:
waiting for a new release from https://github.com/livekit/python-sdks/pull/433

livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -70,6 +71,10 @@ class RoomInputOptions:
     participant_identity: NotGivenOr[str] = NOT_GIVEN
     """The participant to link to. If not provided, link to the first participant.
     Can be overridden by the `participant` argument of RoomIO constructor or `set_participant`."""
+    pre_connect_audio: bool = False
```
pblazej, line 74: Can we enable that by default? There shouldn't be any significant penalty for doing so.

pblazej (approved):
Looks like the Swift example instabilities are caused by sending the stream "too early" as discussed.

I'm fine with unblocking this PR, I wasn't able to break it yet 🤞

livekit-agents/livekit/agents/voice/room_io/room_io.py
pblazej: Per discussion with @bcherry - should we just let the client decide?

livekit-agents/livekit/agents/voice/room_io/room_io.py
bcherry: Is the configurable timeout still needed with the other fixed to SFU?

examples/voice_agents/pre_connect_audio_buffer.py
```diff
@@ -0,0 +1,65 @@
...
+
+load_dotenv()
+
+
+# This example demonstrates the pre-connect audio buffer for instant connect feature.
+# It captures what users say during connection time so they don't need to wait for the connection.
+# The process works in three steps:
+# 1. RoomIO is set up with pre_connect_audio=True
+# 2. When connecting to the room, the client sends any audio spoken before connection
+# 3. This pre-connection audio is combined with new audio after connection is established
```
bcherry, line 20: I'd consider just omitting the specific example and ensuring the other examples are compatible instead

bcherry (approved):

livekit-agents/livekit/agents/voice/room_io/room_io.py
longcw: Why it's not needed? There is still a chance we cannot receive the buffer bc of the network right?

bcherry: just asking back to @theomonnom 's original question: do users need to think about this or can it just be internal? it's confusing anyways because it's not clear from the name if this covers the time from agent start to stream arrive or stream end. it actually turns out it's just the time from stream header to stream trailer.

no strong feelings here just re-raising the original question.

longcw: It's to stream end. Sometimes stream head received but  the tail lost, we don't rely on the stream head for anything.

I still think it's better to expose the timeout option, mainly to notify the user there is a timeout if the buffer is not arrived in time, if the buffer lost there will be a side effect.

pblazej: Missing tail seems to be correlated with the way of sending (frame-by-frame vs max stream chunk size).

bcherry:
@longcw are there any remaining blockers to merge + release?

PR #2173, opened 2025-05-01
Add ElevenLabs plugin support for SSML break tags

tristan-warner-smith:
ElevenLabs has support for SSML pauses in speech using  `<break time="0.3s" />`.
This PR adds support for them, in addition to the existing phonemes, waiting for the full tag to be present before passing it to TTS, not doing this can result in glitchy speech as it tries to speak partial XML.


tristan-warner-smith:
PTAL @theomonnom @davidzhao

theomonnom (approved):
Thanks!!

PR #2175, opened 2025-05-01
feat(plugin-openai): add support for meta llama

MichaelSolati:


davidzhao:
@MichaelSolati given the complexities in supporting all of the tool calling, it'd be better to use their [OpenAI compatible endpoint](https://llama.developer.meta.com/docs/features/compatibility/).

then like Groq, we could just create a wrapper plugin for it

MichaelSolati:
Omg yes that would be a simpler approach, but I'm not sure if long term that's the best strat. I'm willing to bet long term we may see unique offerings via Llama that the compatibility layer won't support. But for now to get something out the door I'll take this approach.

davidzhao:
agreed. long term we should use their native apis. just a bit tricky before we get all the tests set up on v1

MichaelSolati:
> ### ❌ Invalid Changeset Format Detected
> One or more changeset files in this PR have an invalid format. Please ensure they adhere to:
> 
> * Start with `---` and include a closing `---` on its own line.
> * Each package line must be in the format:
>   `"package-name": patch|minor|major`
> * No duplicate package entries allowed.
> * A non-empty change description must follow the front matter.
> 
> **Error details:** .github/next-release/changeset-71f8b172.md: Failed to read file from git branch 'pr_head'. .github/next-release/changeset-990dabed.md: Failed to read file from git branch 'pr_head'.

@davidzhao or @bcherry will need guidance on this if/when this is approved.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -481,6 +484,42 @@ def with_telnyx(
             tool_choice=tool_choice,
         )
 
+    @staticmethod
+    def with_meta(
```
bcherry, line 488: we should offer `with_meta` or we should offer livekit-plugins-meta. I don't think we should do both. Although I do kinda like what you did where livekit-plugins-meta just wraps the openai plugin internally. 

IMO though keep it simple and lets put out `openai.LLM.with_meta` for now and roll a custom plugin when it becomes necessary.

MichaelSolati: We do something similar for groq, so I figured this would make sense.

I'll split the features into two commits, though. We can land the `with_meta` and deliberate about `livekit-plugins-meta`. I'll have to update the docs.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -481,6 +484,42 @@ def with_telnyx(
...
+        *,
+        model: str | MetaChatModels = "Llama-4-Maverick-17B-128E-Instruct-FP8",
+        api_key: str | None = None,
+        base_url: str = "https://api.llama.com/compat/v1/",
+        client: openai.AsyncClient | None = None,
+        user: NotGivenOr[str] = NOT_GIVEN,
+        temperature: NotGivenOr[float] = NOT_GIVEN,
+        parallel_tool_calls: NotGivenOr[bool] = NOT_GIVEN,
+        tool_choice: ToolChoice = "auto",
```
davidzhao, line 497: I think we should default this to:
`NotGivenOr[ToolChoice] = NOT_GIVEN,`

davidzhao (commented):
trying this out.. using the default model. it looks like tool calling isn't working. 

does it work for you with the `basic_agent` example

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
MichaelSolati: So I don't think so. Llama will throw an error if tool_choice is not set to `auto`

image

However that also means we can get rid of this arg.

MichaelSolati:
> trying this out.. using the default model. it looks like tool calling isn't working.
> 
> does it work for you with the `basic_agent` example

I've been trying to get it to call a tool/function and have had no luck. I inspected the structure of the tools array we send and it [looks correct](https://llama.developer.meta.com/docs/features/#how-to-use-tool-calling):

image

I've tinkered with explicitly passing in tools as an arg of the agent when initializing it and I've had no luck.

I've also tried different Llama models and still nothing. I'd be curious how this works using the sdk rather than the openai compatability endpoint, but I'm not sure that matters for adding the functionality to the openai plugin.

PR #2143, opened 2025-04-27
streaming for google tts

jayeshp19:
ideally we should persist tts SynthesizeStream, that'll enable proper streaming support for Google TTS.

cc: @davidzhao @theomonnom

PR #2176, opened 2025-05-01
fix(google-tts): Audio encoding for Chirp3 voices not compatible with PCM encoding

danishshaik:
Fixes an issue introduced in: https://github.com/livekit/agents/pull/2139
https://github.com/livekit/agents/pull/2139#issuecomment-2845691859

This PR addresses an issue where Google TTS voices, specifically those in the Chirp3 family, could not be used due to incompatible audio encoding requests. The default PCM encoding used by the agent framework is not supported by these specific voices.

**Problem Details:**
When attempting to synthesize speech using a Chirp3 voice via `livekit.plugins.google.tts.TTS`, the underlying API rejects the request with a `400 Bad Request` status. The error message clearly states the supported output encodings: `ALAW`, `LINEAR16`, `MULAW`, `MP3`, and `OGG_OPUS`.

**Example Error Snippet**
`livekit.agents._exceptions.APIStatusError: This voice currently only supports ALAW, LINEAR16, MULAW, MP3 and OGG_OPUS output. (status_code=400, request_id=None, body=None) {"tts": "livekit.plugins.google.tts.TTS", "attempt": 2, "streamed": false, "pid": 31836, "job_id": "AJ_xTbpV4gK3xgb"}`

**Solution:**
This change exposes the audio encoding configuration for the TTS plugin. This allows users to explicitly set a compatible encoding, such as `LINEAR16`, enabling the use of Chirp3 voices.

### DIFF
--- a/.github/next-release/changeset-63a32de3.md
+++ b/.github/next-release/changeset-63a32de3.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-google": patch
+---
+
+fix(google-tts): Audio encoding for Chirp3 voices not compatible with PCM encoding (#2176)
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
@@ -56,6 +56,7 @@ class TTS(tts.TTS):
         effects_profile_id: str = "",
         speaking_rate: float = 1.0,
         location: str = "global",
+        audio_encoding: texttospeech.AudioEncoding = texttospeech.AudioEncoding.PCM,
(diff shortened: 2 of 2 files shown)


danishshaik:
@davidliu @chrisackermann

theomonnom (approved):

jayeshp19:
we don't these errors with streaming enabled: https://github.com/livekit/agents/pull/2143

theomonnom (approved):

danishshaik:
> we don't get these errors with streaming enabled: #2143

@jayeshp19 ah interest, potentially worth setting the encoding based on the `use_streaming` flag once https://github.com/livekit/agents/pull/2143/files is merged. Or at least providing an appropriate default.

Issue #2174, opened 2025-05-01
The agent saying words like: "tools_output" during function-call - gemini real time API


shashwatsanket997:
Hi Team, Good Day!

We are facing issues with function calling while using the gemini-live api. The agent sometimes says words like: "tools_output" etc.

I also see an improvement PR: https://github.com/livekit/agents/pull/2089  
But the issue is still reproducible with this fix. 

One possible reason found was that we are using a deprecated API from google genai library to send message to model. Updating the latest one fixes the problem. 

Current:
```python
if isinstance(msg, LiveClientContent):
      await session.send(input=msg)
```

Possible Fix: 
```python 
if isinstance(msg, LiveClientContent):
      await session.send(input=msg)
elif isinstance(msg, LiveClientToolResponse):
      await session.send_tool_response(function_responses=msg.function_responses)
````

**Better Approach:**
Google document suggest to use ***more specific methods***: `send_client_content`, `send_realtime_input`, or `send_tool_response` instead.

Document link: https://googleapis.github.io/python-genai/genai.html#module-genai.types

Using the latest apis - seems to be fixing this issue.

cc: @davidzhao

Denin-Siby:
yes, even I see so many warnings about not using the latest versions. Some of them are:

`venv/lib/python3.10/site-packages/livekit/plugins/google/beta/realtime/realtime_api.py:425: DeprecationWarning: Setting `LiveConnectConfig.generation_config` is deprecated, please set the fields on `LiveConnectConfig` directly. This will become an error in a future version (not before Q3 2025)`

`venv/lib/python3.10/site-packages/livekit/plugins/google/beta/realtime/realtime_api.py:480: DeprecationWarning: The `session.send` method is deprecated and will be removed in a future version (not before Q3 2025).
Please use one of the more specific methods: `send_client_content`, `send_realtime_input`, or `send_tool_response` instead.
  await session.send(input=msg)`

Denin-Siby:
I get this error now while I try to use it with Vertex-AI:

`ERROR:livekit.plugins.google:error in send task: id parameter is not supported in Vertex AI.
Traceback (most recent call last):
  File "/Users/deninsiby/Documents/GitHub/atomic-aiagent-service/venv/lib/python3.10/site-packages/livekit/plugins/google/beta/realtime/realtime_api.py", line 490, in _send_task
    await session.send_tool_response(function_responses=msg.function_responses)
[...]

PR #2181, opened 2025-05-02
Fix: FunctionOutputResponse not getting recognised to gemini live-api

shashwatsanket997:
Hi Team, 

**Contribution** to fix: #2174 

**Context**: FunctionOutputResponse to a tool-call for realtime api was not getting sent to the gemini live api.  Because of deprecated function getting used.  

**Problem**: 
1. Whenever the tool response is injected back - the model speaks out as "tool_outputs" 
2. Model is not able to derive the context - from the tool-response(a.k.a) unable to understand. 

**Solution**: Using the latest apis - seems to be fixing this issue.

**Document link:** https://googleapis.github.io/python-genai/genai.html#module-genai.types

### Important Note: (Maybe we can update the documentation)
During several debugging found that the functionResponse should be always string - and it should not contain any special characters. 

```python
types.FunctionResponse(
    id=msg.call_id,
    name=msg.name,
    response={"output": msg.output}, #the msg.output should be str and should not contain special character.
)
```

Presence of special character in the response, is also leading to model not recognising the tool-call.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
@@ -18,6 +18,7 @@ from google.genai.types import (
     GenerationConfig,
     LiveClientContent,
     LiveClientRealtimeInput,
+    LiveClientToolResponse,
     LiveConnectConfig,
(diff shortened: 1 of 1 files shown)


shashwatsanket997:
Created new one due to sync issue related to CLA: 
https://github.com/livekit/agents/pull/2185

PR #2182, opened 2025-05-02
feat: Add `with_letta` OpenAI plugin

mattzh72:
Add the `with_letta` convenience function for the OpenAI plugin. This allows developers to more easily integrate Letta with Livekit.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -506,6 +506,13 @@ class LLM(llm.LLM):
         if not parsed.netloc:
             raise ValueError(f"URL '{base_url}' is missing a network location (e.g., domain name).")
 
+        api_key = api_key or os.environ.get("LETTA_API_KEY")
+        # Might not be necessary if self-hosted Letta instance
+        if base_url.startswith("https://api.letta.com/v1/") and api_key is None:
+            raise ValueError(
+                "Letta API key is required, either as argument or set LETTA_API_KEY environmental variable"
+            )
+
         return LLM(
             model="letta-fast",
             api_key=api_key,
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -478,6 +479,44 @@ def with_telnyx(
             tool_choice=tool_choice,
         )
 
+    @staticmethod
+    def with_letta(
+        *,
+        agent_id: str,
+        letta_server_url: str,
```
theomonnom, line 486: Should we provide a default url? 

Also let's rename the parameter to `base_url` for consistency

theomonnom (approved):
Otherwise lgtm! Thanks!

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
mattzh72: Ack!

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -478,6 +479,44 @@ def with_telnyx(
...
+        base_url = f"{base_url}/{agent_id}"
+        parsed = urlparse(base_url)
+        if parsed.scheme not in {"http", "https"}:
+            raise ValueError(f"Invalid URL scheme: '{parsed.scheme}'. Must be 'http' or 'https'.")
+        if not parsed.netloc:
+            raise ValueError(f"URL '{base_url}' is missing a network location (e.g., domain name).")
+
+        return LLM(
+            model="letta-fast",
+            api_key=api_key,
```
longcw, line 511: It's better to be able to set the key from env like `LETTA_API_KEY`
```python
        api_key = api_key or os.environ.get("LETTA_API_KEY")
        if api_key is None:
            raise ValueError(...)
```

mattzh72: Ack

PR #2184, opened 2025-05-03
Add missing agents submodule docs, improve package readmes and docstrings

bcherry:
The generated reference docs were missing a bunch of submodules that weren't in __all__ so I added them.

While I was here I noticed that the `voice` module isn't meant to be exported so I cleaned up references to it in examples and will do the same in docs. I also updated all the pypi readmes for consistency and added doc links.

image


theomonnom (approved):
Thanks! This looks good to me.

We still want the `voice` submodule tho.
We're not exporting stuff like transcription or RoomIO directly on the livekit-agents (at least not yet)

bcherry:
@theomonnom ok that makes sense - I had assumed not because voice wasn't exported in agents at all so I added it there. I still think updating the examples to use the base imports where possible is valuable so I'll leave those as is

PR #2189, opened 2025-05-04
supress stream timeout errors in google stt

jayeshp19:
Supress google stt stream timeout error when disconnecting/5 minutes timeout

PTAL @davidzhao @theomonnom

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
@@ -513,7 +513,6 @@ class SpeechStream(stt.SpeechStream):
             except GoogleAPICallError as e:
                 if e.code == 409:
                     logger.debug("stream timed out, restarting.")
-                    pass
                 else:
                     raise APIStatusError(
                         f"{e.message} {e.details}", status_code=e.code or -1
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
@@ -512,7 +512,7 @@ class SpeechStream(stt.SpeechStream):
                 raise APITimeoutError() from None
             except GoogleAPICallError as e:
                 if e.code == 409:
-                    logger.debug("stream timeout, restarting")
+                    logger.debug("stream timed out, restarting.")
                     pass
                 else:
                     raise APIStatusError(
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -505,7 +505,11 @@ async def process_stream(client: SpeechAsyncClient, stream):
             except DeadlineExceeded:
                 raise APITimeoutError() from None
             except GoogleAPICallError as e:
-                raise APIStatusError(e.message, status_code=e.code or -1) from None
+                if e.code == 409:
+                    logger.debug("stream timeout")
+                    pass
```
davidzhao, line 510:
```suggestion
                    self._pool.remove(client)
```

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -505,7 +505,11 @@ async def process_stream(client: SpeechAsyncClient, stream):
             except DeadlineExceeded:
                 raise APITimeoutError() from None
             except GoogleAPICallError as e:
-                raise APIStatusError(e.message, status_code=e.code or -1) from None
+                if e.code == 409:
+                    logger.debug("stream timeout")
```
davidzhao, line 509:
```suggestion
                    logger.debug("stream timed out, restarting")
```

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
davidzhao: I see.. we had been doing it in the other spot.. 

if we don't need to recreate it.. then we should remove the pass here.

jayeshp19: Updated

jayeshp19: the reason we've pool connection in google stt is because they have timelimit of 5 minutes, we don't actually need to create new client, we just need to create new `streaming_recognize` request

jayeshp19: Thinking about it, we don't actually need pool connection google stt

davidzhao (approved):
lg

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
davidzhao: right, that makes sense.

PR #2192, opened 2025-05-04
google stt: support cofigurable streaming

jayeshp19:
Some of the languages does not support streaming in google stt
StreamingRecognize does not support the "gu-IN" language code for the "chirp_2" model at location "us-central1"

+ better error logs

PTAL @theomonnom  @davidzhao


davidzhao (approved):
lg, thanks!

Issue #1865, opened 2025-04-02
MCP Support


ArmanJR:
Hello Livekit team!

I was wondering if you guys are planning to support the [Model Context Protocol](https://www.anthropic.com/news/model-context-protocol) for LLM tool usage and many more applications. It would be nice to have an MCP client attached to the agent.

IngLP:
+1!

theomonnom:
Hey,

This is on our roadmap. While I can't provide a precise ETA, I’d estimate it will happen within the next two months.

quancore:
+1

longcw:
for anyone interested in using MCP in agents, I created a helper function here to convert a MCPTool to the agent FunctionTool dynamically https://github.com/longcw/home-assistant-mcp-agent/blob/main/mcp_client/mcp_utils.py#L15

I think the agents will support loading function tools from MCP, pydanitc or json schema soon

jezell:
@longcw yes, @theomonnom is working on using json + json schema directly in tool definitions, which will clean that up quite a bit.

res63661:
Bruh this WOOOORKS! AAAaaargh! @longcw @theomonnom you just saved my friggin weekend

PR #2200, opened 2025-05-05
add MCP support

theomonnom:
closes https://github.com/livekit/agents/issues/1865

`mcp` is an optional dependency, install using `livekit-agents[mcp]` >=py3.10


livekit-agents/livekit/agents/__init__.py
```diff
@@ -67,6 +69,19 @@
 from .voice.room_io import RoomInputOptions, RoomIO, RoomOutputOptions
 from .worker import SimulateJobInfo, Worker, WorkerOptions, WorkerPermissions, WorkerType
 
+if typing.TYPE_CHECKING:
+    from .llm import mcp  # noqa: F401
+
+
+def __getattr__(name: str) -> typing.Any:
```
longcw, line 76: why this is needed, for solving circular imports? I am wondering is there a better way to do it to avoid the `__getattr__` here?

longcw: or it's to avoid `ImportError` when `mcp` not installed? can we do the lazy import inside the mcp module?

livekit-agents/livekit/agents/llm/mcp.py
```diff
@@ -0,0 +1,195 @@
...
+        self._read_timeout = client_session_timeout_seconds
+
+        self._cache_dirty = True
+        self._lk_tools: list[MCPTool] | None = None
+
+    @property
+    def initialized(self) -> bool:
+        return self._client is not None
+
+    def invalidate_cache(self) -> None:
```
longcw, line 40: why do we need a method instead of an arg like `use_cache` for `list_tools`?

livekit-agents/livekit/agents/__init__.py
theomonnom: Yeah it's for the latter & also allow `from livekit.agents import mcp` instead of `from livekit.agents.llm import mcp`

livekit-agents/livekit/agents/llm/mcp.py
theomonnom: list_tools is used internally mostly mostly, so it may not help much.

invalidate_cache is for the users to say, now all calls made by whatever internal code inside AgentSession will refetch the data

longcw: That makes sense, `invalidate_cache` only when update the agent in agent session. Do we need to check that flag for each llm call? basically don't cache the tools in `AgentActivity` but they cached separately in the mcp servers so we always gather the tools using `list_tools` in `AgentActivity.tools`?

longcw (approved):
looks good, one thing nit about the caching

theomonnom:
> looks good, one thing nit about the caching
It's kind of hard to call it for each LLM calls (it doesn't work for the Realtime API)

jezell:
Great work @theomonnom

PR #2204, opened 2025-05-05
Fix error handling and STT errors

Shubhrakanti:


livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -202,7 +202,7 @@ async def start(self) -> None:
         self._agent_session.on("user_input_transcribed", self._on_user_input_transcribed)
         self._agent_session._room_io = self
 
-    async def aclose(self) -> None:
+    async def aclose(self, disconnect_room: bool) -> None:
```
theomonnom, line 205: Closing the room should be done by the user, not the AgentSession

Shubhrakanti: > Closing the room should be done by the user, not the AgentSession

This disconnects the agent from the room. It doesn't delete the room

theomonnom: sry it's what I meant, we shouldn't disconnect either

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -618,13 +619,19 @@ async def _connect_ws(self) -> aiohttp.ClientWebSocketResponse:
...
+                    _to_deepgram_url(live_config, base_url=self._base_url, websocket=True),
+                    headers={"Authorization": f"Token {self._api_key}"},
+                ),
+                self._conn_options.timeout,
+            )
+        except aiohttp.client_exceptions.WSServerHandshakeError as e:
+            raise APIConnectionError(
+                f"failed to connect to deepgram: {e.code} {e.message} {e.headers}",
+                retryable=False,
+            ) from e
```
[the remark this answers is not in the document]
longcw: Yeah, I noticed the same issue in openai STT with `use_realtime=True`. It won't retry but just fail directly and won't emit the error to close the session as well.

PR #2205, opened 2025-05-05
11labs: fix default model and chunk_length_schedule

davidzhao:
- using turbo as default model, as docstring indicates
- default chunk length as recommended by 11labs, instead of specifying an override


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -99,7 +99,7 @@ def __init__(
         *,
         voice_id: str = DEFAULT_VOICE_ID,
         voice_settings: NotGivenOr[VoiceSettings] = NOT_GIVEN,
-        model: TTSModels | str = "eleven_flash_v2_5",
+        model: TTSModels | str = "eleven_turbo_v2_5",
```
theomonnom, line 102: isn't flash their newer model?

theomonnom (approved):

PR #2206, opened 2025-05-05
Increases buffered words count to Cartesia's TTS service

itsnicjohn:
Cartesia has an issue (first noticed on the "Parvati" voice) where, if a continuation is done on a very short amount of audio (e.g. one word), the audio can get wonky. They intend to improve this soon, but increasing our value here should help without costing much in performance.

Increases BUFFERED_WORDS_COUNT from 3 to 10 to alleviate voice stability issue (changing accents)


davidzhao (approved):
lg

PR #2207, opened 2025-05-06
supress StopResponse exception in llm inference

jayeshp19:
users might want to stop response in LLM node, similar to how we do it in `on_user_turn_completed`https://github.com/livekit/agents/blob/b9b18d37525350dbf9a222ff01f2b369cbf101e3/livekit-agents/livekit/agents/voice/agent_activity.py#L891

PTAL @davidzhao @theomonnom @longcw

### DIFF
--- a/livekit-agents/livekit/agents/voice/generation.py
+++ b/livekit-agents/livekit/agents/voice/generation.py
@@ -112,6 +112,8 @@ def perform_llm_inference(
                         logger.warning(
                             f"LLM node returned an unexpected type: {type(chunk)}",
                         )
+            except StopResponse:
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -112,6 +112,8 @@ async def _inference_task():
                         logger.warning(
                             f"LLM node returned an unexpected type: {type(chunk)}",
                         )
+            except StopResponse:
+                return  # ignore this turn
```
longcw, line 116: you can return a `None` in the llm_node to skip this turn

PR #2218, opened 2025-05-06
Add log when closing agent session from error

Shubhrakanti:


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -417,6 +417,9 @@ async def _aclose_impl(
 
             self.emit("close", CloseEvent(error=error))
 
+            if error is not None:
```
theomonnom, line 420: The log should be done inside the AgentSession._on_error method IMO.

Also are we sure the string representation of error is actually readable?

theomonnom: Maybe we should log the real error?
https://github.com/livekit/agents/blob/62330477dde147dbf8108e38c56089269de90e80/livekit-agents/livekit/agents/llm/llm.py#L62

using exc_info

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -565,6 +562,8 @@ def _on_error(
         if self._closing_task or error.recoverable:
             return
 
+        logger.error(f"AgentSession is closing due to unrecoverable error: {error.error}")
```
longcw, line 565:
```suggestion
        logger.error("AgentSession is closing due to unrecoverable error", exc_info=error.error)
```

Shubhrakanti: Oh man missed this from @theomonnom 's comment. Thanks!

longcw (approved):

PR #2225, opened 2025-05-07
Share _to_deepgram_url to parse booleans correctly

MartinGotelli:
Fix issue with mip_opt_out for TTS.

Share repeated logic under _utils file


MartinGotelli:
@theomonnom can you take a look? It seems I broke the Deepgram TTS, sorry 🙈

theomonnom (approved):

PR #2227, opened 2025-05-08
improve error handling in stt, tts and llm

longcw:


theomonnom (approved):

livekit-agents/livekit/agents/stt/stream_adapter.py
```diff
@@ -47,7 +42,7 @@ def stream(
         self,
         *,
         language: NotGivenOr[str | None] = NOT_GIVEN,
-        conn_options: APIConnectOptions = DEFAULT_STREAM_ADAPTER_API_CONNECT_OPTIONS,
+        conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
```
theomonnom, line 45: I don't think we want to change the default here

theomonnom: Maybe it's fine if we put something different inside the StreamAdapterWrapper constructor when calling super()

examples/voice_agents/error_handling.py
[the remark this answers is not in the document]
longcw: oh yeah, there was a `error_callback.py`, will update that.

livekit-agents/livekit/agents/stt/stream_adapter.py
longcw: there is no retry in the fallback adapter, so before it was a bug with 0 retry.

examples/voice_agents/error_handling.py
```diff
@@ -0,0 +1,79 @@
...
+    @session.on("error")
+    def _on_error(ev: ErrorEvent):
+        if ev.error.recoverable:
+            return
+
+        # TTS and LLM errors can be marked as recoverable
+        # since these components are recreated for each response
+        if isinstance(ev.source, (tts.TTS, llm.LLM)):
+            ev.error.recoverable = True
+            return
```
theomonnom, line 54: I'm not sure to like this pattern, the assumption here is that when it's unrecoverable, it's final

livekit-agents/livekit/agents/stt/stream_adapter.py
theomonnom: It's the StreamAdapter here, here even if the default is 3 retry, it'll do 9 retry (or even more if there is multiple segments)

examples/voice_agents/error_handling.py
longcw: It's hard to say a tts or llm is unrecoverable since they are created per response, it's just to tell the user they can ignore it if they don't want to close the session. I'll move this to the other example as commented by default.

livekit-agents/livekit/agents/stt/stream_adapter.py
theomonnom: The bug here is that we were passing this argument inside the recognize method

examples/voice_agents/error_handling.py
theomonnom: Right, but if an LLM failed 3 times, it's likely it is down?

livekit-agents/livekit/agents/stt/stream_adapter.py
longcw: fixed

examples/voice_agents/error_callback.py
```diff
@@ -44,6 +48,22 @@ def on_error(ev: ErrorEvent):
...
         )
 
+        # If want to continue the conversation, we can set the recoverable to True
+
+        # TTS and LLM errors can be marked as recoverable
+        # since these components are recreated for each response
+
+        # if isinstance(ev.source, (tts.TTS, llm.LLM)):
+        #     ev.error.recoverable = True
+        #     return
```
theomonnom, line 58: I don't even recommend it fwiw lol

theomonnom (approved):

livekit-agents/livekit/agents/tts/stream_adapter.py
```diff
@@ -87,7 +90,9 @@ async def _forward_input():
         async def _synthesize():
             async for ev in self._sent_stream:
                 last_audio: SynthesizedAudio | None = None
-                async for audio in self._wrapped_tts.synthesize(ev.token):
+                async for audio in self._wrapped_tts.synthesize(
+                    ev.token, conn_options=self._wrapped_tts_conn_options
```
theomonnom, line 94: Mmh nevermind I think we do want individual retry for the synthesize

theomonnom (changes requested):
Otherwise it may not retry it when using the StreamAdapter, the segment could be lost

livekit-agents/livekit/agents/tts/stream_adapter.py
longcw: since we emit the errors from `_wrapped_tts.synthesize`, the actual connection options has to be passed to the wrapped tts, the stream adapter always have 0 retries. what do you mean `individual retry for the synthesize`?

examples/voice_agents/error_callback.py
longcw: failed 3 times in 6 seconds doesn't really mean the service is down IMO, for example reached the max concurrency at the moment. It's to allow ppl to continue the conversation if they don't want to close the session, also they can have a self-defined error counter or condition in the callback.

theomonnom: Right, but I'm really doubtful this pattern will be kept in the future. I would say better to do nothing for now. 
(I don't feel like we're going to guarantee changes made to the error inside this callback will really impact the decision to close or not in our internal code)

longcw: or an api that closing the session, so we don't do that internally

livekit-agents/livekit/agents/tts/stream_adapter.py
theomonnom: In this case it's the opposite, the StreamAdapter has 3 retries but the individual requests have 0.
It should be the opposite

longcw: No, it's opposite as you thought, the 0 retry was passed to the stream adapter, this one is the default one (the one user specified)

examples/voice_agents/error_callback.py
theomonnom: Yes, it is closing automatically because it was designed to be fatal (no recovery possible)

theomonnom (approved):

PR #2238, opened 2025-05-08
use the forkserver on Linux by default 

theomonnom:
- also load every registered plugins
- processes are now launched instantly 0.02seconds instead of 3 seconds on the basic_agent.py example

Users can still use the old behaviour using `WorkerOptions(multiprocessing_context="spawn")`

This is safe as long as our plugins don't start threads at import time

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -52,6 +52,7 @@ from .job import (
     RunningJobInfo,
 )
 from .log import DEV_LEVEL, logger
+from .plugin import Plugin
 from .types import NOT_GIVEN, NotGivenOr
 from .utils import is_given
 from .utils.hw import get_cpu_monitor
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/worker.py
```diff
@@ -215,6 +216,13 @@ class WorkerOptions:
 
     By default it uses ``HTTP_PROXY`` or ``HTTPS_PROXY`` from environment
     """
+    multiprocessing_context: Literal["spawn", "forkserver"] = (
+        "spawn" if not sys.platform.startswith("linux") else "forkserver"
```
davidzhao, line 220: nit: easier to read

```suggestion
        "forkserver" if sys.platform.startswith("linux") else "spawn"
```

livekit-agents/livekit/agents/worker.py
```diff
@@ -215,6 +216,13 @@ class WorkerOptions:
...
+    )
+    """The multiprocessing context to use.
```
davidzhao, line 222: unnecessary comment

davidzhao (approved):

livekit-agents/livekit/agents/worker.py
theomonnom: will do as a nit in a next unrelated PR

PR #2245, opened 2025-05-08
set kind and kind-detail for agent participant info

real-danm:


livekit-agents/livekit/agents/worker.py
```diff
@@ -629,8 +634,11 @@ async def _connection_task(self) -> None:
                 join_jwt = (
                     api.AccessToken(self._opts.api_key, self._opts.api_secret)
                     .with_grants(api.VideoGrants(agent=True))
-                    .to_jwt()
+                    .with_kind("agent")
```
real-danm, line 637: not sure why this wasn't set previously, maybe there was a reason?

theomonnom: This isn't a jwt token for the rtc connection, but for the websocket connection of the agent endpoint

PR #2270, opened 2025-05-12
fix raw function tool for google gemini

longcw:


longcw:
I think we should fix the type annotation for `FunctionTool | RawFunctionTool` across the agents, rn most of the tools are annotated as `FunctionTool`.

theomonnom (approved):

Issue #2222, opened 2025-05-07
Realtime Multimodal metrics and transcription callbacks not firing in `livekit-agent@1.0.19`


wahidnj:
After upgrading from `livekit-agents@0.12.20` to `livekit-agent@1.0.19`, my Realtime Multimodal session no longer emits any of the old metrics or transcription events:

session.on('metrics_collected', _on_metrics_collected)
session.on('user_speech_committed', text => asyncio.create_task(on_user_speech(text)))
session.on('agent_speech_committed', text => asyncio.create_task(on_agent_speech(text)))

I tried with Agent Session with realtime but didnt get any metrics data. Also used metric usageCollector. But no result as well for Realtime.

ChenghaoMou:
Both have been updated in the newer visions.

You can check out the docs for more details:
- [Metrics](https://docs.livekit.io/agents/build/metrics/)
- [Events](https://docs.livekit.io/agents/build/events/)

tl;dr
- `conversation_item_added` should be able to replace the old transcription events.
- I couldn't find any **metric emitting code** in both OpenAI and Gemini Realtime implementation, unlike their non-realtime counterparts. It's likely those events are not ready. And `RealtimeSession` only emits following events:

```
EventTypes = Literal[
    "input_speech_started",  # serverside VAD (also used for interruptions)
    "input_speech_stopped",  # serverside VAD
    "input_audio_transcription_completed",
    "generation_created",
    "error",
]
```


If the docs can be updated to reflect the metric availability (like the EOU metric) in more details, that would be great.

wahidnj:
@ChenghaoMou Can you help How I will calculate input and output tokens of realtime openai in latest version.

ChenghaoMou:
Probably they will be added back soon. They were removed in some commits not long ago ([1](https://github.com/livekit/agents/commit/f12ed5152b62de18c0a5e435b991ffb7ec8e20a8) and [2](https://github.com/livekit/agents/commit/28a0851851dec611bd3a289ef99d5fc675e18d22)) but I have no idea why. 

cc @theomonnom might have a better idea.

wahidnj:
@theomonnom please reply whats the problem?

theomonnom:
Hey, we're going to bring back metrics very soon.

Transcriptions are already handled in https://github.com/livekit/agents/pull/2247

jayant-codebuddy:
I’m using LiveKit Agents (livekit-agents==1.3.2) + OpenAI Realtime API via:

[...]

PR #2275, opened 2025-05-13
add realtime model metrics

longcw:
fix https://github.com/livekit/agents/issues/2222


theomonnom (approved):

theomonnom:
Il wondering how we should expose the user transcriptions latency when using RealtimeModels

longcw:
> I’m wondering how we should expose the user transcriptions latency when using the RealtimeModel

add a `transcription_delay` in RealtimeModelMetrics?

Issue #2271, opened 2025-05-12
GatheringFuture CancelledError in Deepgram STT when using Push-to-Talk


JeisonNovoa:
After enabling Push-to-Talk (PTT) in our LiveKit Agents setup, we started seeing unhandled asyncio.CancelledError exceptions in the Deepgram STT plugin. This error did not occur when running without PTT.

Steps to reproduce

Install and run a LiveKit Agents voice bot with Deepgram STT and PTT enabled.

Connect a participant and use the “start_turn” and “end_turn” RPCs to push and release talk.

Watch the server logs for repeated entries like:

vbnet
Copiar código
ERROR:asyncio:_GatheringFuture exception was never retrieved
future: <_GatheringFuture finished exception=CancelledError()>
Traceback (most recent call last):
  File ".../livekit/plugins/deepgram/stt.py", line 536, in recv_task
    msg = await ws.receive()
  ...
asyncio.exceptions.CancelledError
Expected behavior
When PTT ends the speech turn, the Deepgram websocket should close cleanly without logging unhandled CancelledError exceptions.

Actual behavior
The STT recv_task coroutine is cancelled when the websocket closes, but the CancelledError bubbles up and is logged as an unhandled exception.

Environment
We are running on Python 3.11 with these key package versions (full requirements.txt attached):

livekit==1.0.6

livekit-agents==1.0.20

livekit-plugins-deepgram==1.0.19

aiohttp==3.10.11

httpx==0.28.1

deepgram-sdk==3.7.7

Additional context

PTT audio start/stop behavior is working correctly.

The same Deepgram STT configuration in continuous-listen mode (no PTT) does not produce any errors.

We have a small monkey-patch on aiohttp.ClientSession.__init__ only to drop an unsupported proxy kwarg; it does not affect websocket behavior.

Please let me know if you need any more information or a minimal repro repository.

longcw:
The error is raised when we reset the stt in `clear_user_turn`. We reset it by closing the old one then creating a new one.

The stt may raise this error since it was not closed gracefully. So it shouldn't affect anything but I'll try to make the reset more graceful.

PR #2276, opened 2025-05-13
fix gather not retrieved error in stt

longcw:
fix https://github.com/livekit/agents/issues/2271

`asyncio.gather` wrapped by `asyncio.wait` needs await to avoid `_GatheringFuture exception was never retrieved` error

https://stackoverflow.com/questions/63353358/weird-behaviour-of-asyncio-cancellederror-and-gatheringfuture-exception-was-ne


davidzhao (approved):
ah this is pretty tricky.. nice catch

Issue #2281, opened 2025-05-13
Neuphonic and Cartesia websockets raise an error after long periods of silence


adnansiddiquei:
To recreate
1. Follow the [Voice AI Quickstart](https://docs.livekit.io/agents/start/voice-ai/) guide on the docs to create a simple chat room.
2. Use either Neuphonic or Cartesia as the TTS.
3. Start the room with `python main.py dev` and connect to the room using the LiveKit agents playground in the browser.
4. Let the Agent greet you. Give it a response. Let the agent speak its response. Then go silent for about a minute.
5. Speak to the Agent.
6. You will see the attached errors in your console (see below). See attachments `neuphonic_error.log` and `cartesia_error.log`. In the Cartesia case, the chatbot does not break, the warning is raised in the console but the chatbot continues to speak. In the Neuphonic case, this error breaks the chatbot. The user needs to wait a few seconds and speak again before the chatbot actually responds.

**neuphonic_error.log**
```
2025-05-13 15:04:51,577 - ERROR livekit.agents - Error in _inference_task
Traceback (most recent call last):
  File "/Users/adnansiddiquei/Desktop/Neuphonic/tmp/test/Mar25/my-app/.venv/lib/python3.12/site-packages/livekit/agents/utils/log.py", line 16, in async_fn_logs
    return await fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adnansiddiquei/Desktop/Neuphonic/tmp/test/Mar25/my-app/.venv/lib/python3.12/site-packages/livekit/agents/voice/generation.py", line 142, in _inference_task
    async for audio_frame in tts_node:
  File "/Users/adnansiddiquei/Desktop/Neuphonic/tmp/test/Mar25/my-app/.venv/lib/python3.12/site-packages/livekit/agents/voice/agent.py", line 473, in tts_node
    async for ev in stream:
  File "/Users/adnansiddiquei/Desktop/Neuphonic/tmp/test/Mar25/my-app/.venv/lib/python3.12/site-packages/livekit/agents/tts/tts.py", line 449, in __anext__
    raise exc from None
  File "/Users/adnansiddiquei/Desktop/Neuphonic/tmp/test/Mar25/my-app/.venv/lib/python3.12/site-packages/livekit/agents/tts/tts.py", line 301, in _main_task
    return await self._run()
           ^^^^^^^^^^^^^^^^^
[...]

PR #2282, opened 2025-05-13
Partial fix for Neuphonic bug detailed in issue #2281. 

adnansiddiquei:
This PR is a partial fix for https://github.com/livekit/agents/issues/2281.

This does NOT fix the underlying issue. But this small code change will result in Neuphonic having the same underlying behaviour to this bug as Cartesia currently does. I.e., this bug will result in the chatbot not crashing entirely, but simply logging an exception to the console and continuing as normal.

### DIFF
--- a/livekit-plugins/livekit-plugins-neuphonic/livekit/plugins/neuphonic/tts.py
+++ b/livekit-plugins/livekit-plugins-neuphonic/livekit/plugins/neuphonic/tts.py
@@ -366,7 +366,14 @@ class SynthesizeStream(tts.SynthesizeStream):
             )
 
             while True:
-                msg = await ws.receive()
+                try:
+                    msg = await ws.receive()
+                except Exception as e:
+                    raise APIStatusError(
+                        "Neuphonic connection closed unexpectedly",
+                        request_id=request_id,
+                    )
+
                 if msg.type in (
                     aiohttp.WSMsgType.CLOSED,
                     aiohttp.WSMsgType.CLOSE,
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-neuphonic/livekit/plugins/neuphonic/tts.py
```diff
@@ -366,7 +366,14 @@ async def _recv_task(ws: aiohttp.ClientWebSocketResponse):
             )
 
             while True:
-                msg = await ws.receive()
+                try:
+                    msg = await ws.receive()
```
longcw, line 370: Actually the cartesia plugin doesn't have this, the error of cartesia is a connection issue that the ws got closed then the APIStatusError was raised after `ws.receive`
```python
                msg = await ws.receive()
                if msg.type in (
                    aiohttp.WSMsgType.CLOSED,
                    aiohttp.WSMsgType.CLOSE,
                    aiohttp.WSMsgType.CLOSING,
                ):
                    raise APIStatusError(
                        "Cartesia connection closed unexpectedly",
                        request_id=request_id,
                    )
```

From your traceback it's the session in agent not properly closed so it raised an error from `ws.receive()`, do you have a full traceback log for debugging?

longcw (approved):

livekit-plugins/livekit-plugins-neuphonic/livekit/plugins/neuphonic/tts.py
adnansiddiquei: For additional context, these are the exact versions of my livekit dependencies.
I am using these slightly older versions, but the exact error is also present in exactly the same way in the newest versions of the packages as well.
```
(my-app) adnansiddiquei@Mac my-app % uv pip list | grep livekit
livekit                            1.0.6
livekit-agents                     1.0.13
livekit-api                        1.0.2
livekit-plugins-cartesia           1.0.13
livekit-plugins-deepgram           1.0.13
livekit-plugins-neuphonic          1.0.13
livekit-plugins-noise-cancellation 0.2.1
livekit-plugins-openai             1.0.13
livekit-plugins-silero             1.0.13
livekit-plugins-turn-detector      1.0.13
livekit-protocol                   1.0.2
```

PR #2289, opened 2025-05-14
add call_later_if_silent to wait for a silence period

longcw:


examples/voice_agents/basic_agent.py
```diff
@@ -37,8 +37,10 @@ def __init__(self) -> None:
 
     async def on_enter(self):
         # when the agent is added to the session, it'll generate a reply
-        # according to its instructions
-        self.session.generate_reply()
+        # according to its instructions if the user remains silent
+        if await self.session.ensure_silence_for(timeout=3):
```
longcw, line 41: open to the api name

bcherry: `ensure` is an odd word in this context IMO. it feels like an active word that implies that you're actually creating silence instead of merely detecting it.

i'd suggest

`await self.session.user_speech_timeout(3)`

or maybe "timeout" isn't quite right either. you're looking for a silence interval:

`await self.session.detect_silence(interval=3)`

but no strong feelings. thanks for adding this!

longcw: replaced it with `session. call_later_if_silent` that accepts a callback. wdyt?

bcherry: works for me. a callback seems reasonable.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -544,6 +566,34 @@ def interrupt(self) -> asyncio.Future:
...
+            *args: Arguments to pass to callback
+
+        Returns:
+            asyncio.TimerHandle: Timer handle for the scheduled callback
+        """
+        if self._activity is None:
+            raise RuntimeError("AgentSession isn't running")
+
+        handle = self._loop.call_later(delay, callback, *args)
+        if self._user_state == "speaking":
```
davidzhao, line 590: I think the call back condition would probably need to be done differently? what if the following occurs:
- VAD is firing (could be false positive)
- no transcription is received yet

in this case should we cancel the event altogether? 

another thing is.. if agent is speaking, then this timer should not count. we need to count the time after the agent has switched to listening *and* the user is not speaking

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -116,6 +126,7 @@ def __init__(
         max_endpointing_delay: float = 6.0,
         max_tool_steps: int = 3,
         video_sampler: NotGivenOr[_VideoSampler | None] = NOT_GIVEN,
+        user_away_timeout: NotGivenOr[float] = NOT_GIVEN,
```
theomonnom, line 129: Let's provide a default value, 15seconds seems like a good default.

Can we add an example using the user_state_changed callback to listen for the `away` state.
Then do:

```
# if away then ...
session.generate_reply(instructions="the user has been inactive—send a follow-up message to re-engage them."]
```

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -544,6 +566,34 @@ def interrupt(self) -> asyncio.Future:
 
         return self._activity.interrupt()
 
+    def call_later_if_silent(
```
theomonnom, line 569: I'm not a fan of callbacks, what do you think about this API instead (Let's discuss in Slack ideally, I'm still OOO so I'll answer faster there). Ideally we provide async-first APIs

Note that this is entirely separate from our `away` state, which addresses different issues

```python
async def on_enter(self):
    try:
        await self.session.wait_for_user_turn(timeout=3.0)
    except asyncio.TimeoutError:
        self.session.generate_reply() # let the agent speaks first
```


At the moment, it doesn't return anything, but this gives us the flexibility to return whatever we need in the future — for example, a helper that lets you directly await the transcription.

livekit-agents/livekit/agents/voice/agent_session.py
theomonnom: Agree, we need to add redundancy here, listen for transcriptions & VAD events. None of them are 100% reliable signals 


> another thing is.. if agent is speaking, then this timer should not count. we need to count the time after the agent has switched to listening and the user is not speaking

This reinforces my earlier suggestion for using "wait_for_user_turn" — it clearly signals the agent turn isn't counted IMO

livekit-agents/livekit/agents/voice/agent_session.py
theomonnom, line 129: Just thinking out loud:
```suggestion
        user_inactivity_timeout: NotGivenOr[float] = NOT_GIVEN,
```

theomonnom (changes requested):

livekit-agents/livekit/agents/voice/agent_session.py
theomonnom: also see https://github.com/livekit/agents/pull/2289#discussion_r2094652854

livekit-agents/livekit/agents/voice/agent_session.py
longcw: What if you call `wait_for_user_turn` when agent is speaking, should we wait for a longer time than user specified to count only after the agent finished? I don't think that the expected behavior when user specified the timeout.

longcw: also what should we do when it's the realtime model without VAD, for OAI there is `input_speech_started` and `input_speech_stopped` events, but for gemini it only have the input audio transcription events.

this feature is not urgent, we can hold until it's clear.

longcw:
will create another pr for user `away` state first.

PR #2291, opened 2025-05-14
Expose google live api input config

Terisback:
Sometimes you'd like to disable VAD


davidzhao:
when this is disabled, currently there isn't a way to generate a turn explicitly. we are working on this.

Terisback:
I'm using disabled VAD for moments of "elevator music" when model just awaits new state from backend (which can be a long time) and should just shut up. It was handy in openai api.

PR #2292, opened 2025-05-14
feat: allow to switch between Gemini and Vertex AI using env vars

mateuszkulpa:
This PR adds support for configuring Google LLM using environment variables for both Gemini and Vertex AI.
This can be useful, for example, when developing an agent worker locally using Gemini, while using Vertex AI in production. This behavior aligns with the `python-genai` library: https://googleapis.github.io/python-genai/#create-a-client

Additionally, it removes the warning about a missing `GOOGLE_APPLICATION_CREDENTIALS`, as Google supports other authentication methods such as Workload Identity, and the Google SDK handles credential checks and throws appropriate errors.

It also adds a missing default value for `location` in the Google LLM provider, in accordance with both the documentation and the Realtime API provider.


davidzhao (approved):
nice, lgtm!

PR #2295, opened 2025-05-15
add to_provider_format for ChatContext

longcw:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -84,6 +84,7 @@ def __init__(
         metadata: NotGivenOr[dict[str, str]] = NOT_GIVEN,
         max_completion_tokens: NotGivenOr[int] = NOT_GIVEN,
         timeout: httpx.Timeout | None = None,
+        _provider: NotGivenOr[Literal["openai"] | str] = NOT_GIVEN,
```
longcw, line 87: we can specify a different provider when creating the LLM from `openai.LLM.with_xxx` if it needs a different format.

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -315,6 +317,45 @@ def to_dict(
...
+        self, format: Literal["aws"], *, generating_reply: bool = True
+    ) -> tuple[list[dict], _provider_format.aws.AWSFormatData]: ...
+
+    @overload
+    def to_provider_format(
+        self,
+        format: Literal["anthropic"],
+        *,
+        generating_reply: bool = True,
+        cache_control: dict[str, Any] | None = None,
```
theomonnom, line 341: Do we need to support this for Anthropic? If not I would make it a separate PR, we 're just very bad at it today.

We always replace the first instructions for e.g

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -315,6 +317,45 @@ def to_dict(
...
+    ) -> tuple[list[dict], Literal[None]]: ...
+
+    @overload
+    def to_provider_format(
+        self, format: Literal["google"], *, generating_reply: bool = True
+    ) -> tuple[list[dict], _provider_format.google.GoogleFormatData]: ...
+
+    @overload
+    def to_provider_format(
+        self, format: Literal["aws"], *, generating_reply: bool = True
```
theomonnom, line 332: Let's remove generate_reply (as discussed in Slack)

livekit-agents/livekit/agents/llm/_provider_format/openai.py
```diff
@@ -0,0 +1,186 @@
...
+            group_id = item.id.split("/")[0]
+            if group_id not in item_groups:
+                item_groups[group_id] = _ChatItemGroup().add(item)
+            else:
+                item_groups[group_id].add(item)
+        elif item.type == "function_call_output":
+            tool_outputs.append(item)
+        else:
+            item_groups[item.id] = _ChatItemGroup().add(item)
+
```
theomonnom, line 32: Seems like this logic is needed for every LLMs no? not just OAI

livekit-agents/livekit/agents/llm/_provider_format/__init__.py
```diff
@@ -0,0 +1,8 @@
+from typing import Literal
+
+from . import anthropic, aws, google, openai
+
+LLMFormatName = Literal["openai", "google", "aws", "anthropic"]
```
theomonnom, line 5: Is this used anywhere?

livekit-agents/livekit/agents/llm/chat_context.py
longcw: `cache_control` was originally there, wdym we always replace the first instructions?

livekit-agents/livekit/agents/llm/chat_context.py
longcw: remove it and always enable it by default?

livekit-agents/livekit/agents/llm/_provider_format/__init__.py
longcw: I was thinking to make openai LLM._run [support different formats](https://github.com/livekit/agents/pull/2295/files#diff-0186148b58ae1916b3785d774437ad73ec257f2cf4a83d6adb686413aa041febR581) so we can specify the formats in `openai.LLM.with_xxx()`, in case any of them needs a specific format.

Then we need to make add the type for the `format` arg with this type for openai.LLM.

theomonnom: Ah I see, but we should do that automatically, so it isn't very relevant to export the types, we can directly put our Literals inside the to_provider_format method

livekit-agents/livekit/agents/llm/_provider_format/openai.py
theomonnom: I see, in any case, it's incorrect to have an function_output without a function_call, so this should be a shared code,

livekit-agents/livekit/agents/llm/chat_context.py
theomonnom: Yeah remove it, we should be very explicit, generate_reply is stateful (e.g what will the AgentSession do).
ChatContext can be used by whatever application, not necessarily a "running" agent.

We can easily remove it by adding new kwargs like: 

"requires_first_user_message"/"requires_last_user_message", ...

livekit-agents/livekit/agents/llm/chat_context.py
theomonnom: I see, let's remove it, it's just broken IMO

> wdym we always replace the first instructions?

generate_reply and update_instructions are doing that

theomonnom (changes requested):

livekit-agents/livekit/agents/llm/_provider_format/aws.py
```diff
@@ -0,0 +1,96 @@
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass
+
+from livekit.agents import llm
+
+
+@dataclass
+class AWSFormatData:
```
theomonnom, line 10: nit
```suggestion
class BedrockFormatData:
```

livekit-agents/livekit/agents/llm/_provider_format/anthropic.py
```diff
@@ -0,0 +1,107 @@
...
+                {
+                    "id": msg.call_id,
+                    "type": "tool_use",
+                    "name": msg.name,
+                    "input": json.loads(msg.arguments or "{}"),
+                    "cache_control": cache_ctrl_i,
+                }
+            )
+        elif msg.type == "function_call_output":
+            content.append(
```
theomonnom, line 60: Anthropic supports "errored" tool output, I think this PR is a great opportunity to support it. Our OutputItem already has a is_error bool

livekit-agents/livekit/agents/llm/_provider_format/__init__.py
longcw: you need to pass the format name to the `LLM` ctor, then the `to_provider_format` is called in `_run` of LLM. My point is when the `format` is a str, it cannot find the correct overload to have the correct type of return values. Tho this only affects the type check.

livekit-agents/livekit/agents/llm/chat_context.py
longcw: Can you explain why it's broken and how to fix that?

The caching was exposed to anthropic LLM, I have the concern that removing it seems to be a breaking change? or we keep the option but not actually apply it?
```python
class LLM(llm.LLM):
    def __init__(
        self,
        *,
        model: str | ChatModels = "claude-3-5-sonnet-20241022",
        ...
        caching: NotGivenOr[Literal["ephemeral"]] = NOT_GIVEN,
    ) -> None:
```

livekit-agents/livekit/agents/llm/_provider_format/__init__.py
longcw: ok, added below as fallback then removed the `LLMFormatName`
```python
    @overload
    def to_provider_format(self, format: str, **kwargs: Any) -> tuple[list[dict], Any]: ...
```

livekit-agents/livekit/agents/llm/_provider_format/openai.py
longcw: applied it to all formats

livekit-agents/livekit/agents/llm/chat_context.py
theomonnom: I see, I think the annoying thing is that Anthropic caching is per-message. 

From a design perspective, it's kind of weird that this decision is made by the ChatContext.

longcw: > From a design perspective, it's kind of weird that this decision is made by the ChatContext.

That's true, maybe we add the cache options in the anthropic plugin after `to_provider_format`?

theomonnom: yeah that would make more sense

longcw: updated, also updated the caching checkpoints https://github.com/livekit/agents/pull/2295/commits/2df1b9ec88a8985c09f93856761c00074983e51b

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -330,6 +331,50 @@ def to_dict(
...
+        format: Literal["anthropic"],
+        *,
+        requires_first_user_message: bool = True,
+    ) -> tuple[list[dict], _provider_format.anthropic.AnthropicFormatData]: ...
+
+    @overload
+    def to_provider_format(self, format: str, **kwargs: Any) -> tuple[list[dict], Any]: ...
+
+    def to_provider_format(
+        self, format: Literal["openai", "google", "aws", "anthropic"] | str, **kwargs: Any
```
theomonnom, line 365: Ok last change I promise 😅

I think we can standardize `requires_first_user_message` & `requires_last_user_message` are actually needed for all formats.

For example I realized that a lot of the OpenAI compatible endpoints using Llama also has this limitation that the last or first message needs to be a user one. So OpenAI also needs it.

So I think we should just put those arguments directly to the common implementation.

Also let's rename them to:

`inject_dummy_user_message`, I think we don't need 2 arguments, it is basically saying: the assistant should answer a user's question. 

This is what both `requires_first_user_message` & `requires_last_user_message` were doing really

theomonnom (approved):

livekit-agents/livekit/agents/llm/chat_context.py
longcw: Then it's similar to the original `generating_reply` but with a different name lol, but yeah a specific name is better, just if we won't need to inject any assistant messages for some LLM providers

PR #2297, opened 2025-05-15
aws: fixed STT error at end of session

### LINKED ISSUE #2099: AWS STT (Transcribe) Timeout
## Description
When using AWS Transcribe with LiveKit Agents, the STT service times out after 15 seconds and stops transcribing. This creates an issue for my use case where I need to play a required legal consent message (~20 seconds long) before user interaction begins. Users cannot interrupt this message, but the transcription service stops and further interactions are hampered.

## Environment

LiveKit Agent Version: 1.0.11
AWS STT Version: 1.0.11

## Steps to Reproduce

1. Initialize a LiveKit Agent with AWS Transcribe as the STT provider
2. Begin a session that starts with a mandatory legal consent message (approximately 20 seconds in duration)
3. Observe that AWS Transcribe stops transcribing after approximately 15 seconds

## Expected Behavior

[error_log.txt](https://github.com/user-attachments/files/19881789/error_log.txt)

AWS Transcribe should continue transcribing post the duration of the consent message (full 20+ seconds) without timing out.

## Actual Behavior

AWS Transcribe times out after 15 seconds and stops transcribing.

## Additional Context

This is particularly problematic for my application which must deliver complete legal consent messages without interruption for compliance reasons. The consent message cannot be shortened due to legal requirements.

davidzhao:
also simplified Session creation, removed boto3 dep

closes #2099

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
@@ -13,22 +13,26 @@
 from __future__ import annotations
 
 import asyncio
+import os
 from dataclasses import dataclass
 
-import aioboto3
-from amazon_transcribe.auth import StaticCredentialResolver
+from amazon_transcribe.auth import AwsCrtCredentialResolver
 from amazon_transcribe.client import TranscribeStreamingClient
-from amazon_transcribe.model import Result, TranscriptEvent
+from amazon_transcribe.exceptions import BadRequestException
+from amazon_transcribe.model import Result, StartStreamTranscriptionEventStream, TranscriptEvent
 
 from livekit import rtc
-from livekit.agents import DEFAULT_API_CONNECT_OPTIONS, APIConnectOptions, stt, utils
+from livekit.agents import (
+    DEFAULT_API_CONNECT_OPTIONS,
+    APIConnectOptions,
+    stt,
+    utils,
+)
 from livekit.agents.types import NOT_GIVEN, NotGivenOr
 from livekit.agents.utils import is_given
 
 from .log import logger
-from .utils import DEFAULT_REGION, get_aws_async_session
-
-REFRESH_INTERVAL = 1800
+from .utils import DEFAULT_REGION
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
```diff
@@ -53,9 +57,7 @@ def __init__(
         self,
         *,
         region: NotGivenOr[str] = NOT_GIVEN,
-        api_key: NotGivenOr[str] = NOT_GIVEN,
-        api_secret: NotGivenOr[str] = NOT_GIVEN,
```
longcw, line 57: why the `api_key` and `api_secret` are not needed?

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/tts.py
```diff
@@ -90,10 +90,10 @@ def __init__(
             sample_rate=sample_rate,
             num_channels=TTS_NUM_CHANNELS,
         )
-        self._session = session or get_aws_async_session(
```
longcw, line 93: should we remove the `get_aws_async_session` from utils if it's not used anywhere?

longcw (approved):
looks good

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
davidzhao: aws has a credentials loader that automatically looks at env vars along with iam roles and other authorizations. it's generally discouraged to pass in explicit keys. When running in AWS all of these can be automated via iam.

for the other services we handle it properly because they would take in those parameters.. but transcription service doesn't.. so I think it's better to remove these fields.

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/tts.py
davidzhao: good call. removed

51750:
In our use case, the STT object needs to be dynamically created with credentials passed through metadata. Since direct support for passing the API key has been removed, is there any recommended workaround or alternative approach? We can not retrieving the credentials from environment variables.

PR #2311, opened 2025-05-16
Add support for decoding preconnect buffer

lukasIO:


theomonnom (approved):

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
```diff
@@ -107,19 +107,45 @@ async def _read_audio_task(self, reader: rtc.ByteStreamReader, participant_id: s
...
+                )
+
+                async for chunk in reader:
+                    decoder.push(chunk)
+
+                decoder.end_input()
+
+                async for decoded_frame in decoder:
+                    buf.frames.append(decoded_frame)
+                    duration += decoded_frame.duration
```
theomonnom, line 126: This isn't streamed, but prob fine, we receive faster than realtime

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
```diff
@@ -107,19 +107,45 @@ async def _read_audio_task(self, reader: rtc.ByteStreamReader, participant_id: s
             num_channels = int(reader.info.attributes["channels"])
 
             duration = 0
-            audio_stream = utils.audio.AudioByteStream(sample_rate, num_channels)
-            async for chunk in reader:
-                for frame in audio_stream.push(chunk):
+
+            # Check if we need to decode opus
+            is_opus = reader.info.mime_type == "audio/opus"
```
longcw, line 112: let's check the mime_type non empty or is not `audio/pcm` instead of only `audio/opus`? we can do some refactoring and pass the `format` into the decoder after https://github.com/livekit/agents/pull/2152.

longcw (approved):

livekit-agents/livekit/agents/voice/room_io/_pre_connect_audio.py
longcw: It's okay, we don't need the buffer to be a stream

PR #2315, opened 2025-05-16
Mike/update livekit-protocol and livekit-api dependencies to match 1.0

mike-r-mclaughlin:


davidzhao (approved):
thanks Mike!

livekit-agents/setup.py
```diff
@@ -49,8 +49,8 @@
     install_requires=[
         "click~=8.1",
         "livekit>=0.18.1",
```
davidzhao, line 51: you'd want to change this too to `"livekit-api>1,<2"`

PR #2321, opened 2025-05-17
fix chat context item order

longcw:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1242,20 +1248,29 @@ def _on_first_frame(_: asyncio.Future) -> None:
...
+                role="assistant",
+                content=text_out.text,
+                id=llm_gen_data.id,
+                interrupted=False,
+                created_at=reply_started_at,
             )
-            self._agent._chat_ctx.items.append(msg)
+            idx = self._agent._chat_ctx.find_insertion_index(created_at=reply_started_at)
+            self._agent._chat_ctx.items.insert(idx, msg)
+            idx += 1
```
theomonnom, line 1273:
```suggestion
```

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1353,8 +1368,9 @@ def _on_first_frame(_: asyncio.Future) -> None:
                 )
             elif len(new_fnc_outputs) > 0:
                 # add the tool calls and outputs to the chat context even no reply is generated
-                self._agent._chat_ctx.items.extend(new_calls)
-                self._agent._chat_ctx.items.extend(new_fnc_outputs)
+                for msg in tool_messages:
+                    msg.created_at = reply_started_at
```
theomonnom, line 1372: We can just reuse the `find_insertion_index`, this is safer, otherwise we must ensure there is no async calls in-between, and this is too hard to guarantee (Imagine a user adding a new message)

theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1170,6 +1170,8 @@ async def _pipeline_reply_task(
             await utils.aio.cancel_and_wait(*tasks)
             return
 
+        reply_started_at = time.time()
```
theomonnom, line 1173: Is replacing the created_at really necessary? I think the fact that we now use `find_insertion_index` should be sufficient? We could reduce LOC

longcw: The tool messages may be created much earlier than the reply started, as the LLM starts when the reply is still in the queue. Update the timestamp of them to make sure the `created_at` is ordered in the chat ctx.

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -188,9 +190,34 @@ def add_message(
...
 
-        self._items.append(message)
+        if is_given(created_at):
+            idx = self.find_insertion_index(created_at=created_at)
+            self._items.insert(idx, message)
+        else:
+            self._items.append(message)
         return message
 
+    def insert_item(
```
longcw, line 200: added a `insert_item`, pls let me know if this makes sense

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -188,9 +190,34 @@ def add_message(
...
+        self,
+        item: ChatItem | list[ChatItem],
+        *,
+        created_at: NotGivenOr[float] = NOT_GIVEN,
```
theomonnom, line 204: Isn't the created_at here confusing? When inserting an item I would expect that every singular item.created_at is respected. 
I see that the docstrings say it is indeed the intent, but I feel like it isn't obvious.

longcw: I was just thinking respecting every singular item.created_at of a list is a little bit inefficient. but probably it's fine...

PR #2324, opened 2025-05-19
Expose all realtime model parameters

Shubhrakanti:
See [this slack thread](https://live-kit.slack.com/archives/C07TWVC0W4A/p1745985834276339) to see why we only allow these options for the realtime session to be modified.


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -705,6 +705,10 @@ def update_options(
         voice: NotGivenOr[str] = NOT_GIVEN,
         temperature: NotGivenOr[float] = NOT_GIVEN,
         turn_detection: NotGivenOr[TurnDetection | None] = NOT_GIVEN,
+        model: NotGivenOr[str] = NOT_GIVEN,
+        max_response_output_tokens: NotGivenOr[int | Literal["inf"] | None] = NOT_GIVEN,
+        modalities: NotGivenOr[list[Literal["text", "audio"]]] = NOT_GIVEN,
```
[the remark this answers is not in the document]
Shubhrakanti: Ah good catch we can't.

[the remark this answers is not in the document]
Shubhrakanti: It's kind of confusing because if you go to the [docs](https://platform.openai.com/docs/api-reference/realtime-client-events/session/update#realtime-client-events/session/update-session). One of the acceptable parameters is model, but in the example request they don't have model.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -90,6 +90,8 @@ class _RealtimeOptions:
     tool_choice: llm.ToolChoice | None
     input_audio_transcription: InputAudioTranscription | None
     turn_detection: TurnDetection | None
+    max_response_output_tokens: int | Literal["inf"] | None
+    modalities: list[Literal["text", "audio"]]
```
Shubhrakanti, line 94: @theomonnom @longcw I wonder if it makes sense at all to allow people to update `modalities`.

Right now we have a retry mechanism that gets triggered if we get text only output. https://github.com/livekit/agents/pull/2015

Shubhrakanti: If I do 

```python 

self.llm.update_options(
            modalities=["text"],
        )
```

I just get choppy audio output back.

Shubhrakanti: Right now if I do 

```python
self.llm.update_options(
            modalities=["text"],
        )
```

I just get choppy audio output back.

Shubhrakanti: Since we don't support a text only realtime session, It might be best to disable this?

longcw: yeah we only support audio mode right now. We can re-enable it when the realtime + tts is added.

Shubhrakanti: Okay then I'll remove this for now.

Shubhrakanti: removed

theomonnom (approved):

PR #2328, opened 2025-05-19
Support empty array of parameters when using raw_schema

koen-boost:
It's possible to have function tools that do not need any arguments. The current code fails when passing an empty parameters array.

### DIFF
--- a/.github/next-release/changeset-053d451c.md
+++ b/.github/next-release/changeset-053d451c.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
+Support empty array of parameters when creating a function tool using raw_schema (#2328)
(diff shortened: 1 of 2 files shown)


theomonnom (approved):

davidzhao:
this commit is breaking `function_tool`

```
% python basic_agent.py dev
  File "/Users/davidzhao/dev/livekit/agents/examples/voice_agents/basic_agent.py", line 30, in <module>
    class MyAgent(Agent):
    ...<31 lines>...
            return "sunny with a temperature of 70 degrees."
  File "/Users/davidzhao/dev/livekit/agents/examples/voice_agents/basic_agent.py", line 45, in MyAgent
    @function_tool
     ^^^^^^^^^^^^^
  File "/Users/davidzhao/dev/livekit/agents/livekit-agents/livekit/agents/llm/tool_context.py", line 178, in function_tool
    return deco(f)
  File "/Users/davidzhao/dev/livekit/agents/livekit-agents/livekit/agents/llm/tool_context.py", line 171, in deco
    name=name or func.__name__,
         ^^^^
UnboundLocalError: cannot access local variable 'name' where it is not associated with a value
````

PR #2332, opened 2025-05-19
fix tool choice

jayeshp19:
Currently, tool_choice is not usable when set to "required" or a specific function. This PR ensures tool_choice is reset to "auto" or "none" temporarily when passing back  tool response to the LLM, so it can be used practically.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1341,7 +1341,9 @@ class AgentActivity(RecognitionHooks):
                         chat_ctx=chat_ctx,
                         tools=tools,
                         model_settings=ModelSettings(
-                            tool_choice=model_settings.tool_choice if not draining else "none",
+                            tool_choice="none"
+                            if draining or model_settings.tool_choice == "none"
+                            else "auto",
                         ),
                         _tools_messages=[*new_calls, *new_fnc_outputs],
                     ),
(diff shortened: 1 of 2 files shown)


davidzhao:
> Currently, tool_choice is not usable when set to "required" or a specific function.

what do you mean by this? could you provide any examples that would illustrate the problem?

jayeshp19:
> > Currently, tool_choice is not usable when set to "required" or a specific function.
> 
> what do you mean by this? could you provide any examples that would illustrate the problem?


When using function calling, we typically need two LLM calls:

1. First call: the user query → LLM decides which function to call.
2. Second call: we pass the function result back to LLM for final response.

If `tool_choice="required"` or a specific function is set, then *every* call to the LLM will invoke the tool again, including the second call, where we just want it to continue the conversation using the tool's output. This leads to redundant or incorrect behavior.

Example:

```python
llm = openai.ChatOpenAI(tool_choice="required", tools=[get_weather])
```

Even after we call `get_weather()` and pass its output back to the model, it'll still try to call `get_weather()` again, instead of generating a final response.
This PR avoids that by resetting `tool_choice` to `"auto"` or `"none"` after the tool has been executed.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1341,7 +1341,9 @@ def _on_first_frame(_: asyncio.Future) -> None:
                         chat_ctx=chat_ctx,
                         tools=tools,
                         model_settings=ModelSettings(
-                            tool_choice=model_settings.tool_choice if not draining else "none",
+                            tool_choice="none"
```
longcw, line 1344: can you add a comment here, basically the tool response shouldn't have `tool_choice=required`

longcw (approved):

PR #2335, opened 2025-05-20
fix type check

longcw:


longcw:
fixed type issue for `livekit.agents`, will work on the plugins as next step.

livekit-agents/livekit/agents/cli/log.py
```diff
@@ -106,7 +106,7 @@ def default(self, o: Any):
                 except Exception:
                     return None
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
```
Shubhrakanti, line 110: Thoughts on [`Any` vs `object`](https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object)?

> If you’re not sure whether you need to use [object](https://docs.python.org/3/library/functions.html#object) or Any, use [object](https://docs.python.org/3/library/functions.html#object) – only switch to using Any if you get a type checker complaint.

Some more context: https://stackoverflow.com/questions/75127785/mypy-seems-to-think-that-args-kwargs-could-match-to-any-funtion-signature

Shubhrakanti (approved):

livekit-agents/livekit/agents/cli/log.py
longcw: tested with object, for this case it cannot pass the type check with error 
```bash
livekit-agents/livekit/agents/cli/log.py:110: error: Argument 1 to "__init__" of "Formatter" has incompatible type "*tuple[object, ...]"; expected "Optional[str]"  [arg-type]
livekit-agents/livekit/agents/cli/log.py:110: error: Argument 1 to "__init__" of "Formatter" has incompatible type "*tuple[object, ...]"; expected "Literal['%', '{', '$']"  [arg-type]
livekit-agents/livekit/agents/cli/log.py:110: error: Argument 1 to "__init__" of "Formatter" has incompatible type "*tuple[object, ...]"; expected "bool"  [arg-type]
livekit-agents/livekit/agents/cli/log.py:110: error: Argument 2 to "__init__" of "Formatter" has incompatible type "**dict[str, object]"; expected "Optional[str]"  [arg-type]
livekit-agents/livekit/agents/cli/log.py:110: error: Argument 2 to "__init__" of "Formatter" has incompatible type "**dict[str, object]"; expected "Literal['%', '{', '$']"  [arg-type]
livekit-agents/livekit/agents/cli/log.py:110: error: Argument 2 to "__init__" of "Formatter" has incompatible type "**dict[str, object]"; expected "bool"  [arg-type]
Found 6 errors in 1 file (checked 122 source files)
```

I think it makes sense to use `Any` for most of the cases.

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -93,7 +93,7 @@ def chat(
 
     async def aclose(self) -> None: ...
 
-    async def __aenter__(self) -> LLM:
+    async def __aenter__(self) -> LLM[Any]:
```
theomonnom, line 96: Any way to avoid this? Could it just be LLM

mypy.ini
```diff
@@ -0,0 +1,4 @@
+[mypy]
+
+[mypy-mcp.*]
+ignore_missing_imports = True
```
theomonnom, line 4: Can we remove this file and move this to our pyproject.toml

theomonnom: https://github.com/livekit/agents/blob/8ff1efa7d7543c1e9f07f9b1b27da3a5e870968b/pyproject.toml#L77

theomonnom:
I don't think we need to ignore the MCP types, we need to update the CI to install livekit-agents[mcp] before running mypy

longcw:
> I don't think we need to ignore the MCP types, we need to update the CI to install livekit-agents[mcp] before running mypy

It's our ci running in python 3.9 does have mcp installed, for python 3.10 it's not needed

theomonnom:
> It's our ci running in python 3.9 does have mcp installed, for python 3.10 it's not needed

ah I see, let's bump our CI version

longcw:
> > It's our ci running in python 3.9 does have mcp installed, for python 3.10 it's not needed
> 
> 
> 
> ah I see, let's bump our CI version

If we still support python 3.9 maybe we can add type check for both, sometimes the type annotation like X|Y is not supported and will fail at the runtime in 3.9, so a check is still needed IMO.

PR #2338, opened 2025-05-20
Add diarization support

ShayneP:
This PR adds support for STT-provided diarization, meaning that a single track can have multiple speakers logged in the transcription.

Today, the only plugin that supports realtime diarization is Speechmatics, but I imagine there will be more widespread support for this in the future.

I tried to be minimally invasive here, if people want the speaker to be passed to the LLM they can add it into the STT node.

### DIFF
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -44,6 +44,7 @@ class SpeechData:
     start_time: float = 0.0
     end_time: float = 0.0
     confidence: float = 0.0  # [0, 1]
+    speaker: str | None = None
 
 
 @dataclass
--- a/livekit-agents/livekit/agents/voice/events.py
+++ b/livekit-agents/livekit/agents/voice/events.py
@@ -88,6 +88,7 @@ class UserInputTranscribedEvent(BaseModel):
     type: Literal["user_input_transcribed"] = "user_input_transcribed"
     transcript: str
     is_final: bool
+    speaker: str | None = None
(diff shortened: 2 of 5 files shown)


livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -44,6 +44,7 @@ class SpeechData:
     start_time: float = 0.0
     end_time: float = 0.0
     confidence: float = 0.0  # [0, 1]
+    speaker: str | None = None
```
theomonnom, line 47: Should speaker be a int instead?

theomonnom: makes sense, can we rename it to speaker_id?

theomonnom, line 47:
```suggestion
    speaker_id: str | None = None
```

theomonnom (approved):

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -88,6 +88,7 @@ class UserInputTranscribedEvent(BaseModel):
     type: Literal["user_input_transcribed"] = "user_input_transcribed"
     transcript: str
     is_final: bool
+    speaker: str | None = None
```
theomonnom, line 91:
```suggestion
    speaker_id: str | None = None
```

raghavjaistra:
Why did this support did not reflect in the latest release today (livekit-agents 1.0.23)??

PR #2343, opened 2025-05-20
[Rime] Increase timeout for `arcana` model to allow for synthesis of long audio 

MaCaki:
## What changed
The new `arcana` model can take much longer than `mistv2` due to it's architecture. As long as 80% of the duration of the actual audio. 

demonstrated by this curl request:  

```bash
curl -v -X POST "https://users.rime.ai/" \
    -H "Authorization: Bearer $RIME_API_KEY" \
    -H 'Accept: audio/wav' \
    -H 'Content-Type: application/json' \
    -d '{
        "text": "oh, totally! like, um, okay so there was this one time, um, me and my friends went to this art festival in san francisco, right? and it was, like, super fun, but also kinda chaotic <laugh>. so, like, we were checking out all these booths with, like, amazing art and stuff, and I- I even brought along my sketchbook, um, to doodle, you know? so, like, at one point, I see this artist who was doing live paintings and, like, Ive never seen anything like it, right? anyway, Im standing there just vibing, and I accidentally bumped into this guy who was, like, super into astrology like me! <laugh>. and we started talking about our signs, and he was a gemini, and, like, for some reason we, like, bonded over how geminis always get called two-faced, and I was just like, dude, chill, we\u2019re just versatile, you know?",
        "speaker":"celeste",
        "modelId": "arcana"
    }' \
    --output celeste_long_text.mp3

100 2856k    0 2856k  100   901  73169     22  0:00:40  0:00:39  0:00:01 75809
```
0:00:39 for a 0:01:00 long audio.

### DIFF
--- a/livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
+++ b/livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
@@ -39,6 +39,11 @@ from .langs import TTSLangs
 from .log import logger
 from .models import ArcanaVoices, TTSModels
 
+# arcana can take as long as 80% of the total duration of the audio it's synthesizing.
+ARCANA_MODEL_TIMEOUT = 60 * 4
+MISTV2_MODEL_TIMEOUT = 30
+RIME_BASE_URL = "https://users.rime.ai/v1/rime-tts"
+
 
 @dataclass
 class _TTSOptions:
@@ -73,7 +78,7 @@ class TTS(tts.TTS):
     def __init__(
         self,
         *,
-        base_url: str = "https://users.rime.ai/v1/rime-tts",
+        base_url: str = RIME_BASE_URL,
         model: TTSModels | str = "arcana",
         speaker: NotGivenOr[ArcanaVoices | str] = NOT_GIVEN,
         # Arcana options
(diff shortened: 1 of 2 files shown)


examples/avatar_agents/audio_wave/__init__.py
davidzhao, line 1: these shouldn't be checked in

.github/workflows/ci.yml
```diff
@@ -49,4 +49,4 @@ jobs:
         run: uv sync --all-extras --dev
 
       - name: Check Types
-        run: uv run mypy --install-types --non-interactive .
+        run: uv run mypy --install-types --non-interactive --explicit-package-bases .
```
davidzhao, line 52: what's the reason for this change?

livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
```diff
@@ -133,6 +138,8 @@ def __init__(
         self._session = http_session
         self._base_url = base_url
 
+        self._tts_http_timeout = ARCANA_MODEL_TIMEOUT if model == "arcana" else MISTV2_MODEL_TIMEOUT
```
davidzhao, line 141: nit: let's call this `total_timeout` ? to match aiohttp parameter

tts feels redundant in context of the TTS class

davidzhao (commented):
this is a good call. thanks for the PR

.github/workflows/ci.yml
MaCaki: same as above, was fiddling with things to understand why a typecheck workflow step was failing.

livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
MaCaki: yea good point.

MaCaki: changed

davidzhao (approved):
lgtm!

PR #2357, opened 2025-05-22
fix(deepgram): mip_opt_out was passed incorrectly for TTS

davidzhao:


davidzhao:
nvm, i was on an old branch. this is already fixed

Issue #2341, opened 2025-05-20
Feature request: Add a auto reconnect/reset for the realtime model


bml1g12:
When the OpenAI Realtime API reaches 30 minutes, the session expires and leads to an outage of the agent. I would love to have either a graceful handling of this available, or a recommended work around.

I could see a few possible solutions, like gracefully handing over the context to a new session just before the session expires, or catching the relevant error(s) when the timeout is reached.

As far as I am aware, the Azure flavour of the model raises the following in v0 livekit:

```
voc-voice-agent-livekit-dev  | 2025-04-08 03:46:05,944 - ERROR livekit.plugins.openai.realtime - response generation failed {"code": null, "error": "The server had an error while processing your request. Sorry about that! You can retry your request, or contact us through our help center at aoai-realtimeapi-dri@microsoft.com if the error persists. (Please include the session ID sess_BJtlOCDH4TegSOFQWOTHw in your message.)", "session_id": "sess_BJtlOCDH4TegSOFQWOTHw", "pid": 18, "job_id": "AJ_ZZasQ5mPo87N"}
```

whereas the OpenAI flavour might give a clearer error:

```
[2025-05-20 07:49:03,436] 1313 livekit.plugins.openai.realtime - ERROR - OpenAI S2S error {'type': 'error', 'event_id': 'event_BZC0diqUciYtBOplwd5i9', 'error': {'type': 'invalid_request_error', 'code': 'session_expired', 'message': 'Your session hit the maximum duration of 30 minutes.', 'param': None, 'event_id': None}} [2025-05-20 07:49:03,437] 21 livekit.plugins.openai.realtime - ERROR - Error in _recv_task
```

**Related issues**

Related to: https://github.com/livekit/agents/issues/1073
And this thread:  https://livekit-users.slack.com/archives/C07FY8WHGPM/p1747732664342909
CWilson suggested I open a feature request for this.

longcw:
duplicates https://github.com/livekit/agents/issues/1252

Paawan13:
is there any solution to this till date?

longcw:
It's added in #2360

Paawan13:
> It's added in [#2360](https://github.com/livekit/agents/pull/2360)

2025-06-01 16:01:17,769 - ERROR livekit.agents - unhandled exception while running the job task
Traceback (most recent call last):
  File "/root/Desktop/breezeflow-backend-python-main/livekit-voice-main/agent.py", line 176, in entrypoint
[...]

Issue #2358, opened 2025-05-22
Add a auto reconnect/reset for the OpenAI Realtime model


Parva101:
After 30mins of session. When using OpenAI Realtime 4o mini model with MultimodalAgent, we are facing this error: 

`[2025-05-20 07:49:03,436] 1313 livekit.plugins.openai.realtime - ERROR - OpenAI S2S error {'type': 'error', 'event_id': 'event_BZC0diqUciYtBOplwd5i9', 'error': {'type': 'invalid_request_error', 'code': 'session_expired', 'message': 'Your session hit the maximum duration of 30 minutes.', 'param': None, 'event_id': None}} [2025-05-20 07:49:03,437] 21 livekit.plugins.openai.realtime - ERROR - Error in _recv_task`

Also currently are there any workarounds for this error? We have already tried using the FallbackAdapter  method as suggested by @yepher but couldn't make it work.

longcw:
fixed in https://github.com/livekit/agents/pull/2360

Issue #1252, opened 2024-12-18
Reset the realtime model connection


longcw:
The realtime API session may expired in 15mins, we can add a method in `RealtimeSession` to reset the connection and sync the user/agent transcripts to the new connection.

Related discussion: https://community.openai.com/t/realtime-api-session-expired/975036/9

robin-intbot:
Is there any plan to add the issue/feature to roadmap?

PR #2360, opened 2025-05-22
add session_lifetime and auto reconnection for OAI realtime api

longcw:
fix https://github.com/livekit/agents/issues/2341 https://github.com/livekit/agents/issues/1252  https://github.com/livekit/agents/issues/2358

To enable the auto reconnection after x mins
```
llm=openai.realtime.RealtimeModel(session_lifetime_s=20 * 60)  # 20mins
```


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -489,6 +503,58 @@ def send_event(self, event: RealtimeClientEvent | dict) -> None:
...
+                    break
+            finally:
+                if not wait_reconnect_task.done():
+                    wait_reconnect_task.cancel()
+
+            if self._current_generation:
+                # wait for the current generation to complete
+                await self._current_generation._closed_fut
+
+            logger.debug("reconnecting realtime session")
```
longcw, line 536: should we emit an event when reconnecting and reconnected?

theomonnom:
Hey, I need a reconnect task is the wrong pattern, can we do something similar to how we handle our TTS reconnection.

Basically a loop inside the main_task.


Can you explain what is the `_text` logic needed for?

longcw:
>
> 
> Can you explain what is the `_text` logic needed for?

The agent reply was not updated to our remote conversation items before, it's needed to restore the chat history.

longcw:
> Hey, I think a reconnect task is the wrong pattern, can we do something similar to how we handle our TTS reconnection.
> 
> Basically a loop inside the main_task.

updated

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -206,6 +211,7 @@ def __init__(
         temperature: NotGivenOr[float] = NOT_GIVEN,
         tool_choice: NotGivenOr[llm.ToolChoice | None] = NOT_GIVEN,
         http_session: aiohttp.ClientSession | None = None,
+        session_lifetime_s: float | None = None,
```
theomonnom, line 214:
```suggestion
        max_session_duration: float | None = None,
```

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -83,6 +83,8 @@ async def aclose(self) -> None: ...
     "input_speech_stopped",  # serverside VAD
     "input_audio_transcription_completed",
     "generation_created",
+    "session_reconnected",
+    "session_restored",
```
theomonnom, line 87: I don't we need this event, any reason to keep it? ideally users just use session_reconnected
```suggestion
```

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -505,6 +514,50 @@ def send_event(self, event: RealtimeClientEvent | dict) -> None:
...
+                    )
+
+                restore_task.add_done_callback(_on_restored)
+
+            try:
+                await self._main_task()
+            finally:
+                if restore_task:
+                    await utils.aio.cancel_and_wait(restore_task)
+
```
theomonnom, line 558: This isn't synchronized, if the user called generate_reply just before a reconnection is happening, the new generation isn't going to have the new chat_ctx.

The solution here is to just not call the main_task and to not use our queue and send messages individually.
I think the main_task can be replaced by a `async def run_ws(ws: aiohttp.Websocket...): `

There is no new task anymore (no _restore_chat_context)
```suggestion
        @utils.log_exceptions(logger=logger)
        async def _restore_chat_context() -> None:
            # sync chat messages to the new session


        reconnecting = False
        while not self._msg_ch.closed:
            # Open the websocket here!
            ws = ...
            
            if reconnecting:
                chat_ctx = self.chat_ctx.copy(
                    exclude_function_call=True, exclude_instructions=True, exclude_empty_message=True
                )
                self._remote_chat_ctx = llm.remote_chat_context.RemoteChatContext()
                # send update without using our self.updat_chat_ctx
                
                await ws.send({...})
                
                # emit event when everything is done
                logger.debug("reconnected to OpenAI Realtime API, restoring chat context")
                self.emit("session_reconnected", llm.RealtimeSessionReconnectedEvent())

             await self._run_ws(ws)
```

Ofc this need cleanup but this is the idea

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -507,6 +520,68 @@ def send_event(self, event: RealtimeClientEvent | dict) -> None:
...
+                            ),
+                            recoverable=False,
+                        ),
+                    )
+                    raise e
+
+                logger.debug("reconnected to OpenAI Realtime API")
+                self.emit("session_reconnected", llm.RealtimeSessionReconnectedEvent())
+
+            await self._run_ws(ws_conn)
```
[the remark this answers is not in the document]
longcw: maybe in a separate pr with the [gemini reconnection](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py#L851-L856)?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -797,42 +899,54 @@ async def update_chat_ctx(
...
+            if _build_event:
+                return events
 
             try:
                 await asyncio.wait_for(asyncio.gather(*futs, return_exceptions=True), timeout=5.0)
             except asyncio.TimeoutError:
                 raise llm.RealtimeError("update_chat_ctx timed out.") from None
 
-    async def update_tools(self, tools: list[llm.FunctionTool | llm.RawFunctionTool]) -> None:
+    async def update_tools(
```
theomonnom, line 944: if this easy to change, I generally like to avoid the double-purpose methods

theomonnom (approved):
otherwise lgtm!

Issue #2361, opened 2025-05-22
Bug Report: Incorrect transcription_delay calculation results in extremely large values


hsjun99:
## Summary
The `transcription_delay` parameter occasionally receives extremely large values (e.g., `1747888559.8039`) that appear to be timestamps instead of proper delay calculations.

## Environment
- **STT**: Streaming STT
- **Livekit Agent Version**: 1.0.22

## Expected Behavior
`transcription_delay` should be a small positive value representing the delay in seconds.

## Actual Behavior
Sometimes `transcription_delay` gets values like `1747888559.8039` which look like Unix timestamps.

## Code Context
The issue occurs in this code:

```python
committed = await self._hooks.on_end_of_turn(
    _EndOfTurnInfo(
        new_transcript=self._audio_transcript,
        transcription_delay=max(
            self._last_final_transcript_time - last_speaking_time, 0
        ),
        end_of_utterance_delay=time.time() - last_speaking_time,
    )
)
```

## Suspected Cause
`last_speaking_time` appears to be `0` sometimes, causing `transcription_delay` to equal `self._last_final_transcript_time` (which is a timestamp).

## Reproduction
Cannot reproduce intentionally, but happens quite often during normal usage with Streaming STT.

longcw:
This may happen when VAD enabled but VAD didn't detect the speech started (like a false negative of vad) before STT returns the final transcription. Will fix it soon.

hsjun99:
@longcw Thanks for the reply !

zaheerabbas-prodigal:
I have notice these large values occur in the TTS Metrics too. Has anyone else seen that. I am using Cartesia TTS.

Pls lmk if I should open a new issue for that.

hsjun99:
@zaheerabbas-prodigal Haven't seen any when using Elevenlabs.

PR #2363, opened 2025-05-22
support stt END_OF_SPEECH for stt turn detection

longcw:
when `turn_detection="stt"`, it will respect to the `stt.SpeechEventType.END_OF_SPEECH` for stt turn detection.


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -246,13 +250,21 @@ async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
...
         elif ev.type == stt.SpeechEventType.INTERIM_TRANSCRIPT:
             self._hooks.on_interim_transcript(ev)
             self._audio_interim_transcript = ev.alternatives[0].text
 
+        elif ev.type == stt.SpeechEventType.END_OF_SPEECH and self._turn_detection_mode == "stt":
+            self._user_turn_committed = True
+            if not self._speaking:
+                # start response after vad fires END_OF_SPEECH to avoid vad interruption
+                chat_ctx = self._hooks.retrieve_chat_ctx().copy()
+                self._run_eou_detection(chat_ctx)
```
theomonnom, line 266: Is this reliable? What if VAD is wrong and the user isn't speaking? Will it still answer?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -65,8 +68,9 @@ def __init__(
         self._turn_detector = turn_detector
         self._stt = stt
         self._vad = vad
-        self._manual_turn_detection = manual_turn_detection
-        self._user_turn_committed = False
+        self._turn_detection_mode = turn_detection_mode
+        self._vad_base_turn_detection = turn_detection_mode in ("vad", None)
+        self._user_turn_committed = False  # true if user turn ended but EOU task not done
```
theomonnom, line 73: Our EOU logic is currently stateless, can we keep it this way or it is to avoid a double speech?

livekit-agents/livekit/agents/voice/audio_recognition.py
theomonnom: Our interruption may also be too aggressive.

longcw: if VAD didn't speak any speaking, the STT END_OF_SPEECH will fire the eou detection directly as `self._spkeaing` is False. This is the same logic as the STT final transcription event.

longcw: But there indeed a bug caused by the false negative of VAD https://github.com/livekit/agents/issues/2361, the last_speaking_time will be wrong if VAD didn't detect a speech.

livekit-agents/livekit/agents/voice/audio_recognition.py
longcw: It's to allow appending transcript when we are waiting for the EOU timeout. It's mainly for the `manual` turn detection, also valid for `stt` mode when any final transcription comes after the STT END_OF_SPEECH but before EOU timeout.

longcw: > Ideally we don't need states because we know only on of the task can run at a time with the bouncing logic

so we can check the task running for the same purpose?

longcw: checked the code, I think `self._user_turn_committed` is still needed, to indicate "user turn is ended" by manual or by stt END_OF_SPEECH event, there could be no EOU task in this case.

If the turn detection mode is manual or stt, this flag is used to decide if we should fire a new EOU task in VAD.END_OF_SPEECH or STT.FINAL_TRANSCRIPT events.

theomonnom: Ah that makes sense! Lgtm

theomonnom (approved):

Issue #2346, opened 2025-05-21
Elevenlabs issue livekit v1


youssefwalid7:
Bug description: run the agent in a dockerized environment when the TTS service starts speaking it pronounces the words letter by letter especially in any language other than English. 

Docker version:  28.0.4
Python version: python:3.12-slim
Livekit version: latest

Elevenlabs configuration:  
             model="eleven_multilingual_v2",
             voice_id="nPczCjzI2devNBz1zQrb",
             voice_settings=elevenlabs.VoiceSettings(
             stability=0.5,
             similarity_boost=0.64,
             style=0.0,
              speed=1.0,
              use_speaker_boost=False,
             ),

longcw:
can you share a audio recording with the issue? I cannot reproduce the issue with your settings using the [basic agent example](https://github.com/livekit/agents/blob/main/examples/voice_agents/basic_agent.py)
```
        tts=elevenlabs.TTS(
            model="eleven_multilingual_v2",
            voice_id="nPczCjzI2devNBz1zQrb",
            voice_settings=elevenlabs.VoiceSettings(
                stability=0.5, similarity_boost=0.64, style=0.0, speed=1.0, use_speaker_boost=False
            ),
        ),
```

yepher:
These are the sample audio files from OP

[sample_audio.zip](https://github.com/user-attachments/files/20369190/sample_audio.zip)

longcw:
do you have the transcription of those audios?

rashad-leapai:
I had an agent that use elevenlabs TTS, streaming was working for non-English. but since last update of livekit.plugins.elevenlabs the streaming broken, it only work for English now.

Can someone confirm whether updates raised this issue, whether its from livekit or elevenlabs side?

longcw:
@youssefwalid7 @rashad-leapai could you share an example transcript with the issue and the versions of livekit agents that works if possible.

rashad-leapai:
@longcw This resolved elevenlabs TTS streaming for non-English, in my case: 
livekit-agents[openai, elevenlabs, deepgram, silero, turn-detector]>=1.0.0,<=1.0.11

longcw:
I have found the issue, will create a fix soon. Thanks!

PR #2366, opened 2025-05-22
disable split characters for tts

longcw:
disable split characters for STT, fix https://github.com/livekit/agents/issues/2346


theomonnom:
Does it means users have to do something different when using languages like Thay/Chinese?

longcw:
No, I just didn't realized the word tokenizer is used for tts, we should split the characters for tts.

theomonnom (approved):

PR #2373, opened 2025-05-22
revert OpenAI text recovery

theomonnom:
we're going to use the OpenAI TTS instead.

This reduces code complexity


longcw:
are you going to implement the TTS support in this pr? if now I think we  should wait until that before merging this one. The text recovery actually works well as expected rn.

theomonnom:
> are you going to implement the TTS support in this pr? if not I think we should wait until that before merging this one. The text recovery actually works well as expected rn.

Yep, I'll do

longcw:
replaced by https://github.com/livekit/agents/pull/2628

PR #2397, opened 2025-05-26
fix transcription sync when on_playback_finished missing after flush

longcw:
when `on_playback_finished` called more times than capture, the extra `on_playback_finished` was ignored with the following and won't be propagated to the transcription synchronizer
```python
        if self.__playback_finished_count >= self.__playback_segments_count:
            logger.warning(
                "playback_finished called more times than playback segments were captured"
            )
            return
```

The synchronizer may not rotate the `_SegmentSynchronizerImpl` when this happens that causes errors like `RuntimeError: livekit.agents.tokenize.token_stream.BufferedSentenceStream is closed` when capturing the audio and text

### DIFF
--- a/livekit-agents/livekit/agents/voice/transcription/synchronizer.py
+++ b/livekit-agents/livekit/agents/voice/transcription/synchronizer.py
@@ -161,6 +161,14 @@ class _SegmentSynchronizerImpl:
     def closed(self) -> bool:
         return self._close_future.done()
 
+    @property
+    def audio_input_ended(self) -> bool:
+        return self._audio_data.done
+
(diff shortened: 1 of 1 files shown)


longcw:
> When is it possible for on_playback_finished to be called more time than a capture? Should we do the opposite and guard against that? There should be no on_playback_finished if there is no data pushed because this means we're emitting an event that shouldn't even exist? since there is no data

For some implementation issue in the avatar plugin for example, that should be fixed. But this pr is to make sure the agent reply is not crashed due to that.

theomonnom (approved):

PR #2401, opened 2025-05-26
fix(google): Live syncs context, supports manual turns

### LINKED ISSUE #1594: Gemini Live API tools support
I noticed that the package only supports function declarations for using tools with the Gemini Live APi, is there a planned support for the other tools types described in gemini's API, such as code execution and google search?

### LINKED ISSUE #2300: Ground and elastic search for Gemini LLM and Gemini live
So i looked into googles Groundings: https://cloud.google.com/vertex-ai/generative-ai/docs/grounding/overview
[-> Grounding with Google Search](https://cloud.google.com/vertex-ai/generative-ai/docs/grounding/grounding-with-google-search)
[-> Grounding with Elasticsearch](https://cloud.google.com/vertex-ai/generative-ai/docs/grounding/grounding-with-elasticsearch)
[and even Grounding with Google Maps](https://cloud.google.com/vertex-ai/generative-ai/docs/grounding/grounding-with-google-maps)

Are any of the features above are support in gemini LLM  or gemini live?
This is a wanted feature ,
Please implement this and inform us about this or tell us how we should do it.

Thank you

davidzhao:
- added `start_user_activity` to notify model when user starts speaking
- support for using LiveKit turn detection
- fixed using the same id for input & output
- added chat context syncing locally, better coverage of update_chat_ctx
- enables context window compression
- supports Gemini tools (googleSearch, etc) (note: this is still early on the model side)

closes #2300
closes #1594


livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -73,6 +75,7 @@ def __init__(
         frequency_penalty: NotGivenOr[float] = NOT_GIVEN,
         tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
         thinking_config: NotGivenOr[types.ThinkingConfigOrDict] = NOT_GIVEN,
+        gemini_tools: NotGivenOr[list[LLMTool]] = NOT_GIVEN,
```
theomonnom, line 78: I'm not sure about this pattern, feels like this should be inside `Agent.tools` instead

davidzhao: how do we handle the different types by diff vendors?

I'll be glad to back this change out if we are thinking about an alternative path

vishnukool:
Is there a sample code on how to turn on search grounding in livekit for gemini ?

PR #2405, opened 2025-05-27
Making auto_tool_reply_generation parameter configurable for the Gemini RealtimeModel

DeninSiby:
This PR adds the ability to configure auto_tool_reply_generation for the Gemini Realtime Model. Previously, this was hardcoded to True, which meant the model would automatically generate responses for tool calls. Now, users can control this behavior by setting auto_tool_reply_generation=False when initializing the model.

## Why is this useful?
1. *Custom Tool Response Handling*: When auto_tool_reply_generation=False, developers can implement their own custom logic for handling tool responses. This is particularly useful when:
   - You need to perform additional validation or processing of tool results
   - You want to implement custom error handling for tool calls
   - You need to aggregate results from multiple tool calls before generating a response

2. *Fine-grained Control*: This change gives developers more control over the conversation flow, allowing them to:
   - Decide when and how to respond to tool calls
   - Implement custom retry logic for failed tool calls
   - Add additional context or processing before generating responses

3. *Integration Flexibility*: This makes it easier to integrate the Gemini Realtime Model with existing systems that have their own tool response handling mechanisms.

## Example Usage
```
python
model = RealtimeModel(
    auto_tool_reply_generation=False,  # Disable automatic tool response generation
    # ... other configuration options ...
)

```

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
@@ -132,7 +132,7 @@ class RealtimeModel(llm.RealtimeModel):
         frequency_penalty: NotGivenOr[float] = NOT_GIVEN,
         input_audio_transcription: NotGivenOr[AudioTranscriptionConfig | None] = NOT_GIVEN,
         output_audio_transcription: NotGivenOr[AudioTranscriptionConfig | None] = NOT_GIVEN,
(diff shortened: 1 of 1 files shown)


DeninSiby:
@davidzhao Can you check this PR

Denin-Siby:
This helps as a workaround for this issue to: [gemini tool call issue](https://github.com/livekit/agents/issues/2174) @davidzhao 

I also did try not returning anything and generating a reply instead but the model behaviour is very random in that case. That is sometimes it generates a reply sometimes it hallucinates and generates an answer.

PR #2419, opened 2025-05-28
fix type checks and tts fallback adapter

longcw:


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -330,6 +330,9 @@ async def _recv_task(ws: aiohttp.ClientWebSocketResponse):
                     output_emitter.push(b64data)
                 elif data.get("done"):
                     current_segment_id = None
+                    # TODO(long): when input not ended, it will raise
+                    # "start_segment() called before the previous segment was ended"
+                    # since current_segment_id is reset to None while `end_input` is not called?
```
longcw, line 335: @theomonnom this seems to be a bug but I am not sure how sure we fix it, always call `end_input` or only reset the segment_id only input end?

PR #2423, opened 2025-05-28
fix type checks for plugins

longcw:
- fix all type checks
- fix openai and google stt for the new ConnectionPool `async def connection(self, *, timeout: float) -> AsyncGenerator[T, None]`
- fix raw function tools for anthropic and aws llm


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -198,7 +202,7 @@ def __init__(
         entra_token: str | None = None,
         api_key: str | None = None,
         api_version: str | None = None,
-        base_url: str | None = None,
+        base_url: NotGivenOr[str] = NOT_GIVEN,
```
theomonnom, line 205: nit, I generally prefer to use `str | None` over NotGivenOr when we don't need to. 
(we tend to over-use NotGiven in our codebase, but it's OK)

theomonnom (approved):
🙌

Issue #2451, opened 2025-05-30
Wrong agent state when interrupting due to new lk.chat sendText


brianzaubar:
We realized if the agent in the STT-LLM-TTS pipeline is being interrupted the agent switches from `speaking` to `thinking` and then directly (automatically) from `thinking` to `listening` instead of switching directly from `speaking` to `thinking`. 
And then he switches directly from `listening` to `speaking`, which does not make any sense, since the agent would always **think** before he **speaks**, wouldn't he? 
This is very confusing, especially since he is in a `listening` state even though he is currently `thinking`.

We are using the latest livekit-agents 1.0 package and livekit-client@2.11.4.

davidzhao:
@brianzaubar feel free to use the main branch if this is blocking. we'll be making a release sometimes next week

PR #2453, opened 2025-05-30
fix agent state for pipeline agent

longcw:
fix https://github.com/livekit/agents/issues/2451

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -511,7 +511,7 @@ class AgentActivity(RecognitionHooks):
             SpeechCreatedEvent(speech_handle=handle, user_initiated=True, source="say"),
         )
 
-        self._create_speech_task(
+        task = self._create_speech_task(
             self._tts_task(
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -690,6 +690,9 @@ async def _main_task(self) -> None:
                 await speech.wait_for_playout()
                 self._current_speech = None
 
+            if not isinstance(self.llm, llm.RealtimeModel) and not self._speech_tasks:
+                self._session._update_agent_state("listening")
+
```
theomonnom, line 695: Any way to not use this task, but use the pipeline_task instead? Maybe at the end?

longcw: maybe add a done callback to the pipeline and tts task and check the pending speech there?

longcw: updated https://github.com/livekit/agents/pull/2453/commits/84069918838f7bbe991681a446c7224779c9de19

PR #2459, opened 2025-05-30
use pyht SDK

theomonnom:


theomonnom:
merging even if it doesn't pass the tests to avoid breaking the user API for the new release

PR #2460, opened 2025-05-30
Add custom request timeout for OpenAI TTS

anishnag:
We want to be able to modify the OpenAI request timeout, so we expose it as a parameter in the OpenAI `TTS` object.


davidzhao:
thanks for the PR @anishnag! could you run `ruff format` on this?

theomonnom:
We recently added tts_conn_options inside the AgentSession, you should be able to edit this parameter from there.
https://github.com/livekit/agents/blob/eb017ef363824314ae914836dd27a6a6a45bab31/livekit-agents/livekit/agents/voice/agent_session.py#L44

theomonnom:
I think another issue here is that the 30s default isn't a good one. We should never timeout if we continue to receive data from the server. (We don't want a a max request time if there is still audio)

PR #2492, opened 2025-06-03
add artificial delay between consecutive speech handles

longcw:


theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -119,6 +120,7 @@ def __init__(
         max_tool_steps: int = 3,
         video_sampler: NotGivenOr[_VideoSampler | None] = NOT_GIVEN,
         user_away_timeout: float | None = 15.0,
+        min_consecutive_speech_delay: float = 0.0,
```
davidzhao, line 123: is this the right default? or should we consider raising?

longcw: I think the artificial delay is only required for some specific tts providers without any silent padding. So by default there is no delay.

PR #2503, opened 2025-06-04
Add new speed and tracing options to OpenAI RealtimeModel and RealtimeSession

mikevin920:


theomonnom:
I see the right type would be to use: https://github.com/openai/openai-python/blob/56540b32873df335aca9270715a839c2a9770639/src/openai/types/beta/realtime/session.py#L90

It does seems weird tho: `TracingTracingConfiguration` 🤔

kevin-mii:
> I see the right type would be to use: https://github.com/openai/openai-python/blob/56540b32873df335aca9270715a839c2a9770639/src/openai/types/beta/realtime/session.py#L90
> 
> It does seems weird tho: `TracingTracingConfiguration` 🤔

Yeah using this type makes more sense. Also which file do I change so that we can require the latest openai version to support this

MajorTal:
This is awesome - can't wait!

longcw:
> > I see the right type would be to use: https://github.com/openai/openai-python/blob/56540b32873df335aca9270715a839c2a9770639/src/openai/types/beta/realtime/session.py#L90
> > It does seems weird tho: `TracingTracingConfiguration` 🤔
> 
> Yeah using this type makes more sense. Also which file do I change so that we can require the latest openai version to support this

Should it be `session_update_event.SessionTracing` https://github.com/openai/openai-python/blob/56540b32873df335aca9270715a839c2a9770639/src/openai/types/beta/realtime/session_update_event.py#L112?
The session update is using `Session` from `SessionUpdateEvent`
```python
        return SessionUpdateEvent(
            type="session.update",
            # Using model_construct since OpenAI restricts voices to those defined in the BaseModel.  # noqa: E501
            # Other providers support different voices, so we need to accommodate that.
            session=session_update_event.Session.model_construct(**kwargs),  # type: ignore
            event_id=utils.shortuuid("session_update_"),
        )
```

theomonnom (approved):

theomonnom:
> > > I see the right type would be to use: https://github.com/openai/openai-python/blob/56540b32873df335aca9270715a839c2a9770639/src/openai/types/beta/realtime/session.py#L90
> > > It does seems weird tho: `TracingTracingConfiguration` 🤔
> > 
> > 
> > Yeah using this type makes more sense. Also which file do I change so that we can require the latest openai version to support this
> 
> Should it be `session_update_event.SessionTracing` https://github.com/openai/openai-python/blob/56540b32873df335aca9270715a839c2a9770639/src/openai/types/beta/realtime/session_update_event.py#L112? The session update is using `Session` from `SessionUpdateEvent`
> 
> ```python
>         return SessionUpdateEvent(
>             type="session.update",
>             # Using model_construct since OpenAI restricts voices to those defined in the BaseModel.  # noqa: E501
>             # Other providers support different voices, so we need to accommodate that.
>             session=session_update_event.Session.model_construct(**kwargs),  # type: ignore
>             event_id=utils.shortuuid("session_update_"),
>         )
> ```

It was initially using it in the first commit. But I don't think it's the right type (knowing that every other types like input_audio_noise_reduction & turn_detection is not using types from the session_update in our constructor)

longcw:
> It was initially using it in the first commit. But I don't think it's the right type (knowing that every other types like input_audio_noise_reduction & turn_detection is not using types from the session_update in our constructor)

But `SessionUpdateEvent` is using a `Session` defined in `session_update_event` and the `tracing` is annotated as `SessionTracing` from there too. Why that's not the correct type?

longcw:
> knowing that every other types like input_audio_noise_reduction & turn_detection is not using types from the session_update in our constructor

When creating the session update event they are converted to the correct type from it https://github.com/livekit/agents/blob/livekit-agents%401.0.23/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py#L675-L699

PR #2505, opened 2025-06-04
run room_io.start and ctx.connect concurrently in session.start

longcw:


longcw:
`room_io.start` for now is just hooking the the callbacks. so actually the `session.start` including `room_io.start` inside only takes 6ms on my computer. Do we still need this?

theomonnom:
> `room_io.start` for now is just hooking the the callbacks. so actually the `session.start` including `room_io.start` inside only takes 6ms on my computer. Do we still need this?

I think since the STT connection is done in the background (after calling `stream()`), the 6ms doesn't include this delay?

longcw:
> I think since the STT connection is done in the background (after calling `stream()`), the 6ms doesn't include this delay?

Yes, the STT connection is done in AudioRecognition when initialize the `stt_node`.

theomonnom:
> > I think since the STT connection is done in the background (after calling `stream()`), the 6ms doesn't include this delay?
> 
> Yes, the STT connection is done in AudioRecognition when initialize the `stt_node`.

So I think it's worth it, stream() must be called as early as possible

bcherry:
is there any implication in how avatar sessions hook into this? I noticed that our basic example now does session.start then ctx.connect, to allow for instant audio. but the avatar examples do ctx.connect then avatar.start then session.start, since avatar.start requires the room connection. if we make this change, where should avatar.start fit?

longcw:
> is there any implication in how avatar sessions hook into this? I noticed that our basic example now does session.start then ctx.connect, to allow for instant audio. but the avatar examples do ctx.connect then avatar.start then session.start, since avatar.start requires the room connection. if we make this change, where should avatar.start fit?

This is a good point. We need to access `local_participant.identity`  in `avatar_session.start`, it seems there is no to start the avatar session before join the room?

bcherry:
local participant identity should be knowable from the job context before connection since it's in the token, right?

longcw:
> local participant identity should be knowable from the job context before connection since it's in the token, right?

no, the local participant is created after room is connected. I am not sure if there is a way to specify the local participant identity before that.

bcherry:
i think participant identity is [set by the server](https://github.com/livekit/livekit/blob/758e176250feb5865557cedbeda6ba318710bf96/pkg/agent/worker.go#L370) during job assignment so as long as you have the token and api secret you can [decode it](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/worker.py#L792) to extract the participant identity, no?

theomonnom:
> i think participant identity is [set by the server](https://github.com/livekit/livekit/blob/758e176250feb5865557cedbeda6ba318710bf96/pkg/agent/worker.go#L370) during job assignment so as long as you have the token and api secret you can [decode it](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/worker.py#L792) to extract the participant identity, no?

Yes, you can know it beforehand, but I'm not sure if we expose anything to access it

longcw:
@bcherry @theomonnom updated the avatar plugins, they can start before joining the room now.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -392,28 +393,34 @@ async def start(
...
-                except RuntimeError:
-                    pass  # ignore
+            except RuntimeError:
+                pass  # ignore
 
             # it is ok to await it directly, there is no previous task to drain
-            await self._update_activity_task(self._agent)
+            tasks.append(asyncio.create_task(self._update_activity_task(self._agent)))
+
+            await asyncio.gather(*tasks)
```
theomonnom, line 423: Can we cancel the tasks in case of a failure?

theomonnom (approved):

PR #2513, opened 2025-06-05
Fix Google plugin realtime interrupt

amon-ra:
When using Google Gemini realtime plugin if agent is interrupted it keeps listening forever. This fix this error

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py
@@ -819,9 +819,9 @@ class RealtimeSession(llm.RealtimeSession):
             current_gen._completed_timestamp = time.time()
 
         if server_content.interrupted:
+            self._mark_current_generation_done()
             self._handle_input_speech_started()
-
-        if server_content.turn_complete:
+        elif server_content.turn_complete:
             self._mark_current_generation_done()
 
     def _mark_current_generation_done(self) -> None:


davidzhao:
I'm unable to reproduce it.. could you share a repro please?

amon-ra:
With this code:

```python
 async def entrypoint(ctx: agents.JobContext):
    log.info("starting entrypoint")
    await ctx.connect(
        # auto_subscribe=AutoSubscribe.AUDIO_ONLY
    )
    session = agents.AgentSession(
            llm=google.beta.realtime.RealtimeModel(
                model="gemini-2.0-flash-exp",
                voice="Puck",
                temperature=0.8,
                instructions="You are a helpful assistant",
            ),
    )

    await session.start(
        # room=ctx.room,
        room=ctx.room,
        agent=Assistant(),
        # room_input_options=RoomInputOptions(
        #     # LiveKit Cloud enhanced noise cancellation
        #     # - If self-hosting, omit this parameter
        #     # - For telephony applications, use `BVCTelephony` for best results
        #     noise_cancellation=noise_cancellation.BVC(),
        # ),
    )

    await session.generate_reply(instructions="Greet the user and offer your assistance.")
```


When gemini is interrupted, you speak before agent end of speak then it stop responding.

dophlin:
@amon-ra dear, did you find a solution to fix this issue?

pashanitw:
Why hasn’t this been merged yet? I believe we’re still encountering the issue with Gemini Live interruptions. When you interrupt the agent mid-speech, it stops speaking entirely. This is easy to reproduce — just interrupt while the Realtime API is active. It throws this warning and then stops:
WARNING livekit.plugins.google - truncate is not supported by the Google Realtime API

PR #2514, opened 2025-06-05
fix tts fallback adapter test and stream adapter

longcw:


davidzhao (commented):
great catch.. thanks for fixing this. will let @theomonnom give the final ✅

livekit-agents/livekit/agents/tts/stream_adapter.py
```diff
@@ -102,8 +102,7 @@ async def _synthesize() -> None:
                 ) as tts_stream:
                     async for audio in tts_stream:
                         output_emitter.push(audio.frame.data.tobytes())
-
-                    output_emitter.flush()
```
theomonnom: We can't do that actually. The reason is that one request may take quite some time to finish. 

And when a new request starts, it'll start playing the old request but 5s later if it was slow

longcw: okay in that case we need a flush there, but this will make the StreamAdapter always add a 0.01s empty audio to the output when we call the `end_input`, any way to avoid that or it's just okay...

theomonnom: It's okay, it's a limitation that we have, we have no way of signaling the end of a segment otherwise.

PR #2518, opened 2025-06-05
[draft] test cases for agent session and agent activity

longcw:


tests/test_agent_session.py
```diff
@@ -0,0 +1,614 @@
...
+
+
+def check_timestamp(
+    t_event: float, t_target: float, *, speed_factor: float = 1.0, max_abs_diff: float = 0.5
+) -> None:
+    t_event = t_event * speed_factor
+    print(
+        f"check_timestamp: t_event: {t_event}, t_target: {t_target}, max_abs_diff: {max_abs_diff}"
+    )
+    assert abs(t_event - t_target) <= max_abs_diff, (
```
theomonnom, line 577:
```suggestion
    assert abs(t_event - t_target) <= max_abs_diff * speed_factor, (
```

longcw: the `t_event` is already scaled by the speed_factor, and `t_target` is the ground truth without scaling.

PR #2524, opened 2025-06-06
feat(openai): strip thinking tokens

davidzhao:
when using reasoning models, we are currently speaking out thinking tokens.

this change strips all content between `<think>....</think>`

### DIFF
--- a/livekit-agents/livekit/agents/llm/utils.py
+++ b/livekit-agents/livekit/agents/llm/utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import base64
 import inspect
 import sys
@@ -37,6 +38,9 @@ from .tool_context import (
 
 if TYPE_CHECKING:
     from ..voice.events import RunContext
+
+THINK_TAG_START = "<think>"
+THINK_TAG_END = "</think>"
 
 
 def _compute_lcs(old_ids: list[str], new_ids: list[str]) -> list[str]:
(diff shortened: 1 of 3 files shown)


longcw (approved):

livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -400,3 +404,23 @@ def _shallow_model_dump(model: BaseModel, *, by_alias: bool = False) -> dict[str
         key = field.alias if by_alias and field.alias else name
         result[key] = getattr(model, name)
     return result
+
+
+def strip_thinking_tokens(content: str | None, thinking: asyncio.Event) -> str | None:
```
theomonnom, line 409: I'm wondering if there is a better way to do this. We end up having a lot of duplicated logic across all LLMs for that.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -685,7 +687,9 @@ async def _run(self) -> None:
         except Exception as e:
             raise APIConnectionError(retryable=retryable) from e
 
-    def _parse_choice(self, id: str, choice: Choice) -> llm.ChatChunk | None:
+    def _parse_choice(
+        self, id: str, choice: Choice, thinking: asyncio.Event
```
theomonnom, line 691: I'm wondering if we can just store a self._thinking state like we do for other states we have.

livekit-agents/livekit/agents/llm/utils.py
davidzhao: this is shared and not duplicated.. the duplication is how to pass in the provider specific streams. 

where would you place this stripping logic? we could keep it a feature of the superclass, by having a helper function to produce chunks

theomonnom: yeah maybe the superclass. But I guess it's fine for now

PR #2526, opened 2025-06-06
fix: LLM to honor custom timeouts

davidzhao:
previously, timeouts passed to `LLM.chat` were not being used.

also improves fallback adapter defaults.

### DIFF
--- a/livekit-agents/livekit/agents/llm/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/fallback_adapter.py
@@ -38,9 +38,10 @@ class FallbackAdapter(
         self,
         llm: list[LLM],
         *,
-        attempt_timeout: float = 10.0,
-        max_retry_per_llm: int = 1,
-        retry_interval: float = 5,
+        attempt_timeout: float = 5.0,
+        # use fallback instead of retrying
+        max_retry_per_llm: int = 0,
+        retry_interval: float = 0.5,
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/llm/fallback_adapter.py
```diff
@@ -38,9 +38,10 @@ def __init__(
         self,
         llm: list[LLM],
         *,
-        attempt_timeout: float = 10.0,
-        max_retry_per_llm: int = 1,
-        retry_interval: float = 5,
+        attempt_timeout: float = 5.0,
+        # use fallback instead of retrying
+        max_retry_per_llm: int = 0,
+        retry_interval: float = 0.5,
```
theomonnom, line 44: Fwiw a lot of those arguments were removed for the TTS FallbackAdapter.

I think directly retrying after one failure isn't a good tradeoff between reliability and speed. A lot of time, for OpenAI at least, one request fails and the next one directly resolves

davidzhao: here are the trade offs for the two cases:

1. transient failures (retry would fix it)
2. permanent failures (retry would not)

when it's 2, with the default it would mean the agent would remain silent for 10+ seconds.. this is way too long in conversational AI.

I'm ok with a retry if we've reduced timeout to 2.5s

PR #2538, opened 2025-06-09
Add Cartesia STT integration

DineshTeja:
- Implemented STT for livekit's cartesia plugin (includes logic for both batch and streaming STT)
- Local audio energy filter configuration
- Speech event handling
- Proper termination logic with finalize, flush_done, done, etc.
- Added STT model definitions and language support in models.py
- API version for STT: 2025-04-16


davidzhao (approved):
lgtm!

PR #2541, opened 2025-06-10
bithuman avatar refresh token after prewarm

longcw:


examples/avatar_agents/bithuman/agent_worker.py
```diff
@@ -33,11 +33,17 @@ async def entrypoint(ctx: JobContext):
...
-    )
+    runtime: AsyncBithuman | None = ctx.proc.userdata.get("bithuman_runtime")
+    if runtime:
+        # refresh token if runtime is already initialized
+        await runtime._initialize_token()
+    else:
+        runtime = await AsyncBithuman.create(
+            model_path=bithuman_model_path,
+            api_secret=bithuman_api_secret,
+        )
```
theomonnom: Can this be done inside the BitHuman plugin instead? Ideally users don't need to think about this

theomonnom (approved):

PR #2542, opened 2025-06-10
use not given for room io options

longcw:


examples/avatar_agents/bey/agent_worker.py
```diff
@@ -34,8 +26,6 @@ async def entrypoint(ctx: JobContext):
     await session.start(
         agent=Agent(instructions="Talk to me!"),
         room=ctx.room,
-        # audio is forwarded to the avatar, so we disable room audio output
-        room_output_options=RoomOutputOptions(audio_enabled=False),
```
davidzhao, line 38: are these set automatically now?

longcw: yes, if `session.output.audio` is set, we set this to False automatically if it's `NOT_GIVEN` or log a warning if it's True.

PR #2544, opened 2025-06-10
add connect options and error retry for realtime model

longcw:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1457,25 +1491,26 @@ def _handle_error(self, event: ErrorEvent) -> None:
...
             extra={"error": event.error},
         )
+        self._emit_error(
+            APIError(
+                message="OpenAI Realtime API returned an error",
+                body=event.error,
+                retryable=True,
+            ),
+            recoverable=True,
+        )
```
theomonnom, line 1501: It's kind of a grey area here, because errors from the OpenAI realtime server are generally always safe unrecoverable & never fatal in our case. So retryable/recoverable is misleading. Tho I'm not sure how we can improve this.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1457,25 +1491,26 @@ def _handle_error(self, event: ErrorEvent) -> None:
...
+                error=error,
+                recoverable=recoverable,
             ),
         )
 
-        # if event.error.event_id:
-        #     fut = self._response_futures.pop(event.error.event_id, None)
-        #     if fut is not None and not fut.done():
-        #         fut.set_exception(multimodal.RealtimeError(event.error.message))
-
```
theomonnom, line 1478: I don't think we should remove that? Should we?

theomonnom (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
theomonnom: I guess it's unused..

longcw: they are commented? what is that

theomonnom: Yeah today we only timeout on messages that fail. Ideally we propagate the errors tho.

for e.g: if we send bad inputs when updating the chat_context, we can also listen to errors coming from their server.
And use them to set the exception on our pending future. Today we ignore them

longcw: I see, maybe add a TODO there instead

PR #2548, opened 2025-06-10
cancel tasks on start failure

theomonnom:


theomonnom:
we may also need to close the "already" started components

Issue #2564, opened 2025-06-11
Missing Metrics from Cartesia STT


ChenghaoMou:
The Cartesia STT does not send any RECOGNITION_USAGE event and therefore no metrics would be reported.

PR #2565, opened 2025-06-11
add Cartesia STT usage event

ChenghaoMou:
This should address #2564:

- Add input duration from frame duration values (similar to AssemblyAI)
- Send usage event when it is final


theomonnom (approved):

theomonnom:
Thanks!

PR #2566, opened 2025-06-11
Feature: add try except if there could be problems when tty and sounddevice are not available

wlbksy:
On some virtual machines, the `tty` and `sounddevice` code may throw errors due to the absence of corresponding virtualization hardware. Here, a `try-except` block is added to prevent such errors.

### DIFF
--- a/.github/next-release/changeset-76274fba.md
+++ b/.github/next-release/changeset-76274fba.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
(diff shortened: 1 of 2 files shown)


davidzhao:
console mode wouldn't function without a mic & speaker.. so this isn't super useful to swallow the error.

wlbksy:
It is useful if you want to use fake stt and fake tts to implement unit test. However, I'm ok with this.

PR #2568, opened 2025-06-11
Remove unnecessary model param from baseten tts

bcherry:
This parameter isn't needed (and wasn't used) as the model_endpoint is the correct method to choose among your deployed models with baseten.

(not tested yet - @davidzhao I sent a request to join our org so I can verify and finish up the docs)


davidzhao (approved):

davidzhao:
@bcherry could you run ruff first?

PR #2520, opened 2025-06-05
Baseten Livekit plugin integration

htrivedi99:

PR #2576, opened 2025-06-11
Fix baseten STT api key lookup

bcherry:
it was using `is_given` on a param that defaults to `None` which didn't work, so the env var lookup doesn't work. I changed it to match the way model_endpoint is checked and also how the TTS plugin is set up.


bcherry:
actually there's more broken with this STT than just this. will reopen in a bit

bcherry:
this plugin has [other issues](https://github.com/livekit/agents/pull/2520#issuecomment-2964349585) but I want to get this merged anyways so I don't have to fix it later

theomonnom (approved):

Issue #2326, opened 2025-05-19
Feature Request: Utterance timestamps in the ChatContext or Transcript


zaheerabbas-prodigal:
I want to capture the word level or atleast utterance level timestamps for both the `user` and `agent` transcripts and store the timestamp details in the `ChatContext`. The usecase is to use the `ChatContext` as the transcript to run some post-processing. Example use cases- displaying the calls in UI, redaction, summarization etc.

Currently the transcription data the SDK exposes for the user transcript is just the `text` part of the transcript and the utterance or word level timestamps are NOT exposed at all from the SDK.

The agent's transcript however does not even have the timestamps even though [elevenlabs](https://elevenlabs.io/docs/api-reference/text-to-speech/v-1-text-to-speech-voice-id-stream-input#receive.Audio-Output.alignment) and [cartesia](https://docs.cartesia.ai/2025-04-16/api-reference/tts/tts#receive.Receive.timestamps) TTS support timestamps in their API.

Has anyone tried a way to get these timestamp data from the livekit-agents SDK?

I am happy to submit a PR to add this feature. Would the PR be merged if I added this feature?

hsjun99:
@zaheerabbas-prodigal How have you ended up regarding this issue?

longcw:
It's added in https://github.com/livekit/agents/pull/2580, it supports the word level timestamp for cartesia TTS or sentence level timestamp for non-streaming TTS using StreamAdapter.

We are not adding the timed transcripts to the chat context, but you can access the it from the `transcription_node` of the agent when it's supported by the TTS.

hsjun99:
@longcw Great news thanks. Would it also work for elevenlabs?
Also, what is the best strategy to create a word-level timestamp(relative X, absolute O) for the user transcript?

longcw:
elevenlabs is not supported right now as we are using the websocket api and they only support timestamps in post api https://elevenlabs.io/docs/api-reference/text-to-speech/stream-with-timestamps

hsjun99:
@longcw Isn't the normalizedAlignment/Alignment field that provides the timestamps?
![Image](https://github.com/user-attachments/assets/409ef766-dc07-444a-9dd9-e540db732e07)

longcw:
oh yeah I'll take a look

zaheerabbas-prodigal:
Thanks for adding this functionality - @longcw 

> We are not adding the timed transcripts to the chat context

Any reason for not adding this to ChatContext?

[...]

PR #2580, opened 2025-06-12
support timed transcripts from tts

longcw:
Fix https://github.com/livekit/agents/issues/2607, https://github.com/livekit/agents/issues/2326#issuecomment-2975325213
needs https://github.com/livekit/python-sdks/pull/456


livekit-agents/livekit/agents/voice/io.py
```diff
@@ -31,28 +48,24 @@
...
+            Union[
+                AsyncIterable[rtc.AudioFrame],
+                tuple[AsyncIterable[rtc.AudioFrame], Optional[AsyncIterable[TimedString]]],
+            ]
+        ],
+        Awaitable[
+            Optional[
+                Union[
+                    AsyncIterable[rtc.AudioFrame],
+                    tuple[AsyncIterable[rtc.AudioFrame], Optional[AsyncIterable[TimedString]]],
```
theomonnom: let's mode that to some hidden userdata on the AudioFrame? or a io.TimedAudio.

Otherwise it lgtm!

longcw: added a pr to python sdk https://github.com/livekit/python-sdks/pull/456

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1089,6 +1076,12 @@ def _on_first_frame(_: asyncio.Future[None]) -> None:
                     model_settings=model_settings,
                 )
                 tasks.append(tts_task)
+                if (
+                    (tts := self.tts)
+                    and (tts.capabilities.timed_transcript or not tts.capabilities.streaming)
```
longcw: I have a concern here if user created new `AudioFrame` in a customized `tts_node` but didn't forward the timed transcripts, we may missing the text response. wdyt? @theomonnom

livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -233,8 +234,12 @@ def llm_node(
         return Agent.default.llm_node(self, chat_ctx, tools, model_settings)
 
     def transcription_node(
-        self, text: AsyncIterable[str], model_settings: ModelSettings
-    ) -> AsyncIterable[str] | Coroutine[Any, Any, AsyncIterable[str]] | Coroutine[Any, Any, None]:
+        self, text: AsyncIterable[str | TimedString], model_settings: ModelSettings
```
theomonnom, line 247: TimedString is a str, to be fair I'm not even sure if we should encourage people to use the timed transcripts here
```suggestion
        self, text: AsyncIterable[str], model_settings: ModelSettings
```

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -697,9 +722,11 @@ def _emit_frame(frame: rtc.AudioFrame | None = None, *, is_final: bool = False)
                             is_final=True,
                         )
                     )
+                    timed_transcripts = []
                     return
 
             if last_frame is not None:
+                last_frame.user_data["timed_transcripts"] = timed_transcripts
```
theomonnom: Is the idea to send the timed transcripts at the end of segment? 
if so maybe this could just be a new field to SynthesizedAudio. (This also means that we don't have synchronized transcripts until the whole generation is done?)

longcw: It's not at the end of segment, usually at the start of the tts.

longcw: the buffered `timed_transcripts` are always added to the next audio frame, not only the last frame.

livekit-agents/livekit/agents/voice/agent.py
longcw: do you have any suggestions for the alternative for ppl to access the timed transcripts?

tbachlechner:
just adding my support for this one, excited to see you hopefully get this out soon!

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -144,17 +146,30 @@ async def _inference_task() -> bool:
             tts_node = await tts_node
 
         if isinstance(tts_node, AsyncIterable):
+            timed_text_ch = aio.Chan[io.TimedString]()
+            timed_texts_fut.set_result(timed_text_ch)
+
             async for audio_frame in tts_node:
+                for text in audio_frame.userdata.get("timed_transcripts", []):
```
theomonnom: Let's move the key as a constant somewhere. Like https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/types.py

(with `lk.` prefix)

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -491,5 +519,37 @@ def _stream_url(opts: _TTSOptions) -> str:
         url += f"&language_code={language}"
     if is_given(opts.streaming_latency):
         url += f"&optimize_streaming_latency={opts.streaming_latency}"
+    if opts.sync_alignment:
+        url += "&sync_alignment=true"
 
     return url
+
+
+def _to_timed_words(
```
theomonnom, line 547: It's a bit difficult to understand this code without actually running it. Not urgent, but it would be ideal to have some tests in place at some point

theomonnom:
Should we release a new version of livekit-rtc before merging?

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -41,6 +44,8 @@ class SynthesizedAudio:
 class TTSCapabilities:
     streaming: bool
     """Whether this TTS supports streaming (generally using websockets)"""
+    timed_transcript: bool = False
```
theomonnom: we use `aligned_transcript` on other parts of the code (AgentSession has `use_tts_aligned_transcript`). Let's use `aligned` here too?

longcw: updated

sansjack:
> Should we release a new version of livekit-rtc before merging?

when is the EST for this to be added?

examples/voice_agents/timed_agent_transcript.py
```diff
@@ -0,0 +1,54 @@
...
+# This example shows how to obtain the timed transcript from the TTS.
+# Right now, it's supported for Cartesia and ElevenLabs TTS (word level timestamps)
+# and non-streaming TTS with StreamAdapter (sentence level timestamps).
+
+
+class MyAgent(Agent):
+    def __init__(self):
+        super().__init__(instructions="You are a helpful assistant.")
+
+        self._closing_task: asyncio.Task[None] | None = None
```
theomonnom, line 27:
```suggestion
```

livekit-agents/livekit/agents/tts/stream_adapter.py
```diff
@@ -29,14 +29,14 @@ def __init__(
...
-                streaming=True,
-            ),
+            capabilities=TTSCapabilities(streaming=True, aligned_transcript=True),
             sample_rate=tts.sample_rate,
             num_channels=tts.num_channels,
         )
         self._wrapped_tts = tts
-        self._sentence_tokenizer = sentence_tokenizer or tokenize.blingfire.SentenceTokenizer()
+        self._sentence_tokenizer = sentence_tokenizer or tokenize.blingfire.SentenceTokenizer(
+            retain_format=True
```
theomonnom, line 38: Oh actually we were not using retain_format for the StreamAdapter before. Since it is only used to generate a sentence. 

In the PR I did, I was actually keeping the basic.SentenceTokenizer inside the transcription synchronization code.

livekit-agents/livekit/agents/voice/agent.py
theomonnom: IMO we should just keep it implicit for now

livekit-agents/livekit/agents/tts/stream_adapter.py
longcw: It was used in `agent`'s tts_node

longcw: or maybe I added it in this pr, we need to format if we use the timed transcript from stream adapter.

theomonnom: Ah ok, the synchronizer also needs the exact same formatting?

longcw: no, it can be different. They process the sentences separately.

longcw: btw, I added `ev.token.strip()` for the sentence sent to TTS in the adapter, so the TTS should accept the same data, it's just to keep the format for transcription node.

theomonnom: I see, but I thought the aligned transcripts returned by the TTSs were not including new lines/special characters. So I assumed retain_format was not needed.

theomonnom: When using the StreamAdapter with OpenAI, the transcription_node is coming from the llm_node right?
In this case I really don't think we should wait for the TTS? Since we have the opt-in flag `use_tts_aligned_transcript`

longcw: If use tts aligned transcript enabled, the input of the transcription node is from tts.

livekit-agents/livekit/agents/voice/agent.py
longcw: > IMO we should just keep it implicit for now

I understand the concern this may change in the future, but we should expose the timed transcripts to user in some way IMO, this was asked by some folks for awhile. Any alternatives for this?

livekit-agents/livekit/agents/tts/stream_adapter.py
longcw: wdym for we shouldn't wait for the tts when using steam adapter?

theomonnom: Ok that makes sense, so by default, even if we use the StreamAdapter, it'll use the llm output for the transcription_node

PR #2591, opened 2025-06-13
Hume plugin fixes

zgreathouse:
**Changes to Hume Plugin**

- Updates `PostedContext` type to support specifying context with a `generation_id`
    - This is the more common way to specify [context](https://dev.hume.ai/reference/text-to-speech-tts/synthesize-json#request.body.context.ContextGenerationId.generation_id) in Hume's TTS API.
- Updates plugin to no longer default `context` to `None`, despite presence of user provided `context`.
- Updates default sample rate from `24kHz` to `48kHz` and removes the `sample_rate` option.
    - Sample rate should not be configurable for this plugin, because the Hume TTS API only outputs audio with a sample rate of 48kHz. The sample rate of speech audio is not currently configurable through the API.)
- Add `audio_format` option to make audio format configurable.
    - The plugin currently defaults to 'mp3' when the Hume TTS API also supports 'wav' and 'pcm'.
- Removes default `speed` and `trailing_silence` from base utterance in favor of API defaults when `speed` or `trailing_silence` options are not specified.

### DIFF
--- a/.github/next-release/changeset-4fe732da.md
+++ b/.github/next-release/changeset-4fe732da.md
@@ -1,0 +1,6 @@
+---
+"livekit-agents": patch
+"livekit-plugins-hume": patch
+---
+
+Hume plugin fixes (#2591)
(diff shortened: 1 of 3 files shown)


zgreathouse:
@theomonnom I saw you made some changes recently to improve this plugin's code. Mind taking a look at these changes?

theomonnom (approved):

Issue #2581, opened 2025-06-12
PTT: Groq STT returning empty transcripts if end_turn called before STT completes


s-hamdananwar:
When using the LiveKit push-to-talk example with Groq STT, if end_turn is triggered before the user's audio has been transcribed, the transcript is lost — `on_user_turn_completed` is called with an empty transcript.

Repro Steps:

1. Call start_turn → audio is enabled
2. User speaks
3. Call end_turn before STT finishes
4. commit_user_turn is called, but transcript is empty

PR #2597, opened 2025-06-13
wait for final transcript in manual turn detection

longcw:
fix https://github.com/livekit/agents/issues/2581

fix push to talk when using stt.StreamAdapter, slow STT and Google STT

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -57,6 +57,7 @@ class VoiceOptions:
     max_tool_steps: int
     user_away_timeout: float | None
     min_consecutive_speech_delay: float
+    commit_user_turn_timeout: float
 
 
 Userdata_T = TypeVar("Userdata_T")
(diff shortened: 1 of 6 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -130,6 +131,7 @@ def __init__(
         video_sampler: NotGivenOr[_VideoSampler | None] = NOT_GIVEN,
         user_away_timeout: float | None = 15.0,
         min_consecutive_speech_delay: float = 0.0,
+        commit_user_turn_timeout: float = 2.0,
```
theomonnom: I think this should be part of a new parameter for 'commit_turn' instead?

longcw: sounds good

longcw: updated

theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -614,12 +614,12 @@ def clear_user_turn(self) -> None:
 
         self._activity.clear_user_turn()
 
-    def commit_user_turn(self) -> None:
+    def commit_user_turn(self, *, final_transcript_timeout: float = 2.0) -> None:
```
theomonnom: IMO final isn't needed
```suggestion
    def commit_user_turn(self, *, transcript_timeout: float = 2.0) -> None:
```

PR #2608, opened 2025-06-16
Add SSML support for Google TTS

kechako:
This pull request adds optional SSML support to the Google TTS integration in LiveKit Agent.

Changes:
- Introduced a new `enable_ssml` option in the TTS class constructor.
- When `enable_ssml` is set to true, input text is wrapped in `<speak>...</speak>` tags and sent to the Google Text-to-Speech API as SSML.
- SSML is only applied to non-streaming TTS requests, as Google TTS does not support SSML in StreamingSynthesize.

Notes:
- The default behavior remains unchanged; plain text is used unless `enable_ssml` is explicitly set.
- Care should be taken to avoid enabling SSML with streaming requests, as this is not supported by the API and will result in an error.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
@@ -47,6 +47,8 @@ class _TTSOptions:
     speaking_rate: float
     tokenizer: tokenize.SentenceTokenizer
     volume_gain_db: float
+    enable_ssml: bool
+
 
 class TTS(tts.TTS):
     def __init__(
@@ -66,6 +68,7 @@ class TTS(tts.TTS):
         credentials_file: NotGivenOr[str] = NOT_GIVEN,
         tokenizer: NotGivenOr[tokenize.SentenceTokenizer] = NOT_GIVEN,
         use_streaming: bool = True,
+        enable_ssml: bool = False,
     ) -> None:
         """
         Create a new instance of Google TTS.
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -206,10 +214,21 @@ def __init__(self, *, tts: TTS, input_text: str, conn_options: APIConnectOptions
         self._tts: TTS = tts
         self._opts = replace(tts._opts)
 
+    def _build_ssml(self) -> str:
```
jayeshp19, line 217: we shouldn't manually build ssml, we should expect ssml in the stream instead. wdyt @theomonnom

tarekasishm:
Hi! 👋

I might be missing something, but after reviewing this solution, I get the feeling it could be risky in scenarios where LiveKit is used with TTS implementations that don't support streaming.

From what I understand, in those cases LiveKit simulates streaming by splitting the text using SequenceTokenizer. This could potentially lead to invalid SSML in certain situations. Here are a couple of examples I’ve come across:

- ✅ " <prosody rate=\"slow\">Can you hear me now?</prosody>"
→ correctly tokenized as: ["<prosody rate=\"slow\">Can you hear me now?</prosody>"]

- ❌ "But if we make this sentence a little bit longer <prosody rate=\"slow\">Can you hear me now?</prosody>"
→ ends up as: ["But if we make this sentence a little bit longer <prosody rate=\"slow\">Can you hear me now?", "</prosody>"]

This might result in malformed SSML being sent to the TTS engine.

Let me know what you think—happy to be corrected if I'm misunderstanding something!

PR #2611, opened 2025-06-16
add transcript_confidence to ChatMessage

theomonnom:


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/stt.py
```diff
@@ -41,6 +41,7 @@
 API_VERSION_HEADER = "Cartesia-Version"
 API_VERSION = "2025-04-16"
 
+
```
Shubhrakanti, line 44: might mess with your change set? Prolly only what this for the agents package?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -347,18 +352,26 @@ async def _bounce_eou_task(last_speaking_time: float) -> None:
             await asyncio.sleep(max(extra_sleep, 0))
 
             tracing.Tracing.log_event("end of user turn", {"transcript": self._audio_transcript})
+            confidence_avg = (
+                sum(self._final_transcript_confidence) / len(self._final_transcript_confidence)
+                if self._final_transcript_confidence
+                else 0
+            )
```
Shubhrakanti, line 359: Do you think it would be good to have a comment in `ChatMessage.transcription_confidence` on how it's calculated or you think it's implementation details - users shouldn't have to worry about it?

theomonnom: I don't think they should worry about it. But it may depends on the STT tho. I'm not 100% sure

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/stt.py
theomonnom: it's ruff

livekit-agents/livekit/agents/voice/audio_recognition.py
toubatbrian: What's the main purpose of adding transcript confidence? Just curious @theomonnom

PR #2617, opened 2025-06-16
Nova Sonic realtime plugin

BumaldaOverTheWater94:
Notes
Audio works but text and tool use are WIP
There are also schemas for Sonic events.


livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
```diff
@@ -0,0 +1,770 @@
...
+                )
+            self._current_generation.message_ch.send_nowait(
+            llm.MessageGeneration(
+                        message_id=msg_gen.message_id,
+                        text_stream=msg_gen.text_ch,
+                        audio_stream=msg_gen.audio_ch,
+                    )
+                )
+            self._current_generation.messages[self._current_generation.response_id] = msg_gen
+            self._emit_generation_event()
```
BumaldaOverTheWater94: @theomonnom

if I emit this event later it works, but not here, up top

Issue #2606, opened 2025-06-16
Bug: Incorrectly large transcription_delay when VAD is missed but STT provides a transcript (v1.1.1)


hsjun99:
**Describe the bug**
There are circumstances where the `transcription_delay` metric, calculated in `_EndOfTurnInfo`, is reported as an unexpectedly large value, sometimes in the range of 10-15 seconds.

This issue occurs when the Voice Activity Detector (VAD) fails to trigger for a user's utterance, but the Speech-to-Text (STT) engine successfully transcribes the audio and produces a final transcript.

**Root Cause Analysis**
The `transcription_delay` is calculated as `self._last_final_transcript_time - last_speaking_time`.

1.  `_last_speaking_time` is primarily updated by the VAD system upon detecting the end of speech (in `_on_vad_event` via `VADEventType.END_OF_SPEECH`).
2.  `_last_final_transcript_time` is updated when the STT engine emits a `FINAL_TRANSCRIPT` event (in `_on_stt_event`).
3.  If a user speaks, but the VAD does not fire, `_last_speaking_time` is not updated and retains its value from the previous utterance.
4.  However, the STT can still process the audio and produce a transcript. When this `FINAL_TRANSCRIPT` is received, `_on_stt_event` is called.
5.  Inside `_on_stt_event`, the code checks `if not self._speaking:`. Because VAD never triggered, `self._speaking` is `False`, and this block is entered.
6.  The `_run_eou_detection` function is then called, which uses the current (and now stale) `self._last_speaking_time` to schedule the `_bounce_eou_task`.
7.  The final calculation becomes `time.time()` (from the final transcript) minus `_last_speaking_time` (from the *previous* turn), resulting in a very large and incorrect delay value.

The problematic code section is in `AudioRecognition._on_stt_event`:

```python
# file: livekit-agents/livekit/agents/ipc/audio_recognition.py

# ... inside AudioRecognition class
    async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
        # ...
        if ev.type == stt.SpeechEventType.FINAL_TRANSCRIPT:
            # ...
            self._last_final_transcript_time = time.time()
            # ...

            if not self._speaking: # This is True if VAD didn't detect speech
                if not self._vad:
                    # This block only runs if VAD is completely disabled.
                    # The issue arises when VAD is enabled but misses an utterance.
                    self._last_speaking_time = time.time()

                if self._vad_base_turn_detection or self._user_turn_committed:
[...]

PR #2620, opened 2025-06-17
fix transcription delay when VAD false negative

longcw:
partially fix https://github.com/livekit/agents/issues/2361 and https://github.com/livekit/agents/issues/2606 when STT fired but VAD failed to detect the speech. In the middle of the conversation it may still use a out-of-date `last_speaking_time` if VAD has false negative.

### DIFF
--- a/livekit-agents/livekit/agents/metrics/base.py
+++ b/livekit-agents/livekit/agents/metrics/base.py
@@ -62,10 +62,14 @@ class EOUMetrics(BaseModel):
     type: Literal["eou_metrics"] = "eou_metrics"
     timestamp: float
     end_of_utterance_delay: float
-    """Amount of time between the end of speech from VAD and the decision to end the user's turn."""
+    """Amount of time between the end of speech from VAD and the decision to end the user's turn.
+    Negative value if the end of speech was not detected.
+    """
 
     transcription_delay: float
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/metrics/base.py
```diff
@@ -62,10 +62,14 @@ class EOUMetrics(BaseModel):
     type: Literal["eou_metrics"] = "eou_metrics"
     timestamp: float
     end_of_utterance_delay: float
```
davidzhao, line 64: do you think setting it to `None` vs a negative value would be more intuitive?

longcw: using `None` is more intuitive, but I am concerning this may break user code if they didn't check if it's None.

longcw: perhaps it's better to break it for this case...

davidzhao: right.. would we leave as 0 if unknown? I know it's not technically correct.. but something still bugs me about negative.. 

for example.. if you were adding these values together, it seems that unknown latency for one of the components shouldn't make it negative?

longcw: updated to 0

theomonnom (approved):

PR #2622, opened 2025-06-17
Disable ensure_ascii

tg-bomze:
Disable ensure_ascii (`livekit-agents/livekit/agents/cli/log.py`)

This update will help log user and agent transcripts in languages that use the Cyrillic character set.


theomonnom:
Thanks!

Issue #2550, opened 2025-06-10
Extend SpeechHandle with a list of tool calls


mrkowalski:
In unit testing scenario, it is currently not possible to use `AgentSession.generate_reply` to verify tool calls as it returns `SpeechHandle` that does not contain this information. The solution is to call llm node directly, but it circumvents session code flow, which I believe should not happen in testing.

There even is a todo for that in `SpeechHandle` already: _"# TODO(theomonnom): should we introduce chat_items property as well for generated tools?"_

I propose implementing this todo.

theomonnom:
This is going to be added in livekit-agents 1.2.0

Issue #1999, opened 2025-04-14
speech_id in metrics changes when tool calls are involved


yuyuma:
Using v1, I'm interested in measuring the conversation latency as defined in the [v0 docs](https://docs.livekit.io/agents/v0/build/metrics/#measuring-conversation-latency)

While the various ttfts and durations are exposed in the v1 metrics models, I'm unsure how to identify each "turn" of a conversation to correlate latencies of the same.  Is there an example that demonstrates how this might be implemented?

To add onto this, and assuming there is a way to identify a specific conversation turn, a cascading pipeline approach may involve the use of tool calls in which case the conversation latency needs to include the full duration of the tool calling LLM inference step.  Is there a way to easily identify the type of LLM call made through the metrics framework or would we have to rely heuristics (i.e. only count `ttft` of the last llm call and any prior llm calls count its full `duration`)?

ChenghaoMou:
From what I can see, it's bit tricky with STT. In some plugins, audio chunks are sent every x seconds nonstop, and responses from those services aren't linked back to any chunk timestamps. So it is not that straightforward to calculate the latency in STT. see #1135 

However, you can try to measure the time difference between some events to get a **rough** idea:
- user state change (speaking -> _) which marks the end of the user speech;
- agent state change (_ -> speaking) which marks the start of the agent speech;

LilaRest:
@ChenghaoMou Yes, same experience here. I’ll try your alternative solution, thanks for sharing!

Anyway LiveKit team, I’d love to see metrics a bit more polished and showcased in the docs, that could be helpful to a lot of devs after us! If I can help, ping me.

ChenghaoMou:
Right, EOU is not available in traditional STT scenarios. 

To your question, I think the LLM replying based on tool results is considered a new turn here:

![Image](https://github.com/user-attachments/assets/c6cef204-c59b-4857-86c6-0a3daa83246b)

A new speech handle is created when replying and it will have a new speech_handle/speech_id. But it will have the `parent` attribute pointing to the previous `speech_handle` and an increased `step_index`. However, I don't think they are exposed in metrics. It would be easier if a `parent_speech_id` is included.

davidzhao:
> However, in a conversation turn with a function call, the speech_id appears to change in the middle. Is this intentional?

This isn't intentional.. I think having a new speech_id kind of breaks the original intent of the feature. let us think through this one and will get back to you. I'm changing the title of this issue to better reflect this

[...]

PR #2623, opened 2025-06-17
reuse SpeechHandle for all generations inside a single turn

theomonnom:
multiple benefits:
- all items created by a generation (including tools) can be accessed using `SpeechHandle.chat_items`
- all generations inside a turn will have the same `speech_id` (now `turn_id==speech_id`)
- required by our evals framework 

handles: https://github.com/livekit/agents/issues/1999, https://github.com/livekit/agents/issues/2550


longcw:
it seems there is a bug that calling `session.say` in `on_user_turn_completed` will hang the agent. Noticed from the agent session unit tests... 

I can reproduce it with this example https://github.com/livekit/agents/blob/theo/reuse-speech-handle/examples/voice_agents/fast-preresponse.py

Issue #2488, opened 2025-06-03
Fallback Adapter not catching Gemini parsing failures


bryanhoulton:
I had a user today report a crashed room - when I looked into it there was an error from Gemini:
```
raise APIConnectionError(\nlivekit.agents._exceptions.APIConnectionError: gemini llm: error generating content Expecting property name enclosed in double quotes: line 1 column 2 (char 1)",
```
I traced this down and it doesn't seem to be a LiveKit issue - Gemini is returning bad JSON and Google's Python SDK isn't catching it properly. I found more information about this bug in [this issue on another repo](https://github.com/BerriAI/litellm/issues/5650).

Obviously this isn't LiveKit's fault, but it does end up crashing the worker,  even with a FallbackAdapter. Is it possible to include an error like this within the scope of the FallbackAdapter?

longcw:
could you share an example or more logs that can reproduce this issue with the `FallbackAdapter`, it works when I test with the following when I raise an error from the gemini LLM.
```python
        llm=FallbackAdapter(
            llm=[
                google.LLM(model="gemini-2.0-flash-001"),
                openai.LLM(model="gpt-4o-mini"),
            ]
        ),
```

bryanhoulton:
```
[...]

PR #2626, opened 2025-06-18
add error log when llm fallback adapter failed because chunk_sent

longcw:


longcw:
related to https://github.com/livekit/agents/issues/2488

theomonnom (approved):

Issue #2601, opened 2025-06-13
`generate_reply` doesn't work in `on_enter` when using tavus avatar


koakuma-chan:
Currently, `on_enter` is called after the avatar participant joins the room. 

However, it appears that is too early and `generate_reply` doesn't work:
```python
async def on_enter(self):
    # doesn't work
    await self.session.generate_reply(
        instructions="Greet the user with a warm welcome",
    )
```

The workaround is to wait for the avatar participant to publish its video track first, and by that time `generate_reply` starts to work.

PR #2627, opened 2025-06-18
wait for video track from avatar plugins

longcw:
fix https://github.com/livekit/agents/issues/2601

### DIFF
--- a/livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
+++ b/livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import asyncio
 import json
-import logging
 import math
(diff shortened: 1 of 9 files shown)


livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
```diff
@@ -54,7 +58,25 @@ async def start(self) -> None:
             self._room.local_participant.register_rpc_method(
                 RPC_PLAYBACK_FINISHED, self._handle_playback_finished
             )
+            logger.debug(
+                "waiting for the remote participant",
+                extra={"identity": self._destination_identity},
+            )
             await utils.wait_for_participant(room=self._room, identity=self._destination_identity)
```
theomonnom, line 65: We should maybe add a timeout in case the avatar provider is down?

PR #2628, opened 2025-06-18
[wip] support tts with realtime model (audio in, text out)

longcw:


livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -89,7 +94,30 @@ def compute_chat_ctx_diff(old_ctx: ChatContext, new_ctx: ChatContext) -> DiffOps
...
+
+        if old_item.type == "message" and new_item.type == "message":
+            is_same = old_item.text_content == new_item.text_content
+        else:
+            is_same = True
+
+        checked[(old_id, new_id)] = is_same
+        return is_same
+
+    lcs_ids = set(_compute_lcs(old_ids, new_ids, equal=equal))
```
theomonnom: Can we create a new `to_update` op instead? So for IDs that are the same, we do nothing, we just update the content if it changed

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -163,7 +164,9 @@ def interrupt(self) -> None: ...
 
     # message_id is the ID of the message to truncate (inside the ChatCtx)
     @abstractmethod
-    def truncate(self, *, message_id: str, audio_end_ms: int) -> None: ...
+    def truncate(
+        self, *, message_id: str, audio_end_ms: int, forwarded_text: str | None = None
```
theomonnom:
```suggestion
        self, *, message_id: str, audio_end_ms: int, audio_transcript: NotGivenOr[str] = NOT_GIVEN
```

livekit-agents/livekit/agents/llm/utils.py
longcw: updated

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1509,7 +1524,14 @@ async def _read_messages(
                         )
                         break
 
-                    tr_node = self._agent.transcription_node(msg.text_stream, model_settings)
+                    tts_text_input: AsyncIterable[str] | None = None
+                    if not self.llm.capabilities.audio_output and self.tts:
+                        tts_text_input, tr_text_input = utils.aio.itertools.tee(msg.text_stream)
```
theomonnom: I think the object returned by tee needs to be closed

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1509,7 +1524,14 @@ async def _read_messages(
...
+                    else:
+                        tr_text_input = msg.text_stream.__aiter__()
```
theomonnom, line 1666: We should still detect when the realtime API provides only text without audio, and in that case, fallback to TTS as well. Some users might prefer to keep using the S2S API since OpenAI can produce the same TTS voice as a fallback.

theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: right now I [raised an error](https://github.com/livekit/agents/pull/2628/files#diff-ff18b8060d7fec5a4f3840b596c32f3c8d4a4331794ce9909e0e685dd76ada0cR1246-L1344) and notify user to use text modality + TTS.

I am not sure if automatic fallback is expected... we need user to provide a TTS to the agent to allow fallback but usually ppl won't do that when using the realtime model. Should we automatically create a TTS for that?

theomonnom: I think it's fine for now, let's see if people ask about an automatic fallback.

examples/voice_agents/realtime_load_chat_history.py
```diff
@@ -41,7 +41,8 @@ async def entrypoint(ctx: JobContext):
     session = AgentSession()
     agent = Agent(
         instructions="You are a helpful travel planner.",
-        llm=openai.realtime.RealtimeModel(),
+        llm=openai.realtime.RealtimeModel(modalities=["text"]),
+        tts=openai.TTS(voice="alloy"),
```
theomonnom, line 48: Can you add a comment on why it's needed in this case (when adding chat items manually)

theomonnom (approved):
lgtm! this is awesome!

PR #2630, opened 2025-06-18
Cartesia TTS: gracefully handle websocket connection closing after long period of silence

mcuelenaere:
Cartesia seems to kill the WebSocket after 60 seconds of no messages being sent. If we wrap these exceptions as `APIError`, the built-in retry mechanism will automatically create a fresh connection.

This is similar to https://github.com/livekit/agents/issues/2281 and https://github.com/livekit/agents/pull/2282

### DIFF
--- a/.github/next-release/changeset-600494da.md
+++ b/.github/next-release/changeset-600494da.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-cartesia": patch
+---
+
+fix: gracefully handle WS conn closing after long period of silence (#2630)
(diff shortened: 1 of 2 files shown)


davidzhao:
I'm not able to reproduce this. could you share any steps that could repro?

exceptions from `_sentence_stream_task` should already be caught from the parent spawner [here](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py#L342)

mcuelenaere:
We were still on livekit 1.0.22, seems like this got fixed in a later version. Upgrading to 1.1.11 solved the issue for us, will close this now!

PR #2631, opened 2025-06-18
correctly passing speaking_rate to StreamingAudioConfig

david-rodriguez:
Previously, speaking_rate was being configured and stored in _TTSOptions, but it wasn't included in the StreamingAudioConfig used during streaming synthesis. As a result, the speech rate had no effect during streaming playback.


theomonnom (approved):
Thank you!

PR #2634, opened 2025-06-18
feat(plugin-inworldai): add inworld ai tts plugin

MichaelSolati:


livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
```diff
@@ -0,0 +1,287 @@
...
+    async def _process_audio_stream(
+        self,
+        stream: AsyncGenerator[models.SynthesizeSpeechResponse, None],
+        output_emitter: tts.AudioEmitter,
+    ) -> None:
+        """Process audio chunks from the synthesized stream."""
+        async for chunk in stream:
+            if chunk and chunk.get("audioContent"):
+                try:
+                    audio_data = base64.b64decode(chunk["audioContent"])
```
davidzhao: you don't need to parse this yourself, the output emitter would handle this for you, you can control it by passing in `audio/wav` as mime type

it's not great to duplicate wave parsing code in places.

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
```diff
@@ -0,0 +1,287 @@
...
+from __future__ import annotations
+
+import asyncio
+import base64
+import os
+from collections.abc import AsyncGenerator
+from dataclasses import dataclass, replace
+from typing import Literal, cast
+
+from inworld_sdk import InworldClient, models
```
davidzhao: i don't think we should import another SDK. often we see third-party SDKs not handling timeouts properly. and it can be a pain to work with.

if this is a standard HTTP API that produces data, we should just use that directly

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
```diff
@@ -0,0 +1,287 @@
...
+            language (TTSLanguageCodes, str, optional): The language code for synthesis.
+            voice (TTSVoices, str, optional): The voice to use.
+            speed (float, optional): The speed of the voice. Defaults to 1.0.
+            sample_rate (int, optional): The audio sample rate in Hz. Defaults to 24000.
+            base_url (str, optional): The base URL for the Inworld AI API.
+            auth_type (Literal["basic", "bearer"], optional): The authentication type to use.
+                Defaults to "basic".
+        """
+        super().__init__(
+            capabilities=tts.TTSCapabilities(streaming=True),
```
davidzhao: if it's using their HTTP endpoint (and not a websocket), the pattern would be to set streaming=False, and producing a ChunkedStream. in that case we would already automatically handle tokenization, so there isn't the need to duplicate it here

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
MichaelSolati: So I'm not getting audio when I try to pass the `chunk["audioContent"]` to the `output_emitter.push()` and initing it as such

```py
        output_emitter.initialize(
            request_id=utils.shortuuid(),
            sample_rate=self._opts.sampleRateHertz,
            num_channels=NUM_CHANNELS,
            mime_type="audio/wav",
            stream=True,
        )
 ```
 
 I'm guessing it's because the content when decoded is including it's mimetype (I've included the first 44 bytes of the decoded chunk for context)
 
 
 ```txt
 RIFF\xec'\x01\x00WAVEfmt \x10\x00\x00\x00\x01\x00\x01\x00\xc0]\x00\x00\x80\xbb\x00\x00\x02\x00\x10\x00data\xc8'\x01\x00
 ```

File: livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py (hunk 2 above)
Comment (MichaelSolati) [reply livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py]: Lol, by third party its just me... I made the plugin, I was talking to Inworld to maybe pass control of it to them cause I assumed they'd want it, and looking at how we pass our own http clients into some sdks I made sure this was an option for this, but I'll roll in the logic into this plugin.

File: livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py [hunk 4]
```diff
@@ -0,0 +1,321 @@
...
+            pitch=pitch,
+            speed=speed,
+            sampleRateHertz=sample_rate,
+            tokenizer=tokenize.basic.SentenceTokenizer(
+                min_sentence_len=10,
+                stream_context_len=5,
+            ),
+        )
+
+        self._http_client = httpx.AsyncClient(
```
davidzhao: can we use the shared session here? so that it would be able to re-use the connection pool

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
```diff
@@ -0,0 +1,321 @@
...
+                await self._process_audio_stream(stream, output_emitter)
+            output_emitter.end_segment()
+        except httpx.TimeoutException:
+            raise APITimeoutError() from None
+        except Exception as e:
+            raise APIConnectionError() from e
+
+    def _create_synthesis_stream(
+        self, sentence: tokenize.TokenData
+    ) -> AbstractAsyncContextManager[httpx.Response]:
```
davidzhao: this looks like a simple HTTP Post request, which makes it a non-streaming TTS.. and the right path to implement is a `ChunkedStream`. I know the naming might be confusing here.. but here's what the terminology means:
- streaming=True - it's able to keep the connection open and allow us to stream in input tokens as quickly as possible
- streaming=False - we cannot stream in input tokens, and must create a synthesis explicitly

what you are doing here is the latter.. but it's kind of awkwardly implemented as the former.

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
davidzhao: this is handled in the decoder ([code](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/tts/tts.py#L549))

the reason why it's not working for you is because [you are passing in](https://github.com/livekit/agents/pull/2634/files#diff-52850da42dd8b430d471f2f6bde941edad6eebc6f207e5ee3b25d751a9265c43R37) mime type of `audio/pcm`. that tells the emitter to *not* decode it, but to treat it as raw pcm.

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
MichaelSolati: I was under the impression it already was as the `self._http_client` is used by the `ChunkedStream` and we never call `close()` on the http client. Unless you're talking about using `aiohttp` ?

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
MichaelSolati: I think this is the only hold up, we synced offline so I'll wait to hear from you regarding this. However I think everything else should be good to go.

PR #2639, opened 2025-06-18
Fix Hume TTS

### LINKED ISSUE #2619: Incomplete implementation of hume plugin in version ~1.1.0
I wanted to upgrade my livekit agent to 1.1.0 because in version 1.0.23 we faced a problem with AssemblyAI format_turns which got fixed in 1.1.0 but in this version we're facing issues with hume plugin as well as it's custom voice implementation.

The only way to pass the voice is to pass utterance_options and the text is a mandatory field. Even when passing the text as an empty string, it throws some error regarding the stream.

I think the hume tts is still incomplete for version ~1.1.0

bcherry:
The primary fix here is just setting `streaming=False` but I also made some cleanups to the interface so it is more directly aligned to the documentation. those fixes also neatly close #2619

### DIFF
--- a/livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
+++ b/livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
@@ -20,7 +20,7 @@ import json
 import os
 from dataclasses import dataclass, replace
 from enum import Enum
-from typing import Any, TypedDict, Union, cast
+from typing import Any, TypedDict
 
 import aiohttp
 
@@ -34,26 +34,38 @@ DEFAULT_BASE_URL = "https://api.hume.ai"
 SUPPORTED_SAMPLE_RATE = 48000
 
 
-class PostedUtterance(TypedDict, total=False):
+class VoiceId(TypedDict, total=False):
+    id: str
+    provider: VoiceProvider | None
+
+
+class VoiceName(TypedDict, total=False):
+    name: str
+    provider: VoiceProvider | None
+
+
+class UtteranceOptions(TypedDict, total=False):
+    description: str | None
+    speed: float | None
+    voice: VoiceId | VoiceName | None
+    trailing_silence: float | None
+
+
+class Utterance(UtteranceOptions, total=False):
     text: str
-    description: str
-    voice: dict[str, Any]
-    speed: float
-    trailing_silence: float
 
 
-class PostedContextWithGenerationId(TypedDict, total=False):
+class ContextGenerationId(TypedDict, total=False):
     generation_id: str
 
 
-class PostedContextWithUtterances(TypedDict, total=False):
-    utterances: list[PostedUtterance]
+class ContextUtterances(TypedDict, total=False):
+    utterances: list[Utterance]
 
 
-PostedContext = Union[
-    PostedContextWithGenerationId,
-    PostedContextWithUtterances,
-]
+class VoiceProvider(str, Enum):
+    hume = "HUME_AI"
+    custom = "CUSTOM_VOICE"
 
 
 class AudioFormat(str, Enum):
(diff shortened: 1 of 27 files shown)


livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
```diff
@@ -80,29 +92,25 @@ def __init__(
...
+        utterance_options: NotGivenOr[UtteranceOptions] = NOT_GIVEN,
+        context: ContextGenerationId | ContextUtterances | None = None,
         instant_mode: bool = True,
         audio_format: AudioFormat = AudioFormat.mp3,
         base_url: str = DEFAULT_BASE_URL,
         http_session: aiohttp.ClientSession | None = None,
     ):
         super().__init__(
-            capabilities=tts.TTSCapabilities(streaming=True),
+            capabilities=tts.TTSCapabilities(streaming=False),
```
bcherry, line 131: an alternate version of this PR would just be this single line. maybe better because it wouldn't be an interface change for developers but i think the way the interface is done now is a lot cleaner

davidzhao (changes requested):
actually we have some type issues here: https://github.com/livekit/agents/actions/runs/15747758231/job/44387144189?pr=2639#step:7:90

davidzhao:
we'd also want to remove the streaming test cases, those are failing

bcherry:
> we'd also want to remove the streaming test cases, those are failing

yeah I see - I thought this didn't have tests but it did, just not for streaming (which it didn't support but said it did).

bcherry:
updated after feedback from hume and with some useful simplifications from doc writing

livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
```diff
@@ -80,31 +98,63 @@ def __init__(
...
         api_key: str | None = None,
-        utterance_options: NotGivenOr[PostedUtterance] = NOT_GIVEN,
-        context: PostedContext | None = None,
-        instant_mode: bool = True,
+        voice: VoiceById | VoiceByName | None = DEFAULT_VOICE,
+        description: str | None = None,
+        speed: float | None = None,
+        trailing_silence: float | None = None,
+        context: str | list[Utterance] | None = None,
+        instant_mode: NotGivenOr[bool] = NOT_GIVEN,
```
theomonnom, line 106: I think we should always enable instant_mode by default in our case

livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
```diff
@@ -20,56 +20,74 @@
...
 class AudioFormat(str, Enum):
+    """Audio format for the synthesized speech."""
+
     mp3 = "mp3"
     wav = "wav"
     pcm = "pcm"
 
 
+DEFAULT_HEADERS = {
+    "X-Hume-Client-Name": "LiveKit",
```
zgreathouse:
```suggestion
    "X-Hume-Client-Name": "livekit",
```

zgreathouse: @bcherry  We very much appreciate this being added! (our system just expects this to be lowercase)

livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
zgreathouse: @theomonnom @bcherry 

**Context:** Very soon `instant_mode` will be enabled by default in the API. (We plan to roll this change out in the next week.)

The only caveat is that when `instant_mode` is enabled the API requires a `voice` to be specified. 

Preferred behavior for when `instant_mode` is enabled and no `voice` is specified is to use the default `voice`. In other words: the default `voice` would be applied if no voice option is specified—unless `instant_mode` is set to `false`.

bcherry: the behavior in this PR is exactly that - on by default unless you explicitly pass None for the voice, then its turned off because otherwise the API would just throw error. but that's too complex to model in the params so it uses NOT_GIVEN in the param list and sorts out the specific behavior in the method body.

PR #2643, opened 2025-06-19
Issue with STT in Deepgram freezing after update_uptions with language in later versions of agents

vvv-001:
Ihave created this fix myself and it seems to work in 1.1.0, also when doing multiple language switches. Please feel free to look at it and review.

The issue that is caused is with newer versions of LiveKit Agents, after 1.0.20 (and even still in 1.1.0), after updating options for Deepgram STT, specifically Language, the STT model doesn't respond anymore. There is additional loggin added and can be removed eventually.

Full description in this issue here as well:
https://github.com/livekit-examples/python-agents-examples/issues/19

This is preventing us from upgrading from 1.0.19, since the issues was introduced in version 1.0.20 and still exists in current version.

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -20,9 +20,11 @@ import json
 import os
 import weakref
 from dataclasses import dataclass
+from enum import Enum
 from typing import Any
 
 import aiohttp
+import numpy as np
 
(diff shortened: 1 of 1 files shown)


davidzhao:
are you having issues with TTS or STT?

davidzhao:
we can reproduce this and create a fix

vvv-001:
@davidzhao this is with STT

longcw:
@vvv-001 this should be fixed in https://github.com/livekit/agents/pull/2654, will be included in next release.

PR #2646, opened 2025-06-19
Fix : Increase audio mixer timeout

CyprienRicqueB2L:
We observe frequent timeouts for the background audio.

I increased the default timeout to fix it.

Other options are 
- Increasing default timeout of `AudioMixer`
- Adding param `audio_mixer_kwargs` to `BackgroundAudioPlayer`
- Adding param `stream_timeout_ms` to `BackgroundAudioPlayer`

What would be the preferred approach?


theomonnom (approved):
Thanks that lg!

PR #2670, opened 2025-06-21
fix: silence local agent in bey example

niqodea:
This PR disables room audio output for the local agent to avoid conflicts with the avatar agent in the `bey` avatar example. 

Without this, the local agent speaks first, causing the avatar agent to remain silent. This ensures the avatar can respond as expected.

### DIFF
--- a/examples/avatar_agents/bey/agent_worker.py
+++ b/examples/avatar_agents/bey/agent_worker.py
@@ -3,7 +3,14 @@ import os
 
 from dotenv import load_dotenv
 
-from livekit.agents import Agent, AgentSession, JobContext, WorkerOptions, WorkerType, cli
+from livekit.agents import (
+    Agent,
+    AgentSession,
+    JobContext,
+    RoomOutputOptions,
+    WorkerOptions,
+    WorkerType,
+    cli,
 from livekit.plugins import bey, openai
 
 logger = logging.getLogger("bey-avatar-example")
@@ -26,6 +33,9 @@ async def entrypoint(ctx: JobContext):
     await session.start(
         agent=Agent(instructions="Talk to me!"),
         room=ctx.room,
+        # Disable room audio output to prevent the local agent from speaking.
+        # Otherwise, it will interrupt the avatar agent, which will then stay silent.
+        room_output_options=RoomOutputOptions(audio_enabled=False),
     )
 
     session.generate_reply(instructions="say hello to the user")


theomonnom:
Hey, this lgtm, can you fix the ruff error?

davidzhao:
@niqodea on the bey side, I think you should not be subscribing to audio tracks from the room. that way it would not be confused by other audio going around.

examples/avatar_agents/bey/agent_worker.py
```diff
@@ -26,6 +33,9 @@ async def entrypoint(ctx: JobContext):
     await session.start(
         agent=Agent(instructions="Talk to me!"),
         room=ctx.room,
+        # Disable room audio output to prevent the local agent from speaking.
+        # Otherwise, it will interrupt the avatar agent, which will then stay silent.
+        room_output_options=RoomOutputOptions(audio_enabled=False),
```
longcw, line 38: With the latest version of agents, we don't need to manually disable the audio output, when the session.output.audio is already set by the AvatarSession, we won't create another room audio output.

longcw: We removed these lines intentionally in https://github.com/livekit/agents/pull/2542/files#diff-2783da2a53a7ea8994c2ccc6a0c6ba9dcc40fadd3eeb3665b77dd4344821a9bf since the room audio output is disabled automatically when avatar plugin is used.

niqodea: I see, thanks for the clarification!

niqodea:
Closing as this is simply caused by a breaking change in `livekit-agents`

niqodea:
@davidzhao 

> @niqodea on the bey side, I think you should not be subscribing to audio tracks from the room. that way it would not be confused by other audio going around.

Are we sure about this? We would like to listen to the room audio so that we can interrupt our speech in case the human in the room is speaking, right?

longcw:
> We would like to listen to the room audio so that we can interrupt our speech in case the human in the room is speaking, right?

The interruption event is sent to the avatar through [rpc call](https://github.com/livekit/agents/pull/2670), you don't need to estimate that from audio. Basically, if you are using the AvatarRunner, the `clear_buffer` [method](https://github.com/livekit/agents/blob/main/examples/avatar_agents/audio_wave/avatar_runner.py#L75-L81) is called when interruption happends.

Issue #2674, opened 2025-06-22
Expose configuration for max_automatic_function_calls


macastro9714:
Hello,

When using the google llm plugin with automatic function calling, it appears the maximum number of sequential function calls is hardcoded. The logs show a default limit of 10 calls:

```
INFO:google_genai.models:AFC is enabled with max remote calls: 10.
```

This default limit is restrictive for agents whose primary purpose is routing or executing complex, multi-step function sequences that may require more than 10 calls.

Google docs for Python allows this value to be configured to adjust it based on the use case. (https://cloud.google.com/python/docs/reference/vertexai/latest#automatic-function-calling)

**Example from Google's documentation:**
```python
afc_responder = AutomaticFunctionCallingResponder(
    max_automatic_function_calls=5,
)

model = GenerativeModel("gemini-pro")
chat = model.start_chat(responder=afc_responder)
```

Could a similar configuration be exposed in the LiveKit implementation to allow developers to override the default limit for `max_automatic_function_calls`?

Thank you.

macastro9714:
For reference, in case it's useful here is how I worked around it:

```
class GoogleLLMWithAFC(GoogleLLM):
    """
    Google LLM with automatic AFC configuration.
    
    This class extends the standard Google LLM to automatically configure
    Automated Function Calling with the maximum remote calls setting from
    application configuration.
    """

    def chat( # type: ignore[override]
        self,
        *,
        chat_ctx: llm.ChatContext,
        tools: list[FunctionTool] | None = None,
        conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
        parallel_tool_calls: NotGivenOr[bool] = NOT_GIVEN,
        tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
        response_format: NotGivenOr[
            types.SchemaUnion | type[llm_utils.ResponseFormatT]
        ] = NOT_GIVEN,
        extra_kwargs: NotGivenOr[dict[str, Any]] = NOT_GIVEN,
    ) -> LLMStream:
        """
        Create a chat completion with automatic AFC configuration.
        
        This method extends the base chat method to automatically configure
        AFC when function tools are available.
        """

        extra = {}
        if is_given(extra_kwargs):
            extra.update(extra_kwargs)
        
        if tools and len(tools) > 0:
            try:
                afc_config = types.AutomaticFunctionCallingConfig(
                    maximum_remote_calls=20 
                )
[...]

PR #2675, opened 2025-06-23
expose automatic_function_calling config for google LLM

longcw:
fix https://github.com/livekit/agents/issues/2674


davidzhao (approved):
lkgtm

PR #2687, opened 2025-06-23
Add Fireworks AI STT plugin

yunyichi:
- Add fireworks AI Streaming STT plugin.
- Add plugin to CI

For STT parameters, only `language`, `prompt`, `temperature`, and `base_url` are needed to use this STT plugin. Other parameters are already included but not currently required; they’re just there in case future needs arise.

### DIFF
--- a/livekit-plugins/livekit-plugins-fireworksai/livekit/plugins/fireworksai/stt.py
+++ b/livekit-plugins/livekit-plugins-fireworksai/livekit/plugins/fireworksai/stt.py
@@ -85,12 +85,7 @@ class STTOptions:
     temperature: NotGivenOr[float] = NOT_GIVEN
     skip_vad: NotGivenOr[bool] = NOT_GIVEN
     vad_kwargs: NotGivenOr[dict] = NOT_GIVEN
-    audio_window_seconds: NotGivenOr[float] = NOT_GIVEN
-    max_audio_window_seconds: NotGivenOr[float] = NOT_GIVEN
-    min_duration_sec: NotGivenOr[float] = NOT_GIVEN
-    n_prefix_words: NotGivenOr[int] = NOT_GIVEN
-    n_speculation_words: NotGivenOr[int] = NOT_GIVEN
-    text_timeout_seconds: float = 2.0
+    text_timeout_seconds: float = 1.0
     response_format: str = "verbose_json"
     timestamp_granularities: NotGivenOr[list[str]] = NOT_GIVEN
     base_url: NotGivenOr[str] = NOT_GIVEN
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-fireworksai/livekit/plugins/fireworksai/stt.py
```diff
@@ -0,0 +1,551 @@
...
+            "response_format": self._opts.response_format,
+            "timestamp_granularities": (
+                self._opts.timestamp_granularities
+                if is_given(self._opts.timestamp_granularities)
+                else None
+            ),
+        }
+
+        headers = {
+            "User-Agent": "Python/3.12",
```
longcw:
```suggestion
            "User-Agent": "LiveKit Agents",
```

livekit-plugins/livekit-plugins-fireworksai/livekit/plugins/fireworksai/stt.py
```diff
@@ -0,0 +1,551 @@
...
+        temperature: NotGivenOr[float] = NOT_GIVEN,
+        skip_vad: NotGivenOr[bool] = NOT_GIVEN,
+        vad_kwargs: NotGivenOr[dict] = NOT_GIVEN,
+        audio_window_seconds: NotGivenOr[float] = NOT_GIVEN,
+        max_audio_window_seconds: NotGivenOr[float] = NOT_GIVEN,
+        min_duration_sec: NotGivenOr[float] = NOT_GIVEN,
+        n_prefix_words: NotGivenOr[int] = NOT_GIVEN,
+        n_speculation_words: NotGivenOr[int] = NOT_GIVEN,
+        text_timeout_seconds: float = 2.0,
+        timestamp_granularities: NotGivenOr[list[str]] = NOT_GIVEN,
```
longcw, line 107: can you add some comments to these options?

longcw (approved):
lgtm! I tested the plugin and it works, one thing I have noticed is that the delay for final user transcript is like 2s? is that controlled by `text_timeout_seconds`? personally I think the default value should be lower to reduce the latency.

livekit-plugins/livekit-plugins-fireworksai/livekit/plugins/fireworksai/stt.py
yunyichi: Thank you for your advice, already changed text_timeout_seconds to 1.0, and also changed head to "LiveKit Agents".

livekit-plugins/livekit-plugins-fireworksai/livekit/plugins/fireworksai/stt.py
yunyichi: Sure, I removed some unhelpful parameters that might confuse user and added comments to clarify the remaining ones. Thank you very much~

PR #2689, opened 2025-06-23
add type=object to functiondeclaration parameters to match gemini api expectation

ewsiegel:
Gemini API Issue: https://github.com/googleapis/python-genai/issues/981

Gemini API fails when using mode=FunctionCallingConfigMode.ANY, unless tool parameters have been annotated with type=object

Currently the plugin fails to interact with gemini api with error code 400 INVALID_ARGUMENT

### DIFF
--- a/.github/next-release/changeset-72dd288a.md
+++ b/.github/next-release/changeset-72dd288a.md
@@ -1,0 +1,5 @@
+---
+"livekit-plugins-google": patch
+---
+
+add type=object to functiondeclaration parameters to match gemini api expectation (#2689)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/utils.py
```diff
@@ -103,6 +103,12 @@ def get_tool_results_for_realtime(
 def _build_gemini_fnc(function_tool: FunctionTool) -> types.FunctionDeclaration:
     fnc = llm.utils.build_legacy_openai_schema(function_tool, internally_tagged=True)
     json_schema = _GeminiJsonSchema(fnc["parameters"]).simplify()
+    # Gemini API expects type object here:
+    # https://github.com/googleapis/python-genai/issues/981
+    if json_schema:
+        json_schema["type"] = "object"
```
longcw, line 109: dose this happen only when there is no parameter? if so maybe we do this only when `json_schema` is empty?

ewsiegel: I've observed it happen when there are parameters there too

PR #2695, opened 2025-06-24
add markdown filter for tts and transcription nodes

longcw:


longcw:
> Should we always do it by default?

we can enable it by default for tts node, maybe add an option in `ModelSettings`?

theomonnom:
> > Should we always do it by default?
> 
> we can enable it by default for tts node, maybe add an option in `ModelSettings`?

yeah here: https://github.com/livekit/agents/blob/1000cdbf560c193bf8aac80be6194ebc6829abab/livekit-agents/livekit/agents/voice/agent.py#L383

I don't think we can use ModelSettings for that

longcw:
> yeah here:
> 
> https://github.com/livekit/agents/blob/1000cdbf560c193bf8aac80be6194ebc6829abab/livekit-agents/livekit/agents/voice/agent.py#L383
> 
> I don't think we can use ModelSettings for that

how to make it configurable? I'll merge this one first, we can create another pr to enable it by default.

PR #2728, opened 2025-06-27
preemptive generation before end of user turn

longcw:
add `preemptive_generation` option to start llm generation before end of user turn, reduce the latency if
1. the STT final transcript is faster than VAD end of speech
2. turn detection model enabled (e.g. 80~130ms on a Mac mini)

This should wait for https://github.com/livekit/agents/pull/2482


theomonnom:
Can we do that on the v2 branch? The way we're now doing speech scheduling changed quite a bit.

longcw:
> Can we do that on the v2 branch? The way we're now doing speech scheduling changed quite a bit.

yeah, that what I am thinking, this one should wait for the v1.2 to avoid conflict

longcw:
I can merge the v1.2 branch

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -452,6 +452,24 @@ def from_dict(cls, data: dict[str, Any]) -> ChatContext:
...
+                item_a.type == "message"
+                and item_b.type == "message"
+                and item_a.text_content != item_b.text_content
+            )
+            for item_a, item_b in zip(self.items, other.items)
+        ):
+            return False
+
+        return True
+
```
theomonnom: I don't think we should override the equality of our chat_ctx because it isn't a dataclass. Even if the content is the same, the chat_ctx inside the AgentSession and the chat_ctx inside an Agent is different IMO. different references. I know we can use the `is` keyword but IMO the content is too volatile inside the chat_ctx. 

Let's create a separate method instead, wdyt? maybe is_equivalent?

longcw: `is_equivalent` sounds good, but why you think
> if the content is the same, the chat_ctx inside the AgentSession and the chat_ctx inside an Agent is different

theomonnom: what I mean is that `session.chat_ctx == agent.chat_ctx` must always returns False even if the content is the same.

e.g adding a new item to one of them will not add it to the other. The references are different

longcw: tho I disagree with that.. IMO `==` is for content equivalent, `is` is for the same reference, it's okay to use a `is_equivalent` method instead of `__eq__` (better for avoiding confusion like this for users).

theomonnom: Right, it's just that by default, Python uses reference and not content if it isn't a dataclass. IMO it feels weird for the ChatContext to be content checked by default when using ==

Also == is generally for the exact content match, the above is_equivalent ignores some fields

PR #2741, opened 2025-06-29
Minimal mistralai integration

fabitokki:
New plugin with integration of MistralChatModel 
No tool/function calling yet


fabitokki:
Anyone knows how to pass this test: 

Run tar -czf lk-dump-all.tar.gz -C all_lk_dump .
tar: all_lk_dump: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
Error: Process completed with exit code 2.

I could not find any info on this ?

Issue #2729, opened 2025-06-27
Unclosed Google LLM `aiohttp.Client` Causes Continuous Logger Errors


anlagbr:
# Unclosed Google LLM `aiohttp.Client` Causes Continuous Logger Errors

This document addresses the issue of continuous `logger.error` messages caused by an unclosed `aiohttp.Client` session when using the Google LLM.

## Problem Description

When initializing the Google LLM with the following code:

```python
llm = google.LLM(
    model="gemini-2.5-flash",
    thinking_config=types.ThinkingConfig(thinking_budget=0),
)
```

The application continuously logs `ERROR` messages related to "Unclosed client session" and "Unclosed connector" from the `asyncio` library.

## Error Messages

The following `ERROR` messages are repeatedly observed in the logs:

```
2025-06-27 10:38:59,485 - ERROR asyncio - Unclosed client session---------]
client_session: <aiohttp.client.ClientSession object at 0x0000020CEA6F82C0>
2025-06-27 10:38:59,485 - ERROR asyncio - Unclosed connector
connections: ['deque([(<aiohttp.client_proto.ResponseHandler object at 0x0000020CE727F2F0>, 1800540.953)])']
connector: <aiohttp.connector.TCPConnector object at 0x0000020CE8A52A50>
```

## Analysis of the Issue

The repeated "Unclosed client session" and "Unclosed connector" errors indicate that `aiohttp` client sessions and their underlying connectors are *not being properly closed* after use. This can lead to:

*   **Resource Leaks:** Open connections consume system resources (e.g., file descriptors, memory), which can eventually exhaust available resources and cause application instability or crashes.
*   **Performance Degradation:** Accumulation of unclosed connections can slow down the application as it struggles to manage these resources.

anlagbr:
# Package Versions

I am using `livekit` related packages:

*   **LiveKit Core:** `livekit>=1.0.11`
*   **LiveKit Agents:** `livekit-agents[google,silero,turn-detector,openai,cartesia,deepgram,groq,assemblyai,speechmatics]>=1.1.2`
*   **LiveKit Plugins - Noise Cancellation:** `livekit-plugins-noise-cancellation>=0.2.4`

longcw:
This is an issue in google genai v1.21, it should be fixed with https://github.com/livekit/agents/pull/2743

macastro9714:
fyi @longcw I'm still getting this errors using gemini with livekit v1.2.15. I'm getting:

[...]

PR #2743, opened 2025-06-30
upgrade google genai to 1.23

longcw:
https://github.com/googleapis/python-genai/releases/tag/v1.23.0

the async error was fixed

fix https://github.com/livekit/agents/issues/2729

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/utils.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/utils.py
@@ -32,9 +32,7 @@ def to_fnc_ctx(fncs: list[FunctionTool | RawFunctionTool]) -> list[types.Functio
                 types.FunctionDeclaration(
                     name=info.name,
                     description=info.raw_schema.get("description", ""),
-                    parameters=info.raw_schema.get("parameters", {}),
-                    # will change in genai 1.20.1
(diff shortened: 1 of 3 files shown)


theomonnom (approved):

cyruszad:
I'm still experiencing the "Unclosed client session" error in google-genai version 1.23.0 and 1.24.0 when using with LiveKit agents.

  Environment:
  - google-genai: 1.23.0 (also tested with 1.24.0)
  - livekit-plugins-google: 1.1.5
  - Python 3.12
  - macOS

  The issue occurs when:
  1. Using thinking_config with any thinking_budget value
  2. The model makes function/tool calls (even with automatic_function_calling_config={"disable": True})

  Error logs:
  2025-07-05 06:11:57,356 - INFO google_genai.models - AFC is enabled with max remote calls: 10
  2025-07-05 06:11:57,359 - ERROR asyncio - Unclosed client session
  client_session: <aiohttp.client.ClientSession object at 0x11e59dfd0>

  Observations:
  - Error only appears when thinking_config is set
  - Disabling AFC reduces but doesn't eliminate the errors
  - Sessions are created internally by google-genai but not properly closed
  - The error appears immediately after "AFC is enabled" log message

This seems to be the same issue - the internal aiohttp ClientSession created for AFC/function calling isn't being properly closed. The issue persists even with the aiohttp backend installed via pip install google-genai[aiohttp].

PR #2766, opened 2025-07-01
Install optional dependencies during docs gen

bcherry:
the docs workflow doesn't work currently because it wasn't installing the optional aws realtime dependency

https://github.com/livekit/agents/actions/runs/15986741157/job/45092524003#step:8:594

This PR ensures the docs are generated on the widest scope of optional dependencies, which makes intuitive sense to me. I also cleaned up the the docs action to make more correct and complete use of `uv`.

original scope:

~But actually i think we should prefer not making this an optional dependency, so it comes along with `pip install livekit-agents[aws]`. The downside is it requires python 3.12+...~

~There may be an alternative fix available, I'm not sure. Getting this branch up will unblock docs publishing at least since I'll publish off the branch.~


davidzhao (commented):
we need to update the [README](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-aws/README.md) as well to reflect this

bcherry:
@davidzhao after further thought I've decided to leave the plugin as it was and make the docs work across optional plugin dependencies instead. PR is updated.

theomonnom (approved):

Issue #2764, opened 2025-06-30
Interrupt user example throws error and conversation turn order disrupted


richard-sobey:
Using 1.1.5:

When running the interrupt user [example](https://github.com/livekit-examples/python-agents-examples/blob/main/pipeline-llm/interrupt_user.py) from the docs it throws the below error:

`TypeError: '<' not supported between instances of 'SpeechHandle' and 'SpeechHandle'`

The trace comes from line 693 in livekit/agents/voice/agent_activity.py:

`heapq.heappush(self._speech_q, (-priority, time.monotonic_ns(), speech))`

The agent interrupts correctly on cue but once it's message is complete, the agent generates an immediate second follow-up message in response to the previous user's message which was originally interrupted: this is usually completely unrelated to the interrupt message and makes no sense in context of conversation.

Tried a few variations of manually adjusting agent state but with no luck. Help for a quick fix appreciated

longcw:
> the agent generates an immediate second follow-up message in response to the previous user's message which was originally interrupted

most likely this is because the STT sent the final transcript after the `handle_interruption` so it triggered the reply again. You can add a `session.clear_user_turn()` in `handle_interruption` to clear the STT after interruption.

longcw:
> TypeError: '<' not supported between instances of 'SpeechHandle' and 'SpeechHandle'

Can you share the steps to reproduce this error?

richard-sobey:
@longcw sure see below - it's pretty much just running the livekit example with latest 1.1.5 version

Packages
``` python
asyncio                            3.4.3
livekit                            1.0.11
livekit-agents                     1.1.5
livekit-api                        1.0.3
livekit-plugins-deepgram           1.1.5
livekit-plugins-noise-cancellation 0.2.4
livekit-plugins-openai             1.1.5
livekit-plugins-silero             1.1.5
livekit-plugins-turn-detector      1.1.5
livekit-protocol                   1.0.4
```

Code taken exactly from the livekit examples [here](https://github.com/livekit-examples/python-agents-examples/blob/main/pipeline-llm/interrupt_user.py):
``` python
import logging
from pathlib import Path
from dotenv import load_dotenv
from livekit.agents import JobContext, WorkerOptions, cli
from livekit.agents.voice import Agent, AgentSession
from livekit.plugins import openai, deepgram, silero
from livekit.agents.llm import ChatContext, ChatMessage
import re
import asyncio

load_dotenv(dotenv_path='.env.local')

[...]

PR #2770, opened 2025-07-01
fix schedule speech on windows when monotonic_ns resolution is rough

longcw:
Before python 3.13, on Windows, time.monotonic() uses gGetTickCount64() clock which has a resolution of 15.6 ms.
https://github.com/python/cpython/pull/116781

Replaced it with `time.perf_counter_ns` for python < 3.13 on Windows, fix https://github.com/livekit/agents/issues/2764

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -695,6 +695,8 @@ class AgentActivity(RecognitionHooks):
                 heapq.heappush(self._speech_q, (-priority, time.perf_counter_ns(), speech))
                 break
             except TypeError:
+                # handle TypeError when identical timestamps cause speech comparison failure
+                # with perf_counter_ns(), collisions should be rare
                 pass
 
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -689,8 +689,14 @@ def _schedule_speech(
...
             raise RuntimeError("cannot schedule new speech, the agent is draining")
 
-        # Negate the priority to make it a max heap
-        heapq.heappush(self._speech_q, (-priority, time.monotonic_ns(), speech))
+        while True:
+            try:
+                # negate the priority to make it a max heap
+                heapq.heappush(self._speech_q, (-priority, time.perf_counter_ns(), speech))
+                break
+            except TypeError:
```
davidzhao, line 697: maybe comment here on why TypeError is thrown? (it's clear when reading the bug, but that context is lost when reading the code alone)

davidzhao (approved):
nice fix!

PR #2772, opened 2025-07-01
Feat/mistralai plugins

fabitokki:
Minimal Mistral AI plugins


livekit-plugins/livekit-plugins-mistralai/livekit/plugins/mistralai/llm.py
```diff
@@ -0,0 +1,192 @@
...
+    ChatCompletionStreamRequestMessages,
+    Mistral,
+    SystemMessage,
+    UserMessage,
+)
+
+from .models import ChatModels
+
+
+def to_async_stream_mistral_format(
```
theomonnom: We should move this utility inside our ChatContext.to_provider_format. https://github.com/livekit/agents/blob/cddf578e4460a7f3ed893086dd9068472769acca/livekit-agents/livekit/agents/llm/chat_context.py#L362

Perhaps Mistral AI supports the OpenAI API format? If not, we could consider moving the code there.

fabitokki: Hey theo! Thanks for the check
I will integrate the to_provider_format.

Should I also rebase main and solve conflicts ?
Thanks for your availability

fabitokki: I moved it to llm/chat_context.py to use the ChatContext.to_provider_format. However it seems to ue a different format that OpenAI API.

I also used standard names LLM and LLMStream for my class definition instead of MistralLLM to keep it standardize.

class LLM(llm.LLM): instead of class MistraLLM(llm.LLM), etc etc 


Should be cleaner now

fabitokki:
@theomonnom I pulled the new main and rebase/clean my history in my branch. Some CI checks for livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental are not passing but apparently this is another issue. I used the ChatContext.to_provider_format. as you asked so should be fine now :)

theomonnom (approved):
lgtm!!

PR #2778, opened 2025-07-02
fix types for agents 1.2

longcw:


theomonnom (approved):
Oh thanks so much man!

PR #2780, opened 2025-07-02
feat: ability to use remote EOT inference when deployed in Cloud

davidzhao:
when `LIVEKIT_EOT_MULTI_URL` is set, we'll use it instead of running turn detector locally


livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/multilingual.py
```diff
@@ -14,10 +22,63 @@ def __init__(self) -> None:
...
+                    extra=data,
+                )
+                return probability
+            else:
+                # default to indicate no prediction
+                return 1
+
+
+def remote_inference_url() -> str | None:
+    return os.getenv("LIVEKIT_EOT_MULTI_URL")
```
theomonnom: IMO: 
```suggestion
    return os.getenv("LIVEKIT_REMOTE_EOT_URL")
```

And we append the eot/multi to it at runtime

theomonnom (approved):

PR #2785, opened 2025-07-02
fix blingfire tokenizer test

longcw:


livekit-agents/livekit/agents/tokenize/blingfire.py
```diff
@@ -22,7 +23,7 @@ def _split_sentences(
 
     for _, end in offsets:
         raw_sentence = text[start:end]
-        sentence = raw_sentence.strip()
+        sentence = re.sub(r"\s+", " ", raw_sentence).strip()
```
theomonnom: are we sure we want to reformat? 
I'm pretty sure the tests were passing before the retain_format changes

theomonnom: I think the issue here is that it is now using the text[start:end]
but text_to_sentences_with_offsets also directly return a list of sentences we can use when we don't need retain_format

longcw: it's just to replace the multiple `\n` to a single space if `retain_format=False`

theomonnom: I see, but can we directly use the sentences returned by blingfire instead? since they seem to already do those kind of normalization

theomonnom: or this is a different issue?

tests/test_tokenizer.py
```diff
@@ -65,8 +65,8 @@
     "This is a test. Another test.",
     "A short sentence. A longer sentence that is longer than the previous sentence. f(x) = x * 2.54 + 42.",
     "Hey! Hi! Hello! This is a sentence.",
-    "这是一个中文句子。 これは日本語の文章です。",
-    "你好！ LiveKit是一个直播音频和视频应用程序和服务的平台。",
+    "这是一个中文句子。これは日本語の文章です。",
+    "你好！LiveKit是一个直播音频和视频应用程序和服务的平台。",
```
longcw, line 72: before the test passed but actually there are redundant space added to the expected result. Those ` ` are not in the original text because these languages don't have a space between sentences.

livekit-agents/livekit/agents/tokenize/blingfire.py
longcw: it's just to keep the code simple, but we can have to split sentences for with or without format, so we don't add any operation on the output of the bingfire.

theomonnom (approved):

livekit-agents/livekit/agents/tokenize/blingfire.py
longcw: `re.sub(r"\s*\n+\s*", " ", raw_sentence).strip()` maybe this is safer?

theomonnom: I'm not sure, I don't know what's blingfire is doing internally. So it's why I suggested to use the returned strings instead of the offsets. But that's probably fine?

PR #2795, opened 2025-07-02
disconnect from room if session closed due to participant disconnected

longcw:


theomonnom:
I'm not sure if we should introduce this behaviour. It means that the whole Job is now tied to the AgentSession lifecycle. 
We can't know if some other processes are happening alongside the AgentSession

longcw:
> I'm not sure if we should introduce this behaviour. It means that the whole Job is now tied to the AgentSession lifecycle. We can't know if some other processes are happening alongside the AgentSession

This behavior was already in the 1.1.5 https://github.com/livekit/agents/blob/livekit-agents%401.1.5/livekit-agents/livekit/agents/voice/room_io/room_io.py#L443, it was introduced as discussed in [slack link](https://live-kit.slack.com/archives/C05P6DRC6MV/p1750905378907939?thread_ts=1750875213.711229&cid=C05P6DRC6MV), that the agent should leave the room if it was closed to avoid confusion (agent is in the room but doesn't response).

It disconnects from the room only the reason is participant disconnected.

theomonnom:
Oh ok I see, but what is this PR doing then? Isn't it enough inside the RoomIO?

`close_on_disconnect` needs to be respected

theomonnom:
ah ok this is livekit-agents 1.2, I think I accidentally removed this [line](https://github.com/livekit/agents/blob/975cf0d885b3fa024d3c025c43dcd4c3df733929/livekit-agents/livekit/agents/voice/room_io/room_io.py#L443) from RoomIO

longcw:
> Oh ok I see, but what is this PR doing then? Isn't it enough inside the RoomIO?
> 
> `close_on_disconnect` needs to be respected

it was removed when changing to `self._agent_session._close_soon(reason=CloseReason.PARTICIPANT_DISCONNECTED)` in agents 1.2.

`close_on_disconnect` is respected in room io before calling `session._close_soon`.

longcw:
maybe we make `close_soon` return a `Future` and still call the disconnect in room io as a done callback?

longcw:
> ah ok this is livekit-agents 1.2, I think I accidentally removed this [line](https://github.com/livekit/agents/blob/975cf0d885b3fa024d3c025c43dcd4c3df733929/livekit-agents/livekit/agents/voice/room_io/room_io.py#L443) from RoomIO

yeah, it's just we cannot call `room.disconnect` after `session.close_soon` now, we don't know when the session will be closed.

theomonnom:
> > ah ok this is livekit-agents 1.2, I think I accidentally removed this [line](https://github.com/livekit/agents/blob/975cf0d885b3fa024d3c025c43dcd4c3df733929/livekit-agents/livekit/agents/voice/room_io/room_io.py#L443) from RoomIO
> 
> yeah, it's just we cannot call `room.disconnect` after `session.close_soon` now, we don't know when the session will be closed.

Could RoomIO listen to the close event?

longcw:
> Could RoomIO listen to the close event?

`room.disconnect` needs to be called after close event? or it's fine as the disconnect needs to be in an async task

longcw:
> Could RoomIO listen to the close event?

oh we cannot do that, room io was closed before `close` event emit from the agent session.

livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -429,7 +428,17 @@ def _on_participant_disconnected(self, participant: rtc.RemoteParticipant) -> No
...
+                reason=CloseReason.PARTICIPANT_DISCONNECTED
+            )
+
+            def on_closed(task: asyncio.Task[None]) -> None:
+                if not task.cancelled() and (
+                    not self._disconnect_room_atask or self._disconnect_room_atask.done()
+                ):
+                    self._disconnect_room_atask = asyncio.create_task(self._room.disconnect())
+
+            closing_task.add_done_callback(on_closed)
```
longcw: @theomonnom  how about this version?

but I found a bug that disconnecting from the room when agent just started speaking, the job process didn't shutdown, and then it never shutdown after the timeout. This happens in main branch as well. any thoughts about this?

longcw:
some ppl are confused that the room is closed right after the agent disconnected, close this one for now.

PR #2796, opened 2025-07-02
Feature : Add audio_mixer_kwargs to BackgroundAudioPlayer

CyprienRicqueB2L:
We use custom musics in `BackgroundAudioPlayer` and face frequent timeouts in `rtc.AudioMixer`.

This PR allows to pass `audio_mixer_kwargs` to `BackgroundAudioPlayer` so we can set custom `stream_timeout_ms` considering the music duration.


theomonnom:
Maybe it would make more sense to only expose stream_timeout_ms?
Since the other parameters don't really need to be configured

CyprienRicqueB2L:
Yes both work, it seemed weird to expose only partially the configuration of the audio.
I change that

sarthakgoyal23:
@CyprienRicqueB2L We are facing similar timeout issue what value you configured to avoid the issue?

Issue #2387, opened 2025-05-23
Minor issue with GOOGLE_APPLICATION_CREDENTIALS and project_id


ChenghaoMou:
The comment in the code [mentions](https://github.com/livekit/agents/blob/9d8e9a3e2de777b73a2ca190f07abf17cce295d5/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py#L145) project is inferred from key file, but it still [checks](https://github.com/livekit/agents/blob/9d8e9a3e2de777b73a2ca190f07abf17cce295d5/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py#L211) for the project regardless of the GOOGLE_APPLICATION_CREDENTIALS status. There are many ways to deal with this:

1. change the comment;
2. check for GOOGLE_APPLICATION_CREDENTIALS and project_id first;
3. skip checking project_id if GOOGLE_APPLICATION_CREDENTIALS is present;

What do you think?

PR #2798, opened 2025-07-02
Improve gcp vertex credential check

ChenghaoMou:
This unifies two realtime/llm implementations of checking vertex ai project and location:

- if project is not provided, it is read from the GOOGLE_APPLICATION_CREDENTIALS, which could still be None
- if either project or location is still None, raise an error

This should address #2387


davidzhao:
@ChenghaoMou do you mind rebasing this one? we'd be glad to merge it

PR #2805, opened 2025-07-03
parameterize inference worker init timeout

### LINKED ISSUE #2496: cannot load both english & multilingual EOU
If the code import both EOU, it will fail to start because the code has a hardcoded timeout of 30s

Here is the code in question in worker.py

```
        self._inference_executor: ipc.inference_proc_executor.InferenceProcExecutor | None = None
        if len(_InferenceRunner.registered_runners) > 0:
            self._inference_executor = ipc.inference_proc_executor.InferenceProcExecutor(
                runners=_InferenceRunner.registered_runners,
                initialize_timeout=30,
                close_timeout=5,
                memory_warn_mb=2000,
```

Not sure why it ignores the initialize_process_timeout in WorkerOptions

levity:
Fixes #2496

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -296,7 +296,7 @@ class Worker(utils.EventEmitter[EventTypes]):
         if len(_InferenceRunner.registered_runners) > 0:
             self._inference_executor = ipc.inference_proc_executor.InferenceProcExecutor(
                 runners=_InferenceRunner.registered_runners,
-                initialize_timeout=30,
+                initialize_timeout=opts.initialize_process_timeout,
                 close_timeout=5,
                 memory_warn_mb=2000,
                 memory_limit_mb=0,  # no limit


theomonnom:
That makes sense, maybe we should add a new parameter instead? like `initialize_inference_process_timeout`

levity:
Just having one parameter works fine for my use case, at least. And it
makes sense to me in general to have a single timeout since both kinds of
worker start up at the same time.

yuyuma:
Eager to have see this merged and incorporated into next release because our deployments on Fly.io started failing recently due to associated timeouts

levity:
@yuyuma you may be interested in this monkeypatching step that I added to my Dockerfile as a temporary workaround:

```
# Monkeypatch livekit to remove hardcoded init timeout
RUN sed -i 's/initialize_timeout=30,/initialize_timeout=opts.initialize_process_timeout,/' \
    .venv/lib/python3.11/site-packages/livekit/agents/worker.py
```

change the path to match your own Python configuration ofc. hope this helps! ✌️

longcw:
@levity could you add a new `initialize_inference_process_timeout` param to the `WorkerOptions`?

levity:
> @levity could you add a new `initialize_inference_process_timeout` param to the `WorkerOptions`?

personally, i would advocate against adding another option. IMO the additional flexibility doesn't seem worth the mental overhead cost of adding another option.

speaking from my own experience: AFAIK the documentation never explicitly explains that there are multiple separate processes. so it's not really part of the developer's mental model that you would want or need to manage them separately. it's just, "run this command, and the whole thing either succeeds or fails as a unit."

isaaccs:
any update on it ?

theomonnom (approved):

PR #2817, opened 2025-07-03
Nova Sonic Example Agent

BumaldaOverTheWater94:


theomonnom (approved):
thanks! will cleanup some stuff after this PR

PR #2821, opened 2025-07-04
add test for restaurant example

longcw:


examples/restaurant/eval/test_agent.py
```diff
@@ -0,0 +1,168 @@
...
+@pytest.mark.asyncio
+async def test_reservation() -> None:
+    agents = _create_agents()
+    userdata = UserData(agents=agents)
+
+    async with _llm_model() as llm, AgentSession(llm=llm, userdata=userdata) as sess:
+        await sess.start(agents["greeter"])
+        if sess.current_speech:
+            # `generate_reply` is called in `on_enter`
+            await sess.current_speech
```
theomonnom: Do you know if this is really necessary?

examples/restaurant/eval/test_agent.py
```diff
@@ -0,0 +1,168 @@
...
+        if sess.current_agent != agents["checkout"]:
+            await sess._update_activity(agents["checkout"], new_activity="start")
+        await sess.current_agent.update_chat_ctx(checkout_chat_ctx)
+        result = await sess.run(user_input="oh I want to add another coffee")
+        try:
+            result.expect[:].contains_function_call(name="to_takeaway")
+        except AssertionError:
+            result.expect.message(role="assistant").judge(
+                llm, intent="ask user if they want to back to takeaway", verbose=True
+            )
```
theomonnom, line 169: I think we need to think about a better way to do this. I think it will not be possible for v1 (at least cleanly)

longcw: yeah, it's tricky. basically we cannot restore all states to a checkpoint right now

examples/restaurant/eval/test_agent.py
longcw: Yes if `sess.run()` with a `user_input`. The created timestamp of the user input may be earlier than agent reply in `on_enter`, causing the order issue in chat context.

theomonnom: Oh interesting, can you add a comment?

longcw: yeah sure, I forgot the reason without the comment lol

PR #2825, opened 2025-07-05
Non-blocking Long Running Function Tools

toubatbrian:
WIP non-blocking tool running in background.

Try to play around with [example](https://github.com/livekit/agents/blob/brian/non-blocking-tool-call/examples/voice_agents/long_running_tools.py).

- [x] Make long running tool work with voice pipeline agent
- [ ] Make  long running tool work with realtime model
    - [x] Allow realtime model to schedule tool, and been notified with `when_idle` reply mode
    - [ ] Allow realtime model to be interrupted with `interrupt` mode
- [ ] Integrate long running tool to gemini's native async function api

### DIFF
--- a/livekit-agents/livekit/agents/llm/tool_context.py
+++ b/livekit-agents/livekit/agents/llm/tool_context.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 import inspect
-from collections.abc import Awaitable
+from collections.abc import AsyncIterator, Awaitable
 from dataclasses import dataclass
 from typing import (
     Any,
@@ -31,8 +31,6 @@ from typing import (
 
 from typing_extensions import NotRequired, Required, TypedDict, TypeGuard
 
-from livekit.agents.llm.chat_context import FunctionCall
-
 
 # Used by ToolChoice
 class Function(TypedDict, total=False):
@@ -46,18 +44,9 @@ class NamedToolChoice(TypedDict, total=False):
 
 ToolChoice = Union[NamedToolChoice, Literal["auto", "required", "none"]]
 
-AsyncToolReplyMode = Literal["when_idle", "interrupt", "silent"]
+ToolReplyMode = Literal["when_idle", "interrupt", "silent"]
 
-AsyncToolPendingResult = Union[Callable[[FunctionCall], Any], dict[str, Any], str, None]
-
-DEFAULT_ASYNC_TOOL_REPLY_MODE: AsyncToolReplyMode = "when_idle"
-
-
-def default_async_tool_pending_result(call: FunctionCall) -> dict[str, Any]:
-    return {
-        "id": call.call_id,
-        "result": f"Tool {call.name} is executing, please wait...",
-    }
+DEFAULT_TOOL_REPLY_MODE: ToolReplyMode = "when_idle"
 
 
 class ToolError(Exception):
@@ -94,6 +83,7 @@ class StopResponse(Exception):
 class _FunctionToolInfo:
     name: str
     description: str | None
+    reply_mode: ToolReplyMode | None
 
 
 @runtime_checkable
@@ -123,6 +113,7 @@ class RawFunctionDescription(TypedDict):
 class _RawFunctionToolInfo:
     name: str
     raw_schema: dict[str, Any]
+    reply_mode: ToolReplyMode | None
 
 
 @runtime_checkable
@@ -134,38 +125,55 @@ class RawFunctionTool(Protocol):
 
 F = TypeVar("F", bound=Callable[..., Awaitable[Any]])
 Raw_F = TypeVar("Raw_F", bound=Callable[..., Awaitable[Any]])
+Async_F = TypeVar("Async_F", bound=Callable[..., AsyncIterator[Any]])
+Async_Raw_F = TypeVar("Async_Raw_F", bound=Callable[..., AsyncIterator[Any]])
 
 
 @overload
 def function_tool(
-    f: Raw_F, *, raw_schema: RawFunctionDescription | dict[str, Any]
+    f: Raw_F,
+    *,
+    raw_schema: RawFunctionDescription | dict[str, Any],
+    reply_mode: ToolReplyMode | None = None,
 ) -> RawFunctionTool: ...
 
 
 @overload
 def function_tool(
-    f: None = None, *, raw_schema: RawFunctionDescription | dict[str, Any]
+    f: None = None,
+    *,
+    raw_schema: RawFunctionDescription | dict[str, Any],
+    reply_mode: ToolReplyMode | None = None,
 ) -> Callable[[Raw_F], RawFunctionTool]: ...
 
 
 @overload
 def function_tool(
-    f: F, *, name: str | None = None, description: str | None = None
+    f: F,
+    *,
+    name: str | None = None,
+    description: str | None = None,
+    reply_mode: ToolReplyMode | None = None,
 ) -> FunctionTool: ...
 
 
 @overload
 def function_tool(
-    f: None = None, *, name: str | None = None, description: str | None = None
+    f: None = None,
+    *,
+    name: str | None = None,
+    description: str | None = None,
+    reply_mode: ToolReplyMode | None = None,
 ) -> Callable[[F], FunctionTool]: ...
 
 
 def function_tool(
-    f: F | Raw_F | None = None,
+    f: F | Raw_F | Async_F | Async_Raw_F | None = None,
     *,
     name: str | None = None,
     description: str | None = None,
     raw_schema: RawFunctionDescription | dict[str, Any] | None = None,
+    reply_mode: ToolReplyMode | None = None,
 ) -> (
     FunctionTool
     | RawFunctionTool
@@ -182,7 +190,11 @@ def function_tool(
             # support empty parameters
             raise ValueError("raw function description must contain a parameters key")
 
-        info = _RawFunctionToolInfo(raw_schema={**raw_schema}, name=raw_schema["name"])
+        info = _RawFunctionToolInfo(
+            raw_schema={**raw_schema},
+            name=raw_schema["name"],
+            reply_mode=reply_mode,
+        )
         setattr(func, "__livekit_raw_tool_info", info)
         return cast(RawFunctionTool, func)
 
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1427,6 +1440,98 @@ def _on_first_frame(_: asyncio.Future[None]) -> None:
...
+        # TODO(brian): should we allow user to run agent handoff in long running tools?
+        if sanitized_out.agent_task is not None:
+            logger.error(
+                "Unable to run AgentTask from the async tool execution",
+                extra={
+                    "function": sanitized_out.fnc_call.name,
+                    "speech_id": speech_handle.id,
+                },
+            )
+            return
```
toubatbrian, line 1489: Doing handoff in long running tools seems a bit weird, but technically achievable.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1427,6 +1440,98 @@ def _on_first_frame(_: asyncio.Future[None]) -> None:
...
+        )
+
+        chat_ctx = self._agent._chat_ctx.copy()
+        # TODO(brian): allow user to customize this message?
+        chat_ctx.add_message(
+            role="user",
+            content=[
+                f"Long running tool {scheduled_task.fnc_call.name} finished execution with output:\n{content}"  # noqa: E501
+            ],
+        )
```
toubatbrian, line 1520: Allow user to customize?

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -374,11 +426,24 @@ async def _execute_tools_task(
...
-                task = asyncio.create_task(
-                    function_tool(*fnc_args, **fnc_kwargs),
-                    name=f"function_tool_{fnc_call.name}",
-                )
+                if is_async_tool(function_tool):
+                    async_tool_stream = function_tool(*fnc_args, **fnc_kwargs)
+                    task = asyncio.create_task(
+                        async_tool_stream.__anext__(),
+                        name=f"function_tool_{fnc_call.name}",
+                    )
```
toubatbrian, line 441: When `function_tool` decorates an `AsyncGenerator`, it automatically behaves as a long-running tool. The first `yield` should immediately provide the tool response—this signals that the tool is running or has been scheduled. 

The second `yield` then marks the completion of the long-running tool. This pattern mirrors the behavior of an async context manager, using two yields to denote the start and end of execution.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1427,6 +1440,98 @@ def _on_first_frame(_: asyncio.Future[None]) -> None:
...
+            return
+
+        if reply_mode == "when_idle":
+            self._generate_reply(chat_ctx=chat_ctx, priority=SpeechHandle.SPEECH_PRIORITY_LOW)
+            return
+
+        if reply_mode == "interrupt":
+            self.interrupt()
+            self._generate_reply(chat_ctx=chat_ctx, priority=SpeechHandle.SPEECH_PRIORITY_HIGH)
+            return
```
toubatbrian, line 1533: Not sure if this also work with realtime model?

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -119,43 +125,100 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any: ...
 
 F = TypeVar("F", bound=Callable[..., Awaitable[Any]])
 Raw_F = TypeVar("Raw_F", bound=Callable[..., Awaitable[Any]])
+Async_F = TypeVar("Async_F", bound=Callable[..., AsyncGenerator[Any]])
+Async_Raw_F = TypeVar("Async_Raw_F", bound=Callable[..., AsyncGenerator[Any]])
```
toubatbrian, line 129: When function_tool decorates an AsyncGenerator, it automatically behaves as a long-running tool.

longcw: the current implementation is already non-blocking function tool, a executed tool won't be awaited before the speech marked as done, when the tool returns it will create a new reply to the speech queue. So it's basically `non-blocking` with `reply_node=when_idle`, any reason to add an explicit Async Function?

I was thinking what we missed is the blocking mode, the tools can be cancelled only when user interrupts the agents before audio playout is done. If the audio is done, new user audio won't cancel the tools running in background. In blocking mode, we may want to cancel the tools as well.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1427,6 +1440,98 @@ def _on_first_frame(_: asyncio.Future[None]) -> None:
...
+        # TODO(brian): allow user to customize this message?
+        chat_ctx.add_message(
+            role="user",
+            content=[
+                f"Long running tool {scheduled_task.fnc_call.name} finished execution with output:\n{content}"  # noqa: E501
+            ],
+        )
+
+        if reply_mode == "silent":
+            await self.update_chat_ctx(chat_ctx=chat_ctx)
```
longcw, line 1523: we don't need a `silent` mode? user can just return `None` from the tool to skip the reply, and they can call `update_chat_ctx` inside the tool if they want to change anything.

livekit-agents/livekit/agents/llm/tool_context.py
longcw: I originally think what we are missing is the blocking mode, the tools can be cancelled only when user interrupts the agents before audio playout is done. If the audio is done, new user audio won't cancel the tools running in background. In blocking mode, we may want to cancel the tools as well.

livekit-agents/livekit/agents/voice/agent_activity.py
toubatbrian: > the current implementation is already non-blocking function tool, a executed tool won't be awaited before the speech marked as done, when the tool returns it will create a new reply to the speech queue. So it's basically non-blocking with reply_node=when_idle, any reason to add an explicit Async Function?

Yes, you’re absolutely right—I totally missed that! After double-checking the implementation, I can confirm it’s indeed non-blocking. However, I believe the current non-blocking logic has some flaws, especially for tools that run over an extended period. Here’s an example of the behavior I’m seeing:

Let’s say we have a tool called `model_training` that takes 10 minutes to complete. When the user asks the agent to train a model, the LLM issues a tool call to `model_training`. The sequence looks like this:

```
[user] Train a model  
[llm] tool_call: model_training  
...10 minutes pass...  
[llm] Received tool_output
[llm] Respond user with "Your training has completed"
```

This works fine if there’s no interaction during those 10 minutes. But things get weird if the user speaks in the meantime. For example, if the user says “Thanks” shortly after the tool call, the model will still respond with nothing but execute the model_training tool again. This can repeat with every new user message:

```
[user] Train a model  
[llm] tool_call: model_training  
...3 seconds later...  
[user] Thanks!  
[llm] tool_call: model_training  
...3 seconds later...  
[user] Tell me a long story  
[llm] tool_call: model_training  
```
....

Why is this happening? Currently, `tool_call` and `tool_output` messages are grouped together and only added to the chat context once the tool completes. So during those 10 minutes, the LLM has no memory of having issued the tool call. As a result, it keeps reissuing the tool call, regardless of what the user says.

This limitation also affects scenarios involving parallel tool calls. For example, when the LLM issues one long-running tool call alongside several instant ones. In such cases, all tool outputs are held back and only populated into the chat context after the longest-running tool finishes. This delays responses even for tools that completed almost immediately, which could impacts the responsiveness of the system.

This is why I propose we use an async generator (`Async_F` and `Async_Raw_F` generics) to implement a more proper long-running background tool behavior. With this approach, the tool yields twice:
1. The first yield is an instant response — this ensures that the `tool_call` and a "placeholder" `tool_output` are immediately populated in the chat context.
2. The second yield delivers the actual result once the tool finishes.

This would aligns well with the current reply_mode='when_idle' behavior. And it also opens up the possibility of supporting an interrupt mechanism down the line, allowing the long-running tools to interrupt any active user / agent communication. 

One extra feature we could do is we can let tool to yield potentially more than twice, and we allow more granular control of reply mode for each individual yielded result chunk:
```py
@function_tool
async def long_running_workflow():
   yield "Tool is running, waiting for result...", "when_idle"

   result = await some_long_running_step_1()
   yield f"Finished step 1 with result: {result}", "when_idle"

   result = await some_long_running_step_2()
   yield f"Finished step 2 with result: {result}", "silent"

   ...

   result = await some_long_running_step_final()
   yield f"Finished final result: {result}", "interrupt"
```

Let me know what you think! @longcw

longcw: > For example, if the user says “Thanks” shortly after the tool call, the model will still respond with nothing but execute the model_training tool again. This can repeat with every new user message

Yeah, this is a bug we should fix, but I am not sure if the LLMs support to put a function tool message without a tool output in the chat context. How do you fix that for async tools?

regarding the `AsyncGenerator` vs `asycnio.Task` discussed in slack, I am not sure if it's a good to introduce the `AsyncGenerator`, like @theomonnom said user can get that using a `Task` with `generate_reply` in it.

But I do agree we should explicitly support `non-blocking` and `blocking` tools. Right now, it looks like non-blocking, but do have the issues of triggering multiple times and users cannot track/cancel the background tasks.

toubatbrian: > I am not sure if the LLMs support to put a function tool message without a tool output in the chat context. How do you fix that for async tools?

I would suggest not leaving a tool call unpaired. Let's ensure each tool call has a corresponding tool output for 2 reasons:

1. Our `ChatContext` does not support "dangling" tool calls. During the conversion from `ChatItem` to OpenAI format, any tool call or tool output without a matching counterpart is discarded.
2. Some providers (e.g., Anthropic) may raise errors if a tool call is passed without an associated output.

The main reason for introducing `AsyncGenerator` is because long-running tools inherently need **two** tool outputs. As mentioned above, we can’t leave the tool call on its own, which is why the initial tool output acts as a placeholder—something like `"The tool is running, waiting for result..."` to indicate that the tool has started but isn’t done yet.

Once the tool completes, we generate a second output (the actual result) and process it accordingly to add it to the chat context. `AsyncGenerator` is a natural fit for this, since it lets us emit multiple values from a single function.

That said, whether we use `AsyncGenerator` or not eventually, we should be careful with how we handle both the initial and final tool responses for long-running tools.

longcw: > Once the tool completes, we generate a second output (the actual result) and process it accordingly to add it to the chat context

this makes sense, there some options to achieve that
+ create a new pair of tool call and output to the end of the chat context
+ move the original pair to the end and modify the content
+ add a user message at the end (seems like you are using this in the code)

did you try the first two options and which one works best? In another word, the question is what is the optimal way to insert the starting/intermediate/final messages to the chat context in a long-running tool.

`AsyncGenerator` is not the critical part IMO, maybe user can call a method like `run_context.report(message)` in the regular tool to report the progress if we figured out how to insert them to the chat context.

toubatbrian: > did you try the first two options and which one works best?

bth I’m not sure which approach is definitively best. Most LLM providers’ chat interfaces likely aren’t designed or trained to handle deferred tool responses. We’ll need proper evaluation and benchmarking to make a meaningful comparison, though that’s probably a separate discussion.

To clarify on the options you mentioned, I’ll sketch out some message history examples to make sure we’re aligned. Suppose I ask the agent to execute train_model, which is a long-running tool:

**Option 1: Append a new pair of tool call/output at the end**
```
[user] "Train a model"
[tool_call] `train_model()`
[tool_output] "Model started training, waiting for result..."
[user] "Tell me a story"
[llm] "Once upon a time, ..."
(.... after a while until tool is done ...)
[tool_call] `train_model()`  # we append a "simulated" a too_call as if LLM calls the tool again
[tool_output] "Model finished training, result: ..."  # append result
```

**Option 2: Move original tool call/output to the end**
_Before tool finishes:_
```
[user] "Train a model"
[tool_call] `train_model()`
[tool_output] "Model started training, waiting for result..."
[user] "Tell me a story"
[llm] "Once upon a time, ..."
(.... waiting for tool to be done ...)
```

_After tool finishes:_
```
[user] "Train a model"
(tool call re-ordered to below)
[user] "Tell me a story"
[llm] "Once upon a time, ..."
[tool_call] `train_model()`
[tool_output] "Model finished training, result: ..."
```

**Option 3: Add a user message to indicate tool completion**
```
[user] "Train a model"
[tool_call] `train_model()`
[tool_output] "Model started training, waiting for result..."
[user] "Tell me a story"
[llm] "Once upon a time, ..."
(.... after a while until tool is done ...)
[user] "Tool `train_model` with id=xyz finished. Output: 'Model finished training, result: ...'"
```

Personally, option 3 feels the cleanest to me since it keeps the original chat history intact. Even though it uses a user message to represent tool output, LLMs usually handle this fine if prompted properly.

Option 1 modifies the history by injecting a simulated tool call, which slightly shifts the context. Option 2 reorders the original messages, which might change the flow of the conversation.

Having said that, my hunch is that we should aim to support all three patterns by either:
1. Providing internal abstractions that make it easy for users to implement them (maybe add a `long_running_response_mode` with options `simulate_extra_tool_call`, `tool_call_reorder`, and `respond_as_user_message`).
2. Documenting how to achieve the same outcomes using existing LiveKit agent APIs (e.g., ChatContext, generateReply) through guided docs.

Issue #2431, opened 2025-05-28
[NEW] Gemini Native [TTS] Support [AMAZING]


mercuryyy:
Hey Team, have you guys seen the new - https://ai.google.dev/gemini-api/docs/speech-generation

You can test it on AI studio - https://aistudio.google.com/generate-speech

Plans on supporting the Native TTS for Gemini as an addon to the google plugin? 

The quality of this module is better then 11 Labs and PlayHT by 2 Levels up. 
Multi Language support is amazing, and price is much cheaper then 11 Labs and others

No Steaming support yet only from the live API

but this is something to look at for the future

Goktug:
Actually, it seems like there is streaming support as well.

https://ai.google.dev/gemini-api/docs/speech-generation#stream

mercuryyy:
I saw that, tried few variations its not actual streaming from the API endpoint, please share snippet if you can get it working!

IPROTAGON1ST:
@mercuryyy
> I saw that, tried few variations its not actual streaming from the API endpoint, please share snippet if you can get it working!

```Python
# pip install google-genai

import base64
import mimetypes
import os
import re
import struct
from google import genai
from google.genai import types


def save_binary_file(file_name, data):
    f = open(file_name, "wb")
    f.write(data)
    f.close()
    print(f"File saved to to: {file_name}")


def generate():
    client = genai.Client(
        api_key=os.environ.get("GEMINI_API_KEY"),
    )

    model = "gemini-2.5-flash-preview-tts"
    contents = [
        types.Content(
            role="user",
            parts=[
                types.Part.from_text(text="""INSERT_INPUT_HERE"""),
            ],
        ),
    ]
    generate_content_config = types.GenerateContentConfig(
        temperature=1,
        response_modalities=[
            "audio",
        ],
        speech_config=types.SpeechConfig(
            voice_config=types.VoiceConfig(
                prebuilt_voice_config=types.PrebuiltVoiceConfig(
                    voice_name="Zephyr"
                )
            )
        ),
    )

    file_index = 0
    for chunk in client.models.generate_content_stream(
        model=model,
        contents=contents,
        config=generate_content_config,
    ):
        if (
            chunk.candidates is None
            or chunk.candidates[0].content is None
            or chunk.candidates[0].content.parts is None
        ):
            continue
        if chunk.candidates[0].content.parts[0].inline_data and chunk.candidates[0].content.parts[0].inline_data.data:
[...]

PR #2834, opened 2025-07-07
add gemini tts (beta)

longcw:
address https://github.com/livekit/agents/issues/2431

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/__init__.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/__init__.py
@@ -1,6 +1,7 @@
 from . import realtime
+from .gemini_tts import TTS as GeminiTTS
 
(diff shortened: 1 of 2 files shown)


theomonnom:
Should we provide a default prompt that automatically workaround this limitation?

longcw:
> Should we provide a default prompt that automatically workaround this limitation?

added a default instruction, the quality of the gemini tts is really good, the speed is also very slow, not sure if it's only on my end... It might be useful for ppl don't care the speed that much.

921kiyo:
Thank you for the PR! I was also looking to integrate gemini tts to livekit and came across this PR. 

For the speed, is the process of generating speech being slow, or the playing the generated speech being slow?

theomonnom (approved):

921kiyo:
The reason for the delay is because the Gemini TTS doesn't support streaming? I also experience the speech synthesis taking a few seconds for every sentence

PR #2843, opened 2025-07-08
Fix: Use synchronized transcript for interrupted session.say() responses

eliotsamuelmiller:
We started our proof of concept using `session.generate_reply()`.

When the user interrupted the agent, it was useful to have only the part which was actually spoken for `interrupted` messages.

When we moved to generating responses separately in our platform and then using `session.say()`, `interrupted` messages would contain the full content - not just the portion which was read aloud.

This PR aims to align the behaviour of `session.say()` to `session.generate_reply()` by only adding the spoken content to the chat context for interrupted messages.


theomonnom (approved):
Thanks!

PR #2852, opened 2025-07-09
support resume agent from a false interruption

longcw:


examples/voice_agents/resume_interrupted_agent.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @session.on("agent_speech_resume")
+    def _agent_speech_resume(ev: AgentSpeechResumeEvent):
+        logger.info(
+            "Resuming agent from interruption",
+            extra={"instructions": ev.instructions, "forwarded_text": ev.forwarded_text},
+        )
+        session.generate_reply(instructions=ev.instructions or NOT_GIVEN)
+
+    await session.start(agent=Agent(instructions="You are a helpful assistant."), room=ctx.room)
```
toubatbrian: Do we expect the resume to be handled automatically? e.g. maybe inside a pipeline reply task, we starts streaming llm/tts/tools and as soon as interrupt happens, we pause audio output from receiving audio frames from tts and somehow buffer it and schedule a resume event. When the timer passes the resume delay threshold, we then re-stream those buffered frames back to audio output? This way we can reuse a single tts stream, segmented by the small interrupts. So it could look like

```
[user] Tell me a story.
[llm] Once upon a time, in ...
[user] wait, ok go on
[llm] ...a small village where ...
```

where the TTS "Once upon a time, in a small village where..." is cut into two pieces by the interrupts.

However, this might require a lot of code changes on the pipeline_reply_task as current pipeline implementation might not support resume-after-interrupt.

longcw: that's another way of resume mentioned by @theomonnom as well. it's possible if we add an inactive state to the SpeechHandle. but I am not sure if it's worth to do that for now, the LLM is good enough to start a new generation from where it left IMO.

toubatbrian: Yep, no doubt about the LLM’s intelligence, my main concern is around cost. A false interrupt could potentially be one LLM call instead of two. 

> but I am not sure if it's worth to do that for now, the LLM is good enough to start a new generation from where it left IMO.

Agree, let’s move forward with a new generation for now and maybe revisit the more complex path later if needed.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -153,6 +156,7 @@ def __init__(
         max_tool_steps: int = 3,
         video_sampler: NotGivenOr[_VideoSampler | None] = NOT_GIVEN,
         user_away_timeout: float | None = 15.0,
+        agent_false_interruption_timeout: float | None = 4.0,
```
theomonnom, line 159: 4.0 seems like a lot? I'm not sure..

PR #2853, opened 2025-07-09
Prevent console mode from crashing

donalffons:
This PR fixes the issue described in #2854.


theomonnom (approved):

theomonnom:
Thanks! I think this can happen when the delay between the speaker and the microphone is too large

donalffons:
Thank you for merging this!

PR #2887, opened 2025-07-13
update/mistralai-plugins

fabitokki:
Update mistralai-plugins 
Added an example of LLM usage in examples
Few fixes 
set provider format now use group tools calls like every other providers


fabitokki:
Hello @theomonnom 

Any news from this PR ? :)

fabitokki:
I pulled main to integrate latest change

fabitokki:
Hello @davidzhao ! I think Theo might not be available atm. Is there anyone that could check this PR :) ?

Issue #2885, opened 2025-07-13
TTS characters always 0 and tts metrics not shown


aayush1036:
I have created an agent with the following setup
1. LLM fallback (openai and claude)
2. TTS fallback (elevenlabs and cartesia)
3. Turn detection plugin
4. Deepgram plugin 
5. Noise cancellation plugin
6. Silero VAD

I want to maintain latency logs for debugging purposes (saw this latency calculation on their website - https://docs.livekit.io/agents/build/metrics/#measuring-conversation-latency)

I'm getting all other metrics except the metrics related to TTS. 

My agent code 

```python3
import logging

logging.basicConfig(
    level=logging.DEBUG,
    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
    handlers=[
        logging.StreamHandler(),
        logging.FileHandler('test.log')
    ]
)

logger = logging.getLogger("tavus-avatar-example")

from livekit import agents
from livekit.agents import AgentSession, RoomInputOptions, MetricsCollectedEvent, cli, WorkerOptions, WorkerType, Agent, RoomOutputOptions
from livekit.plugins import tavus, openai, anthropic, elevenlabs, cartesia, noise_cancellation, deepgram, silero
from livekit.agents.llm import FallbackAdapter as LLMFallbackAdapter
from livekit.agents.tts import FallbackAdapter as TTSFallbackAdapter
from livekit.plugins.turn_detector.english import EnglishModel
from livekit.agents.metrics import UsageCollector
from credentials import TAUVUS_REPLICA_ID, TAUVUS_PERSONA_ID, ELEVENLABS_VOICE_ID, CARTESIA_VOICE_ID
from dotenv import load_dotenv
load_dotenv()

async def entrypoint(ctx: agents.JobContext):
    await ctx.connect()
    llm = LLMFallbackAdapter(
        llm=[
            openai.LLM(model='gpt-4o'),
            anthropic.LLM(model='claude-3-5-sonnet-20240620')
        ]
    )

    tts = TTSFallbackAdapter(
        tts=[
            elevenlabs.TTS(voice_id=ELEVENLABS_VOICE_ID, model='eleven_turbo_v2'),
            cartesia.TTS(voice=CARTESIA_VOICE_ID, model='sonic-2')
        ]
    )
    stt = deepgram.STT(model='nova-3-general')
    vad = silero.VAD.load(min_silence_duration=0.5)
    turn_detection = EnglishModel()
[...]

PR #2890, opened 2025-07-14
fix TTS fallback adapter metrics_collected event

longcw:
fix https://github.com/livekit/agents/issues/2885

This pr forward the tts metrics to the fallback adapter. Another solution is to call `_mark_started` in the fallback adapter, but this makes the `ttfb` much higher than the regular tts where `_mark_started` is called after the sentence/word tokenizer.


davidzhao (approved):
lgtm

theomonnom (approved):

PR #2895, opened 2025-07-14
add option to cancell ambient sound while in think mode

yasfatft:


davidzhao:
what is the use case here? why would you want to disable ambient sound in this case?

yasfatft:
I kinda agree that this is not a vary popular change or neither neccessary, because it corrupt the flow of agent that tryies to mimic a human operator bahaiviour or so, but if it is acknoledged that the responser is not a human and it is indeed AI agent, then in a case which the participant is a sip client, the ambient sound is kinda like a noise which ensure the client the call is ongoing still (kinda like a call status feedback) and not ended (or buged or failed) because there is no ui feedback (the functionality of the ambient sound is not just a background sound anymore) and when switching to the thinking mode, the mix of those two sound may not be so pleasant!

yasfatft:
No suggestion or comment!?

PR #2873, opened 2025-07-11
add opentelemetry trace

longcw:
replace #2844, address https://github.com/livekit/agents/issues/2260

supports
+ user_turn: user_speaking, eou_detection
+ assistant_turn: llm_node, tts_node, tts_request, agent_speaking, function_tool
+ handoff: on_exit, on_enter
+ realtime model: realtime_assistant_turn
+ llm and tts metrics in llm/tts_request

example (updated): https://us.cloud.langfuse.com/project/cmcvboshr08hoad071me1a5dj/traces/73eb07fa53663b10551f256bcec88286?timestamp=2025-07-14T08%3A16%3A31.812Z&display=timeline&observation=7ff6a3a683365762

Issue #2260, opened 2025-03-26
feat: add opentelemetry instrumentation for tracing


marcklingen:
## Context

I'm one of the maintainers of [github.com/langfuse/langfuse](https://github.com/langfuse/langfuse). livekit is growing in popularity and I've gotten multiple requests by community members who'd like to get traces of their livekit applications in order to evaluate them.

## Suggestion

I'd suggest to follow the example of the AI SDK ([docs](https://sdk.vercel.ai/docs/ai-sdk-core/telemetry)) by adding native Opentelemetry-based instrumentation to livekit. Thereby, the implementation is vendor-agnostic and teams can use an LLM-specific backend like Langfuse (tracing, evals, prompt management) or their traditional observability platforms (Grafana, Datadog, ...).

Potential addition next to traces: Exporting metrics ([livekit docs](https://docs.livekit.io/agents/build/metrics/)) via OpenTelemetry would allow teams to easily track them within their observability stack.

## Resources

- https://opentelemetry.io/blog/2025/ai-agent-observability/
- https://opentelemetry.io/docs/specs/semconv/gen-ai/
- https://langfuse.com/docs/opentelemetry/get-started
- https://github.com/orgs/langfuse/discussions/5235

davidzhao:
@marcklingen we'd love to get this integrated into the framework. will take a look at this

VadzimBelski-ScienceSoft:
+1

theomonnom:
Hey, we're going to work on it this month!

marcklingen:
Excited for it, ping me in case you have a preview. Would love to support this in a good state in Langfuse once this is available.

Side note: pipecat just released this. Might be interesting to check out their attribute mapping.

vishal-bluebash:
+1

Trivedi-grv:
+1

TomCC7:
+1

isaaccs:
+1

xavphi:
+1

danishshaik:
+1

us-asad:
+1

Dramilio:
+1

ferreirapmartin:
+1

ShaheerCH:
hello. any update regarding this

ma08:
+1

nischalj10:
Hey team - any update on this?

longcw:
we are working on it in https://github.com/livekit/agents/pull/2844, will be ready soon!

[...]

PR #2901, opened 2025-07-15
feat(otel): add gen_ai attributes to llm_request and add plugin override

dmcaulay:
Enhances the existing OpenTelemetry tracing implementation (#2873) by adding standard GenAI semantic conventions to LLM spans and enabling plugins to override with official provider instrumentation for more accurate tracing.

### **OpenTelemetry GenAI Attributes**
- **Standard generation span**: Added OpenTelemetry GenAI semantic conventions so `llm_request` spans are automatically recognized as generation spans
- **Complete LLM layer attributes**: Includes all standard attributes available at the LLM abstraction layer (model, provider, tokens, etc.)
- **Conversation events**: Structured events for system/user/assistant messages with multimodal support
- **Tool information**: Function parameter tracking in span attributes

### **Improved Retry Observability**
- **Per-attempt tracing**: Each retry attempt now gets its own span for granular observability of retry behavior
- **Individual timing**: See duration and status of each specific attempt
- **Error attribution**: Exceptions recorded on the specific attempt that failed

### **Plugin Override Support**
- **Provider instrumentation detection**: New `has_opentelemetry_instrumentation` property allows plugins to indicate official OpenTelemetry support
- **Accurate provider spans**: When available, plugins can override the `llm_request` span with official provider instrumentation
- **Removes LiveKit abstractions**: Provider spans capture exactly what's sent to the client, without LiveKit's LLM abstraction layer modifications

### **Langfuse Integration**
- **Time to first token**: Added `langfuse.observation.completion_start_time` attribute with ISO UTC timestamp

### **OpenAI Example Implementation**
- **Official instrumentation**: Integrated `opentelemetry-instrumentation-openai` package as example of plugin override
- **Auto-initialization**: OpenAI instrumentor automatically initialized in plugin startup
- **Model/provider properties**: Expose accurate model and provider information for span attributes

### Files Modified
- `livekit-agents/livekit/agents/llm/llm.py` - Main tracing logic with hybrid approach
- `livekit-agents/livekit/agents/debug/trace_types.py` - OpenTelemetry GenAI attribute constants
- `livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py` - Model/provider properties
- `livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/__init__.py` - Auto-instrumentation setup
- `examples/voice_agents/langfuse_trace.py` - Adjusted global tracer provider setup

## Testing Status

✅ Basic functionality verified  
✅ OpenAI plugin integration tested  
✅ New attribute availability confirmed  
✅ Import compatibility validated  
⚠️ **Video frame encoding needs testing** - Multimodal content with `rtc.VideoFrame` requires validation

## Questions for Review

1. **Plugin override approach**: Is the `has_opentelemetry_instrumentation` property the right way to detect and enable provider instrumentation override? (Note: I chose override vs. nesting to avoid double-counting generation spans and maintain clear semantic boundaries - when provider instrumentation is available, it replaces the `llm_request` span entirely rather than creating nested spans)
2. **GenAI attribute completeness**: Are we missing any important OpenTelemetry GenAI semantic conventions at the LLM layer?
3. **Video frame handling**: The current approach encodes `rtc.VideoFrame` to base64 JPEG - is this the preferred method for multimodal traces?
4. **Langfuse-specific attributes**: Should we include custom attributes like `langfuse.observation.completion_start_time` for vendor-specific requirements, or keep the tracing implementation purely standards-compliant?

## Related

- Enhances #2873 
- Addresses https://github.com/livekit/agents/issues/2260
- Follows OpenTelemetry GenAI conventions: https://opentelemetry.io/docs/specs/semconv/gen-ai/

### Default LLM Request Span
Screenshot 2025-07-14 at 10 35 21 PM

### Plugin Override
Screenshot 2025-07-14 at 10 35 50 PM

### DIFF
--- a/examples/voice_agents/langfuse_trace.py
+++ b/examples/voice_agents/langfuse_trace.py
@@ -23,6 +23,7 @@ load_dotenv()
 def setup_langfuse(
     host: str | None = None, public_key: str | None = None, secret_key: str | None = None
 ):
+    from opentelemetry import trace
     from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
     from opentelemetry.sdk.trace import TracerProvider
     from opentelemetry.sdk.trace.export import BatchSpanProcessor
@@ -41,6 +42,9 @@ def setup_langfuse(
     trace_provider = TracerProvider()
     trace_provider.add_span_processor(BatchSpanProcessor(OTLPSpanExporter()))
     set_tracer_provider(trace_provider)
+    
+    # Also set the global OpenTelemetry tracer provider for OpenAI instrumentation
+    trace.set_tracer_provider(trace_provider)
 
 
 @function_tool
@@ -101,7 +105,7 @@ class Alloy(Agent):
 async def entrypoint(ctx: JobContext):
     setup_langfuse()  # set up the langfuse tracer
 
-    session = AgentSession(vad=silero.VAD.load())
+    session = AgentSession()
 
     @session.on("metrics_collected")
     def _on_metrics_collected(ev: MetricsCollectedEvent):
--- a/examples/voice_agents/langfuse_trace.py
+++ b/examples/voice_agents/langfuse_trace.py
@@ -8,7 +8,7 @@ from livekit.agents import Agent, AgentSession, JobContext, RunContext, WorkerOp
 from livekit.agents.debug import set_tracer_provider
 from livekit.agents.llm import function_tool
 from livekit.agents.voice import MetricsCollectedEvent
-from livekit.plugins import deepgram, openai, silero
+from livekit.plugins import deepgram, openai
 from livekit.plugins.turn_detector.multilingual import MultilingualModel
 
 logger = logging.getLogger("langfuse-trace-example")
@@ -42,7 +42,7 @@ def setup_langfuse(
     trace_provider = TracerProvider()
     trace_provider.add_span_processor(BatchSpanProcessor(OTLPSpanExporter()))
     set_tracer_provider(trace_provider)
-    
+
     # Also set the global OpenTelemetry tracer provider for OpenAI instrumentation
     trace.set_tracer_provider(trace_provider)
 
--- a/livekit-agents/livekit/agents/debug/trace_types.py
+++ b/livekit-agents/livekit/agents/debug/trace_types.py
@@ -1,6 +1,5 @@
 # Import official OpenTelemetry semantic conventions
 from opentelemetry.semconv._incubating.attributes import gen_ai_attributes
-from opentelemetry.semconv._incubating.attributes import error_attributes
 
 ATTR_SPEECH_ID = "lk.speech_id"
 ATTR_AGENT_LABEL = "lk.agent_label"
--- a/livekit-agents/livekit/agents/debug/trace_types.py
+++ b/livekit-agents/livekit/agents/debug/trace_types.py
@@ -1,3 +1,7 @@
+# Import official OpenTelemetry semantic conventions
+from opentelemetry.semconv._incubating.attributes import gen_ai_attributes
+from opentelemetry.semconv._incubating.attributes import error_attributes
+
 ATTR_SPEECH_ID = "lk.speech_id"
 ATTR_AGENT_LABEL = "lk.agent_label"
 ATTR_START_TIME = "lk.start_time"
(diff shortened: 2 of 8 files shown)


dmcaulay:
cc @longcw and @theomonnom what do you think?

examples/voice_agents/langfuse_trace.py
```diff
@@ -101,7 +105,7 @@ async def transfer_to_kelly(self) -> Agent:
 async def entrypoint(ctx: JobContext):
     setup_langfuse()  # set up the langfuse tracer
 
-    session = AgentSession(vad=silero.VAD.load())
```
longcw, line 104: can you keep the vad in the example, the pipeline agent needs it.

livekit-agents/livekit/agents/debug/trace_types.py
```diff
@@ -44,3 +47,31 @@
...
+ATTR_GEN_AI_SYSTEM = gen_ai_attributes.GEN_AI_SYSTEM
+ATTR_GEN_AI_OPERATION_NAME = gen_ai_attributes.GEN_AI_OPERATION_NAME
+ATTR_GEN_AI_RESPONSE_ID = gen_ai_attributes.GEN_AI_RESPONSE_ID
+ATTR_GEN_AI_RESPONSE_MODEL = gen_ai_attributes.GEN_AI_RESPONSE_MODEL
+ATTR_GEN_AI_USAGE_INPUT_TOKENS = gen_ai_attributes.GEN_AI_USAGE_INPUT_TOKENS
+ATTR_GEN_AI_USAGE_OUTPUT_TOKENS = gen_ai_attributes.GEN_AI_USAGE_OUTPUT_TOKENS
+
+# OpenLLMetry standard attributes
+ATTR_LLM_USAGE_TOTAL_TOKENS = "llm.usage.total_tokens"
+ATTR_LLM_IS_STREAMING = "llm.is_streaming"
```
longcw, line 62: can you add a link where these attributes are defined, I found this https://www.traceloop.com/docs/openllmetry/contributing/semantic-conventions#traces-definitions but there is no `llm.is_streaming`.

longcw (commented):
thanks! overall it looks good to me, if you don't mind we can merge this first and I'll do some cleaning.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/__init__.py
```diff
@@ -54,6 +54,17 @@ class OpenAIPlugin(Plugin):
     def __init__(self) -> None:
         super().__init__(__name__, __version__, __package__, logger)
 
+        # Initialize OpenAI OpenTelemetry instrumentation
+        try:
+            from opentelemetry.instrumentation.openai import OpenAIInstrumentor
```
longcw, line 59: I think we don't need to enable it by default. The trace added by this OpenAIInstrumentor is basically the same as attributes added to the `llm_request` in this pr, using `llm_request` is more consistent for all llm providers. users can enable it from the entrypoint if they want to use it.

longcw:
created a pr with some cleanups https://github.com/livekit/agents/pull/2905: always create the `llm_request` span and made openai instrument optional, removed some attributes that not needed.

examples/voice_agents/langfuse_trace.py
dmcaulay: yeah, my system had issues finding the model so i killed it for testing, since it wasn't crucial for my use case.

livekit-agents/livekit/agents/debug/trace_types.py
dmcaulay: i pulled them from openllmetry code, not sure why it's not in their docs

attribute names- https://github.com/traceloop/openllmetry/blob/e22a48891ff86e2ac0153b50886c8a38246f2406/packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py#L36

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/__init__.py
dmcaulay: i liked it because it tells me exactly what's sent to the provider, but it's fine if you think the override capability is too much. i added a comment on your pr for why i didn't include both.

dmcaulay:
thanks for the review and quick merge! should i close this pr?

dmcaulay:
> created a pr with some cleanups #2905: always create the `llm_request` span and made openai instrument optional, removed some attributes that not needed.

the main issue with the decorator is that you don't see failed calls- latency and other metrics like token usage may be off.

longcw:
yes you can close this one, and feel free to create a new issue or pr if needed. I'll remove the comment about the openai instrumentation, basically we don't want each LLM provider has a different span. The usage, error and ttft are all added to the `llm_request` span now.

PR #2905, opened 2025-07-15
add gen_ai attributes to llm_request

longcw:
clean up https://github.com/livekit/agents/pull/2901

example: https://us.cloud.langfuse.com/project/cmcvboshr08hoad071me1a5dj/traces/60bf505145bf9f12bb96f590637fbd7b?timestamp=2025-07-15T10%3A30%3A19.372Z&display=preview&observation=dc510ba2c7b5d8d2


examples/voice_agents/langfuse_trace.py
```diff
@@ -42,6 +42,17 @@ def setup_langfuse(
     trace_provider.add_span_processor(BatchSpanProcessor(OTLPSpanExporter()))
     set_tracer_provider(trace_provider)
 
+    # (optional) enable OpenAI OpenTelemetry instrumentation
```
dmcaulay, line 45: you may want to remove this comment. if llm_request is the generation span, then adding both will most likely result in double counting generation calls, tokens, etc in most observability platforms.

dmcaulay: you can see the double counting here.
Screenshot 2025-07-15 at 11 06 26 AM

dmcaulay: it looks like it's double counting the price. not sure why the token usage is different, maybe a livekit calculation issue or something with langfuse?

longcw: actually, the token and price is not double counting when the `openai.chat` is nested under the `llm_request`. 

the `openai.chat` has a different token counts (or it doesn't expose the token explicitly so the langfuse is counting it differently). The usage count in `llm_request` is directly from the api returned usage, and it's the same when you use [opentelemetry-instrumentation-openai-v2](https://pypi.org/project/opentelemetry-instrumentation-openai-v2/).

longcw: maybe the prices is double counted? I'll double check, but anyway we can remove the comment. put it there just in case someone curious about the llm-specific instrumentation.

PR #2906, opened 2025-07-15
[DRAFT] Add inference process health check endpoint

alfredguiaugment:


theomonnom:
It's a great call, I think you can do it directly inside /health

theomonnom (approved):
Thanks lgtm!

PR #2911, opened 2025-07-15
fix Datastream IO race

theomonnom:


bcherry:
this seems to resolve the issue for me

PR #2919, opened 2025-07-16
remove openai prewarm

theomonnom:


amjd:
Hey @theomonnom, curious.. why was this change removed? Are there any plans to bring this back if the current implementation was imperfect?

PR #2926, opened 2025-07-16
fix: don't hardcode role=assistant on openai

jdeloach:
There are several valid potential roles that could be returned: https://github.com/openai/openai-python/blob/35df552d032873b62c2ae127a0efce60947dbed0/src/openai/types/chat/chat_completion_chunk.py#L73

This avoids overwriting them, to give more flexibility when building systems with the OpenAI LLM implementation as a base.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -707,7 +707,7 @@ class LLMStream(llm.LLMStream):
                     call_chunk = llm.ChatChunk(
                         id=id,
                         delta=llm.ChoiceDelta(
-                            role="assistant",
+                            role=delta.role,
                             content=delta.content,
                             tool_calls=[
                                 llm.FunctionToolCall(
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -707,7 +707,7 @@ def _parse_choice(
                     call_chunk = llm.ChatChunk(
                         id=id,
                         delta=llm.ChoiceDelta(
-                            role="assistant",
+                            role=delta.role,
```
davidzhao, line 710: tool call would need to be mapped this way

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -735,7 +735,7 @@ def _parse_choice(
             call_chunk = llm.ChatChunk(
                 id=id,
                 delta=llm.ChoiceDelta(
-                    role="assistant",
```
davidzhao, line 738: same as the above

PR #2943, opened 2025-07-18
clean up avatar example and add retry for datastream io rpc call

longcw:
1. rewrote the audio_wave example
2. added a retry logic for playback finished rpc call before we have that in rust


examples/avatar_agents/audio_wave/agent_worker.py
```diff
@@ -36,9 +36,24 @@ class AvatarConnectionInfo:
...
+    try:
+        decoded = ctx.decode_token()
+        agent_identity = decoded["sub"]
+    except (RuntimeError, KeyError):
+        if not ctx.room.isconnected():
+            raise RuntimeError(
+                "local participant identity not found in token, and room is not connected"
+            ) from None
+        agent_identity = ctx.room.local_participant.identity
+
```
theomonnom: I think we should provide a different API for the token. 

Something like ctx.token_details(), it doesn't check any signature and return the typed api.VideoGrants, wdyt?

theomonnom: https://github.com/livekit/python-sdks/blob/main/livekit-api/livekit/api/access_token.py

longcw: maybe an `api.Claims`? then read the local participant identity from `claims.identity`

theomonnom: Oh yeah sry, api.Claims lgtm!

theomonnom: This way we don't even need to catch for errors. It's safe to assume the token `content` is always OK (we ignore validation)

longcw: updated

livekit-agents/livekit/agents/job.py
```diff
@@ -439,6 +442,39 @@ def decode_token(self, api_secret: NotGivenOr[str] = NOT_GIVEN) -> dict[str, Any
             api_secret = ""
         return jwt.decode(self._info.token, api_secret, options=options, algorithms=["HS256"])  # type: ignore
 
+    def token_details(self) -> Claims:
+        claims = self.decode_token()
```
theomonnom: IMO we should make decode_token private.

livekit-agents/livekit/agents/job.py
```diff
@@ -439,6 +442,39 @@ def decode_token(self, api_secret: NotGivenOr[str] = NOT_GIVEN) -> dict[str, Any
...
+
+        if sip_dict := claims.get("sip"):
+            sip_dict = {camel_to_snake(k): v for k, v in sip_dict.items()}
+            sip_dict = {k: v for k, v in sip_dict.items() if k in SIPGrants.__dataclass_fields__}
+            claims["sip"] = SIPGrants(**sip_dict)
+
+        if room_config := claims.get("roomConfig"):
+            claims["roomConfig"] = ParseDict(
+                room_config, RoomConfiguration(), ignore_unknown_fields=True
+            )
```
theomonnom: https://github.com/livekit/python-sdks/blob/0389db8349de3dc83bbc4fe7a1b0eae53d028d57/livekit-api/livekit/api/access_token.py#L209

I'm thinking that we could add a new argument to the TokenVerifier in livekit-api to skip verifications. So we don't duplicate code

theomonnom: Otherwise it lgtm!

longcw: created a pr for python sdk https://github.com/livekit/python-sdks/pull/473

longcw:
needs https://github.com/livekit/python-sdks/pull/473

livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
```diff
@@ -225,33 +233,60 @@ def _handle_stream_received(
...
+                except rtc.RpcError as e:
+                    if retry_count == self._rpc_max_retries - 1:
+                        logger.error(
+                            f"failed to notify playback finished after {retry_count + 1} retries",
+                            exc_info=e,
+                        )
+                        raise
+                    retry_count += 1
+                    logger.warning("failed to notify the agent playback finished, retrying...")
+                    await asyncio.sleep(0.01)
```
theomonnom: 10ms is quite small

theomonnom (approved):

livekit-agents/livekit/agents/job.py
```diff
@@ -432,12 +431,8 @@ def _participant_available(self, p: rtc.RemoteParticipant) -> None:
                 lambda _, coro=coro: self._participant_tasks.pop((p.identity, coro))  # type: ignore
             )
 
-    def decode_token(self, api_secret: NotGivenOr[str] = NOT_GIVEN) -> dict[str, Any]:
-        options = {}
-        if not is_given(api_secret):
-            options["verify_signature"] = False
-            api_secret = ""
-        return jwt.decode(self._info.token, api_secret, options=options, algorithms=["HS256"])  # type: ignore
+    def token_details(self) -> Claims:
```
theomonnom:
```suggestion
    def token_claims(self) -> Claims:
```

PR #2949, opened 2025-07-20
fix duplicated user messages when preemptive generation canceled

longcw:


theomonnom (approved):

quocatsg:
@longcw we upgraded Livekit agents to 1.2 and started observing duplicated user messages even we don't enable preemptive generation. It's not consistently happening to every turn though. Would it be something resolved by this MR? Thanks!

longcw:
@quocatsg what LLM model are you using, can you share the steps for reproducing?

quocatsg:
@longcw we're using a custom agent that sends to an in-process custom developed LLM orchestrator (backed by Langgraph). It's not consistently happening, so we can't know the pattern to reproduce it yet. We could observe it just by having a few normal calls, as we did in the previous Livekit agent versions, nothing special involved.

PR #2953, opened 2025-07-21
tune vad min_silence_duration and min_endpointing_delay

longcw:


davidzhao (approved):

Panmax:
May I ask why this parameter needs to be modified? I found that when using version 1.2.2 of the agent, it tends to enter the thinking state too easily, which does not meet my expectations.

longcw:
@Panmax  It's a trade off between responding speed and the end of user tuen accuracy. You can adjust the min_endingpoint_delay based on your needs.

PR #2960, opened 2025-07-22
fix wait_for_playout to wait for generation instead of done future

longcw:


FilipKubis:
Hi, I am facing issues with waiting for speech playout hanging indefinitely in the new half duplex regime (realtime model + custom tts). 

Is this something which fixes the problem? I'm also not sure if this change will work in other setups (e.g.: realtime only or STT -> LLM -> TTS). What would be the recommended approach for a hotfix?

longcw:
> Hi, I am facing issues with waiting for speech playout hanging indefinitely in the new half duplex regime (realtime model + custom tts).
> 
> Is this something which fixes the problem? I'm also not sure if this change will work in other setups (e.g.: realtime only or STT -> LLM -> TTS).

I think so, this should fix the issue for both pipeline mode and half-duplex mode.

FilipKubis:
> > Hi, I am facing issues with waiting for speech playout hanging indefinitely in the new half duplex regime (realtime model + custom tts).
> > Is this something which fixes the problem? I'm also not sure if this change will work in other setups (e.g.: realtime only or STT -> LLM -> TTS).
> 
> I think so, this should fix the issue for both pipeline mode and half-duplex mode.

Awesome, tried it out and seems to do the trick. Thanks Long!

longcw:
replaced by https://github.com/livekit/agents/pull/2966

Issue #1378, opened 2025-01-15
Feature Request: Ultravox example


thiswillbeyourgithub:
Hi,

[Ultravox](https://github.com/fixie-ai/ultravox) is a suite of open weight models that are designed for getting the time to first token as low as possible with audio input. Basically they trained a good and fast projector to project the whisper large v3 encoder into llama 4.1 LLMs, both in 8B and 70B size.

I think it would be a great fit for livekit's agents so it would be nice to add an example and demo for it!

zkoch:
Thanks -- happy to work with folks at Livekit to make this happen!

therealron:
Any update on this?

psinojiya:
Hello Everyone, any update on this?

thiswillbeyourgithub:
I'm very interested still, especially as nothing as simple as ollama exists for chat (audio+text) but I lack the skills to implement it. I'm still surprised no one has created it since it seems to be in the best interest of all parties involved: livekit, ultravox, kyutai (they made moshi), etc. And everyone seems to advertise their solution as easy to implement.

jayeshp19:
Hi It will be very easy to implement with our next big release.

psinojiya:
Sounds great @jayeshp19 , any approximate idea on dates for that?

jayeshp19:
In the next week or two, you can track progress here: https://github.com/livekit/agents/tree/dev-1.0

mdwoicke:
@jayeshp19 Any current updates on this? I am not seeing anything related to this in the link provided. Thanks!

NelsonHotel:
I'm looking to use ultravox and livekit would love to know where this project stands.

akarray:
There isn't any working example yet for the community ?

ChenghaoMou:
I am working on a PR for this: https://github.com/livekit/agents/pull/2409

So far I got it working. But most of the code is generated by AI so I have to double check.

nischalj10:
Any update on this? Would really like ultravox work with livekit

thevijaydeore:
Is any  update on the livekit  support the ultravox

tongclement:
+1

nischalj10:
hey @jayeshp19 - just following up on this

tongclement:
Would be great if it would support fine-tuned versions of the model too (fine tuned using custom data to support new languages like cantonese) 
(still investigating how this would work on a technical level)

mercuryyy:
Any update ?

[...]

PR #2992, opened 2025-07-24
Add Ultravox Realtime API Plugin

moadel321:
# Ultravox Plugin for LiveKit Agents

This PR introduces a new Ultravox plugin for the LiveKit Agents framework, enabling real-time speech-to-speech interactions using the Ultravox Realtime API

Used the Websockets integration guide found here : https://docs.ultravox.ai/apps/websockets

## Components

**realtime_model.py** - Primary implementation containing RealtimeModel and RealtimeSession classes. Handles WebSocket connection management, binary audio streaming, client tool execution, and API parameter configuration including temperature, language hints, and voice selection.

**events.py** - Pydantic event models for Ultravox WebSocket protocol. Defines bidirectional message types including ping/pong, transcripts, tool invocations, debug events, and state management following the Ultravox data message specification.

**models.py** - Type definitions for available Ultravox models and voices. Includes static validation lists and default fallback configurations for model and voice selection.

**agent.py** - test example implementation demonstrating plugin usage with tools 

## API Authentication 

You'll need an API key from Ultravox. Set it as an environment variable:
export ULTRAVOX_API_KEY="your_api_key_here"

## Manual Tests Done 

Test Case 1: "What time is it in New York?"

Tests: get_time tool
Expected: Agent calls the time tool

Test Case 2: "Look up customer John Smith"
Tests: lookup_customer tool
Expected: Returns "Customer: John Smith, Account: ACC-001, Status: VIP, Balance: $1,250" from mock 

Test Case 3: "Search for pricing information"
Tests: search_knowledge tool
Expected: Returns pricing details from knowledge base mock

Test Case 4: Say "Hi, this is John Smith"
Tests: Auto context injection via on_user_turn_completed
Expected: Agent acknowledges VIP status and account info without using tools
Look for logs: [RAG] Auto-injecting context for: John Smith

Test Case 5: Ask "What are your hours?"
Expected: Agent provides hours info immediately
Look for logs: [RAG] Auto-injecting knowledge for: hours

Tried to run /tests running, but i'm on windows 11, so:

```
error: Distribution `bithuman==0.5.20 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're on Windows (`win_amd64`), but `bithuman` (v0.5.20) only has wheels for the following platforms: `manylinux_2_28_aarch64`, `manylinux_2_28_x86_64`, `macosx_10_9_x86_64`, `macosx_11_0_arm64`
```

## API Parameters

Supports essential Ultravox API parameters:
- temperature - Response randomness control
- languageHint - Primary language specification
- maxDuration - Session duration limits
- enableGreetingPrompt - Automatic greeting configuration
- timeExceededMessage - Custom timeout messaging
- vadSettings - Voice activity detection configuration
- firstSpeaker - Initial speaker designation

new parameters can be added from here in the future : https://docs.ultravox.ai/api-reference/calls/calls-post

## Known Limitations

- Struggles with numbers and dates, but Ultravox v0.6 should improve performance

### DIFF
--- a/livekit-plugins/livekit-plugins-ultravox/examples/agent.py
+++ b/livekit-plugins/livekit-plugins-ultravox/examples/agent.py
@@ -23,7 +23,6 @@ from livekit.agents.voice.events import (
     MetricsCollectedEvent,
     UserInputTranscribedEvent,
 )
-from livekit.agents import metrics 
 from livekit.plugins import silero
 from livekit.plugins.ultravox.realtime import RealtimeModel
 


moadel321:
unit tests are failing due to "httpcore.LocalProtocolError: Illegal header value b'Bearer '" 

This looks like an API key is missing.... don't want to modify any tests so I'll wait for a response

moadel321:
Linking https://github.com/livekit/agents/issues/1378

davidzhao:
thank you! @moadel321 

I'll be reviewing & testing this in a few days.

moadel321:
> thank you! @moadel321
> 
> I'll be reviewing & testing this in a few days.

Anytime ! 
First PR to LiveKit Agents... open to feedback and improvements

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1162 @@
...
+        User messages are sent as regular text messages.
+        Assistant messages are skipped (managed by Ultravox internally).
+        Function calls/results are handled via the existing tool mechanism.
+
+        Args:
+            chat_ctx: The updated chat context to inject
+        """
+        logger.info(f"[ultravox] Injecting context with {len(chat_ctx.items)} items")
+
+        for item in chat_ctx.items:
```
longcw: it needs to compute the diff between the `chat_ctx` and remote chat ctx, and it seems this model only support append system or user messages?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1162 @@
...
+                logger.debug(f"[ultravox] Injecting user message: {content_text[:100]}...")
+                self._send_client_event(
+                    InputTextMessageEvent(text=content_text, defer_response=False)
+                )
+            # Skip assistant messages (handled by Ultravox)
+            # Skip function calls (handled by existing tool mechanism)
+
+    async def update_tools(self, tools: list[llm.FunctionTool | llm.RawFunctionTool]) -> None:
+        """Update the available tools."""
+        self._tools.update_tools(tools)
```
longcw, line 468: also need to restart the session?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1162 @@
...
+                        "generate_reply timed out waiting for generation_created event."
+                    )
+                )
+                if self._pending_generation_fut is fut:
+                    self._pending_generation_fut = None
+
+        timeout_handle = asyncio.get_event_loop().call_later(5.0, _on_timeout)
+        fut.add_done_callback(lambda _: timeout_handle.cancel())
+        # Starting a new generation here since the output medium might be changed.
+        self._start_new_generation()
```
longcw: is this enough to notify the server to start a new response? I didn't see it send a message to the server.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1162 @@
...
+    @utils.log_exceptions(logger=logger)
+    def generate_reply(
+        self, *, instructions: NotGivenOr[str] = NOT_GIVEN
+    ) -> asyncio.Future[llm.GenerationCreatedEvent]:
+        """Generate a reply from the LLM based on the instructions."""
+        if self._pending_generation_fut and not self._pending_generation_fut.done():
+            logger.warning(
+                "generate_reply called while another generation is pending, cancelling previous."
+            )
+            self._pending_generation_fut.cancel("Superseded by new generate_reply call")
```
longcw, line 527: how to determine the new response is the one request this time but not from the previous `generate_reply`?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1162 @@
...
+                # Set first token timestamp on first text delta (TTFT measurement)
+                if msg_gen._first_token_timestamp is None:
+                    msg_gen._first_token_timestamp = time.time()
+                    ttft = msg_gen._first_token_timestamp - msg_gen._created_timestamp
+                    logger.info(f"[ultravox] first text token received - TTFT: {ttft:.3f}s")
+                msg_gen.text_ch.send_nowait(event.delta)
+
+            # Handle final transcript
+            if event.final:
+                if event.text and not event.delta:  # Only send full text if no delta was sent
```
longcw: no need to send final agent transcript if it's already sent as deltas.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1162 @@
...
+            name=event.tool_name,
+            arguments=json.dumps(event.parameters),
+        )
+
+        if self._current_generation is None:
+            self._start_new_generation()
+        self._current_generation.function_ch.send_nowait(function_call)
+
+        # Execute tool directly and send result to Ultravox immediately
+        asyncio.create_task(self._execute_tool_and_send_result(event))
```
longcw: tool execution is handled by the agent, no need to execute it inside the model.

maybe use `send_tool_result` in the `update_chat_ctx` and set `auto_tool_reply_generation` to True if the model will generate reply after `send_tool_result`.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1162 @@
...
+        except Exception as e:
+            logger.error(f"Error processing audio data: {e}")
+
+    def commit_audio(self) -> None:
+        """Commit the current audio input segment."""
+        pass
+
+    def clear_audio(self) -> None:
+        """Clear the audio buffer."""
+        self._bstream._buf.clear()
```
longcw, line 1231: clear the bstream buffer is not enough, this method is called to clear all buffered audio that not used for response generation. this is optional, left a warning is okay.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1162 @@
...
+        # Starting a new generation here since the output medium might be changed.
+        self._start_new_generation()
+
+        return fut
+
+    def interrupt(self) -> None:
+        """Interrupt the current generation."""
+        # Ultravox doesn't have a specific interrupt message, but we can clear the buffer
+        logger.debug("Interrupted current generation")
+        self.clear_audio()
```
longcw: `interrupt` is for interrupting the agent, is there a way to interrupt the agent?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1162 @@
...
+                self._current_generation.message_ch.send_nowait(
+                    llm.MessageGeneration(
+                        message_id=self._current_generation.response_id,
+                        text_stream=self._current_generation.text_ch,
+                        audio_stream=self._current_generation.audio_ch,
+                    )
+                )
+
+        # Also clear local audio buffer
+        self.clear_audio()
```
longcw, line 1181: instead of clearing the user input buffer, emitting a user speaking started event will interrupt the agent audio playout.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Added  diff tracking using the llm.utils.compute_chat_ctx_diff() utility, similar to how OpenAI plugin handles it f95c2e37a0a227342d8c2c32421595789ecdf130

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Added restart logic to update_tools(), matches the approach used in Google's realtime plugin
d65aab2785b7f0d23756c6cf9d13c42fac0d2a9c

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: yes, streaming PCM frames over the websocket is sufficient to trigger a response.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Added in [2f990b802314c06d2686636680837b3567c3e147](https://github.com/livekit/agents/pull/2992/commits/2f990b802314c06d2686636680837b3567c3e147)
Cancels any previous pending generation when new generate_reply called and uses timestamps to ensure correlation with correct generation....this  guarantees the response corresponds to the latest generate_reply call.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Removed last agent transcript send, now we stream deltas -> close on final
no final full text resend
44a1c5a125c752ee35c2ada5f5e3bbc6e5bd375f

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: I spent some time trying to get this to work inside the agent, tool calls were unreliable (mostly timeouts)

Ultravox expects an immediate client_tool_result Websocket message keyed by the original invocationId; deferring via update_chat_ctx causes the timeout ; https://docs.ultravox.ai/apps/datamessages#clienttoolresult

Tried:
- update_chat_ctx + auto_tool_reply_generation: no function_call_output arrives before Ultravox “speaking”; no client_tool_result sent → timeouts.
- function_tools_executed bridge (plugin- and agent-side): event doesn’t fire in time → same timeout.

Not like OpenAI/Gemini

- They accept tool results via their native realtime streams; Ultravox requires an immediate client_tool_result (invocationId) or it times out.

What worked: 

Execute tool in plugin and send result immediately 
or send from inside the tool (with RunContext). The latter breaks cross-agent tool  defintion onsistency.

I need help on this one... left as is for now, tools works reliably but there's a deviation from handling executing tools inside the agent

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Updated clear_audio to clear all local input buffers and added a warning. Ultravox doesn’t expose a client-side input buffer clear
3cc65c469b48aee5832d06eb54a81c20c6263b4e

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Ultravox handles interrupts via server-side VAD, when the user starts speaking over the agent, the server detects it and sends PlaybackClearBuffer.... we already handle that and drop unplayed audio, so the agent stops speaking and interrupt() does local cleanup; 

if programmatic (non‑voice) stop is needed, we can add a text barge-in (user_text_message with urgency: immediate ). But I'm afraid this might be treated as real user input.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Ultravox handles interrupts via server-side VAD, when the user starts speaking over the agent, the server detects it and sends PlaybackClearBuffer.... we already handle that and drop unplayed audio, so the agent stops speaking and interrupt() does local cleanup; 

if programmatic (non‑voice) stop is needed, we can add a text barge-in (user_text_message with urgency: immediate )

longcw: yeah I meant a programmatic interruption when user calls `session.interrupt()`.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: tried replacing `clear_audio` with `self.emit("input_speech_started", llm.InputSpeechStartedEvent())` as suggested, but this broke consecutive function calls during testing

Ultravox docs define `PlaybackClearBuffer` as a server message to clear buffered output audio, It’s not a user-interruption signal : https://docs.ultravox.ai/apps/datamessages#playbackclearbuffer

restored `self.clear_audio()` for now

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: now session.interrupt() sends an input_text_message with urgency: "immediate" and deferResponse: true,  VAD interruptions remain unchanged

call session.interrupt() to stop, if need to iterrupt with real text, call interrupt() then append a user ChatMessage to session_handle.chat_ctx

eb12cd26dccbce568fe01c978f75bf44ad348946

moadel321:
examples/agent.py demo : https://www.loom.com/share/5d4bc3d5f4b540fc829fd5246af985e8?sid=ba415eb7-cd77-4b2b-a5a3-5423f4bbfaba

longcw:
@moadel321 thanks! I'll take a look asap.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1343 @@
...
+
+        # Process new items only (Ultravox doesn't support deletions)
+        for _, msg_id in diff_ops.to_create:
+            item = chat_ctx.get_by_id(msg_id)
+            if not item:
+                continue
+
+            if item.role in ("system", "developer"):
+                # System and developer messages become deferred instructions
+                content_text = self._safe_text_content(item)
```
longcw: check the `item.type` before accessing the .type and .role, there are three types: `message`, `function_call`, and `function_call_output`. After checking the .type the `_safe_text_content` is not needed.
```suggestion
            if item.type == "message" and item.role in ("system", "developer"):
                text = item.text_content
```

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1343 @@
...
+        # Update remote context to track what we've sent
+        for previous_msg_id, msg_id in diff_ops.to_create:
+            item = chat_ctx.get_by_id(msg_id)
+            if (
+                item
+                and item.role in ("system", "developer", "user")
+                and self._safe_text_content(item)
+            ):
+                # Use previous_msg_id from diff operations for correct ordering
+                self._remote_chat_ctx.insert(previous_msg_id, item)
```
longcw: `_remote_chat_ctx` needs to be synced with the chat history on the server side, so it also should be updated when a new agent or user message received, like that in [gemini live plugin](https://github.com/livekit/agents/blob/livekit-agents@1.2.6/livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/realtime/realtime_api.py#L924-L937).

since the ultravox doesn't support inserting but only appending, it can use `chat_ctx` directly just like that in gemini live.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1343 @@
...
+    async def update_tools(self, tools: list[llm.FunctionTool | llm.RawFunctionTool]) -> None:
+        """Update the available tools."""
+        # Get current and new tool names for comparison
+        current_tool_names = set(self._tools.function_tools.keys())
+        new_tool_names = {
+            tool.name for tool in tools if hasattr(tool, "name") and tool.name is not None
+        }
+
+        # Always update the tools
+        self._tools.update_tools(tools)
```
longcw:
```suggestion
        current_tool_names = set(self._tools.function_tools.keys())

        # Always update the tools
        self._tools.update_tools(tools)
        new_tool_names = set(self._tools.function_tools.keys())
```

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1343 @@
...
+    def push_audio(self, frame: rtc.AudioFrame) -> None:
+        """Push audio frames to the session for transcription by Ultravox."""
+        if self._closed or not self._ws:
+            return
+
+        for resampled_frame in self._resample_audio(frame):
+            for audio_frame in self._bstream.write(resampled_frame.data.tobytes()):
+                if self._ws and not self._ws.closed:
+                    # Audio data is sent directly to the WebSocket
+                    asyncio.create_task(self._ws.send_bytes(audio_frame.data.tobytes()))
```
longcw: audio bytes should be added to the `_msg_ch` and sent via the `send_task` as well.

`asyncio.create_task(self._ws.send_bytes(audio_frame.data.tobytes()))` will create a lot of untracked tasks and cannot guarantee the order of the audio bytes sent to the server

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1343 @@
...
+            if self._ws and not self._ws.closed:
+                # Use text barge-in with immediate urgency to interrupt
+                # deferResponse=true prevents Ultravox from generating a response
+                interrupt_msg = {
+                    "type": "input_text_message",  # Correct type per events.py
+                    "text": "(barge-in)",  # Clear indication this is an interruption
+                    "urgency": "immediate",
+                    "deferResponse": True,  # Don't generate a response, just stop
+                }
+                self._send_client_event(interrupt_msg)  # Direct call, already non-blocking
```
longcw: should we add `urgency` to the `InputTextMessageEvent`?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1343 @@
...
+                        )
+                    )
+
+            elif item.role == "user":
+                # User messages sent normally
+                content_text = self._safe_text_content(item)
+                if content_text:
+                    logger.debug(f"[ultravox] Injecting user message: {content_text[:100]}...")
+                    self._send_client_event(
+                        InputTextMessageEvent(text=content_text, defer_response=False)
```
longcw: I assume `deferResponse=True` is to not generate a response, should it be True here? we don't want a response after `update_chat_ctx`.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1343 @@
...
+                )
+
+                # Cancel remaining tasks
+                for task in pending:
+                    await utils.aio.cancel_and_wait(task)
+
+                # Close current WebSocket
+                if self._ws:
+                    await self._ws.close()
+                    self._ws = None
```
longcw:
```suggestion
                try:
                    # Wait for any task to complete
                    done, _ = await asyncio.wait(
                        [send_task, recv_task, restart_wait_task], return_when=asyncio.FIRST_COMPLETED
                    )

                    for task in done:
                        if task != restart_wait_task:
                            # propagate exception if any
                            task.result()
                finally:
                    # Close current WebSocket
                    if self._ws:
                        await self._ws.close()
                        self._ws = None
                    await utils.aio.cancel_and_wait(send_task, recv_task, restart_wait_task)
```

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1343 @@
...
+                        logger.info(f"<<< {data}")
+                    event = parse_ultravox_event(data)
+                    self._handle_ultravox_event(event)
+
+                except Exception as e:
+                    logger.error(f"Error handling message: {e}", exc_info=True)
+
+            elif msg.type == aiohttp.WSMsgType.BINARY:
+                self._handle_audio_data(msg.data)
+            elif msg.type == aiohttp.WSMsgType.CLOSED:
```
longcw: raise an error when connection closed unexpectedly? for example https://github.com/livekit/agents/blob/livekit-agents@1.2.6/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py#L705-L714

longcw (commented):
left a few more comments, overall I think it's almost there. I'll test with the tool execution and see if I can help with it.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Updated to check item.type == "message" before accessing .role and to use item.text_content directly. removed safe_text_content completely
eb2f7fe043cbeb2f9fbde723153571e73f34fa24

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: replaced s_remote_chat_ctx with append only self._chat_ctx (like gemini)

On events, we append:
- user final transcript in  _handle_transcript_even
- assistant output on completion in _mark_current_generation_done

35a1af2cf130e334eb1c2951ce6a139497230ded

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: removed per chunk tasks, enqueing audio bytes in msg_ch and using send_task now similar to OpenAI and Gemini realtime plugins 
dc430122df0e8ca8baa85721c23d8e2a0afaf703

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Added in 07a0bff90cce2cfbaa1995d5440d30a662e5853e

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1346 @@
...
+        # Could also emit as a custom event for advanced debugging
+        # self.emit("debug_message", event.message)
+
+    @utils.log_exceptions(logger=logger)
+    def _handle_audio_data(self, audio_data: bytes) -> None:
+        """Handle binary audio data from Ultravox."""
+        try:
+            # Check if we have a current generation before processing audio
+            if not self._current_generation or self._current_generation._done:
+                logger.debug("[ultravox] Received audio data but no current generation, ignoring")
```
longcw: I got this log during the testing and the agent response at the beginning got cut off, should it create a new generation when agent audio received?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -0,0 +1,1346 @@
...
+        else:
+            logger.warning(f"Unhandled Ultravox event: {event}")
+
+    @utils.log_exceptions(logger=logger)
+    def _handle_transcript_event(self, event: TranscriptEvent) -> None:
+        """Handle transcript events from Ultravox."""
+        # Gate by ordinal to avoid late/stale events
+        if event.ordinal < self._last_seen_ordinal:
+            logger.debug(
+                f"Skipping stale TranscriptEvent with ordinal {event.ordinal} <= {self._last_seen_ordinal}"
```
longcw: got this log as well, like `Skipping stale TranscriptEvent with ordinal 1 <= 2`, the user transcripts got ignored, and it worked after commented it out. I guess the user and agent transcription has different ordinals?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
longcw: I think I found the issue: agent framework needs to mark the generation done before return the function outputs to the API (via `update_chat_ctx()`).

It seems the ultravox model is treating the tool call as a single generation? we can mark the generation done after sent it to the function_ch
```suggestion
        self._mark_current_generation_done()
```

then send the tool outputs to the api in `update_chat_ctx`
```python
            elif item.type == "function_call_output":
                tool_result = ClientToolResultEvent(
                    invocationId=item.call_id,
                    agent_reaction="speaks",
                )
                if item.is_error:
                    tool_result.error_message = item.error_message
                else:
                    tool_result.result = item.output

                self._send_client_event(tool_result)
```

longcw (commented):
I tested the plugin and the tool call issue can be fixed as mentioned, and found that the agent transcription is not stable? usually I lost the first part of it, and it's slow than audio?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: Done ,  tools now execute in the agent now, matches gemini realtime plugin
68ff8723f528b37f2e351cad3d61ba431ad5f45d

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: addressed in 080b071f3012bf17e09755d9d07de789de0f6ea6

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
moadel321: fixed in a2307714739899e054af87b5d98098946e8e093d

moadel321:
if im correct this leaves us with :

- transcripts ordering issue, text transcripts not stable 
- chained functions calls not working, might be an issue with the model 

will look into these more in a few hours

moadel321:
Moved from one global ordinal to per-role ordinals in 5803c5627143f36d2d9fde3e2e6185c992262783 , that fixed the transcripts ordering issue, tested on LiveKit Playground

moadel321:
- Transcripts ordering issue is fixed
- Model can perform chained function calls but really needs to be pushed to do so, either via system prompt or during user input 

Only remaining issue i can see is incorrect TTFT metrics during non tool use turns

longcw:
@moadel321 thanks for the updates, I tested it with the example and it overall works well!

I think we can merge this one and I'd like to do some cleaning in a separate pr, for some issues I can see
- the text input has duplicated user transcript
- Ultravox API has a rough audio/transcription sync internally, we need some changes in our synchronizer to make it work better when the text is slow as the audio.

moadel321:
Sounds good. Also, in addition to the two issues you listed, there's a third one related to TTFT being near zero in non-function call turns... I'll open another PR to continue working on these issues.

longcw (approved):

PR #2996, opened 2025-07-24
wait on_exit before pause scheduling

longcw:


theomonnom (approved):
Thanks! Let's wait for it as well

theomonnom (approved):

PR #3001, opened 2025-07-24
chore(assemblyai): update termination message

dan-ince-aai:
Update to new termination message, docs here: https://www.assemblyai.com/docs/speech-to-text/universal-streaming#session-termination


theomonnom (approved):
thanks, lgtm

PR #3027, opened 2025-07-26
pin the hedra avatar sample rate to 16k

longcw:


longcw:
> this seems fine, but why is it necessary?

they want 16kHz audio, and if not specified, the audio sent to the avatar may have different sample rate when using `session.say` on audios with different sample rate.

PR #3029, opened 2025-07-27
feat: allow custom track name to be used with RoomIO

davidzhao:
also exports transcription & output types so custom agents could be built using them

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/__init__.py
+++ b/livekit-agents/livekit/agents/voice/room_io/__init__.py
@@ -1,4 +1,5 @@
 from ...types import ATTRIBUTE_PUBLISH_ON_BEHALF
+from ._output import _ParticipantAudioOutput, _ParticipantTranscriptionOutput
 from .room_io import (
     DEFAULT_ROOM_INPUT_OPTIONS,
     DEFAULT_ROOM_OUTPUT_OPTIONS,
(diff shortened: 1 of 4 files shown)


davidzhao:
I'm building a live translator example that would benefit from this

livekit-agents/livekit/agents/voice/room_io/__init__.py
```diff
@@ -16,6 +17,8 @@
     "RoomOutputOptions",
     "ATTRIBUTE_PUBLISH_ON_BEHALF",
     "TextInputEvent",
+    "_ParticipantTranscriptionOutput",
+    "_ParticipantAudioOutput",
```
longcw, line 25: should we remove the _ before exposing these? otherwise looks good to me.

longcw (approved):
looks good to me! something nit

livekit-agents/livekit/agents/voice/room_io/__init__.py
davidzhao: are we reasonably confident for others to build with it? if so I'll remove it

longcw: yeah... let's keep the name for now

theomonnom: agree, let's not expose them

theomonnom (approved):

livekit-agents/livekit/agents/voice/room_io/__init__.py
davidzhao: I have to export them in order to use the transcription sync with audio.

translation apps need it. though it's fine to include _ to indicate it might change.

longcw: agreed, let's merge this pr without changing the name.

PR #3032, opened 2025-07-27
Fix RuntimeError on scheduling speech and on mark_generation_done

jayeshp19:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1152,6 +1152,13 @@ async def _user_turn_completed_task(
             # is detected. So the previous execution should complete quickly.
             await old_task
 
+        if self._scheduling_paused:
```
longcw, line 1155: `_scheduling_paused` is already checked before creating the `_user_turn_completed_atask` in `on_end_of_turn`.

livekit-agents/livekit/agents/voice/speech_handle.py
```diff
@@ -207,7 +207,8 @@ def _mark_done(self) -> None:
         with contextlib.suppress(asyncio.InvalidStateError):
             # will raise InvalidStateError if the future is already done (interrupted)
             self._done_fut.set_result(None)
-            self._mark_generation_done()
+            if self._generations:
```
longcw, line 210: do you know in what case the `_mark_done()` is called while `_generations` is empty?

livekit-agents/livekit/agents/voice/agent_activity.py
jayeshp19: if scheduling pauses after the end‑of‑turn but before _user_turn_completed_task reaches _generate_reply(), the call will raise an exception

livekit-agents/livekit/agents/voice/speech_handle.py
jayeshp19: if preemptive generation is cancelled before the speech is actually scheduled

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: I don't think this solves the problem, the activity can be still closed during `await self._agent.on_user_turn_completed` below.

longcw: I am wondering if that an expected raise when agent activity closed after that `on_end_of_turn`, or we should add a check right before the `_generate_reply` called. cc @theomonnom

longcw (approved):

livekit-agents/livekit/agents/voice/agent_activity.py
theomonnom: Yeah we need to think about it, but this PR already solve parts of it

theomonnom: we should probably raise an error on close, but still allow it when draining?

theomonnom (approved):

PR #3033, opened 2025-07-28
add retry_on_chunk_sent to FallbackAdapter for allowing retry after chunk sent

longcw:
…hunk sent


theomonnom:
Isn't this going to mix 2 unrelated audio files?

longcw:
> Isn't this going to mix 2 unrelated audio files?

so it's not enabled by default, I think it's better than say nothing when the gemini LLM failed, also would like to see what is the chunk it sent in https://github.com/livekit/agents/issues/2488#issuecomment-3027029014

theomonnom (approved):

PR #3041, opened 2025-07-29
Updated to send out stt results for deepgram that has a minimum amoun…

meetakshay99:
…t of confidence to avoid transcriptions generated without user actually speaking. Happens for Hindi language.

### DIFF
--- a/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
+++ b/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
@@ -57,6 +57,7 @@ class STTOptions:
     speech_endpoint: NotGivenOr[str] = NOT_GIVEN
     profanity: NotGivenOr[speechsdk.enums.ProfanityOption] = NOT_GIVEN
     phrase_list: NotGivenOr[list[str] | None] = NOT_GIVEN
+    explicit_punctuation: bool = False
 
 
 class STT(stt.STT):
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -46,6 +46,8 @@
 
 BASE_URL = "https://api.deepgram.com/v1/listen"
 
+# Deepgram is sometimes sensitive, so we'll ignore results with low confidence
+_default_min_confidence = 0.4
```
davidzhao: what is the range of values that we are seeing? if we are going to set a default, it'd be great to set it to a comfortable level that it won't trigger any false negatives

davidzhao (commented):
thanks for the PR, just one comment

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
meetakshay99: In my tests I had noticed them to be between 0.1 and 0.35
So had set it to 0.4 to be on the safe side.

meetakshay99: Pls suggest next steps to be done for getting this change merge with main.

davidzhao: could you merge from main?

I would feel more comfortable merging when the default is NOT_GIVEN. While you've tested in Hindi and it works well, I'm not sure that we can assume it'll work across languages.

Perhaps include in the comments of your findings (0.4 works well for Hindi), but leaving the default alone

PR #3050, opened 2025-07-30
use tts as fallback when realtime model returns text response

longcw:


livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -26,11 +26,15 @@ class InputSpeechStoppedEvent:
...
 
+MessageType = Literal["text", "audio"]
+
+
 @dataclass
 class MessageGeneration:
     message_id: str
     text_stream: AsyncIterable[str]  # could be io.TimedString
     audio_stream: AsyncIterable[rtc.AudioFrame]
+    message_type: MessageType | Awaitable[MessageType]
```
theomonnom: can we name this, modalities
so it contains both `text` and `audio` when we have both

theomonnom: should we always make it an Awaitable?
so basically

: 
```suggestion
    modalities: asyncio.Future[list[MessageType]]
```

longcw: it seems the text is always available for realtime models? OAI realtime API only accepts `['text']` and `['audio', 'text']`. I think it's enough to use a single flag to indicate if the message contains audio or only text, otherwise if it's a list, the check becomes more complex, e.g.
```python
if "audio" not in modalities and "text" in modalities:
    # fallback to tts
```

theomonnom: I feel like it aligns more with https://github.com/livekit/agents/blob/3b9965981932666b52b262a54ec9955af11e5a09/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py#L183

longcw: updated

theomonnom (approved):

PR #3055, opened 2025-07-31
feat: support bithuman cloud avatar

CathyL0:


livekit-plugins/livekit-plugins-bithuman/livekit/plugins/bithuman/avatar.py
```diff
@@ -35,26 +58,90 @@ class AvatarSession:
     def __init__(
         self,
         *,
-        model_path: NotGivenOr[str | None] = NOT_GIVEN,
+        mode: NotGivenOr[Literal["local", "cloud_gpu", "cloud_cpu", "cloud"]] = "local",
```
longcw: what's the difference between `cloud_gpu`, `cloud_cpu` and `cloud`, could you add a docstring?

CathyL0: adjusted and added comments

PR #3066, opened 2025-08-01
Add mem0 example

parshvadaftari:
Added mem0 example with the latest livekit verion.

### DIFF
--- a/examples/voice_agents/mem0-livekit/README.md
+++ b/examples/voice_agents/mem0-livekit/README.md
@@ -1,0 +1,92 @@
+## [Mem0](https://mem0.ai/)
+
+This guide demonstrates how to create a memory-enabled voice assistant using LiveKit, Deepgram, OpenAI, and Mem0, focusing on creating an intelligent, context-aware travel planning agent.
+
+## Prerequisites
+
+Before you begin, make sure you have:
+
+1. Installed Livekit Agents SDK with voice dependencies of silero and deepgram:
+```bash
+pip install "livekit-agents[deepgram,openai,cartesia,silero,turn-detector]~=1.0" "livekit-plugins-noise-cancellation~=0.2" "python-dotenv"
+```
+
+2. Installed Mem0 SDK:
+```bash
+pip install mem0ai
+```
+
+3. Set up your API keys in a `.env` file:
+```sh
+LIVEKIT_URL=your_livekit_url
+LIVEKIT_API_KEY=your_livekit_api_key
+LIVEKIT_API_SECRET=your_livekit_api_secret
+DEEPGRAM_API_KEY=your_deepgram_api_key
+MEM0_API_KEY=your_mem0_api_key
+OPENAI_API_KEY=your_openai_api_key
+```
+
+## Key Features of This Implementation
+
+1. **Semantic Memory Retrieval**: Uses Mem0 to store and retrieve contextually relevant memories
+2. **Voice Interaction**: Leverages LiveKit for voice communication with proper turn detection
+3. **Intelligent Context Management**: Augments conversations with past interactions
+4. **Travel Planning Specialization**: Focused on creating a helpful travel guide assistant
+5. **Function Tools**: Modern tool definition for enhanced capabilities
+
+## Running the Example
+
+To run this example:
+
+1. Install all required dependencies
+2. Set up your `.env` file with the necessary API keys
+3. Ensure your microphone and audio setup are configured
+4. Run the script with Python 3.11 or newer and with the following command:
+```sh
+python mem0-livekit-voice-agent.py start
+```
+or to start your agent in console mode to run inside your terminal:
+
+```sh
+python mem0-livekit-voice-agent.py console
+```
+5. After the script starts, you can interact with the voice agent using [Livekit's Agent Platform](https://agents-playground.livekit.io/) and connect to the agent inorder to start conversations.
+
+## Best Practices for Voice Agents with Memory
+
+1. **Context Preservation**: Store enough context with each memory for effective retrieval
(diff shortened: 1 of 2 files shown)


parshvadaftari:
@theomonnom @bcherry Can you review this?

davidzhao:
this is awesome! we'll take a look

examples/voice_agents/mem0-livekit/README.md
```diff
@@ -0,0 +1,92 @@
...
+## Running the Example
+
+To run this example:
+
+1. Install all required dependencies
+2. Set up your `.env` file with the necessary API keys
+3. Ensure your microphone and audio setup are configured
+4. Run the script with Python 3.11 or newer and with the following command:
+```sh
+python mem0-livekit-voice-agent.py start
```
longcw, line 46: can you use the actual example file name?

examples/voice_agents/mem0-livekit/mem0_livekit_agent.py
```diff
@@ -0,0 +1,119 @@
...
+                    if paragraph:
+                        source = "mem0 Memories"
+                        if "from [" in paragraph:
+                            source = paragraph.split("from [")[1].split("]")[0]
+                            paragraph = paragraph.split("]")[1].strip()
+                        context_parts.append(f"Source: {source}\nContent: {paragraph}\n")
+                if context_parts:
+                    full_context = "\n\n".join(context_parts)
+                    logger.info(f"Injecting RAG context: {full_context}")
+                    turn_ctx.add_message(role="assistant", content=full_context)
```
longcw, line 84: instead of inject memory as an assistant message, should we inject it to the instructions or a system message as in the mem0 basic usage https://github.com/mem0ai/mem0/blob/v0.1.115/README.md#basic-usage?

examples/voice_agents/mem0-livekit/mem0_livekit_agent.py
```diff
@@ -0,0 +1,119 @@
...
+    async def on_enter(self):
+        self.session.generate_reply(
+            instructions="Briefly greet the user and offer your assistance."
+        )
+
+    async def on_user_turn_completed(self, turn_ctx: ChatContext, new_message: ChatMessage) -> None:
+        # Persist the user message in Mem0
+        try:
+            logger.info(f"Adding user message to Mem0: {new_message.text_content}")
+            add_result = await mem0_client.add(
```
longcw, line 55: in this example every user message (and only user message) is added to the memory. according to their [document](https://docs.mem0.ai/core-concepts/memory-operations/add#when-should-you-add-memory%3F), should we add memories using a tool call and add both user and agent messages?  
> When Should You Add Memory?
Add memory whenever your agent learns something useful:
A new user preference is shared
A decision or suggestion is made
A goal or task is completed
A new entity is introduced
A user gives feedback or clarification

examples/voice_agents/mem0-livekit/mem0_livekit_agent.py
```diff
@@ -0,0 +1,119 @@
...
+            logger.info(f"mem0_client.search returned: {search_results}")
+            if search_results and isinstance(search_results, list):
+                context_parts = []
+                for result in search_results:
+                    paragraph = result.get("memory") or result.get("text")
+                    if paragraph:
+                        source = "mem0 Memories"
+                        if "from [" in paragraph:
+                            source = paragraph.split("from [")[1].split("]")[0]
+                            paragraph = paragraph.split("]")[1].strip()
```
longcw, line 79: why this is needed?

examples/voice_agents/mem0-livekit/README.md
parshvadaftari: Sure will do that!

examples/voice_agents/mem0-livekit/mem0_livekit_agent.py
parshvadaftari: Well this needs to be removed as it was added for a debugger point to check whether or not the memory is retrieved,

examples/voice_agents/mem0-livekit/mem0_livekit_agent.py
parshvadaftari: No not every user message is added to memory. Every user message goes into the check for mem0, and we have a default fact extraction to check whether the message is a useful memory or not. Inorder to make it more memory efficient in terms of the personal memories, every user message needs to be added to the mem0. Definetly we can add user and agent messages, but this example give a general idea of how mem0 <> livekit can be integrated and used to build memory enhanced voicebot.

Again this is just a gist of what this integration can help other build a memory enhanced voice assisstant.

examples/voice_agents/mem0-livekit/mem0_livekit_agent.py
parshvadaftari: Yes doable, but I wasn't sure how to handle the same.

examples/voice_agents/mem0-livekit/mem0_livekit_agent.py
longcw: I see, can you point me the document how mem0 check the user messages?

examples/voice_agents/mem0-livekit/mem0_livekit_agent.py
longcw: any better way to decode source and paragraph from the result of mem0? or can we just put it in the chat ctx as is?

parshvadaftari:
@longcw Hey can we add a new example. As we are planning to bump up the version of mem0 with many major changes which will break the current integration example. So I'll be making changes once we release the bumped up version of mem0 to 1.0. Thanks!

PR #3067, opened 2025-08-02
Fixing chunk size bug on inworld plugin-- matches hume and neuphonic

thavidu:
Someone at inworld sent me this bug fix after seeing it somewhere- posting it so it helps everyone :)


theomonnom (approved):

theomonnom:
Thanks

Issue #2981, opened 2025-07-23
TTS Optimization to generate speech lazily


anlagbr:
Currently, all speech is generated sequentially. However, if the agent is interrupted mid-speech, some audio will not be played, and a significant cost will still be incurred (TTS is typically the most expensive component among STT, LLM, and TTS).

I propose generating speech in a more efficient manner. For instance, the initial speech segment could be generated immediately. Subsequently, there would be a 1.5-second delay (depending on the provider) before generating the next chunk of audio (1.5 delay before current speech ends) coinciding with the completion of the current segment. This approach aims to minimize wasted computational resources and expenditure.

I couldn't find an official API solution for this problem.

Thanks.

longcw:
this feature is included in agents 1.2.3, here is an example https://github.com/livekit/agents/blob/livekit-agents@1.2.3/examples/voice_agents/tts_text_pacing.py

shivamkataria2000:
@longcw Does it work with Elevenlabs TTS?

longcw:
> [@longcw](https://github.com/longcw) Does it work with Elevenlabs TTS?

for 11labs it needs a `tts.StreamAdapter` to make it work in non-streaming mode. 11labs also has the `auto_mode` that accepts full sentences in the stream, we can support it with streaming laterly.

PR #3074, opened 2025-08-04
add tts stream pacer to generate speech lazily

longcw:
address https://github.com/livekit/agents/issues/2981 and https://linear.app/livekit/issue/AGT-1149/ttsstreamadapter-reduce-nb-requests-and-improve-speech-quality

### DIFF
--- a/livekit-agents/livekit/agents/tts/stream_adapter.py
+++ b/livekit-agents/livekit/agents/tts/stream_adapter.py
@@ -6,6 +6,7 @@ from typing import Any
 
 from .. import tokenize, utils
 from ..types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, APIConnectOptions, NotGivenOr
+from .stream_pacer import SentenceStreamPacer
 from .tts import (
     TTS,
     AudioEmitter,
@@ -27,6 +28,7 @@ class StreamAdapter(TTS):
         *,
         tts: TTS,
         sentence_tokenizer: NotGivenOr[tokenize.SentenceTokenizer] = NOT_GIVEN,
+        stream_pacer: NotGivenOr[SentenceStreamPacer | None] = NOT_GIVEN,
     ) -> None:
         super().__init__(
             capabilities=TTSCapabilities(streaming=True, aligned_transcript=True),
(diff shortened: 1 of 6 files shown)


livekit-agents/livekit/agents/tts/stream_adapter.py
```diff
@@ -27,6 +28,7 @@ def __init__(
         *,
         tts: TTS,
         sentence_tokenizer: NotGivenOr[tokenize.SentenceTokenizer] = NOT_GIVEN,
+        stream_pacer: NotGivenOr[SentenceStreamPacer | None] = NOT_GIVEN,
```
theomonnom: This is awesome! 🔥  Do you think we can simplify the API?
Something like just `pacing=True`? So users don't even have to think about it.

I also think this should be enabled by default

longcw: sounds good, but any ideas of the params `min_remaining_audio` and `max_text_length`, I think it's better to allow user to customize them.

I didn't enable it by default bc the transcription audio sync is worse in the case, and `use_tts_aligned_transcript` is not enabled by default right now, don't want user to see this as a "regression".

theomonnom: Is it also impacting the synchronization for the StreamAdapter, or only with use_tts_aligned_transcript?

longcw: also for StreamAdapter since the `use_tts_aligned_transcript` is not enabled for it by default, but I think we can enable that for StreamAdapter since it has no any side effect.

theomonnom: Makes sense, should we allow the pacing parameter to be `bool | SentenceStreamPacer`. So it is easier to use for users. (Just have to make it to True) 

Also maybe we should name it `text_pacing`?

theomonnom: Also mb `SentenceStreamPacer` -> `TextStreamPacer` ?

theomonnom:
Otherwise lgtm!

livekit-agents/livekit/agents/tts/stream_adapter.py
longcw: sounds good

longcw: updated

theomonnom (approved):

PR #3079, opened 2025-08-05
google stt: add enable_word_time_offsets option

longcw:


davidzhao (approved):
lg

what's the disadvantage of leaving these as enabled?

longcw:
> what's the disadvantage of leaving these as enabled?

it doesn't work for some languages https://live-kit.slack.com/archives/C066ZQ0T8NL/p1753985892421379

PR #3080, opened 2025-08-05
fix cartesia error when input text is empty

longcw:
cartesia tts returns an error `{'type': 'error', 'context_id': '49c376fe1e6a', 'status_code': 400, 'done': True, 'error': 'transcript empty or contains only punctuation: first transcript for WS is empty or contains only punctuation'}` when input text is empty.

ignore the error message if `done` is True.

fix https://github.com/livekit/agents/issues/3084


davidzhao:
why would we push an empty string through the TTS? is it related to the TTS pacing change?

longcw:
> why would we push an empty string through the TTS? is it related to the TTS pacing change?

no, it's from LLM when it's a tool call without message. before we ignored this error message. I exposed the cartesia error in https://github.com/livekit/agents/pull/3028 but it actually should check `done` first to ignore such cases.

PR #3091, opened 2025-08-06
fix: check if scheduling_paused after on_user_turn_completed callback

longcw:
the scheduling may fail after `on_user_turn_completed` callback. this pr adds check after the `on_user_turn_completed` callback, and there is no async calls between the check and `_generate_reply`.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1182,6 +1182,13 @@ class AgentActivity(RecognitionHooks):
             if self._rt_session is not None:
                 self._rt_session.interrupt()
 
+        if self._scheduling_paused:
+            logger.warning(
+                "skipping on_user_turn_completed, speech scheduling is paused",
+                extra={"user_input": info.new_transcript},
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1189,6 +1182,13 @@ async def _user_turn_completed_task(
             if self._rt_session is not None:
                 self._rt_session.interrupt()
 
+        if self._scheduling_paused:
```
longcw, line 1185: I am wondering if we should remove the check here, this will skip the `on_user_turn_completed` callback if agent is closing, and users may use this callback to save the user message, then they will miss the last user message.

and even we check `_scheduling_paused` here, seems like that calling `generate_reply` inside `on_user_turn_completed` still has the chance to raise a runtime error if agent is closing after that since the callback is async?

longcw: cc @theomonnom wdyt?

theomonnom (approved):

PR #3097, opened 2025-08-06
add the sdk version to the worker http endpoint

real-danm:
add the sdk version used by the agent the /worker http endpoint to allow versioning checks

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -358,7 +358,7 @@ class Worker(utils.EventEmitter[EventTypes]):
                     "agent_name": self._opts.agent_name,
                     "worker_type": agent.JobType.Name(self._opts.worker_type.value),
                     "active_jobs": len(self.active_jobs),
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/worker.py
```diff
@@ -358,6 +358,7 @@ async def worker(_: Any) -> web.Response:
                     "agent_name": self._opts.agent_name,
                     "worker_type": agent.JobType.Name(self._opts.worker_type.value),
                     "active_jobs": len(self.active_jobs),
+                    "version": __version__,
```
davidzhao: should we also include if it's python or node? or do you already have that info?

davidzhao (approved):

livekit-agents/livekit/agents/worker.py
real-danm: will add it

longcw:
this reminder me that we should add the agent and plugin versions to the logs.

PR #3107, opened 2025-08-07
feat: add reasoning effort option to OpenAI LLM configuration

mateuszkulpa:
Added [reasoning effort](https://platform.openai.com/docs/guides/latest-model#minimal-reasoning-effort) option to the OpenAI plugin. This is crucial because the default value is `medium`, which can lead to increased latency compared to non-thinking models.

According to [OpenAI's migration guide](https://platform.openai.com/docs/guides/latest-model?lang=python&reasoning-effort-mode=chat&verbosity-mode=chat#migrating-from-other-models-to-gpt-5), when migrating from GPT-4.1 to GPT-5, it is recommended to use the `low` or `minimal` reasoning effort.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -49,6 +49,7 @@ from .models import (
     DeepSeekChatModels,
     OctoChatModels,
     PerplexityChatModels,
+    ReasoningEffort,
     TelnyxChatModels,
     TogetherChatModels,
     XAIChatModels,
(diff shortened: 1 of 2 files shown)


yuyuma:
@mateuszkulpa thanks for this. While you're at it, could you also please add `verbosity` as an option?

https://platform.openai.com/docs/guides/latest-model?lang=python&reasoning-effort-mode=chat&verbosity-mode=chat#verbosity

If you aren't able to I would be happy to submit a separate PR myself.

davidzhao:
I have a PR here: #3108

davidzhao:
closing this since the alternative is merged.

PR #3108, opened 2025-08-07
expose reasoning_effort and other new params.

davidzhao:
also added compatibility with gpt-oss across providers

### DIFF
--- a/livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/llm.py
+++ b/livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/llm.py
@@ -4,6 +4,7 @@ import os
 
 import httpx
 import openai
+from openai.types import ReasoningEffort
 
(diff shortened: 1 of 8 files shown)


yuyuma:
@davidzhao thanks, can't wait to test out gpt-5 for voice use. Can you please add the [`verbosity` parameter](https://platform.openai.com/docs/guides/latest-model?lang=python&reasoning-effort-mode=chat&verbosity-mode=chat#verbosity) as well?  For our use case we'd like to also explore more succinct responses via this new configuration.

davidzhao:
> @davidzhao thanks, can't wait to test out gpt-5 for voice use. Can you please add the [`verbosity` parameter](https://platform.openai.com/docs/guides/latest-model?lang=python&reasoning-effort-mode=chat&verbosity-mode=chat#verbosity) as well? For our use case we'd like to also explore more succinct responses via this new configuration.

will do!

longcw (approved):

PR #3109, opened 2025-08-08
add resume_false_interruption callback to AgentSession

longcw:


longcw:
@theomonnom can you review this one?

theomonnom (commented):
this lg, I'm just not sure about the scenario where a SpeechHandle created using say/generate_reply should be "restarted".

theomonnom:
Just saw one of those examples in dz's [PR](https://github.com/livekit/agents/pull/3125):
```
        await self.session.say("connecting you to the customer now.")
        await self.session_manager.merge_calls()
```

longcw:
> this lg, I'm just not sure about the scenario where a SpeechHandle created using say/generate_reply should be "restarted".

we can add the `source` back to the event, and skip retry for source `say` in the default callback.

livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -26,24 +26,31 @@ def __init__(
...
         self._lock = asyncio.Lock()
-        self._audio_source = rtc.AudioSource(sample_rate, num_channels, queue_size_ms)
+        self._audio_source = rtc.AudioSource(sample_rate, num_channels, queue_size_ms=200)
         self._publish_options = track_publish_options
         self._publication: rtc.LocalTrackPublication | None = None
         self._subscribed_fut = asyncio.Future[None]()
 
+        self._audio_buf = utils.aio.Chan[rtc.AudioFrame]()
+        self._audio_bstream = utils.audio.AudioByteStream(
+            sample_rate, num_channels, samples_per_channel=sample_rate // 20
```
theomonnom, line 41: The tricky thing if splitting the frames in smaller size, is that as soon as the asyncio loop is slower or the cpu usage is high, the audio may be stuttering.

You should be able to see that when using the `stress` cmd on mac

theomonnom: I remember having a lot of issues around that when initially developing agents. This is why we can push faster than realtime from the Python side (from this [PR](https://github.com/livekit/rust-sdks/pull/425))

longcw: but we still have a reasonable buffer (200ms) for the AudioSource, the chunking here is just to ensure a single frame is not too big that we paused too late.

theomonnom: I see, so the interruptions has a min of 200ms latency here? Would be good to check if 200ms is enough. The only real reason we need big buffers is because users may write slow code on the event loop

longcw: we can use 100ms or 200ms, but I am wondering, what will happen if we call `audio_source.clear_queue()` when `await audio_source.capture(frame)` didn't return, how can we get the queued audio back from the `audio_source`?

if we can get the audio back, then the queue size and frame size don't matter, I can implement the pause in a different way when `pause` is called.

theomonnom: There is no easy way to get the audio back. But it's a complete OK assumption to assume that the playout was realtime. So 2s passed = 2s to discard

longcw: right now we will have a max of `frame_size` latency to pause the audio, and will drop max to 200ms (audio_source queue size) of audio when pause.

theomonnom: maybe 200ms is completely fine in both naturalness and cpu stress

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2132,6 +2156,53 @@ def _tool_execution_completed_cb(out: ToolExecutionOutput) -> None:
...
+                return
+
+            resumed = False
+            if self._session.options.resume_false_interruption and self._session.output.audio:
+                self._session.output.audio.resume()
+                resumed = True
+                logger.debug("resumed false interrupted speech", extra={"timeout": timeout})
+
+            self._session.emit(
+                "agent_false_interruption", AgentFalseInterruptionEvent(resumed=resumed)
```
longcw, line 2176: keep this event or replace it with a callback?

theomonnom: We should deprecate it for now, to avoid breaking changes, logging a warning is fine for now

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -156,7 +148,8 @@ def __init__(
         max_tool_steps: int = 3,
         video_sampler: NotGivenOr[_VideoSampler | None] = NOT_GIVEN,
         user_away_timeout: float | None = 15.0,
-        agent_false_interruption_timeout: float | None = 4.0,
+        agent_false_interruption_timeout: float | None = 2.0,
```
```suggestion
        false_interruption_timeout: float | None = 2.0,
```
I think it's fine to call it this way, `user_false_interruption_timeout` wouldn't make a ton of sense anyway

longcw: this will be a breaking change, should we change a name and log deprecate warning if user set `agent_false_interruption_timeout`?

PR #3117, opened 2025-08-09
feat: cgroups v1 support, lower idle processes

davidzhao:
the way we get CPU count isn't accurate. on cgroupsv1 instances, the quota is often unset, leading it to still use the host CPU count.

here we'll put a upper cap on concurrent processes spun up as well as idle processes, to reduce the likelihood that it'll try to use the entire machine

### DIFF
--- a/livekit-agents/livekit/agents/ipc/proc_pool.py
+++ b/livekit-agents/livekit/agents/ipc/proc_pool.py
@@ -22,7 +22,7 @@ EventTypes = Literal[
     "process_job_launched",
 ]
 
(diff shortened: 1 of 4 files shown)


real-danm (approved):

theomonnom (approved):
lgtm! tho using the whole machine is generally best practice here

(to have the max capacity)

davidzhao:
> lgtm! tho using the whole machine is generally best practice here
> 
> (to have the max capacity)

I agree.. when cgroups is working correctly, it still allows for it. (i.e. max worker count remains unchanged). 

I only changed the pieces where it relates to concurrent spinup. because the cost of getting it wrong is quite high (and these numbers are wrong often, when running cgroupsv1)

Issue #3133, opened 2025-08-11
Bug: StreamAdapter (text_pacing =True) + ElevenLabs causes tools to loop & agent to go silent (works without text_pacing)


affablebloke:
**Package/version**
	•	livekit-agents==1.2.5
	•	ElevenLabs plugin from livekit.plugins.elevenlabs.
	•	Observed only when wrapping ElevenLabs TTS with livekit.agents.tts.StreamAdapter. https://github.com/livekit/agents/pull/3074

**What’s happening**

**When the agent invokes any function/tool:**
	1.	Spoken output stops (agent goes quiet), then
	2.	The agent repeatedly re-invokes the same tool call in a loop when it hears new audio from the user.

If I remove text_pacing=True from StreamAdapter or use ElevenLabs TTS directly, it works as usual.

How to reproduce:

```
from livekit.agents.tts.stream_adapter import StreamAdapter
from livekit.plugins import elevenlabs

class MyAgent(Agent):
    def __init__(self) -> None:
        super().__init__(
            tts=StreamAdapter(
                tts=elevenlabs.TTS(
                    model=XXXXXX,
                    voice_id=XXXXXX,
                    voice_settings=elevenlabs.VoiceSettings(
                        stability=XXXXXX, similarity_boost=XXXXXX
                    ),
                    api_key=XXXXXX,
                ),
                text_pacing=True,
      ),
   tools=[...]
)
```

longcw:
I cannot reproduce the issue, can you share the logs and a reproducible example? and what is the output of `pip list | grep livekit`?

affablebloke:
@longcw I'll try to come up with a minimal repro. I have a PR that fixes it for my large dynamic tool calls, but it's too large and bespoke to submit as a repro right now.

https://github.com/livekit/agents/pull/3140

longcw:
~can you share the steps to reproduce your issue? I'd like to understand it better.~

ok I found the reason, it's when the text is empty `""` the `_wakeup_event` is never set since there is no sentences pushed, and then the stream never closed. The empty text usually happens when LLM returns a tool call.

created a fix https://github.com/livekit/agents/pull/3146.

affablebloke:
Thanks @longcw great work!

PR #3140, opened 2025-08-12
fix: enhance StreamPacerWrapper

affablebloke:
Fixes: https://github.com/livekit/agents/issues/3133

### Always wake the sender on flush() / end_input()

**Problem:** flush() only forwards to the underlying SentenceStream and doesn't wake the pacer loop. end_input() only sets the wake event if the audio emitter’s channel is already closed. That means a tool boundary + pacing can leave buffered sentences that never get considered for sending. 

**Fix:**
- In flush(), after delegating, call self._wakeup_event.set()
- In end_input(), always call self._wakeup_event.set() (regardless of emitter channel state)

### Note: 
I have not tested this fix with Cartesia TTS


livekit-agents/livekit/agents/tts/stream_pacer.py
```diff
@@ -115,49 +119,56 @@ async def _send_task(self) -> None:
...
+                remaining_audio = max(0.0, audio_start_time + audio_duration - curr_time)
+
+            should_send = first_sentence or (
                 generation_stopped and remaining_audio <= self._options.min_remaining_audio
-            ):
+            )
+
+            # If input ended, drain regardless of remaining_audio
+            if self._input_ended and self._sentences:
+                should_send = True
```
longcw: > 3) Drain any remaining sentences when input ends
Problem: The send loop only emits when remaining_audio <= min_remaining_audio or first_sentence. If input has ended and we still have _sentences, we should ignore min_remaining_audio and drain them immediately to avoid deadlock at the end of a turn.
Fix: If _input_ended is True, send whatever remains right away (bounded by max_text_length) without waiting for remaining_audio.


`input_end` only means the LLM generation is done, we shouldn't send all texts to TTS after that, otherwise the pacing is not actually working.

livekit-agents/livekit/agents/tts/stream_pacer.py
```diff
@@ -115,49 +119,56 @@ async def _send_task(self) -> None:
...
-            if audio_duration > 0.0 and audio_start_time == 0.0:
-                audio_start_time = curr_time
 
-            # check if audio generation stopped
-            if curr_time - prev_check_time >= 0.1:
-                if prev_audio_duration < audio_duration:
+            # Detect a new generation window and reset window start
+            if audio_duration > prev_audio_duration:
+                if not generation_started or generation_stopped:
+                    audio_start_time = curr_time  # Reset window
```
longcw: > 1) Reset the pacing window when synthesis restarts
Problem: audio_start_time is set only once (when pushed_duration()>0) and never reset for subsequent utterances/turns. Since pushed_duration() is cumulative, remaining_audio = audio_start_time + audio_duration - now can become very large/incorrect on later turns so the pacer keeps waiting and never sends the next text batch, hence the audio stall.

we assume the audio playout starts after `audio_start_time` and in a realtime speed, in what case `start_time + duration - now` (or `duration - (now - start_time)`) becomes very large (larger than actual unplayed audio duration), like the playout speed is faster than audio speed?

longcw: I expect `duration - (now - start_time)` is slightly smaller than the actual duration of unplayed audio, since there might be delays to audio playout starts, but it should be fine to have it smaller than actual.

livekit-agents/livekit/agents/tts/stream_pacer.py
affablebloke: Thanks @longcw. I reverted this change.

livekit-agents/livekit/agents/tts/stream_pacer.py
affablebloke: I think this is fine too after testing.

affablebloke:
@longcw this PR may be redundant now since your fix https://github.com/livekit/agents/pull/3146 landed in main.

davidzhao:
thanks for the report @affablebloke and appreciate the attempt to fix the issue 🙏

affablebloke:
> thanks for the report @affablebloke and appreciate the attempt to fix the issue 🙏

Thanks! Happy I could help get this resolved.

PR #3141, opened 2025-08-12
feat(aws-polly): add support for ssml text type

choso:
[AWS Polly](https://docs.aws.amazon.com/polly/latest/dg/API_SynthesizeSpeech.html#polly-SynthesizeSpeech-request-TextType) supports `ssml` text type, adds `text_type` to TTSOptions with `text` as default.


longcw (approved):
for next: need to check/update the sentence tokenizer to not break ssml tags into multiple chunks.

PR #3155, opened 2025-08-14
fix deadlock when calling _update_activiy in on_enter of a new agent

longcw:
fix the deadlock when calling AgentTask in the `on_enter` of an AgentTask.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -812,7 +812,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         wait_on_enter: bool = True,
     ) -> None:
         async with self._activity_lock:
-            # _update_activity is called directy sometimes, update for redundancy
+            # _update_activity is called directly sometimes, update for redundancy
             self._agent = agent
 
             if new_activity == "start":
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -849,9 +849,10 @@ async def _update_activity(
             elif new_activity == "resume":
                 await self._activity.resume()
 
-            if wait_on_enter:
-                assert self._activity._on_enter_task is not None
-                await asyncio.shield(self._activity._on_enter_task)
+        # move it outside the lock to allow calling _update_activity in on_enter of a new agent
```
longcw, line 853: @theomonnom do you see any risk of releasing the lock before `on_enter` done? right now except entering an AgentTask, only `session.update_agent` has `wait_on_enter = True`.

theomonnom (approved):
Added more context in Slack.
This lgtm! Nice catch man!

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -849,9 +849,10 @@ async def _update_activity(
...
+        if wait_on_enter:
+            assert self._activity._on_enter_task is not None
+            await asyncio.shield(self._activity._on_enter_task)
 
     @utils.log_exceptions(logger=logger)
     async def _update_activity_task(self, task: Agent) -> None:
```
```suggestion
    async def _update_activity_task(self, old_task: asyncio.Task | None, task: Agent) -> None:
```

PR #3157, opened 2025-08-14
fix mistralai llm and stt

longcw:
the original mistralai plugin doesn't work, and doesn't support tool calls.


davidzhao (approved):
lgtm!

Issue #2033, opened 2025-04-17
Proposal: adding more information in metrics


ChenghaoMou:
Inspired by the conversation in #1999 and my personal experience.

Currently there are certain metrics included in Metric objects, but some useful information can still be added to enrich them for better utility, for example:

- `parent_speech_id`: for better conversation lineage tracking, from speech_handle's parent ref.
- model_name and vendor/provide_name: particularly useful when used with a fallback adapter for better observability over models or third-party services. 

What do you think?

jlavasyou:
+1 for model and provider/system

System is a required attribute, and model is a recommended attribute for [OpenTelemetrys conventions for gen-ai metrics](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/
)

Naive implementation I did that worked:
- add model field in `LLMMetrics`
- set model field in `LLMSteam` on `_metrics_monitor_task`
- track the current model being used in `FallbackAdapter` with an index and then return that index as part of a `model` field

ChenghaoMou:
Adding relevant tickets here:

- #3165 
- #3272

PR #3165, opened 2025-08-15
Add model and provider in metrics for all services

ChenghaoMou:
This PR attempts to add model and service provider names into STT/LLM/TTS/VAD/EOU metrics. This should address https://github.com/livekit/agents/issues/2033

- used `provider` instead of `system` to accommodate all services instead of just LLM;
- some of them uses base_url domain name as the default provider name instead of fixed values;
- in case of unclear value, `unknown` is used;

Tested with a simple cascading and an OpenAI realtime agent:

- [x] Observing STT metrics with metadata
- [x] Observing TTS metrics with metadata
- [x] Observing LLM metrics with metadata
- [x] Observing Realtime metrics with metadata (w separate turn_detection or server-side semantic-vad)


davidzhao:
this is awesome work @ChenghaoMou !

ChenghaoMou:
Hey @longcw @theomonnom would love to hear your thoughts on this.

vipulnandan-ng:
@ChenghaoMou Thanks for this! Trying to integrate langfuse. this will be great!

longcw:
I'll fix the conflicts to #3439 in a separate pr.

PR #3167, opened 2025-08-15
feat: Add OpenRouter plugin for LiveKit Agents

Hormold:
This PR adds a new OpenRouter plugin for LiveKit Agents, providing access to LLM through OpenRouter's routing platform with advanced provider routing and cost optimization features.

## Why not just use the OpenAI plugin?
OpenRouter is OpenAI-compatible but requires a dedicated plugin for OpenRouter-specific features:
1. **Provider routing API**: Custom headers for routing across providers by cost/latency
2. **Web search integration**: `:online` model suffix triggers different API endpoints
3. **Analytics headers**: OpenRouter-specific tracking requires custom request formatting
4. **Provider preferences**: Additional JSON payloads not supported by standard OpenAI client

These features need custom request handling that would clutter the OpenAI plugin.

This is my first PR for LiveKit - feedback welcome! 

Another idea: it can be part of the OpenAI package like a method -> with_openrouter, with typed params specified for openrouter like for xAI

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Literal, TypedDict, Union
+from typing import Literal, Optional, TypedDict, Union
 
 from openai.types import AudioModel
 
@@ -260,7 +260,7 @@ class OpenRouterWebPlugin:
     """OpenRouter web search plugin configuration"""
 
     max_results: int = 5
-    search_prompt: str | None = None
+    search_prompt: Optional[str] = None
     id: str = "web"
 
 


univers3:
any update on this?
sorting Openrouter's providers by latency makes it a very good option, even with an additional hop.
I have tested it and it's extremely good

livekit-plugins/livekit-plugins-openrouter/livekit/plugins/openrouter/llm.py
```diff
@@ -0,0 +1,214 @@
...
+        # Create OpenAI client instance with custom headers
+        client = oai_client.AsyncOpenAI(
+            api_key=resolved_api_key,
+            base_url="https://openrouter.ai/api/v1",
+            default_headers=headers,
+            timeout=timeout,
+        )
+
+        # Create LiveKit OpenAI LLM wrapper with the configured client
+        self._client = openai.LLM(
```
longcw: can you add the support as `openai.LLM.with_openrouter(...)`?

angelsocastro:
Between the two options, I think the openai.LLM.with_openrouter(...) approach that @longcw  suggested makes more sense, as it keeps the API consistent with the existing OpenAI client and avoids having to use a separate plugin. That would fully cover my use case.

livekit-plugins/livekit-plugins-openrouter/livekit/plugins/openrouter/llm.py
Hormold: Sure, will do soon.

sifrgh:
any update on this?

Hormold:
> any update on this?

I'm back to coding, so it will be finished soon

Hormold:
The OpenRouter implementation uses an **inner class** pattern instead of direct LLM instantiation because it's the only provider that requires injecting specific parameters (like provider preferences, web plugins, fallback models, and analytics headers) into every chat API call via extra_body and extra_headers, necessitating a custom chat method override that other providers don't need.
Is it looks good? Or anyone has other ideas?

livekit-plugins/livekit-plugins-openrouter/livekit/plugins/openrouter/llm.py
Hormold: @longcw can you check it one more time?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -346,6 +380,125 @@ def with_x_ai(
...
+                merged: dict[str, Any] = {}
+                if is_given(extra_kwargs):
+                    merged.update(extra_kwargs)
+
+                # Add OpenRouter-specific body parameters
+                if self.__or_body:
+                    body = dict(self.__or_body)
+                    if "extra_body" in merged and isinstance(merged["extra_body"], dict):
+                        body.update(merged["extra_body"])
+                    merged["extra_body"] = body
```
longcw: I think we can expose `extra_body`, `extra_headers`, and `extra_query` to the base LLM class so you don't need to  create a OpenRouterLLM here.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -74,6 +74,40 @@ class _LLMOptions:
...
+    data_collection: Literal["allow", "deny"] | None = None
+    only: list[str] | None = None
+    ignore: list[str] | None = None
+    quantizations: list[str] | None = None
+    sort: Literal["price", "throughput", "latency"] | None = None
+    max_price: dict[str, float] | None = None
+
+    def to_dict(self) -> dict[str, Any]:
+        return {k: v for k, v in self.__dict__.items() if v is not None}
+
```
longcw: move them to `models.py`? and can we use a TypedDict instead?

Hormold: About TypedDict - OpenRouterWebPlugin can't be TypedDict because it needs default values (max_results=5, id="web") and TypedDict doesn't support defaults. Had to keep it as @dataclass. Any ideas?

Hormold: OpenRouterProviderPreferences could be TypedDict since all fields are optional

Hormold:
> lgtm! can you fix the ruff and type check before merging? see https://github.com/livekit/agents/blob/main/CONTRIBUTING.md#typechecking-linting-and-formatting for details.

Fixed, but there's still one type error in the branch. Same as in the main branch currently.

PR #3169, opened 2025-08-16
fix: skip done speech when scheduling

longcw:
if a speech was interrupted when it's in the queue, it will be marked as done before scheduling. when it's scheduled, `await speech._wait_for_generation()` in `_scheduling_task` will never return


theomonnom (approved):
nice catch!

PR #3176, opened 2025-08-18
speechmatics stt: fix speech data start and end time and clean up

longcw:


livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -757,10 +722,8 @@ def _get_speaker_fragments_from_fragment_group(
         if not group:
             return None
 
-        min_frag_time = min(frag.start_time for frag in group)
-        ts = (self._start_time + datetime.timedelta(seconds=min_frag_time)).isoformat(
```
longcw, line 761: start and end timestamps of `SpeechData` should be float instead of a str

PR #3177, opened 2025-08-18
add `cancel_tool_reply` to function_tools_executed event

longcw:
if there are parallel tool calls, the tool executions won't know each other. this is useful when ppl want to skip the tool response when another tool returns a new Agent.

### DIFF
--- a/livekit-agents/livekit/agents/voice/events.py
+++ b/livekit-agents/livekit/agents/voice/events.py
@@ -152,9 +152,25 @@ class FunctionToolsExecutedEvent(BaseModel):
     function_calls: list[FunctionCall]
     function_call_outputs: list[FunctionCallOutput | None]
     created_at: float = Field(default_factory=time.time)
+    _reply_required: bool = Field(exclude=True)
+    _handoff_required: bool = Field(exclude=True)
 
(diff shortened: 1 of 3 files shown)


theomonnom:
It makes sense to support that. But I'm not sure about the API. Ideally the event listeners are "read-only".

longcw:
yeah I agree it's "weird" to raise `StopResponse` from an event callback... any suggestion on the api, should we add a function executed node?

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -152,10 +152,26 @@ class FunctionToolsExecutedEvent(BaseModel):
     function_calls: list[FunctionCall]
     function_call_outputs: list[FunctionCallOutput | None]
     created_at: float = Field(default_factory=time.time)
+    _reply_required: bool = PrivateAttr(default=False)
+    _handoff_required: bool = PrivateAttr(default=False)
 
     def zipped(self) -> list[tuple[FunctionCall, FunctionCallOutput | None]]:
         return list(zip(self.function_calls, self.function_call_outputs))
 
+    def prevent_tool_reply(self) -> None:
```
theomonnom: Maybe nit, but actually I also like `cancel_tool_reply` (following the Python cancel of asyncio.Task)

PR #3178, opened 2025-08-18
Use heartbeat param for websocket connection

wlbksy:
Use a heartbeat parameter to the websocket connection between the worker and the LiveKit server so that the worker can detect earlier whether the websocket has become unresponsive.

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -58,6 +58,7 @@ from .version import __version__
 ASSIGNMENT_TIMEOUT = 7.5
 UPDATE_STATUS_INTERVAL = 2.5
 UPDATE_LOAD_INTERVAL = 0.5
+HEARTBEAT_INTERVAL = 30
 
 
 def _default_initialize_process_fnc(proc: JobProcess) -> Any:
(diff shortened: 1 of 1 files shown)


davidzhao:
I'm not sure that this is needed. the worker is periodically sending load information to the server. the channel should have plenty of activity.

wlbksy:
> I'm not sure that this is needed. the worker is periodically sending load information to the server. the channel should have plenty of activity.

Thanks for the feedback! You’re right that under normal conditions the worker is already sending periodic `update_worker_status`, so the channel has regular activity. That does provide a good indication of liveness in most cases.
One caveat is that the server doesn’t actually reply to `update_worker_status` — it just processes them internally. This means the worker has no confirmation that the messages actually reached the server. If the connection becomes half-open, or if the worker stalls due to a bug, the kernel may keep the socket “alive” for many minutes before reporting an error. Similarly, NAT timeouts or silent drops at the network layer may not be detected quickly.
Adding a lightweight heartbeat with pong responses gives us an explicit acknowledgment from the server side. It acts as a fail-fast mechanism: even if the channel usually has activity, the heartbeat provides an extra safeguard that helps us detect unresponsive connections earlier and recover faster in edge cases.

davidzhao:
got it! thanks for the explanation. that makes sense

PR #3186, opened 2025-08-18
primary speaker detection with STT diarization

longcw:
1. detect and switch primary speaker based on audio RMS level
2. support formatting primary and background speakers' transcripts separately
3. or suppress the background speaker from STT

### DIFF
--- a/livekit-agents/livekit/agents/stt/multi_speaker_adapter.py
+++ b/livekit-agents/livekit/agents/stt/multi_speaker_adapter.py
@@ -26,6 +26,23 @@ class MultiSpeakerAdapter(STT):
         primary_format: str = "{text}",
         background_format: str = "{text}",
     ):
+        """MultiSpeakerAdapter is an adapter that allows to detect and suppress background speakers.
+        It needs STT with diarization capability and works for a single audio track.
+
+        Args:
+            stt (STT): STT instance to wrap
+            detect_primary_speaker (bool, optional): Whether to detect primary speaker. Defaults to True.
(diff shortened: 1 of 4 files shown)


theomonnom:
added some feedbacks on Slack

livekit-agents/livekit/agents/stt/multi_speaker_adapter.py
```diff
@@ -0,0 +1,346 @@
...
+from livekit import rtc
+
+from .. import utils
+from ..log import logger
+from ..types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, APIConnectOptions, NotGivenOr
+from ..utils.audio import AudioByteStream
+from .stt import STT, RecognizeStream, SpeechData, SpeechEvent, SpeechEventType
+
+
+class MultiSpeakerAdapter(STT):
```
theomonnom, line 18: I think we should call this DiarizationAdapter

theomonnom (approved):

livekit-agents/livekit/agents/stt/multi_speaker_adapter.py
longcw: I was about to use `DiarizationAdapter`, but it sounds like we are going to add diarization in this adapter, like `StreamAdapter` adds streaming support for non-streaming stt.

longcw: I am going to merge this, we can change the name before next release if you have a different opinion.

PR #3199, opened 2025-08-19
feat: new CLI & new AgentServer API

theomonnom:


examples/frontdesk/frontdesk_agent.py
```diff
@@ -152,7 +151,11 @@ async def list_available_slots(
         return "\n".join(lines) or "No slots available at the moment."
 
 
-async def entrypoint(ctx: JobContext):
+server = AgentServer()
+
+
+@server.realtime_session()
```
longcw, line 158: what are the `realtime_session` and `rtc_session`?

theomonnom: This is going to be the new way of specifying the job entrypoint

PR #3203, opened 2025-08-20
fix(google): Instant Voice Cloning Key Parameter is incorrect

gnomefin:
## Overview
According to the official cloud_tts the passing parameter should be `voice_cloning_key` not `voice_clone_key`. I had tested my custom livekit plugin it and works properly.

```
class VoiceCloneParams(proto.Message):
    r"""The configuration of Voice Clone feature.

    Attributes:
        voice_cloning_key (str):
            Required. Created by GenerateVoiceCloningKey.
    """

    voice_cloning_key: str = proto.Field(
        proto.STRING,
        number=1,
    )
```

I believe the related issue is here: https://github.com/livekit/agents/issues/3013
@davidzhao


davidzhao:
thank you for the fix. this is on me thinking I had the right parameter and was over-reliant on type checking vs testing out the path.

PR #3215, opened 2025-08-21
feat: enable keyterm prompting in deepgram for de, nl, sv, da

danielgrittner:
Deepgram released new monolingual models (see [here](https://developers.deepgram.com/changelog/speech-to-text-changelog#nova-3-model-update)) that all support keyterm prompting.


davidzhao (approved):
lgtm

giandiego:
Hi, Spanish support has already been added, now es and es-419 should be added as well.

Regards.

PR #3218, opened 2025-08-21
livekit-agents 1.3

theomonnom:


livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -123,7 +125,15 @@ def _setup_cloud_tracer(
     headers = (("Authorization", f"Bearer {access_token.to_jwt()}"),)
     metadata = {"room_id": room_id, "job_id": job_id}
 
-    tracer_provider = TracerProvider()
+    resource = Resource.create(
```
paulwe, line 130: this is adding it to the resource or the scope...? resource is host, scope is record, entry is... whatever

PR #3219, opened 2025-08-21
CLI: Check if microphone is receiving input through RMS

kushagrob:
Ran into an issue during local Agent setup where Terminal didn't have mic permissions but Livekit wasn't able to detect this. `sounddevice` in Python still creates an input stream even when the permissions are off, so we check the RMS to see if any input audio is coming in after 5 seconds and log an error if that's not the case. 

- For some reason, even when the permission is off, RMS goes slightly above 0 when the TTS is talking (but never above 1.0). 
- From local testing, when the mic is working and the user is not speaking, background noise is always >5.0 and usually above >10.0. Hence, I set MIN_RMS to 2.0

### DIFF
--- a/livekit-agents/livekit/agents/voice/chat_cli.py
+++ b/livekit-agents/livekit/agents/voice/chat_cli.py
@@ -28,6 +28,8 @@ MAX_AUDIO_BAR = 30
 INPUT_DB_MIN = -70.0
 INPUT_DB_MAX = 0.0
 FPS = 16
+MIC_CHECK_DELAY = 5.0
+MIN_RMS = 2.0
 
 
 AEC_RING_BUFFER_SIZE = 24000 * 4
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/chat_cli.py
```diff
@@ -28,6 +28,8 @@
 INPUT_DB_MIN = -70.0
 INPUT_DB_MAX = 0.0
 FPS = 16
+MIC_CHECK_DELAY = 5.0
+MIN_RMS = 2.0
```
Shubhrakanti, line 32: better name? So we know this so for detecting audio.

theomonnom:
This is nice! It should be straightforward to add on 1.3 by creating a new column [here](https://github.com/livekit/agents/blob/c378bb455abcc1f334152b38aaa50c1628a8f9d2/livekit-agents/livekit/agents/cli/cli.py#L613C40-L613C46)

PR #3224, opened 2025-08-22
✨ Add `start_tine`, `end_time`, and `speaker_id` to 11labs `SpeechData`

dvschuyl:
ElevenLabs exposes word-level timings for the speech-to-text api: https://elevenlabs.io/docs/api-reference/speech-to-text/convert .

This PR uses these timings to add additional info to the generated `SpeechData`.


theomonnom:
lgtm! thanks

PR #2636, opened 2025-06-18
docs: add development instructions to README and example READMEs

bcherry:
## Summary
- Expand the Contributing section in `README.md` with development setup, testing, formatting, and documentation instructions
- Add links to the `examples/` directory and [`python-agents-examples`](https://github.com/livekit-examples/python-agents-examples) repo in the Examples section
- Add brief READMEs for example subdirectories that were missing them (`primitives/`, `other/`, `drive-thru/`, `dtmf/`, `frontdesk/`, `survey/`)

## Test plan
- [x] `make check` passes (format, lint, type-check)
- [ ] Verify rendered markdown looks correct on GitHub

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3225, opened 2025-08-22
add readme for examples

longcw:


bcherry (approved):
this is good but needs instructions to run them. i have a pretty stale PR that includes instructions for that which I haven't gotten back to - maybe you could incorporate the important parts here and we'll discard the rest of mine? https://github.com/livekit/agents/pull/2636

(I do have a minor concern that including lists like this in the README adds a maintenance burden since they may become outdated, but I don't feel strongly about it)

PR #3229, opened 2025-08-22
feat: support interruptions via interim transcripts for AssemblyAI plugin

dan-ince-aai:
Hi team! This PR adds support for interruptions via the AssemblyAI plugin using interim transcripts. It mimics the VAD inference done event when STT start of speech is received. This is to support AssemblyAI's plugin if VAD is not included in the agent pipeline, since we have now integrated VAD into the turn detection model and therefore don't want to run it twice.

Please feel free to adjust the PR as needed and happy to address any comments! :)

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -45,9 +45,7 @@ class _TurnDetector(Protocol):
     async def unlikely_threshold(self, language: str | None) -> float | None: ...
     async def supports_language(self, language: str | None) -> bool: ...
 
-    async def predict_end_of_turn(
-        self, chat_ctx: llm.ChatContext, *, timeout: float | None = None
-    ) -> float: ...
+    async def predict_end_of_turn(self, chat_ctx: llm.ChatContext) -> float: ...
 
 
 class RecognitionHooks(Protocol):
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -296,12 +296,40 @@ async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
...
+                self._hooks.on_start_of_speech(start_vad_event)
+                # Also trigger inference done event for interruption logic
+                inference_vad_event = vad.VADEvent(
+                    type=vad.VADEventType.INFERENCE_DONE,
+                    samples_index=0,
+                    timestamp=0.0,
+                    speech_duration=1.0,  # Set to 1.0 to pass min_interruption_duration check
+                    silence_duration=0.0,
+                )
+                self._hooks.on_vad_inference_done(inference_vad_event)
```
theomonnom, line 319: I'm not sure hacking the VAD event is a good way to handle this.

We could maybe create a new method inside our Hooks? If we're 100% sure AssemblyAI detects an interruption, creating a new method like `self._hooks.on_interrupt` is maybe fine?

Or maybe `on_start_of_speech` could accept a stt.SpeechEvent argument

theomonnom (commented):
I'm curious to know why there is a need to stop using the local VAD. It should be more reliable since we don't need to wait for any transcript. Interruptions may be handled better and the arguments inside AgentSession will be respected (e.g the `min_interruption_duration`)

livekit-agents/livekit/agents/voice/audio_recognition.py
dan-ince-aai: @theomonnom in production environments for our customers with noisy audio, the VAD is being constantly triggered leading to the agent being interrupted mid-speech, therefore we have received many requests to add interruptions via the first partial instead. It is also more consistent to handle interruptions via a transcript as you can be certain that a word has been transcribed and not just a backchannel or background noise. The partial transcript is returned in ~300ms after speech so it is also similar in speed to VAD.

dan-ince-aai: Also more than happy to go another route instead of hacking the VAD event, @theomonnom let me know what you would prefer

May you join our shared Slack channel for closer collaboration? Happy to share more details about why we made this PR too

theomonnom: Hey, we started looking into this internally. We should have a solution for this pretty soon

PR #3238, opened 2025-08-22
fix Speech interruption attempted when allow_interruption is False

### LINKED ISSUE #3230: Speech interruption attempted when allow_interruption is False
I'm getting this error message `This generation handle does not allow interruptions`

I'm yet to replicate it but i found that it's due to this code 
```py
if self._user_turn_completed_atask != asyncio.current_task():
    # If a new user turn has already started, interrupt this one since it's now outdated
    # (We still create the SpeechHandle and the generate_reply coroutine, otherwise we may
    # lose data like the beginning of a user speech).
    speech_handle.interrupt()
```
in `agent_activity.py`

There are no checks here on whether or not interruptions are allowed which results in a runtime error because of 
```py
if not self._allow_interruptions:
    raise RuntimeError("This generation handle does not allow interruptions")
```

AdvaitS:
fix #3230

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1276,7 +1276,7 @@ class AgentActivity(RecognitionHooks):
                 user_message=user_message, chat_ctx=temp_mutable_chat_ctx
             )
 
-        if self._user_turn_completed_atask != asyncio.current_task():
+        if self._user_turn_completed_atask != asyncio.current_task() and not speech_handle._allow_interruptions:
             # If a new user turn has already started, interrupt this one since it's now outdated
             # (We still create the SpeechHandle and the generate_reply coroutine, otherwise we may
             # lose data like the beginning of a user speech).


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1276,7 +1276,10 @@ async def _user_turn_completed_task(
                 user_message=user_message, chat_ctx=temp_mutable_chat_ctx
             )
 
-        if self._user_turn_completed_atask != asyncio.current_task():
+        if (
+            self._user_turn_completed_atask != asyncio.current_task()
+            and not speech_handle._allow_interruptions
```
longcw, line 1281: should we cancel the speech in this case? cc @theomonnom

davidzhao: IMO manual interruption should always be allowed..

`allow_interruption` should be referencing if a user is able to interrupt via voice

theomonnom: Another PR added `force=True` to the `interrupt` method.

theomonnom: Otherwise I think this PR makes sense, but shouldn't it be the opposite? We check if interruptions is allowed, not disallowed

PR #3239, opened 2025-08-22
cpu.py looks in more places for cgroupv2 cpu.max

michael-salient:


livekit-agents/livekit/agents/utils/hw/cpu.py
```diff
@@ -50,7 +50,10 @@ def cpu_count(self) -> float:
             return env_cpus
         quota, period = self._read_cpu_max()
         if quota == "max":
-            return psutil.cpu_count() or 1.0
+            try:
+                return len(psutil.Process().cpu_affinity())
```
michael-salient, line 54: https://psutil.readthedocs.io/en/latest/#psutil.cpu_count

PR #3241, opened 2025-08-24
chore(11labs): update plugin to use multi-context websocket API

davidzhao:


davidzhao:
> nice! Did you see any ttfb improvements?

i did a bunch of side by side tries today. it improves it by ~50ms on avg, but only if `auto_mode` is enabled (on both). otherwise it's actually slower than before, likely due to different internal chunking logic.

since their official docs suggests flushing manually, I've enabled auto_mode by default and flushing ourselves

```
When generating long responses, stream the text in smaller chunks and use the flush: true flag at the end of complete sentences. This improves the quality of the generated audio and improves responsiveness.
```

IlyaUmanets:
@davidzhao , A connection is dropped after updating the TTS config. How can I open a connection immediately after using the "update_options" function?

PR #3245, opened 2025-08-25
deepgram: support for Flux

davidzhao:


shivamkataria2000:
This looks cool, it would be great if this works out.

lasryaric:
hey @davidzhao, thank you so much for this PR, that's just what I was looking for.
What about VAD and turn detection in Livekit? Are they still needed when using Deepgram flux model?

PR #3247, opened 2025-08-25
chore: allow types-protobuf>=4

davidzhao:
previously it required <5, which isn't necessary

### DIFF
--- a/livekit-agents/pyproject.toml
+++ b/livekit-agents/pyproject.toml
@@ -28,12 +28,12 @@ classifiers = [
 dependencies = [
     "click~=8.1",
     "livekit>=1.0.12,<2",
-    "livekit-api>=1.0.4,<2",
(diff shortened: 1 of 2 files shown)


longcw (approved):

livekit-agents/pyproject.toml
```diff
@@ -93,6 +93,7 @@ speechmatics = ["livekit-plugins-speechmatics>=1.2.6"]
 spitch = ["livekit-plugins-spitch>=1.2.6"]
 tavus = ["livekit-plugins-tavus>=1.2.6"]
 turn-detector = ["livekit-plugins-turn-detector>=1.2.6"]
+upliftai = ["livekit-plugins-upliftai>=1.2.6"]
```
longcw, line 96: also needs to add to https://github.com/livekit/agents/blob/main/pyproject.toml#L42

Issue #3251, opened 2025-08-25
Bug Report: Groq Plugin LLM Initialization Fails with "safety_identifier" Parameter Error


asdkjasdfla:
@theomonnom @davidzhao 

## Summary
The `livekit-plugins-groq` package fails to initialize LLM instances due to passing an unsupported `safety_identifier` parameter to the parent `LLM` class constructor.

## Error
```
TypeError: LLM.__init__() got an unexpected keyword argument 'safety_identifier'
```

**Full Traceback:**
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/path/to/site-packages/livekit/plugins/groq/services.py", line 49, in __init__
    super().__init__(
TypeError: LLM.__init__() got an unexpected keyword argument 'safety_identifier'
```

## Environment
- **Python:** 3.12.3
- **OS:** Linux
- **livekit-agents:** 1.2.6
- **livekit-plugins-groq:** 1.2.6 (also occurs in 1.2.5)
- **livekit:** 1.0.12

## Minimal Reproduction Case
```python
from livekit.plugins import groq

# This fails with TypeError
llm = groq.LLM(api_key='test-key', model='llama-3.3-70b-versatile')
```

## Root Cause
The Groq plugin's `__init__` method passes a `safety_identifier` parameter to the parent `LLM` class via `super().__init__()`, but the base `LLM` class constructor doesn't accept this parameter.

**Analysis:**
- `groq.LLM.__init__()` signature includes `safety_identifier: NotGivenOr[str] = NOT_GIVEN`
- `livekit.agents.llm.LLM.__init__()` signature is `(self) -> None` (accepts no parameters)

## Impact
- **Severity:** Critical - Complete inability to instantiate Groq LLM
- **Affected Versions:** 1.2.5, 1.2.6 (likely other recent versions)
- **Workaround:** None available

## Expected Behavior
```python
from livekit.plugins import groq
llm = groq.LLM(api_key='my-api-key', model='llama-3.3-70b-versatile')
# Should successfully create LLM instance
```

## Suggested Fix
The Groq plugin should handle the `safety_identifier` parameter internally without passing it to the parent constructor:

```python
# In livekit/plugins/groq/services.py
def __init__(self, *, safety_identifier=NOT_GIVEN, **kwargs):
    # Handle safety_identifier internally
    self._safety_identifier = safety_identifier
    
    # Don't pass safety_identifier to parent class
[...]

PR #3256, opened 2025-08-26
fix groq plugin dependencies

longcw:
fix https://github.com/livekit/agents/issues/3251


theomonnom (approved):

davidzhao:
wow.. I can't believe this was still stuck like this.. nice catch

PR #3260, opened 2025-08-26
feat: update silero_vad model to v6

mateuszkulpa:
Updated silero model to newest version: https://github.com/snakers4/silero-vad/releases/tag/v6.0


theomonnom:
done in https://github.com/livekit/agents/pull/3275 because the file doesn't have a checksum

PR #3267, opened 2025-08-26
chore(silero): Update Silero VAD model to v6

imsakg:
Upgrade bundled silero_vad.onnx to v6 to pick up model improvements and fixes.


theomonnom:
done in https://github.com/livekit/agents/pull/3275 because the file doesn't have a checksum

PR #3269, opened 2025-08-27
add AudioOutput capabilities and disable pause/resume for if output not supported

longcw:


livekit-agents/livekit/agents/voice/io.py
```diff
@@ -225,6 +232,12 @@ def sample_rate(self) -> int | None:
         """The sample rate required by the audio sink, if None, any sample rate is accepted"""
         return self._sample_rate
 
+    @property
+    def supports_pause(self) -> bool:
```
theomonnom: For STT and TTS, we directly expose the capabilities dataclass:
https://github.com/livekit/agents/blob/2c84fc9e7cdb3878456e01a31d66fe442d01ad2c/livekit-agents/livekit/agents/tts/tts.py#L88

Let's do the same here

longcw: the issue is that the `pause` flag needs to check the capabilities from `next_in_chain`, then we need to create a new `TTSCapabilities` that actually not the same as the one passed in?

theomonnom: Ok that makes sense then, let's keep this. 

nit: we have can_subscribe, can_publish, so mb can_pause is better

PR #3274, opened 2025-08-27
fix OpenAI Realtime event race conditions

maurice2k:
Wait for session.created and session.updated events before sending other events like response.create as it might fail if the session has not been created or the session.updates (like instructions) have not been applied yet.

This fix waits for alle session.(created|updated) events and blocks further events except for valid session.update events.


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -677,6 +709,15 @@ async def _run_ws(self, ws_conn: aiohttp.ClientWebSocketResponse) -> None:
...
             nonlocal closing
             async for msg in self._msg_ch:
+                # do not send any events until the session has been created ("session.created" event)
+                # or pending session updates have been acknowledged ("session.updated" event)
+                # except for further "session.update" events
+                if isinstance(msg, SessionUpdateEvent):
+                    await self._session_updates_pending.inc()
+                else:
+                    await self._session_created_future
+                    await self._session_updates_pending.wait_until_zero()
```
longcw: my concern is this will block the agent silently if a single `SessionUpdateEvent` failed. should we have a retry and raise an error instead?

maurice2k: I think it's crucial to wait at least for session.created, as this caused a lot of errors lately with response.create being sent too early (generate_reply failed).
I think it's worse if OpenAI API is slower.

Not waiting for all session.updated is causing issues if response.create is coming too fast and i.e. tools are have not been updated.

We have experienced so many issues due to this, so I would rather implement some timeout if session.updated is blocking too long rather than not having this fix...

longcw: A timeout error sounds good.

maurice2k:
hope this is fine now.
should _ZeroNotifyCounter be moved to utils.py maybe?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -508,6 +517,48 @@ def process_base_url(
...
+        self._lock = asyncio.Lock()
+
+    async def inc(self) -> None:
+        async with self._lock:
+            self._count += 1
+            if self._count == 1:
+                self._zero_event.clear()
+
+    async def dec(self) -> None:
+        async with self._lock:
```
longcw: the lock is not needed, this can be a sync method.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -678,6 +732,21 @@ async def _run_ws(self, ws_conn: aiohttp.ClientWebSocketResponse) -> None:
...
+                    await self._session_updates_pending.inc()
+                else:
+                    await self._session_created_future
+                    timeout_occurred = not await self._session_updates_pending.wait_until_zero(
+                        timeout=self._realtime_model._opts.session_updates_waiting_timeout
+                    )
+                    if timeout_occurred:
+                        logger.warning(
+                            f"Waiting for session update acknowledgements (session.updated events) timed out after {self._realtime_model._opts.session_updates_waiting_timeout}s"
+                        )
```
longcw:
```suggestion
                    try:
                        await asyncio.wait_for(self._pending_session_updates.done(), timeout=5)
                    except asyncio.TimeoutError:
                        raise APIConnectionError(
                            message="OpenAI Realtime API session update timed out",
                        )
```

I am wondering should we actually block the sending task when update session is not done, especially for audio event `InputAudioBufferAppendEvent`? if the sending task is blocked the audio will be blocked and out of sync with realtime audio input.

Maybe it's better to make the `update_options` and `update_instructions` return a `Future` that set to done when update session counts is zero, and add a timeout for the future like that in `generate_reply`. It was [planed initially](https://github.com/livekit/agents/blob/livekit-agents@1.2.8/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py#L1085-L1086), but the `session.updated` doesn't have a metadata for id tracking, so it needs to use counts of events.

longcw: @maurice2k does this fix your issue https://github.com/livekit/agents/pull/3346?

maurice2k: @longcw not exactly. i would not add it there as it slows down updates to the session. and these updates are already sent way to often (i.e. 3 updates for each session).

sending session.update events **should not** block. maybe same for input audio. but everything that requires a certain session state (i.e. response.create) must be blocked unless all updates are through.

longcw: > 3 updates for each session

if you mean the session initialization, we can modify it to wait for the updates concurrently. I think we shouldn't block the sending task, but just the single operation and you can choose to wait it or not.

we already did that for `generate_reply` (response.create), adding that for the session.update should be able to fix your original issue
> response.create as it might fail if the session has not been created or the session.updates (like instructions) have not been applied yet.

please let me know if ^ is not true and if there is any remaining issues that awaiting the single operation cannot fix, we can iterate.

maurice2k: I've refactored the logic after taking a deep dive into OpenAI's Realtime API documentation and their Python Agents SDK.

New logic is: nothing is being blocked except for response.create, which is waiting for session.created and all pending session.updated events.
This makes sense as response.create is the only event truly dependent on session updates.

PR #3278, opened 2025-08-28
support interrupting agent by interim transcripts

longcw:
replace https://github.com/livekit/agents/pull/3229

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -63,6 +63,7 @@ class VoiceOptions:
     discard_audio_if_uninterruptible: bool
     min_interruption_duration: float
     min_interruption_words: int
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -143,6 +144,7 @@ def __init__(
         discard_audio_if_uninterruptible: bool = True,
         min_interruption_duration: float = 0.5,
         min_interruption_words: int = 0,
+        interrupt_by_interim_transcript: bool = True,
```
theomonnom: I think we should remove this parameter, and automatically enable it when you don't have VAD

longcw: oh I thought it was removed... now it's always enabled.

theomonnom (approved):

PR #3280, opened 2025-08-28
allow SpeechHandle to be interrupted when long-running function tools are pending

longcw:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -199,6 +199,9 @@ def __init__(self, agent: Agent, sess: AgentSession) -> None:
         self._on_enter_task: asyncio.Task | None = None
         self._on_exit_task: asyncio.Task | None = None
 
+        # speeches that audio playout finished but not done because of tool calls
+        self._background_speeches: set[SpeechHandle] = set()
```
theomonnom, line 203: not a fan of the name `_background_speeches`, but can't think of better lol

theomonnom (approved):
lgtm!

livekit-agents/livekit/agents/voice/agent_activity.py
davidzhao: `_pending_speech_handles` ?

theomonnom: _idle_speeches?

PR #3281, opened 2025-08-28
feat(Google): Add markup for TTS HD voices

itskyf:
## Summary
Adds support for [Google TTS pause control markup](https://cloud.google.com/text-to-speech/docs/chirp3-hd#pause_control) by extending the `tts_node`.

## Notes
I noticed that **when `use_tts_aligned_transcript=True` (default), the transcript also outputs the `[pause]` markup**, which might not be ideal for frontend display.  
Would you recommend flagging this in the code itself, or is it better documented in the integration guide?

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
@@ -52,6 +52,7 @@ class _TTSOptions:
     volume_gain_db: float
     custom_pronunciations: CustomPronunciations | None
     enable_ssml: bool
+    use_markup: bool
 
 
 class TTS(tts.TTS):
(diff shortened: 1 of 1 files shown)


davidzhao:
this looks great! could you share any tips on how one could produce these markups? we'd be glad to include the example in our docs

itskyf:
@davidzhao In Thai there are no natural spaces, so I first ask my LLM to insert them. Before sending to Chirp 3 HD, I replace spaces with `[pause short]` so words are spoken cleanly.

**Important**: I need to set `use_tts_aligned_transcript=False` in Agent.__init__(), otherwise the modified text (with tags) will leak into transcripts.

```python
class MyAgent(Agent):
    def __init__(self):
        super().__init__(use_tts_aligned_transcript=False)

    async def tts_node(
        self, text: AsyncIterable[str], model_settings: ModelSettings
    ) -> AsyncIterable[rtc.AudioFrame]:
        async def add_pauses(input_text: AsyncIterable[str]) -> AsyncIterable[str]:
            async for chunk in input_text:
                yield chunk.replace(" ", "[pause short]")  # ...add other rules if needed
        async for frame in Agent.default.tts_node(self, add_pauses(text), model_settings):
            yield frame
```

longcw (approved):

PR #3296, opened 2025-08-29
Tool class & EndCallTool

theomonnom:


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -562,6 +562,9 @@ async def drain(self) -> None:
 
         await self._activity.drain()
 
+    def close_soon(self) -> None:
+        self._close_soon(reason=CloseReason.USER_INITIATED, drain=True)
```
bcherry, line 721: for telephony agents, i think this would also need to remove the user participant so their phone hangs up on their end

also I think offering a public close & hangup method is a good idea on its own, so it might make sense to name this method `end_call` instead of `close_soon`, for the developer's benefit

davidzhao: it's not just for telephony clients.. SDK clients would also still stay in the room. I think end call is in a 1 on 1 scenario, and the idea is that it ends it for the other side as well. (i.e. delete_room)

theomonnom: This will be done with https://linear.app/livekit/issue/AGT-1935/add-delete-room-on-close-to-roomoptions instead

livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -51,14 +52,16 @@ def __init__(
...
         if type(self) is Agent:
             self._id = "default_agent"
         else:
             self._id = id or misc.camel_to_snake_case(type(self).__name__)
 
         self._instructions = instructions
-        self._tools = tools.copy() + find_function_tools(self)
+        self._tools = find_function_tools(self)
+        for tool in tools or []:
+            self._tools += tool.__livekit_tools__() if isinstance(tool, ToolSet) else [tool]
```
longcw: @theomonnom do you think if we need to keep `ToolSet` in the `self._tools` and expand it until calling the LLM?

theomonnom: That's a good question, I think we should keep it. Seems more "future-proof"

longcw: updated to keep the ToolSet in Agent, and unroll it when calling `ToolContext.update_tools` or `RealtimeSession.update_tools`.  https://github.com/livekit/agents/pull/3296/commits/15868b03bfda1c07cabb5d573dcec54c82a65d58
 
I am not sure if it's worth actually since it adds more extra checks and complexity, do you have some future use cases in mind?

theomonnom: One of the reason is that we could have some "special" tools. For e.g OpenAI has their own WebSearchTool
We coule enable this by doing a fake placeholder like:

class OpenAIWebSearch(Tool): ...

theomonnom:
Now that I'm thinking about it, maybe we should still name this `Tool` instead of `ToolSet`.

I feel like it's kind of weird to have the EndCallTool to be a "ToolSet"

marioplumbeer:
I'd love to use this instead of my own implementation. The doctoring defines the logic to end the call, but it can vary depending on the use case.

Would you be open to build the function tool at runtime and inject keywords into the function tool description, such that it could be used to determine if it should be called or not?

marioplumbeer:
In addition to that, it would be nice to be able to inject a callback, so that users can perform cleanup before ending the call.

longcw:
@theomonnom this one is ready for review

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -717,6 +717,9 @@ async def drain(self) -> None:
...
```
chenghao-mou, line 721: Should we take a reason parameter here instead? So we can use something like `AGENT_INITIATED` in the EndCallTool.

examples/voice_agents/session_close_callback.py
```diff
@@ -15,19 +25,26 @@
...
+            await self.session.current_speech
 
-        logger.info("Closing session from function tool")
-        await self.session.generate_reply(instructions="say goodbye to the user")
-
-        self.session.shutdown()
+        logger.info("generating goodbye message")
+        await self.session.generate_reply(
+            instructions="say goodbye to the user", tool_choice="none"
+        )
```
chenghao-mou, line 47: Maybe worth a comment here to mention why no tools are wanted here.

examples/voice_agents/session_close_callback.py
chenghao-mou, line 80: I don't think backslash is supported <=3.11

chenghao-mou (approved):
Overall lgtm. 

One small thing: could you help me understand the rationale behind ToolSet? I can see `EndCallTool` uses it, but I'm not sure what problem it's solving or when I'd reach for it considering `tools` already takes a list of things.

longcw:
> could you help me understand the rationale behind ToolSet? I can see `EndCallTool` uses it, but I'm not sure what problem it's solving or when I'd reach for it considering `tools` already takes a list of things.

from my understanding, we want to provide some built-in tools like `EndCallTool` and these tools may have a list of `FunctionTool` bound together.

PR #3301, opened 2025-08-30
feat: add environment variable to customize the logger name

efontan:
Allow applications to customize the logger name for better integration with their logging infrastructure. , instead of `livekit.agents`.

The default behavior stays the same if the environment variable isn't set.


efontan:
Please let me know if this approach makes sense or if there's a better way to do it. It would help us a lot to keep consistent names in our agents.

theomonnom:
I'm not sure to understand the need of this, is it to have different agents in the same logger?

PR #3305, opened 2025-09-01
use markdown and emoji filters for tts_node by default

longcw:


examples/voice_agents/basic_agent.py
```diff
@@ -49,10 +48,12 @@ async def on_enter(self):
...
-        # TTS node allows us to process the text before it's sent to the model
-        # here we'll strip out markdown
-        filtered_text = filter_markdown(text)
-        return super().tts_node(filtered_text, model_settings)
+        # Markdown and emoji are filtered before sending it to the TTS model by default,
+        # you can disable this via model_settings in tts_node
+
+        # model_settings.tts_filter_markdown = False
+        # model_settings.tts_filter_emoji = False
+        return super().tts_node(text, model_settings)
```
theomonnom, line 57: I don't think we need new fields inside ModelSettings.

Let's directly do it here, if the users don't want the default filter, they can just not call super()

longcw: sounds good. moved the filters to `Agent.tts_node`.

marctorsoc: Q: should these filters be applied to `transcription_node` by default too?  @theomonnom @longcw 

in my case, I had them in both, so now it means I keep them for `transcription_node` but remove them from `tts_node` 🤷‍♂️

longcw: @marctorsoc no, it's only applied to `tts_node` by default.

marctorsoc: @longcw yes, my question is what are your thoughts about "extending the default" to `transcription_node` too

longcw: it's not planned.

PR #3316, opened 2025-09-02
add tool call and output to session.history

longcw:


livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -131,7 +131,9 @@ async def update_instructions(self, instructions: str) -> None:
 
         await self._activity.update_instructions(instructions)
 
-    async def update_tools(self, tools: list[llm.FunctionTool | llm.RawFunctionTool]) -> None:
+    async def update_tools(
+        self, tools: list[llm.FunctionTool | llm.RawFunctionTool], *, filter_chat_ctx: bool = True
```
theomonnom: Do we need `filter_chat_ctx`? I feel like it's always necessary

longcw: I think it's better to give a choice to the user, also make the filtering more explicit. it might be useful if you just want to disable a tool temporally?

theomonnom: I think it's better for users to don't even have to think about it IMO (keeping it simple).

Maybe we can improve tool_choice for this purpose in the future.

longcw: I think it's fine to remove the option from `update_tools`, what do you think about having `exclude_invalid_function_calls` for `update_chat_ctx`?

theomonnom: I kind of feel like it's the same thing, I don't see when `exclude_invalid_function_calls` should be False (at least for 99% of the usecase, probably not worth adding it)

longcw: I still feel it's better to make it configurable and explicit, even it's required for most of use cases, we should let users know we are doing that for them, silently removing some items from the chat context is confusing IMO.

theomonnom: What happens if we set it to False and a request to the LLM is made? Is it going to fail?

longcw: setting it to False is to not filter the tool call/outputs in chat context, doesn't mean the LLM will call the invalid tools following the history, and if that happens it will log a warning for not founded tools.

PR #3326, opened 2025-09-02
feat(azure-stt): add configurable punctuation option to Azure STT

sarthakgoyal23:
This pull request adds support for configuring punctuation behavior in the Azure Speech-to-Text (STT) plugin. The main change is the introduction of a new `punctuation` option, allowing users to explicitly control how punctuation is handled during speech recognition.

**Azure STT punctuation option support:**

* Added a `punctuation` field to the `STTOptions` class, allowing users to specify punctuation handling.
* Updated the `STT` class constructor to accept the new `punctuation` parameter and documented its behavior in the docstring. [[1]](diffhunk://#diff-e385d376ee185b8e05288d55722a2672de5aae917b2d20f50a0bc86be4af9259R81) [[2]](diffhunk://#diff-e385d376ee185b8e05288d55722a2672de5aae917b2d20f50a0bc86be4af9259R97-R98)
* Passed the `punctuation` option through to the internal configuration in the `STT` class.
* Modified the `_create_speech_recognizer` function to set the Azure service property for punctuation when the option is explicitly set to `False`.

### DIFF
--- a/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
+++ b/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
@@ -57,7 +57,7 @@ class STTOptions:
     speech_endpoint: NotGivenOr[str] = NOT_GIVEN
     profanity: NotGivenOr[speechsdk.enums.ProfanityOption] = NOT_GIVEN
     phrase_list: NotGivenOr[list[str] | None] = NOT_GIVEN
-    punctuation: NotGivenOr[bool] = NOT_GIVEN
+    explicit_punctuation: bool
 
 
 class STT(stt.STT):
@@ -78,7 +78,7 @@ class STT(stt.STT):
         profanity: NotGivenOr[speechsdk.enums.ProfanityOption] = NOT_GIVEN,
         speech_endpoint: NotGivenOr[str] = NOT_GIVEN,
         phrase_list: NotGivenOr[list[str] | None] = NOT_GIVEN,
-        punctuation: NotGivenOr[bool] = NOT_GIVEN,
+        explicit_punctuation: bool = False,
     ):
         """
         Create a new instance of Azure STT.
@@ -94,8 +94,9 @@ class STT(stt.STT):
         Args:
             phrase_list: List of words or phrases to boost recognition accuracy.
                         Azure will give higher priority to these phrases during recognition.
-            punctuation: Controls punctuation behavior. If False, sets explicit punctuation mode.
-                        If True or NOT_GIVEN, uses Azure's default punctuation behavior.
+            explicit_punctuation: Controls punctuation behavior. If True, enables explicit punctuation mode
+                        where punctuation marks are added explicitly. If False (default), uses Azure's
+                        default punctuation behavior.
         """
 
         super().__init__(capabilities=stt.STTCapabilities(streaming=True, interim_results=True))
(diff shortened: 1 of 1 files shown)


sarthakgoyal23:
@theomonnom  @longcw Can you please review the changes?

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
```diff
@@ -77,6 +78,7 @@ def __init__(
         profanity: NotGivenOr[speechsdk.enums.ProfanityOption] = NOT_GIVEN,
         speech_endpoint: NotGivenOr[str] = NOT_GIVEN,
         phrase_list: NotGivenOr[list[str] | None] = NOT_GIVEN,
+        punctuation: NotGivenOr[str] = NOT_GIVEN,
```
longcw: use `Literal["explicit", "implicit"]` if the values are bounded?

longcw (commented):
left a small comment, otherwise looks good to me

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
sarthakgoyal23: @longcw I refered this comment for the fix 

https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/667#issuecomment-690840772

As this configuration is not documented in azure docs.

I tried with explicit and implicit previously but after your comment i check instead of implicit if i pass any string in punctuation it keeps punctuation enabled so now i have made this as bool and if it will be false then only i will pass punctuation as explicit where it disables the punctuation

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
```diff
@@ -367,6 +372,12 @@ def _create_speech_recognizer(
     if is_given(config.profanity):
         speech_config.set_profanity(config.profanity)
 
+    # Set punctuation behavior if specified
+    if is_given(config.punctuation) and config.punctuation is False:
```
longcw: nit:
```suggestion
    if config.punctuation is False:
```

and according to https://learn.microsoft.com/en-us/azure/ai-services/speech-service/display-text-format?pivots=programming-language-python#punctuation, set it to `explicit` means the STT will recognize the explicitly read punctuations but not add it automatically.

should we call this param `explicit_punctuation: bool` instead of `punctuation: bool`? the later one makes people think it's fully disable the punctuation which is not the case.

sarthakgoyal23: Thanks for the suggestion @longcw.  I have updated the changes please check now

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
```diff
@@ -57,6 +57,7 @@ class STTOptions:
     speech_endpoint: NotGivenOr[str] = NOT_GIVEN
     profanity: NotGivenOr[speechsdk.enums.ProfanityOption] = NOT_GIVEN
     phrase_list: NotGivenOr[list[str] | None] = NOT_GIVEN
+    explicit_punctuation: bool
```
longcw: can you fix the type check here?

longcw (approved):

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
sarthakgoyal23: Done

Issue #3297, opened 2025-08-30
Add image input support to the OpenAI gpt-realtime plugin


chensterman:
Would be really great to support image inputs into the gpt-realtime model, as described in the [OpenAI launch](https://openai.com/index/introducing-gpt-realtime/).

Would be happy to get started on this feature.

MatheusRDG:
Waiting for it, it will be amazing!

Nate711:
Me too

nischalj10:
Eagery waiting for this

tusheet-geoiq:
me too

davidzhao:
added in #3420, it'll be available in the next release

Shoshin23:
@davidzhao the always on video_enabled=true works but its too slow for non-image questions. i have a separate function call that gets triggered for image questions so i want to push video frames manually with push video but that wasn't working for me.

can you share a manual frame push example?

Shoshin23:
i figured out that its model.session.push_video() but my first question remains. is there a specific way to send frames via push_video() if i set video_enabled=False? so far whatever i've tried on that front hasnt worked.

when i start an agentsession with video_enabled=True then i could fetch the latest frame and send it via push_video() but setting to false isn't sending frames so the model hallucinates a response

longcw:
@Shoshin23 a solution for now is you can create a custom `_VideoSampler` https://github.com/livekit/agents/blob/livekit-agents@1.2.11/livekit-agents/livekit/agents/voice/agent_session.py#L110 similar to the `VoiceActivityVideoSampler`, it will be called for every video frame, and the frames are passed to the model only it returns true.

So for your use case, you can add a flag for this sampler, and set it when you want video input, and turn it off after the frame is pushed.

You can also turn the default `VoiceActivityVideoSampler` off with `speaking_fps=0` and `silent_fps=0` so no video frames are pushed automatically, but you can still grab a frame and call the `model.session.push_video()` by yourself.

Shoshin23:
hi @longcw thanks for the input. i'm going to give this custom sampler a shot but for the time being this configuration returns an error

```
 assistant = AgentSession(
        llm=openai.realtime.RealtimeModel(
            model="gpt-realtime",
            voice=config["voice"] if "voice" in config else "alloy",
            video_sampler = VoiceActivityVideoSampler(
                speaking_fps=0,
                silent_fps=0,
            )
        ))
```
i get this error in my console:

```
[...]

PR #3327, opened 2025-09-02
placeholder Realtime image input support (NEED UPDATE TO OPENAI PYTHO…

chensterman:
Placeholder code for Realtime image input support. The functionality of this code is dependent on an update to the OpenAI Python SDK, which also has not seemed to have updated to support image inputs for the Realtime model on their own API.

#3297

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1583,7 +1583,35 @@ def _livekit_item_to_openai_item(item: llm.ChatItem) -> ConversationItem:
                 )
 
             elif isinstance(c, llm.ImageContent):
-                continue  # not supported for now
+                # Serialize the image using LiveKit's utility
+                cache_key = "serialized_image"
+                if cache_key not in c._cache:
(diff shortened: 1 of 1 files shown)


davidzhao:
thanks for the PR @chensterman! which update are you referring to on their end? I can see `ConversationItemContent` being exported as of 2 months ago.

we should also update to the latest version of their SDK.

chensterman:
> thanks for the PR @chensterman! which update are you referring to on their end? I can see `ConversationItemContent` being exported as of 2 months ago.
> 
> we should also update to the latest version of their SDK.

Ah yes what I meant was I don't even think the OpenAI Python SDK supports the latest image input feature of the Realtime API. You can see that its lacking here in the [ConversationItemContent](https://github.com/openai/openai-python/blob/a52463c93215a09f9a142e25c975935523d15c10/src/openai/types/beta/realtime/conversation_item_content.py#L28)

I've submitted an [issue](https://github.com/openai/openai-python/issues/2600) on their repo and working on a PR for that. Does this seem to make sense to you?

davidzhao:
Got it. that makes sense! thanks for the clarification. Let me know if I can help getting this feature across the finish line.

davidzhao:
it turns out quite a bit has been updated in the API. this is implemented in #3420

PR #3328, opened 2025-09-02
feat: OpenAI Realtime Model Token/Cost Reporting for LangFuse via OpenTelemetry

bml1g12:
# Realtime Model Metrics Attribution and Span Hierarchy for Langfuse Integration (Cost Reporting)

This PR follows this slack thread https://livekit-users.slack.com/archives/C07FY8WHGPM/p1754548905020589?thread_ts=1754489320.734619&cid=C07FY8WHGPM with @longcw 

## Problem Statement

The LiveKit Agents framework needed improvements to OpenTelemetry span attribution and Langfuse integration for realtime LLM models:

1. **Span Hierarchy**: `realtime_assistant_turn` spans sometimes appeared as separate root spans instead of being nested under `agent_session` for some speech generation events 
2. **Metrics Attribution**: Realtime model metrics (token counts, costs) were not attributed to any spans. 
3. **Langfuse Classification**: Spans with token metrics weren't being recognized as "generations" in Langfuse 

## Solution

Some challenges encountered in this PR and my solutions:

1. OpenTelemetry does not yet have an [official namespace ](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-token-type)for cached and audio tokens

 --> To solve this I have used the langfuse recommended approach [here](https://langfuse.com/integrations/native/opentelemetry#usage) of e.g. `gen_ai.usage.input_cached_tokens` namespace 

2. Realtime APIs are highly concurrent, and can have e.g. multiple tool calls and speech generation async within a `realtime_assistant_turn`, or even finishing after the `realtime_assistant_turn` has ended. OpenTelemetry does not support adding metadata to ended spans, so we cannot simply retrospectively add the metrics (token counts) to each span, we need to do it whilst they are still in progress, or in a new span.
--> Assuming we wish to correctly nest everything within a `realtime_assistant_turn` and a single trace, the solution I came to in this PR is a span manager class, which intelligently allocates spans based on their type to the most relevant ongoing span, and if they have all finished by the time the metrics are collected (`agent_activity/_on_metrics_collected()`) then just make a new span called `realtime_metrics`

To illustrate this, if we naively just append data to current span in the `_on_metrics_collected()` it will sometimes go to the top level span and sometimes there will not be any spans at all as e.g. the function call span has already ended.

```
realtime_assistant_turn span starts
├── agent_speaking span starts (if audio output)
├── function_tool span(s) start (if tool calls)
├── realtime_assistant_turn span ENDS
├── agent_speaking continues...
├── function_tool(s) continue...
├── Metrics emitted (orphaned!)
└── agent_speaking/function_tool spans end
```
In thie above case, we create a new span, and in the simpler case that the function tool span has not ended, this PR correctly attributes the tokens/cost to that span. 
  

3. In order to add metrics like token counts to spans, we need a reference to the spans in the metrics event handler (`(`agent_activity/_on_metrics_collected()`) `) . We have already the "request_id" there, so I made use of this to link back to the associated realtime generation (which maybe a little confusingly is called `response_id` in the existing codebase). This required a small edit to the `GenerationCreatedEvent` across plugs to persist this metadata consistently.    

### Core Architecture Changes

**New `RealtimeSpanManager` class** (`telemetry/utils.py`)
- Encapsulates realtime span management logic
- Implements hierarchical span attribution system:
  1. Active `function_tool` span (most specific)
  4. Active `agent_speaking` span 
  5. Active `realtime_assistant_turn` span
  6. Fallback metrics span if needed

** :bug: Fixed Context Propagation** (`agent_activity.py`)
- Updated `_create_speech_task` to propagate OpenTelemetry context (AsyncIO tasks needed to inherit OpenTelemetry context)

**Enhanced Langfuse Integration**
- Provide token counts for realtime generations
- Added "model" attributes for automatic generation classification. By tagging all realtime generations with the model name, we ensure that LangFuse recognises them as "generations" and can automatically attribute costs (given token counts are provided)

#### Framework Updates
- **`llm/realtime.py`**: Added `response_id` field to `GenerationCreatedEvent`
- **`telemetry/trace_types.py`**: Added realtime metrics and Langfuse-compatible attributes
- **`telemetry/utils.py`**: New `RealtimeSpanManager`, `RealtimeSpanContext`, and utility functions (+310 lines)
- **`voice/agent_activity.py`**: Integrated span manager and fixed context propagation
- **`voice/agent_session.py`**: Registered `agent_speaking` spans with manager
- **`voice/generation.py`**: Added timing attributes to `function_tool` spans

#### Plugin Updates
All realtime model plugins updated to set `response_id` in `GenerationCreatedEvent`:
- **OpenAI Plugin**: `response_id=event.response.id`
- **AWS Plugin**: `response_id=self._current_generation.response_id`
- **Google Plugin**: `response_id=self._current_generation.response_id`

## Benefits

1. **Consistent Observability**: All realtime spans properly nested under `agent_session`
2. **Accurate Metrics**: Token counts and costs correctly attributed to relevant spans
3. **Proper Langfuse Integration**: Spans correctly classified with accurate latency
4. **Memory Efficient**: If there was either a memory leak (due to incorrect cleanup of references to realtime spans) or high consumption due to a huge number of parallel realtime calls, the use of a BoundedDict class prevents the number of references being stored in memory from exploding. Of course there is also a try/finally to correctly cleanup references to spans after they are processed. 
5. **Intelligent Attribution**: Realtime Metrics attached to most relevant active span

Simply adding metrics to `realtime_assistant_turn` is not possible as you can have more than one tool per span and also because the span can actually end before the metric associated it emited as the metric is emitted when the function/tool speech finishes. It is not possible in OpenTelemetry (unlike langfuse) to edit ended spans.
```
realtime_assistant_turn span starts
├── agent_speaking span starts (if audio output)
├── function_tool span(s) start (if tool calls)
├── realtime_assistant_turn span ENDS
├── agent_speaking continues...
├── function_tool(s) continue...
├── Metrics emitted (orphaned!)
└── agent_speaking/function_tool spans end
```

## Notes for Future Improvement 

- I limited the scope of this PR to the OpenAI Realtime API, as AWS and Google realtime models would need more testing and also because I can see some core differences. e.g. it think the AWS one streams metrics during generation, so this approach would not work without some small tweaks.
- I could not find a way to get a handle on the name of the model without hackery  - is there a better way that would work for _all_ realtime models? Abstract realtime model class seems like its missing a model name attribute 
```python
        try:
            model_name = self.llm._opts.model  # type: ignore[attr-defined]
        except AttributeError:
            model_name = "unknown-realtime-model"
```
- I seems to me uv.lock needs updating - but was unclear if I should be commiting that as its unrelated to my PR
## Example output

image


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -88,6 +89,9 @@ class AgentActivity(RecognitionHooks):
     def __init__(self, agent: Agent, sess: AgentSession) -> None:
         self._agent, self._session = agent, sess
         self._rt_session: llm.RealtimeSession | None = None
+        # OpenTelemetry span manager for realtime generations
+        self._realtime_span_manager = RealtimeSpanManager(maxsize=100)
```
bml1g12: This class is only needed when we have a realtime session - so in principle could be initialised to None then later set like the self._rt_session but I was not sure the idiomatic way to achieve that in the codebase. As it is, this class is very small - its just holds an empty dictionary if not in use, so no problem per say in initialising it not to None

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1888,8 +1909,28 @@ async def _realtime_generation_task(
...
+            )
+
         assert self._rt_session is not None, "rt_session is not available"
         assert isinstance(self.llm, llm.RealtimeModel), "llm is not a realtime model"
+        try:
+            # This attribute is specific to OpenAI realtime model.
+            # TODO: implement a generic way across all realtime models to obtain the model name
+            # from livekit.plugins.openai.realtime.realtime_model import RealtimeModel
+            # would be the correct type, but this class should be plugin agnostic presumably.
+            model_name = self.llm._opts.model  # type: ignore[attr-defined]
```
longcw: we can add a `label` property to the `llm.RealtimeModel`, like that in `llm.LLM`.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1888,8 +1909,28 @@ async def _realtime_generation_task(
...
+        except AttributeError:
+            model_name = "unknown-realtime-model"
+
+        current_span.set_attribute(trace_types.ATTR_GEN_AI_REQUEST_MODEL, model_name)
+
+        # Store the model name in the span manager for metrics attribution
+        self._realtime_span_manager.model_name = model_name
```
longcw: I think it's better to set the model name when creating the `_realtime_span_manager` (so create the manager later after rt_session created).

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1028,6 +1028,11 @@ def _update_agent_state(self, state: AgentState) -> None:
             if self._agent_speaking_span is None:
                 self._agent_speaking_span = tracer.start_span("agent_speaking")
                 self._agent_speaking_span.set_attribute(trace_types.ATTR_START_TIME, time.time())
+                # Register the speaking span with active realtime contexts
+                if self._activity:
+                    self._activity._realtime_span_manager.register_agent_speaking_span(
+                        self._agent_speaking_span
```
longcw: if we are sure the `_update_agent_state` is called inside the `_realtime_generation_task` (a realtime turn), do we still need this? and I think one agent speaking span should only nested to one realtime span.

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -512,6 +512,19 @@ async def _traceable_fnc_tool(
                         trace_types.ATTR_FUNCTION_TOOL_ARGS, fnc_call.arguments
                     )
 
+                    # Register the function tool span with active realtime contexts
+                    # Get the current activity from the context variable
+                    from .agent_activity import _AgentActivityContextVar
+
+                    try:
+                        if activity := _AgentActivityContextVar.get(None):
+                            activity._realtime_span_manager.register_function_tool_span(
```
longcw: same question here, `perform_tool_executions` is called inside the `_realtime_generation_task`, I thought it should always be nested under a realtime span. if that's not the case, do you know why it would happen?

bml1g12:
Thanks for the review, I will respond to each comment, but unrelated to that whilst I remember I just wanted to highlight one part of this PR. As currently implemented:

1) If an ongoing agent_speech or function_tool span has not ended when metrics are collected, we associate the token counting directly with that span
2) If the span is ended by the time the metrics are collected (which seems to happen fairly rarely, but sometimes - it depends on the asyncio event loop timings I guess), I create a new span nested under the same level as the relevant span, called "realtime_metrics"
You can see an example in attached screenshot 
image

I think this is no problem, in that its fairly readable and you can aggregate costs over the parent spans to get the data, but the only other alternative I see is that we never do (1) and instead always do (2) - keep me posted if you prefer that and I can revise it, but I favour it as-is personally as it means most of the time the metrics are tightly coupled to the relevant span as you'd expect

livekit-agents/livekit/agents/voice/agent_activity.py
bml1g12: Thanks, that would be very useful - keep me posted if implemented and I can revise this, or feel free to revise this directly when implemented; I assume it should be decoupled from this PR or done by a core contributor, given its part of the core library

livekit-agents/livekit/agents/voice/agent_activity.py
bml1g12: addressed this here https://github.com/livekit/agents/pull/3328/commits/35bb1fb17ee15a963a4ead02519cb0da89945778

livekit-agents/livekit/agents/voice/agent_activity.py
bml1g12: That makes sense, I have moved it here https://github.com/livekit/agents/pull/3328/commits/35bb1fb17ee15a963a4ead02519cb0da89945778

livekit-agents/livekit/agents/voice/agent_session.py
bml1g12: > if we are sure the _update_agent_state is called inside the _realtime_generation_task (a realtime turn), do we still need this?

It's pretty tough to follow the code execution flow but I am fairly confident it is indeed called inside the  `_realtime_generation_task`. 

It is called within it due to the callback `_on_first_frame` which seems to be called for first frame of audio or text generation from even from the realtime models

```python
        def _on_first_frame(_: asyncio.Future[None]) -> None:
            self._session._update_agent_state("speaking")
```


```
Path 1: User-initiated generation (via AgentSession.generate_reply())
agent_session.py:777-832 - AgentSession.generate_reply()
Public API method called by user code
agent_session.py:823 - self._activity._generate_reply()
Delegates to the current activity
agent_activity.py:787-839 - AgentActivity._generate_reply()
Checks if LLM is RealtimeModel at line 828
agent_activity.py:830 - self._realtime_reply_task()
Handles realtime-specific reply logic
agent_activity.py:1898 - await self._realtime_generation_task()
Called from within _realtime_reply_task
agent_activity.py:1959 - self._session._update_agent_state("speaking")
Called in _on_first_frame callback when audio/text generation starts
agent_session.py:1033-1040 - _update_agent_state("speaking")
Registers agent speaking span with realtime span manager (only for RealtimeModel)

Path 2: Server-initiated generation (via realtime model events)
Realtime LLM Server - Generates GenerationCreatedEvent (server-side turn detection)
agent_activity.py:1080 - _on_generation_created()
Event handler for server-initiated generations
agent_activity.py:1097 - self._realtime_generation_task()
Directly calls the generation task
agent_activity.py:1959 - self._session._update_agent_state("speaking")
Same callback as Path 1
agent_session.py:1033-1040 - _update_agent_state("speaking")
Same registration logic as Path 1
```

It would be great to be able to encapsulate all the realtime span logic in one place but I do not see how to do this


>  and I think one agent speaking span should only nested to one realtime span.

I suspect livekit already enforces no more than one speech generation simultanously (even though I believe OpenAI Realtime Model does actually support this via out of band communication) so I do not think there is a risk of more than one speech span per realtime_assistant_turn - at least I have not seen this happen personally yet   

If for whatever reason it did happen, I presume we would want to have more than one span per realtime_assistant_turn though to make it clear that it has happened?

bml1g12: Ah! On re-reading your comment, I think maybe there is a misunderstanding of what these `register_function_tool_span` and `register_agent_speaking_span` methods are doing - they are not creating the span, they are taking a note of the a reference to the span, so that later when we collect metrics and push tokens, we can _associate_ the metrics with the most relevant span

Maybe it would help if I rename these from `register_functon_tool_span` to `track_function_tool_span`? Like you imply, the span is already correctly nested regardless of whether we call this, its just a case of trying to get a hold of the relevant spans created by the realtime session.

livekit-agents/livekit/agents/voice/generation.py
bml1g12: yes this will be correctly nested under a realtime_assistant_turn

https://github.com/livekit/agents/pull/3328#discussion_r2322705194 I think this answers your question

bml1g12:
Thanks @longcw  @theomonnom I have responded to review

livekit-agents/livekit/agents/voice/agent_session.py
bml1g12: So I think I have clarified this with this refactor commit: https://github.com/livekit/agents/pull/3328/commits/11b7c031aeea8a40987f3452f93dc8ab5af28615

bml1g12:
It does occur to me that _maybe_ we could create a much simpler and more generic way across all models (STT, TTS, LLM, Realtime) for token reporting (=cost reporting) with opentelemetry if we were to extend the metrics object emitted by models (on_metrics_collected parameter) to include the reference to the opentelemetry span associated with those metrics. What do you think of this idea? I've not tested it but if you are interested in that approach, then I think whether it works or not would depend on whether the relevant span is always the "current span" when we emit metrics, which I'm not sure if the case or not. If that was the case though I think you'd have a much more generic solution and avoid the need to track references to realtime spans across the code. If we wanted to try this, I'd suggest it as a separate improvement PR as not sure how soon until I have time to implement it.

It might be helpful to establish some style guidelines for OpenTelemetry usage in the code base (e.g. it seems maybe there is a convention that plugins should not use tracing? I also see from the existing code that LiveKit is adopting a convention that one conversation within an agent session should remain as _one_ nested trace etc.)

longcw:
@bml1g12 thanks for the reply! I'll take a look on the updates soon.

bml1g12:
> @bml1g12 thanks for the reply! I'll take a look on the updates soon.

Thank you, I eagerly await your feedback, as will save our organisation needing to run from a custom fork of the repo if merged and deployed to livekit-agents package

bml1g12:
@longcw  sorry I know you must be busy, was just curious if you had any time to finalise the review here?

longcw:
@bml1g12 overall I think the solution makes sense, the core problem to solve is that the realtime metrics come too late after the `realtime_assistant_turn` finished, so you track a map from response_id to the span and add the metrics later.

one more question, why do you add the metrics to the `agent_speaking` span? IMO we should add it directly to the `realtime_assistant_turn`, in that case, we only track the `realtime_assistant_turn` span but not others, this should simplify the logic?

bml1g12:
> @bml1g12 overall I think the solution makes sense, the core problem to solve is that the realtime metrics come too late after the `realtime_assistant_turn` finished, so you track a map from response_id to the span and add the metrics later.

Thats right

> one more question, why do you add the metrics to the `agent_speaking` span? IMO we should add it directly to the `realtime_assistant_turn`, in that case, we only track the `realtime_assistant_turn` span but not others, this should simplify the logic?

I believe one  `realtime_assistant_turn` can correspond to more than one `on_metrics_collected()` invocation, as each invocation of that corresponds to a realtime API requests, and one `realtime_assistant_turn` could in principle call multiple tools, or call a tool and generate a response. For this reason I suspect it would not be easy to nest it under `realtime_assistant_turn` directly at least in a generic way. 

Conceptually, I think its also neat we can most of the time associate the token counts directly with the span taken i.e. whether it was speaking/tool calling (with the caveat in point 2. [here](https://github.com/livekit/agents/pull/3328#issuecomment-3254222475))

longcw:
> I believe one realtime_assistant_turn can correspond to more than one on_metrics_collected() invocation, as each invocation of that corresponds to a realtime API requests, and one realtime_assistant_turn could in principle call multiple tools, or call a tool and generate a response.

I don't think so, a `_realtime_generation_task` is corresponding to a single generation from the realtime api, the tool calls are executed in this task but the response will has a new `_realtime_generation_task`.


```python
    @tracer.start_as_current_span("realtime_assistant_turn")
    async def _realtime_generation_task(
        self,
        *,
        speech_handle: SpeechHandle,
        generation_ev: llm.GenerationCreatedEvent,
        model_settings: ModelSettings,
        instructions: str | None = None,
    ) -> None:
```
it's possible to have a generation id in the `llm.GenerationCreatedEvent` and use it to match the metrics sent from the realtime api.

bml1g12:
> > I believe one realtime_assistant_turn can correspond to more than one on_metrics_collected() invocation, as each invocation of that corresponds to a realtime API requests, and one realtime_assistant_turn could in principle call multiple tools, or call a tool and generate a response.
> 
> I don't think so, a `_realtime_generation_task` is corresponding to a single generation from the realtime api, the tool calls are executed in this task but the response will has a new `_realtime_generation_task`.
> 
> ```python
>     @tracer.start_as_current_span("realtime_assistant_turn")
>     @utils.log_exceptions(logger=logger)
>     async def _realtime_generation_task(
>         self,
>         *,
>         speech_handle: SpeechHandle,
>         generation_ev: llm.GenerationCreatedEvent,
>         model_settings: ModelSettings,
>         instructions: str | None = None,
>     ) -> None:
> ```
> 
> it's possible to have a generation id in the `llm.GenerationCreatedEvent` and use it to match the metrics sent from the realtime api.

Thanks for this suggestion, it indeed means I was able to simplify the code - works nicely!

image

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -360,7 +364,19 @@ def _create_speech_task(
...
+        current_context = otel_context.get_current()
+
+        # Create a wrapper coroutine that runs in the captured context
+        async def _context_aware_coro() -> Any:
+            # Attach the captured context before running the original coroutine
+            token = otel_context.attach(current_context)
+            try:
+                return await coro
+            finally:
+                otel_context.detach(token)
```
longcw, line 377: is this really needed? I thought it was automatically done when you create an async task. any issues did you see if we don't attach/detach the context explicitly?

bml1g12:
This PR was refactored into https://github.com/livekit/agents/pull/3439 and has been merged, so I will close this PR

I will followup with a seperate PR to address one bug I have noticed with above though: https://github.com/livekit/agents/pull/3454

Issue #3330, opened 2025-09-02
SSML Tags are getting dropped in TTS node for ElevenLabs on 1.2.7 and above


shivamkataria2000:
# SSML Tags are getting dropped in TTS node for ElevenLabs on 1.2.7 and above

## Problem
SSML `<break>` tags are being stripped from ElevenLabs TTS output after upgrading from v1.2.6 to v1.2.7+. Tags appear in LLM generation (visible in Langfuse traces) but disappear from final transcripts.

## Environment
- **Broken**: LiveKit v1.2.7, v1.2.8
- **Working**: LiveKit v1.2.6
- **TTS**: ElevenLabs with `enable_ssml_parsing=True`

## Reproduction
**Expected**: `Hey... <break time="0.4s" /> how's your Wednesday going?`
**Actual**: `Hey... . how's your Wednesday going?`

## Configuration
```python
{
    'model': 'eleven_flash_v2_5',
    'auto_mode': True,
    'enable_ssml_parsing': True
}
```

## Impact
Regression breaks natural speech timing/pacing control in conversational AI. 

<img width="1568" height="327" alt="Image" src="https://github.com/user-attachments/assets/0084fd45-11a4-4e24-811f-e5a9b972d828" /> 

<img width="1568" height="513" alt="Image" src="https://github.com/user-attachments/assets/588381f3-6e3e-4941-a1f5-abf97904fae6" />

longcw:
can you share your full agent configuration, do you have the `use_tts_aligned_transcript` enabled? if so the tag was actually sent to the TTS but filtered out from the output transcription by elevenlabs.

shivamkataria2000:
Yes, use_tts_aligned_transcript is set to True. I found the issue. This got introduced with this commit https://github.com/livekit/agents/commit/81c9cc8f46a93c8dc47e5c93c49d51acfefffc65 where we started using normalizedAlignment instead of alignment from 11labs response.

Create a PR to resolve it https://github.com/livekit/agents/pull/3336

yuyuma:
@shivamkataria2000 thanks for the issue and PR.  I was triaging a regression and came across this post which pointed us in the right direction.

[...]

PR #3336, opened 2025-09-03
feat: (elevenlabs-tts) Add option for preferred_transcript_alignment

shivamkataria2000:
### Description
Resolves [#3330](https://github.com/livekit/agents/issues/3330) - SSML tags are getting dropped in TTS node for ElevenLabs on 1.2.7 and above

### Problem
When _use_tts_aligned_transcript_ is enabled in ElevenLabs TTS, SSML tags were being stripped from the transcript output, causing users to lose important markup information that may be needed for downstream processing or display purposes.

### Solution
Added a new preferred_transcript_alignment configuration option to the ElevenLabs TTS node that allows users to choose how transcript text should be handled:

"_normalized_" (default): Uses the processed/normalized text without SSML tags (existing behavior)
"_original_": Retains the original input text including SSML tags in the transcript

### Technical Details

The new option only takes effect when _use_tts_aligned_transcript_ is set to true
Maintains backward compatibility by defaulting to "_normalized_" behavior
Allows users who need SSML preservation to opt into "_original_" mode

### Usage
```
tts=elevenlabs.TTS(
    preferred_transcript_alignment="original"  # Retains SSML tags
)

```
### Testing

 

- [x] Verified that "normalized" mode works as before
- [x]  Verified that "original" mode preserves SSML tags in transcript
- [x]  Confirmed backward compatibility with existing configurations

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/models.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/models.py
@@ -20,3 +20,8 @@ TTSEncoding = Literal[
     "mp3_44100_128",
     "mp3_44100_192",
 ]
+
+TranscriptAlignment = Literal[
+    "normalized",
+    "original",
+]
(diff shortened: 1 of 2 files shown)


longcw:
fixed in #3341

PR #3337, opened 2025-09-03
Transport additional arguments passed to Groq to wrapped OAI LLM

msaelices:
Groq LLM is a wrapper of the OpenAI one, and it is not accepting any additional argument the OAI LLM could accept, like the `max_completion_tokens`, so we cannot use it.


davidzhao:
we have a preference for explicit params here. added max_completion_tokens in #3370

PR #3338, opened 2025-09-03
feat(AWS): Enable prompt caching

itskyf:
Since AWS Bedrock only supports the "default" caching mode, I only add two boolean flags to enable the cache.

Document: https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html


longcw (approved):
looks good to me! can you fix the types before merge?

itskyf:
> looks good to me! can you fix the types before merge?

Sure, I have fixed it.

davidzhao (approved):
thanks for the PR!

PR #3340, opened 2025-09-03
Fix warning in model field access

marctorsoc:
Resolve a warning by updating the way model fields are accessed in the `_shallow_model_dump` function.


davidzhao (approved):
lgtm

Issue #3344, opened 2025-09-04
Tools throw error when working with OpenAI Realtime API


samdaviesudina:
Hi! I have an agent that uses the OpenAI Realtime API, and I'm seeing a consistent error like this one every time my agent calls a `function_tool`:

```2025-09-04 18:34:41,194 - ERROR livekit.plugins.openai - OpenAI Realtime API returned an error {"error": "Error(message=\"Tool call ID 'call_Xz0z4X8aSYDdkTf2' not found in conversation.\", type='invalid_request_error', code='invalid_tool_call_id', event_id=None, param=None)", "pid": 60986, "job_id": "AJ_ZSfSWq5stXMo"}```

Any help greatly appreciated! ☺️ 

Here is a minimal set-up to reproduce the error. The error happens in `collect_name`, when executing `return name`.

```python
# src/agent.py

from livekit.plugins import openai
from livekit.agents import function_tool, Agent, RunContext

from src.tasks.collect_name_task import CollectNameTask


class MyAgent(Agent):
    def __init__(self) -> None:
        super().__init__(
            instructions="Say hi, and then gather some info about the user",
            llm=openai.realtime.RealtimeModel(),
        )

    async def on_enter(self):
        await self.session.generate_reply()

    @function_tool()
    async def collect_name(self, _: RunContext) -> str:
        """Use this tool to collect the user's name."""
        name = await CollectNameTask()
        return name  # here is where the error happens
```

```python
# src/tasks/collect_name_task.py

from livekit.agents import AgentTask, function_tool
from livekit.plugins import openai


class CollectNameTask(AgentTask[str]):
    def __init__(self):
        super().__init__(
            instructions="Collect the user's name",
            llm=openai.realtime.RealtimeModel(),
        )

    async def on_enter(self) -> None:
        await self.session.generate_reply()

    @function_tool
    async def name_collected(self, name: str) -> None:
        """Use this when you've collected the user's name."""
        self.complete(name)

```

```python
# main.py

from dotenv import load_dotenv
from livekit import agents
from livekit.agents import (
    AgentSession,
    RoomInputOptions,
)
from livekit.plugins import noise_cancellation

from src.agents.agent import MyAgent

load_dotenv(".env")


async def entrypoint(ctx: agents.JobContext):
    session = AgentSession()
    await session.start(
        room=ctx.room,
        agent=MyAgent(),
        room_input_options=RoomInputOptions(
            noise_cancellation=noise_cancellation.BVCTelephony(),
[...]

PR #3345, opened 2025-09-05
add realtime model tool calls to chat ctx

longcw:
fix https://github.com/livekit/agents/issues/3344

tool call and output of realtime session are not added to `agent.chat_ctx`, so calling AgentTask and resuming the old agent will fail when it's in a function call.

### DIFF
--- a/livekit-agents/livekit/agents/tts/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/tts/fallback_adapter.py
@@ -268,7 +268,9 @@ class FallbackSynthesizeStream(SynthesizeStream):
         recovering: bool = False,
     ) -> AsyncGenerator[SynthesizedAudio, None]:
         # If TTS doesn't support streaming, wrap it with StreamAdapter
-        if not tts.capabilities.streaming:
+        if tts.capabilities.streaming:
+            stream = tts.stream(conn_options=conn_options)
+        else:
             from .. import tokenize
 
             wrapped_tts = StreamAdapter(
@@ -276,8 +278,6 @@ class FallbackSynthesizeStream(SynthesizeStream):
                 sentence_tokenizer=tokenize.blingfire.SentenceTokenizer(retain_format=True),
             )
             stream = wrapped_tts.stream(conn_options=conn_options)
-        else:
-            stream = tts.stream(conn_options=conn_options)
 
         @utils.log_exceptions(logger=logger)
         async def _forward_input_task() -> None:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -617,6 +617,9 @@ async def pause(self, *, blocked_tasks: list[asyncio.Task]) -> None:
         # When resuming, the AgentSession.update_agent must use the same AgentActivity instance!
         async with self._lock:
             await self._pause_scheduling_task(blocked_tasks=blocked_tasks)
+            if self._rt_session is not None:
+                # backup chat_ctx of the session for resume
```
theomonnom: Can you add a more explicit comment, it isn't obvious that tool calls aren't added to the agent chat_ctx when using the RealtimeSession

longcw: I am wondering should we add the tool call and output to the chat ctx or always use the rt_session.chat_ctx for realtime model?

theomonnom: I think we should add the tool calls to the chat_ctx

longcw: modified this pr to add tool call and output to the chat ctx for realtime model, this solve the issue for OAI realtime model as well.  cc @theomonnom 

but calling AgentTask in a function tool in other realtime time model doesn't work because it doesn't support resume the tool call context.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2042,6 +2042,9 @@ async def _read_messages(
         async def _read_fnc_stream() -> None:
             async for fnc in fnc_stream_for_tracing:
                 function_calls.append(fnc)
+                # add the function call to the chat context once it's received
+                self._agent._chat_ctx.items.append(fnc)
```
theomonnom: should we do this inside 
https://github.com/livekit/agents/blob/f8dd2bcb5a0215ad06407669d9c47ca2243fb1d9/livekit-agents/livekit/agents/voice/agent_activity.py#L2056
?

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -58,6 +58,7 @@ class RealtimeCapabilities:
     user_transcription: bool
     auto_tool_reply_generation: bool
     audio_output: bool
+    resumable_function_calls: bool
```
theomonnom:
```suggestion
    manual_function_calls: bool
```
maybe manual?

PR #3346, opened 2025-09-05
OAI realtime model wait for updated event

longcw:
wait for update event to be responded for `update_tools`, `update_instructions` and session starts.

this will raise an error if session update event is not responded before timeout, to avoid a silent failure.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1212,7 +1212,12 @@ class RealtimeSession(
             ev = self._create_tools_update_event(tools)
             self.send_event(ev)
 
-            await self._session_update_done.wait()  # TODO: add timeout
+            try:
+                await asyncio.wait_for(
+                    self._session_update_done.wait(), timeout=SESSION_UPDATE_TIMEOUT
+                )
+            except asyncio.TimeoutError:
+                raise llm.RealtimeError("update_tools timed out.") from None
 
             assert isinstance(ev.session, RealtimeSessionCreateRequest)
             assert ev.session.tools is not None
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -754,14 +774,31 @@ async def _reconnect() -> None:
             logger.debug("reconnected to OpenAI Realtime API")
             self.emit("session_reconnected", llm.RealtimeSessionReconnectedEvent())
 
+        async def _check_session_update_done() -> None:
+            try:
+                await asyncio.wait_for(
```
theomonnom, line 779: Could it also be done by just adding a lock to update_instructions, update_chat_ctx, etc?

longcw: I think it's different, this is to make sure the update session is successful, sometimes the update failed but the session continues and user didn't get notified.

theomonnom: do we know why session.update is failing?

longcw: for example #3431, the update failed when reconnection due to API mismatch, causing the voice changed in the middle of conversation. It might be confusing for the users and hard to debug, like the instruction is not applied, the voice is incorrect... we should make such error more explicit by raising an error when update failed.

theomonnom: It seems to add a lot of stuff to keep track of. So it may be harder for us to debug lol. 
Isn't just logging all the OpenAI errors enough?

longcw: perhaps raise an error when update option failed? users may ignore the error logs

longcw: btw, I think we already have the error logs from OpenAI

PR #3377, opened 2025-09-08
commit user turn before closing the AgentSession

longcw:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1222,6 +1222,16 @@ def on_end_of_turn(self, info: _EndOfTurnInfo) -> bool:
...
             )
+
+            # add user input to chat context
+            user_message = llm.ChatMessage(
+                role="user",
+                content=[info.new_transcript],
+                transcript_confidence=info.transcript_confidence,
+            )
+            self._agent._chat_ctx.items.append(user_message)
+            self._session._conversation_item_added(user_message)
```
theomonnom: It may happen when we do a "temporary" handoff to another agent (With AgentTask). Not only when closing

theomonnom: I think we need to add a new "closing" flag to the AgentSession?

longcw: do we also want to record the user message even it's a temporal handoff, otherwise the message is just lost?

theomonnom: I'm not sure, I'm just worried that inserting a message at this time may "drift" the agent

theomonnom: Fwiw, I've never seen the above log, so it seems very rare. I think it's OK to lose this message but just making sure we have it when closing the session

longcw: okay, I'll add the closing flag.

longcw: updated. btw, we didn't see this warning before because we didn't commit and wait for the user transcript before closing, after this pr I guess we will see it more often when user disconnects right after speaking.

theomonnom: lgtm, yeah let's see

PR #3381, opened 2025-09-08
Add minimum confidence level filter for deepgram STT.

meetakshay99:
Updated to send out stt results for deepgram that has a minimum amount of confidence to avoid transcriptions generated without user actually speaking.

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -39,7 +39,7 @@ from livekit import api, rtc
 from livekit.protocol import agent, models
 
 from . import ipc, telemetry, utils
-from ._exceptions import AssignmentTimeoutError
+from ._exceptions import APIStatusError, AssignmentTimeoutError
 from .inference_runner import _InferenceRunner
 from .job import (
     JobAcceptArguments,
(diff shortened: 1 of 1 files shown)


meetakshay99:
@davidzhao  Can you pls review and merge this PR if all fine?

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -63,6 +63,7 @@ class STTOptions:
     keywords: list[tuple[str, float]]
     keyterms: list[str]
     profanity_filter: bool
+    min_confidence_threshold: float
```
longcw: this should be added to the agent framework, not just for deepgrams IMO, e.g. maybe a option for the [AudioRecognition](https://github.com/livekit/agents/blob/livekit-agents@1.2.8/livekit-agents/livekit/agents/voice/audio_recognition.py#L66).

meetakshay99: yes, if its generalized for all STT Providers that would be great. But until then good to have here.

davidzhao:
@meetakshay99 it's currently failing CI, could you please correct the errors?

meetakshay99:
@davidzhao Have fixed the issue. Please check.

meetakshay99:
There is some formatting related error which I am unable to fix. I tried using the "black" command as suggested in the help but that too did not work. Hence reverted that change.
image

PR #3385, opened 2025-09-09
inference gateway api

longcw:


examples/voice_agents/basic_agent.py
```diff
@@ -89,9 +90,12 @@ async def entrypoint(ctx: JobContext):
...
         # any combination of STT, LLM, TTS, or realtime API can be used
-        llm=openai.LLM(model="gpt-4o-mini"),
-        stt=deepgram.STT(model="nova-3", language="multi"),
-        tts=openai.TTS(voice="ash"),
+        # llm=openai.LLM(model="gpt-4o-mini"),
+        # stt=deepgram.STT(model="nova-3", language="multi"),
+        # tts=openai.TTS(voice="ash"),
+        llm=inference.LLM(model="openai/gpt-4o-mini"),
+        stt=inference.STT(model="deepgram"),
+        tts=inference.TTS(model="elevenlabs", voice="cgSgspJ2msm6clMCkdW9"),
```
bcherry: are we still planning to support a simple string interface like `llm="openai/gpt-4o-mini"`?

livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -0,0 +1,393 @@
...
+    api_secret: str
+    verbosity: NotGivenOr[Verbosity]
+    extra_kwargs: dict[str, Any]
+
+
+class LLM(llm.LLM):
+    def __init__(
+        self,
+        *,
+        model: LLMModels | str = DEFAULT_MODEL,
```
theomonnom: I would make model `arg` instead of `kwarg`

livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -0,0 +1,393 @@
...
+        parallel_tool_calls: NotGivenOr[bool] = NOT_GIVEN,
+        tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
+        max_completion_tokens: NotGivenOr[int] = NOT_GIVEN,
+        base_url: NotGivenOr[str] = NOT_GIVEN,
+        api_key: NotGivenOr[str] = NOT_GIVEN,
+        api_secret: NotGivenOr[str] = NOT_GIVEN,
+        timeout: httpx.Timeout | None = None,
+        max_retries: NotGivenOr[int] = NOT_GIVEN,
+        verbosity: NotGivenOr[Verbosity] = NOT_GIVEN,
+        extra_kwargs: NotGivenOr[dict[str, Any]] = NOT_GIVEN,
```
theomonnom, line 214: Should top_p, temperature, tool_choice etc, be extra_kwargs now?

livekit-agents/livekit/agents/inference/tts.py
```diff
@@ -0,0 +1,311 @@
...
+        model: NotGivenOr[TTSModels | str] = NOT_GIVEN,  # TODO: add a default model
+        voice: NotGivenOr[str] = NOT_GIVEN,
+        language: NotGivenOr[str] = NOT_GIVEN,
+        encoding: NotGivenOr[TTSEncoding] = NOT_GIVEN,
+        sample_rate: NotGivenOr[int] = NOT_GIVEN,
+        base_url: NotGivenOr[str] = NOT_GIVEN,
+        api_key: NotGivenOr[str] = NOT_GIVEN,
+        api_secret: NotGivenOr[str] = NOT_GIVEN,
+        http_session: aiohttp.ClientSession | None = None,
+        extra_kwargs: NotGivenOr[dict[str, Any]] = NOT_GIVEN,
```
theomonnom, line 171: I think depending on the model "Literal", we could make extra_kwargs a TypedDict with all the available fields.

livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -0,0 +1,393 @@
...
+from ..types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, APIConnectOptions, NotGivenOr
+from ..utils import is_given
+from ._utils import create_access_token
+from .models import LLMModels
+
+lk_oai_debug = int(os.getenv("LK_OPENAI_DEBUG", 0))
+
+Verbosity = Literal["low", "medium", "high"]
+DEFAULT_BASE_URL = "https://agent-gateway.livekit.cloud/v1"
+DEFAULT_MODEL = "openai/gpt-4o-mini"
```
adrian-cowham: @theomonnom @longcw We don't have a relationship with OpenAI, but we do with Azure. What do you think about changing this to `azure/gpt-4o-mini`?

adrian-cowham: Technically, `openai/gpt-4o-mini` will work, but since we don't have a business agreement with OpenAI we can't make any guarantees to our customers regarding reliability.

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -0,0 +1,370 @@
...
+
+from .. import stt, utils
+from .._exceptions import APIConnectionError, APIError, APIStatusError
+from ..log import logger
+from ..types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, APIConnectOptions, NotGivenOr
+from ..utils import is_given
+from ._utils import create_access_token
+from .models import STTLanguages, STTModels
+
+STTEncoding = Literal["pcm_s16le"]
```
adrian-cowham, line 67: The gateway also supports `pcm_mulaw`

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -0,0 +1,370 @@
...
+    api_key: str
+    api_secret: str
+    extra_kwargs: dict[str, Any]
+
+
+class STT(stt.STT):
+    def __init__(
+        self,
+        *,
+        model: NotGivenOr[STTModels | str] = NOT_GIVEN,  # TODO: add a default model
```
adrian-cowham: You can add a default model on the agent side if you wish, the Gateway will also choose one if the agent doesn't provide one.

Language can also be empty, we choose `en` as the default. So, the Gateway will do the following:

| language | provider | model      | result.    |
|----------|----------|----------|---------|
| none        | none       |  none       | defaults to english, choose preferred provider for english and default model from the provider|
| lang    |  none     |  none | chooses the preferred provider for lang and the default model from the provider |
| none | assemblyai | none |  chooses assemblyai and sets the lang to english|
| lang | assemblyai | none |  chooses assemblyai and sets the lang to lang and validates language support|
| lang | | model_name |  error: if model is specified, provider needs to also be specified |

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -0,0 +1,370 @@
...
+                elif msg_type == "interim_transcript":
+                    self._process_transcript(data, is_final=False)
+                elif msg_type == "final_transcript":
+                    self._process_transcript(data, is_final=True)
+                elif msg_type == "session.finalized":
+                    pass
+                elif msg_type == "session.closed":
+                    pass
+                elif msg_type == "error":
+                    raise APIError(f"LiveKit STT returned error: {msg.data}")
```
adrian-cowham, line 373: Will the agent retry after this error is raised?

longcw: yes

livekit-agents/livekit/agents/inference/stt.py
longcw: what is the decoding format for `pcm_mulaw`, same as `pcm_s16le`?

livekit-agents/livekit/agents/inference/llm.py
longcw: if they are not supported by all models then yes, they should be in extra_kwargs.

examples/voice_agents/basic_agent.py
longcw: yes, it's supported. we will create a `inference.LLM` if llm is a string in `AgentSession` and `Agent`.

livekit-agents/livekit/agents/inference/stt.py
adrian-cowham: No, `pcm_mulaw` is compressed. We're probably ok not supporting it.

livekit-agents/livekit/agents/inference/tts.py
```diff
@@ -0,0 +1,311 @@
...
+            api_secret=lk_api_secret,
+            extra_kwargs=extra_kwargs if is_given(extra_kwargs) else {},
+        )
+        self._session = http_session
+        self._pool = utils.ConnectionPool[aiohttp.ClientWebSocketResponse](
+            connect_cb=self._connect_ws,
+            close_cb=self._close_ws,
+            max_session_duration=300,
+            mark_refreshed_on_get=True,
+        )
```
adrian-cowham, line 266: How does connection pooling affect the user's connection limit?

livekit-agents/livekit/agents/inference/stt.py
longcw: this sounds good to me.

livekit-agents/livekit/agents/inference/tts.py
longcw: it will reuse the ws connections and reconnect it after `max_session_duration`, right now set as 5mins.

livekit-agents/livekit/agents/inference/stt.py
longcw: does this also work for tts?

livekit-agents/livekit/agents/inference/tts.py
longcw: one issue is that `model` will have a lot of combinations with format `provider/model:voice`, we will need to add a `provider` option for TypedDict.

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -0,0 +1,378 @@
...
+        lk_base_url = (
+            base_url
+            if is_given(base_url)
+            else os.environ.get("LIVEKIT_GATEWAY_URL", DEFAULT_BASE_URL)
+        )
+
+        lk_api_key = (
+            api_key
+            if is_given(api_key)
+            else os.getenv("LIVEKIT_GATEWAY_API_KEY", os.getenv("LIVEKIT_API_KEY", ""))
```
longcw, line 188: wdyt about reading from `LIVEKIT_GATEWAY_API_KEY` first, if not set then use `LIVEKIT_API_KEY`?

livekit-agents/livekit/agents/inference/models.py
```diff
@@ -0,0 +1,19 @@
...
+    "cartesia/ink-whisper",
+    "deepgram/nova-3",
+    "assemblyai",
+]
+
+STTLanguages = Literal["en", "de", "es", "fr", "ja", "pt", "zh"]
+
+LLMModels = Literal[
+    "openai/gpt4o",
+    "...",
```
longcw: @adrian-cowham  do we have a list of supported tts, stt, llm models?

livekit-agents/livekit/agents/inference/models.py
```diff
@@ -0,0 +1,19 @@
...
+]
+
+STTModels = Literal[
+    "cartesia/ink-whisper",
+    "deepgram/nova-3",
+    "assemblyai",
+]
+
+STTLanguages = Literal["en", "de", "es", "fr", "ja", "pt", "zh"]
```
longcw: will all STT support these languages?

livekit-agents/livekit/agents/inference/stt.py
adrian-cowham: Why does are there separate API keys? The Agent is making the request on the developer's behalf and the API key needs to be attached to the developer. The Gateway uses the `project_id` for quotas and billing. Does this help answer your question?

livekit-agents/livekit/agents/inference/models.py
adrian-cowham: No. For example, AssemblyAI only supports `en`. Here is a list of supported providers and their languages.
https://github.com/livekit/agent-gateway/blob/main/pkg/provider/stt/assemblyai/assemblyai.go#L11-L24
https://github.com/livekit/agent-gateway/blob/main/pkg/provider/stt/cartesia/cartesia.go#L64-L178
https://github.com/livekit/agent-gateway/blob/main/pkg/provider/stt/deepgram/deepgram.go#L11-L76

livekit-agents/livekit/agents/inference/models.py
adrian-cowham: Yes, @tmshapland put together a beautiful Notion doc here:

https://www.notion.so/livekit/Gateway-Metering-Quotas-and-Pricing-2393c4901a428036b0d3c266b28ce4fc?source=copy_link

livekit-agents/livekit/agents/inference/stt.py
adrian-cowham: Not yet, but it will. It's setting in a PR. I'm going to add `DefaultVoice` to each provider as well too.

livekit-agents/livekit/agents/inference/llm.py
bcherry: IMO we probably shouldn't use a default model at all for LLM. the user should be required to specify one. otherwise we will put ourselves in a tricky place if we ever want to change the default, because it could immediately break or alter behavior of any agents in production on the old default. LLM behavior is just too variable to handle this way...

we can make `azure/gpt-4o-mini` the default in examples, templates, and docs. 

although I'm also unsure whether we should really encourage most users to specify the provider at all. there's not necessarily meaningful differences between them in terms of behavior, the differences are all in reliability / performance / etc. Those are all decisions that the gateway product is designed to abstract away from developers - so why make them think about it? If they want gpt-4o-mini, the default behavior should be it can serve from openai or from azure interchangeably based on current conditions.

this maybe breaks down when the price of an individual model varies by provider, though.

Also do we support the openai platform at all? it's not listed in the pricing sheet @tmshapland shared

theomonnom: agree, we don't need to specify `azure`, `openai/gpt-4o` should just use whatever best provider we have

adrian-cowham: Totally agree. And we should do this for open source models as well. However, we currently don't have the ability to determine the best provider based on performance metrics...which is OK. For OpenAI models we can just default to Azure since it's actually supposed to be the better provider.

This is a change I will have to make in the Gateway. If the provider isn't supplied AND it's a supported OpenAI model, serve inference from Azure for the requested model.

livekit-agents/livekit/agents/inference/stt.py
longcw: just for the case ppl want to use a different key for the gateway, like OSS livekit server but still use cloud gateway. maybe removing the env is fine for most of the cases.

livekit-agents/livekit/agents/inference/stt.py
longcw: It seems that when I put a wrong model name the gateway returns an empty dict `received unexpected message from LiveKit STT: {}` and closes the the connection directly. Can we add an error message?

livekit-agents/livekit/agents/inference/models.py
```diff
@@ -0,0 +1,179 @@
...
+
+class _ElevenLabsVoiceSettings(TypedDict, total=False):
+    stability: float  # [0.0 - 1.0]
+    similarity_boost: float  # [0.0 - 1.0]
+    style: float  # [0.0 - 1.0]
+    speed: float  # [0.8 - 1.2]
+    use_speaker_boost: bool
+
+
+class TTSOptions_ElevenLabs(TypedDict, total=False):
```
longcw: are these options supported by the gateway? @adrian-cowham

livekit-agents/livekit/agents/inference/models.py
```diff
@@ -0,0 +1,179 @@
+from __future__ import annotations
+
+from typing import Literal, TypedDict, Union
+
+TTSModels_Cartesia = Literal[
```
longcw: open to better variable names

theomonnom: I would mb just do CartesiaModels? 
mb it's fine if we move this inside the tts.py

livekit-agents/livekit/agents/inference/models.py
```diff
@@ -0,0 +1,179 @@
...
+    "cartesia", "cartesia/sonic", "cartesia/sonic-2", "cartesia/sonic-turbo"
+]
+TTSModels_ElevenLabs = Literal[
+    "elevenlabs",
+    "elevenlabs/eleven_flash_v2",
+    "elevenlabs/eleven_flash_v2_5",
+    "elevenlabs/eleven_turbo_v2",
+    "elevenlabs/eleven_turbo_v2_5",
+    "elevenlabs/eleven_multilingual_v2",
+    "elevenlabs/eleven_multilingual_v3",
```
adrian-cowham: I think we should remove `eleven_multilingual_v3`, it's not supported by the Gateway. Checkout their docs:
https://elevenlabs.io/docs/models#eleven-v3-alpha

livekit-agents/livekit/agents/inference/models.py
```diff
@@ -0,0 +1,179 @@
...
+    "deepgram/nova-3-medical",
+    "deepgram/nova-2",
+    "deepgram/nova-2-general",
+    "deepgram/nova-2-medical",
+    "deepgram/nova-2-phonecall",
+]
+STTModels_Cartesia = Literal["cartesia", "cartesia/ink-whisper"]
+STTModels_Assemblyai = Literal[
+    "assemblyai",
+    # "assemblyai/universal-streaming",  # wrong name?
```
adrian-cowham: Yeah, the API doesn't accept a model name.

livekit-agents/livekit/agents/inference/models.py
adrian-cowham: @longcw Not currently, the only supported options are `auto_mode` and `apply_text_normalization`. I can add support pretty easily tho. However, for v1 I don't think we need to support every option. @theomonnom do you have an opinion?

longcw: I think we doesn't need to support all the options in v1, it would be great if you have a list of supported options for different providers.

livekit-agents/livekit/agents/inference/models.py
longcw: updated

livekit-agents/livekit/agents/inference/models.py
adrian-cowham: @long you're right, let me put together a doc for you.

adrian-cowham: @longcw Here's quick and dirty doc, let me know if you have questions or feedback.
https://www.notion.so/livekit/Gateway-Supported-Provider-Params-2703c4901a428035a52df37c9f799a85?source=copy_link

longcw: updated accordingly

adrian-cowham: @longcw I updated that notion doc, I added `keyterms_prompt` for AssemblyAI

bcherry: @adrian-cowham can you give me access to that too? i'm working on docs

adrian-cowham: @bcherry Done. I made it public: https://www.notion.so/livekit/Gateway-Supported-Provider-Params-2703c4901a428035a52df37c9f799a85?source=copy_link

livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -0,0 +1,582 @@
...
+    "groq/qwen/qwen3-32b",
+]
+BasetenModels = Literal[
+    "baseten/deepseek-ai/DeepSeek-V3-0324",
+    "baseten/meta-llama/Llama-4-Scout-17B-16E-Instruct",
+    "baseten/meta-llama/Llama-4-Maverick-17B-128E-Instruct",
+    "baseten/moonshotai/Kimi-K2-Instruct",
+    "baseten/openai/gpt-oss-120b",
+    "baseten/Qwen/Qwen3-235B-A22B-Instruct-2507",
+]
```
davidzhao, line 83: @adrian-cowham do we want to standardize these in terms of lowercasing/etc? or maybe the server-side should just always register everything in a case insensitive way

livekit-agents/livekit/agents/inference/stt.py
longcw: @adrian-cowham could you take a look on this as well, the error message like wrong configuration, missing voice id is not forwarded to the agent.

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -0,0 +1,496 @@
...
+        sample_rate: NotGivenOr[int] = NOT_GIVEN,
+        api_key: NotGivenOr[str] = NOT_GIVEN,
+        api_secret: NotGivenOr[str] = NOT_GIVEN,
+        http_session: aiohttp.ClientSession | None = None,
+        extra_kwargs: NotGivenOr[dict[str, Any]] = NOT_GIVEN,
+    ) -> None: ...
+
+    def __init__(
+        self,
+        model: NotGivenOr[STTModels | str] = NOT_GIVEN,
```
theomonnom, line 149: is it possible to directly do "lang/en-US"?

livekit-agents/livekit/agents/inference/stt.py
adrian-cowham: @longcw Absolutely, I will add better error messaging, this has been on my todo. What's you opinion on closing the connection? Should the gateway still close the connection?

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -0,0 +1,497 @@
...
+    "deepgram/nova-2",
+    "deepgram/nova-2-general",
+    "deepgram/nova-2-medical",
+    "deepgram/nova-2-phonecall",
+]
+CartesiaModels = Literal[
+    "cartesia",
+    "cartesia/ink-whisper",
+]
+AssemblyaiModels = Literal["assemblyai"]
```
bcherry, line 36: for consistency can we also support `"assemblyai/default"` as an option? that way you're always allowed to include a model name

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -0,0 +1,497 @@
...
+
+DeepgramModels = Literal[
+    "deepgram",
+    "deepgram/nova-3",
+    "deepgram/nova-3-general",
+    "deepgram/nova-3-medical",
+    "deepgram/nova-2",
+    "deepgram/nova-2-general",
+    "deepgram/nova-2-medical",
+    "deepgram/nova-2-phonecall",
```
bcherry, line 30:
```suggestion
    "deepgram/nova-2-conversational",
```

davidzhao: IIRC phonecall is the right thing

https://developers.deepgram.com/docs/models-languages-overview#nova-2

livekit-agents/livekit/agents/inference/llm.py
bcherry, line 83: the inference gateway code shows that we have nicer aliases for most of these models, which is probably what we should focus on?

here's what's going into docs

```
   azure: [
    { slug: 'azure/gpt-4o' },
    { slug: 'azure/gpt-4o-mini', default: true },
    { slug: 'azure/gpt-4.1' },
    { slug: 'azure/gpt-4.1-mini' },
    { slug: 'azure/gpt-4.1-nano' },
    { slug: 'azure/gpt-5' },
    { slug: 'azure/gpt-5-mini' },
    { slug: 'azure/gpt-5-nano' },
  ],
  baseten: [
    { slug: 'baseten/qwen-3-235b', default: true },
    { slug: 'baseten/llama-4-maverick' },
    { slug: 'baseten/llama-4-scout' },
    { slug: 'baseten/deepseek-v3' },
    { slug: 'baseten/kimi-k2' },
    { slug: 'baseten/gpt-oss-120b' },
  ],
  cerebras: [
    { slug: 'cerebras/llama-3.1-8b' },
    { slug: 'cerebras/llama-3.3-70b', default: true },
    { slug: 'cerebras/llama-4-scout-17b-16e-instruct' },
    { slug: 'cerebras/qwen-3-32b' },
    { slug: 'cerebras/qwen-3-235b-a22b-instruct-2507' },
  ],
  groq: [
    { slug: 'groq/llama-3-8b' },
    { slug: 'groq/llama-3-70b' },
    { slug: 'groq/llama-3.3-70b', default: true },
    { slug: 'groq/llama-4-scout' },
    { slug: 'groq/llama-4-maverick' },
    { slug: 'groq/gpt-oss-120b' },
    { slug: 'groq/kimi-k2' },
    { slug: 'groq/qwen-3-32b' },
  ],
```

livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -0,0 +1,582 @@
...
+    "openai/gpt-4o-2024-05-13",
+    "openai/gpt-4o-2024-07-18",
+    "openai/gpt-4o-mini",
+]
+GoogleModels = Literal[
+    "google/gemini-2.5-pro-preview-05-06",
+    "google/gemini-2.5-flash-preview-04-17",
+    "google/gemini-2.5-flash-preview-05-20",
+    "google/gemini-2.0-flash-001",
+    "google/gemini-2.0-flash-lite-preview-02-05",
```
bcherry: i don't think google is ready

davidzhao (approved):
nice work!

livekit-agents/livekit/agents/inference/llm.py
adrian-cowham: Gemini is not supported yet.

livekit-agents/livekit/agents/inference/stt.py
longcw: right now no, it needs to be `STT(language="en-US")`

livekit-agents/livekit/agents/inference/stt.py
longcw: `nova-2-conversationalai` and `nova-2-phonecall` are two models

livekit-agents/livekit/agents/inference/llm.py
longcw: just tested with all providers, it seems baseten and cerebras are not working, can you confirm @adrian-cowham?

and these aliases seem don't work right now, for example, `groq/meta-llama/llama-4-maverick-17b-128e-instruct` works but `groq/llama-4-maverick` doesn't.

IMO, we should use the original names from the providers in v1, it's easier for users to copy-paste names from the provider's document, and it's supported automatically for any new models from the providers that not in our list.

For example, `llama-3.1-8b-instant` is supported in https://console.groq.com/docs/models#production-models but not in our list (I'll add it), user can still use it with `groq/llama-3.1-8b-instant`. FYI, `groq/llama3-8b` is no longer supported. We should add link to the provider's model page in the doc IMO.

longcw: `azure/gpt-4o` and `azure/gpt-4o-mini` got a timeout error from azure, `azure/gpt-5` and other gpt-5 models is not supported and raises a 400 `Error proxying completions: model not found in provider config`.

I guess there is a config list in the gateway, should we allow unknown models for a provider and just pass it through when it's not in our list?

longcw: @bcherry what is the default per provider? do we have a default model for each provider so we only use `llm="azure"`? I don't think it's supported right now and we should always require a model name IMO.

adrian-cowham: @long and I spoke on slack, the model discrepancies are because he was using production instead of staging.

adrian-cowham: @longcw I should note that even though groq supports `llama-3.1-8b-instant`, we do not. It is not in Tom's list of supported models for the Gateway.

longcw: @adrian-cowham okay, but why `llama-3.1-8b-instant` is not included, and `llama3-70b-8192` and `llama3-8b-8192` are deprecated by groq, we should remove them.
> on May 31, 2025, we emailed users to announce the deprecation of llama3-70b-8192 and llama3-8b-8192 in favor of llama-3.3-70b-versatile and llama-3.1-8b-instant respectively. 

my question is do we have a hard selection of the models per provider, or users can use what ever models listed in their document?

livekit-agents/livekit/agents/inference/stt.py
bcherry: I was going off of the gateway code, which only has one of them included right now https://github.com/livekit/agent-gateway/blob/main/pkg/provider/stt/deepgram/deepgram.go cc @adrian-cowham

livekit-agents/livekit/agents/inference/stt.py
bcherry: this is fine with me - do we think this will be the final design? i wrote the docs based on theo's comment but it's very easy to change

livekit-agents/livekit/agents/inference/llm.py
bcherry: I don't think we can allow people to use models we haven't specifically included, because we need to set up pricing for each

bcherry: also while we do have a default model per provider i don't think we should encourage it and would be fine if we even removed support. i also agree devs should have to specify the model

adrian-cowham: I will remove the deprecated models, thanks @longcw

adrian-cowham: Yup, gateway should handle this. I have aliases defined in the code that are all lowercase and friendlier to read.

livekit-agents/livekit/agents/inference/stt.py
adrian-cowham: I'll add the `phonecall` model

longcw: and the other one should be `nova-2-conversationalai`?

adrian-cowham: Yup! I'll fix that.

livekit-agents/livekit/agents/inference/stt.py
longcw: for the first version I think we won't support string config like `lang/en-US`, but it could be a nice feature in the future, let's see what's the feedback on v1?

livekit-agents/livekit/agents/inference/llm.py
longcw: > I don't think we can allow people to use models we haven't specifically included, because we need to set up pricing for each

yeah this makes sense, I'll update the model list when it's finalized.

PR #3403, opened 2025-09-10
Fix PydanticDeprecatedSince211 warning - Accessing the 'model_fields' attribute on the instance is deprecated

hudson-worden:
Fixes the following warning

packages/livekit/agents/llm/utils.py:410: PydanticDeprecatedSince211: Accessing the 'model_fields' attribute on the instance is deprecated. Instead, you should access this attribute from the model class. Deprecated in Pydantic V2.11 to be removed in V3.0.


hudson-worden:
This is currently failing type checks for a problem that's already in main

> livekit-agents/livekit/agents/tts/fallback_adapter.py:280: error: Incompatible types in assignment (expression has type "SynthesizeStream", variable has type "StreamAdapterWrapper")  [assignment]

davidzhao:
I don't see the failure in main, maybe rebase?

livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -405,9 +405,12 @@ def _is_optional_type(hint: Any) -> bool:
     return is_union and type(None) in get_args(hint)
 
 
-def _shallow_model_dump(model: BaseModel, *, by_alias: bool = False) -> dict[str, Any]:
+T = TypeVar("T", bound=BaseModel)
+
+
+def _shallow_model_dump(model_type: type[T], model: T, *, by_alias: bool = False) -> dict[str, Any]:
     result = {}
-    for name, field in model.__class__.model_fields.items():
```
davidzhao, line 421: this is already accessing the type on the class, how are you reproducing this warning?

hudson-worden: @davidzhao you are correct, we're on an older version of the lib so I mistakenly thought that the line was still an issue.

Issue #3396, opened 2025-09-10
user_speaking span sometimes has enormous duration_ms


pesterhazy:
In opentelemetry, the user_speaking span sometimes has an excessive length, e.g. 18446744073700.02 ms

<img width="2166" height="610" alt="Image" src="https://github.com/user-attachments/assets/c5700e49-8818-431e-87c4-557e43ed4895" />

This skews the trace when looked at in, e.g., Honeycomb

pesterhazy:
Probably related to this line? https://github.com/livekit/agents/blob/f2fb5705c865ee7f220f60b5baeaf4c716db4cf6/livekit-agents/livekit/agents/voice/agent_session.py#L1060

tinalenguyen:
@pesterhazy I made a PR, let me know if that solves the issue!

longcw:
@pesterhazy can you share the start and end time of the user speaking span?

Bnowako:
Hey @longcw and @tinalenguyen I am from Paulus team and sharing more context:

The end_time is earlier:

```
lk.start_time: 1757614656.088785
lk.end_time: 1757614655.7160554
```
also take a look here on the distribution of user_speaking durations
<img width="1070" height="719" alt="Image" src="https://github.com/user-attachments/assets/464669d5-d810-4f2f-8a57-cdc9fe5ee8ed" />

Also will let you know if it worked out after we update :)

longcw:
so the actual problem is the end_time < start_time, will take a look. cc @tinalenguyen

Bnowako:
@longcw @tinalenguyen this issue returned when we bumped from 1.3.9 to 1.3.12

<img width="2462" height="1118" alt="Image" src="https://github.com/user-attachments/assets/0fffb7e4-b0be-4285-95ba-3657f3512aab" />

PR #3404, opened 2025-09-10
fix user speaking span duration

tinalenguyen:
addresses #3396 

removed converting the end time to nanoseconds for an accurate speaking duration span

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1057,7 +1057,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         elif self._user_speaking_span is not None:
             end_time = last_speaking_time or time.time()
             self._user_speaking_span.set_attribute(trace_types.ATTR_END_TIME, end_time)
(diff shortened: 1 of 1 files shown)


longcw:
The end time is intended to use the last speaking time for more accurate latency calculation. We should check if nano seconds is needed instead of remove it IMO. cc @tinalenguyen

tinalenguyen:
@longcw When would nano seconds be needed? Wouldn't that also be the case for the agent speaking span?

longcw:
> @longcw When would nano seconds be needed? Wouldn't that also be the case for the agent speaking span?

not for the agent speaking span, `last_speaking_time` is needed for user span because the VAD needs 0.4-0.5s to confirm user speaking ended after silence, so the last_speaking_time is not the time we call this function.

longcw:
OTEL doesn't specify the unit of the timestamp https://opentelemetry.io/docs/specs/otel/trace/api/#time, but python sdk uses nano seconds https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L946

I think the issue might not come from this line, I am asking if they can provide the start and end time of the span. The duration `18446744073700.02 ms` the user reported also doesn't make sense if you interrupt it as nano seconds (5124095 hours to 5 hours).

PR #3412, opened 2025-09-12
Removed OpenAI-Beta header as per the docs

MajorTal:
In the docs:
https://platform.openai.com/docs/guides/realtime#beta-to-ga-migration
For REST API requests, WebSocket connections, and other interfaces with the Realtime API, beta users had to include the following header with each request:
OpenAI-Beta: realtime=v1
This header should be removed for requests to the GA interface. To retain the behavior of the beta API, you should continue to include this header.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -649,7 +649,6 @@ class RealtimeSession(
                 headers["api-key"] = self._realtime_model._opts.api_key
         else:
             headers["Authorization"] = f"Bearer {self._realtime_model._opts.api_key}"
-            headers["OpenAI-Beta"] = "realtime=v1"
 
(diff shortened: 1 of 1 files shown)


davidzhao:
Thanks Tal!

longcw:
it needs more changes to migrate to GA version, e.g. importing from `openai.types.realtime` instead of `openai.types.beta.realtime` (paths of some events changed as well), some event names have changed
> response.text.delta → response.output_text.delta
response.audio.delta → response.output_audio.delta
response.audio_transcript.delta → response.output_audio_transcript.delta

I'll revert this change and create another with other needed migrations next week. right now it's broken with an error when updating the session
> 2025-09-14 11:20:21,698 - ERROR livekit.plugins.openai - OpenAI Realtime API returned an error {"error": "Error(message=\"Missing required parameter: 'session.type'.\", type='invalid_request_error', code='missing_required_parameter', event_id='session_update_0717c534dbb6', param='session.type')", "pid": 2599092, "job_id": "AJ_qNwx5FUC6a7j"}

PR #3417, opened 2025-09-12
add new room options

longcw:
the original `RoomInputOptions` and `RoomOutputOptions` are compatible but marked as deprecated.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -819,6 +819,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
 
             self.emit("close", CloseEvent(error=error, reason=reason))
 
+            self._cancel_user_away_timer()
             self._user_state = "listening"
             self._agent_state = "initializing"
             self._llm_error_counts = 0
@@ -1018,7 +1019,11 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
             self._agent = agent
 
             if new_activity == "start":
-                if agent._activity is not None:
+                previous_agent = self._activity.agent if self._activity else None
+                if agent._activity is not None and (
+                    # allow updating the same agent that is running
+                    agent is not previous_agent or previous_activity != "close"
+                ):
                     raise RuntimeError("cannot start agent: an activity is already running")
 
                 self._next_activity = AgentActivity(agent, self)
--- a/livekit-agents/livekit/agents/voice/room_io/types.py
+++ b/livekit-agents/livekit/agents/voice/room_io/types.py
@@ -52,6 +52,8 @@ class TextInputOptions:
 class AudioInputOptions:
     sample_rate: int = 24000
     num_channels: int = 1
+    frame_size_ms: int = 50
+    """The frame size in milliseconds for the audio input."""
     noise_cancellation: rtc.NoiseCancellationOptions | None = None
     pre_connect_audio: bool = True
     """Pre-connect audio enabled or not."""
@@ -175,6 +177,7 @@ class RoomOptions:
                 AudioInputOptions(
                     sample_rate=input_options.audio_sample_rate,
                     num_channels=input_options.audio_num_channels,
+                    frame_size_ms=input_options.audio_frame_size_ms,
                     noise_cancellation=input_options.noise_cancellation,
                     pre_connect_audio=input_options.pre_connect_audio,
                     pre_connect_audio_timeout=input_options.pre_connect_audio_timeout,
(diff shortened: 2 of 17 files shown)


examples/voice_agents/basic_agent.py
```diff
@@ -123,11 +121,12 @@ async def log_usage():
     await session.start(
         agent=MyAgent(),
         room=ctx.room,
-        room_input_options=RoomInputOptions(
-            # uncomment to enable Krisp BVC noise cancellation
-            # noise_cancellation=noise_cancellation.BVC(),
-        ),
-        room_output_options=RoomOutputOptions(transcription_enabled=True),
+        room_options={
+            "audio_input": {
```
tinalenguyen: what are your thoughts on importing room_io and then something like:

```
await session.start(
         agent=MyAgent(),
         room=ctx.room,
                  input=room_io.RoomInputOptions(...),
                  output=room_io.RoomOutputOptions(...),
         )
)
```
or even:

```
await session.start(
         agent=MyAgent(),
         room=ctx.room,
                  input= {
                           "text_enabled": False,
                           "video_input": True,
                   },
                   output= {
                            "transcription_enabled": False
                  }
         )
)
```
and in `RoomOptions`, group together input and output options

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -405,6 +405,7 @@ async def start(
         *,
         capture_run: Literal[True],
         room: NotGivenOr[rtc.Room] = NOT_GIVEN,
+        room_options: NotGivenOr[room_io.RoomOptions | dict[str, Any]] = NOT_GIVEN,
```
theomonnom: the tricky thing with dict[str, Any] is that we don't have any typing. So most IDE will not suggest any auto-completion

longcw: yeah I am not happy with this version, but it seems that there is no way to support both dataclasses/pydantic and typed dict without defining the options twice.

If we only support the first one, it will need to import a lot of xxxOptions to create the config...

livekit-agents/livekit/agents/voice/room_io/types.py
```diff
@@ -0,0 +1,266 @@
...
+class RoomOptions(_BaseOptions):
+    text_input: NotGivenOr[TextInputOptions | bool] = NOT_GIVEN
+    """The text input options. If not provided, default to True."""
+    audio_input: NotGivenOr[AudioInputOptions | bool] = NOT_GIVEN
+    """The audio input options. If not provided, default to True."""
+    video_input: NotGivenOr[VideoInputOptions | bool] = NOT_GIVEN
+    """The video input options. If not provided, default to False."""
+    audio_output: NotGivenOr[AudioOutputOptions | bool] = NOT_GIVEN
+    """The audio output options. If not provided, default to True."""
+    transcription_output: NotGivenOr[TranscriptionOutputOptions | bool] = NOT_GIVEN
```
davidzhao: should we call this text_output? since for the agent it's actually outputting LLM text.. 

transcription is a type of text.. and feels like this is controlling both

davidzhao (commented):
nice, seems much cleaner to organize this way.

livekit-agents/livekit/agents/voice/room_io/types.py
longcw: rename to text_output and `TextOutputOptions`

examples/voice_agents/basic_agent.py
longcw: yeah, I think the following makes sense
```python
    await session.start(
        agent=MyAgent(),
        room=ctx.room,
                # uncomment to enable the Krisp BVC noise cancellation
                # noise_cancellation=noise_cancellation.BVC(),
            ),
            audio_output=room_io.AudioOutputOptions(sample_rate=24000),
            text_output=room_io.TextOutputOptions(sync_transcription=True),
        ),
    )
```

livekit-agents/livekit/agents/voice/agent_session.py
longcw: removed the `dict` input.

livekit-agents/livekit/agents/voice/room_io/types.py
```diff
@@ -0,0 +1,266 @@
...
+                    pre_connect_audio=input_options.pre_connect_audio,
+                    pre_connect_audio_timeout=input_options.pre_connect_audio_timeout,
+                )
+                if input_options.audio_enabled is not False
+                else False
+            )
+
+            opts.video_input = input_options.video_enabled
+            opts.close_on_disconnect = input_options.close_on_disconnect
+            opts.delete_room_on_close = input_options.delete_room_on_close
```
chenghao-mou, line 193: Do we want to copy over `participant_kinds` and `participant_identity`?

theomonnom (approved):

theomonnom:
lgtm overall, it just feels a bit annoying to have to import all those different types

longcw:
> lgtm overall, it just feels a bit annoying to have to import all those different types

all the types are exposed via `livekit.agents.room_io` so users can use it like 
```python
from livekit.agents import room_io

...
            await session.start(
                room=ctx.room,
                        sample_rate=24000,
                        num_channels=1,
                    ),
                    close_on_disconnect=False,
                ),
            )

PR #3422, opened 2025-09-15
fix structured_output example when use_tts_aligned_transcript is enabled

longcw:


davidzhao (approved):

longcw:
replaced by https://github.com/livekit/agents/pull/3423

PR #3424, opened 2025-09-15
add transcription_filters as an option to AgentSession

longcw:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1470,6 +1470,7 @@ def _on_first_frame(_: asyncio.Future[None]) -> None:
                     node=self._agent.tts_node,
                     input=audio_source,
                     model_settings=model_settings,
+                    transcription_filters=self._session.options.transcription_filters,
```
theomonnom: I'm wondering if we can find a better name, it may not be clear that `transcription_filters` is also changing the TTS input.

longcw: yeah that's right. maybe `tts_node_filters`, `tts_input_filters`?

longcw: renamed to `tts_text_transforms`.

livekit-agents/livekit/agents/voice/transcription/filters.py
```diff
@@ -1,5 +1,26 @@
...
+    text: AsyncIterable[str], transforms: Sequence[TextTransforms]
+) -> AsyncIterable[str]:
+    all_transforms = {
+        "filter_markdown": filter_markdown,
+        "filter_emoji": filter_emoji,
+    }
+
+    for transform in transforms:
+        if transform not in all_transforms:
+            raise ValueError(
```
marctorsoc, line 18: @longcw so we cannot add any custom transform through text_transforms? Why?

longcw: this is for our built-in transforms right now, to add custom processor for text stream, you can customize the `tts_node` or `transcription_node` directly.

PR #3429, opened 2025-09-15
add shutdown method for AgentSession

tinalenguyen:
`shutdown()` closes and drains the session. it can also be used in function calls

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -639,11 +639,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         )
 
     def shutdown(self) -> None:
-        if self._closing_task:
-            return
-        self._closing_task = asyncio.create_task(
-            self._aclose_impl(error=None, drain=True, reason=CloseReason.USER_INITIATED)
-        )
+        self._close_soon(error=None, drain=True, reason=CloseReason.USER_INITIATED)
 
     @utils.log_exceptions(logger=logger)
     async def _aclose_impl(


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -625,18 +625,18 @@ async def drain(self) -> None:
 
         await self._activity.drain()
 
-    def _close_soon(
+    def shutdown(
```
theomonnom: I think we should create a separate method, so we don't expose the `drain` and the `reason` to users

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -625,18 +625,18 @@ async def drain(self) -> None:
...
         self,
         *,
-        reason: CloseReason,
+        reason: CloseReason = CloseReason.USER_INITIATED,
         drain: bool = False,
         error: llm.LLMError | stt.STTError | tts.TTSError | llm.RealtimeModelError | None = None,
+        close_room: bool = True,
```
theomonnom: I think this should be inside the RoomOptions instead ([here](https://github.com/livekit/agents/blob/5fd278c90ad89b25258ee4f5e210b3226aa47bf1/livekit-agents/livekit/agents/voice/room_io/room_io.py#L61))
A new argument like `delete_room_on_close`, ideally we merge RoomInputOptions and RoomOutputOptions together, but feel free to ignore for now

theomonnom (commented):
Otherwise lgtm!

livekit-agents/livekit/agents/voice/agent_session.py
tinalenguyen: i removed the logic for now, will it always be the case that the room should be deleted on agentsession close? i wonder if there are any specialized cases that need flexibility

i'll keep thinking on it but yeah this could be another PR

tinalenguyen: i guess for specialized cases, `delete_room_on_close` could be set to `False` and the user could do `get_job_ctx().delete_room()` when needed

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -634,11 +634,17 @@ def _close_soon(
...
         self._closing_task = asyncio.create_task(
             self._aclose_impl(error=error, drain=drain, reason=reason)
         )
 
+    def shutdown(self) -> None:
+        if self._closing_task:
+            return
+        self._closing_task = asyncio.create_task(
+            self._aclose_impl(error=None, drain=True, reason=CloseReason.USER_INITIATED)
+        )
```
theomonnom: nit:
```suggestion
    def shutdown(self) -> None:
        self._close_soon(error=None, drain=True, reason=CloseReason.USER_INITIATED)
```

theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_session.py
theomonnom: yeah exactly, if the users don't want that, they will put delete_room_on_close to False

Mokin711:
"ERROR calling session.shutdown(): 'AgentSession' object has no attribute 'shutdown' " is this a new version that added the shutdown function?

marctorsoc:
@theomonnom @longcw how is `session.shutdown()` different than `ctx.delete_room()`. Is the difference only (let's assume 1 user - 1 agent in a room only, no more participants) than the latter hangs up, whereas the former allows the user to stay alone in the room?

UPDATED: Sorry, I now saw [this](https://docs.livekit.io/agents/worker/job/#ending-the-session), explaining `ctx.shutdown` vs `ctx.delete_room`

examples/voice_agents/inactive_user.py
```diff
@@ -39,8 +39,7 @@ async def user_presence_task():
             )
             await asyncio.sleep(10)
 
-        await asyncio.shield(session.aclose())
-        ctx.delete_room()
+        session.shutdown()
```
marctorsoc, line 42: @tinalenguyen I'm not sure this change is equivalent. See my comment in the PR

longcw:
> "ERROR calling session.shutdown(): 'AgentSession' object has no attribute 'shutdown' " is this a new version that added the shutdown function?

yes, it will be included in next release.

PR #3432, opened 2025-09-16
expose LocalInferenceRunner

longcw:


livekit-agents/livekit/agents/inference_runner.py
```diff
@@ -9,15 +9,14 @@ class _RunnerMeta(Protocol):
     INFERENCE_METHOD: ClassVar[str]
 
 
-_RunnersDict = dict[str, type["_InferenceRunner"]]
+_RunnersDict = dict[str, type["LocalInferenceRunner"]]
 
 
-# kept private until we stabilize the API (only used for EOU today)
-class _InferenceRunner(ABC, _RunnerMeta):
+class LocalInferenceRunner(ABC, _RunnerMeta):
```
longcw, line 15: anything else we need to do except renaming? @theomonnom

I think it would be good to have an example to use the inference runner, any recommendations? whisper.cpp needs some custom building seems not good for an example.

PR #3436, opened 2025-09-16
docs(CONTRIBUTING.md): add linting/formatting command guideline

bml1g12:
It wasn't immediately clear to me as a contibutor what commands to run for linting/typechecking/formatting without inspecting what the CI scripts were doing, so thought it would be nice to make it more explicit. In particular, even after reading the CI scripts, I didn't notice you need to run a `uv pip install pip` before the `uv run mypy` will work, so I hope this will help other contributors get up and running faster.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,7 @@ uv pip install pip && uv run mypy --install-types --non-interactive \
 ### Linting
 
 ```bash
-uv run ruff check .
+uv run ruff check --output-format=github .
 ```
 
 ### Formating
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -45,3 +45,61 @@ be approved:
...
+    -p livekit.plugins.sarvam \
+    -p livekit.plugins.inworld \
+    -p livekit.plugins.simli \
+    -p livekit.plugins.anam
+```
+
+### Linting
+
+```bash
+uv run ruff check .
```
longcw: keep consistency with the ci
```suggestion
uv run ruff check --output-format=github .
```

PR #3444, opened 2025-09-17
fix(openai-realtime): nest session.update audio fields; omit null tur…

MajorTal:
…n_detection

OpenAI Realtime v1 expects audio-related options to be nested under `session.audio`.
Sending them at the top level (e.g., `session.turn_detection`, `session.voice`,
`session.input_audio_transcription`) results in `unknown_parameter` errors.
Additionally, `turn_detection.create_response: null` is rejected with `invalid_type`.

What’s changed
- RealtimeSession.update_options:
  - voice → session.audio.output.voice
  - speed → session.audio.output.speed
  - turn_detection → session.audio.input.turn_detection
  - input_audio_transcription → session.audio.input.transcription
  - input_audio_noise_reduction → session.audio.input.noise_reduction
  - Only include `audio` in the payload when at least one of the above is set.
- utils.to_turn_detection:
  - Build kwargs conditionally (only include fields when not None), so
    `create_response` is never serialized as `null`.

Why
- Fixes these runtime errors from OpenAI Realtime:
  - Unknown parameter: 'session.turn_detection'
  - Unknown parameter: 'session.voice'
  - Unknown parameter: 'session.input_audio_transcription'
  - Unknown parameter: 'session.input_audio_noise_reduction'
  - Invalid type for 'session.audio.input.turn_detection.create_response': expected boolean, got null

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1051,13 +1051,21 @@ class RealtimeSession(
             self._realtime_model._opts.tool_choice = tool_choice
             kwargs["tool_choice"] = to_oai_tool_choice(tool_choice)
 
+        audio_cfg: RealtimeAudioConfig | None = None
+
         if is_given(voice):
             self._realtime_model._opts.voice = voice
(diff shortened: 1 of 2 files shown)


davidzhao:
I've updated it to use explicit types rather than kwargs in #3448. going to merge that one instead. thanks for the PR!

PR #3447, opened 2025-09-18
fix: send interrupt only when server vad not enabled or forced

sebitokazu:
## Description

When we're using a Realtime Model with server side VAD, the server is responsible of generating `input_speech_started` events. In that cases, we don't want to send a `cancel` event to the model as we would be self interrupting ourselves.

Example of current flow:
- We receive an `input_speech_started` event ([see handle](https://github.com/livekit/agents/blob/06f5321bba0eb90863396df72eb4972d70ebf86a/livekit-agents/livekit/agents/voice/agent_activity.py#L1013-L1024))
- Since we use server interruption, we allow_interruptions can't be false
- Then in `interrupt` we would be sending an interruption to the server (from the server's own event)

Appreciate any comment or possible scenario where this behaviour isn't expected.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -868,7 +868,9 @@ class AgentActivity(RecognitionHooks):
             speech.interrupt(force=force)
             interrupted_speeches.append(speech)
 
-        if self._rt_session is not None:
+        if self._rt_session is not None and (
+            not self._rt_session.realtime_model.capabilities.turn_detection or force
+        ):
             self._rt_session.interrupt()
 
         if not interrupted_speeches:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -868,7 +868,9 @@ def interrupt(self, *, force: bool = False) -> asyncio.Future[None]:
             speech.interrupt(force=force)
             interrupted_speeches.append(speech)
 
-        if self._rt_session is not None:
+        if self._rt_session is not None and (
```
longcw: maybe add an option to skip the rt_session to the `AgentActivity.interrupt()` method and set it in `_on_input_speech_started`. this method is also used by other interruption logic, it should interrupt the rt_session by default.

sebitokazu: added `omit_rt_interrupt` param here ([8f0fd0b](https://github.com/livekit/agents/pull/3447/commits/8f0fd0b5c42dae72ce9f93bb5ce0ecd276a3ebed))

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -854,7 +854,9 @@ def generate_reply(
 
         return handle
 
-    def interrupt(self, *, force: bool = False) -> asyncio.Future[None]:
+    def interrupt(
+        self, *, force: bool = False, omit_rt_interrupt: bool = False
+    ) -> asyncio.Future[None]:
```
longcw, line 859:
```suggestion
    def interrupt(self, *, force: bool = False) -> asyncio.Future[None]:
```

I don't think we need to expose this to AgentSession.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -864,7 +866,7 @@ def interrupt(self, *, force: bool = False) -> asyncio.Future[None]:
         if self._activity is None:
             raise RuntimeError("AgentSession isn't running")
 
-        return self._activity.interrupt(force=force)
+        return self._activity.interrupt(force=force, omit_rt_interrupt=omit_rt_interrupt)
```
longcw, line 869:
```suggestion
        return self._activity.interrupt(force=force)
```

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1017,7 +1019,9 @@ def _on_input_speech_started(self, _: llm.InputSpeechStartedEvent) -> None:
         # self.interrupt() is going to raise when allow_interruptions is False, llm.InputSpeechStartedEvent is only fired by the server when the turn_detection is enabled.  # noqa: E501
         # When using the server-side turn_detection, we don't allow allow_interruptions to be False.
         try:
-            self.interrupt()  # input_speech_started is also interrupting on the serverside realtime session  # noqa: E501
+            self.interrupt(
+                omit_rt_interrupt=True
+            )  # input_speech_started is also interrupting on the serverside realtime session  # noqa: E501
```
longcw, line 1024: we should update the comment here

longcw (commented):
left some comments, otherwise looks good to me. cc @theomonnom do you see any risk of doing this?

Issue #3449, opened 2025-09-18
Getting Connection error and Retry error in Azure Realtime api


akashicMarga:
# LiveKit Agents: NotGiven Serialization Error with Pydantic 2.8+ - Need Guidance

## Issue Summary

I'm new to LiveKit and encountered an error when trying to use Azure OpenAI Realtime API with LiveKit agents. The error occurs with Pydantic versions 2.8+ and prevents the agent from connecting.

## Environment

- **LiveKit Agents Version**: 1.2.9
- **Pydantic Version**: 2.8.2, 2.9.2, 2.11.9 (all affected)
- **Python Version**: 3.13
- **Platform**: macOS
- **OpenAI Provider**: Azure OpenAI Realtime API

## Error Details

```
AttributeError: 'NotGiven' object has no attribute '__pydantic_fields_set__'
  File "/path/to/livekit/plugins/openai/realtime/realtime_model_beta.py", line 583, in _reconnect
    msg = ev.model_dump(by_alias=True, exclude_unset=True, exclude_defaults=False)
```

The error happens when LiveKit tries to serialize events containing `NotGiven` objects using Pydantic's `model_dump()` method. The `NotGiven` class doesn't have the attributes that Pydantic 2.8+ expects.

## My Solution (Working but Need Validation)

I fixed it by modifying the error handling in `realtime_model_beta.py` around line 583:

```python
for ev in events:
    try:
        msg = ev.model_dump(by_alias=True, exclude_unset=True, exclude_defaults=False)
    except AttributeError as e:
        if "'NotGiven' object has no attribute" in str(e):
            try:
                msg = ev.model_dump(by_alias=True, exclude_unset=True, exclude_defaults=False, exclude_none=True)
            except AttributeError:
                # Manual serialization when Pydantic completely fails
                msg = {}
                for field_name, field_value in ev.__dict__.items():
                    if hasattr(field_value, '__class__') and 'NotGiven' in str(field_value.__class__):
                        continue  # Skip NotGiven fields
                    elif hasattr(field_value, 'model_dump'):
                        try:
                            msg[field_name] = field_value.model_dump(by_alias=True, exclude_unset=True, exclude_defaults=False, exclude_none=True)
                        except AttributeError:
                            msg[field_name] = str(field_value)
                    else:
                        msg[field_name] = field_value
        else:
            raise e
```

This solution works - my agent now connects successfully without the Pydantic error.

## Questions

[...]

PR #3451, opened 2025-09-18
fix openai realtime with azure when input_audio_noise_reduction is not given

longcw:
fix https://github.com/livekit/agents/issues/3449


davidzhao (approved):
lgtm! thanks for the fix

PR #3457, opened 2025-09-18
Bug fix: Safe handling in ChatContext.truncate() when fewer items than max_items

adianliusie:
This PR resolves a bug in the ChatContext.truncate() method where the system message could be duplicated when truncation was applied on a context with fewer items than max_items. The fix is simple and quite self explanatory.


theomonnom (approved):

theomonnom:
thanks

PR #3467, opened 2025-09-19
add delete_room_on_close to RoomInputOptions

tinalenguyen:
when the `AgentSession` closes, the room is deleted by default

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/room_io.py
+++ b/livekit-agents/livekit/agents/voice/room_io/room_io.py
@@ -485,5 +485,6 @@ class RoomIO:
         self._update_state_atask = asyncio.create_task(_set_state())
 
     def _on_agent_session_close(self, ev: CloseEvent) -> None:
-        job_ctx = get_job_context()
-        self._delete_room_task = job_ctx.delete_room()
+        if self._input_options.delete_room_on_close:
+            job_ctx = get_job_context()
+            self._delete_room_task = job_ctx.delete_room()
--- a/livekit-agents/livekit/agents/voice/room_io/room_io.py
+++ b/livekit-agents/livekit/agents/voice/room_io/room_io.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any, Callable, Optional
 from livekit import rtc
 
 from ... import utils
+from ...job import get_job_context
 from ...log import logger
 from ...types import (
     ATTRIBUTE_AGENT_STATE,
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -146,6 +149,7 @@ def __init__(
         self._tasks: set[asyncio.Task[Any]] = set()
         self._update_state_atask: asyncio.Task[None] | None = None
         self._close_session_atask: asyncio.Task[None] | None = None
```
tinalenguyen, line 151: i don't see `self._close_session_atask` ever being set, was its purpose supposed to be similar to `self._delete_room_task`?

livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -478,3 +483,7 @@ async def _set_state() -> None:
             self._update_state_atask.cancel()
 
         self._update_state_atask = asyncio.create_task(_set_state())
+
+    def _on_agent_session_close(self, ev: CloseEvent) -> None:
+        job_ctx = get_job_context()
```
theomonnom: Let's make sure `delete_room_on_close` is True here

livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -478,3 +483,8 @@ async def _set_state() -> None:
             self._update_state_atask.cancel()
 
         self._update_state_atask = asyncio.create_task(_set_state())
+
+    def _on_agent_session_close(self, ev: CloseEvent) -> None:
+        if self._input_options.delete_room_on_close:
+            job_ctx = get_job_context()
+            self._delete_room_task = job_ctx.delete_room()
```
theomonnom, line 490: We're already storing the task internally here:
https://github.com/livekit/agents/blob/4bb8407f48668ec096d55d9059ad4ddeb8fbc6f5/livekit-agents/livekit/agents/job.py#L295

So we don't necessarily need to store it here

tinalenguyen: @theomonnom I stored the task to check it [here](https://github.com/livekit/agents/blob/tina/delete-room-on-close/livekit-agents/livekit/agents/voice/room_io/room_io.py#L433) so `_on_participant_disconnected(`) doesn't try to close the agentsession when it is already closed

let me know if there's a better way to go about that

PR #3481, opened 2025-09-23
inference: support lang/ for stt and update models

longcw:


livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -62,8 +66,22 @@ class AssemblyaiOptions(TypedDict, total=False):
     keyterms_prompt: list[str]  # default: not specified
 
 
-STTModels = Union[DeepgramModels, CartesiaModels, AssemblyaiModels]
-STTLanguages = Literal["en", "de", "es", "fr", "ja", "pt", "zh"]
+STTLanguages = Literal["multi", "en", "de", "es", "fr", "ja", "pt", "zh", "hi"]
+_LanguageAsModels = Literal[
```
davidzhao, line 70: @adrian-cowham to confirm. if we passed in `multi`, which model/language does it get matched to?

longcw: btw, do we support to specify a list of languages?

adrian-cowham: @davidzhao Yes, the API accepts `multi` as a language and the Gateway will choose `deepgram/nova-3`.
`nova-3` supports these languages:
```
English, Spanish, French, German, Hindi, Russian, Portuguese, Japanese, Italian, Dutch, Turkish, Norwegian, and Indonesian
```
https://developers.deepgram.com/docs/models-languages-overview#nova-3

Here is some code I was using for testing:
```
    def __init__(self):
        super().__init__(
            instructions=_instructions_haley,
            stt=inference.STT(model="", language="multi"),
            llm="azure/gpt-4o-mini",
            tts="elevenlabs:cgSgspJ2msm6clMCkdW9",
        )
```

adrian-cowham: @longcw We do not support a list of languages, but we can. To the best of my knowledge, `nova-3` (and `nova-2`) are the only models that simultaneously supports multiple languages. The Cartesia and Assembly APIs only allow you to specify a single language.

PR #3490, opened 2025-09-23
brianyin/agt-1946-built-in-dtmf-support

toubatbrian:


livekit-agents/livekit/agents/beta/tools/dtmf.py
```diff
@@ -0,0 +1,80 @@
+from enum import Enum
+
+from livekit.agents.job import get_job_context
+from livekit.agents.llm import function_tool
+
+
+class DtmfEvent(str, Enum):
```
toubatbrian, line 7: use enum to have strict tool schema instead of raw string

livekit-agents/livekit/agents/beta/tools/dtmf.py
```diff
@@ -0,0 +1,80 @@
...
+        return 11
+    elif event.value in ["A", "B", "C", "D"]:
+        # DTMF codes 10-15 are used for letters A-D
+        return ord(event.value) - ord("A") + 12
+    else:
+        raise ValueError(f"Invalid DTMF event: {event}")
+
+
+@function_tool
+async def send_dtmf_events(
```
toubatbrian, line 41: For now I'll wait for `ToolSet` class: https://github.com/livekit/agents/pull/3296/files#diff-f443ca72981c01a0780b2742eef4232b4a313c99d36f462900034b138df2c016. And convert those function_tool into `ToolSet` class once done.


But the general idea is that we expose 2 tools in the toolsets, one allows pressing a single digit and the other one allows pressing a sequence (which may be a usage pattern so agent only need one tool call to handle multiple dtmf events)

livekit-agents/livekit/agents/beta/tools/dtmf.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Do not call when:
+    - User only wants to send a single DTMF event
+    """
+    job_ctx = get_job_context()
+
+    for event in events:
+        try:
+            code = dtmf_event_to_code(event)
+            await job_ctx.room.local_participant.publish_dtmf(code=code, digit=event.value)
```
toubatbrian, line 55: For sending a sequence, not sure if we want to add a sleep between each. @theomonnom

examples/voice_agents/dtmf_agent.py
```diff
@@ -0,0 +1,81 @@
...
+load_dotenv()
+
+
+DTMF_AGENT_DISPATCH_NAME = os.getenv("DTMF_AGENT_DISPATCH_NAME", "my-telephony-agent")
+
+
+class DtmfAgent(Agent):
+    def __init__(self) -> None:
+        super().__init__(
+            instructions="You are a voice assistant that can press number on the phone to interact with the user or a IVR system.",
```
toubatbrian, line 32: Will improving the system message

livekit-agents/livekit/agents/beta/tools/dtmf.py
theomonnom: Do we need 2 tools? The LLM could also just use a single element list?

livekit-agents/livekit/agents/beta/tools/dtmf.py
theomonnom: No I don’t think we need to sleep

livekit-agents/livekit/agents/beta/tools/dtmf.py
toubatbrian: yeah, I think that also work

toubatbrian:
See https://github.com/livekit/agents/pull/3547

PR #3494, opened 2025-09-24
brianyin/agt-1947-automatically-parse-dtmf-input-from-users

toubatbrian:


livekit-agents/livekit/agents/beta/tools/dtmf.py
```diff
@@ -57,3 +62,83 @@ async def send_dtmf_events(
             return f"Failed to send DTMF event: {event.value}. Error: {str(e)}"
 
     return f"Successfully sent DTMF events: {', '.join(events)}"
+
+
+async def collect_dtmf_inputs(
```
toubatbrian, line 71: @theomonnom @longcw  for receiving dtmf from user, which one do you think is better:

**Option 1**: give LLM a tool called `collect_dtmf_inputs` like it is defined here, and LLM choose to call the tool which wait for specified number of digits with timeout/terminator key.

I have tested this approach, it's really hard to make LLM explicitly call this tool. It usually just generate text "Please enter phone number" without followed by tool call, unless user explicitly mention it and prompt it.

**Option 2**: just expose a general utility `collect_dtmf_inputs`, and user can use it e.g. in a Task to collect particular information and wait for it.

Pro is that it gives user more flexibility to design their agent while hiding certain level of abstraction. Cons is that it is not part of the agentic loop, the function has to be written programmcally usually with combination of a Task to collect info.

Option 3: Instead of conditionally listen to dtmf_received room event, we create a utility that listen to it all time, define certain timeout / terminator key so that when a sequence of number been pressed and timeout reaches (or terminator key been pressed at the end) we append this as a message to agent but not trigger it to generate a reply. 

Usage:
[llm] please provide your phone number
[user] enter 1 2 3  4 5 6 7 8 9 0 # --> `chat_ctx.add_message("User has typed 123-456-7890")`
[user] Ok I have entered my info
[llm] Ok I see you have typed ......, just to confirm is this correct?
[user] yes
[llm] tool_execution save_phone_number("1234567890")

Pros is this does not require any tool call during dtmf input collection stage, just modify the chat context. But cons is it's hard to define the start and end of a dtmf segment. For example, the following edge case:
```
[llm] please provide your phone number
[user] enter 1 2 4
[user] oh sorry there's a typo let me re-enter
[user] 1 2 3 4 5 6 7 8 9 0 
```
In this case we need to be really careful of constructing our message history so that two sequences of dtmf inputs are well separated by the middle user message.

PR #3497, opened 2025-09-24
return RealtimeModelBeta for Azure OpenAI Realtime

tinalenguyen:
azure does not support GA API


theomonnom (approved):

bruno-wlodarczyk:
Are there any updates in the case? As this is in GA https://learn.microsoft.com/en-us/azure/ai-services/speech-service/voice-live

PR #3499, opened 2025-09-24
add google stt chirp_3

tinalenguyen:
note that Chirp_3 does not support word-level confidence scores but supports diarization

[source](https://cloud.google.com/speech-to-text/v2/docs/chirp_3-model#transcribe_using_chirp_3_batch_request_with_diarization)

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/models.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/models.py
@@ -10,6 +10,7 @@ SpeechModels = Literal[
     "medical_conversation",
     "chirp",
     "chirp_2",
+    "chirp_3",
     "latest_long",
     "latest_short",
 ]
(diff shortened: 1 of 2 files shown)


theomonnom:
I think we should handle that inside `_streaming_recognize_response_to_speech_data`

tinalenguyen:
@theomonnom How would we check if the model is Chirp 3 in `_streaming_recognize_response_to_speech_data`? Would another argument be passed for that?

theomonnom:
> @theomonnom How would we check if the model is Chirp 3 in `_streaming_recognize_response_to_speech_data`? Would another argument be passed for that?

yes we can add another argument

mrkowalski:
Hi @theomonnom, are you going to continue with this PR? I have just stumbled upon this issue and I am happy to pick it up.

mrkowalski:
Hi @theomonnom, @tinalenguyen. chirp_3 does not return transcript confidences. The only change that is correct in this PR is addition of `chirp_3` in models.py. The rest is invalid.

I do not think there should be any separate logic for chirp_3 as the model is using the same API as chirp_2 and others. A user may set `min_confidence_threshold=0.0` on `livekit.plugins.google.STT` to make chirp_3 work without code changes - addition of chirp_3 to models.py helps only with type checking.

davidzhao:
@mrkowalski feel free to take a stab at this. we'd be happy to merge your version instead.

mrkowalski:
> @mrkowalski feel free to take a stab at this. we'd be happy to merge your version instead.

https://github.com/livekit/agents/pull/3628

PR #3511, opened 2025-09-26
realtime model: wait for generate_reply before update tool results

longcw:


longcw:
Tested with different realtime api providers:
1. OpenAI (auto_tool_reply_generation=False): works as expected
2. Gmini (auto_tool_reply_generation=True): works when `tool_behavior="NON_BLOCKING"`
3. AWS: `generate_reply` is not supported
4. Ultravox (auto_tool_reply_generation=True): doesn't work, it has a timeout for tool outputs, sending the tool output to the model after `generate_reply` cannot create tool response.

I am wondering should we just tell users to use `await session.generate_reply()` in tool call for the supported models directly? adding a wait in the framework doesn't actually work for all models. cc @theomonnom @toubatbrian  @Shubhrakanti

Shubhrakanti:
maybe there's a way to do this just for the open ai realtime model for now?

Shubhrakanti:
> maybe there's a way to do this just for the open ai realtime model for now?

we could wait for the previous generation before sending the `generation_reply` in the open ai realtime implementation. But this is good enough for now.

PR #3512, opened 2025-09-26
brian/receive-dtmf-input

toubatbrian:


livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,175 @@
...
+import logging
+from dataclasses import dataclass
+
+from livekit import rtc
+from livekit.agents.beta.tools.dtmf import DtmfEvent, format_dtmf
+from livekit.agents.job import get_job_context
+from livekit.agents.llm.chat_context import ChatContext
+from livekit.agents.llm.tool_context import function_tool
+from livekit.agents.types import NOT_GIVEN, NotGivenOr
+from livekit.agents.utils.aio.debounce import debounced
```
theomonnom: We must use `relative` import when inside the agents framework

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class DtmfInputsTask(AgentTask[list[DtmfEvent]]):
+    def __init__(
+        self,
+        input_config: SingleDigitConfig | MultiDigitConfig,
+        chat_ctx: NotGivenOr[ChatContext] = NOT_GIVEN,
+        input_timeout: float = 5.0,
+        interrupt_on_dtmf_sent: bool = False,
+    ) -> None:
```
theomonnom, line 45: I think we could design a better API without having to import more dataclasses?
We could use @overload too in case we need very different argument depending on single-digit vs multi-digit.

In an ideal word, you use have to use GetDtmfTask alone

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def format_instructions(self) -> str:
+        return MULTI_DIGIT_INSTRUCTIONS.format(
+            name=self.name,
+            num_digits=self.num_digits,
+            description=self.description,
+        )
+
+
+class DtmfInputsTask(AgentTask[list[DtmfEvent]]):
```
theomonnom: Thinking about: 
```suggestion
class GetDtmfTask(AgentTask[list[DtmfEvent]]):
```

theomonnom: I think a list of DtmfEvent is hard to parse for the users. Maybe just returning a string is easier?

examples/dtmf/dtmf_agent.py
```diff
@@ -0,0 +1,120 @@
...
+        choices = {
+            DtmfEvent.ONE: "Check credit card balance",
+            DtmfEvent.TWO: "Check recent transactions",
+            DtmfEvent.THREE: "Connect to a live agent",
+        }
+
+        result = await DtmfInputsTask(
+            input_config=SingleDigitConfig(choices=choices),
+            interrupt_on_dtmf_sent=True,
+        )
```
theomonnom: I don't think we need to constrain users to use a specific format.

I think something like this is sufficient:
```suggestion
        result = await DtmfInputsTask(
            extra_instructions="Press 1 to Check credit card balance, Press 2 to ...",
            interrupt_on_dtmf_sent=True,
        )
```

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@dataclass
+class MultiDigitConfig:
+    name: str
+    num_digits: int
+    description: str
+
+    def validate(self, inputs: list[DtmfEvent]) -> bool:
+        return len(inputs) == self.num_digits
```
theomonnom: Let's completely ignore validation for this Task.

I think this should be done by the user, outside of this AgentTask.
Like this pattern I sent on Slack

https://live-kit.slack.com/archives/C07TWVC0W4A/p1758826767640969?thread_ts=1758723237.629889&cid=C07TWVC0W4A

examples/dtmf/dtmf_agent.py
toubatbrian: > My take is there aren't really too many formats. From my understanding there are only 2, either press single digits but can have multiple choices, or press a sequence of digits for a specific info. I would still think it's helpful to we provide a built-in template for those formats. 

Please ignore above

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
toubatbrian: > In an ideal word, you use have to use GetDtmfTask alone

I can think of one benefit of having a Config class, basically for any kind of input format, it requires a validator, a format instruction, and a stop condition for which we call generate_reply immediately without having to wait for more digits.

If we can make the config to be a protocol having those methods, user can leverage that and even define their own custom format without the need to extending the GetDtmfTask or override its instructions. The `SingleDigitConfig` and `MultiDigitConfig` are examples (we don't event need them in our codebase actually, we can just let user to define the config protocol). I think this could make `GetDtmfTask` very flexible and adaptive to different format by **composing** this config protocol.

```py
class DtmfInputProtocol(Protocol):
    def format_instruction(self) -> str: ...

    def validate(self, inputs: list[DtmfEvent]) -> bool: ...
    
    def stop_reached(self, inputs: list[DtmfEvent]) -> bool: ...
```

examples/dtmf/dtmf_agent.py
toubatbrian: Also ^ I explained the benefit of potentially having a config Protocol class to bundle validation, instruction, stop_condition logics, which allows user to define their desired format.

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
toubatbrian: Again, ^ I think having a config Protocol would be a right balance of making GetDtmfTask flexible enough while abstracting certain common logics such as output validation:

```py
class DtmfInputProtocol(Protocol):
    def format_instruction(self) -> str: ...

    def validate(self, inputs: list[DtmfEvent]) -> bool: ...
    
    def stop_reached(self, inputs: list[DtmfEvent]) -> bool: ...
```

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,175 @@
...
+from livekit.agents.types import NOT_GIVEN, NotGivenOr
+from livekit.agents.utils.aio.debounce import debounced
+
+from ...voice.agent import AgentTask
+
+logger = logging.getLogger("dtmf-inputs")
+
+
+SINGLE_DIGIT_INSTRUCTIONS = """
+You are a voice assistant that can collect DTMF inputs from the user on a phone call.
```
theomonnom: I think we shouldn't mention DTMF in the prompt? I'm not sure, I'm just thinking that from a user/caller perspective, they may have no idea what it is

toubatbrian: yeah, the prompt is still tentative, open to any modifications

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
toubatbrian: To give an example, let's say user can define a hypothetical format such that, when the last 4 digits typed are 1 2 3 4, then we accept the input but need its length to be < 10:
```py
class HypotheticalInputFormat(DtmfInputProtocol):
    def format_instruction(self) -> str: 
        return "You are a xxxxx, input should end with 1 2 3 4 and length < 10"

    def validate(self, inputs: list[DtmfEvent]) -> bool: 
        return len(inputs) < 10 and inputs[-4:] == ["1", "2", "3", "4"]

    def stop_reached(self, inputs: list[DtmfEvent]) -> bool: ...
        return inputs[-4:] == ["1", "2", "3", "4"]
```

and use it

```py
format = HypotheticalInputFormat()

result = await GetDtmfTask(
    input_format=format
    ...
)
```

For input that can be ended with terminator key, here's an example format user can define:
```py
@dataclass
class TerminatableFormat(DtmfInputProtocol):
    terminator_key: DtmfEvent

    def format_instruction(self) -> str: 
        return "You are a xxxxx"

    def validate(self, inputs: list[DtmfEvent]) -> bool: 
        return ...

    def stop_reached(self, inputs: list[DtmfEvent]) -> bool: ...
        return len(inputs) > 0 and inputs[-1] == self.terminator_key

format = TerminatableFormat(terminator_key=DtmfEvent.POUND)

result = await GetDtmfTask(
    input_format=format
    ...
)
```

@theomonnom what do you think about this API?

theomonnom: I don't think this task must be that advanced for validation, I think num_digits is probably enough.
With the following example, it is already very easy for the users to validate the input in a loop for e.g
```python
@function_tool
async def find_my_order(self) -> None:
    self.session.say("Can you provide your order ID by typing it", allow_interruptions=False)
    while True:
        order_id = await GetDTMFTask(num_digits=10, ask_for_confirmation=True)
    
        order = self.db.lookup(order_id) 
        if order is not None:
            break
    
    return f"The order status is {order.status}, it is going to arrive in ..." 
```

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
theomonnom: See https://github.com/livekit/agents/pull/3512#discussion_r2383181998

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
toubatbrian: I think validation is needed since it's also part of the conventional IVR system, TwilloML have the validation mechanism as part of their programmable workflow. My main thought is we can't let LLM alone to handle and judge input format correctness (also LLM are pretty bad dealing with long sequence of numbers, could likely induce hallucinations).

toubatbrian: > I think num_digits is probably enough.

With the config protocol setup, users can customize format validation rules—whether that's a simple num_digits check or more complex custom logic, depending on their needs.

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,149 @@
...
+        else:
+            self._generate_dtmf_reply.schedule()
+
+    async def on_enter(self) -> None:
+        ctx = get_job_context()
+
+        ctx.room.on("sip_dtmf_received", self._on_sip_dtmf_received)
+        self.session.on("agent_state_changed", self._on_user_state_changed)
+        self.session.on("agent_state_changed", self._on_agent_state_changed)
+        self.session.generate_reply()
```
toubatbrian, line 177: @theomonnom I'm not sure if we want to hardcode a generate reply inside `on_enter` or we let user define this outside of the Task?

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,150 @@
...
+                "</dtmf_inputs>"
+            )
+
+            await self.session.generate_reply(instructions=instructions)
+            self._curr_dtmf_inputs = []
+
+        def _on_sip_dtmf_received(ev: rtc.SipDTMF) -> None:
+            if self.received_full_digits():
+                # temporarily pause DTMF receive to prevent new DTMF inputs interrupts the reply generation
+                return
```
toubatbrian: @theomonnom would love to know your thoughts here

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
theomonnom: Yes, it's fine to do it inside on_enter

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,168 @@
...
+    def __init__(
+        self,
+        name: str,
+        num_digits: int,
+        *,
+        chat_ctx: NotGivenOr[ChatContext] = NOT_GIVEN,
+        input_timeout: float = 5.0,
+        ask_for_confirmation: bool = False,
+        interrupt_on_complete_input: bool = False,
+    ) -> None:
```
theomonnom, line 45:
```suggestion
    def __init__(
        self,
        *,
        num_digits: int,
        extra_instructions: str.
        chat_ctx: NotGivenOr[ChatContext] = NOT_GIVEN,
        ask_for_confirmation: bool = False,
        dtmf_input_timeout: float = 5.0,
    ) -> None:
```
Let's always interrupt, no need for `interrupt_on_complete_input` IMO

We can replace `name` with `extra_instructions`, it's more flexible

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,168 @@
...
+from ...llm.chat_context import ChatContext
+from ...types import NOT_GIVEN, NotGivenOr
+from ...utils.aio.debounce import Debounced, debounced
+from ...voice.agent import AgentTask
+from ..tools.dtmf import DtmfEvent, format_dtmf
+
+logger = logging.getLogger("dtmf-inputs")
+
+
+class GetDtmfTask(AgentTask[Optional[str]]):
```
theomonnom: Can we create a Result dataclass like:
https://github.com/livekit/agents/blob/a4ad0dba2137e1b972dc97b38c41502817aa97cb/livekit-agents/livekit/agents/beta/workflows/email_address.py#L23

Also in case we have an error, let's raise ToolError (`AgentTask.complete` also accepts Error as the complete value) instead of None.

examples/dtmf/dtmf_agent.py
```diff
@@ -0,0 +1,109 @@
...
+                result = await GetDtmfTask(
+                    num_digits=10,
+                    chat_ctx=self.chat_ctx.copy(
+                        exclude_instructions=True, exclude_function_call=True
+                    ),
+                    ask_for_confirmation=True,
+                    extra_instructions="Ask user to provide a phone number in the format of +1234567890",
+                )
+            except ToolError as e:
+                self.session.generate_reply(instructions=e.message, allow_interruptions=False)
```
toubatbrian: `allow_interruptions=False` this should prevent tool been interrupted by any user speech

livekit-agents/livekit/agents/beta/workflows/utils.py
theomonnom, line 1: Can we move everything from this file inside the workflow itself?

toubatbrian: done

toubatbrian:
Merging it for now, will update the branch if anything new comes up.

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -0,0 +1,185 @@
...
+            # immediately kick off the DTMF reply generation if matches the stop event
+            if ev.digit == dtmf_stop_event.value:
+                self._generate_dtmf_reply()
+                return
+
+            self._curr_dtmf_inputs.append(DtmfEvent(ev.digit))
+            logger.info(f"DTMF inputs: {format_dtmf(self._curr_dtmf_inputs)}")
+            self._generate_dtmf_reply.schedule()
+
+        @debounced(delay=dtmf_input_timeout)
```
theomonnom, line 99: How can we make sure the task is always "cancelled/closed" ? 
Like a close method for e.g?

toubatbrian: In `on_exit` I have:
```py
    async def on_exit(self) -> None:
        ctx = get_job_context()

        ctx.room.off("sip_dtmf_received", self._on_sip_dtmf_received)
        self.session.off("agent_state_changed", self._on_user_state_changed)
        self.session.off("agent_state_changed", self._on_agent_state_changed)
        self._generate_dtmf_reply.cancel()

```

Issue #3407, opened 2025-09-11
Message paired with tool-call gets lost


pedroedgex:
I've been using livekit for the past couple of weeks and have noticed a strange behavior: Sometimes the model will repeat a message (or something similar to it) after stating it.

When I started to investigate (using Portkey to evaluate traces), I realized that whenever a text message gets sent along-side a tool call, livekit will not register the text message in the chat context for the next round, causing the model to, sometimes, repeat the same message, given that it don't see it in its context.

Here is an example of what I saying, the image below is a message generated by GPT-4o in a STT-LLM-TTS framework:

<img width="962" height="380" alt="Image" src="https://github.com/user-attachments/assets/6dda86f7-162b-4a5c-b41e-0ca64340d2c6" />

<img width="1365" height="511" alt="Image" src="https://github.com/user-attachments/assets/cfd2f94a-b948-4489-8715-177593f65574" />

On the next turn, that same message only have the tool call registered to it:

<img width="971" height="649" alt="Image" src="https://github.com/user-attachments/assets/97646938-e2b3-4ab8-97e3-54aaa11c2bbd" />

That causes problems on how the model generated the message for this turn.

I am using Python 3.12, livekit-agents and livekit-plugins-openai version 1.2.8, as well as portkey-ai==1.14.0

I believe the error is likely located within the _pipeline_reply_task for the Agent class, since I observed this for multiple models and with or without portkey (though I did not collect the evidence to add here)

furious-luke:
Hey @pedroedgex ! I believe I've tracked down the source of this issue. I've raised a PR here https://github.com/livekit/agents/pull/3517

PR #3517, opened 2025-09-27
fix missing simultaneous message & tool call

furious-luke:
When the LLM returns both a message, and one or more tool calls, the message was not being added to the locally copied chat context. This resulted in the tool call result being executed without the first message. Typically this caused the model to repeat itself.

I believe this resolves the issue #3407.


theomonnom (approved):

theomonnom:
lgtm! Thank you!

PR #3524, opened 2025-09-29
feat(speechmatics): add max_speakers parameter for speaker diarization

nsepehr:
## Summary
This PR adds support for the `max_speakers` parameter to the Speechmatics STT plugin, allowing developers to limit the number of unique speakers detected during diarization.

## Problem
Currently, when using the Speechmatics STT plugin with diarization enabled, there's no way to specify the maximum number of speakers. The `transcription_config` parameter (which is deprecated) accepts a `speaker_diarization_config` with `max_speakers`, but this value is not preserved when the plugin processes the configuration.

## Solution  
- Added `max_speakers` as a direct parameter to the STT `__init__` method
- Updated the `STTOptions` dataclass to include the `max_speakers` field
- Modified `_process_config` to include `max_speakers` in the `speaker_diarization_config` when sending to the Speechmatics API
- Added proper handling for extracting `max_speakers` from the deprecated `transcription_config` parameter for backward compatibility
- Updated documentation to explain the new parameter

## Use Case
This parameter is particularly useful for scenarios where the number of participants is known in advance, such as:
- Couples therapy sessions (exactly 2 speakers)
- One-on-one interviews
- Small group discussions with a fixed number of participants

## Testing
- Tested locally with a couples therapy agent implementation
- Verified that the parameter is correctly passed to the Speechmatics API configuration
- Confirmed backward compatibility with the deprecated `transcription_config` parameter

## Example Usage
```python
stt = speechmatics.STT(
    language="en",
    enable_diarization=True,
    max_speakers=2,  # Limit to 2 speakers
    diarization_sensitivity=0.5,
    speaker_active_format="@[{speaker_id}]: {text}",
)
```

## Breaking Changes
None - this is a backward-compatible addition.


livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -259,6 +266,13 @@ def __init__(
                 if is_given(punctuation_overrides)
                 else config.punctuation_overrides
             )
+            # Extract max_speakers from speaker_diarization_config if present
+            if not is_given(max_speakers) and hasattr(config, "speaker_diarization_config"):
+                if (
+                    config.speaker_diarization_config
+                    and "max_speakers" in config.speaker_diarization_config
+                ):
+                    max_speakers = config.speaker_diarization_config["max_speakers"]
```
longcw: `speaker_diarization_config` is a dataclass? if it requires a specific version of `speechmatics`, we can specify it in pyproject.toml.

```suggestion
            if (
                not is_given(max_speakers)
                and (dz_cfg := config.speaker_diarization_config)
                and dz_cfg.max_speakers is not None
            ):
                max_speakers = dz_cfg.max_speakers
```

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -382,6 +397,8 @@ def _process_config(self) -> None:
             dz_cfg: dict[str, Any] = {}
             if self._stt_options.diarization_sensitivity is not None:
                 dz_cfg["speaker_sensitivity"] = self._stt_options.diarization_sensitivity
+            if self._stt_options.max_speakers is not None:
+                dz_cfg["max_speakers"] = self._stt_options.max_speakers
```
longcw: we should also refactor `_process_config` to replace the dataclass value of `transcription_config` instead of assigning a dict to it.

nsepehr: Are you suggesting we do that in this PR? This would be more of an overhaul to the way that the _process_config works, right? I could do a follow up for that if it makes more sense... I would like the max_speakers feature to work for the time being while that would be done

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -374,25 +372,28 @@ def _process_config(self) -> None:
...
+        # Build base config
+        config_kwargs: dict[str, Any] = {
+            "language": self._stt_options.language,
+            "domain": self._stt_options.domain,
+            "output_locale": self._stt_options.output_locale,
+            "operating_point": self._stt_options.operating_point,
+            "diarization": "speaker" if self._stt_options.enable_diarization else None,
+            "enable_partials": self._stt_options.enable_partials,
+            "max_delay": self._stt_options.max_delay,
+        }
```
nsepehr: @longcw is this what you were asking for?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
nsepehr: @longcw let me know if this is what you were looking for.

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -233,9 +240,7 @@ def __init__(
         """
 
         super().__init__(
-            capabilities=stt.STTCapabilities(
-                streaming=True, interim_results=True, diarization=enable_diarization
-            ),
+            capabilities=stt.STTCapabilities(streaming=True, interim_results=True),
```
longcw: `enable_diarization` is missing?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
longcw: not really, what I mean is the assigned types don't match the annotated types of `TranscriptionConfig`. For example, in TranscriptionConfig `additional_vocab: Optional[dict] = None` needs a dict, but here the assigned value is a list of dict,
```python
config_kwargs["additional_vocab"] = [
                {
                    "content": e.content,
                    "sounds_like": e.sounds_like,
                }
                for e in self._stt_options.additional_vocab
            ]
```

same issue with `speaker_diarization_config: Optional[SpeakerDiarizationConfig] = None`, here assigning a `dz_cfg: dict[str, Any]` to the var wants a dataclass `SpeakerDiarizationConfig`, `TranscriptionConfig(**config_kwargs)` won't convert the dict to dataclass.

We should add the following to the end of https://github.com/livekit/agents/blob/main/pyproject.toml to enable the type checking with speechmatics package, 
```toml
[[tool.mypy.overrides]]
module = "speechmatics.*"
follow_untyped_imports = true
```
then `mypy --install-types --non-interactive -p livekit.plugins.speechmatics` will raise errors like 
```bash
livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py:248: error: Incompatible types in assignment (expression has type "str | None", variable has type "str | NotGiven")  [assignment]
livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py:249: error: Incompatible types in assignment (expression has type "str | None", variable has type "str | NotGiven")  [assignment]
livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py:252: error: Incompatible types in assignment (expression has type "bool | None", variable has type "bool")  [assignment]
livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py:253: error: Incompatible types in assignment (expression has type "float | None", variable has type "float")  [assignment]
livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py:255: error: Incompatible types in assignment (expression has type "list[AdditionalVocabEntry] | dict[Any, Any] | None", variable has type "list[AdditionalVocabEntry] | NotGiven")  [assignment]
...
```

longcw: I was trying to fix the types in https://github.com/livekit/agents/pull/3599 and noticed that the values set to `transcription_config.speaker_diarization_config` and `transcription_config.additional_vocab` are not the same as the type annotation that they have, and these values are converted to a dict that sent to the API.

It looks like the type annotation in the sdk is not the same to what the API expected?

nsepehr: Here's what I found

  1. The speechmatics-rt library is flexible - it accepts BOTH:
    - Plain dict objects
    - SpeakerDiarizationConfig dataclass instances
    - Both produce the exact same API payload when .to_dict() is called
  2. The type hints ARE technically incorrect:
    - additional_vocab is typed as Optional[dict] but the API expects a list of dicts
    - speaker_diarization_config is typed as Optional[SpeakerDiarizationConfig] but it also accepts dict
  3. Both approaches work at runtime because:
    - The library doesn't enforce strict type checking
    - The .to_dict() method handles both cases correctly
    - The API gets the same JSON payload either way

longcw: thanks for validation! can you merge https://github.com/livekit/agents/pull/3599 to your updates and use the correct values with a `type: ignore` for now.

nsepehr: Merged

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -370,26 +385,20 @@ def _process_config(self) -> None:
...
-                }
-            if dz_cfg:
-                transcription_config.speaker_diarization_config = dz_cfg
+            # Create SpeakerDiarizationConfig with explicit parameters
+            transcription_config.speaker_diarization_config = SpeakerDiarizationConfig(
+                max_speakers=self._stt_options.max_speakers,
+                speaker_sensitivity=self._stt_options.diarization_sensitivity,
+                prefer_current_speaker=self._stt_options.prefer_current_speaker,
+                # TODO: speakers field is not supported by SpeakerDiarizationConfig yet
+                # speakers={s.label: s.speaker_identifiers for s in self._stt_options.known_speakers},
```
longcw: can you use the dict for now with a `type: ignore` to ignore the type check?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -370,26 +385,20 @@ def _process_config(self) -> None:
...
-            transcription_config.additional_vocab = [
-                {
-                    "content": e.content,
-                    "sounds_like": e.sounds_like,
-                }
-                for e in self._stt_options.additional_vocab
-            ]
+            # Using dict format as per type annotation
+            transcription_config.additional_vocab = {
+                e.content: e.sounds_like for e in self._stt_options.additional_vocab
```
longcw: same here, use the list of dict if that's required by the API.

nsepehr: For additional_vocab: You're correct. The API expects a list (array), not a dict. According to the docs, it should be:
  [
      {"content": "hello", "sounds_like": ["helo", "halo"]},
      {"content": "world", "sounds_like": ["wurld"]}
  ]

  NOT:
  {
      "hello": ["helo", "halo"],
      "world": ["wurld"]
  }

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
nsepehr: The SpeakerDiarizationConfig IS a dataclass provided by the Speechmatics SDK. We're using it correctly here. Whether you pass the dataclass or a dict, they serialize to the same JSON structure. The dataclass approach is cleaner and type-safe. I don't think we should change that.

nsepehr:
@longcw can you please review this so it can be landed for the next release 🙏

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
longcw: I think we should keep the support of `speakers` options using dict for now.

nsepehr: Ok... updated it with using dict. Can you review now please.

Issue #3525, opened 2025-09-29
After being wrapped with FallbackAdapter, it is not possible to obtain model_name and model_provider in metrics.


Panmax:
To improve availability, my STT, LLM, and TTS are all wrapped with FallbackAdapter, resulting in model_name and model_provider under metadata in the metrics being unknown.

Panmax:
@longcw I have already upgraded to 1.2.14, but the model_name and model_provider in the LLM metrics logs are still "unknown".

longcw:
@Panmax the fix #3526 was not released yet, will be included in next release.

PR #3526, opened 2025-09-29
fix model provider and metrics for FallbackAdapter and StreamAdapter

longcw:
fix https://github.com/livekit/agents/issues/3525

- always forward metrics from the actual tts/llm/stt when using fallback adapter
- fix provider and name for fallback adapter and stream adapter


livekit-agents/livekit/agents/llm/fallback_adapter.py
```diff
@@ -73,6 +73,17 @@ def __init__(
             _LLMStatus(available=True, recovering_task=None) for _ in self._llm_instances
         ]
 
+        for llm_instance in self._llm_instances:
+            llm_instance.on("metrics_collected", self._on_metrics_collected)
+
+    @property
+    def model(self) -> str:
+        return "FallbackAdapter"
```
Panmax, line 81: So, when using FallbackAdapter, is it impossible to obtain the actual model being used?

longcw: the metrics emitted are still from the actual model. and if you use tracing, both fallback adapter and the actual model will create (nested) spans including model names.

Panmax: get

PR #3534, opened 2025-09-29
ignore agent context on generate_reply if flag is set

s-hamdananwar:


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -777,17 +777,25 @@ def _generate_reply(
         tool_choice: NotGivenOr[llm.ToolChoice] = NOT_GIVEN,
         allow_interruptions: NotGivenOr[bool] = NOT_GIVEN,
         schedule_speech: bool = True,
+        ignore_agent_instructions: bool = False,
```
theomonnom, line 780: I'm not sure if `ignore_agent_instructions` is the right API to add. 

Maybe we can just allow to pass-in a custom chat_ctx?

PR #3535, opened 2025-09-29
feat: submit final transcripts from "utterance" field

dan-ince-aai:
We have added a new field to the Turn message response called "utterance" which returns Final transcript on any short pause (160ms). This is so users can do preemptive generation with the transcript text without waiting for end_of_turn=True. API reference: https://www.assemblyai.com/docs/api-reference/streaming-api/streaming-api#receive.receiveTurn.utterance

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -339,12 +339,8 @@ class SpeechStream(stt.SpeechStream):
             end_of_turn = data.get("end_of_turn")
             utterance = data.get("utterance")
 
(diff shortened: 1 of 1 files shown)


tmshapland:
@longcw @theomonnom hold off on merging this. I'll let you know when it's ready.

dan-ince-aai:
Hi @theomonnom could we merge this? Utterances is working successfully in the inference service

longcw:
> added a new field to the Turn message response called "utterance" which returns Final transcript on any short pause (160ms)

we recently introduced a `PREFLIGHT_TRANSCRIPT` speech event type https://github.com/livekit/agents/blob/livekit-agents@1.2.14/livekit-agents/livekit/agents/stt/stt.py#L32, do you think if it's suitable for the "utterance" here?

PR #3536, opened 2025-09-30
support STT with `model:lang` and parse model specs outside ctor

longcw:


livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -302,6 +302,11 @@ def __init__(
             ),
         )
 
+    @classmethod
+    def parse_model(cls, model: str) -> LLM:
```
davidzhao: nit: should we call this `from_short_id` or `from_model_string`

longcw: `from_model_string` sounds good, or `from_model_spec`?

bcherry: I have the same question in docs - what do we call this thing?

"spec" or "specifier" seem ok. I also like "descriptor":

```
@classmethod
def parse_descriptor(cls, descriptor: str) -> LLM:
```

and then docs can say something like "use a model descriptor for a shorthand way to specify the model and common options, or a class instance if additional configuration is needed"

longcw: I like `descriptor`, and the method becomes `from_model_descriptor()`?

bcherry: i think just `from_descriptor` would be fine

bcherry: but this is internal anyways so not a big deal either way. i ended up using "descriptor" for the concept in the docs

Issue #3089, opened 2025-08-05
AgentTask awaiting: RealtimeError: update_chat_ctx timed out with openai.realtime.RealtimeModel


lajd:
I get the following error when running the example AgentTask:

```
-08-05 15:57:27,705 - DEBUG livekit.agents - tools execution completed {"speech_id": "speech_485e492e87ed"}
2025-08-05 15:57:32,705 - ERROR livekit.agents - Error in _traceable_on_enter 
Traceback (most recent call last):
  File "/testenvlib/python3.11/site-packages/livekit/agents/utils/log.py", line 16, in async_fn_logs
    return await fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/testenvlib/python3.11/site-packages/livekit/agents/voice/agent_activity.py", line 388, in _traceable_on_enter
    await self._agent.on_enter()
  File "/testenv/tasks.py", line 47, in on_enter
    if await CollectConsent():
       ^^^^^^^^^^^^^^^^^^^^^^
  File "/testenvlib/python3.11/site-packages/livekit/agents/voice/agent.py", line 712, in __await_impl
    await old_agent.update_chat_ctx(
  File "/testenvlib/python3.11/site-packages/livekit/agents/voice/agent.py", line 169, in update_chat_ctx
    await self._activity.update_chat_ctx(chat_ctx)
  File "/testenvlib/python3.11/site-packages/livekit/agents/voice/agent_activity.py", line 290, in update_chat_ctx
    await self._rt_session.update_chat_ctx(chat_ctx)
  File "/testenvlib/python3.11/site-packages/livekit/plugins/openai/realtime/realtime_model.py", line 978, in update_chat_ctx
    raise llm.RealtimeError("update_chat_ctx timed out.") from None
livekit.agents.llm.realtime.RealtimeError: update_chat_ctx timed out.
```

Here's the code:

```
import os
from dotenv import load_dotenv
from livekit import agents
from livekit.agents import AgentSession
from livekit.agents import AgentTask, function_tool
from livekit.plugins import openai
from livekit.agents import Agent, function_tool
from livekit.agents.log import logger as livekit_logger


load_dotenv()



class CollectConsent(AgentTask[bool]):
    def __init__(self):
        super().__init__(
            instructions="Ask for recording consent and get a clear yes or no answer."
        )

    async def on_enter(self) -> None:
[...]

PR #3537, opened 2025-09-30
draft for workflow design

longcw:


livekit-agents/livekit/agents/beta/workflows/email_address.py
```diff
@@ -27,38 +27,21 @@ class GetEmailResult:
...
         self,
+        *,
         chat_ctx: NotGivenOr[llm.ChatContext] = NOT_GIVEN,
         turn_detection: NotGivenOr[TurnDetectionMode | None] = NOT_GIVEN,
         stt: NotGivenOr[stt.STT | None] = NOT_GIVEN,
         vad: NotGivenOr[vad.VAD | None] = NOT_GIVEN,
         llm: NotGivenOr[llm.LLM | llm.RealtimeModel | None] = NOT_GIVEN,
         tts: NotGivenOr[tts.TTS | None] = NOT_GIVEN,
         allow_interruptions: NotGivenOr[bool] = NOT_GIVEN,
+        extra_instructions: NotGivenOr[str] = NOT_GIVEN,
```
longcw, line 38: for these built-in workflows, I think we should allow users to add extra instructions or even modify instructions completely. for a real productions, users may want to add a lot of other prompts like context, tones, fallback handling, examples…

livekit-agents/livekit/agents/beta/workflows/email_address.py
```diff
@@ -74,6 +57,32 @@ def __init__(
         # used to ignore the call to confirm_email_address in case the LLM is hallucinating and not asking for user confirmation
         self._email_update_speech_handle: SpeechHandle | None = None
 
+    @property
+    def base_instructions(self) -> str:
```
longcw, line 61: besides adding a `extra_instructions` option, how about expose `base_instructions` as a property, so users can modify it by creating a sub class if they want.

livekit-agents/livekit/agents/beta/workflows/consent_collection.py
```diff
@@ -0,0 +1,77 @@
...
+    from ...voice.agent_session import TurnDetectionMode
+
+
+@dataclass
+class CollectConsentResult:
+    consent: bool
+    denied_reason: str | None
+
+
+class CollectConsentTask(AgentTask[CollectConsentResult]):
```
longcw, line 21: CollectConsent might be another common task (e.g. https://github.com/livekit/agents/issues/3089), it will require more customization on instructions to add the context

livekit-agents/livekit/agents/beta/workflows/consent_collection.py
```diff
@@ -0,0 +1,77 @@
...
+            turn_detection=turn_detection,
+            stt=stt,
+            vad=vad,
+            llm=llm,
+            tts=tts,
+            allow_interruptions=allow_interruptions,
+        )
+
+    @property
+    def base_instructions(self) -> str:
```
theomonnom, line 49: Maybe we can override `instructions`?

longcw: the original `instructions` property is for users to read the instructions, we won't set the instructions from it?

PR #3562, opened 2025-10-02
added drain param to session.shutdown()

tinalenguyen:
flexibility for draining tasks/activity


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -662,8 +662,8 @@ def _close_soon(
             self._aclose_impl(error=error, drain=drain, reason=reason)
         )
 
-    def shutdown(self) -> None:
-        self._close_soon(error=None, drain=True, reason=CloseReason.USER_INITIATED)
+    def shutdown(self, drain: bool = True) -> None:
```
```suggestion
    def shutdown(self, *, drain: bool = True) -> None:
```

PR #3021, opened 2025-07-25
feature/enable-prometheus-multiprocess-collector

johncDepop:
adding the ability to use the multiprocess collector if the env variable `PROMETHEUS_MULTIPROC_DIR` is present

PR #3565, opened 2025-10-03
Configure Prometheus in multi process mode

efontan-dialpad:
# Add Prometheus multi-process metrics Support

## Overview

This implementation adds support for Prometheus metrics collection from child job processes in the LiveKit workers. 
When multi-process mode is enabled, metrics from all child job processes are aggregated and exposed through the `/metrics` endpoint.

## Problem

Prometheus runs in single-process mode by default, but LiveKit workers are multi-process. 
Because of this, the `/metrics` endpoint only exposed metrics from the main worker process because:

1. Each process has its own separate instance of the Prometheus global `REGISTRY`
2. Child processes write metrics to their own memory space
3. The HTTP server in the main process only reads from its own registry

If we want to create new metrics inside the entrypoint function, we need to put the Prometheus client in multi-process mode.

## Solution

I've followed the docs to implement multi-process mode using the `prometheus_client`: https://prometheus.github.io/client_python/multiprocess/

1. We need to set a `PROMETHEUS_MULTIPROC_DIR` environment variable to a temporary directory that Prometheus can use to write metrics (before startup).
2. The metrics endpoint aggregates data from all files when scraped
3. Stale metrics from dead processes are automatically cleaned up

## Limitations

Prometheus multi-process mode has some inherent limitations:

- Registries can not be used as normal, all instantiated metrics are exported
- Registering metrics to a registry later used by a MultiProcessCollector may cause duplicate metrics to be exported
- Custom collectors do not work (e.g. cpu and memory metrics)
- Gauges cannot use set_function
- Info and Enum metrics do not work
- The pushgateway cannot be used
- Gauges cannot use the pid label
- Exemplars are not supported
- Remove and Clear of labels are currently not supported in multiprocess mode.

### Option 1: Set Environment Variable (Recommended)

Set the `PROMETHEUS_MULTIPROC_DIR` environment variable before starting your agent:

```bash
export PROMETHEUS_MULTIPROC_DIR=/tmp/prometheus
mkdir -p $PROMETHEUS_MULTIPROC_DIR
python your_agent.py
```

### Option 2: Configure via WorkerOptions

```python
from livekit.agents import WorkerOptions
from livekit import agents

opts = WorkerOptions(
    entrypoint_fnc=entrypoint,
    prometheus_port=9999,
    prometheus_multiproc_dir="/tmp/prometheus"
)

agents.cli.run_app(opts)
```

## Backward Compatibility

**This change is fully backward compatible:**

- If `PROMETHEUS_MULTIPROC_DIR` is **not set**, the system works exactly as before (single-process mode)
- If `PROMETHEUS_MULTIPROC_DIR` **is set**, multi-process mode is automatically enabled

## Files Modified

1. **`worker.py`**
   - Detects `prometheus_multiproc_dir` option or `PROMETHEUS_MULTIPROC_DIR` environment variable
   - Sets environment variable for child processes
   - Cleans directory on startup
   - Periodically updates child process count metric

2. **`telemetry/metrics.py`**
   - Added `multiprocess_mode` parameter to Gauge metrics
   - Removed `set_function()` (not supported in multiprocess mode)
   - Added explicit update function for child process count

3. **`telemetry/http_server.py`**
   - Detects if multiprocess mode is enabled
   - Uses `MultiProcessCollector` to aggregate metrics from all processes
   - Falls back to default registry if multiprocess mode is disabled

4. **`ipc/supervised_proc.py`**
   - Calls `mark_process_dead(pid)` when child processes exit
   - Ensures proper cleanup of stale metrics

## Disk Permissions

The process must have write permissions in the directory specified by `PROMETHEUS_MULTIPROC_DIR`.
If you are running your worker inside a Kubernetes pod, you can use a `emptyDir` volume to store the metrics.

## Migration Guide to enable Multi-Process metrics

1. Set `PROMETHEUS_MULTIPROC_DIR` environment variable
2. Create the directory with write permissions
3. Restart your agent
4. Verify metrics are being aggregated

Example for Kubernetes:
```yaml
apiVersion: v1
kind: Pod
spec:
  containers:
  - name: agent
    env:
    - name: PROMETHEUS_MULTIPROC_DIR
      value: /tmp/prometheus
    volumeMounts:
    - name: metrics
      mountPath: /tmp/prometheus
  volumes:
  - name: metrics
    emptyDir: {}
```

### DIFF
--- a/livekit-agents/livekit/agents/ipc/supervised_proc.py
+++ b/livekit-agents/livekit/agents/ipc/supervised_proc.py
@@ -4,6 +4,7 @@ import asyncio
 import contextlib
 import logging
 import multiprocessing as mp
+import os
 import socket
 import sys
 import threading
@@ -278,6 +279,23 @@ class SupervisedProc(ABC):
 
         await self._join_fut
         self._exitcode = self._proc.exitcode
+
+        # Mark process as dead for prometheus multiprocess mode.
+        # This is required to clean up stale metrics.
+        if "PROMETHEUS_MULTIPROC_DIR" in os.environ and self._pid:
+            try:
+                from prometheus_client import multiprocess
+                multiprocess.mark_process_dead(self._pid)
+                logger.debug(
+                    "marked process as dead for prometheus multiprocess mode",
+                    extra=self.logging_extra(),
+                )
+            except Exception as e:
+                logger.warning(
+                    f"failed to mark process as dead for prometheus: {e}",
+                    extra=self.logging_extra(),
+                )
+
         self._proc.close()
         await aio.cancel_and_wait(ping_task, read_ipc_task, main_task)
 
--- a/livekit-agents/livekit/agents/telemetry/http_server.py
+++ b/livekit-agents/livekit/agents/telemetry/http_server.py
@@ -1,12 +1,15 @@
 from __future__ import annotations
 
 import asyncio
+import os
 
 import aiohttp.web_request
 from aiohttp import web
 from prometheus_client import (
     CONTENT_TYPE_LATEST,
+    CollectorRegistry,
     generate_latest,
+    multiprocess,
 )
 
 from .. import utils
@@ -14,9 +17,15 @@ from .. import utils
 
 async def metrics(_request: aiohttp.web_request.Request) -> web.Response:
     def _get_metrics() -> bytes:
-        # registry = CollectorRegistry(auto_describe=True)
-        # multiprocess.MultiProcessCollector(registry)
-        return generate_latest()
+        # Check if multiprocess mode is enabled
+        if "PROMETHEUS_MULTIPROC_DIR" in os.environ:
+            # Create a new registry for this request to collect metrics from all processes
+            registry = CollectorRegistry(auto_describe=True)
+            multiprocess.MultiProcessCollector(registry)
+            return generate_latest(registry)
+        else:
+            # Use default global registry if multiprocess mode is not enabled
+            return generate_latest()
 
     loop = asyncio.get_running_loop()
     data = await loop.run_in_executor(None, _get_metrics)
--- a/livekit-agents/livekit/agents/telemetry/metrics.py
+++ b/livekit-agents/livekit/agents/telemetry/metrics.py
@@ -12,18 +12,34 @@ PROC_INITIALIZE_TIME = prometheus_client.Histogram(
     buckets=[0.1, 0.5, 1, 2, 5, 10],
 )
 
+# Use 'livesum' mode to aggregate active jobs across all processes
+# This sums the values from processes that are still running
 RUNNING_JOB_GAUGE = prometheus_client.Gauge(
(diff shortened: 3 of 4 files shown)


efontan-dialpad:
> lgtm! do you know if this could prevent the worker from running if the multiprocess dir doesn't exist?

Good question, and yes. In my tests, I had to create the directory in the pods (either create the full path or use `/tmp`); otherwise, the worker crashes during startup. I didn't want to include it inside the worker code, but if you think it makes sense, I can add a commit with this.

Also, where is the best place to add all of this documentation for configuring Prometheus?

theomonnom:
cc @bcherry for docs!

efontan-dialpad:
@bcherry, Let me know how I can help with documenting this change. I would love to merge this PR and start using it. We are currently pointing the dependency to my fork, and it works well!

theomonnom:
> if you think it makes sense, I can add a commit with this.

Yeah I think we should do it, otherwise it'll crash the worker for every user at the first run

bcherry:
@efontan-dialpad thank you so much! We're planning on adding prometheus docs in general and will include this when we do. It looks pretty straightforward and I think you've given us the info we need already

btw @theomonnom I think https://github.com/livekit/agents/pull/3021 is a duplicate and can be closed in favor of this one

sohbit:
@bcherry @theomonnom is there an estimation on when this gets merged and released?

theomonnom:
@efontan-dialpad should I take over this PR?

efontan-dialpad:
> @efontan-dialpad should I take over this PR?

@theomonnom Seems I don't have permissions to merge the PR in this repo. Maybe because I forked it... If you can, yes please.

theomonnom:
Oh I meant to make sure we create the prometheus directory at startup. (to avoid the worker crashing)
https://github.com/livekit/agents/pull/3565#issuecomment-3383582786

efontan-dialpad:
> Oh I meant to make sure we create the prometheus directory at startup. (to avoid the worker crashing) [#3565 (comment)](https://github.com/livekit/agents/pull/3565#issuecomment-3383582786)

Oh I can make the change and push another commit.

efontan-dialpad:
> Oh I meant to make sure we create the prometheus directory at startup. (to avoid the worker crashing) [#3565 (comment)](https://github.com/livekit/agents/pull/3565#issuecomment-3383582786)

@theomonnom Done, I just added a line in the worker if a directory is passed:

```
if self._prometheus_multiproc_dir:
    os.makedirs(self._prometheus_multiproc_dir, exist_ok=True)
``` 
- if the directory does not exist, `os.makedirs` creates the directory recursively. For example, if I pass "/tmp/livekit/prometheus" but "/tmp/livekit/" does not exist, it will be created automatically.
- if the directory exists, do nothing.

I've tested it with the `minimal_worker.py` and that works.
Let me know if there's anything else I can do to get this PR merged. Thanks!

theomonnom (approved):
Thanks! lgtm!

PR #3570, opened 2025-10-03
Updating Cartesia Version

namantalreja:
## Changes
- Updating Cartesia TTS Version to `2025-04-16`
- Bumping Version to `1.3.0`

## Testing
- Tested with Agent setup locally
- Tested with both Stream and Chunked Outputs

### DIFF
--- a/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
+++ b/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
@@ -49,7 +49,7 @@ from .models import (
 
 API_AUTH_HEADER = "X-API-Key"
 API_VERSION_HEADER = "Cartesia-Version"
-API_VERSION = "2024-06-10"
+API_VERSION = "2025-04-16"
 
 
 @dataclass
--- a/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/version.py
+++ b/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/version.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "1.2.14"
+__version__ = "1.3.0"


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "1.2.14"
+__version__ = "1.3.0"
```
davidzhao: this is managed by the framework, so you don't have to manually bump this.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -49,7 +49,7 @@
 
 API_AUTH_HEADER = "X-API-Key"
 API_VERSION_HEADER = "Cartesia-Version"
-API_VERSION = "2024-06-10"
+API_VERSION = "2025-04-16"
```
davidzhao, line 52: any incompatibilities between the versions?

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/version.py
namantalreja: fixed

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
namantalreja: In this version we’re removing support for:

- Embeddings
- stability cloning mode
- Experimental controls for adjusting speed and emotion.

Here is a [link](https://docs.cartesia.ai/developer-tools/changelog) to our Changelog.

wlbksy:
This is a breaking commit. "2025-04-16" does not support
1. Embeddings
2. stability cloning mode
3. Experimental controls for adjusting speed and emotion.

You should revert this commit, or you should branching on this features.

wlbksy:
>> Users who absolutely need embeddings or experimental controls should use API version 2024-11-13 with model ID sonic-2-2025-03-07, both of which are still available. Please be aware that these features could lead to reduced model stability even with that model.

These part of code still exists and are actually used by some users in the `TTS.__init__()` in the same file.

PR #3578, opened 2025-10-05
cast deepgram sttv2 status_code to int from str

hwuiwon:
`data.get("code")` returns string. Cast it to int to prevent

```
TypeError
'>=' not supported between instances of 'str' and 'int'
```


hwuiwon:
@davidzhao

davidzhao:
good catch, though this isn't the right fix. they are returning a string enum as an error code. see #3578

PR #3580, opened 2025-10-05
chore: ensure a recent version of certifi is installed

davidzhao:
on some machines, there could be an old version of certifi on the machine
that isn't able to handle LiveKit Cloud's certs


livekit-agents/pyproject.toml
```diff
@@ -27,6 +27,7 @@ classifiers = [
 ]
 dependencies = [
     "click~=8.1",
+    "certifi>=2025.6.15",
```
theomonnom, line 30: Is it being used by default tho?

davidzhao: yes. we need SSL root certs in order to connect to anything https. `requests`, `httpx`, `httpcore` all depends on this with a versionless dependency.

the problem is the default version on some systems are too old that they cannot connect to cloud. it's more of an issue on Mac dev machines, since some OS bundle an really old version of python

PR #3587, opened 2025-10-06
feat(AWS): Bedrock Selective Guardrails

itskyf:
## Selective Bedrock Guardrail
- The plugin now runs the guardrail only on the **latest user message** when `guardrailConfig` is set.  
- Since guardrails are charged per token, guard the latest user message will reduce the cost.

## Typing for aioboto3
- Added `types-aioboto3[bedrock-runtime]` to the **dev dependencies** for better IDE support and type hints.  
- It’s used only in `TYPE_CHECKING`, I try to not change any logic.


livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
```diff
@@ -289,29 +321,28 @@ def _parse_chunk(self, request_id: str, chunk: dict) -> llm.ChatChunk | None:
                 ),
             )
         elif "contentBlockStop" in chunk:
```
itskyf, line 323: The `if self._tool_call_id` and `if self._tool_call_id is None` is conflict, so I remove the `if self._tool_call_id` check

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
```diff
@@ -289,29 +321,28 @@ def _parse_chunk(self, request_id: str, chunk: dict) -> llm.ChatChunk | None:
...
-            if self._tool_call_id:
```
itskyf, line 292: This line conflicts with the `if self._tool_call_id is None` below, so I delete it.

livekit-agents/livekit/agents/llm/_provider_format/aws.py
```diff
@@ -33,6 +33,8 @@ def to_chat_ctx(
             role = "assistant"
         elif msg.type == "function_call_output":
             role = "user"
+        else:
+            raise ValueError(f"Unknown message type: {msg.type!r}")
```
davidzhao: can we log here instead of raise? don't want to break users applications in case this is happening

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
davidzhao: this should not be deleted. instead the next line `if self._tool_call_id is None` should be

PR #3594, opened 2025-10-07
feat(telemetry/utils): add ttft reporting

bml1g12:


livekit-agents/livekit/agents/telemetry/trace_types.py
```diff
@@ -77,3 +77,7 @@
 
 # Platform-specific attributes
 ATTR_LANGFUSE_COMPLETION_START_TIME = "langfuse.observation.completion_start_time"
+# Langfuse only allows searching/filtering/graphing over fields if they are prefixes with this key OR
+# which use a set of reserved keys like ATTR_GEN_AI and ATTR_LANGFUSE items above.
+# so e.g. ATTR_LANGFUSE_SEARCHABLE_PREFIX + "foo" will allow `foo` to be a filterable attribute in langfuse.
+ATTR_LANGFUSE_SEARCHABLE_PREFIX = "langfuse.trace.metadata."
```
longcw: is this used anywhere? if not it's better to not include it in the types.

bml1g12: you don't think its useful reference for devs (i.e. to sign post you cannot assume simply adding an atrribute makes it useful for searching/graphing in langfuse)? This information was a little hard for me to find as a developer and surprising behaviour in LangFuse (that metadata attached to a trace is not searchable unless nested under very specific namespaces). 

But OK I have removed it now

longcw: I think it's langfuse should document this somewhere, we may not want to add too many platform-specific vars in the codebase.

PR #3596, opened 2025-10-07
feat(tts): integrate AsyncAI TTS engine into livekit

ashotbagh:
## Summary by CodeRabbit

* **New Features**
  * Adds an AsyncAI text-to-speech plugin with real-time streaming audio, configurable model, language (en, de, es, fr, it), encoding (PCM variants), voice selection, tokenizer support, and connection pooling with lifecycle management.
* **Documentation**
  * Adds a README with installation steps, API key prerequisite, and integration links.
* **Chores**
  * Adds packaging and version metadata for distribution.

### DIFF
--- a/livekit-plugins/livekit-plugins-asyncai/livekit/plugins/asyncai/tts.py
+++ b/livekit-plugins/livekit-plugins-asyncai/livekit/plugins/asyncai/tts.py
@@ -23,6 +23,7 @@ import weakref
 from collections import deque
 from dataclasses import dataclass, replace
 from typing import Union, cast
+from urllib.parse import urlencode
 
 import aiohttp
 
@@ -58,7 +59,7 @@ class _TTSOptions:
     sample_rate: int
     voice: str | list[float]
     api_key: str
-    language: str
+    language: str | None
     base_url: str
 
     def get_http_url(self, path: str) -> str:
@@ -90,7 +91,7 @@ class TTS(tts.TTS):
 
         Args:
             model (TTSModels, optional): The Async TTS model to use. Defaults to "asyncflow_multilingual_v1.0".
-            language (str, optional): The language code for synthesis. Defaults to "en".
+            language (str, optional): The language code for synthesis.
             encoding (TTSEncoding, optional): The audio encoding format. Defaults to "pcm_s16le".
             voice (str, optional): The voice ID.
             sample_rate (int, optional): The audio sample rate in Hz. Defaults to 32000.
@@ -143,9 +144,8 @@ class TTS(tts.TTS):
 
     async def _connect_ws(self, timeout: float) -> aiohttp.ClientWebSocketResponse:
         session = self._ensure_session()
-        url = self._opts.get_ws_url(
-            f"/text_to_speech/websocket/ws?{API_AUTH_HEADER}={self._opts.api_key}&{API_VERSION_HEADER}={API_VERSION}"
-        )
+        query = urlencode({API_AUTH_HEADER: self._opts.api_key, API_VERSION_HEADER: API_VERSION})
+        url = self._opts.get_ws_url(f"/text_to_speech/websocket/ws?{query}")
 
         init_payload = {
             "model_id": self._opts.model,
@@ -209,8 +209,8 @@ class TTS(tts.TTS):
 
     def synthesize(
         self, text: str, *, conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS
-    ):
-        pass
+    ) -> tts.ChunkedStream:
+        raise NotImplementedError("AsyncAI TTS supports streaming only; use tts.stream().")
 
     async def aclose(self) -> None:
         for stream in list(self._streams):
@@ -245,7 +245,7 @@ class SynthesizeStream(tts.SynthesizeStream):
             ws: aiohttp.ClientWebSocketResponse, asyncai_context_id: str
         ) -> None:
             async for ev in sent_tokenizer_stream:
-                token_pkt = {}
+                token_pkt: dict[str, object] = {}
                 token_pkt["transcript"] = ev.token + " "
                 token_pkt["context_id"] = asyncai_context_id
                 token_pkt["force"] = True
@@ -294,7 +294,8 @@ class SynthesizeStream(tts.SynthesizeStream):
                 if current_segment_id is None:
                     current_segment_id = segment_id
                     output_emitter.start_segment(segment_id=segment_id)
-                if data.get("final") and data["final"] is True:
+                final = data.get("final")
+                if isinstance(final, bool) and final:
                     if sent_tokenizer_stream.closed:
                         output_emitter.end_input()
                         break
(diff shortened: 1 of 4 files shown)


davidzhao:
Thanks for the PR!

While testing it, I'm finding a few issues:

1. the websocket connection would time out, giving users an error:

should the connection be kept open? if not then you shouldn't use a connection pool, but just using it for each synthesis
```
2025-12-14 00:51:52,631 - WARNING livekit.plugins.asyncai - unexpected message {'error_code': 'TIMEOUT', 'message': 'Timeout: no message received in 60.0 seconds', 'request_id': 'unknown', 'extra': None} {"room": "sbx-104ei0-mcZwM3K6KcttVEyXaVjnRX", "pid": 16917, "job_id": "AJ_gjHdmUFbjtCx"}
2025-12-14 00:51:52,641 - WARNING livekit.agents - failed to synthesize speech, retrying in 0.1s
  File "/Users/davidzhao/dev/livekit/agents-async/livekit-plugins/livekit-plugins-asyncai/livekit/plugins/asyncai/tts.py", line 283, in _run
    await asyncio.gather(*tasks)
  File "/Users/davidzhao/dev/livekit/agents-async/livekit-plugins/livekit-plugins-asyncai/livekit/plugins/asyncai/tts.py", line 253, in _recv_task
    raise APIStatusError(
        "Async connection closed unexpectedly", request_id=request_id
    )
livekit.agents._exceptions.APIStatusError: Async connection closed unexpectedly (status_code=-1, request_id=74718f843e3b, body=None, retryable=True)
```

2. the `final` message never comes, so the synthesis isn't successful

please use `basic_agent.py` as a test case to ensure the TTS could perform in a conversation of 5 mins long. also, the branch is a bit out of date from main, it'd be great if you could rebase the PR to main

livekit-plugins/livekit-plugins-asyncai/livekit/plugins/asyncai/tts.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def synthesize(
+        self, text: str, *, conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS
+    ):
+        pass
+
+    async def aclose(self) -> None:
+        for stream in list(self._streams):
+            await stream.aclose()
+
```
tinalenguyen, line 218:
```suggestion
await self._sent_tokenizer_stream.aclose()
```

livekit-plugins/livekit-plugins-asyncai/livekit/plugins/asyncai/tts.py
```diff
@@ -0,0 +1,293 @@
...
+                    )
+
+                if msg.type != aiohttp.WSMsgType.TEXT:
+                    logger.warning("unexpected Async message type %s", msg.type)
+                    continue
+
+                data = json.loads(msg.data)
+                if current_segment_id is None:
+                    current_segment_id = "new_segment"
+                    output_emitter.start_segment(segment_id="new_segment")
```
tinalenguyen: It may be better to keep track of the current segment id as `self._current_segment_id`
```suggestion
                if current_segment_id is None:
                    current_segment_id = utils.shortuuid()
                    output_emitter.start_segment(segment_id=current_segment_id)
```

livekit-plugins/livekit-plugins-asyncai/livekit/plugins/asyncai/tts.py
```diff
@@ -0,0 +1,293 @@
...
+            sample_rate=self._opts.sample_rate,
+            num_channels=1,
+            mime_type="audio/pcm",
+            stream=True,
+        )
+
+        async def _sentence_stream_task(ws: aiohttp.ClientWebSocketResponse) -> None:
+            async for ev in self._sent_tokenizer_stream:
+                token_pkt = {}
+                token_pkt["transcript"] = ev.token + " "
```
tinalenguyen, line 249: I think we should pass a `context_id` here related to each synthesis and keep track of it as `self._current_segment_id` to be used later in `output_emitter.start_segment(segment_id=self._current_segment_id)`

ashotbagh:
Hey @davidzhao , @tinalenguyen 
Thanks for the review.

The earlier version of this PR was indeed out of date. Since then, our AsyncAI WebSocket endpoint has changed and now supports multi-context sessions, which required updating the connection handling logic. I’ve updated the implementation accordingly and rebased the branch on the latest main.
I’ve tested the updated version locally using basic_agent.py, and did not observe the timeout or any other issue.

Happy to make further adjustments if you still see any issues or have any other suggestions.

tinalenguyen:
@ashotbagh Thanks for making the changes! 

I still see these errors when interrupting the agent:
```
    raise APIStatusError(
        "Async connection closed unexpectedly", request_id=request_id
    )
livekit.agents._exceptions.APIStatusError: Async connection closed unexpectedly (status_code=-1, request_id=33b99ecfaa23, body=None, retryable=True)
```

Could it be that each context must be flushed or closed after each synthesis? 

I also occasionally see empty audio messages like:
`livekit.…s.asyncai unexpected message {'final': False, 'audio': '', 'context_id': '470f507b-0a3a-46c6-a5af-796eb148a5ad'} `

ashotbagh:
@tinalenguyen Thanks for the comments!

Regarding the “connection closed unexpectedly” issue: I’ve tried to reproduce it extensively (including interruption scenarios), but haven’t been able to trigger it locally so far. At the moment, the context flush happens on the final chunk, where we explicitly send:
`end_pkt = {}
end_pkt["transcript"] = ""
end_pkt["context_id"] = asyncai_context_id
await ws.send_str(json.dumps(end_pkt))`

I’ve also made a small update to silently ignore empty audio messages (e.g. audio == "" on non-final frames), instead of logging them as unexpected, and applied a few minor fixes to satisfy ruff checks.

Could you please let me know if the issue still persists with the latest changes, and, if possible, share a reliable way to reproduce it? I’d be happy to dig deeper once I can observe it on my side.

davidzhao:
testing with `basic_agent`, after it performs a tool call, the agent would freeze on me.

ashotbagh:
Thanks for the report.

The freeze after a tool call was caused by a deadlock in the Async TTS streaming path: the receive loop was waiting for input_sent_event, which was only set when at least one sentence token was emitted. In tool-call flows, it’s possible for the tokenizer to emit no tokens before the end packet is sent, which caused the recv task to wait indefinitely.

I’ve fixed this by ensuring the receive loop is unblocked even when no tokens are emitted, so the stream always progresses correctly after tool calls.

I also cleaned up import ordering to satisfy ruff checks.

I’ve re-tested with basic_agent.py, including repeated tool calls, and the agent no longer hangs. Please let me know if you still see any issues or edge cases I should cover.

ashotbagh:
Hi @davidzhao , @tinalenguyen . 
Just a quick follow-up, I pushed a fix for the tool-call hang (now reproducible and resolved) and cleaned up the remaining ruff issues.
When you have a moment, could you please take another look and let me know if everything looks good on your side? Happy to make further adjustments if needed. 
Thanks!

ashotbagh:
@davidzhao 
I’ve addressed the review comments and CI failures:
	•	Fixed mypy issues,
	•	Resolved formatting and linting issues,
	•	Rebased the branch on the latest main

All checks should be passing now. Please let me know if there’s anything else you’d like me to adjust - happy to iterate. Thanks!

davidzhao (approved):

PR #3601, opened 2025-10-08
Add a format provider for VLLM models.

LaurentMazare:
Some VLLM models, e.g. gemma 3 12B, expect some strict ordering on message roles, i.e. these should be `system`, `user`, `assistant`, `user`, `assistant`, etc. I haven't been able to get this to work with any of the existing `provider_fmt` options, `mistrlai` ensures that the last message is a `user` message but doesn't compact the existing `user` or `assistant` messages.

This PR adds a new format provider option, `vllm`, which ensures that the roles are presented in the proper ordering. Consecutive `user` messages are compacted together. This was tested to work properly with vllm 0.9.1 and various models (gemma 3, mistral).

Not sure if this the right way to solve this issue so let me know if there is a better alternative.

### DIFF
--- a/livekit-agents/livekit/agents/llm/_provider_format/vllm.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/vllm.py
@@ -19,7 +19,7 @@ def to_chat_ctx(
         messages.append({"role": "user", "content": dummy_user_message})
     if len(messages) > 1 and messages[0]["role"] == "system" and messages[1]["role"] == "assistant":
         messages.insert(1, {"role": "user", "content": "Hello"})
-    collated = []
+    collated: list[dict] = []
     for msg in messages:
         if len(collated) > 0 and collated[-1]["role"] == msg["role"]:
             collated[-1]["content"] += " " + msg["content"]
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/llm/_provider_format/vllm.py
```diff
@@ -0,0 +1,28 @@
...
+    *,
+    inject_dummy_user_message: bool = True,
+    dummy_user_message: str = "",
+) -> tuple[list[dict], Literal[None]]:
+    messages, _ = openai_to_chat_ctx(chat_ctx, inject_dummy_user_message=inject_dummy_user_message)
+
+    if len(messages) == 1 and messages[0]["role"] == "system":
+        messages.append({"role": "user", "content": dummy_user_message})
+    if len(messages) > 1 and messages[0]["role"] == "system" and messages[1]["role"] == "assistant":
+        messages.insert(1, {"role": "user", "content": "Hello"})
```
theomonnom: Do we have to inject "Hello" here?

theomonnom: I think it's a good solution, I know a lot of LLMs doesn't support chat_ctx starting with an assistant role.

`inject_dummy_user_message` should already handle this

LaurentMazare: Ah thanks for catching this, it was indeed meant to be `dummy_user_message` parameter rather than the hardcoded `"Hello"`.

PR #3607, opened 2025-10-08
Allow Recording Verbal DTMF Input when ask_confirmation is turned off

toubatbrian:


theomonnom:
Can we write evals similar to https://github.com/livekit/agents/blob/4a09684f97e770529a059f38e50b910d572564b9/tests/test_workflows.py#L8?
So we can validate the system prompts, it seems quite complex

livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -73,14 +80,16 @@ async def confirm_inputs(inputs: list[DtmfEvent]) -> None:
 
         if ask_for_confirmation:
             instructions += "Once user has confirmed the digits (by verbally spoken or entered manually), call `confirm_inputs` with the inputs."
+        else:
+            instructions += "If user provides the digits through voice and it is valid, call `record_inputs` with the inputs."
```
theomonnom, line 89: I don't think the LLM can make the difference between voice/text.

theomonnom: Should we use generate_reply with user_input instead of the instructions? When receiving dtmf inputs

toubatbrian: Oh just to clarify, for the case where `ask_confirmation` is `False`. A valid dtmf input will directly trigger a `self.complete()` no tool call is needed. This `record_inputs` tool is only used if 
1) user say out loud the dtmf inputs verbally
2) ask_confirmation is set to `False`

toubatbrian: Without the tool there's no way to exit the `GetDtmfTask` with verbal inputs only. Unlike when `ask_confirmation=True` we have a `confirm_inputs` tool that LLM have access to and is forced to call regardless of verbal / dtmf number press inputs

PR #3610, opened 2025-10-09
Agent IVR System Example

toubatbrian:


toubatbrian:
> This example is quite complex, lgtm but I don't think this usecase is very common.
It's generally the opposite (agent speaking to an IVR).
Voice AI is used to replace those legacy inputs lol

Yeah 100%, also I'll try to simplify this and make a more mainstream example: https://live-kit.slack.com/archives/C07TWVC0W4A/p1760425950611219?thread_ts=1760341830.898979&cid=C07TWVC0W4A

PR #3614, opened 2025-10-09
lift google realtime api out of beta

tinalenguyen:


bcherry:
Is it possible for us to make this change backwards compatible?

davidzhao:
good point, let's alias the beta namespace to avoid breaking user code

PR #3616, opened 2025-10-09
Oracle STT, LLM, and TTS plug-ins.

kschnable:
This is the first version of new Oracle STT, LLM, and TTS Python plug-ins. Will be sending testing instructions, SVG file, 3 .md files, and authentication files to CWilson soon. CWilson indicated that he would forward all these files to client team.


kschnable:
Removing - will re-add after a few other changes.

PR #3632, opened 2025-10-13
Restore audio enabled flags after exit GetDtmfTask

toubatbrian:
For building an agentic IVR system, user may want to disable audio enabled flag globally to prevent user interrupting the DTMF collection task flow. However, since we also support voice input inside `GetDtmfTask`, we would like to enable audio to true inside the task and restore it's original state outside of the task.


theomonnom:
I don't think the AgentTask should edit this enabled flag at all, it shouldn't touch the IO

Issue #3401, opened 2025-09-10
Langfuse / OTEL for llm judge when testing


marctorsoc:
Hi team, I would like to monitor and use the playground of langfuse for the LLM judge when [testing](https://docs.livekit.io/agents/build/testing/#llm-based-judgment). I found that this works:
```
from langfuse import openai as langfuse_openai
    import sys

    monkeypatch.setitem(sys.modules, "openai", langfuse_openai)
```
and I get a trace for the judge too. But it's really hacky. Is there a better way to do this?

davidzhao:
I'm not sure this is a use case we fully understand. Could you share some examples on the output you are looking to achieve?

marctorsoc:
I'm looking into having traces of the LLM judge when testing an agent. Let's say I have
```
@pytest.mark.asyncio # Or your async testing framework of choice
async def test_your_agent(setup_langfuse) -> None:
    async with (
        openai.LLM(model="gpt-4o-mini") as llm_judge,
        AgentSession(llm=llm) as session,
    ):
        # Start the agent in the session
        await session.start(Assistant())
        
        # Run a single conversation turn based on the given user input
        result = await session.run(user_input="Hello")
        
        await (
            result.expect.next_event().is_message(role="assistant")
            .judge(
                llm, intent="Offers a friendly introduction and offer of assistance."
            )
        )

```
where `setup_langfuse` is a fixture that configures the trace provider. When I run the test, I'll get traces for the agents in the session, BUT the trace for the llm_judge of the assertion does not appear in langfuse. Please let me know if you need more details

longcw:
there should be`llm_request` and `llm_request_run` span when using `llm.chat` in llm judge. maybe you need to setup setup_langfuse in the global level and add a root span for `.judge(...)` since it's outside the agent session.

marctorsoc:
@longcw created an example here: https://github.com/marctorsoc/livekit-example. Could you provide what I need to change so the LLM requests of the llm_judge are also captured in langfuse?

marctorsoc:
@longcw I do see a `llm_request_run` but it's empty

<img width="1126" height="844" alt="Image" src="https://github.com/user-attachments/assets/683e86ed-e327-421d-b681-e020bf05766f" />

However, if I do 
```
from langfuse import openai as langfuse_openai
    import sys

    monkeypatch.setitem(sys.modules, "openai", langfuse_openai)
```
I see

[...]

PR #3639, opened 2025-10-13
feat(voice/run_result): OTEL tracing of judge

bml1g12:
This PR enhances the OpenTelemetry tracing for the LLM-as-a-judge functionality in Livekit to provide:
- The model name (so e.g. LangFuse can estimate price)
- The token counting (so e.g. LangFuse can estimate price)
- A named parent span (so e.g. its clear that it is a span not part of a regular conversation with the agent) - before this PR it was an unnamed parent span (see screenshots below for comparison)
- The LLM-as-a-judge results (shared via the already existing OTEL attributes for function call results)

_____

**Before this PR:**

image

_____

**After this PR:**

image

### DIFF
--- a/livekit-agents/livekit/agents/voice/run_result.py
+++ b/livekit-agents/livekit/agents/voice/run_result.py
@@ -894,7 +894,7 @@ class ChatMessageAssert:
 
         arguments: str | None = None
         usage: llm.CompletionUsage | None = None
-          
+
         extra_kwargs = {}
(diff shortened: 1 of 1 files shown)


marctorsoc:
This is great since now I get this
image
but still not the full input-output. So I still need what I mentioned [here](https://github.com/livekit/agents/issues/3401#issuecomment-3474167446) to see
image
and also be able to play with the playground e.g. to adjust the intent

PR #3648, opened 2025-10-14
fix: heartbeat

zachkamran:
The new deepgram api requires WS pings otherwise it will close the connection.

https://developers.deepgram.com/docs/flux/flux-nova-3-comparison#connection-management


davidzhao (approved):
nice catch! thanks Zach

PR #3649, opened 2025-10-15
Unprompted STT Reconnection at startup

adrian-cowham:
Addressing an issue that was causing the LiveKit Inference STT to reconnect immediately after the initial connection. Note, this erroneous behavior does not impact functionality, it may delay startup a bit but the agent is fully functional.


theomonnom (approved):
lgtm, just need ruff

PR #3651, opened 2025-10-15
fix #3650 cartesia version backward compatibility

### LINKED ISSUE #3650: PR #3570 is breaking for livekit Cartesia user
https://docs.cartesia.ai/developer-tools/changelog

api version "2025-04-16" does not support
1. Embeddings
2. stability cloning mode
3. Experimental controls for adjusting speed and emotion.

> Users who absolutely need embeddings or experimental controls should use API version 2024-11-13 with model ID sonic-2-2025-03-07, both of which are still available. Please be aware that these features could lead to reduced model stability even with that model.

These part of code still exists and are actually used by some users in the `TTS.__init__()` in the same file.

ref pr: https://github.com/livekit/agents/pull/3570

wlbksy:
Reset Cartesian api version to an older one and add a api version parameter to the constructor for easier user adjustment.

fix #3650

### DIFF
--- a/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
+++ b/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
@@ -50,6 +50,8 @@ from .models import (
 API_AUTH_HEADER = "X-API-Key"
 API_VERSION_HEADER = "Cartesia-Version"
 API_VERSION = "2025-04-16"
+API_VERSION_WITH_EMBEDDINGS_AND_EXPERIMENTAL_CONTROLS = "2024-11-13"
+MODEL_ID_WITH_EMBEDDINGS_AND_EXPERIMENTAL_CONTROLS = "sonic-2-2025-03-07"
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -89,6 +92,7 @@ def __init__(
         tokenizer: NotGivenOr[tokenize.SentenceTokenizer] = NOT_GIVEN,
         text_pacing: tts.SentenceStreamPacer | bool = False,
         base_url: str = "https://api.cartesia.ai",
+        api_version: str = API_VERSION_WITH_EMBEDDINGS_AND_EXPERIMENTAL_CONTROLS,
```
davidzhao, line 95: we should default to the non-deprecated version. we have to follow API deprecation from the model providers.

davidzhao (approved):

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
wlbksy: That would be another issue for updating 3rd api version, which should be highlighted in livekit change log. Because some users might be using the old version.

PR #3654, opened 2025-10-15
feat: add preflight transcript via utterance

dan-ince-aai:
added utterance as a preflight transcript as per @longcw guidance, docs: https://www.assemblyai.com/docs/api-reference/streaming-api/streaming-api#receive.receiveTurn.utterance

the utterance is emitted if a short pause is detected and all words have been finalised, so that it can be used for preemptive generation

one issue I did notice though is that the preemptive generation validation is an absolute comparison so even though no words have changed, if the final transcript includes an extra comma it cancels, is this intended or could it be edited to first lowercase the text and strip punc and compare just words?

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -64,7 +64,7 @@ class STT(stt.STT):
         api_key: NotGivenOr[str] = NOT_GIVEN,
         sample_rate: int = 16000,
         encoding: Literal["pcm_s16le", "pcm_mulaw"] = "pcm_s16le",
-        speech_model: Literal[
+        model: Literal[
             "universal-streaming-english", "universal-streaming-multilingual"
         ] = "universal-streaming-english",
         end_of_turn_confidence_threshold: NotGivenOr[float] = NOT_GIVEN,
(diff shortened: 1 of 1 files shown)


theomonnom:
lgtm! can you fix the type checks?

> one issue I did notice though is that the preemptive generation validation is an absolute comparison so even though no words have changed, if the final transcript includes an extra comma it cancels, is this intended or could it be edited to first lowercase the text and strip punc and compare just words?

fair point, I think we can do this change, do you want to contribute it?

dan-ince-aai:
Done, thanks @theomonnom!

I also added speech_model to support the multilingual model which is now in prod, the docs unfortunately haven't been updated yet to display the speech_model param instead of language but this was deployed to prod yesterday: https://www.assemblyai.com/docs/universal-streaming/multilingual-transcription

Will contribute the preemptive validation change in a separate PR

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -45,7 +46,10 @@
 class STTOptions:
     sample_rate: int
     buffer_size_seconds: float
-    encoding: str = "pcm_s16le"
+    encoding: Literal["pcm_s16le", "pcm_mulaw"] = "pcm_s16le"
+    speech_model: Literal["universal-streaming-english", "universal-streaming-multilingual"] = (
```
davidzhao, line 50: could you update this to `model` to be consistent? `speech` is redundant since this is an STT module

davidzhao (approved):

PR #3665, opened 2025-10-16
use inference gateway in the readme

theomonnom:


README.md
bcherry: you can remove this

bcherry, line 119: remove this

README.md
```diff
@@ -177,9 +177,9 @@ async def entrypoint(ctx: JobContext):
     userdata = StoryData()
     session = AgentSession[StoryData](
         vad=silero.VAD.load(),
-        stt=deepgram.STT(model="nova-3"),
-        llm=openai.LLM(model="gpt-4o-mini"),
-        tts=openai.TTS(voice="echo"),
+        stt="deepgram/nova-3",
+        llm="openai/gpt-4o",
```
bcherry, line 181: we're recommending `openai/gpt-4.1.mini` as the new default

theomonnom: This model hallucinates a lot, can't get the drive-thru tests to pass

bcherry: oh we should probably not be using it as default then - will revive the convo in slack

README.md
```diff
@@ -177,9 +177,9 @@ async def entrypoint(ctx: JobContext):
...
+        tts="elevenlabs/eleven_turbo_v2_5:Xb7hH8MSUJpSbSDYk0k2",
```
bcherry:
```suggestion
        tts="cartesia/sonic-2:9626c31c-bec5-4cca-baa8-f8ba9e84c8bc",
```

README.md
tmshapland: @bcherry @theomonnom I can't find the slack thread where you're discussing what the default should be. Looks like you landed on 4o. Should we have the examples and the quickstart also use 4o by default? 

When we were talking about this for the gateway before launching, I remember DZ saying we should use 4.1 bc OAI will eventually deprecate 4o. But I was also just talking to a user who said, "4o 2024-05-13 is better at tool calling, and oai models are not robotic, and the intelligence doesn’t drop with large context, all for the best trade off on latency. 4.1 and gemini 2.5 flash comes close, but  the may version of 4o is the best. They even charge more for it." Lots of users have different preferences, but it was interesting to hear this guy back up what he likes about 4o, and specify the may version specifically.

image

bcherry: i posted in slack but got no replies on it https://live-kit.slack.com/archives/C05P6DRC6MV/p1760744284787699

I don't have an opinion, happy to do whatever the team thinks is the right answer but want to be consistent across the board and also get this PR landed!

PR #3674, opened 2025-10-17
brianyin/agt-2076-support-repeat-instruction-in-dtmf-gathering

toubatbrian:


livekit-agents/livekit/agents/beta/workflows/dtmf_inputs.py
```diff
@@ -40,6 +44,7 @@ def __init__(
         *,
         num_digits: int,
         ask_for_confirmation: bool = False,
+        repeat_instructions: int = 1,
```
theomonnom, line 47: I'm not sure if this is actually a feature that is useful for users? 
Seems like it's only useful for us while debugging? Do we want to expose it?

toubatbrian: Honestly, I’m not sure either. We won’t really know until users start using it. That said, Twilio’s digit gather does support specifying the number of repeats, so having some level of parity seems reasonable. In any case, this feature is non-intrusive. If users stick with the default, the behavior remains exactly as before.

theomonnom: Can we revert it for now? It's harder to remove features than adding new ones

toubatbrian: I'm still using it to test the dtmf agent

PR #3680, opened 2025-10-17
TaskGroup

theomonnom:


livekit-agents/livekit/agents/beta/workflows/task_group.py
```diff
@@ -0,0 +1,128 @@
...
+    def _build_out_of_scope_tool(self, *, active_task_id: str) -> FunctionTool | None:
+        if not self._visited_tasks:
+            return None
+
+        # Only allow to regress to already visited tasks
+        task_ids = self._visited_tasks.copy()
+        task_ids.discard(active_task_id)
+        task_repr = {f.id: f.description for f in self._registered_factories if f.id in task_ids}
+
+        description = (
```
davidzhao, line 106: this description seems very specific to the style of agent where it's modifying information in another task.. is it robust enough for other use cases as a task?

should we also allow users to skip around? currently it's only allowing you to go back to previous tasks. I'm wondering if this should be specified on each task itself either:
* can be revisited
* can be skipped ahead

tinalenguyen: I think in future versions of `TaskGroup`, we can allow additional parameters to tasks for those:

`mutable: bool` will determine whether or not a task can be revisited
`global: bool` will determine if a task can be called regardless of whether it was visited or not

(Open to other name suggestions)

theomonnom: @davidzhao Do you have any examples or scenarios where you’d want a task that can be skipped ahead?

> `is it robust enough for other use cases as a task?`

For now we're mostly focused on input collection, do you have other use cases in mind?

PR #3681, opened 2025-10-18
add survey agent example

tinalenguyen:


examples/survey/survey_agent.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+@dataclass
+class Userdata:
+    filename: str
+    task_results: dict
+
+
+def write_to_csv(filename: str, data: dict):
+    with open(filename, "a", newline="") as csvfile:
```
theomonnom: open() is a blocking IO function, we should use aiofiles

tinalenguyen: thoughts on using [aiocsv](https://pypi.org/project/aiocsv/#description) instead? the csv writer is synchronous

theomonnom: fwiw, csv is very easy to do. Just do quote escape on values and separate columns using a comma

examples/survey/survey_agent.py
```diff
@@ -0,0 +1,392 @@
...
+
+        Args:
+            chosen_scenario (str): The scenario the candidate chose, either 'old_project' or 'new_project'
+            old_project_scale_response (str): An overview of the candidate's response
+        """
+        self.session.generate_reply(
+            instructions="Express interest in seeing the candidate scale their project as they described in the future."
+        )
+        results = {"scenario": chosen_scenario, "response": scale_response}
+        self._result["scaling_project_response"] = results
```
theomonnom: Can we use dataclasses for the results, I think it would be cleaner since we keep everything typed. 
Can we also have tasks that get a deterministic input? Like yes or no, or an enumeration of choices

tinalenguyen: sounds good for dataclass results!

as for deterministic input, maybe for `CommuteTask` we can have `can_commute: bool` and an enumeration of transportation choices?

examples/survey/survey_agent.py
tinalenguyen: when writing values that way, any commas in the values would split the answer across multiple columns. i'm now using both aiofiles and aiocsv

PR #3682, opened 2025-10-18
chore: correct minimax-ai package name

davidzhao:
also removed play.ai plugin and updated keywords


davidzhao:
> should we remove the playai plugin from the docs too?

yeah I'll get rid of it

PR #3694, opened 2025-10-21
fix(google): omit scheduling parameter for Vertex AI function responses

dkristak:
Fixes `ValueError` when using `RealtimeModel` with Vertex AI where function calls execute successfully but sending responses back to the model fails with: `scheduling parameter is not supported in Vertex AI.`

```
Error traceback:
  File "livekit/plugins/google/realtime/realtime_api.py", line 732
    await session.send_tool_response(function_responses=msg.function_responses)
  File "google/genai/live.py", line 407, in send_tool_response
    tool_response_dict = live_converters._LiveClientToolResponse_to_vertex(...)
  File "google/genai/_live_converters.py", line 2203
    raise ValueError('scheduling parameter is not supported in Vertex AI.')
```

Root cause: `get_tool_results_for_realtime()` was setting the `scheduling` parameter to `WHEN_IDLE` by default for all `FunctionResponse` objects, including when `vertexai=True`. Vertex AI's converter explicitly rejects any scheduling parameter in function responses.

Impact: Functions would execute successfully but Gemini never received the results, causing conversation flow to break, repeated function calls, or the model to get stuck.

Fix: Conditionally construct FunctionResponse objects - only include `scheduling` and `id` parameters when `vertexai=False`, matching the pattern already used for the `id` parameter.

Related: Similar to how `id` is excluded for Vertex AI per:
https://github.com/googleapis/python-genai/blob/85e00bc/google/genai/_live_converters.py#L1435

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/utils.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/utils.py
@@ -105,17 +105,25 @@ def get_tool_results_for_realtime(
     function_responses: list[types.FunctionResponse] = []
     for msg in chat_ctx.items:
         if msg.type == "function_call_output":
-            res = types.FunctionResponse(
-                name=msg.name,
-                response={"output": msg.output},
-                scheduling=tool_response_scheduling
-                if is_given(tool_response_scheduling)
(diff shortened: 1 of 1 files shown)


davidzhao:
issue is fixed in #3781

PR #3704, opened 2025-10-22
Fix/aws realtime tooluse barge in

kachenjr:
# Fix Nova Sonic audio routing, tool use, and barge-in handling
This commit fixes three critical bugs in the Nova Sonic realtime plugin that
prevent it from working correctly in production scenarios.

### 1. Audio Routing After Tool Calls
**Problem**: Audio frames not playing after tool execution
**Root Cause**: Audio routed to wrong generation after tool calls complete
**Solution**: Create new generation per ASSISTANT SPECULATIVE text event
**Impact**: Audio now plays correctly for each assistant response

Nova Sonic sends ASSISTANT SPECULATIVE text events to signal new assistant
turns, including after tool calls. Each turn needs its own generation to
ensure audio frames route to the correct audio channel.

### 2. Tool Use Across Multiple Turns
**Problem**: Tool calls fail or behave incorrectly across multiple turns
**Root Cause**: Generation not closed after tool call, preventing framework
from delivering tool results via update_chat_ctx()
**Solution**: Close generation immediately after emitting tool call
**Impact**: Tool use now works reliably across multiple conversation turns

The LiveKit framework expects the generation to close so it can call
update_chat_ctx() with tool results. A new generation is created when
Nova Sonic sends the next ASSISTANT SPECULATIVE event with the response.

### 3. Crashes on User Interruption (Barge-In)
**Problem**: Session crashes when user interrupts assistant mid-response
**Root Cause**: Race conditions with future initialization and None pointer
access after barge-in sets _current_generation to None
**Solution**: 
- Initialize futures as None, create lazily in initialize_streams()
- Add defensive None checks throughout event handlers
**Impact**: Interruptions handled gracefully without crashes

Creating futures in __init__ causes race conditions during session restart.
Lazy initialization ensures the event loop exists before future creation.

### Simplified Architecture
- **Message tracking**: Single content_id_map dict instead of 4 separate dicts
  (messages, user_messages, speculative_messages, tool_messages)
- **Restart tracking**: Per-turn _restart_attempts instead of session-level
  tracking for better barge-in metrics
- **Timestamps**: Float (time.time()) instead of ISO-8601 strings for easier
  duration calculations

The single dict approach is simpler, easier to debug, and sufficient for
tracking content IDs and their types. Both approaches have identical memory
characteristics (no leaks) since dicts live inside _ResponseGeneration
instances that are created and destroyed per turn.

### Adopted from Origin
- Added ModelStreamErrorException to recoverable errors (from origin/main
  commit c6747052, Oct 16, 2025)

## Testing

All features verified working:
- Audio playback ✓
- Tool use (multiple turns) ✓
- Barge-in/interruptions ✓
- Multi-turn conversations ✓

Tested against origin/main and confirmed tool use does not work without
these fixes.

## Breaking Changes

None. Public API unchanged. Metrics format unchanged. Only internal
implementation differs.

## AI Assistance

Portions of this code were developed with assistance from AI tools for
debugging, testing, and implementation of the fixes described above.

---

Co-authored-by: Amazon Q Developer

### DIFF
--- a/livekit-agents/livekit/agents/ipc/supervised_proc.py
+++ b/livekit-agents/livekit/agents/ipc/supervised_proc.py
@@ -4,6 +4,7 @@ import asyncio
 import contextlib
 import logging
 import multiprocessing as mp
+import os
 import socket
 import sys
 import threading
@@ -278,6 +279,24 @@ class SupervisedProc(ABC):
 
         await self._join_fut
         self._exitcode = self._proc.exitcode
+
+        # Mark process as dead for prometheus multiprocess mode.
+        # This is required to clean up stale metrics.
+        if "PROMETHEUS_MULTIPROC_DIR" in os.environ and self._pid:
+            try:
+                from prometheus_client import multiprocess
+
+                multiprocess.mark_process_dead(self._pid)  # type: ignore[no-untyped-call]
+                logger.debug(
+                    "marked process as dead for prometheus multiprocess mode",
+                    extra=self.logging_extra(),
+                )
+            except Exception as e:
+                logger.warning(
+                    f"failed to mark process as dead for prometheus: {e}",
+                    extra=self.logging_extra(),
+                )
+
         self._proc.close()
         await aio.cancel_and_wait(ping_task, read_ipc_task, main_task)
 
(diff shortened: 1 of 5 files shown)


theomonnom:
Thanks!

PR #3705, opened 2025-10-23
Fix: Connection Pool race condition 

adrian-cowham:
This PR guards against a get() and prewarm() race condition in `ConnectionPool`. 

During testing I noticed a single agent making 2 simultaneous connections to the inference service. I was able to debug this and track it down to a race condition in `ConnectionPool` between the TTS implementation calling [`_pool.connection()`](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/inference/tts.py#L460) and the `AgentActivity` calling the TTS's [`prewarm`](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/voice/agent_activity.py#L421) method.

This is reliably reproducible by starting an agent that uses a TTS implementation. Note, the TTS implementation must also be using a `ConnectionPool` and it must `prewarm` the `ConnectionPool`.

Here's my understanding of the order of operations:
1. `AgentActivity` calls `tts.prewarm()`
2. `tts.prewarm` calls `self._pool.prewarm()`
3. `pool.prewarm()` kicks off an async task to make the WebSocket connection
4. `AgentActivity` proceeds
5. yada yada yada
6. While the async task is making the WebSocket connection, the agent starts the TTS implementation and the TTS implementation calls `_pool.connection()`. This is the race condition. If `pool.prewarm()` hasn't completed, `_pool.connection()` will create a connection **AND** when `pool.prewarm()` finished, it will have created an additional connection.

### Impact
Each agent will create 2 WebSocket connections to the TTS provider for the duration of the session. They are reliably closed at the end of a session.

### Proposed Fix:
I added a lock around getting a connection in the `ConnectionPool`. I feel like this is a bit of a sledge hammer, feedback is encouraged and appreciated.

### DIFF
--- a/livekit-agents/livekit/agents/utils/connection_pool.py
+++ b/livekit-agents/livekit/agents/utils/connection_pool.py
@@ -4,6 +4,7 @@ import weakref
 from collections.abc import AsyncGenerator, Awaitable
 from contextlib import asynccontextmanager
 from typing import Callable, Generic, Optional, TypeVar
+
 from . import aio
 
 T = TypeVar("T")
@@ -40,7 +41,7 @@ class ConnectionPool(Generic[T]):
         self._connections: dict[T, float] = {}  # conn -> connected_at timestamp
         self._available: set[T] = set()
         self._connect_timeout = connect_timeout
-        self.connect_lock = asyncio.Lock()
+        self._connect_lock = asyncio.Lock()
 
         # store connections to be reaped (closed) later.
         self._to_close: set[T] = set()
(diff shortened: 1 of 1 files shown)


davidzhao (commented):
change lgtm, could you run ruff to clear up [CI](https://github.com/livekit/agents/actions/runs/18734923483/job/53439568479?pr=3705)?

livekit-agents/livekit/agents/utils/connection_pool.py
```diff
@@ -41,6 +40,7 @@ def __init__(
         self._connections: dict[T, float] = {}  # conn -> connected_at timestamp
         self._available: set[T] = set()
         self._connect_timeout = connect_timeout
+        self.connect_lock = asyncio.Lock()
```
theomonnom: can you prefix with `_`

adrian-cowham: On it.

PR #3707, opened 2025-10-23
Support Agent Session Tools 

toubatbrian:
Support adding tools directly to agent sessions that are shared by all agents


toubatbrian:
@theomonnom as we discussed, here's PR for adding agent session tools, I'm going to merge this small PR for now to unblock the ivr activity once. But let me know if you want to add anything else, I can do a quick follow up

PR #3711, opened 2025-10-24
add extra instructions + tools params into GetEmailTask

tinalenguyen:
allows for `GetEmailTask()` to be used in more specialized cases


tinalenguyen:
@theomonnom In the SurveyAgent use case, `decline_email_capture()` causes TaskGroup to break since ToolError is an exception. I bypassed this by passing in the `disqualify()` tool and adding extra instructions to call that instead. I think it can be helpful to "override" things like that

Issue #3617, opened 2025-10-09
STT Transcripts Don't Trigger User "Speaking" State When VAD Misses Speech


philipp-eisen:
## Environment

`livekit-agents==1.2.11`

We are using the livekit session with:
```
...
turn_detection = MultilingualModel(),
vad=silero.VAD.load(),
...
```
as well as Deepgram as STT

## Problem 
We've noticed an issue that we at times receive a transcript from the STT model. 
In traces we can see that we have a `eou_detection` span, however we don't have a `user_speaking` span and we also don't receive any `user_state_changed` event.
If the EOU model detects the speech as final the agent will continue in the conversation.

However, this is very problematic for us when the previous state of the user was `away`, because then our agent will continue to think that our the user is away and after some time end the call. 

I have seen PR #3559 -- however this does not appear to change the user state.

longcw:
> In traces we can see that we have a eou_detection span, however we don't have a user_speaking span and we also don't receive any user_state_changed event.

since we update the user state based on VAD results, so this is expected if VAD has a miss detection.

> this is very problematic for us when the previous state of the user was away

maybe we can set user state to `listen` when a final user transcript received and the current user state is `away`.

philipp-eisen:
yeah that would be helpful

Pn7Hao:
Hi, I can confirm that sometimes VAD doesn’t trigger the user stage change from listening to speaking. If you use the turn_detector model, it can have an even more serious impact. Therefore, I suggest turning off VAD and letting the STT/ASR front end handle everything—it will improve performance.

zaheerabbas-prodigal:
Hello @longcw - the VAD missing the speech also causes issues in `transctiption_delay` calculation as it relies on the `last_speaking_time`. So if the user state is not updated because VAD misses the detection of speech - this emits incorrect transcription_delay numbers which are NOT accurate.

Any reason why the VAD misses the speech and also how we can improve the transcription delay calculation in such cases? If the VAD hasn't fired - we should use the final transcript timestamp itself as the last speech timestamp - this could help avoid these miscalculations?

This is happening more frequently in `livekit-agents=1.2.15`

longcw:
@zaheerabbas-prodigal 
> Any reason why the VAD misses the speech and also how we can improve the transcription delay calculation in such cases?

it's the false negative of VAD, you may tune the `min_speech_duration` and `activation_threshold` of VAD but basically it's a trade-off between false positive and false negative. 

[...]

PR #3716, opened 2025-10-25
reset user state from away when user input transcribed

longcw:
fix https://github.com/livekit/agents/issues/3617

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1094,6 +1094,10 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         )
 
     def _user_input_transcribed(self, ev: UserInputTranscribedEvent) -> None:
+        if self.user_state == "away" and ev.is_final:
+            # reset user state from away to listening in case VAD has a miss detection
+            self._update_user_state("listening")
+
         self.emit("user_input_transcribed", ev)
 
     def _conversation_item_added(self, message: llm.ChatMessage) -> None:


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1094,6 +1094,10 @@ def _update_user_state(
         )
 
     def _user_input_transcribed(self, ev: UserInputTranscribedEvent) -> None:
+        if self.user_state == "away" and ev.is_final:
```
davidzhao, line 1097: similarly, should we trigger user_state changes to speaking when any input is detected?

longcw: what inputs do you mean, like the interim transcripts, or text input?

davidzhao: I meant interim transcripts.. that should set the user to `speaking`, if it's away.

but I think text input should also set it to `listening`

theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_session.py
theomonnom: like firing speaking for 0.5s?

theomonnom: I think it's fine to keep it to `listening` IMO

longcw: there is a potential issue, the state will stay in speaking if there is no final transcript after interim transcripts.. so I prefer to only set to listening with stt.

davidzhao: I guess it's fine as well. though it'd be good to handle "no final transcripts" in a generic fashion.. we should time it out and just stick with interim if final never comes

PR #3722, opened 2025-10-27
brianyin/agt-2058-dtmf-write-evals-for-getdtmftask

toubatbrian:


tests/test_workflows.py
```diff
@@ -92,3 +93,63 @@ async def test_get_dtmf_sip_event_with_confirmation() -> None:
...
+            result = RunResult[None](output_type=None)
+            sess._global_run_state = result
+
+            events = [rtc.SipDTMF(code=1, digit=str(i)) for i in range(10)]
+            events = [*events, rtc.SipDTMF(code=1, digit="#")]
+            for event in events:
+                ctx.room.emit_event("sip_dtmf_received", event)
+                await asyncio.sleep(0.1)
+
+            await result
```
toubatbrian: @theomonnom I’m trying to verify that a generate_reply is triggered after sending this sequence of DTMF events. Right now, the assertion utility only initiates an assertion result by actively calling generate_reply via `sess.run`, so it doesn’t work for detecting passively generated replies.

Any ideas for handling this kind of assertion more effectively? This is the best workaround I’ve come up with so far, unless we invest in building some dedicated assertion utilities for this case.

theomonnom: if the generate_reply isn't called inside a function_tool, we have no way to detect it

theomonnom: This definitely looks hacky tho..

toubatbrian: Should we rethink how this test is written? Curious if you have any better ideas.

I’d also like to keep it, since triggering generate_reply on confirmation is a key piece of logic I want the task to enforce.

toubatbrian: How about this:

```py
async with sess.eval(output_type=None) as run_result:
    events = [rtc.SipDTMF(code=1, digit=str(i)) for i in range(10)]
    events = [*events, rtc.SipDTMF(code=1, digit="#")]
    for event in events:
        ctx.room.emit_event("sip_dtmf_received", event)
        await asyncio.sleep(0.1)

run_result.expect.next_event()....
```

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -436,6 +437,19 @@ def run(self, *, user_input: str, output_type: type[Run_T] | None = None) -> Run
...
+            raise RuntimeError("nested runs are not supported")
+
+        try:
+            run_state = RunResult(output_type=output_type)
+            self._global_run_state = run_state
+            yield run_state
+            await run_state
+        finally:
+            self._global_run_state = None
+
```
toubatbrian, line 452: ^

PR #3724, opened 2025-10-27
turn detection: model v0.4.0-intl

lwestn:
Updating turn detector model. You can compare with current prod model for phone number/address etc issue here: https://huggingface.co/spaces/livekit/eot-visualization.

I confirmed this is behaving as expected when running the agent locally.


jeradf (approved):

taresh18:
@lwestn Hi, great update. My use-case is phone numbers and emails and its accuracy seems to be much better than before. However, still not as accurate as https://huggingface.co/fixie-ai/ultraVAD
What are your views on using audio context as well for turn detection along with the chat context?

Issue #3732, opened 2025-10-28
Feature: Support Google LLM safety_settings param


tamirh:
### Feature Type

Would make my life easier

### Feature Description

Gemini supports various content filtering parameters under `safety_settings` in the `GenerateContentConfig` call to restrict the LLM output based on different `HarmCategory` definitions and `HarmBlockThreshold` levels.

Could we update the google `llm.py` plugin so we can pass through the `safety_settings` as a param in the `google.LLM()` constructor?

### Workarounds / Alternatives

* Filtering and restrictions via prompts
* Secondary agent watching text stream to monitor for safety

### Additional Context

* https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#harmcategory
* https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/configure-safety-filters#gemini-api-in-vertex-ai

### PR
https://github.com/livekit/agents/pull/3733

tinalenguyen:
Hi, thank you for the PR! I reviewed it, once it's finalized feel free to mark it ready for review :)

tinalenguyen:
@tamirh Hi, are you still able to work on the PR? If not, I can take over for you :)

tinalenguyen:
completed in #3815

PR #3733, opened 2025-10-28
Add support for safety_settings in google LLM constructor

tamirh:
## Description
Feature Request: https://github.com/livekit/agents/issues/3732

Adding in a way to define safety restrictions on Google LLM via the constructor params

## Changes

Take `safety_settings` list/dict and pass it on to the LLMStream via `_extra_kwargs.`

Note: Also added doc line for missing doc on `seed` param

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
@@ -63,6 +63,7 @@ class _LLMOptions:
     gemini_tools: NotGivenOr[list[_LLMTool]]
     http_options: NotGivenOr[types.HttpOptions]
     seed: NotGivenOr[int]
+    safety_settings: NotGivenOr[list[types.SafetySettingOrDict] | dict[str, str]]
 
 
 class LLM(llm.LLM):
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -88,6 +89,7 @@ def __init__(
         gemini_tools: NotGivenOr[list[_LLMTool]] = NOT_GIVEN,
         http_options: NotGivenOr[types.HttpOptions] = NOT_GIVEN,
         seed: NotGivenOr[int] = NOT_GIVEN,
+        safety_settings: NotGivenOr[list[types.SafetySettingOrDict] | dict[str, str]] = NOT_GIVEN,
```
tinalenguyen, line 92: I think it would be best to only accept `SafetySettingOrDict` types for this parameter. If you want to support accepting `dict[str, str]` then you would need to cast it to `SafetySetting` anyway to pass to `GenerateContentConfig`
```suggestion
        safety_settings: NotGivenOr[list[types.SafetySettingOrDict]] = NOT_GIVEN,
```

PR #3738, opened 2025-10-28
check room connection state + rename to on_emit + taskgroup fix

tinalenguyen:
encountered `Exception: cannot access local participant before connecting` from 

```
\livekit\agents\voice\agent_activity.py", line 1672, in _pipeline_reply_task  
           _set_participant_attributes(current_span, room_io.room.local_participant)                                                                      
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^   
```

- confirms connection state is 1 (`CONN_CONNECTED`)
- rename `emit` to `on_emit`, revert change
- remove usage of keys() from taskgroup visited task set

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1668,7 +1668,7 @@ class AgentActivity(RecognitionHooks):
         if new_message:
             current_span.set_attribute(trace_types.ATTR_USER_INPUT, new_message.text_content or "")
 
-        if (room_io := self._session._room_io) and room_io.room.connection_state == 1:
+        if (room_io := self._session._room_io) and room_io.room.isconnected():
             _set_participant_attributes(current_span, room_io.room.local_participant)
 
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1668,7 +1668,7 @@ async def _pipeline_reply_task(
         if new_message:
             current_span.set_attribute(trace_types.ATTR_USER_INPUT, new_message.text_content or "")
 
-        if room_io := self._session._room_io:
```
theomonnom, line 1671: You can import the actual value rtc.ConnectionState.{enum_here}

theomonnom: fwiw i think we have isconnected()

tinalenguyen: ah you're right, great call!!

theomonnom (approved):

PR #3754, opened 2025-10-29
Add Speechmatics TTS

aaronng91:
Add support for Speechmatics TTS

### DIFF
--- a/livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/tts.py
+++ b/livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/tts.py
@@ -84,6 +84,14 @@ class TTS(tts.TTS):
         )
         self._session = http_session
 
+    @property
+    def model(self) -> str:
+        return "unknown"
+
+    @property
+    def provider(self) -> str:
+        return "Speechmatics"
+
     def _ensure_session(self) -> aiohttp.ClientSession:
         if not self._session:
             self._session = utils.http_context.http_session()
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/tts.py
```diff
@@ -0,0 +1,185 @@
...
+            base_url=base_url,
+            api_key=api_key,
+        )
+        self._session = http_session
+
+    def _ensure_session(self) -> aiohttp.ClientSession:
+        if not self._session:
+            self._session = utils.http_context.http_session()
+        return self._session
+
```
longcw, line 99: adding the model and provider properties

```suggestion
    @property
    def model(self) -> str:
        return "model name"

    @property
    def provider(self) -> str:
        return "Speechmatics"

```

aaronng91: I've added the properties, and added "unknown" as model name to follow convention from other plugins and we only have 1 TTS model at the moment.

aaronng91:
Not sure if there's an issue with the CLA assistant, but I've tried a few times to sign it and it is still showing as "not signed"

longcw (approved):

aaronng91:
> Not sure if there's an issue with the CLA assistant, but I've tried a few times to sign it and it is still showing as "not signed"

ok this has now been signed

PR #3755, opened 2025-10-29
add missing file error for turn detection model

tinalenguyen:


livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/base.py
```diff
@@ -172,14 +184,6 @@ def __init__(
...
                 self._languages = json.load(f)
 
-    @property
-    def model(self) -> str:
-        return self._model_type
-
-    @property
-    def provider(self) -> str:
-        return "livekit"
-
```
theomonnom, line 182: Any reason to remove that?

theomonnom (approved):
Thanks lgtm!

livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/base.py
tinalenguyen: ah mistake, my bad

Issue #3756, opened 2025-10-29
Add Language Parameter Support for Rime Arcana TTS Model


gokuljs:
### Feature Type

I cannot use LiveKit without it

### Feature Description

The Rime TTS plugin currently supports the `lang` parameter only for the `mistv2` model. The `arcana` model does not allow language selection, which blocks customers who need multilingual text-to-speech support.

This feature adds `lang` parameter support to both `arcana` and `mistv2` models, ensuring consistent behavior and enabling multilingual voice synthesis for global users.

📌 Reason:  This feature is customer-requested and essential for multilingual deployments. Without `lang` support in the Arcana model, customers can’t generate non-English TTS, making the product unusable for them.

### Workarounds / Alternatives

There are no viable workarounds. The only option is to use the `mistv2` model, which does not provide the same voice quality or configuration flexibility as the Arcana model.

tinalenguyen:
Thank you for the PR, we'll take a look!

PR #3757, opened 2025-10-29
Add Language Parameter Support for Rime Arcana TTS Model

gokuljs:
The Rime TTS plugin currently supports the `lang` parameter only for the `mistv2` model. The `arcana` model does not allow language selection, which blocks customers who need multilingual text-to-speech support.

This feature adds `lang` parameter support to both `arcana` and `mistv2` models, ensuring consistent behavior and enabling multilingual voice synthesis for global users.


gokuljs:
#3756

davidzhao (approved):
lg

gokuljs:
@davidzhao and idea when this will be published

PR #3765, opened 2025-10-30
Enable changing turn detection on existing session

emgeee:
This PR adds a new method to the `AgentSession` called `set_turn_detection()` that enables users to swap out the turn detection strategy mid session.

### DIFF
--- a/examples/bank-ivr/README.md
+++ b/examples/bank-ivr/README.md
@@ -1,112 +1,77 @@
-# Bank IVR Example
+# IVR Navigation Agent Example
 
-This example reenvisions the LiveKit telecom demo as a retail banking assistant. It combines a scripted IVR tree, a DTMF helper agent, and a mock banking backend so you can experiment with full call flows—authentication, menu navigation, and data lookups—without touching production systems.
+This example demonstrates how to build a Voice AI agent that can navigate IVR (Interactive Voice Response) systems. 
 
-## What's Included
+The goal is to show how an AI agent can act as a human caller: listening to menu prompts (e.g., "Press 1 for accounts..."), interpreting them, and sending DTMF tones (keypad presses) to navigate the tree and extract information or perform actions.
 
-- Read-only banking data served by `MockBankService`, covering deposits, credit cards, loans, and rewards.
-- A multi-agent IVR implemented with LiveKit Agents. The main `agent.py` orchestrates authentication, menu routing, and task execution.
-- A DTMF-focused assistant (`dtmf_agent.py`) that reads the caller's intent from dispatch metadata and presses keypad digits on the caller's behalf.
-- A dialer script (`dial_bank_agent.py`) that provisions a LiveKit dispatch, embeds the user's request, and places an outbound SIP call.
+## Overview
+
+The example consists of three parts:
+
+1.  **The Navigator (`ivr_navigator_agent.py`)**: The AI agent that calls the bank. It listens to the IVR, interprets the prompts, and sends DTMF tones to navigate. **This is the main focus of the example.**
+2.  **The Mock Bank (`ivr_system_agent.py`)**: A simulated banking IVR system. In a real application, this would be an actual phone number (like an airline or pharmacy). We provide this mock so you can test the navigator against a known system.
+3.  **The Dialer (`dial_bank_agent.py`)**: A script that connects the two. It dispatches the Navigator agent and has it dial the Mock Bank via SIP.
 
 ## File Tour
 
-- `agent.py` - primary IVR worker. Hosts the authentication flow, menu prompts, and per-domain tasks.
-- `dtmf_agent.py` - companion agent dedicated to DTMF entry. Pulls the request from `ctx.room.metadata`, drives the IVR with keypad events, and records the outcome.
-- `dial_bank_agent.py` - command-line helper that triggers a dispatch and SIP call. Accepts `--phone` and `--request` so you can enqueue different intents.
-- `mock_bank_service.py` - loads structured customer data from `data.json` into immutable dataclasses for safe read access.
-- `data.json` - sample dataset featuring two customers with realistic balances, transactions, cards, loans, and rewards.
-- `test_mock_bank_service.py` - pytest coverage for the data service and formatting helpers.
+-   **`ivr_navigator_agent.py`**: The agent logic for navigating the IVR. It receives a user intent (e.g., "Check my balance") and uses an LLM to determine which DTMF digits to press based on what it hears.
+-   **`ivr_system_agent.py`**: The mock IVR system. It plays prompts and waits for DTMF input. It uses `MockBankService` to simulate real banking data.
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -85,6 +86,9 @@ class _PreemptiveGeneration:
 
 
 # NOTE: AgentActivity isn't exposed to the public API
+_StrTurnDetection = Literal["stt", "vad", "realtime_llm", "manual"]
```
longcw: I think we can move `TurnDetectionMode` from agent_session to here or the audio_recognition to avoid duplicated definition.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -116,94 +120,126 @@ def __init__(self, agent: Agent, sess: AgentSession) -> None:
...
+            if self._false_interruption_timer is not None:
+                self._false_interruption_timer.cancel()
+                self._false_interruption_timer = None
+
+            if self._interrupt_paused_speech_task is not None:
+                if not self._interrupt_paused_speech_task.done():
+                    self._interrupt_paused_speech_task.cancel()
+                self._interrupt_paused_speech_task = None
+        elif previous_mode == "manual":
+            if self._false_interruption_timer is not None:
```
longcw: why only cancel the timer when the mode or previous mode is manual?

emgeee: Switching from one auto turn taking strategy to another one is more orthogonal to the the interruption state so it should be fine to keep the same timer running.

Going to/from manual is a little different and we want to avoid firing stale resume events.

livekit-agents/livekit/agents/voice/agent_activity.py
emgeee: good call, I've moved it to `agent_session.py`

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -116,94 +115,126 @@ def __init__(self, agent: Agent, sess: AgentSession) -> None:
...
+        self._on_enter_task: asyncio.Task | None = None
+        self._on_exit_task: asyncio.Task | None = None
+
+        # speeches that audio playout finished but not done because of tool calls
+        self._background_speeches: set[SpeechHandle] = set()
+
+    def _determine_turn_detection_mode(
+        self,
+    ) -> tuple[TurnDetectionStr | None, _TurnDetector | None]:
+        turn_detection = self.turn_detection
```
longcw: maybe better to make the turn_detection as an arg of this method?
```suggestion
    def _validate_turn_detection(
        self, turn_detection: TurnDetectionMode | None
    ) -> tuple[TurnDetectionStr | None, _TurnDetector | None]:
```

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -772,6 +772,41 @@ def update_options(
                 max_endpointing_delay=max_endpointing_delay,
             )
 
+    def set_turn_detection(self, turn_detection: TurnDetectionMode) -> None:
```
longcw: I think we can move this to `session.update_options`

```python
    def update_options(
        self,
        *,
        min_endpointing_delay: NotGivenOr[float] = NOT_GIVEN,
        max_endpointing_delay: NotGivenOr[float] = NOT_GIVEN,
        turn_detection: NotGivenOr[TurnDetectionMode] = NOT_GIVEN,
    ) -> None:
```

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -772,6 +772,41 @@ def update_options(
...
+        """
+
+        previous_setting = self._turn_detection
+        previous_mode = self._activity._turn_detection_mode if self._activity is not None else None
+
+        self._turn_detection = turn_detection
+
+        # propagate changes to running or pending activities
+        if self._activity is not None:
+            self._activity.update_turn_detection()
```
longcw: in AgentActivity, turn_detection is read from agent then the session
```python
    @property
    def turn_detection(self) -> TurnDetectionMode | None:
        return cast(
            "TurnDetectionMode | None",
            self._agent.turn_detection
            if is_given(self._agent.turn_detection)
            else self._session.turn_detection,
        )
```

so when the Agent has turn_detection defined, calling `session.update_turn_detection` won't actually be applied. 

this makes me wonder can you just change the `turn_detection` of the current agent and call `session.update_agent(current_agent)` to change the turn detection for the session?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -116,94 +115,126 @@ def __init__(self, agent: Agent, sess: AgentSession) -> None:
...
             )
 
-        self._mcp_tools: list[mcp.MCPTool] = []
+        return mode, turn_detector
 
-        self._on_enter_task: asyncio.Task | None = None
-        self._on_exit_task: asyncio.Task | None = None
+    def update_turn_detection(self) -> None:
+        previous_mode = self._turn_detection_mode
+        mode, turn_detector = self._determine_turn_detection_mode()
```
longcw:
```suggestion
        mode, turn_detector = self._validate_turn_detection(self.turn_detection)
```

longcw (commented):
overall looks good to me. I am wondering can you just change the `turn_detection` of the current agent and call `session.update_agent(current_agent)` to change the turn detection for the session?

livekit-agents/livekit/agents/voice/agent_activity.py
emgeee: since this method already depends on self properties I think it could go either way but happy to update the logic

livekit-agents/livekit/agents/voice/agent_session.py
emgeee: it looks like update_agent() will try and start the new agent, which will throw a RuntimeError since it would already be running.

longcw: I see, yeah will try to fix that in #3814 

> so when the Agent has turn_detection defined, calling session.update_turn_detection won't actually be applied.

this is my only concern left, maybe adding a `turn_detection_mode` in the activity which default is None?

emgeee: Good callout, I added a session.turn_detection_updated property that tracks if the session's turn detector was updated and the overrides the priority order, which should hopefully address the issue.

longcw: If you don't mind I can take this one, would like to add some refactoring on this.

emgeee: Not at all, please do!

longcw: created a pr https://github.com/livekit/agents/pull/3816, and also 
> update_agent() will try and start the new agent, which will throw a RuntimeError since it would already be running

this is fixed in #3814

PR #3768, opened 2025-10-30
changes default base url for spelling

aryeila:
Fixes typo for minimax base url, found while testing for docs.

PR #3769, opened 2025-10-31
fix minimax tts base url

longcw:


longcw:
duplicates https://github.com/livekit/agents/pull/3768

PR #3779, opened 2025-11-03
Add support for using a previous silero vad model file

zaheerabbas-prodigal:
Should help partially resolve cases like https://github.com/livekit/agents/issues/3780


livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
```diff
@@ -66,6 +67,7 @@ def load(
         activation_threshold: float = 0.5,
         sample_rate: Literal[8000, 16000] = 16000,
         force_cpu: bool = True,
+        onnx_file_path: NotGivenOr[Path] = NOT_GIVEN,
```
longcw: I think we should support str file path as well?
```suggestion
        onnx_file_path: NotGivenOr[Path | str] = NOT_GIVEN,
```

livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
```diff
@@ -120,7 +123,13 @@ async def entrypoint(ctx: JobContext):
...
             prefix_padding_duration = padding_duration
 
-        session = onnx_model.new_inference_session(force_cpu)
+        resolved_onnx_path: Path | None
+        if is_given(onnx_file_path):
+            resolved_onnx_path = onnx_file_path
+        else:
+            resolved_onnx_path = None
+
+        session = onnx_model.new_inference_session(force_cpu, resolved_onnx_path)
```
longcw:
```suggestion
        session = onnx_model.new_inference_session(force_cpu, onnx_file_path=onnx_file_path or None)
```

livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
zaheerabbas-prodigal: Allowed `str` to be passed as well

livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
zaheerabbas-prodigal: > onnx_file_path or None

Did you mean `resolved_onnx_path or None` as `onnx_file_path` can be `NotGiven` too.

longcw: `bool(NOT_GIVEN)` is False, so it will be None if it's not given.

zaheerabbas-prodigal: Ack - made the changes. Did not check that - thanks for clarifying

longcw (approved):
lgtm!

can you remove or rename the example, "custom_vad_model" may confuse ppl who what to use a vad model other than silero VAD which is actually not.

zaheerabbas-prodigal:
Removed the `custom_vad_model` script - so that it does not cause confusion

Issue #3713, opened 2025-10-24
[Bug] Ultravox RealtimeModel converts all tool parameters to strings instead of preserving JSON types


ShafiqDevs:
## **Description:**

When using the Ultravox `RealtimeModel` with MCP (Model Context Protocol) tools or any function calling that requires strict type validation, all tool parameters are being converted to strings regardless of their actual types defined in the tool schema. This causes MCP servers and other type-strict tools to reject the function calls with type validation errors.

## **Environment:**

- **LiveKit Agents Version:** Latest (as of 2025-01-24)
- **LiveKit Plugins Ultravox Version:** Latest (as of 2025-01-24)
- **Python Version:** 3.11+
- **OS:** Windows 10
- **MCP Integration:** Yes (HTTP-based MCP server with strict type validation)

## **Expected Behavior:**

Tool parameters should preserve their JSON types as defined in the tool schema:
- Arrays should remain as arrays: `["low_stock"]`
- Numbers should remain as numbers: `10`
- Booleans should remain as booleans: `true`
- Null values should remain as null: `null`

## **Actual Behavior:**

All tool parameters are converted to strings, with nested JSON being double-escaped:
- Arrays become strings: `"[\"low_stock\"]"` (double-escaped)
- Numbers become strings: `"10"`
- Booleans become strings: `"true"`
- Null values become strings: `"null"`

### 1. Set up an MCP server with a tool that has typed parameters:

```python
from livekit.agents import Agent, AgentSession, mcp
from livekit.plugins import ultravox, silero

# Example tool with typed parameters
# Tool expects: alert_type (array), limit (number), include_valuation (boolean)
agent = Agent(
    instructions="You are a helpful assistant with access to tools.",
    mcp_servers=[
        mcp.MCPServerHTTP(
            url="http://localhost:8000/mcp",  # MCP server with strict type validation
            timeout=30.0,
        )
    ],
)
```

### 2. Configure AgentSession with Ultravox RealtimeModel:

```python
session = AgentSession(
    vad=silero.VAD.load(),
    llm=ultravox.realtime.RealtimeModel(
        system_prompt="You are a helpful assistant. Use tools to answer user queries.",
        temperature=0.5,
    ),
)
```

### 3. Connect to room and start session:

```python
await session.start(room=ctx.room, agent=agent)
```

## **Error Logs:**

```
[...]

PR #3791, opened 2025-11-04
fix: coerce parameters to tool schema

omChauhanDev:
#### Summary
This PR fixes a bug where Ultravox `RealtimeModel` emitted tool parameters as strings, breaking MCP and other type-strict tools. We now coerce parameters to the correct JSON types based on each tool’s schema before emitting `FunctionCall`, including proper handling of union types with null.

- Fixes: [#3713](https://github.com/livekit/agents/issues/3713)
- Affects: `livekit-plugins-ultravox`
- Impact: Restores compatibility with MCP servers and any type-validated tool schemas

---

### Problem
Ultravox sometimes sends tool parameters as strings (e.g., `"10"`, `"true"`, `"[\"low_stock\"]"`, `"null"`). We previously forwarded them as-is via `json.dumps(event.parameters)`, causing downstream validation errors:
- Arrays were strings
- Numbers were strings
- Booleans were strings
- Optional unions with null (e.g., `["integer","null"]`) received `"null"` instead of `null`

This made Ultravox `RealtimeModel` incompatible with MCP and other type-strict integrations.

---

### What’s changed
- Add schema-based coercion to convert stringified values to the correct JSON types:
  - Arrays: parse JSON arrays, recursively coerce items
  - Numbers/integers: coerce from strings (supports `1e3`, `-inf`, etc.)
  - Booleans: coerce `"true"/"false"`, `"1"/"0"`, `"yes"/"no"`, etc.
  - Objects: parse JSON objects, recursively coerce fields
  - Null unions: when schema allows `null`, map `"null"`/`"None"` → `null` (also for `anyOf`/`oneOf`/`allOf`)
  - Fallback: best-effort coercion when schema is missing

- Wire coercion into Ultravox tool invocation path:
  - Before: `json.dumps(event.parameters)`
  - After: `json.dumps(coerced_params)` where `coerced_params` honors the tool schema

---

### Files changed
- `livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/utils.py`
  - New: `coerce_parameters_to_schema(tools_ctx, tool_name, params)` utility
  - Includes recursive coercion, union handling, and best-effort fallback
- `livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py`
  - Use `coerce_parameters_to_schema(...)` in `_handle_tool_invocation_event(...)`

---

### Before vs After
- Before (tool args forwarded as-is)
  - `alert_type`: `"[\"low_stock\"]"`
  - `limit`: `"10"`
  - `include_valuation`: `"true"`
  - `category_id`: `"null"`
- After (schema-correct types)
  - `alert_type`: `["low_stock"]`
  - `limit`: `10`
  - `include_valuation`: `true`
  - `category_id`: `null`

This aligns Ultravox behavior with OpenAI Realtime for tool calling.

---

### Screenshots
- Before: MCP validation error (string types)
image
- After: MCP validation passes (correct JSON types)
image

---

### Testing
- Manual:
  - Start the strict MCP server (as in the shared repro)
  - Run Ultravox `RealtimeModel` with the sample tool call
  - Verify MCP logs show proper types; no “Expected number/array/boolean, received string” errors

---

### Backward compatibility
- Safe: If Ultravox starts sending correctly typed parameters upstream, coercion is effectively a no-op.
- The change only affects tool invocation arguments; no changes to other Ultravox message handling.

---

### Release notes
- Ultravox: Fix tool parameter typing. Parameters now preserve JSON types per schema (arrays, numbers, booleans, null), restoring compatibility with MCP and type-strict tools.

---

### DIFF
--- a/livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
@@ -28,7 +28,7 @@ from livekit.agents.utils import is_given
 
 from ..log import logger
 from ..models import UltravoxModel, UltravoxVoice
-from ..utils import parse_tools
+from ..utils import coerce_parameters_to_schema, parse_tools
 from .events import (
     CallStartedEvent,
     ClientToolInvocationEvent,
@@ -1052,11 +1052,14 @@ class RealtimeSession(
                 f"invocationId={event.invocation_id} params_keys={list(event.parameters.keys())}"
             )
 
+        # Coerce parameters to match the tool schema so downstream consumers get correct types
+        coerced_params = coerce_parameters_to_schema(self._tools, event.tool_name, event.parameters)
+
         # Emit FunctionCall to maintain framework compatibility
         function_call = llm.FunctionCall(
             call_id=event.invocation_id,
             name=event.tool_name,
-            arguments=json.dumps(event.parameters),
+            arguments=json.dumps(coerced_params),
         )
 
         if self._current_generation is None:
--- a/livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/utils.py
+++ b/livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 from typing import Any
 
 from livekit.agents import llm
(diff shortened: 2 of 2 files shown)


theomonnom:
Could Pydantic do that automatically?

ShafiqDevs:
Any reviewers working on this? I'm so excited to test this out

omChauhanDev:
@theomonnom 
Pydantic could, but it'd require create_model() to dynamically build models from JSON schemas, then .model_validate() then .model_dump() for each call, avoided it since we're on the hot path of every tool invocation so thought that the direct approach would be some what lighter.

But if you prefer via Pydantic, I'm happy to rework it.

PR #3802, opened 2025-11-05
Chore: AssemblyAI add dynamic keyterms + return language code + normalize preemptive generation validation

dan-ince-aai:
Hi team!

Added dynamic keyterms updating without closing the stream, sends a message with type "UpdateConfiguration" https://www.assemblyai.com/docs/api-reference/streaming-api/streaming-api#send.sendUpdateConfiguration

Language code is returned when "Language_detection" is True (set now by default), and we pass this in the preflight and final transcripts, its only returned with utterances so we store it when we get it
https://www.assemblyai.com/docs/api-reference/streaming-api/streaming-api#receive.receiveTurn.language_code

Also normalized the preemptive generation text in agent_activity before validation, so we only compare if there are new words and not any changes in casing or punctuation.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1393,7 +1393,8 @@ class AgentActivity(RecognitionHooks):
             # otherwise invalidate the preemptive generation
 
             if (
-                (preemptive.info.new_transcript or "").lower() == (user_message.text_content or "").lower()
+                (preemptive.info.new_transcript or "").lower()
+                == (user_message.text_content or "").lower()
                 and preemptive.chat_ctx.is_equivalent(temp_mutable_chat_ctx)
                 and preemptive.tools == self.tools
                 and preemptive.tool_choice == self._tool_choice
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -385,7 +385,7 @@ class SpeechStream(stt.SpeechStream):
             utterance = data.get("utterance", "")
             transcript = data.get("transcript", "")
             confidence = words[-1].get("confidence", 0.0) if words else 0.0
-            
+
             # language_code is only returned with utterances, so track it for final transcript
             if "language_code" in data:
                 self._current_language_code = data["language_code"]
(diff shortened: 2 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1391,8 +1392,14 @@ async def _user_turn_completed_task(
         if preemptive := self._preemptive_generation:
             # make sure the on_user_turn_completed didn't change some request parameters
             # otherwise invalidate the preemptive generation
+
+            # Normalize text by lowercasing and removing punctuation for comparison
+            def normalize_text(text: str) -> str:
+                return text.lower().translate(str.maketrans("", "", string.punctuation))
```
longcw: should we just use `text.strip()` to remove start and end spaces, different punctuation like `?` or `!` may have different TTS results?

dan-ince-aai: @longcw in my testing, the same input to an LLM but changing between blank, ?, ! has no effect on the LLM response

longcw: IMO we shouldn't normalize the punctuations in the agents framework, maybe you can add it to the specific (validated) STT to reuse the latest transcript sent to the preemptive generation as the final version if the normalized text is the same.

dan-ince-aai: Done, thanks @longcw!

Let me know if there's anything else needed for this PR or all good to go

dan-ince-aai:
Hi @longcw @theomonnom may we have this updated PR reviewed? 

The type checks and tests are failing due to OpenAI plugin

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1391,8 +1391,10 @@ async def _user_turn_completed_task(
         if preemptive := self._preemptive_generation:
             # make sure the on_user_turn_completed didn't change some request parameters
             # otherwise invalidate the preemptive generation
+
             if (
-                preemptive.info.new_transcript == user_message.text_content
+                (preemptive.info.new_transcript or "").lower()
+                == (user_message.text_content or "").lower()
```
longcw, line 1397: I think it's better to move the text normalization to the STT plugin, you can compare the final transcript with the previous preflight version before sending it.

longcw (commented):
looks good to me! left a comment

PR #3804, opened 2025-11-05
Fix: BlingFire SentenceTokenizer merges back-to-back sentences with no space

omChauhanDev:
### Problem

The `BlingFire` sentence tokenizer was merging consecutive sentences when there was no intervening space between them. This caused two critical issues:

1. **Batch tokenization**: Two complete sentences were returned as a single token
   - Example: `"Could you please help me with your full name?What could go wrong. <break time=\"0.25s\" />."` → 1 token instead of 2

2. **Streaming delays**: In streaming mode, the first sentence wouldn't be yielded until the entire turn completed, causing significant delays in TTS generation
   - This is particularly problematic when LLM outputs back-to-back sentences (common after tool calls finish)

### Root Cause

BlingFire's `text_to_sentences_with_offsets()` requires whitespace after sentence-ending punctuation (`.`, `!`, `?`) to detect sentence boundaries. When punctuation is immediately followed by a non-space character (e.g., `"Sentence one.Sentence two"`), BlingFire treats it as a single sentence.

## Solution

Added synthetic sentence boundary detection in `_split_sentences()` that identifies sentence endings even when there's no space after punctuation. The fix:

1. **Preserves BlingFire boundaries**: Uses all original BlingFire-detected boundaries
2. **Adds safe synthetic boundaries**: Detects cases where sentence-ending punctuation (`.`, `!`, `?`) is immediately followed by:
   - `<` (SSML tags, common in voice agent outputs)
   - An uppercase letter (typical sentence start)

3. **Normalizes boundaries**: Sorts and deduplicates all boundaries to ensure correct sequential processing
4. **Maintains existing behavior**: Preserves `min_sentence_len` filtering and `retain_format` options

### Technical Details

The fix augments BlingFire's offset detection with additional boundary points:

```python
# BlingFire boundaries are not always correct, so we add additional boundaries 
# where punctuation is immediately followed by '<' or a capital letter.
ends: list[int] = [end for _, end in offsets]
for m in re.finditer(r"[.!?](?=(?:<|[A-Z]))", text):
    ends.append(m.end())

# Normalize and sort unique boundaries
ends = sorted({e for e in ends if 0 < e <= len(text)})
```

The regex pattern `r"[.!?](?=(?:<|[A-Z]))"` uses a positive lookahead to match:
- Sentence-ending punctuation (`.`, `!`, `?`)
- Immediately followed by either `<` or an uppercase letter `[A-Z]`

This conservative approach avoids false positives in common cases like:
- Email addresses: `user@example.com` (not split)
- Decimal numbers: `3.14` (not split)
- URLs: `https://example.com` (not split)
- Abbreviations: `Mr.Smith` (not split, as lowercase follows)

### Before Fix

image

### After Fix

image

### DIFF
--- a/livekit-agents/livekit/agents/tokenize/blingfire.py
+++ b/livekit-agents/livekit/agents/tokenize/blingfire.py
@@ -18,10 +18,20 @@ def _split_sentences(
 ) -> list[tuple[str, int, int]]:
     _, offsets = blingfire.text_to_sentences_with_offsets(text)
 
+    # Added boundaries where punctuation is immediately followed by '<' or a capital letter.
+    ends: list[int] = [end for _, end in offsets]
+    for m in re.finditer(r"[.!?](?=(?:<|[A-Z]))", text):
+        ends.append(m.end())
+
+    # Normalize and sort unique boundaries
+    ends = sorted({e for e in ends if 0 < e <= len(text)})
+
     merged_sentences = []
     start = 0
 
-    for _, end in offsets:
+    for end in ends:
+        if end <= start:
+            continue
         raw_sentence = text[start:end]
         sentence = re.sub(r"\s*\n+\s*", " ", raw_sentence).strip()
         if not sentence or len(sentence) < min_sentence_len:


davidzhao:
AI generated code with breaking tests

PR #3808, opened 2025-11-06
Inference: Allow provider specific parameter updates

adrian-cowham:
I added `extra_kwargs` to `update_options` to mirror the constructor. Feedback welcomed.


adrian-cowham:
I'll look into the type check error.

theomonnom (approved):

PR #3816, opened 2025-11-07
allow turn detection mode to be updated within session

longcw:
refactor https://github.com/livekit/agents/pull/3765

example usage:
```python
session.update_options(turn_detection="manual")
```

### DIFF
--- a/livekit-agents/livekit/agents/beta/workflows/address.py
+++ b/livekit-agents/livekit/agents/beta/workflows/address.py
@@ -11,7 +11,7 @@ from ...voice.events import RunContext
 from ...voice.speech_handle import SpeechHandle
 
 if TYPE_CHECKING:
-    from ...voice.agent_session import TurnDetectionMode
+    from ...voice.audio_recognition import TurnDetectionMode
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


emgeee:
looks good!

theomonnom:
Since `turn_detection` is configured per `Agent`, should updates be made through the `session` object? rather than setting it directly for each `Agent` instance, like `agent.turn_detection = "manual"`?

longcw:
> Since `turn_detection` is configured per `Agent`, should updates be made through the `session` object? rather than setting it directly for each `Agent` instance, like `agent.turn_detection = "manual"`?

we also need to update it for AudioRecognition, so maybe add a setter for `agent.turn_detection`?

emgeee:
Any progress on this?

theomonnom:
> we also need to update it for AudioRecognition, so maybe add a setter for `agent.turn_detection`?

Yes, a setter makes sense? Or maybe we can also introduce `update_options`?

longcw:
> > we also need to update it for AudioRecognition, so maybe add a setter for `agent.turn_detection`?
> 
> Yes, a setter makes sense? Or maybe we can also introduce `update_options`?

updated.

PR #3823, opened 2025-11-07
feat(google): add denoiser configuration parameter to Google STT class

LastMonopoly:
Adds support for Google Cloud Speech-to-Text denoiser configuration to improve recognition accuracy for noisy audio.

- Adds `denoiser_config` parameter to STT constructor
- Properly documented and typed
- Passes all CI checks


LastMonopoly:
Justification
- `uv.lock` file is changed a lot mainly due to newer version of uv always adds " upload-time" fields when making changes
- I have to update dependencies because I want to upgrade `google-cloud-speech` from "2.31.0" to "2.34.0" which supports `denoiser_config` (the whole purpose of this PR)

Reference
- https://cloud.google.com/speech-to-text/docs/models/chirp-3#enable_denoiser_and_snr-filtering
- https://docs.cloud.google.com/python/docs/reference/speech/latest/google.cloud.speech_v2.types.DenoiserConfig

PR #3825, opened 2025-11-08
Close TTS connections when closing the session

adrian-cowham:
I have an agent that transfers to other agents, during testing I unexpectedly hit the Inference TTS quota. After digging around a bit I discovered that after every transfer, the number of TTS connections to my dev Inference service kept increasing.  

As far as I can tell, this only happens for TTS plugins that use a connection pool. 

I tested this change with TTS plugins that do **and** do not use a connection pool. In the case of TTS plugins that do not use a connection pool.

This doesn't happen for STT connections since those get closed [here](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/voice/agent_activity.py#L676-L677) .

Feedback welcome, let me know if there's an alternative way to handle this situation.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -644,6 +644,7 @@ class AgentActivity(RecognitionHooks):
         if isinstance(self.llm, llm.LLM):
             self.llm.off("metrics_collected", self._on_metrics_collected)
             self.llm.off("error", self._on_error)
+            await self.llm.aclose()
 
         if isinstance(self.llm, llm.RealtimeModel) and self._rt_session is not None:
             self._rt_session.off("generation_created", self._on_generation_created)
@@ -659,6 +660,7 @@ class AgentActivity(RecognitionHooks):
         if isinstance(self.stt, stt.STT):
             self.stt.off("metrics_collected", self._on_metrics_collected)
             self.stt.off("error", self._on_error)
+            await self.stt.aclose()
 
         if isinstance(self.tts, tts.TTS):
             self.tts.off("metrics_collected", self._on_metrics_collected)
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -663,6 +663,7 @@ class AgentActivity(RecognitionHooks):
         if isinstance(self.tts, tts.TTS):
             self.tts.off("metrics_collected", self._on_metrics_collected)
             self.tts.off("error", self._on_error)
+            await self.tts.aclose()
 
         if isinstance(self.vad, vad.VAD):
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -663,6 +663,7 @@ async def _close_session(self) -> None:
         if isinstance(self.tts, tts.TTS):
             self.tts.off("metrics_collected", self._on_metrics_collected)
             self.tts.off("error", self._on_error)
+            await self.tts.aclose()
```
longcw, line 668: should we also close the STT and LLM?

adrian-cowham: Hi Long! The call to `self._audio_recognition.aclose()` will close the STT, but given the `None` check, perhaps that's not guaranteed. Happy to also close the STT and LLM, let me add that and do some testing.

adrian-cowham: I updated the PR to also close the LLM and STT. Let me know what you think.

longcw: I just realized the `stt` `tts` and `llm` are reused when Agent handoff if they are set in the AgentSession, it seems we need an api to close the existing connections but not the instance so they can be reused?

longcw: Tho it actually works like that, for openai, cartesia, deepgram, only `tts` has an `aclose` implemented and it's indeed closing the connections but not the instance. not sure how other plugins are implemented.

`aclose` sounds like it will close the instance and may make it not usable. should we add a new method like `close_streams` for them? cc @theomonnom wdyt?

theomonnom: When the AgentActivity closes, we shouldn't close the TTS/STT. Only the stream being used for the current activity. (But AudioRecognition should already do it).

If we use close on the stt/tts instance, it'll also close all the connections inside the Pool, so the reconnection will be slower on every agent handoff

longcw: so the issue is the `tts_node` and `stt_node` may not be closed after interruption or agent handoff?

for example, tts_node is created in https://github.com/livekit/agents/blob/livekit-agents@1.3.4/livekit-agents/livekit/agents/voice/generation.py#L230, it's a `AsyncGenerator` but we never call `tts_node.aclose()`?

theomonnom (approved):
Approving this PR, we can optimize the connection re-use in the future PRs

livekit-agents/livekit/agents/voice/agent_activity.py
theomonnom: One issue is that new agents are creating new instances of the inference.TTS, so we're having more than one connection pool

davidzhao:
@adrian-cowham is this still needed?

PR #3844, opened 2025-11-10
feat(llm/groq): add disable_tool_validation option to LLM configuration

youssmox:
## Summary
- Added support for propagating the Groq-specific `disable_tool_validation` flag through `livekit.plugins.openai.LLM`, storing it in `_LLMOptions`.
- Forward the flag in `LLM.chat` only when the client’s base URL points at Groq, raising a clear error otherwise to avoid misconfiguration.
- This ensures the Groq plugin can opt out of tool validation without breaking other OpenAI-compatible providers.

## Testing
- Not run (feature is configuration-only; recommend running Groq integration test/smoke before merge).


chenghao-mou:
Thanks for creating the PR! 

This flag should be an extra [body parameter](https://console.groq.com/docs/api-reference#chat-create-request-body) instead of a function parameter in `extra/extra_kwargs`. Otherwise, it will throw errors like:

```
TypeError: AsyncCompletions.create() got an unexpected keyword argument 'disable_tool_validation'
```

PR #3847, opened 2025-11-10
fix pickling AgentServer for python 3.9

longcw:
`asyncio.Lock()` is not pickleable in python 3.9 with error `AttributeError: Can't pickle local object 'WeakSet.__init__.<locals>._remove'` (https://github.com/python/cpython/issues/74876). Actually the `WeakSet` is still not pickleable in later pythons, but it seems `asyncio.Lock` is fixed after python 3.10.

this causing `python agent.py dev` failed since AgentServer now is an arg of `watchfiles.arun_process` when watchfiles enabled.

in addition to this one, https://github.com/livekit/agents/pull/3842 is needed for running in python 3.9

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -332,6 +332,21 @@ class AgentServer(utils.EventEmitter[EventTypes]):
 
         self._lock = asyncio.Lock()
 
+    if sys.version_info < (3, 10):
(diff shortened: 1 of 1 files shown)


chenghao-mou (approved):
Do we need to persist lock state? If not, LGTM.

longcw:
> Do we need to persist lock state? If not, LGTM.

yeah, I think we don't need to persist it.

PR #3906, opened 2025-11-12
add stereo audio recorder and example

chenghao-mou:
**Background**

This is the first step in barge-in integration. This PR adds callbacks for RecorderIO to access input audio frames every `_write_interval` second(s).

Sample user recording made with the commented callback:
[history.mp3](https://github.com/user-attachments/files/23626756/history.mp3)


examples/voice_agents/stereo_audio.py
davidzhao, line 1: is the main purpose of this component for internal use / barge-in model? if so it'd be better to make it private instead of exposing in the public API. and we probably won't need an example either

chenghao-mou: Yes, I will make it private. The example is included only for the reviewing/testing. I will remove it before merging.

Issue #3237, opened 2025-08-22
When using an Avatar + Gemini Realtime = Error when interrupting


mercuryyy:
This is very hard to repreduce i'v tested this on various setups and tried finding why this is happening but i think it might have to do with the avatar framework.

Only when an avatar is enabled and on gemini realtime when we interrupt the agent we get 

2025-08-22 15:08:52,614 - WARNING livekit.plugins.google - truncate is not supported by the Google Realtime API. {"pid": 28580, "job_id": "AJ_YfHLDpCzEMao"}

which is normal since gemini does not support cutting the response short but this causes the avatar+llm to freeze up and not recover.

longcw:
which avatar provider are you using, and does that work for other LLMs?

mercuryyy:
using "hedra" This also happens with OpenAI realtime but not as consistently as with Gemini realtime. 

This does not happen when we setup Gemini realtime to use  elevanlabs (or other providers)  for TTS pipeline.


I also tested this on 1.2.1 - 1.2.6

mercuryyy:
@longcw i have tested with various combinations and this happens only with Gemini realtime on full realtime pipeline once you add a TTS provider it does not happen anymore.

Also this sometimes happens even without the avatar.

Finally when this happens the chat transcript keeps going the agent is receiving and responding (inside the chat context) but the audio does not play, this  happens more often when not speaking english.  

I have tried so many different logging steps and debugs i cant understand why this is happening mainly because it keeps responding in the chat context but the audio do not play.

longcw:
it sounds an issue of the gemini realtime? can you share the logs and agent configurations

Issue #3434, opened 2025-09-16
Avatar stops receiving audio after several consecutive interruptions


ChenghaoMou:
Need some help on getting on the right track to debug this issue:

I am observing a pattern where the agent has been interrupted several times in a roll, it stops sending the audio. I can confirm that:
1. The avatar never triggers the logging statement I placed in push_audio function to show audio is being pushed.
2. A previous turn before all these interruptions is finished with the correct "playback finished" log.
3. All metrics seem normal. TTS metrics are still reported, but the audio is not received by the avatar.
4. No errors reported either from Livekit or the avatar.

Where should I start looking to debug this? Thanks in advance.

```
livekit                                  1.0.12
livekit-agents                           1.2.6
livekit-api                              1.0.5
livekit-blingfire                        1.0.0
livekit-plugins-anthropic                1.2.6
livekit-plugins-assemblyai               1.2.6
livekit-plugins-cartesia                 1.2.6
livekit-plugins-deepgram                 1.2.6
livekit-plugins-elevenlabs               1.2.6
livekit-plugins-google                   1.2.6
livekit-plugins-noise-cancellation       0.2.5
livekit-plugins-openai                   1.2.6
livekit-plugins-silero                   1.2.6
livekit-plugins-speechmatics             1.2.6
livekit-plugins-turn-detector            1.2.6
livekit-protocol                         1.0.5
```

Maybe related: #2826

ChenghaoMou:
One thing I noticed is that one of the TTS metrics were reported much later than expected: 

<img width="1071" height="491" alt="Image" src="https://github.com/user-attachments/assets/5b75fa01-cf93-454b-8609-dd0a2fb2cdd8" />

Not sure why the first one here is not cancelled:

```
TTSMetrics(type='tts_metrics', label='livekit.plugins.cartesia.tts.TTS', request_id='76640224f8df', timestamp=1758026777.797612, ttfb=0.2746591499890201, duration=9.530534961988453, audio_duration=26.424249999999937, cancelled=False, characters_count=486, streamed=True, segment_id='e8d722a66bf8', speech_id='speech_9eed13845e5e')

[...]

PR #3917, opened 2025-11-13
add timeout to datastream clear_buffer to avoid deadlock when missing playback finished event

longcw:
when playback finished event is not received from the avatar service, the agent may hang forever to wait for the event. this usually happens during interruption due to some race conditions in the avatar service or network issues.

fix https://github.com/livekit/agents/issues/3237 and https://github.com/livekit/agents/issues/3434

### DIFF
--- a/livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
+++ b/livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
@@ -187,6 +187,9 @@ class DataStreamAudioOutput(AudioOutput):
             self.on_playback_finished(playback_position=pushed_duration, interrupted=True)
             self._reset_playback_count()
 
+        if self._clear_buffer_timeout_handler:
+            self._clear_buffer_timeout_handler.cancel()
+
         if timeout is not None:
             self._clear_buffer_timeout_handler = asyncio.get_event_loop().call_later(
                 timeout, _on_timeout


davidzhao (approved):
lgtm!

livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
```diff
@@ -168,6 +168,30 @@ def pause(self) -> None:
...
+            )
+            self.on_playback_finished(playback_position=pushed_duration, interrupted=True)
+            self._reset_playback_count()
+
+        if timeout is not None:
+            self._clear_buffer_timeout_handler = asyncio.get_event_loop().call_later(
+                timeout, _on_timeout
+            )
+
     def _handle_playback_finished(self, data: rtc.RpcInvocationData) -> str:
```
chenghao-mou, line 198: QQ: Could there be a race condition here? If this function gets called after `await self._room.local_participant.perform_rpc` completes but before `self._clear_buffer_timeout_handler = asyncio.get_event_loop().call_later(` executes, would that trigger multiple on_playback_finished callbacks?

longcw: nice catch! it could be if the playback finished received before clear buffer rpc done... will add a check in the timeout callback.

longcw: checked the code, there is a check in `on_playback_finished` to ignore the redundant events, and ideally this shouldn't happen if avatar service handles the `clear_buffer` RPC before sending the playback finished event.

chenghao-mou: Got it. Thanks!

chenghao-mou (approved):
LGTM!

theomonnom (approved):

PR #3920, opened 2025-11-13
chore: move LK env var checks later

davidzhao:
allows download-files to be completed without needing env


chenghao-mou (approved):
LGTM. Tested with a successful build and deployment.

PR #3932, opened 2025-11-14
Added Support for gpt-5.1-chat-latest

### LINKED ISSUE #3931: gpt-5.1-chat-latest model support
### Bug Description

gpt-5.1-chat-latest isn't present in listed models 

### Expected Behavior

gpt-5.1-chat-latest model should be added

### Reproduction Steps

```bash
gpt-5.1-chat-latest model support isn't there
```

### Operating System

macOS

### Models Used

_No response_

### Package Versions

```bash
.
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

devb-enp:
Fixes #3931

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
@@ -20,6 +20,7 @@ TTSVoices = Literal[
 DalleModels = Literal["dall-e-2", "dall-e-3"]
 ChatModels = Literal[
     "gpt-5.1",
+    "gpt-5.1-chat-latest",
     "gpt-5",
     "gpt-5-chat-latest",
     "gpt-5-mini",


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/models.py
```diff
@@ -20,6 +20,7 @@
 DalleModels = Literal["dall-e-2", "dall-e-3"]
 ChatModels = Literal[
     "gpt-5.1",
+    "gpt-5.1-chat-latest",
```
chenghao-mou, line 23: You might want to add this in `_supports_reasoning_effort` and update [this line](https://github.com/livekit/agents/blob/11a7861a03411cb789ae1c41c876701d442e9446/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py#L121) to something like `if "gpt-5.1" in model:`

devb-enp: gpt-5.1-chat-latest does not support reasoning https://platform.openai.com/docs/models/compare?model=gpt-5.1-chat-latest @chenghao-mou

chenghao-mou: Thanks for the information. I was looking at the [wrong page](https://platform.openai.com/docs/models/gpt-5.1-chat-latest).

Issue #2417, opened 2025-05-28
How to force speech synthesis for partial `llm_node` result?


furious-luke:
Hi there!

I am working on a custom `llm_node` method on one of my agents that performs a sequence of LLM calls for various purposes and stitches them together to create a final output to send to the TTS service. Before I send the final outputs, I want to be able to send some speech to be played while the rest of the LLM calls occur. However, I've found that the speech will not begin; it seems it's waiting for a threshold of available text to be generated.

As an example, this demonstrates largely what I'm aiming for...

```python
class MyAgent(Agent):
    def llm_node(...):
        # Perform an initial LLM call and get some text.
        response = await self.call_some_fast_llm()
        # Send this to the TTS service.
        yield response
        # TODO: How to force the TTS service to say the text that's been sent?
        remaining = await self.call_some_slow_llm()
        # Send the remaining text to the TTS service.
        yield remaining
        # Normally it's only now the speech begins.
```

Is there a way to force the TTS to begin playing the provided text while waiting for the rest to generate, all from within the `llm_node` method? Or perhaps there's a better way to accomplish this?

iandoesallthethings:
> For now, this isn’t supported — the llm_node does not have a “Flush” capability. Attempt [#2023](https://github.com/livekit/agents/pull/2023) was an effort to implement this. We do, however, have a related example that uses [on_user_turn_completed](https://github.com/livekit/agents/blob/main/examples/voice_agents/fast-preresponse.py) to inject “pre-messages” from fast LLMs.

I'm trying to do something similar, triggering buffer responses while rag searching in the background. `on_user_turn_completed` works fine for voice, but text messages don't trigger this event. Is that intentional?

It seems like text messages should count as a user turn for accessibility reasons.

longcw:
@iandoesallthethings you can customize the text input callback from `RoomInputOptions.text_input_cb` https://github.com/livekit/agents/blob/livekit-agents%401.1.0/livekit-agents/livekit/agents/voice/room_io/room_io.py#L76

iandoesallthethings:
> [@iandoesallthethings](https://github.com/iandoesallthethings) you can customize the text input callback from `RoomInputOptions.text_input_cb` https://github.com/livekit/agents/blob/livekit-agents%401.1.0/livekit-agents/livekit/agents/voice/room_io/room_io.py#L76

Ah, I see. So if I want text and voice to be treated the same, I just have to call the same function in on_user_turn_completed and text_input_cb?

[...]

Issue #3798, opened 2025-11-04
BufferedTokenStream never yields the completed tokens (sentence or word) till the next token arrives causing delays in TTS Speech generation


zaheerabbas-prodigal:
### Bug Description

Issue 1:
Streaming tokenizers in [`livekit.agents.tokenize.token_stream.BufferedTokenStream`](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/tokenize/token_stream.py#L14) never emit the first token unless a second token appears. Ref [code](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/tokenize/token_stream.py#L44)

With real LLM output (multi‑second gaps between tokens in some cases) this means Cartesia/ElevenLabs TTS stays silent until another chunk arrives. 

If we relax the [guard](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/tokenize/token_stream.py#L44) to allow a single token, the loop spins forever because sentence/word tokenizers return inclusive end indices (tok[2]), so the buffer slice never shrinks.

Issue 2:
blingfire.SentenceTokenizer treats back‑to‑back sentences with no intervening space as a single token. So when the LLM outputs something like `Sentence one.Sentence two` - common when tool calls finish and new LLM request is made - BlingFire glues everything together, and the streaming path won’t yield a sentence until the entire turn finishes causing further delay in speech generation. 

As an example run the tokenization script I have shared below and see the output as `Could you please help me with your full name?What could go wrong. <break time="0.25s" />.` even though there are 2 complete sentences here and 1 SSML tag but are sent to TTS as a single complete sentence. Please [refer this](https://gist.github.com/zaheerabbas-prodigal/e37077fb6f46484a66ba43ca75dc84d6#:~:text=%5Bblingfire.SentenceTokenizer%5D%5B%20%206.116s%5D%20output%20token%3D%27Could%20you%20please%20help%20me%20with%20your%20full%20name%3FWhat%20could%20go%20wrong.%20%3Cbreak%20time%3D%220.25s%22%20/%3E.%27%20(segment%3D54fa17021f32))

### Expected Behavior

Issue 1:
As soon as a tokenizer detects a complete token - whether that’s a sentence, a word - it should be emitted immediately, even if it’s the only token so far and not depend on the look-ahead logic as the delay from LLM chunks can be in certain cases.

Issue 2:
[...]

Issue #3529, opened 2025-09-29
How to force speech synthesis for partial llm_node result?


muaaz0333:

longcw:
duplicates https://github.com/livekit/agents/issues/2417

here is an attempt to add `flush` ability for `llm_node` https://github.com/livekit/agents/pull/2023, but it's not finalized.

longcw:
it was supported in https://github.com/livekit/agents/pull/3933

PR #3933, opened 2025-11-14
add flush for llm_node

longcw:
replace https://github.com/livekit/agents/pull/2023

close https://github.com/livekit/agents/issues/2417, https://github.com/livekit/agents/issues/3798, https://github.com/livekit/agents/issues/3529

### DIFF
--- a/examples/voice_agents/flush_llm_node.py
+++ b/examples/voice_agents/flush_llm_node.py
@@ -8,6 +8,7 @@ from dotenv import load_dotenv
 from livekit.agents import (
     Agent,
     AgentSession,
+    FlushSentinel,
     JobContext,
     MetricsCollectedEvent,
     ModelSettings,
@@ -71,7 +72,7 @@ class FastResponseAgent(Agent):
         chat_ctx: llm.ChatContext,
         tools: list[llm.FunctionTool],
         model_settings: ModelSettings,
-    ) -> AsyncIterable[llm.ChatChunk | llm.FlushSentinel]:
+    ) -> AsyncIterable[llm.ChatChunk | FlushSentinel]:
         called_tools: list[llm.FunctionToolCall] = []
         has_text_message = False
         async for chunk in Agent.default.llm_node(
@@ -95,7 +96,7 @@ class FastResponseAgent(Agent):
             yield "One moment while I look that up. "
             # flush the response to tts immediately
             # NOTE: this will close the current tts_node and start a new one
-            yield llm.FlushSentinel()
+            yield FlushSentinel()
 
             # simulate additional processing before closing the llm_node
             await asyncio.sleep(3)
--- a/livekit-agents/livekit/agents/llm/llm.py
+++ b/livekit-agents/livekit/agents/llm/llm.py
@@ -75,10 +75,6 @@ class LLMError(BaseModel):
     recoverable: bool
 
 
-class FlushSentinel:
-    pass
-
-
 TEvent = TypeVar("TEvent")
 
 
--- a/livekit-agents/livekit/agents/__init__.py
+++ b/livekit-agents/livekit/agents/__init__.py
@@ -51,6 +51,7 @@ from .types import (
     DEFAULT_API_CONNECT_OPTIONS,
     NOT_GIVEN,
     APIConnectOptions,
+    FlushSentinel,
     NotGiven,
     NotGivenOr,
 )
@@ -170,6 +171,7 @@ __all__ = [
     "BuiltinAudioClip",
     "AudioConfig",
     "PlayHandle",
+    "FlushSentinel",
     "io",
     "avatar",
     "cli",
--- a/livekit-agents/livekit/agents/llm/__init__.py
+++ b/livekit-agents/livekit/agents/llm/__init__.py
@@ -18,7 +18,6 @@ from .llm import (
     ChatChunk,
     ChoiceDelta,
     CompletionUsage,
-    FlushSentinel,
     FunctionToolCall,
     LLMError,
     LLMStream,
@@ -94,7 +93,6 @@ __all__ = [
     "RealtimeSessionReconnectedEvent",
     "RealtimeSessionRestoredEvent",
     "LLMError",
-    "FlushSentinel",
 ]
 
 # Cleanup docs of unexported modules
--- a/livekit-agents/livekit/agents/types.py
+++ b/livekit-agents/livekit/agents/types.py
@@ -34,6 +34,10 @@ The key for the timed transcripts in the audio frame userdata.
 _T = TypeVar("_T")
 
 
+class FlushSentinel:
+    pass
+
+
 class NotGiven:
     def __bool__(self) -> Literal[False]:
         return False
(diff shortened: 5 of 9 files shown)


livekit-agents/livekit/agents/llm/_provider_format/openai.py
```diff
@@ -35,6 +35,9 @@ def to_chat_ctx(
         for tool_output in group.tool_outputs:
             messages.append(_to_chat_item(tool_output))
 
+    import pprint
```
chenghao-mou: Image

longcw: forgot to remove it lol

examples/voice_agents/flush_llm_node.py
```diff
@@ -0,0 +1,104 @@
...
+        self.fast_llm_prompt = llm.ChatMessage(
+            role="system",
+            content=[
+                "Generate a short instant response to the user's message with 5 to 10 words.",
+                "Do not answer the questions directly. For example, let me think about that, "
+                "wait a moment, that's a good question, etc.",
+            ],
+        )
+
+    async def llm_node(
```
davidzhao, line 70: how does this work as part of a function call? I think that's the primary example to show:
- tool call is triggered
- synthesizing a fast response (and feeding them into the LLM), flushing it
- when function call returns, our pipeline generates another response

examples/voice_agents/flush_llm_node.py
```diff
@@ -0,0 +1,104 @@
...
+        if quick_response:
+            yield "\n"
+            logger.info(f"quick response: {quick_response}")
+
+            # flush the quick response to tts immediately
+            # NOTE: this will close the current tts_node and start a new one
+            yield llm.FlushSentinel()
+
+            # (Optional) add the quick response to the chat ctx for the main llm
+            assert isinstance(self.llm, llm.LLM)
```
davidzhao: self.llm might not be defined, if it's defined on AgentSession

why do we need this assert here?

examples/voice_agents/README.md
```diff
@@ -28,6 +28,7 @@ This directory contains a comprehensive collection of voice-based agent examples
 ### 🎯 Pipeline Nodes & Hooks
 
 - [`fast-preresponse.py`](./fast-preresponse.py) - Generating quick responses using the `on_user_turn_completed` node
+- [`flush_llm_node.py`](./flush_llm_node.py) - Flushing partial LLM output to TTS in `llm_node`
```
davidzhao, line 31: think we should combine these examples, and show users how to generate a fast pre-response in one.. 

having both is a bit confusing. there should be one recommended path for doing this

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -76,6 +76,10 @@ class LLMError(BaseModel):
     recoverable: bool
 
 
+class FlushSentinel:
```
theomonnom: Do you think this should be inside the llm module vs a global one?

examples/voice_agents/flush_llm_node.py
longcw: in pervious example the llm is defined in Agent, adding assert here to pass the type check

examples/voice_agents/flush_llm_node.py
longcw: updated the example.

also reverted https://github.com/livekit/agents/pull/2787 to be able to execute the tools before llm_node completed, using `created_at` to keep the order of the RunResult instead. cc @theomonnom

livekit-agents/livekit/agents/llm/llm.py
longcw: I have no strong opinion, is there any other flush use case?

davidzhao: maybe export from the global namespace? so folks are importing it from the main namespace instead of `llm.FlushSentinel` ?

examples/voice_agents/flush_llm_node.py
davidzhao: how does the user decide if they should be using fast response or not? it seems that with this example, each request is going to end up triggering a `One moment while I look that up`

it seems like the real use cases will be driven by a tool call, where it may be useful to notify users for certain tools that would take awhile to process.

longcw: In this example it only triggers when there is a tool call, and they can also add conditions on tool call names and args.
```python
            if (
                not fast_response_sent
                and isinstance(chunk, llm.ChatChunk)
                and chunk.delta
                and chunk.delta.tool_calls
            ):
                yield "One moment while I look that up. "
                # flush the response to tts immediately
                # NOTE: this will close the current tts_node and start a new one
                yield llm.FlushSentinel()
```

davidzhao: got it, I missed the tool_calls part initially. this makes sense. so there seem to be three different ways of doing this

1. using llm_node to detect when the right tool is being called, then yielding a string segment (or other llm chunks) along with a flush sentinel
2. inside the tool call, using `session.say` to yield a static response
3. inside the tool call, using `session.generate_reply` to generate a dynamic response

how does a user decide which one they should be using?

longcw: for simple use cases, like a quick response for a tool call, users can use `session.say` or `generate_reply` directly in the tool IMO.

flush in `llm_node` is for more customized cases, for example, if users only want a quick response for the tool calls that without text message (sometimes LLM response a text message with the tool call), they can check that in the custom `llm_node` and flush the quick response there.

davidzhao: makes sense to me. does it make sense to make it clear in this example, that it's only generating one if:
- the tool matches `get_weather`
- the LLM did not return any existing text (to avoid duplicating responses)

livekit-agents/livekit/agents/llm/llm.py
theomonnom: yeah I think the global namespace makes more sense.

theomonnom: > I have no strong opinion, is there any other flush use case?

It's only OK if we don't think we will need to add any other "arguments" to this llm_flush (e.g `yield FlushSentinel(force=True)`)

livekit-agents/livekit/agents/voice/run_result.py
```diff
@@ -42,25 +43,38 @@
 class ChatMessageEvent:
     item: llm.ChatMessage
     type: Literal["message"] = "message"
+    created_at: float = field(default_factory=time.time)
```
theomonnom: I think we may not need to duplicate `create_at`, we not have an AgentHandoff item in 1.3

livekit-agents/livekit/agents/voice/run_result.py
theomonnom, line 1: Btw is it related to this PR? Or it's safe to ignore in my review?

longcw: it's needed to allow tool execution before `llm_node` returned (for the example added). right now the tool execution will wait for the text generation, i.e. llm_node done, to keep the order of RunResults.

longcw: I split that part to https://github.com/livekit/agents/pull/3962

PR #3944, opened 2025-11-14
New Tools: HTTP pre-built tools

Hormold:
I'd like to propose two new beta tools that address common patterns I've seen when building voice agents.

~~1. SMS Tool - Many voice agents need to send text messages during calls (order confirmations, addresses, booking links, etc.). This tool provides a simple way to send SMS via Twilio, SignalWire, or Vonage. It can auto-detect the caller's phone number from the SIP session or use pre-configured recipients. The goal is to reduce boilerplate for this common use case while keeping it flexible enough for different workflows.~~
UPD: **SMS tools moved to another branch to separate discussion**

HTTP Tool - A lot of agents need to call external APIs - webhooks, CRMs, internal services, etc. Rather than writing custom functions for each HTTP call, this tool lets you declaratively configure requests with parameters, headers, timeouts, and response handling. The idea is to make simple API integrations faster to implement while still supporting customization when needed.

Both tools include error handling, provider auto-detection (SMS), and parameter validation. I've included tests and an example with the weather agent showing how they work together.

Open to feedback on the API design, naming, and what additional features might be useful!


livekit-agents/livekit/agents/beta/tools/http/send_http.py
```diff
@@ -0,0 +1,306 @@
...
+    url: str,
+    method: str,
+    headers: dict[str, str] | None,
+    timeout_ms: int,
+    arguments: dict[str, Any],
+) -> HTTPResponse:
+    """Make HTTP request with given parameters.
+
+    Args:
+        url: Endpoint URL
```
bcherry: in agent builder work I found that it's a must-have to support templating arguments into named path segments if this is going to work for any REST API. so this shouldn't be an actual URL but should be a URL template, and we'd need a way to partition the passed arguments into those that go into the URL and those that go into the payload

livekit-agents/livekit/agents/beta/tools/http/send_http.py
```diff
@@ -0,0 +1,306 @@
...
+    async def http_handler(raw_arguments: dict, context: RunContext) -> str | None:
+        if config.execution_message:
+            await _announce_execution(tool_name, config.execution_message, context.session)
+
+        response = await _make_http_request(
+            url=config.url,
+            method=config.method,
+            headers=config.headers,
+            timeout_ms=config.timeout_ms,
+            arguments=raw_arguments,
```
bcherry: it's important to sanitize the raw_arguments and only call the tool with the arguments actually specified in the schema, to prevent the LLM from accidentally loading extra unspecified params in

livekit-agents/livekit/agents/beta/tools/http/send_http.py
```diff
@@ -0,0 +1,306 @@
...
+
+    if not isinstance(parameters_schema, dict) or "type" not in parameters_schema:
+        logger.error(
+            "HTTP tool creation failed: invalid parameters schema",
+            extra={"schema_type": type(parameters_schema).__name__},
+        )
+        raise ValueError("Invalid parameters schema - must be a dict with 'type' field")
+
+    tool_name = config.name or "http_request"
+    tool_description = config.description or f"Make {config.method} request to {config.url}"
```
bcherry: i don't think a default description makes sense - if the dev leaves it blank, it should remain blank

livekit-agents/livekit/agents/beta/tools/http/send_http.py
```diff
@@ -0,0 +1,306 @@
...
+    parameters_schema = config.parameters or {"type": "object", "properties": {}, "required": []}
+
+    if not isinstance(parameters_schema, dict) or "type" not in parameters_schema:
+        logger.error(
+            "HTTP tool creation failed: invalid parameters schema",
+            extra={"schema_type": type(parameters_schema).__name__},
+        )
+        raise ValueError("Invalid parameters schema - must be a dict with 'type' field")
+
+    tool_name = config.name or "http_request"
```
bcherry: tool name should be required

livekit-agents/livekit/agents/beta/tools/http/send_http.py
```diff
@@ -0,0 +1,306 @@
...
+            url="https://api.weather.com/current",
+            method="GET",
+            parameters={
+                "type": "object",
+                "properties": {
+                    "city": {"type": "string"}
+                },
+                "required": ["city"]
+            },
+            output_normalizer=lambda resp: f"Temp: {resp.body}" if resp.success else None
```
bcherry: this syntax seems pretty verbose - i wonder if there's a more pythonic way to do it?

livekit-agents/livekit/agents/beta/tools/http/send_http.py
```diff
@@ -0,0 +1,306 @@
...
+        timeout_ms: Timeout in milliseconds
+        arguments: Parameters to send
+
+    Returns:
+        HTTPResponse with success/failure details
+    """
+    method = method.upper()
+    timeout = aiohttp.ClientTimeout(total=timeout_ms / 1000.0)
+
+    async with aiohttp.ClientSession(timeout=timeout) as session:
```
bcherry: this should use `utils.http_session`

livekit-agents/livekit/agents/beta/tools/http/send_http.py
```diff
@@ -0,0 +1,306 @@
...
+            if method == "GET":
+                resp_ctx = session.get(url, params=arguments, headers=headers)
+            elif method == "POST":
+                resp_ctx = session.post(url, json=arguments, headers=headers)
+            elif method == "PUT":
+                resp_ctx = session.put(url, json=arguments, headers=headers)
+            elif method == "PATCH":
+                resp_ctx = session.patch(url, json=arguments, headers=headers)
+            else:  # DELETE
+                resp_ctx = session.delete(url, params=arguments, headers=headers)
```
bcherry: this should specify DELETE explicitly, and throw an error i the method is unknown

Hormold: it's validated in tool creation

livekit-agents/livekit/agents/beta/tools/http/send_http.py
```diff
@@ -0,0 +1,306 @@
...
+    Attributes:
+        url: Endpoint URL (required)
+        name: Custom tool name (optional)
+        description: Custom tool description (optional)
+        method: HTTP method (default: "POST")
+        headers: Optional dict of headers to send with request
+        timeout_ms: Request timeout in milliseconds (default: 30000)
+        parameters: JSON schema dict for tool parameters (defaults to empty object schema)
+        execution_message: Optional message to announce before executing (spoken by agent)
+        output_normalizer: Optional function to transform HTTPResponse to string or None
```
bcherry: rather than adding additional complexity to the tool - it may be better to create two layers of abstraction:

1. a simplified utility for performing a URL request from a tool call that handles variable templating and errors consistently, and could be used inside any function_tool
2. a tool factory method that takes an HTTP config and outputs a simple tool using the aforementioned utility

this way devs could add any kind of before/after customization they need by leveraging the lower-level utility, but we'd still offer a no-frills basic wrapper

livekit-agents/livekit/agents/beta/tools/http/send_http.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+    method = method.upper()
+    timeout = aiohttp.ClientTimeout(total=timeout_ms / 1000.0)
+
+    async with aiohttp.ClientSession(timeout=timeout) as session:
+        try:
+            if method == "GET":
+                resp_ctx = session.get(url, params=arguments, headers=headers)
+            elif method == "POST":
+                resp_ctx = session.post(url, json=arguments, headers=headers)
```
bcherry: you need to `context.disallow_interruptions()` for anything besides a GET request, since these aren't otherwise cancelable

Hormold:
- added url templating with :param placeholders, auto-validating that every path param is declared and stripping those fields from the body. if the param is part of the URL template it's substituted into the path, otherwise it stays in the payload.
- built a quick schema builder + arg sanitizer so only whitelisted, type-checked params go over the wire; bad inputs from LLM raise ToolError.
- switched to the shared utils http session and block interruptions for mutating requests
- split the code into a reusable run_http_request() helper plus the existing factory, so custom tools can hook before/after without rewriting http logic.
- http tool name is required, no default description

so now you can either pass a full json_schema dict for complete control, or a shorthand params list (HTTPToolParam) for common string/enum/int/bool cases - the factory turns both into the same schema before validation.

PR #3945, opened 2025-11-15
don't use decorators for setup_fnc & load_fnc

theomonnom:


livekit-agents/livekit/agents/worker.py
```diff
@@ -430,42 +435,7 @@ def decorator(
 
         if func is not None:
             return decorator(func)
-        return decorator
-
-    @overload
-    def setup(
```
longcw, line 436: any reason to not use decorators for setup_fnc? I know a lot of examples are using 
```python
server = AgentServer()


@server.setup()
def prewarm(proc: JobProcess):

```

PR #3948, opened 2025-11-15
heygen liveavatar plugin

tinalenguyen:
[heygen liveavatar docs](https://docs.liveavatar.com/docs/getting-started)

### DIFF
--- a/livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/avatar.py
+++ b/livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/avatar.py
@@ -6,6 +6,7 @@ import contextlib
 import os
 import uuid
 from collections.abc import Iterator
+from typing import Any
 
 import aiohttp
 
@@ -20,6 +21,7 @@ from livekit.agents import (
     get_job_context,
     utils,
 )
+from livekit.agents.voice.avatar import QueueAudioOutput
 from livekit.agents.voice.room_io import ATTRIBUTE_PUBLISH_ON_BEHALF
 
 from .api import LiveAvatarAPI, LiveAvatarException
@@ -48,10 +50,13 @@ class AvatarSession:
 
         self._avatar_participant_identity = avatar_participant_identity or _AVATAR_AGENT_IDENTITY
         self._avatar_participant_name = avatar_participant_name or _AVATAR_AGENT_NAME
-        self._main_atask = asyncio.Task | None
+        self._tasks: set[asyncio.Task[Any]] = set()
+        self._main_atask: asyncio.Task | None
         self._audio_resampler: rtc.AudioResampler | None = None
         self._session_data = None
         self._msg_ch = utils.aio.Chan[dict]()
+        self._audio_playing = False
+        self._playback_position = 0.0
 
     async def start(
         self,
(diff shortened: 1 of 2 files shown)


theomonnom (approved):

theomonnom:
🔥

livekit-plugins/livekit-plugins-heygen/livekit/plugins/heygen/avatar.py
```diff
@@ -0,0 +1,202 @@
...
+        )
+
+        def ws_send(msg: dict) -> None:
+            asyncio.create_task(ws_conn.send_json(data=msg))
+
+        async def _send_task() -> None:
+            @self._agent_session.on("agent_state_changed")
+            def on_agent_state_changed(ev):
+                if ev.old_state == "speaking" and ev.new_state == "listening":
+                    ws_send({"type": "agent.speak_end", "event_id": str(uuid.uuid4())})
```
longcw: add the message to the msg_queue (`aio.Chan`) and send the message in `async def _send_task()` in a for loop. otherwise the order of the message sent by `asyncio.create_task(ws_conn.send_json(data=msg))` is not guaranteed.

btw, every created task should be tracked, to make sure it's done or cancelled at the end.

livekit-plugins/livekit-plugins-heygen/livekit/plugins/heygen/avatar.py
```diff
@@ -0,0 +1,202 @@
...
+                    encoded_audio = base64.b64encode(data).decode("utf-8")
+
+                    msg = {
+                        "type": "agent.speak",
+                        "event_id": str(uuid.uuid4()),
+                        "audio": encoded_audio,
+                    }
+
+                    try:
+                        await ws_conn.send_json(data=msg)
```
longcw: audio should be added to the same `msg_queue`, similar to the OAI realtime API implementation.

livekit-plugins/livekit-plugins-heygen/livekit/plugins/heygen/avatar.py
```diff
@@ -0,0 +1,202 @@
...
+        local_participant = self._room.local_participant
+        track_perms = rtc.ParticipantTrackPermission(
+            participant_identity=_AVATAR_AGENT_IDENTITY, allow_all=True
+        )
+        local_participant.set_track_subscription_permissions(
+            allow_all_participants=False, participant_permissions=[track_perms]
+        )
+
+        if self._agent_audio_track is not None:
+            agent_audio_stream = rtc.AudioStream.from_track(track=self._agent_audio_track)
```
longcw: I'm curious will the heygen avatar publish the audio track to the room? if so are there two audio tracks?

livekit-plugins/livekit-plugins-heygen/livekit/plugins/heygen/avatar.py
```diff
@@ -0,0 +1,202 @@
...
+
+        ws_conn = await self._api._ensure_http_session().ws_connect(url=self._ws_url)
+
+        # TODO check input rate dynamically?
+        self._audio_resampler = rtc.AudioResampler(
+            input_rate=48000, output_rate=24000, num_channels=1
```
longcw: yeah I think we need decide the input rate dynamically.

examples/avatar_agents/heygen/README.md
```diff
@@ -0,0 +1,28 @@
...
+This example demonstrates how to create a animated avatar using [Heygen LiveAvatar](https://www.liveavatar.com/).
+
+## Usage
+
+* Update the environment:
+
+```bash
+# HeyGen Config
+export HEYGEN_API_KEY="..."
+export HEYGEN_AVATAR_ID="..."
```
dyi1: Small nudge if we could rename this (and below) to LIVEAVATAR_API_KEY? A HeyGen API key is something else entirely and is completely incompatible with LiveAvatar. It could lead to confusion with users. 

A quick find and replace for a lot of "HeyGen" to "LiveAvatar" would probably save both our developers some headache.

dyi1 (approved):

livekit-plugins/livekit-plugins-heygen/livekit/plugins/heygen/avatar.py
dyi1: yes iirc, on HeyGen side, we also publish an audio track to the room.

longcw: if that's the case we can use `QueueAudioOutput` to passthrough agent audio to the avatar plugin so it's not published to the room from the agent side. here is an [example](https://github.com/livekit/agents/blob/livekit-agents@1.3.1/livekit-plugins/livekit-plugins-bithuman/livekit/plugins/bithuman/avatar.py#L237) cc @tinalenguyen

tinalenguyen: yes, there are two audio tracks. i set the permissions so that the local track cannot be heard by other participants and only the avatar track will play ([code](https://github.com/livekit/agents/blob/48dadee13f076a0006e55f2f74ae6f0ad40f0063/livekit-plugins/livekit-plugins-heygen/livekit/plugins/heygen/avatar.py#L122-L128)). i will check out `QueueAudioOutput`, thanks @longcw!

PR #3951, opened 2025-11-16
include speech duration in VAD EOS

jayeshp19:


livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
```diff
@@ -493,8 +493,8 @@ def _copy_speech_buffer() -> rtc.AudioFrame:
                         pub_speaking
                         and silence_threshold_duration >= self._opts.min_silence_duration
                     ):
+                        speech_duration = pub_speech_duration
```
davidzhao: good catch, but we don't need to create a new var here. so the change can just be moving `pub_speech_duration = 0.0` to after the event.

jayeshp19: that's much cleaner, updated.

jayeshp19: It seems we get inflated speech duration, it includes the trailing silence window (e.g., 1 s speech + min_silence_duration 0.55 s -> END_OF_SPEECH reports ~1.55 s). should we subtract the accumulated silence?

PR #3954, opened 2025-11-16
STT: ElevenLabs STTv2 (Scribe v2 Realtime) support

varghesepaul:
## Summary

  Adds support for ElevenLabs Scribe v2 Realtime streaming STT with ~150ms latency.

  ## Features

  - WebSocket-based streaming transcription
  - Configurable commit strategies (VAD/manual)
  - Word-level timestamp support
  - Automatic reconnection handling
  - Comprehensive error handling

  ## API Options

  - `model_id`: Model selection (default: scribe_v2_realtime)
  - `language_code`: Language support (optional)
  - `commit_strategy`: "vad" (default) or "manual"
  - `include_timestamps`: Enable word-level timestamps
  - VAD parameters: threshold, silence duration, speech duration

  ## Implementation Details

  - Follows Deepgram STTv2 pattern for consistency
  - Uses RecognizeStream base class (modern API)
  - Proper usage tracking via RECOGNITION_USAGE events
  - Session override support via update_options()

  ## Known Issues

  ElevenLabs API currently returns duplicate transcripts in some scenarios. I've reported this to ElevenLabs
  (https://github.com/elevenlabs/elevenlabs-python/issues/686). No explicit deduplication logic added as it risks removing valid repeated content.

  ## Documentation

 STT - Realtime : https://elevenlabs.io/docs/cookbooks/speech-to-text/streaming , https://elevenlabs.io/docs/api-reference/speech-to-text/v-1-speech-to-text-realtime

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt_v2.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt_v2.py
@@ -22,6 +22,7 @@ import logging
 import os
 import weakref
 from dataclasses import dataclass
+from typing import Literal
 from urllib.parse import urlencode
 
 import aiohttp


varghesepaul:
ElevenLabs has fixed the issue https://github.com/elevenlabs/elevenlabs-python/issues/686, and the latest test results look good.

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt_v2.py
```diff
@@ -0,0 +1,753 @@
...
+
+
+class STTv2(stt.STT):
+    def __init__(
+        self,
+        *,
+        model_id: STTModels | str = "scribe_v2_realtime",
+        language_code: NotGivenOr[str] = NOT_GIVEN,
+        audio_format: STTAudioFormat = "pcm_16000",
+        commit_strategy: str = "vad",
```
longcw:
```suggestion
        commit_strategy: Literal["vad", "manual"] = "vad",
```

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt_v2.py
```diff
@@ -0,0 +1,753 @@
...
+        include_timestamps: bool = False,
+        vad_silence_threshold_secs: float = 1.5,
```
longcw, line 73: what is the default value from 11labs, can we use `NOT_GIVEN` as default? also, during my testing, the transcripts didn't get committed after 1.5s, does that related to the configuration?

varghesepaul: Yeah, this ended up being a config issue. 

-  ElevenLabs’ default commit_strategy="manual" basically turns off all the VAD settings. So when we tested with the API defaults, vad_silence_threshold_secs wasn’t doing anything — which is why transcripts weren’t auto-committing.

- The PR fixes that by switching the default to commit_strategy="vad", which brings back the 1.5s auto-commit behavior.

- We tried a few settings in prod to make things snappier, especially for quick phrases like “Hello?”: vad_silence_threshold_secs = 0.6 (much faster than the 1.5s default)
min_silence_duration_ms = 150 (more stable than 100ms)

The old 1.5s threshold was causing single-word phrases to stall for up to 22 seconds because background noise blocked clean silence detection. Dropping it to 0.6s fixed the lag without hurting accuracy.

Also — should we keep the default at 0.6 in the PR? We can either match ElevenLabs’ default, use NOT_GIVEN, or just go with 0.6 since that aligns with the vad strategy and seems to work best in practice. I’m leaning toward setting it to 0.6 by default.

Our stt config:

  stt_instance = elevenlabs.STTv2(
      model_id="scribe_v2_realtime",
      vad_silence_threshold_secs=.6,
      vad_threshold=.4,
      min_silence_duration_ms=150,
)

Screenshot 2025-11-18 at 9 45 41 AM

Issue #3918, opened 2025-11-13
Allow stt suppression by vad


preciselyV:
### Feature Type

I cannot use LiveKit without it

### Problem
Some less common STT models (especially non-english ones) are prune to misfire - they may falsely detect words while user is silent or recognize words from different language (for multi language models). It gets even worse in more challenging audio conditions like bad user microphone or noisy background.

### Solution
The most obvious solution is to use VAD to detect the start of user utterance and delegate the recognition itself to STT, since VAD models are trained to be resistant to audio disturbance, while with STT's it's not always the case.

But, in current implementation of `AudioRecognition` any user input is sent to both STT and VAD simultaneously:

```Python 
class AudioRecognition:
...
   def push_audio(self, frame: rtc.AudioFrame) -> None:
        self._sample_rate = frame.sample_rate
        if self._stt_ch is not None:
            self._stt_ch.send_nowait(frame)

        if self._vad_ch is not None:
            self._vad_ch.send_nowait(frame)
```
And there is also no direct connection between `vad.VADEventType.END_OF_SPEECH` and `stt.SpeechEventType.FINAL_TRANSCRIPT` events, since `AudioRecognition._on_stt_event()` and `AudioRecognition._on_vad_event` know nothing about each other. 

The simplest way to solve this in my opinion is to introduce something like `suppress_stt_with_vad()` parameter to `livekit.agents.AgentSession` and pass it all the way up to the `AudioRecognition`, which in turn would do smth like:

```Python
class AudioRecognition:
    def __init__(self,...suppress_stt_with_vad:bool = False):
        self._start_of_speech_received = False
        self.suppress_stt_with_vad = suppress_stt_with_vad

    async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
        if not self._start_of_speech_received and self._suppress_stt_with_vad:
            logger.warning('STT is misfiring: received final transcript'
            ' without vads start of speech')
            self.clear_user_turn()
            return
        elif (ev.type == stt.SpeechEventType.FINAL_TRANSCRIPT and
              self._start_of_speech_received and and self._suppress_stt_with_vad):
            self._start_of_speech_received = False
        ...
    async def _on_vad_event(self, ev: vad.VADEvent) -> None:
        if ev.type == vad.VADEventType.START_OF_SPEECH:
            self._start_of_speech_received = True
            await super()._on_vad_event(ev)

...

```

### Workarounds / Alternatives

right now, since 
[...]

PR #3961, opened 2025-11-17
Allow stt suppression by vad

preciselyV:
As discussed in #3918 some STTs may misfire creating false speech recognition. To fix this @chenghao-mou suggested rewriting `StreamAdapter` to be able to work with stream capable STT's and only send STT's events if `VADEvent.START_OF_SPEECH` was generated.

Implementation is a bit different from the suggestion:

1. Input chunks are always sent via `push_frame()` to both STT and VAD streams  instead of starting to send to STT after according VAD events. Decided to go this route, since some STTs may cache previously received frames to improve prediction, and sending after VAD event is guaranteed "eat" at least one chunk with user speech. Don't wanna spoil accuracy at all 
2. Instead of calling `stt._recognize_impl()` to get `NotImplementedError` for stream only STTs, we check for `stt.capabilities` and  new parameter called `force_stream` in `StreamAdapter` . Calling the `stt._recognize_impl()` if it is implemented, will introduce unnecessary API call, and since it will be done during initialization will affect performance. It's also better to let user decide which mode he'd like to use 

By default `force_stream=False` in both `StreamAdapter` and `StreamAdapterWrapper` for backward compatibility. Unless set to `True` it will be using old logic, so no harm will be done to anyone who was already counting on it.

### DIFF
--- a/livekit-agents/livekit/agents/stt/stream_adapter.py
+++ b/livekit-agents/livekit/agents/stt/stream_adapter.py
@@ -25,9 +25,7 @@ class StreamAdapter(STT):
                 interim_results=False,
                 diarization=False,  # diarization requires streaming STT
             )
-        super().__init__(
-            capabilities=capabilities
-        )
+        super().__init__(capabilities=capabilities)
         self._vad = vad
         self._stt = stt
         self._force_stream = force_stream
@@ -74,7 +72,7 @@ class StreamAdapter(STT):
             wrapped_stt=self._stt,
             language=language,
             conn_options=conn_options,
-            force_stream=self._force_stream
+            force_stream=self._force_stream,
         )
 
     def _on_metrics_collected(self, *args: Any, **kwargs: Any) -> None:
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/stt/stream_adapter.py
```diff
@@ -83,17 +91,90 @@ def __init__(
...
+                    continue
+
+                # we let vad handle these events
+                if (
+                    event.type == SpeechEventType.START_OF_SPEECH
+                    or event.type == SpeechEventType.END_OF_SPEECH
+                ):
+                    continue
+
+                if event.type == SpeechEventType.FINAL_TRANSCRIPT and status:
```
chenghao-mou, line 160: You can receive `FINAL_TRANSCRIPT`  while the user is still speaking, especially when there is a short pause:

```txt
[ User speech 1]...........[pause].....[User speech 2]
..........................................[F 1]...............[F2] 
```

The first `FINAL_TRANSCRIPT` will turn off `start_of_speech_received` after VAD has a `START_OF_SPEECH` for speech 2. The second one will be ignored. 

Compared to VAD events, STT events can also be delayed and reliability varies across vendors.

PR #3962, opened 2025-11-17
revert #2787: execute tools without waiting text generation

longcw:
messages in RunResult are ordered by the `created_at` field

### DIFF
--- a/livekit-agents/livekit/agents/voice/run_result.py
+++ b/livekit-agents/livekit/agents/voice/run_result.py
@@ -6,10 +6,9 @@ import contextvars
 import functools
 import json
 import os
-import time
 from collections.abc import Generator
 from contextlib import contextmanager
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/worker.py
```diff
@@ -346,7 +346,7 @@ def setup_fnc(self) -> Callable[[JobProcess], Any] | None:
         return self._setup_fnc
 
     @setup_fnc.setter
-    def setup_fnc(self, value: Callable[[JobProcess], Any] | None):
+    def setup_fnc(self, value: Callable[[JobProcess], Any] | None) -> None:
```
chenghao-mou, line 349: I merged the same change in my last PR.

livekit-agents/livekit/agents/voice/run_result.py
```diff
@@ -42,25 +43,38 @@
...
+
+    def __post_init__(self) -> None:
+        self.created_at = self.item.created_at
 
 
 @dataclass
 class FunctionCallOutputEvent:
     item: llm.FunctionCallOutput
     type: Literal["function_call_output"] = "function_call_output"
+    created_at: float = field(default_factory=time.time)
```
theomonnom: Can we use item.created_at instead?
It should be possible if we add `item` to AgentHandoffEvent

theomonnom (commented):
The original intent was because `RunResult` was going to support streamed events, which is not possible if we rely on sorting the items after insertions

```
async for event in result.stream_events():
    if event.type == "message":
        ...
```

PR #3966, opened 2025-11-17
Fix GPT-5.1 reasoning_effort

ivanpuhachov:
With latest version, default `gpt-5.1` raises the following error:
```
Unsupported value: 'reasoning_effort' does not support null with this model. Supported values are: 'none', 'low', 'medium', and 'high'.\", 'type': 'invalid_request_error', 'param': 'reasoning_effort', 'code': 'unsupported_value'
```


theomonnom:
Thanks, lgtm

Issue #3969, opened 2025-11-17
`get_job_context` not usable inside function tools when running in `console` mode


mateuszkulpa:
### Bug Description

When running an agent in `console` mode, calling `get_job_context()` inside a function tool causes the worker to fail. The same code works in other modes.

### Expected Behavior

`get_job_context()` should return the job context inside function tools in `console` mode, just like in other execution modes.

### Reproduction Steps

1.  Add the following line to `examples/voice_agents/basic_agent.py`:
```
job_context = get_job_context()
```

2.  Start the worker:
```
uv run examples/voice_agents/basic_agent.py console --text
```

3.  Ask a question that triggers a function call, for example:
```
What is the weather in Paris?
```

4.  The worker crashes when the function tool attempts to call `get_job_context()`.

A full minimal reproduction is available here:
https://github.com/mateuszkulpa/agents/blob/get_job_context_reproduction/examples/voice_agents/basic_agent.py#L62-L66

### Operating System

Mac OS 15.6.1

### Package Versions

```bash
1.3.2
```

### Proposed Solution

```python

```

tinalenguyen:
Hi, I believe this is the expected behavior. For `console` mode, a fake job is simulated for the entrypoint function. Since `console` mode is meant to run locally, your code will not connect to Livekit Cloud.

[source code](https://github.com/livekit/agents/blob/b0540a08c8a8227445383ca37a7774f931b8211f/livekit-agents/livekit/agents/cli/cli.py#L1134-L1139)

longcw:
it works in audio mode but not text mode, will investigate

davidzhao:
I think it'd still be good if we could construct a job_context. that'll avoid the API behavioral differences between console and dev

longcw:
created a fix https://github.com/livekit/agents/pull/3972

PR #3972, opened 2025-11-18
fix contextvar when using text mode in console

longcw:
fix https://github.com/livekit/agents/issues/3969

When using `asyncio.run_coroutine_threadsafe()`, the coroutine has the context from the thread it's created, so `_generate(text)` created in main thread doesn't have the `_JobContextVar`. This fix uses  `c.io_loop.call_soon_threadsafe(..., context=c.io_context)` instead.


chenghao-mou (approved):
LGTM

Issue #3319, opened 2025-09-02
Multiple traces created for a single LiveKit call instead of one unified trace


gulamsarvar-bluebash:
Hi team,

I’m integrating Langfuse with LiveKit, and I’m running into an issue with trace management.

Currently, for a single call/session in LiveKit, Langfuse is automatically creating multiple traces instead of keeping everything under one unified trace. My goal is to have one trace per call/session, with child spans inside that trace for different steps (LLM calls, tool usage, events, etc.).

Even though I’m explicitly setting up a root span and trace in my custom method

<img width="555" height="175" alt="Image" src="https://github.com/user-attachments/assets/8fee3b68-5158-46ac-9d44-323b427fcd22" />

Langfuse still creates separate traces for events happening during the same call.

Questions:
How can I prevent Langfuse from creating multiple traces automatically in LiveKit integration?
Is there a way to disable auto-trace creation and use only my custom trace management logic?

longcw:
I am also not sure why this would happen, a workaround is adding a session id `langfuse.session.id` to the spans so they will be grouped into a single session in langfuse https://github.com/livekit/agents/blob/livekit-agents@1.2.8/examples/voice_agents/langfuse_trace.py#L112.

badbye:
Encountered the same issue (occurs unstably):
<img width="1045" height="635" alt="Image" src="https://github.com/user-attachments/assets/d1136b34-5563-4227-bcdd-b23777f9a2e4" />

longcw:
probably fixed by https://github.com/livekit/agents/pull/3454

marctorsoc:
Seeing this again since the update to 3.1 from 2.1.18. All traces have the same session_id. But for some reason, a new trace is created for every interaction (LLM response, TTS, etc). This is my integration code based on the example:
```
def get_trace_provider() -> TracerProvider:
    public_key = os.getenv("LANGFUSE_PUBLIC_KEY")
    secret_key = os.getenv("LANGFUSE_SECRET_KEY")
    host = os.getenv("LANGFUSE_HOST")

    if not public_key or not secret_key or not host:
        raise ValueError("LANGFUSE_PUBLIC_KEY, LANGFUSE_SECRET_KEY, and LANGFUSE_HOST must be set")

    langfuse_auth = base64.b64encode(f"{public_key}:{secret_key}".encode()).decode()
[...]

PR #3987, opened 2025-11-18
Temp workaround for langfuse otel traces

chenghao-mou:
Screenshot 2025-11-18 at 15 28 13

There seems to be a regression where Langfuse stops having a root span to group span. Previously, there would be an empty span at the root level:
Screenshot 2025-11-18 at 15 29 47

This PR adds one root span manually, so it still works:
Screenshot 2025-11-18 at 15 30 12

I suspect some part of the new code affected the Otel context, leading to dangling spans.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -637,8 +637,7 @@ class AgentActivity(RecognitionHooks):
                 attributes={trace_types.ATTR_AGENT_LABEL: self.agent.label},
             )
             try:
-                with trace.use_span(span, end_on_exit=False):
-                    await self._start_session()
+                await self._start_session()
             finally:
                 span.end()
 
@@ -659,9 +658,8 @@ class AgentActivity(RecognitionHooks):
                 attributes={trace_types.ATTR_AGENT_LABEL: self._agent.label},
             )
             try:
-                with trace.use_span(span, end_on_exit=False):
-                    await self._pause_scheduling_task(blocked_tasks=blocked_tasks)
-                    await self._close_session()
+                await self._pause_scheduling_task(blocked_tasks=blocked_tasks)
+                await self._close_session()
             finally:
                 span.end()
 


chenghao-mou:
Related to https://github.com/livekit/agents/issues/3319

chenghao-mou:
Q: Should we create the tracer provider at the server level and create a separate one for each `rtc_session` call?

longcw:
I think the original issue was caused by https://github.com/livekit/agents/pull/3726 where the `agent_session` span is not used as root span anymore. do you know why we made that change?

chenghao-mou:
> I think the original issue was caused by #3726 where the `agent_session` span is not used as root span anymore. do you know why we made that change?

Can confirm. As soon as I add that decorator back, it works as before. I don't know why we made that change, but I will try to figure out a fix without reverting it.

chenghao-mou:
Found the solution. The `agent_session` span was never activated/attached. There are two more (`_agent_speaking_span` and `_user_speaking_span`) that are not activated either. But I think activating them might cause nested span issues with barge-in/interruptions. I will look into them when working on barge-in instead.

longcw:
@chenghao-mou 
> There are two more (_agent_speaking_span and _user_speaking_span) that are not activated either.

yes, this is intended. there is no more spans should be nested under the agent/user speaking span.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -637,7 +637,8 @@ async def resume(self) -> None:
                 attributes={trace_types.ATTR_AGENT_LABEL: self.agent.label},
             )
             try:
-                await self._start_session()
+                with trace.use_span(span, end_on_exit=False):
```
chenghao-mou: Do we need this?

longcw: I think no

theomonnom (approved):

marctorsoc:
@chenghao-mou @longcw this fixed part of the issue, so now the traces are grouped. But I see the `input` empty. I only see the output and metadata. Is this something you're aware / working on?

chenghao-mou:
> But I see the input empty. I only see the output and metadata. Is this something you're aware / working on?

Could you share a trace/screenshot? I don't think we have received similar issue reports.

marctorsoc:
here @chenghao-mou 

image

and

image

marctorsoc:
image

chenghao-mou:
@marctorsoc I can confirm this is a regression on our side. Will look into it. Thanks for reporting!

PR #3995, opened 2025-11-18
AGT-2208 Allow pause in final transcript

chenghao-mou:
image

1. User speaks with an VAD span (hey can you tell me the part… how livekit, how does it work? [brief pause])
2. Receives one FINAL_TRANSCRIPT (preemptive generation triggered, new speech, not interrupted)
3. Agent could start speaking here
4. Receives another FINAL_TRANSCRIPT (no interruption occurred, no preemptive generation due to `or (self._current_speech is not None and not self._current_speech.interrupted)` ***VAD also failed or delayed in this case***  (`And` | `But also the,`)
5. Or agent could start speaking here
6. Since the second transcript is not an EOU, the agent continues speaking while the user is also continuing his speech, until VAD or EOU triggers interruption eventually

This makes sure FINAL TRANSCRIPT will at least pause the speech (and immediately interrupt it) if needed.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1257,14 +1257,9 @@ class AgentActivity(RecognitionHooks):
             ),
         )
 
-        if (
-            ev.alternatives[0].text
-            and self._audio_recognition
-            and self._audio_recognition._turn_detection_mode
-            not in (
-                "manual",
-                "realtime_llm",
-            )
+        if ev.alternatives[0].text and self._turn_detection not in (
+            "manual",
+            "realtime_llm",
         ):
             self._interrupt_by_audio_activity()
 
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -334,6 +333,12 @@ async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
             self._audio_transcript = self._audio_transcript.lstrip()
             self._final_transcript_confidence.append(confidence)
             transcript_changed = self._audio_transcript != self._audio_preflight_transcript
+            self._hooks.on_final_transcript(
+                ev,
+                speaking=self._speaking if self._vad else None,
+                pause_speech=transcript_changed
+                and (self._vad_base_turn_detection or self._user_turn_committed),
```
longcw: it seems I forgot to check the `turn_detection_mode` when interrupting by the interim transcript.

I think we don't need `pause_speech` here, but check the turn_detection_mode in both `on_final_transcript` and `on_interim_transcript` before interruption, like `_turn_detection_mode not in ("manual", "realtime")`?

chenghao-mou: Added the check in both functions and dropped the `pause_speech` flag.

theomonnom (approved):
Nice catch!

longcw (approved):

PR #4023, opened 2025-11-19
pass raw schema for out_of_scope in TaskGroup

tinalenguyen:


theomonnom:
fixed inside https://github.com/livekit/agents/pull/4025

PR #4027, opened 2025-11-20
feat(gemini3) : Add Gemini 3 support with thinking_level and thought_signature

varghesepaul:
Adds support for Gemini 3 models with model-specific parameter handling and multi-turn function calling via thought_signature.

  Changes:
  - Add gemini-3-pro-preview model
  - Support thinking_level parameter for Gemini 3 models
  - Maintain backward compatibility with thinking_budget for Gemini 2.x
  - Implement thought_signature capture and injection for multi-turn function calls
  - Add model version detection to route parameters correctly
  - Optimize thought_signature handling to only process for Gemini 3 models
  - Inject thought_signature inline during format conversion to avoid post-processing

### DIFF
--- a/livekit-agents/livekit/agents/llm/_provider_format/google.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/google.py
@@ -17,7 +17,10 @@ class GoogleFormatData:
 
 
 def to_chat_ctx(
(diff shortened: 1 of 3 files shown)


superMDguy:
Would be awesome to try this out with the new Flash model!

Hormold:
Thanks for PR! I'll try to test and merge it soon!

PR #4029, opened 2025-11-20
fix bithuman avatar getting local participant identity

longcw:


davidzhao (approved):
lg.

does this impact other plugins as well?

davidzhao:
hmm I see a few plugins with the same check.. should we simplify and make it `job_ctx.local_participant_identity`?

longcw:
> does this impact other plugins as well?

other avatar plugins already used this to allow avatar session starts before room connected.

davidzhao:
I understand.. but this is a lot of duplication.. wouldn't it be cleaner to have what I proposed instead of copying the same logic everywhere?

longcw:
> should we simplify and make it job_ctx.local_participant_identity

yeah that makes sense.

PR #4031, opened 2025-11-20
add JobContext.local_participant_identity

longcw:


livekit-agents/livekit/agents/job.py
```diff
@@ -310,6 +310,13 @@ def room(self) -> rtc.Room:
     def agent(self) -> rtc.LocalParticipant:
         return self._room.local_participant
 
+    @property
+    def local_participant_identity(self) -> str:
+        if identity := self.token_claims().identity:
+            return identity
+
+        return self._room.local_participant.identity
```
theomonnom, line 318: can this be different?

longcw: it should be the same, just in case the first one is empty? maybe not possible?

davidzhao: I can't see how they would be different. this has to be the same token we are passing in when connecting to room.

any chance `token_claims` could throw?

longcw: I originally thought `Claims.identity` could be empty since the default value is `""`. it seems it cannot be empty as it's already checked in `Claims.to_jwt`

PR #4032, opened 2025-11-20
AGT-2182: Add inference bargein and examples

chenghao-mou:
- New barge-in detector under inference
- Two stream implementation:
	- HTTP endpoints for working with hosted model
	- WS for working with gateway proxy

Detailed spec can be found in the [notion page](https://www.notion.so/livekit/Barge-in-Detection-Workflow-and-API-Design-2a03c4901a42805c835bda5605495e6e?source=copy_link)

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -511,7 +511,10 @@ class SpeechStream(stt.SpeechStream):
                         ws = None
                     await utils.aio.gracefully_cancel(*tasks, wait_reconnect_task)
                     tasks_group.cancel()
-                    tasks_group.exception()  # retrieve the exception
+                    try:
+                        tasks_group.exception()  # retrieve the exception
+                    except asyncio.CancelledError:
+                        pass
             finally:
                 closing_ws = True
                 if ws is not None and not ws.closed:
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -163,8 +163,8 @@ class AgentActivity(RecognitionHooks):
         )
         self._turn_detection = self._validate_turn_detection(turn_detection)
 
-        # barge-in detection
-        self._bargein_detection_enabled: bool = self.bargein_detection is not None
+        self._bargein_detector: inference.BargeinDetector | None = self._resolve_bargein_detection()
+        self._bargein_detection_enabled: bool = self._bargein_detector is not None
 
         # this allows taking over audio interruption temporarily until barge-in is detected
         # by default it is true unless turn_detection is manual or realtime_llm
@@ -607,7 +607,7 @@ class AgentActivity(RecognitionHooks):
             hooks=self,
             stt=self._agent.stt_node if self.stt else None,
             vad=self.vad,
-            bargein_detection=self.bargein_detection,
+            bargein_detection=self._bargein_detector,
             min_endpointing_delay=self.min_endpointing_delay,
             max_endpointing_delay=self.max_endpointing_delay,
             turn_detection=self._turn_detection,
@@ -2804,8 +2804,7 @@ class AgentActivity(RecognitionHooks):
     def vad(self) -> vad.VAD | None:
         return self._agent.vad if is_given(self._agent.vad) else self._session.vad
 
-    @property
-    def bargein_detection(self) -> inference.BargeinDetector | None:
+    def _resolve_bargein_detection(self) -> inference.BargeinDetector | None:
         if not (
             self.stt is not None
             and self.stt.capabilities.aligned_transcript


examples/bargein_server.py
theomonnom, line 1: Let's move this file to `other`

examples/voice_agents/basic_agent.py
```diff
@@ -14,6 +14,7 @@
     metrics,
     room_io,
 )
+from livekit.agents.inference.bargein import BargeinDetector
```
theomonnom: Can it just be:
```suggestion
from livekit.agents.inference import BargeinDetector
```

examples/voice_agents/basic_agent.py
```diff
@@ -93,6 +94,7 @@ async def entrypoint(ctx: JobContext):
         # See more at https://docs.livekit.io/agents/build/turns
         turn_detection=MultilingualModel(),
         vad=ctx.proc.userdata["vad"],
+        bargein_detector=BargeinDetector(),
```
theomonnom: I like this pattern (it follows how we use STT/TTS in the inference gateway in our docs) 
```suggestion
        bargein_detector=inference.BargeinDetector(),
```

Tho I'm also wondering if it should always be there by default.

livekit-agents/livekit/agents/bargein.py
```diff
@@ -0,0 +1,277 @@
...
+    timestamp: float
+    label: str
+    error: Exception = Field(..., exclude=True)
+    recoverable: bool
+
+
+TEvent = TypeVar("TEvent")
+
+
+class BargeinDetector(
```
theomonnom: It is very unlikely we're going to have third-party/different bargin-in models. (At least they may not work the same way at all).

I think we should remove this abstract baseclass and directly have the implementation as-is in the inference/bargin.py

This way, we have complete freedom to modify the API at any time without worrying about this class being backward compatible.
We just do whatever we need inside our inference/bargin.py implementation

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -343,6 +479,41 @@ async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
...
+                        else None,
+                    },
+                )
+                self._transcript_buffer.append(ev)
+                return
+            elif self._transcript_buffer:
+                # emit the preceding sentinel event immediately before this event
+                # assuming *only one* sentinel event could precede the current event
+                # ignore if the previous event is not a sentinel event
+                logger.debug(
```
theomonnom: I see that we're only re-emitting events inside `_transcript_buffer` when we receive a new STT event. Shouldn't we also re-emit it if we detect it was a bargin? The STT may not send us any additional events, so there is a case where we just ignore this buffered event.

examples/bargein_server.py
chenghao-mou: I will remove it before merging. I don't think the user will need it.

livekit-agents/livekit/agents/voice/audio_recognition.py
chenghao-mou: Yes, I don't have a return in this `elif` branch. So that after-the-barge-in event will be processed normally.

livekit-agents/livekit/agents/bargein.py
chenghao-mou: Sounds good. I will drop the base classes.

livekit-agents/livekit/agents/voice/audio_recognition.py
theomonnom: What I mean is what if this scenario happens: 
1. should_hold_event is True
2. we buffer the transcript
3. we detected it was indeed a barge-in
4. we never receive new transcripts from the STT

In this case it seems like we just lose the buffered transcript and we don't generate any new generation?

chenghao-mou: Great question! Here is a breakdown of what I am thinking when there is **no new transcript**:

## Scenario 1: If there is no actual user speech/barge-in

This is a barge-in false positive. If `resume_false_interruption` is enabled, the audio will pause and then resume. If not enabled, it will interrupt with no new transcript.

## Scenario 2: If there is actual user speech/barge-in but no new transcript

Case 1: This is an STT failure. But it should behave similarly to Scenario 1.

Case 2: The transcript for the barge-in comes before the inference is done

The typical audio needed for the model for a barge-in detection is around **300ms**. This means we can lose the transcript for that 300ms window. I think it is very rare to have a true barge-in for just 300ms of speech and STT finishes transcribing that before it's detected. 

But there is a non-zero probability of barge-in detection being late, in which case, we might lose more transcript. What I can do here is to re-emit transcripts back until the last speaking point (if we have the timing information), but it might include non-barge-in transcript if the user says something like "Right, right [bc], we should ...[barge-in]".

chenghao-mou: Okay, I have updated it so that it will flush either at the end of agent speech or at new STT events up to the last overlap speech start.

examples/voice_agents/basic_agent.py
chenghao-mou: Maybe we can enable this by default after we have released it for a while. You know, just making sure it works well in various production use cases.

livekit-agents/livekit/agents/log.py
```diff
@@ -1,6 +1,22 @@
...
+    def trace(self, message: str, *args: Any, **kwargs: Any) -> None:
+        if self.isEnabledFor(TRACE_LEVEL):
+            self._log(TRACE_LEVEL, message, args, **kwargs)
+
+    def dev(self, message: str, *args: Any, **kwargs: Any) -> None:
+        if self.isEnabledFor(DEV_LEVEL):
+            self._log(DEV_LEVEL, message, args, **kwargs)
+
+
+logging.setLoggerClass(Logger)
```
theomonnom, line 21: TIL we can do that

livekit-agents/livekit/agents/inference/__init__.py
```diff
@@ -1,3 +1,10 @@
+from .bargein import (
+    BargeinDetector,
+    BargeinError,
+    BargeinEvent,
+    BargeinEventType,
+    BargeinStreamBase,
```
theomonnom:
```suggestion
```
Should this one stays internal?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -175,19 +204,181 @@ def update_options(
...
         self.update_stt(self._stt)
         self.update_vad(self._vad)
+        self.update_bargein_detector(self._bargein_detector)
 
     def stop(self) -> None:
         self.update_stt(None)
         self.update_vad(None)
+        self.update_bargein_detector(None)
+
+    def start_barge_in_monitoring(self) -> None:
```
theomonnom: I think it's totally fine to name this:
`on_agent_started_speaking`

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -175,19 +204,181 @@ def update_options(
...
+                    user_speaking_span.attributes
+                    and user_speaking_span.attributes.get(trace_types.ATTR_IS_BARGEIN) is None
+                ):
+                    user_speaking_span.set_attribute(trace_types.ATTR_IS_BARGEIN, "false")
+            else:
+                user_speaking_span.set_attribute(trace_types.ATTR_IS_BARGEIN, "false")
+
+        self._bargein_ch.send_nowait(inference.BargeinStreamBase._OverlapSpeechEndedSentinel())
+
+    def end_barge_in_monitoring(self, ignore_until: float) -> None:
```
theomonnom: `on_agent_stopped_speaking`

theomonnom: Let's make it a kwargs & rename the parameter, otherwise it isn't obvious when reading the code from other places
```suggestion
    def end_barge_in_monitoring(self, *, ignore_transcript_until: float) -> None:
```

theomonnom (approved):

livekit-agents/livekit/agents/inference/__init__.py
chenghao-mou: good point!

livekit-agents/livekit/agents/voice/audio_recognition.py
chenghao-mou: Good point. Looking at the names we have today:
Screenshot 2025-12-18 at 09 57 44

I think this set matches better:
Screenshot 2025-12-18 at 09 58 15

wdyt?

PR #4033, opened 2025-11-20
Support for pronunciation dictionary in Cartesia TTS

cateet:
Add support for pronunciation dictionary in Cartesia TTS as per https://docs.cartesia.ai/api-reference/tts/websocket


chenghao-mou:
Hey @cateet thanks for creating a PR!

It seems that this feature is only supported by `sonic-3` models. Could you add a check in `_check_generation_config` and update the last part in `update_options` to trigger the check when the parameter is passed?

cateet:
> Hey @cateet thanks for creating a PR!
> 
> It seems that this feature is only supported by `sonic-3` models. Could you add a check in `_check_generation_config` and update the last part in `update_options` to trigger the check when the parameter is passed?

Great catch. Updated the PR

Issue #3881, opened 2025-11-11
ElevenLabs Scribe v2 Realtime


IngLP:
### Feature Type

Would make my life easier

### Feature Description

ElevenLabs released Scribe v2 Realtime, which has Realtime support AND great support for many languages.
Currently, ElevenLabs STT is slow due to no Streaming support.
Adding this would make Scribe usable for calls, if you need low latency (and you want to).

### Workarounds / Alternatives

Currently using Deepgram STT, which is the fastest that also supports non-English, the only one to achieve low latency for other languages than English.

davidzhao:
hey all, we are aware of the interest in Scribe v2 and will make this available shortly

samuelcastro:
Let's go! 🚀

vishal-kickcall:
Really excited to test this with our agents

facitoo:
waiting for livekit ...

louisjoecodes:
Excited to see Scribe Realtime V2 in livekit!

amanbmujawar:
any ETA for this ?

sebhs:
+1 please

sanderontwikkelt:
🚀

HeyBreezDev:
🚀🚀

yb90200:
+1

nagarajadiga:
+1

mollaosmanoglu:
+1

lennycampino:
+1

samuelcastro:
No JS?

shashwat-aditya:
Where is the documentation for this? Can it only be used with inference route?

Lokesh8Goenka:
+1

PR #3909, opened 2025-11-12
feat(elevenlabs): add STTv2 with streaming support for Scribe v2

yorrick:
## Summary

Adds streaming speech-to-text support for ElevenLabs Scribe v2 API.

### Changes
- Added `STTv2` class in `stt_v2.py` following the Deepgram and Cartesia pattern
- Implements WebSocket-based real-time streaming transcription (not using elevenlabs-python sdk)
- Only supports VAD commit strategy
- Supports sample rates from 8kHz to 48kHz
- Added type definitions for `STTModels`, `STTAudioFormat`, and `STTCommitStrategy`

## Questions

1. **STTv2 approach**: Is creating a separate `STTv2` class (following the Deepgram pattern) the right approach for maintaining backward compatibility?

2. **Testing**: I wasn't able to update `test_stt.py` - the tests appear to be broken since May 2025 (`make_test_speech` is commented out in commit 64db393f). I noticed that Deepgram STTv2 (PR #3245) was also added without test changes. I did the test with a full pipeline

PR #4041, opened 2025-11-21
add use_realtime to elevenlabs stt and support scribe v2 realtime model

longcw:
clean up https://github.com/livekit/agents/pull/3909
close https://github.com/livekit/agents/issues/3881

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
@@ -155,6 +155,13 @@ class STT(stt.STT):
                 headers={AUTHORIZATION_HEADER: self._opts.api_key},
             ) as response:
                 response_json = await response.json()
+                if response.status != 200:
+                    raise APIStatusError(
+                        message=response_json.get("detail", "Unknown ElevenLabs error"),
+                        status_code=response.status,
+                        request_id=None,
+                        body=response_json,
+                    )
                 extracted_text = response_json.get("text")
                 language_code = response_json.get("language_code")
                 speaker_id = None


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -28,58 +32,89 @@
...
+from .log import logger
+from .models import STTRealtimeSampleRates
+
 API_BASE_URL_V1 = "https://api.elevenlabs.io/v1"
 AUTHORIZATION_HEADER = "xi-api-key"
 
 
+class VADOptions(TypedDict, total=False):
+    vad_silence_threshold_secs: float | None
+    """Silence threshold in seconds for VAD. Default to 1.5"""
```
davidzhao, line 50: this seems kind of long for realtime? does it mean that it'll mark the end of speech after 1.5s?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -170,3 +212,299 @@ def _transcription_to_speech_event(
...
+
+        for stream in self._streams:
+            stream.update_options(server_vad=server_vad)
+
+    def stream(
+        self,
+        *,
+        language: NotGivenOr[str] = NOT_GIVEN,
+        conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
+    ) -> SpeechStream:
```
davidzhao, line 236: should we throw here if `use_realtime` isn't set to true?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
longcw: 1.5s is the default value from 11labs, and yeah I think it's too long. actually the server VAD there is very sensitive to background noise, it tends to never stop during my testing...

their default commit strategy is `manual`... so I think a proper way to use their realtime model is to combine it with a local VAD https://github.com/livekit/agents/pull/4043

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
longcw: the `stt.capabilities.streaming` will be False if `use_realtime` isn't set, ideally the agent framework won't use `stt.stream` in this case

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
davidzhao: so to be clear, they do not send final transcript until VAD is clear?

longcw: yes

PR #4058, opened 2025-11-23
fix: accurate speech duration in VAD EOS

jayeshp19:
https://github.com/livekit/agents/pull/3951#discussion_r2533060419


jayeshp19:
> I think the change makes sense technically.
> 
> But it might change how people should set `min_interruption_duration`. Right now it's used when determining audio activity interruption. Users might need to exclude `min_silence_duration` from it manually.
> 
> cc @livekit/agent-devs what do you think?

We’re using the `INFERENCE_DONE` event for the `min_interruption_duration` comparison, and this change does not affect that event.
INFERENCE_DONE event could publish inflated speech duration, not sure, if we should handle that...

chenghao-mou (approved):
You are right. LGTM!

Issue #4072, opened 2025-11-24
LiveKit Agents 1.3 JobContext.api credentials regression


inickt:
### Bug Description

After updating to 1.3.3 (locally, since this also needs https://github.com/livekit/agents/pull/4046 until a new release is published), `JobContext.api` no longer works with credentials that are loaded at runtime in `WorkerOptions` (or `AgentServer`).

These environment variables used to be exported if they were set at runtime: https://github.com/livekit/agents/pull/3581

### Expected Behavior

`ctx.api` should reuse the same credentials set in `WorkerOptions`/`ServerOptions`/`AgentServer`

### Reproduction Steps

```bash
1. Set `ws_url`/`api_key`/`api_secret` in code, whether through legacy `WorkerOptions` or new `AgentServer`
2. Access `ctx.api`
```

### Operating System

macOS / linux

### Models Used

n/a

### Package Versions

```bash
1.3.3 @ cefeaa7e7e2f1f6f5308b60ae461357a6b45ceec
```

### Proposed Solution

1. Update documentation [here](https://github.com/livekit/agents/blob/cefeaa7e7e2f1f6f5308b60ae461357a6b45ceec/livekit-agents/livekit/agents/job.py#L276-L282) (still would be a regression)
2. Update `AgentServer` to export env variables (like before)
3. Update `JobContext` to actually use the LiveKit API from `AgentServer`

(3) seems ideal to me if the API session is able to be reused, otherwise (2) is fine. If you deem this a mistake it is pretty easy for us to fix (1) in code ourselves, but it is a tiny regression so wanted to flag.

### Additional Context

We have a more complicated server startup with config loading so we have a bit more indirection that normal where we want our own config loader and to explicitly pass in creds vs assuming the environment variable are always set.

Thanks!

tinalenguyen:
Hi, thanks for the detailed report! I made a PR and went with solution (2), we will likely keep the behavior of creating a new API session every time `ctx.api` is accessed

inickt:
Thanks for the quick fix!

PR #4075, opened 2025-11-24
add livekit credentials to environment

tinalenguyen:
for https://github.com/livekit/agents/issues/4072


livekit-agents/livekit/agents/worker.py
```diff
@@ -286,6 +286,11 @@ def __init__(
         self._ws_url = ws_url or os.environ.get("LIVEKIT_URL") or ""
         self._api_key = api_key or os.environ.get("LIVEKIT_API_KEY") or ""
         self._api_secret = api_secret or os.environ.get("LIVEKIT_API_SECRET") or ""
+
+        os.environ["LIVEKIT_URL"] = self._ws_url
+        os.environ["LIVEKIT_API_KEY"] = self._api_key
+        os.environ["LIVEKIT_API_SECRET"] = self._api_secret
+
```
theomonnom: I think we should do that inside the run() method instead

PR #4078, opened 2025-11-24
Support custom auth schemes for MCP servers

cecchi:
This PR adds an `auth` param to the `MCPServerHTTP` class, which is passed through to the underlying `mcp` library.

Livekit supports providing a list of static headers to the MCP client which is useful for basic authentication schemes, but are often insufficient. In our case, our authentication tokens are short-lived and need to be refreshed periodically. Having a more flexible way to configure authentication makes this possible.

```py
session = agents.AgentSession(
    mcp_servers=[
        mcp.MCPServerHTTP(
            url="https://foo.bar/mcp/",
            auth=RefreshingAccessTokenAuth(),
        )
    ],
)
```

### DIFF
--- a/livekit-agents/livekit/agents/llm/mcp.py
+++ b/livekit-agents/livekit/agents/llm/mcp.py
@@ -13,6 +13,7 @@ from urllib.parse import urlparse
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 
 try:
+    from httpx import Auth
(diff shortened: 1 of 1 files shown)


cecchi:
Hi @theomonnom - would love to get your eyes on this if possible, assuming you're the right person. Similar to https://github.com/livekit/agents/pull/4365 where the option to control the underlying transport was added, this just gives consumers slightly more control over the MCP transport mechanics.

PR #4085, opened 2025-11-25
speech_handle's _wait_for_generation hangs forever if cancelled before playout

simllll:
if a speech handle gets cancelled before the playout has started, it's "on the queue", and authorization is cleared, but then the _wait_for_generation is never finished. 

I'm not sure if this is the real fix, but at least it works..maybe someone of the python gurus has a better idea?

I had this issue already in the typescript version, but was wondering why there is nothing like this in the python version... and now I also stumbled upon this in the PY version.

see also https://github.com/livekit/agents-js/pull/842

Added a test suite to reproduce the issue.

disclaimer: new to agent's python sdk ;-)

### DIFF
--- a/livekit-agents/livekit/agents/voice/speech_handle.py
+++ b/livekit-agents/livekit/agents/voice/speech_handle.py
@@ -192,7 +192,12 @@ class SpeechHandle:
         if not self._generations:
             raise RuntimeError("cannot use wait_for_generation: no active generation is running.")
 
-        await asyncio.shield(self._generations[step_idx])
+        # Race against the interrupt future to avoid hanging when speech is interrupted
+        # before the generation completes
+        await asyncio.wait(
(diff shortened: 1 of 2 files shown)


longcw:
a related fix https://github.com/livekit/agents/pull/3169 to avoid deadlock when speech is interrupted before `_authorize_generation` called.

I am wondering do you see any other issues caused by `_wait_for_generation` from the agents?

simllll:
With this fix in place, I had no further issues in typescript. I haven't looked at the other PR in detail,but I'm pretty sure it's not applied in the TS version and therefore I was not affected by it.

This fix is quite easy, and I can't see any side effects so far.

What we could question: what is when a speech out is playing,and cancelled, .. while wait for generation is running. But as we cancel it anyway, I guess there is nothing wrong with it.

PR #4088, opened 2025-11-25
AGT-2269 insert silence during pauses for RecorderIO

chenghao-mou:
This ensures we have silence pauses in the recording by:
- tracking pause and resume wall times
- insert silences once playback is finished

Screenshot 2025-11-26 at 10 31 40
Image: two false interruptions and pauses


longcw (approved):
lgtm!

Issue #4094, opened 2025-11-25
Add full parameter support to update_options in Rime TTS plugin


gokuljs:
### Feature Type

I cannot use LiveKit without it

### Feature Description

The Rime TTS plugin currently allows updates only for model and speaker values through the update_options method. This creates limitations for applications that need to adjust TTS behavior during runtime. Voice agents often modify temperature, speed, stability, latency settings, or other Rime specific parameters based on conversation context or user preference. Since these values cannot be changed without recreating the full TTS instance.

The requested feature is to expand update_options so it supports all parameters accepted during initialization.

gokuljs:
#4095

PR #4095, opened 2025-11-25
feat(rime): expand update_options to accept all TTS parameters

gokuljs:
The Rime TTS plugin currently allows updates only for model and speaker values through the update_options method. This creates limitations for applications that need to adjust TTS behavior during runtime. Voice agents often modify temperature, speed, stability, latency settings, or other Rime specific parameters based on conversation context or user preference. Since these values cannot be changed without recreating the full TTS instance.


gokuljs:
#4094

gokuljs:
@theomonnom something seems off with the Ruff formatter in the GitHub CI. I ran both the format and check commands locally and they passed, but the CI is still failing.

davidzhao (approved):
lgtm

PR #4099, opened 2025-11-26
Fallback API for Inference

adrian-cowham:


adrian-cowham:
Yeah, that's interesting i wonder if we should find a way to report back the model used by Inference. Also, @chenghao-mou can you coordinate with me before merging and releasing?

adrian-cowham:
Also, I just pushed a small change to address 3.9 type checks, hopefully.

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -69,6 +70,59 @@ class AssemblyaiOptions(TypedDict, total=False):
...
+    Extra fields are passed through to the provider.
+
+    Example:
+        >>> FallbackModel(name="deepgram/nova-3", keywords=["livekit"])
+    """
+
+    name: STTModels | str
+    """Model name (e.g. "deepgram/nova-3", "assemblyai/universal-streaming", "cartesia/ink-whisper")."""
+
+    extra: dict[str, Any] | None = None
```
theomonnom: to be consistent
```suggestion
    extra_kwargs: dict[str, Any] | None = None
```

livekit-agents/livekit/agents/inference/tts.py
```diff
@@ -42,6 +44,87 @@
...
+    Extra fields are passed through to the provider.
+
+    Example:
+        >>> FallbackModel(name="cartesia/sonic", voice="")
+    """
+
+    name: Required[str]
+    """Model name (e.g. "cartesia/sonic", "elevenlabs/eleven_flash_v2", "rime/arcana")."""
+
+    voice: Required[str | None]
```
theomonnom: Can the voice be None? Is this different than omitting it (removing the Required flag)

livekit-agents/livekit/agents/inference/tts.py
```diff
@@ -42,6 +44,87 @@
...
+
+
+class Fallback(TypedDict, total=False):
+    """Configuration for fallback models when the primary model fails."""
+
+    models: Required[Sequence[FallbackModelType]]
+    """Fallback models in priority order."""
+
+    connection: ConnectionOptions
+    """Connection options for fallback attempts."""
```
theomonnom: I'm wondering if this class is really useful. 

Maybe the argument inside the TTS/STT can directly accept a Sequence of FallbackModelType and the ConnectionOptions can directly be inside the constructor? 

Probably nitpicking but maybe we can also re-use https://github.com/livekit/agents/blob/f7e42af4732a73d855c6a16365cdb4095d107b95/livekit-agents/livekit/agents/types.py#L54?

I'm not 100% sure, wdyt?

theomonnom (approved):
Otherwise lgtm! Nice work!

livekit-agents/livekit/agents/inference/tts.py
chenghao-mou: Right now, we already have `conn_options` as a parameter in `stream` or `_recognize_impl`, but it is often used with the default value somewhere behind the scenes (not exposed to the user). I think it is a good idea to expose this at the constructor level for all of them (we can start with inference in this PR) with the existing `APIConnectOptions`.

chenghao-mou:
> Yeah, that's interesting i wonder if we should find a way to report back the model used by Inference. Also, @chenghao-mou can you coordinate with me before merging and releasing?

That would be nice-to-have, and we can do that in a later PR. And yes, I should be able to do that.

livekit-agents/livekit/agents/inference/tts.py
```diff
@@ -98,6 +152,8 @@ def __init__(
         api_secret: NotGivenOr[str] = NOT_GIVEN,
         http_session: aiohttp.ClientSession | None = None,
         extra_kwargs: NotGivenOr[CartesiaOptions] = NOT_GIVEN,
+        fallback: NotGivenOr[list[FallbackModelType] | FallbackModelType] = NOT_GIVEN,
+        connect_options: NotGivenOr[APIConnectOptions] = NOT_GIVEN,
```
theomonnom: On the other parts of the framework, it's named `conn_options`. I think we should be consistent

livekit-agents/livekit/agents/inference/tts.py
```diff
@@ -115,6 +171,8 @@ def __init__(
         api_secret: NotGivenOr[str] = NOT_GIVEN,
         http_session: aiohttp.ClientSession | None = None,
         extra_kwargs: NotGivenOr[ElevenlabsOptions] = NOT_GIVEN,
+        fallback: NotGivenOr[list[FallbackModelType] | FallbackModelType] = NOT_GIVEN,
+        connect_options: NotGivenOr[APIConnectOptions] = NOT_GIVEN,
```
theomonnom: Nice! It's much cleaner to have the conn_options outside of the old Fallback dict

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -69,6 +70,48 @@ class AssemblyaiOptions(TypedDict, total=False):
 STTLanguages = Literal["multi", "en", "de", "es", "fr", "ja", "pt", "zh", "hi"]
 
 
+class FallbackModel(TypedDict, total=False):
```
marctorsoc, line 73: could you add some docs for this?

tests/test_inference_stt_fallback.py
```diff
@@ -0,0 +1,259 @@
...
+    def test_string_with_language_suffix_discards_language(self):
+        """Language suffix in string model is discarded."""
+        result = _normalize_fallback("deepgram/nova-3:en")
+        assert result == [{"name": "deepgram/nova-3"}]
+
+    def test_fallback_model_with_extra_kwargs(self):
+        """FallbackModel with extra_kwargs is preserved."""
+        fallback = FallbackModel(
+            name="deepgram/nova-3",
+            extra_kwargs={"keywords": [("livekit", 1.5)], "punctuate": True},
```
marctorsoc, line 137: Q: isn't deepgram moving from keywords to keyterms? https://developers.deepgram.com/docs/keyterm

PR #4102, opened 2025-11-26
AGT-2317: wait for user silence before speaking

chenghao-mou:
Speech will wait for both authorization and user silence.

User silence is defined as:
- VAD: not speaking
- VAD: speaking but has `raw_accumulated_silence`

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1630,7 +1630,7 @@ class AgentActivity(RecognitionHooks):
 
         if speech_handle.interrupted:
             current_span.set_attribute(trace_types.ATTR_SPEECH_INTERRUPTED, True)
-            await utils.aio.cancel_and_wait(wait_for_authorization)
+            await utils.aio.cancel_and_wait(wait_for_authorization, wait_for_user_silence)
             return
 
         text_source: AsyncIterable[str] | None = None
@@ -1893,7 +1893,7 @@ class AgentActivity(RecognitionHooks):
 
         if speech_handle.interrupted:
             current_span.set_attribute(trace_types.ATTR_SPEECH_INTERRUPTED, True)
-            await utils.aio.cancel_and_wait(*tasks, wait_for_authorization)
+            await utils.aio.cancel_and_wait(*tasks, wait_for_authorization, wait_for_user_silence)
             await text_tee.aclose()
             return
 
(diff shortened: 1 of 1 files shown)


theomonnom:
If the user is in a noisy environment, we may never trigger the speech? VAD is a very unstable signal

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1243,6 +1247,11 @@ def on_vad_inference_done(self, ev: vad.VADEvent) -> None:
         if ev.speech_duration >= self._session.options.min_interruption_duration:
             self._interrupt_by_audio_activity()
 
+        if ev.speaking and ev.raw_accumulated_silence <= 0:
```
longcw: we may want a threshold of `raw_accumulated_speech` here?

longcw:
> If the user is in a noisy environment, we may never trigger the speech? VAD is a very unstable signal

I think it's fine, if VAD has a lot of false positives in a noisy env, the agent speech will still be interrupted all the time even without this waiting mechanism. we can add an option to make the threshold configurable, maybe disabled by default right now.

livekit-agents/livekit/agents/voice/agent_activity.py
chenghao-mou: Right, we definitely don't want it to be speaking between user words.

chenghao-mou:
> > If the user is in a noisy environment, we may never trigger the speech? VAD is a very unstable signal
> 
> I think it's fine, if VAD has a lot of false positives in a noisy env, the agent speech will still be interrupted all the time even without this waiting mechanism. we can add an option to make the threshold configurable, maybe disabled by default right now.

How about `min_response_silence`: how long the user silence the agent should wait for before it can respond, which should be shorter than `min_endpointing_delay`.

livekit-agents/livekit/agents/voice/agent_activity.py
chenghao-mou: I have updated it to `self._session.options.min_endpointing_delay / 2` as a tentative threshold.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1611,7 +1624,8 @@ async def _tts_task_impl(
         audio_output = self._session.output.audio if self._session.output.audio_enabled else None
 
         wait_for_authorization = asyncio.ensure_future(speech_handle._wait_for_authorization())
-        await speech_handle.wait_if_not_interrupted([wait_for_authorization])
+        wait_for_user_silence = asyncio.ensure_future(self._user_silence_event.wait())
```
longcw, line 1627: similar to `wait_for_authorization`, `wait_for_user_silence` needs to be canceled if interrupted

Issue #3739, opened 2025-10-28
AWS STT plugin extremely slow when using multi-user transcription


pabloFuente:
### Bug Description

When changing the plugin in the example agent [**multi-user-transcriner.py**](https://github.com/livekit/agents/blob/main/examples/other/transcription/multi-user-transcriber.py) from deepgram to aws, the transcription starts getting **very slow with 3 or more participants connected to the same room**. When transcribing different participants from different Rooms, transcriptions behave well and arrive within a reasonable time.

This does not happen for any of the other [17 STT plugins](https://docs.livekit.io/agents/models/stt/#plugins). Only the Amazon Transcribe plugin behaves this way in this sample agent.

The agent log shows this warning message again and again:

```log
{"message": "inference is slower than realtime", "level": "WARNING", "name": "livekit.plugins.silero", "delay": 13.077118384041238, "pid": 52, "job_id": "AJ_zzYdH8TPndaD", "timestamp": "2025-10-28T19:12:35.291028+00:00"}
{"message": "inference is slower than realtime", "level": "WARNING", "name": "livekit.plugins.silero", "delay": 12.733135948984998, "pid": 52, "job_id": "AJ_zzYdH8TPndaD", "timestamp": "2025-10-28T19:12:35.293962+00:00"}
{"message": "inference is slower than realtime", "level": "WARNING", "name": "livekit.plugins.silero", "delay": 13.187841181981316, "pid": 52, "job_id": "AJ_zzYdH8TPndaD", "timestamp": "2025-10-28T19:12:35.457624+00:00"}
{"message": "inference is slower than realtime", "level": "WARNING", "name": "livekit.plugins.silero", "delay": 13.44968232001003, "pid": 52, "job_id": "AJ_zzYdH8TPndaD", "timestamp": "2025-10-28T19:12:35.458104+00:00"}
{"message": "inference is slower than realtime", "level": "WARNING", "name": "livekit.plugins.silero", "delay": 13.052407222982552, "pid": 52, "job_id": "AJ_zzYdH8TPndaD", "timestamp": "2025-10-28T19:12:35.645356+00:00"}
{"message": "inference is slower than realtime", "level": "WARNING", "name": "livekit.plugins.silero", "delay": 13.400238036039632, "pid": 52, "job_id": "AJ_zzYdH8TPndaD", "timestamp": "2025-10-28T19:12:35.648003+00:00"}
[...]

PR #4111, opened 2025-11-27
Replaced deprecated amazon-transcribe SDK with new aws-sdk-transcribe-streaming

pabloFuente:
The AWS Python SDK used by livekit-plugins-aws in the STT implementation ([`amazon-transcribe`](https://github.com/awslabs/amazon-transcribe-streaming-sdk)) has known **critical CPU bottleneck** performance issues and has been deprecated by AWS in favor of [`aws-sdk-transcribe-streaming`](https://github.com/awslabs/aws-sdk-python/tree/develop/clients/aws-sdk-transcribe-streaming).

- Link to related `amazon-transcribe` issue: https://github.com/awslabs/amazon-transcribe-streaming-sdk/issues/121
- Link to related `livekit/agents` issue: https://github.com/livekit/agents/issues/3739

This PR replaces the old, deprecated, low-performance AWS SDK with the new official AWS SDK. I have tested the problematic scenario explained in this issue (https://github.com/livekit/agents/issues/3739) and it is no longer a problem when using this PR. Transcribing multiple users from the same LiveKit Agent using livekit-plugin-aws STT is now possible, with no delays. The scope of the commit is quite small, so I do not expect any significant side effects.

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
@@ -190,7 +190,7 @@ class SpeechStream(stt.SpeechStream):
                 config_kwargs["aws_session_token"] = self._credentials.session_token
             else:
                 config_kwargs["aws_credentials_identity_resolver"] = (
-                    EnvironmentCredentialsResolver()  # type: ignore
+                    EnvironmentCredentialsResolver()
                 )
 
             client: TranscribeStreamingClient = TranscribeStreamingClient(
--- a/uv.lock
+++ b/uv.lock
@@ -272,18 +272,6 @@ wheels = [
 ]
 
 [[package]]
-name = "amazon-transcribe"
-version = "0.6.4"
-source = { registry = "https://pypi.org/simple" }
-dependencies = [
-    { name = "awscrt" },
-]
-sdist = { url = "https://files.pythonhosted.org/packages/0b/3a/f2bf812e8716d25fad092a21e2f01377d00c15e3a1e9be97d132a1d33850/amazon_transcribe-0.6.4.tar.gz", hash = "sha256:89770100e88446a451241f49ad14820bf3113f4545df0a32e82c73455928432c", size = 33687, upload-time = "2025-05-05T10:02:20.156Z" }
-wheels = [
-    { url = "https://files.pythonhosted.org/packages/91/c9/d13edab9292aff456f10250bc9d2ad7584de69d35878574e24631efac630/amazon_transcribe-0.6.4-py3-none-any.whl", hash = "sha256:ba5a27616106e659f031c975da12c331665d122d2518e2018a88b308e66fdee7", size = 41118, upload-time = "2025-05-05T10:02:19.202Z" },
-]
-
-[[package]]
 name = "annotated-types"
 version = "0.7.0"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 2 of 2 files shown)


pabloFuente:
Unit test `tests/test_chat_ctx.py` is failing with an authorization error:

```
Error code: 401 - {'error': {'message': "You didn't provide an API key. You need to provide your API key in an Authorization header..."}
```

I don't think this has anything to do with the changes in my PR.

livekit-plugins/livekit-plugins-aws/pyproject.toml
```diff
@@ -8,7 +8,7 @@ dynamic = ["version"]
 description = "LiveKit Agents Plugin for services from AWS"
 readme = "README.md"
 license = "Apache-2.0"
-requires-python = ">=3.9.0"
+requires-python = ">=3.12"
```
longcw: I think we should keep 3.9 since tts and llm are still supported by lower versions of python.

livekit-plugins/livekit-plugins-aws/pyproject.toml
```diff
@@ -18,20 +18,19 @@ classifiers = [
...
-    "Programming Language :: Python :: 3.9",
-    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3 :: Only",
 ]
 dependencies = [
     "livekit-agents>=1.3.5",
     "aioboto3>=14.1.0",
-    "amazon-transcribe>=0.6.4",
+    "aws_sdk_transcribe_streaming>=0.2.0",
```
longcw: add `python_version >= '3.12'` here and raise an error from the stt plugin if the `aws_sdk_transcribe_streaming` is not installed.

longcw (commented):
Thanks for creating the pr! Looks good to me, something nit:

livekit-plugins/livekit-plugins-aws/pyproject.toml
pabloFuente: Done https://github.com/livekit/agents/pull/4111/commits/c25967ac4ba8ae10754568bb37c4448b002131a9#diff-3a3f1fe01032d6446a5594e145a51d4a51fdd627f6eccde9c935aea3bbb36fd2L11-R11

livekit-plugins/livekit-plugins-aws/pyproject.toml
pabloFuente: And done: https://github.com/livekit/agents/commit/c25967ac4ba8ae10754568bb37c4448b002131a9#diff-bf9df96379f6c4bd68aa35f8e6a5126a9cab8878003cc1ec82f1e0135aa8f05f

I've also updated the uv.lock to reflect the changes in the aws SDK.

longcw (approved):
lgtm

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
```diff
@@ -145,36 +181,25 @@ def __init__(
...
     async def _run(self) -> None:
         while True:
-            client = TranscribeStreamingClient(
-                region=self._opts.region,
-                credential_resolver=self._credential_resolver(),
+            config_kwargs: dict[str, Any] = {"region": self._opts.region}
+            if self._credentials:
+                config_kwargs["aws_access_key_id"] = self._credentials.access_key_id
+                config_kwargs["aws_secret_access_key"] = self._credentials.secret_access_key
+                config_kwargs["aws_session_token"] = self._credentials.session_token
```
danial-relevanceai, line 190: Thank you @pabloFuente for this fix and @longcw for reviewing, while I wait for a release, I tried to patch this in and I believe we're missing the following in the case credentials are provided to the `STT`
```
                config_kwargs["aws_credentials_identity_resolver"] = (
                    StaticCredentialsResolver()
                )
```

The context is in deployed AWS environments when `self._credentials == None` and it uses `EnvironmentCredentialsResolver` did not work for me and the agents were absolutely silent with no error or debug logs as if session never started, still haven't figured out why. Regardless, since the `STT` module accepts `credentials` I provided that as a workaround passing it like so it gets tokens based on iam access of my containers.

```
            session = botocore.session.get_session()
            credentials = session.get_credentials()
            if credentials is None:
                raise RuntimeError("AWS credentials not found")
            frozen = credentials.get_frozen_credentials()
            return aws.STT(
                ...
                credentials=AWSCredentials(
                    access_key_id=frozen.access_key,
                    secret_access_key=frozen.secret_key,
                    session_token=frozen.token,
                ),
            )
```
but it was still not working and after some digging it seems the AWS SDK requires aws_credentials_identity_resolver to be set to `StaticCredentialsResolver` for it to use those properties

davidzhao: thanks for spotting this case @danial-relevanceai. do you want to open a PR to address this? with your approach, we'd need to make sure there is an async version of `get_frozen_credentials` since we cannot block the async runloop here.

PR #4126, opened 2025-12-01
add WarmTransferTask 

longcw:


examples/warm-transfer/warm_transfer.py
```diff
@@ -1,243 +1,68 @@
...
-logger = logging.getLogger("basic-agent")
-logger.setLevel(logging.DEBUG)
-logging.getLogger().setLevel(logging.DEBUG)
+logger = logging.getLogger("warm-transfer")
 
 load_dotenv()
 
 # ensure the following variables/env vars are set
 SIP_TRUNK_ID = os.getenv("LIVEKIT_SIP_OUTBOUND_TRUNK")  # "ST_abcxyz"
 SUPERVISOR_PHONE_NUMBER = os.getenv("LIVEKIT_SUPERVISOR_PHONE_NUMBER")  # "+12003004000"
```
chenghao-mou, line 25: nitpicking: Should we use `os.env["xxx"]` call here to fail fast? Otherwise, we need to handle the `None` cases later.

chenghao-mou: nitpicking: `SIP_TRUNK_ID` is not used here.

longcw: it can be removed, we read `LIVEKIT_SIP_OUTBOUND_TRUNK` inside the task as well.

longcw:
I tested using agents-playground and RingCentral SIP and they can here each other, perhaps your caller from browser didn't sub to the new audio track automatically?

chenghao-mou:
> I tested using agents-playground and RingCentral SIP and they can here each other, perhaps your caller from browser didn't sub to the new audio track automatically?

Sorry, I was not doing explicit dispatch. Once I have done that, it works properly.

livekit-agents/livekit/agents/voice/room_io/types.py
```diff
@@ -168,6 +168,8 @@ def _ensure_options(
 
         if isinstance(options, RoomOptions):
             return options
+        elif is_given(options):
+            raise ValueError(f"expect RoomOptions, got {type(options)}")
```
theomonnom: nitpicking
```suggestion
            raise ValueError(f"expected RoomOptions, got {type(options)}")
```

livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -433,7 +440,10 @@ def _on_delete_room_task_done(task: asyncio.Future[api.DeleteRoomResponse]) -> N
         if self._options.delete_room_on_close and self._delete_room_task is None:
             job_ctx = get_job_context()
             logger.info(
-                "deleting room on agent session close (disable via `RoomInputOptions.delete_room_on_close=False`)"
+                "deleting room on agent session close (disable via `RoomInputOptions.delete_room_on_close=False`)",
+                extra={
+                    "deleted_room": self._room.name,
```
theomonnom:
```suggestion
                    "room": self._room.name,
```
Let's always use the same attributes for common livekit features. It makes it way easier to search for logs just based on the room name

livekit-agents/livekit/agents/beta/workflows/warm_transfer.py
```diff
@@ -0,0 +1,352 @@
...
+            chat_ctx=NOT_GIVEN,  # don't pass the chat_ctx
+            turn_detection=turn_detection,
+            tools=tools or [],
+            stt=stt,
+            vad=vad,
+            llm=llm,
+            tts=tts,
+            allow_interruptions=allow_interruptions,
+        )
+
```
theomonnom, line 96: Can we add something in the constructor checking for the JobContext and raising an error if this isn't using Cloud (MoveParticipant is only supported on Cloud)

theomonnom: Maybe doing it on_enter is fine too

livekit-agents/livekit/agents/beta/workflows/warm_transfer.py
```diff
@@ -0,0 +1,352 @@
...
+        self._sip_trunk_id = (
+            sip_trunk_id if is_given(sip_trunk_id) else os.getenv("LIVEKIT_SIP_OUTBOUND_TRUNK", "")
+        )
+        if not self._sip_trunk_id:
+            raise ValueError(
+                "`LIVEKIT_SIP_OUTBOUND_TRUNK` environment variable or `sip_trunk_id` argument must be set"
+            )
+
+        # background audio and io
+        self._background_audio = BackgroundAudioPlayer()
```
theomonnom, line 112: I never really tested that, but do we know if two BackgroundAudio could conflict between each other?

livekit-agents/livekit/agents/beta/workflows/warm_transfer.py
```diff
@@ -0,0 +1,352 @@
...
+        self._hold_audio_handle: PlayHandle | None = None
+        self._hold_audio = (
+            cast(Optional[Union[AudioSource, AudioConfig, list[AudioConfig]]], hold_audio)
+            if is_given(hold_audio)
+            else AudioConfig(BuiltinAudioClip.HOLD_MUSIC, volume=0.8)
+        )
+
+        self._original_io_state: dict[str, bool] = {}
+
+    def get_instructions(
```
theomonnom, line 122: I'll try to think more about prompt templates today, ideally we have something similar for every other task we have

examples/warm-transfer/warm_transfer.py
```diff
@@ -254,60 +79,31 @@ async def transfer_to_human(self, context: RunContext):
...
-        # Assistant: I see, looks like your account has been locked out.
-        # Customer: Can you help me?
-        # Assistant: I'm not able to help with that, would you like to speak to a human agent?
-        # Customer: Yes please.
-        # """
-
-        super().__init__(
-            instructions=_supervisor_agent_instructions + "\n\n" + prev_convo,
+        await self.session.say(
+            "you are on the line with my supervisor. I'll be hanging up now.",
```
theomonnom, line 106: Are we sure the human cannot speak at the same time?

examples/warm-transfer/warm_transfer.py
```diff
@@ -254,60 +79,31 @@ async def transfer_to_human(self, context: RunContext):
...
+            assert SUPERVISOR_PHONE_NUMBER is not None
 
+            result = await WarmTransferTask(
+                target_phone_number=SUPERVISOR_PHONE_NUMBER,
+                sip_trunk_id=SIP_TRUNK_ID,
+                chat_ctx=self.chat_ctx,
+            )
+        except Exception as e:
+            logger.error(f"failed to transfer to supervisor: {e}")
+            raise e
```
theomonnom, line 96: Let's raise a ToolError instead

livekit-agents/livekit/agents/voice/background_audio.py
```diff
@@ -30,6 +30,7 @@ class BuiltinAudioClip(enum.Enum):
     OFFICE_AMBIENCE = "office-ambience.ogg"
     KEYBOARD_TYPING = "keyboard-typing.ogg"
     KEYBOARD_TYPING2 = "keyboard-typing2.ogg"
+    HOLD_MUSIC = "hold_music.mp3"
```
Hormold: maybe we can reduce the size of the music file a bit using same ogg format

livekit-agents/livekit/agents/beta/workflows/warm_transfer.py
```diff
@@ -0,0 +1,352 @@
...
+Once the human agent has confirmed, you should call the tool `connect_to_caller` to connect them to the caller.
+
+Start by giving them a summary of the conversation so far, and answer any questions they might have.
+
+## Conversation history with caller
+{conversation_history}
+## End of conversation history with caller
+
+You are talking to the human agent now,
+give a brief introduction of the conversation so far, and ask if they want to connect to the caller.
```
Hormold, line 57: Maybe we can include something like that:

TASK: Write from your perspective as the AI assistant who participated in this call:
- WHO you're talking to (name, role, company if mentioned)
- WHY they contacted you (goal, problem, request)
- WHY a human agent is requested or needed at this point
- Brief summary in 100-200 characters from a first-person perspective

Hormold: Ideally and naturally it should be something like:

> Hey, this is a customer from XYZ Health. They called because their reception agent isn’t routing SIP calls correctly. I’ve verified credentials and connection, but they need someone from engineering to check the server config

>I'm about to connect you with John from GreenTech Solar. He asked about pricing and contract terms for the AI call assistant. I walked him through the basic plan, but he wants to confirm enterprise options with a human representative.

livekit-agents/livekit/agents/voice/room_io/room_io.py
longcw: one issue I noticed is that we may add a `room` to `ctx.log_context_fields` it overwrites the value here, maybe we shouldn't overwrite the value in https://github.com/livekit/agents/blob/livekit-agents@1.3.5/livekit-agents/livekit/agents/job.py#L222?
```python
    ctx.log_context_fields = {
        "room": ctx.room.name,
    }
```

livekit-agents/livekit/agents/beta/workflows/warm_transfer.py
longcw: just tested and it works well with two BackgroundAudioPlayer, since they publish audio tracks separately.

examples/warm-transfer/warm_transfer.py
longcw: the human can speak actually, this message is said after human participant moved to the room, I don't think we have a way to or need to mute the human?

livekit-agents/livekit/agents/beta/workflows/warm_transfer.py
longcw: that's a good point. let's keep the default one simple right now, users can add more specific instructions in `WarmTransferTask(..., extra_instructions="...")`

davidzhao: I mainly worry about the flexibility of a fixed instructions. our messaging is very generic and doesn't include:
- what the use case is. i.e. for debt collection it could be something like: a customer has agreed to pay off their debt.
- customer information, who are they, any information the primary agent has pulled from a DB/etc

Perhaps our example should show both of these cases via extra_instructions?

PR #4131, opened 2025-12-01
AGT-2316: refine timestamps in spans and recording alignment

chenghao-mou:
- Update the user speech start time to include the VAD speech duration;
- Audio capture in output is now synced with the audio source capture, based on @longcw's PR #4149 
- User speech is now padded if the mic is enabled after the agent's first speech;

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1662,7 +1662,7 @@ class AgentActivity(RecognitionHooks):
             try:
                 started_speaking_at = fut.result() or time.time()
             except BaseException:
-                started_speaking_at = time.time()
+                return
 
             self._session._update_agent_state(
                 "speaking",
@@ -1741,7 +1741,11 @@ class AgentActivity(RecognitionHooks):
         if speech_handle.interrupted and audio_output is not None:
             playback_ev = await audio_output.wait_for_playout()
 
-            if audio_out is not None and audio_out.first_frame_fut.done():
+            if (
+                audio_out is not None
+                and audio_out.first_frame_fut.done()
+                and not audio_out.first_frame_fut.cancelled()
+            ):
                 if playback_ev.synchronized_transcript is not None:
                     forwarded_text = playback_ev.synchronized_transcript
             else:
@@ -1946,7 +1950,8 @@ class AgentActivity(RecognitionHooks):
             try:
                 started_speaking_at = fut.result() or time.time()
             except BaseException:
-                started_speaking_at = time.time()
+                return
+
             self._session._update_agent_state(
                 "speaking",
                 start_time=started_speaking_at,
@@ -2034,7 +2039,11 @@ class AgentActivity(RecognitionHooks):
                 audio_output.clear_buffer()
 
                 playback_ev = await audio_output.wait_for_playout()
-                if audio_out is not None and audio_out.first_frame_fut.done():
+                if (
+                    audio_out is not None
+                    and audio_out.first_frame_fut.done()
+                    and not audio_out.first_frame_fut.cancelled()
+                ):
                     # playback_ev is valid only if the first frame was already played
                     if playback_ev.synchronized_transcript is not None:
                         forwarded_text = playback_ev.synchronized_transcript
@@ -2310,7 +2319,8 @@ class AgentActivity(RecognitionHooks):
             try:
                 started_speaking_at = fut.result() or time.time()
             except BaseException:
-                started_speaking_at = time.time()
+                return
+
             self._session._update_agent_state(
                 "speaking",
                 start_time=started_speaking_at,
@@ -2523,7 +2533,11 @@ class AgentActivity(RecognitionHooks):
 
                     playback_ev = await audio_output.wait_for_playout()
                     playback_position = playback_ev.playback_position
-                    if audio_out is not None and audio_out.first_frame_fut.done():
+                    if (
+                        audio_out is not None
+                        and audio_out.first_frame_fut.done()
+                        and not audio_out.first_frame_fut.cancelled()
+                    ):
                         # playback_ev is valid only if the first frame was already played
                         if playback_ev.synchronized_transcript is not None:
                             forwarded_text = playback_ev.synchronized_transcript
--- a/tests/fake_io.py
+++ b/tests/fake_io.py
@@ -51,6 +51,7 @@ class FakeAudioOutput(AudioOutput):
 
         if not self._pushed_duration:
             self._start_time = time.time()
+            self.on_playback_started(created_at=self._start_time)
         self._pushed_duration += frame.duration
 
     def flush(self) -> None:


livekit-agents/livekit/agents/voice/recorder_io/recorder_io.py
```diff
@@ -259,14 +263,35 @@ def __init__(self, *, recording_io: RecorderIO, source: io.AudioInput) -> None:
...
+        ):
+            logger.warning(
+                "input speech started after last agent speech ended",
+                extra={
+                    "last_agent_speech_time": pad_since,
+                    "input_started_time": self.__started_time,
+                },
+            )
+            self.__padded = True
+            frames = [
```
longcw, line 290: if I understand correctly, the issue is the silent duration is not counted when both input and output buffers are empty. so it's not only the duration after the first response to the time user enables the mic, but also before the first response generated (LLM + TTS time).

can we record the time the `RecordIO.start` called as `last_take_time`, then modify `RecorderAudioInput.take_buf` to return a silence frame with duration `current_time - last_take_time` before there is any frame added to the input buffer?

then the total padding duration should be `started_wall_time - RecordIO.start_wall_time`

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2654,7 +2675,9 @@ def _on_false_interruption() -> None:
                 and not self._paused_speech.done()
             ):
                 self._session._update_agent_state(
-                    "speaking", otel_context=self._paused_speech._agent_turn_context
+                    "speaking",
+                    otel_context=self._paused_speech._agent_turn_context,
+                    start_time=time.time_ns(),
```
longcw:
```suggestion
                    start_time_ns=time.time_ns(),
```
to make it more explicit?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1225,7 +1230,7 @@ def on_end_of_speech(self, ev: vad.VADEvent | None) -> None:
             speech_end_time = speech_end_time - ev.silence_duration
         self._session._update_user_state(
             "listening",
-            last_speaking_time=speech_end_time,
+            last_speaking_time=int(speech_end_time * 1_000_000_000),
```
longcw: since for most of case we use seconds for the time arguments, can we keep them as seconds, and convert it to ns in the `update_user_state` and `update_agent_state` for consistency?

livekit-agents/livekit/agents/voice/recorder_io/recorder_io.py
chenghao-mou: Screenshot 2025-12-04 at 10 04 50

Not sure how, but it seems we must have padding somewhere outside agents: 

output speech started at 1764842333.011143 (first output frame capture time)
input speech started at 1764842347.058694 (first input frame capture time)
agent speech span started at 1764842333.011
user speech span started at 1764842347.058
but we have the recording started almost 1 second before the agent turn even started.

livekit-agents/livekit/agents/voice/agent_activity.py
chenghao-mou: Good point, will do!

livekit-agents/livekit/agents/voice/recorder_io/recorder_io.py
longcw: sorry what was the issue in this case, the audio wave looks matched with the span, do you mean it matched because you added padding outside agents, or you think it's not matched?

chenghao-mou: It is matched, but we didn't do any padding in agents before the first `agent_speaking` AFAIK.

chenghao-mou: Here is what we have confirmed: Audio recording in observability only starts playing when the cursor hits the `start_wall_time`/(first agent or user frame capture time), essentially padding silence in both channels;

But if we are padding this in agents, we are essentially pushing the `RecordIO.start_wall_time` to `last_take_time`, right?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1638,10 +1641,14 @@ async def _read_text() -> AsyncIterable[str]:
         started_speaking_at: float | None = None
         stopped_speaking_at: float | None = None
 
-        def _on_first_frame(_: asyncio.Future[None]) -> None:
+        def _on_first_frame(fut: asyncio.Future[float] | asyncio.Future[None]) -> None:
```
theomonnom, line 1655: This isn't obvious that the float inside the future was for the started_speaking_at

livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -57,6 +61,9 @@ def __init__(
         self._playback_enabled = asyncio.Event()
         self._playback_enabled.set()
 
+        self._first_frame: rtc.AudioFrame | None = None
```
theomonnom: Is there any reason to store the first_frame?

livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -97,9 +104,23 @@ async def capture_frame(self, frame: rtc.AudioFrame) -> None:
...
+                self._first_frame = f
+                self._first_frame_fut = asyncio.Future[None]()
+
             await self._audio_buf.send(f)
             self._pushed_duration += f.duration
 
+            # wait for the first frame to be captured
+            if self._first_frame and self._first_frame_fut:
+                try:
+                    await self._first_frame_fut
```
theomonnom: I'm not sure to understand why we're doing this? 
Why do we need to wait for the first frame before directly pushing the rest?

livekit-agents/livekit/agents/voice/agent_activity.py
chenghao-mou: Good point. I will add a comment!

livekit-agents/livekit/agents/voice/room_io/_output.py
chenghao-mou: We start a speech span and calculate pushed duration based on upstream frame push time: 
1. this sometimes create a span a few hundred ms early than the actual speech;
2. if the speech is interrupted before the frame reaches the audio source, we would have created short but invalid spans and recordings

So this makes sure the audio source captures the first frame before committing more frames. The other approach I was thinking of is to create an `on_playback_started` event so callers can use that for span creation or recording processing.

livekit-agents/livekit/agents/voice/room_io/_output.py
chenghao-mou: Good question. I think this is more of a defensive implementation by checking both the future and the identity. I can simplify it.

livekit-agents/livekit/agents/voice/recorder_io/recorder_io.py
```diff
@@ -259,14 +263,42 @@ def __init__(self, *, recording_io: RecorderIO, source: io.AudioInput) -> None:
...
+                _create_silence_frame(padding, frames[0].sample_rate, frames[0].num_channels),
+                *frames,
+            ]
+        # we could pad with silence here with some fixed SR and channels,
+        # but it's better for the user to know that this is happening
+        elif pad_since and self.__started_time is None and not self.__padded and not frames:
+            logger.warning(
```
theomonnom, line 297: Could this be an issue if the user is muted by default?

chenghao-mou: I think muting is fine. It still generates silence frames. The only case this would happen is when the user doesn't allow access to the microphone at the browser level at all.

livekit-agents/livekit/agents/voice/room_io/_output.py
theomonnom: I think it would make more sense to introduce a new `on_playback_started` event. It's going to make things more explicit

chenghao-mou: I don't really have a strong opinion. cc @longcw would love to hear your thoughts as well.

chenghao-mou: This is now updated to use an `on_playback_started` event.

PR #4132, opened 2025-12-01
fix: Remove duplicate _mark_playout_done() call in PlayHandle.stop()

joshiayush:
Fixes a race condition and duplicate logic in `PlayHandle.stop()` by centralizing the `_mark_playout_done()` call to only occur in the `_play_task` method. This addresses the TODO comment at line 414 in `background_audio.py`.


joshiayush:
@davidzhao  Is the following still needed inside the wrapper? It is already getting called in the `finally` block.

```python
            # TODO(theomonnom): the wait_for_playout() may be innaccurate by 400ms
            play_handle._mark_playout_done()
```

PR #4133, opened 2025-12-01
add job_executor_type property to AgentServer

joshiayush:
Add proper property setter for job_executor_type in AgentServer

### DIFF
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -1205,7 +1205,7 @@ class _ConsoleWorker:
                     self._shutdown_cb()
                     return
 
-            self._server._job_executor_type = JobExecutorType.THREAD  # TODO: better setter
+            self._server.job_executor_type = JobExecutorType.THREAD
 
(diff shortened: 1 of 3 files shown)


joshiayush:
@theomonnom

joshiayush:
@davidzhao

theomonnom:
Users can already set the executor type inside the constructor. 

We also have update_options: https://github.com/livekit/agents/blob/80dca444aa6a6dc4bc7356b85a9df74194346cdf/livekit-agents/livekit/agents/worker.py#L753

livekit-agents/livekit/agents/cli/cli.py
```diff
@@ -1205,7 +1205,7 @@ async def _async_main() -> None:
                     self._shutdown_cb()
                     return
 
-            self._server._job_executor_type = JobExecutorType.THREAD  # TODO: better setter
+            self._server.job_executor_type = JobExecutorType.THREAD
```
joshiayush, line 1208: @theomonnom Is it that in this case this should be `self._server.update_options()`?

Issue #3760, opened 2025-10-30
Say and ensure is added to chat context


marctorsoc:
### Feature Type

Nice to have

### Feature Description

I want to use `session.say`, and allow interruptions but if I do so and the user interrupts, the message is half added to the chat context, only to the point that was pronounced. Could you add some param so say to allow to do so? I was not expecting this default

### Workarounds / Alternatives

I wrote
```
async def safe_say(self, text: str, allow_interruptions: bool = True):
        """
        Say something, and make sure it's added to the context
        """
        chat_ctx = self.chat_ctx.copy()
        chat_ctx.add_message(content=[text], role="assistant")
        await self.update_chat_ctx(chat_ctx)
        return await self.session.say(
            text,
            allow_interruptions=allow_interruptions,
            add_to_chat_ctx=False,
        )
```

But then I see that `session.history` (which is what I use to create the full transcript when finishing the session) does not have it :(

davidzhao:
The current behavior is.. when `add_to_chat_ctx` is true and the agent is interrupted, we'll be adding a partial transcript to the context, up to the point where the agent was interrupted. This is needed because otherwise the LLM does not have the context in subsequent generations..

what behavior are you looking for?

marctorsoc:
I want to add the full message as if it was not interrupted. The reason is that if it's a false interruption, the agent won't understand to which stage we arrived e.g. in our case booking confirmed

davidzhao:
I see, have you tried turning on false interruption handling? [https://docs.livekit.io/agents/build/turns/#false-interruptions](https://docs.livekit.io/agents/build/turns/#false-interruptions)

davidzhao:
related, if it's a confirmation of sorts after an action, I would recommend not allowing users to interrupt that. <br><br>though your original ask seems valid. basically that changes made by `update_chat_ctx` is not kept in `session.history` afterwards. we'll look into that

marctorsoc:
yes, this is my current setup
```
def get_turn_detection_config(ctx: JobContext) -> dict:
    """
    We use EnglishModel for turn detection. This helps the VAD (Voice Activity Detector)
    to understand when to stop by providing linguistic context. Otherwise, it'd be based
    on audio activity only. Note that the STT is config'd with language=English to match this
    Note that this requires `uv run main.py download-files` before `uv run main.py dev`
    multilingual model is slower - using English for now
    turn_detection=MultilingualModel(),

    """
    config: dict[str, Any] = dict()

[...]

PR #4137, opened 2025-12-01
Item level merge control

parkerbjur:
Addresses @davidzhao's [comment](https://github.com/livekit/agents/issues/3760#issuecomment-3468978385). 
> changes made by update_chat_ctx [should not be] kept in session.history afterwards.

- I've added the prop `should_merge` to all `ChatItem` types (default `True`)
- I've updated `ChatContext.merge()` to take this parameter into account
- I've updated `ChatContext.add_message()` to accept `should_merge` as a param (default `NOT_GIVEN`)

When should_merge is set to false on a `ChatItem`, `ChatContext.merge` will skip over that item when performing the merge. This allows a user to manually add context to an agent's chat history without it being committed to `session.history` or merged during an agent handoff. To maintain backwards compatibility, I have set the default value of `should_merge` to `True`. 

Edit: This actually only addresses agent handoffs, not session history. will open a new PR shortly. 
Edit 2: [PR](https://github.com/livekit/agents/pull/4140)


theomonnom:
I'm not sure if we want to include states inside ChatItem, it doesn't feel like it is the right place to do it. Especially since they're also going to get serialized

PR #4140, opened 2025-12-01
add persist_to_session_history prop to ChatItem

parkerbjur:
- I've added the prop `persist_to_session_history` to all `ChatItem` types (default `True`)
- I've updated `AgentSession._conversation_item_added()`, `AgentSession._tool_items_added()`, and 
  `AgentSession._update_activity()` to respect this parameter
 - I've updated `ChatContext.add_message()` to accept `persist_to_session_history` as a param (default `NOT_GIVEN`)

When `persist_to_session_history` is set to `False` on a `ChatItem`, the item will be used in the agent's chat context for LLM calls but will not be persisted to `session.history`. This allows agents to add ephemeral context that influences the current response without polluting the permanent conversation history. To maintain backwards compatibility, I have set the default value of `persist_to_session_history` to `True`.

### DIFF
--- a/livekit-agents/livekit/agents/llm/chat_context.py
+++ b/livekit-agents/livekit/agents/llm/chat_context.py
@@ -159,6 +159,7 @@ class ChatMessage(BaseModel):
     metrics: MetricsReport = Field(default_factory=lambda: MetricsReport())
     created_at: float = Field(default_factory=time.time)
     hash: bytes | None = Field(default=None, deprecated="hash is deprecated")
+    persist_to_session_history: bool = True
 
     @property
     def text_content(self) -> str | None:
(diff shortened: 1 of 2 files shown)


theomonnom:
Do you have an example on how it was polluting the AgentSession history before? If you manually add items inside the `Agent.chat_ctx`, it isn't going to be added inside the `AgentSession.history`

parkerbjur:
So I actually read David's prescription incorrectly in [this issue](https://github.com/livekit/agents/issues/3760#issuecomment-3468978385). I understood it as saying that changes made via `update_chat_ctx` should not be kept in `session.history`, implying they were right now. Instead, he's saying they aren't being saved right now and this user is wanting a way to persist them to the session.history. 

Do you think it's reasonable to keep the shape of this change (i.e. adding a prop to the ChatItem to control session persistence). Obviously I'll need to change the rest of the behavior to ensure it does persist when needed. 

Or do you think it's more reasonable to add an argument to the `update_chat_ctx` function which will control if those items will be added to session history.

theomonnom:
Should users manually insert items inside session.history if they want to persist some "custom" items?

PR #4145, opened 2025-12-02
Add support for audio frame processor

lukasIO:
depends on https://github.com/livekit/python-sdks/pull/533 and https://github.com/livekit/python-sdks/pull/534


livekit-agents/livekit/agents/voice/room_io/_input.py
```diff
@@ -209,37 +233,35 @@ def __init__(
...
-        noise_cancellation: rtc.NoiseCancellationOptions | NoiseCancellationSelector | None,
+        noise_cancellation: rtc.NoiseCancellationOptions
+        | rtc.FrameProcessor[rtc.AudioFrame]
+        | None,
         pre_connect_audio_handler: PreConnectAudioHandler | None,
-        frame_size_ms: int = 50,
     ) -> None:
+        audio_processor: rtc.FrameProcessor[rtc.AudioFrame] | None = None
+        if isinstance(noise_cancellation, rtc.FrameProcessor):
+            audio_processor = noise_cancellation
```
longcw, line 248: python sdk already handles the `noise_cancellation` if is a `FrameProcessor` in https://github.com/livekit/python-sdks/pull/533/files#diff-7e2485e8b1480ff233d810a2d222fe14cb1432d0837ce28a6e6182ae8b2422b6R112-R113, here it's processed duplicately?

lukasIO: Ah, sorry, you're right, I meant to delete the `processor.process` call within the input handler and forgot. 
I'll update it right away, thanks for catching that!

The original intention:
The processor is forwarded directly to the audio stream.
The audio stream however does not have any participant and/or room context.
In order to hook into room+participant+publication info, the processor is passed to the input handler additionally which allows the stream metadata to be updated.

lukasIO: addressed in https://github.com/livekit/agents/pull/4145/commits/f9e324cb840fee8b74271f68dfd3de4a2a833eff

Does that look good to you now?

livekit-agents/livekit/agents/voice/room_io/_input.py
```diff
@@ -198,6 +214,14 @@ def _on_track_unavailable(
             if self._on_track_available(publication.track, publication, participant):
                 return
 
+    def _on_token_refreshed(self) -> None:
+        if (
+            self._processor is not None
+            and self._room._token is not None
+            and self._room._server_url is not None
+        ):
+            self._processor._update_credentials(token=self._room._token, url=self._room._server_url)
```
theomonnom: Do you think the livekit-rtc `package` should be responsible for directly updating the credentials?

lukasIO: agree, in theory that would be nice.

I wasn't sure what the best way would be for handling the possibility of the input stream changing here further down in the rtc SDK. More than the credential update this would be about the stream info update.
 
Any thoughts on this?

lukasIO: Had a look at this from the rtc SDK side. 

One challenge I see with shifting responsibility for this to the rtc SDK is that there's no control from within the rtc SDK over when/how users construct audio streams, do you have an idea how to achieve a tighter integration between stream and the room instance?

theomonnom: I see, yes, it's probably fine for now.
Maybe in the future, rtc.Room should keep track of every rtc.AudioStream, I don't see an easy way to do it today tho

livekit-agents/livekit/agents/voice/room_io/_input.py
```diff
@@ -48,6 +49,9 @@ def __init__(
 
         self._room.on("track_subscribed", self._on_track_available)
         self._room.on("track_unpublished", self._on_track_unavailable)
+        self._room.on("token_refreshed", self._on_token_refreshed)
```
xianshijing-lk, line 53: nit,
in line 127,  function aclose(self) -> None:
add 
self._room.off("token_refreshed", self._on_token_refreshed) ?

xianshijing-lk: btw, I think we should also cancel the ongoing task in aclose(self) function, like
for t in list(self._tasks):
        await aio.cancel_and_wait(t)
    self._tasks.clear()
    
    which will syncrhonously wait and cancel the ongoing tasks

lukasIO:
> lgtm! but how are users supposed to use this API?


forgot to add the type to the AudioInputOptions 🤦
I've now added it also for Ben's new NoiseCancellationSelector

livekit-agents/livekit/agents/voice/room_io/_input.py
lukasIO: I implemented the first suggestion. 
Will defer to @theomonnom for the task cancelling suggestion

xianshijing-lk (approved):

lukasIO:
ruff checks are failing on main as well due to the durable stuff, going to merge this

PR #4146, opened 2025-12-02
use log filter for log_context_fields

longcw:
the logging will raise `KeyError("Attempt to overwrite %r in LogRecord" % key)` if user add fields to log records that already existed in `log_context_fields`.

replace the record factory with log filter so the `log_context_fields` is added after log created and will ignore the fields already set.

address https://github.com/livekit/agents/pull/4126#discussion_r2579433155

### DIFF
--- a/livekit-agents/livekit/agents/job.py
+++ b/livekit-agents/livekit/agents/job.py
@@ -103,6 +103,32 @@ DEFAULT_PARTICIPANT_KINDS: list[rtc.ParticipantKind.ValueType] = [
 ]
 
 
+class _ContextLogFieldsFilter(logging.Filter):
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/job.py
```diff
@@ -157,7 +184,10 @@ def __init__(
         self._lock = asyncio.Lock()
 
     def _on_setup(self) -> None:
-        pass
+        root_logger = logging.getLogger()
+        for handler in root_logger.handlers:
+            handler.addFilter(self._log_filter)
```
theomonnom, line 189: Can we directly add the filter to the logger, instead of the handlers? 

Users may add more handler after `on_setup` is called, and we will not add any filter on them

PR #4150, opened 2025-12-02
Gradium integration.

LaurentMazare:


livekit-plugins/livekit-plugins-gradium/livekit/plugins/gradium/tts.py
```diff
@@ -0,0 +1,392 @@
...
+    utils,
+)
+from livekit.agents.types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, NotGivenOr
+from livekit.agents.utils import is_given
+
+from .log import logger
+
+ssl_context = ssl.create_default_context()
+ssl_context.check_hostname = False
+ssl_context.verify_mode = ssl.CERT_NONE
```
theomonnom: Is this still expected?

livekit-plugins/livekit-plugins-gradium/livekit/plugins/gradium/stt.py
```diff
@@ -0,0 +1,418 @@
...
+from .log import logger
+
+STTEncoding = Literal["pcm_s16le"]
+
+# Define bytes per frame for different encoding types
+bytes_per_frame = {
+    "pcm_s16le": 2,
+}
+
+ssl_context = ssl._create_unverified_context()
```
theomonnom: Same here

livekit-plugins/livekit-plugins-gradium/livekit/plugins/gradium/tts.py
LaurentMazare: Ah no it's not expected, I've just removed it, thanks for catching this!

livekit-plugins/livekit-plugins-gradium/livekit/plugins/gradium/stt.py
LaurentMazare: Thanks for catching this one too.

LaurentMazare:
Thanks for the prompt review and all your help along the way @theomonnom !

PR #4155, opened 2025-12-03
AGT-2302: add aligned_transcript to STT

chenghao-mou:
- Add aligned_transcript capability to STT
- Improve start_time and end_time tracking for all vendors (useful for spans/traces and multimodal eot). Now they are all relative to the audio input start time or session time if it is not available.

Tested and confirmed most of the ones using `SpeechData`, but skipped a few:
- Baseten
- Ones that don't support the feature at all

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -579,7 +579,6 @@ class SpeechStream(stt.SpeechStream):
 
         speech_data = stt.SpeechData(
             language=language,
-            # TODO: update gateway to return wall times since we do retry/reconnects internally @chenghao-mou
             start_time=self.start_time_offset + data.get("start", 0),
             end_time=self.start_time_offset + data.get("start", 0) + data.get("duration", 0),
             confidence=data.get("confidence", 1.0),


livekit-agents/livekit/agents/voice/io.py
```diff
@@ -43,22 +43,6 @@
...
-        text: str,
-        start_time: NotGivenOr[float] = NOT_GIVEN,
-        end_time: NotGivenOr[float] = NOT_GIVEN,
-    ) -> TimedString:
-        obj = super().__new__(cls, text)
-        obj.start_time = start_time
-        obj.end_time = end_time
-        return obj
-
-
```
theomonnom, line 61: nit: any reason to move this?

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -72,6 +79,7 @@ class STTCapabilities:
     streaming: bool
     interim_results: bool
     diarization: bool = False
+    aligned_transcript: Literal["word", "chunk", False] = False
```
theomonnom, line 82: What does chunk means? Like a sentence for e.g?

or it can be smaller than a word, e.g vowels?

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -184,7 +184,9 @@ def __init__(
             extra_kwargs (dict, optional): Extra kwargs to pass to the STT model.
         """
         super().__init__(
-            capabilities=stt.STTCapabilities(streaming=True, interim_results=True),
+            capabilities=stt.STTCapabilities(
+                streaming=True, interim_results=True, aligned_transcript="chunk"
+            ),
```
theomonnom: Do we support it on the gateway? Should we have aligned_transcript=None instead ?

livekit-agents/livekit/agents/voice/io.py
chenghao-mou: To avoid circular import so we can import it in `stt.py`:
1. `io` imports `stt` for `SpeechEvent`
2. `stt` imports `TimedString` from `io`

livekit-agents/livekit/agents/stt/stt.py
chenghao-mou: For a specific audio chunk (one or more audio frames), we might only have start and end timestamps, but nothing at word-level. Some might refer it as segment.

livekit-agents/livekit/agents/inference/stt.py
chenghao-mou: Yes, we have some information. But it is somewhat confusing: see [Slack](https://live-kit.slack.com/archives/C07TWVC0W4A/p1765207021561289)

chenghao-mou: Turns out inference does support word level timestamps. But the resetting issue still exists, unless we normalize them into wall times in both agents and gateway.

chenghao-mou: Inference now has been patched with properly start times and durations.

theomonnom (approved):
lgtm! nice!!

Issue #3240, opened 2025-08-23
Prewarm the LLM before session starts


davidzhao:
Currently (1.2) we prewarm the connection with STT and TTS, but not with the LLMs.

This is because LLMs usually require us to perform an inference request, instead of just opening a HTTP connection to `/`. There are a bunch of benefits to prewarming the LLM, the primary one is we could bypass the initial connection setup time, which could add up to be ~2s (DNS, SSL roundtrips).

galigutta:
looking for this as well. i am having better results with gpt-5 but the latency is unbearable on the first turn. This becomes more immediate. i am seeing 4+ seconds for first calls vs 1.2 seconds for cached calls.

abhismatrix1:
@davidzhao where do you prewarm STT and TTS in livekit agents. i guess its not implemented for any plugins. We are seeing very high first speech latency from agent after user message. Can you hep me how you prewarm STT and TTS.

ss14:
@davidzhao Is there a guide how to prewarm LLMs explicitly?

anlagbr:
Are there any updates on this?

Pulkit0729:
Hi @davidzhao,

I've been looking into LLM prewarming and wanted to share my findings.

You mentioned that LLMs typically require an inference request for prewarming, unlike STT/TTS which can just open an HTTP connection to `/`. However, I discovered that for **OpenAI and other HTTP-based LLM APIs**, we can actually achieve the same connection prewarming benefits without needing to send a full inference request.

For **self-hosted LLMs**, sending an inference request during prewarm makes sense to "wake up" the model and load it into memory. However, for **public LLM services** (OpenAI, Anthropic, Google, etc.), the models are already running and serving requests globally. In this case, the primary latency bottleneck is the **client-side connection establishment** (DNS, TCP, TLS), not the model availability.


I've implemented a `prewarm()` method for the OpenAI LLM plugin that:
1. Makes a lightweight GET request to `/` in a background task
2. Establishes the HTTP connection (DNS resolution, TCP handshake, TLS negotiation)


## Test Results



<img width="1170" height="152" alt="Image" src="https://github.com/user-attachments/assets/a02b64cb-5270-41a4-8e0d-0c8eb2e070c5" />
If you believe this approach is correct, I'd be happy to submit a PR for your review!
Let me know your thoughts!

marctorsoc:
@Pulkit0729 could you share your code even if dirty?

Pulkit0729:
[...]

PR #4167, opened 2025-12-03
feat(plugins): add prewarm() support for LLM providers

joshiayush:
Implements connection prewarming for LLM plugins to eliminate the initial response delay that's particularly problematic for telephony agents. This follows the existing prewarming pattern already established for STT and TTS services. (fixes: #3240)


joshiayush:
@davidzhao

jp-paper:
@joshiayush I understand that this PR adds an "internal" prewarm, but do you think this could/should be extended to support  explicit caching, for example with Gemini (https://ai.google.dev/gemini-api/docs/caching?lang=python#explicit-caching)?

I am working on doing this in the Worker setup function right now, which requires a separate GenAI client, but I wonder if this LLM prewarm function could potentially be a more appropriate place for this, design-wise?

joshiayush:
@jp-paper I understood what you are referring to and it's actually a good idea, but shouldn't that be a part of a separate PR, after this one is merged, a successor?

What are you caching right now in the worker setup function? tool definitions? what?

@davidzhao Should that be added to agents what @jp-paper has mentioned?

Hormold:
Hello! I'm curious, do you have any benchmark results from before and after? Like in https://github.com/livekit/agents/issues/3240#issuecomment-3486194037

Also, are you sure that sending "hi" as a message is a good idea? Isn't it enough to just send a simple GET request (i.e., a health check)? It will consume tokens without letting user know about that.

joshiayush:
@Hormold I'm using LiveKit in our products and I'm overriding the existing `LLM` class definition because the current version of `agents` doesn't support `prewarm` and here are the results.

Screenshot 2025-12-09 at 8 45 34 AM

I'm following this method:

> This is because LLMs usually require us to perform an inference request, instead of just opening a HTTP connection to /.

jp-paper:
@joshiayush I don't think this needs to be handled in this PR, and we could move this discussion elsewhere if it's something others would like to cover.

In our project, we create rooms with pre-defined content to be covered during the session. This content can be images and documents which I would like to explicitly cache as Gemini pre-tokenize them compared to implicit caching. I expect that it will reduce the payload transferred on every LLM request as the content will already exist on Gemini's end and improve the latency a bit.

Previously, we have been adding this content as a user message in the initial chat context instead of starting with the default empty chat context, but it does not feel right as it is not provided by the actual participant of the session, but the "platform" managing the sessions. Also caused by a limitation from Gemini to understand only "user" and "assistant" messages, and not "developer" like other models.

joshiayush:
@longcw Are you guys planning to add this support to agents?

joshiayush:
Not needed? Closing it then!

Issue #3973, opened 2025-11-18
JSON logs no longer include levelname, Datadog status mapping breaks


MonkeyLeeT:
### Bug Description

 - PR #3199 switched the CLI logging to a custom JsonFormatter (livekit-agents/livekit/agents/cli/log.py).
- That formatter only copies fields referenced in logging.PercentStyle.default_format, plus two hard-
    coded keys (level and name). It never emits the standard levelname.
- The formatter also drops levelname from extras via _RESERVED_ATTRS, so applications can’t re-add it
    in handlers.
- Result: production JSON logs now only contain level, not levelname. Datadog (and any tooling that
    expects the canonical Python field) can’t determine log status anymore.
  - As a workaround, we added ad hoc remapper logic to look at level field, but not in canonical way.

### Expected Behavior

levelname is still in output json record

### Reproduction Steps

```bash
Upgrade to 1.3.0+, then run agent with prod environment and json logger, and configure logging override as below

logging.PercentStyle.default_format = (
    "%(levelname)s %(message)s %(processName) %(threadName)s"
)

observe that levelname is not present in the json output


```

### Operating System

Linux

### Package Versions

```bash
livekit-agents=1.3.2
```

### Proposed Solution

```python
Always include the levelname attribute in JSON output (the previous dependency did this by default), or provide a supported hook to restore it across processes. Otherwise every consumer has to monkeypatch json formatter before app runs, which is brittle.
```

longcw:
does `levelname` exist in the json logs before agents 1.3? I tested with 1.2.18 and it's `"level": "INFO"`.

MonkeyLeeT:
> does `levelname` exist in the json logs before agents 1.3? I tested with 1.2.18 and it's `"level": "INFO"`.

It does, you need to override the default formatting as below to expose it in final json.

```
    logging.PercentStyle.default_format = (
        "%(levelname)s %(message)s %(processName) %(threadName)s"
    )
```

But this is not working now after 1.2.18.

MonkeyLeeT:
@longcw Any update here?

MonkeyLeeT:
@longcw A friendly ping here.

longcw:
the issue is that the setting of `logging.PercentStyle.default_format` is not used, fixed in https://github.com/livekit/agents/pull/4169

PR #4169, opened 2025-12-04
fix logging style format is not respected

longcw:
fix https://github.com/livekit/agents/issues/3973


chenghao-mou (approved):
LGTM

I can see `{"level name": "INFO", "message": "process initialized", "processName": "MainProcess", "threadName": "MainThread", "level": "INFO", "name": "livekit.agents", "pid": 39013, "elapsed_time": 0.64, "timestamp": "2025-12-04T09:32:02.546449+00:00"}` with the setting from the original issue.

PR #4170, opened 2025-12-04
Adding extra content to OpenAI LLM. Improving function call grouping.

russellmartin-livekit:


theomonnom:
I don't see extra_content from the ChoiceDelta:
https://github.com/openai/openai-python/blob/dc7602151b9042891ccd5042c7be3881337368e5/src/openai/types/chat/chat_completion_chunk.py#L59

russellmartin-livekit:
> I don't see extra_content from the ChoiceDelta: https://github.com/openai/openai-python/blob/dc7602151b9042891ccd5042c7be3881337368e5/src/openai/types/chat/chat_completion_chunk.py#L59

My testing shows it working

OpenAI SDK's Pydantic models use extra='allow' 1, which preserves any additional fields from the JSON response. We access it via getattr(delta, "extra_content", None). This allows providers to include additional metadata like thought signatures.

https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/llm/chat_context.py#L158

Context around google though signatures: https://ai.google.dev/gemini-api/docs/thought-signatures?authuser=1#openai

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -158,6 +158,8 @@ class ChatMessage(BaseModel):
     extra: dict[str, Any] = Field(default_factory=dict)
```
theomonnom, line 158: Should we use extra instead? 
And add a new key?

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -183,6 +185,12 @@ class FunctionCall(BaseModel):
     arguments: str
     name: str
     created_at: float = Field(default_factory=time.time)
+    extra_content: dict[str, Any] | None = None
```
theomonnom: Same here, I would use extra instead

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -52,12 +52,16 @@ class FunctionToolCall(BaseModel):
     name: str
     arguments: str
     call_id: str
+    extra_content: dict[str, Any] | None = None
```
theomonnom: I would still name this `extra` IMO
```suggestion
    extra: dict[str, Any] | None = None
```

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -52,12 +52,16 @@ class FunctionToolCall(BaseModel):
...
+    """Provider-specific extra content (e.g., Google thought signatures)."""
 
 
 class ChoiceDelta(BaseModel):
     role: ChatRole | None = None
     content: str | None = None
     tool_calls: list[FunctionToolCall] = Field(default_factory=list)
+    extra_content: dict[str, Any] | None = None
```
theomonnom: Same here
```suggestion
    extra: dict[str, Any] | None = None
```

theomonnom (approved):
Otherwise lgtm!

PR #4177, opened 2025-12-04
fix(aws): Handle nested schema in Nova Sonic tool parameter extraction

somoore:
## Summary
Fixes tool parameter extraction for Amazon Nova Sonic realtime models. Tool calls were receiving empty `{}` parameters due to incorrect schema access.

## Problem
`build_legacy_openai_schema()` returns a nested structure:
{'type': 'function', 'function': {'name': '...', 'parameters': {...}}}The existing code directly accessed `schema["parameters"]`, which returns `None` when the schema has the nested `function` key, causing all tool parameters to be empty.

## Solution
Safely extract parameters by checking for the nested `function` key:
schema = llm.utils.build_legacy_openai_schema(f, internally_tagged=True)
func_schema = schema.get("function", schema)
input_schema = func_schema.get("parameters", {"type": "object", "properties": {}})This handles both flat and nested schema formats with proper fallback.

## Testing
Tested with function tools using the LiveKit Agents framework. Tools now receive correct parameter schemas instead of empty objects.

## Checklist
- [x] Ran `ruff check --fix` - all checks passed
- [x] Ran `ruff format` - code formatted
- [x] Handles both flat and nested schema structures
- [x] Includes fallback for raw function tools

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
@@ -448,15 +448,9 @@ class RealtimeSession(  # noqa: F811
             for name, f in self.tools.function_tools.items():
                 if llm.tool_context.is_function_tool(f):
                     description = llm.tool_context.get_function_info(f).description
-                    # Robust schema extraction for MCP tool compatibility:
-                    # - internally_tagged=True returns flat: {"name", "parameters", "type"}
-                    # - Other sources may use nested: {"type": "function", "function": {...}}
-                    # Handle both formats safely with fallback
-                    schema = llm.utils.build_legacy_openai_schema(f, internally_tagged=True)
-                    func_schema = schema.get("function", schema)
-                    input_schema = func_schema.get(
-                        "parameters", {"type": "object", "properties": {}}
-                    )
+                    input_schema = llm.utils.build_legacy_openai_schema(f, internally_tagged=True)[
+                        "parameters"
+                    ]
                 elif llm.tool_context.is_raw_function_tool(f):
                     description = llm.tool_context.get_raw_function_info(f).raw_schema.get(
                         "description"


livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
```diff
@@ -448,16 +448,24 @@ def _serialize_tool_config(self) -> ToolConfiguration | None:
             for name, f in self.tools.function_tools.items():
                 if llm.tool_context.is_function_tool(f):
                     description = llm.tool_context.get_function_info(f).description
-                    input_schema = llm.utils.build_legacy_openai_schema(f, internally_tagged=True)[
-                        "parameters"
-                    ]
+                    # Nova Sonic tool schema fix: build_legacy_openai_schema() returns nested structure
+                    # {'type': 'function', 'function': {'name': '...', 'parameters': {...}}}
+                    # Must safely extract parameters from nested 'function' key
```
longcw: when `internally_tagged` is True, the `build_legacy_openai_schema` returns parameters without nested function
```python
    if internally_tagged:
        return {
            "name": info.name,
            "description": info.description or "",
            "parameters": schema,
            "type": "function",
        }
```

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
```diff
@@ -448,16 +448,25 @@ def _serialize_tool_config(self) -> ToolConfiguration | None:
...
                     description = llm.tool_context.get_function_info(f).description
-                    input_schema = llm.utils.build_legacy_openai_schema(f, internally_tagged=True)[
-                        "parameters"
-                    ]
+                    # Robust schema extraction for MCP tool compatibility:
+                    # - internally_tagged=True returns flat: {"name", "parameters", "type"}
+                    # - Other sources may use nested: {"type": "function", "function": {...}}
+                    # Handle both formats safely with fallback
+                    schema = llm.utils.build_legacy_openai_schema(f, internally_tagged=True)
+                    func_schema = schema.get("function", schema)
```
longcw: sorry but I didn't see how `func_schema = schema.get("function", schema)` can handle both `{"name", "parameters", "type"}` and `{"type": "function", "function": {...}}`. when `internally_tagged=True`, it will always the first case.

MCP tools always create raw function tools, I guess what you actually want to fix is https://github.com/livekit/agents/pull/4177/files#diff-062465eace0742ba6f8f049c2b07a2aa892a5a6363ddb6299eb0c358c239cd98L458-R469 when the mcp tools didn't provide a parameters in the `raw_schema`.

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
somoore: Thanks @longcw and @kachenjr!

@longcw you’re correct, `internally_tagged=True` returns a flat structure today. The confusion in my original description came from debugging MCP tools, which surface schemas differently (e.g. via input_schema), and from thinking about the non‑internally‑tagged case that does use the nested `{"type": "function", "function": {...}} `shape.

The goal of this PR is to make the Nova Sonic plugin robust enough to handle both the current flat OpenAI‑style shape and the shapes used by MCP tools, which currently fail with the existing implementation and result in `{}` tool parameters.

I’ve updated the code comments to reflect this more accurately: we’re not claiming that `build_legacy_openai_schema(..., internally_tagged=True)` returns nested, just that the extraction logic is defensive enough to handle both flat and nested formats.

This robustness is what makes Nova Sonic tool calls work correctly with MCP tools in practice.

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
somoore: I don't have bandwidth to iterate further on this. 
Nova Sonic on livekit + MCP is broken. full stop. I've fixed it locally for my use case, feel free to use this PR as reference or close.

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
```diff
@@ -448,16 +448,25 @@ def _serialize_tool_config(self) -> ToolConfiguration | None:
...
+                    input_schema = func_schema.get(
+                        "parameters", {"type": "object", "properties": {}}
+                    )
```
longcw:
```suggestion
                    input_schema = llm.utils.build_legacy_openai_schema(f, internally_tagged=True)[
                        "parameters"
                    ]
```

I mean for MCP tool it's the raw function tool, only the changes in the [line 464](https://github.com/livekit/agents/pull/4177/files#diff-062465eace0742ba6f8f049c2b07a2aa892a5a6363ddb6299eb0c358c239cd98L458-R469) are needed, the changes here break the regular function tool.

longcw: I'll revert the changes on regular function tool as `llm.utils.build_legacy_openai_schema(f, internally_tagged=True)[
                        "parameters"
                    ]`  is always valid, and apply the fix for raw function tool. thanks for the pr!

longcw (approved):

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
somoore: that's great! thanks for working with me on this @longcw @kachenjr

PR #4180, opened 2025-12-05
fix _on_reject when no answer

longcw:


livekit-agents/livekit/agents/worker.py
```diff
@@ -1226,7 +1226,7 @@ async def _job_request_task() -> None:
                     "no answer was given inside the job_request_fnc, automatically rejecting the job",  # noqa: E501
                     extra={"job_request": job_req, "agent_name": self._agent_name},
                 )
-                await _on_reject()
+                await _on_reject(terminate=True)
```
longcw: should we terminate the job when no answer?

theomonnom: No I don't think we should

longcw:
```suggestion
                await _on_reject(terminate=False)
```

longcw: ok updated to False...

theomonnom (approved):

PR #4184, opened 2025-12-05
fix: resolve audio dropout after AgentTask completion

tomc98:
Fixed critical bugs causing agent to go silent after nested workflow agents (AgentTasks) complete:

1. Audio recognition not recreated on resume:
   - When activity is paused for AgentTask handover, _close_session() closed audio_recognition but didn't set it to None
   - On resume, _start_session() skipped recreation because the check `if self._audio_recognition is None` was False
   - Fix: Set audio_recognition, rt_session, and realtime_spans to None after closing them in _close_session()

2. Atomic handover with prestart pattern:
   - Added prestart() to create audio_recognition before activity swap
   - Added complete_start() to finish startup after swap
   - Reordered _update_activity() to prestart new activity before pausing old one, eliminating ~100-200ms audio gap

3. Transcript buffering during handover:
   - Buffer transcripts received during prestart mode
   - Replay them after complete_start() finishes

4. Additional race condition fixes:
   - Fixed generation futures not marked done on interruption
   - Fixed async generator cleanup race conditions
   - Added auth_lock for authorization operations on SpeechHandle

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1789,9 +1789,13 @@ class AgentActivity(RecognitionHooks):
 
         if speech_handle.interrupted:
             current_span.set_attribute(trace_types.ATTR_SPEECH_INTERRUPTED, True)
-            # Mark generation done to prevent scheduling task from waiting forever
-            if speech_handle._generations:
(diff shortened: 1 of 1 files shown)


Hormold:
Could you please share a minimal PoC agent to reproduce this issue? Thanks!

PR #4185, opened 2025-12-05
fix record.exc_info is not pickable when using LogQueueHandler

longcw:
fix issue introduced in https://github.com/livekit/agents/pull/4128, `record.exc_info[0]` or `record.exc_info[1]` can be unpickeable in some cases.


chenghao-mou (approved):
LGTM.

Issue #4190, opened 2025-12-06
Properly handle Gemini thought summaries and export them via OpenTelemetry in Google LLM plugin


giovaborgogno:
### Bug Description

Gemini’s API provides thought summaries when include_thoughts=true is enabled. These appear in content.parts with the part.thought flag set and must be handled separately from normal answer text. https://ai.google.dev/gemini-api/docs/thinking#summaries

Right now, the Google LLM plugin in `agents/livekit-plugins/livekit-plugins-google/livekit/plugins/google
/llm.py` ignores part.thought entirely. The parsing code only looks at part.function_call and part.text:

```python
def _parse_part(self, id: str, part: types.Part) -> llm.ChatChunk | None:
    if part.function_call:
        chat_chunk = llm.ChatChunk(
            id=id,
            delta=llm.ChoiceDelta(
                role="assistant",
                tool_calls=[
                    llm.FunctionToolCall(
                        arguments=json.dumps(part.function_call.args),
                        name=part.function_call.name,
                        call_id=part.function_call.id or utils.shortuuid("function_call_"),
                    )
                ],
                content=part.text,
            ),
        )
        return chat_chunk

    return llm.ChatChunk(
        id=id,
        delta=llm.ChoiceDelta(content=part.text, role="assistant"),
    )
```

There’s no check for part.thought, so thought summaries are treated as regular assistant output.

### Problems

1. Bug: Thoughts are spoken by TTS
When include_thoughts=True, thought summaries are merged into the same content used for user-facing responses. The TTS layer receives them and reads the agent’s internal reasoning out loud, which is not what Gemini’s “thinking” feature is meant for.

2. Missing observability of thoughts in OTEL
LiveKit already uses OpenTelemetry, but the Gemini thought summaries are not surfaced there at all. There is no way to inspect the model’s internal reasoning in traces/logs while keeping it hidden from the end user and TTS.

### Expected Behavior

**Separation of thoughts vs. answer**
Parts with part.thought == True should not be included in the assistant’s user-visible content.
Thought parts should never be forwarded to TTS or any channel that is meant for end-user output.

**Observability via existing OpenTelemetry**
Thought summaries should be attached to the existing OTEL spans/traces for Gemini calls

### Reproduction Steps

```bash
e.g.:


session = AgentSession(
        stt="assemblyai/universal-streaming-multilingual",
        llm=google.LLM(
            model="gemini-2.5-flash-preview-09-2025",
            temperature=0.8,
[...]

PR #4191, opened 2025-12-06
Add support for thought processing in Google LLM (#4190)

giovaborgogno:
Related with https://github.com/livekit/agents/issues/4190

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
@@ -328,7 +328,10 @@ class LLM(llm.LLM):
                     else:
                         _level = "low"
                 # Use thinking_level only (pass as dict since SDK may not have this field yet)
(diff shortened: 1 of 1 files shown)


theomonnom:
I think https://github.com/livekit/agents/pull/4170 is a better solution. So we keep the thought tokens in a separate field

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -449,6 +449,13 @@ def _parse_part(self, id: str, part: types.Part) -> llm.ChatChunk | None:
             )
             return chat_chunk
 
+        if part.thought:
+            content = f"<think>\n{part.text}\n</think>"
```
davidzhao: we need to store these in a separate field, and not mixing them with regular content.

giovaborgogno:
> I think #4170 is a better solution. So we keep the thought tokens in a separate field

I synced my branch with those changes to use the new extra field in `ChoiceDelta`

giovaborgogno:
@theomonnom @davidzhao

PR #4192, opened 2025-12-08
OpenAI Responses API Plugin

tinalenguyen:
To use:
`llm=openai.responses.LLM()` 

The tools included:
`openai.tools.WebSearch()`
`openai.tools.FileSearch()`
`openai.tools.CodeInterpreter()`

### DIFF
--- a/livekit-agents/livekit/agents/llm/tool_context.py
+++ b/livekit-agents/livekit/agents/llm/tool_context.py
@@ -381,12 +381,7 @@ class ToolContext:
 
     @overload
     def parse_function_tools(
-        self, format: Literal["openai"], *, strict: bool = True
-    ) -> list[dict[str, Any]]: ...
-
-    @overload
-    def parse_function_tools(
-        self, format: Literal["openai.responses"], *, strict: bool = True
+        self, format: Literal["openai", "openai.responses"], *, strict: bool = True
     ) -> list[dict[str, Any]]: ...
 
     @overload


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -0,0 +1,292 @@
...
+                        chunk = self._handle_output_items_done(event)
+
+                    if isinstance(event, ResponseTextDeltaEvent):
+                        chunk = self._handle_response_output_text_delta(event)
+
+                    if isinstance(event, ResponseCompletedEvent):
+                        chunk = self._handle_response_completed(event)
+
+                    if chunk is not None:
+                        self._event_ch.send_nowait(chunk)
```
toubatbrian, line 217: Should we set verbosity to 'low' or None and disable thinking by default? This mainly applies when working with the GPT-5 family models. Otherwise, the time to first text token tends to be quite slow, which can make it feel like the LLM request is hanging.

toubatbrian: The alternative would be streaming reasoning token as llm chunks (user can set weather to let LLM "speak out" reasoning token with a flag), or we do stream thinking chunk inside LLM plugin but make it only sends as part of transcripts but not pass into TTS. cc @theomonnom let me know your thoughts

livekit-agents/livekit/agents/llm/_provider_format/openai.py
```diff
@@ -206,3 +208,21 @@ def to_fnc_ctx(tool_ctx: llm.ToolContext, *, strict: bool = True) -> list[dict[s
...
+    for tool in tool_ctx.function_tools.values():
+        if isinstance(tool, llm.RawFunctionTool):
+            schema = tool.info.raw_schema
+            schema["type"] = "function"
+            schemas.append(schema)
+        elif isinstance(tool, llm.FunctionTool):
+            schema = llm.utils.build_legacy_openai_schema(tool, internally_tagged=True)
+            schemas.append(schema)  # type: ignore
+        elif isinstance(tool, llm.ProviderTool):
+            schemas.append(tool.to_dict())
```
theomonnom: We should manually check every ProviderTool individually here (FileSearch, ...).
llm.ProviderTool doesn't have a to_dict() method.

tinalenguyen: just updated, checking `openai.tools.OpenAITool` works

rimiakter5612-arch (approved):

theomonnom (approved):

davidzhao:
woohoo! nice work @tinalenguyen

PR #4194, opened 2025-12-08
fix watchfiles prevent agent prcoess exit on SIGTERM

longcw:
in dev mode with `reload` enabled, watchfiles will raise a `KeyboardInterrupt` on SIGTERM, but not sure why the process doesn't exit after `KeyboardInterrupt` exception, added a force exiting.


chenghao-mou:
Interestingly, I couldn't reproduce the issue with dev mode, even after changing files multiple times with reload. Any tips?

longcw:
@chenghao-mou it happens if you send a SIGTERM to the process when reload is enabled. are using using iterm2 on mac? if so the reload is disabled, otherwise when you send SIGTERM to the process there should be a log 
>     17:29:21.415 WARNI… watchfiles.main    received signal 15, raising KeyboardInterrupt 

but the process never exit after it

PR #4196, opened 2025-12-08
fix(llm): sanitize control characters in function call JSON arguments

ArpitKotecha:
### Problem

LLMs sometimes generate function call JSON with literal control characters (e.g., newlines, tabs) inside string values. For example:

```json
{"prompt": "A timeline showing:
- Event 1
- Event 2"}
```

The literal newline violates the JSON spec, causing `pydantic_core.from_json()` to fail with:
```
ValueError: control character (\u0000-\u001F) found while parsing a string
```

This breaks function tool execution when the LLM outputs multi-line content in tool arguments.

### Solution

Add a `_sanitize_json_control_chars()` helper that escapes control characters within JSON string values before parsing:
- `\n` → `\\n`
- `\r` → `\\r`  
- `\t` → `\\t`
- Other control chars → `\\uXXXX`

The function preserves already-escaped sequences and only modifies content inside JSON string values.

### Changes

- Added `_sanitize_json_control_chars()` helper function in `utils.py`
- Modified `prepare_function_arguments()` to sanitize JSON before calling `from_json()`

### Testing

Tested with real-world LLM output containing multi-line prompts that previously caused the error.


livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -38,6 +45,51 @@
...
+    LLMs sometimes generate JSON with literal control characters (e.g., newlines)
+    inside string values, which violates the JSON spec and causes parsing to fail
+    with: ValueError: control character (\\u0000-\\u001F) found while parsing a string
+
+    This function escapes those control characters before parsing.
+    """
+    if not json_str:
+        return json_str
+
+    def escape_control_chars_in_string(match: re.Match[str]) -> str:
```
davidzhao: could you add some tests for the parsing function? it's important to ensure that it's handling the wide variety of valid JSON inputs without breaking them.

also, isn't it better to run a regex replacement to strip them out?

ArpitKotecha: Good call on both counts. Pushed an update:

- Simplified the implementation — it's now two regex passes (find JSON string spans, then re.sub control chars within each span). No more manual character loop.
- I kept it as escape rather than strip so multi-line content in tool args is preserved (e.g. an LLM passing a multi-line `prompt` keeps its line breaks). Stripping would silently lose data. Happy to switch if you'd rather drop them.
- Added `TestSanitizeJsonControlChars` in `tests/test_tools.py` covering: valid JSON left untouched (parametrized over escapes, unicode, nested, arrays, primitives, empty), literal `\n`/`\r`/`\t` escaped, other control chars (`\x01`, `\x1f`) escaped as `\uXXXX`, already-escaped sequences preserved, escaped quotes mixed with literal newlines, whitespace outside strings left alone, and an end-to-end `prepare_function_arguments` test with a multi-line prompt.

Also rebased on latest main.

ArpitKotecha: @davidzhao let me know if anything else is required on this

Issue #3606, opened 2025-10-08
Sporadically seeing "Error in _play_task" from Background Audio player


danielgrittner:
Hey, we see the following error every now and then:

```
2025-10-08 15:30:30,524 - ERROR livekit.agents - Error in _play_task
Traceback (most recent call last):
  File "/.venv/lib/python3.12/site-packages/livekit/agents/utils/log.py", line 16, in async_fn_logs
    return await fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.12/site-packages/livekit/agents/voice/background_audio.py", line 367, in _play_task
    await gen.aclose()
RuntimeError: aclose(): asynchronous generator is already running

2025-10-08 15:30:30,526 - ERROR asyncio - Task exception was never retrieved
future: <Task finished name='Task-228' coro=<BackgroundAudioPlayer._play_task() done, defined at /.venv/lib/python3.12/site-packages/livekit/agents/utils/log.py:13> exception=RuntimeError('aclose(): asynchronous generator is already running')>
Traceback (most recent call last):
  File "/.venv/lib/python3.12/site-packages/livekit/agents/utils/log.py", line 16, in async_fn_logs
    return await fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.12/site-packages/livekit/agents/voice/background_audio.py", line 367, in _play_task
    await gen.aclose()
RuntimeError: aclose(): asynchronous generator is already running
```

We are on the newest version `1.2.14`.

Unfortunately, I don't know how to reproduce this. I also found some old issues with the same problem stating it is supposed to be fixed. However, for us this is not the case.
I am observing this issue locally as well as in production.

longcw:
previous attempt for fixing this issue https://github.com/livekit/agents/pull/2814

I can still reproduce it when stop the PlayHandle very fast cc @theomonnom 

It seems to be the race condition between `await gen.aclose()` and `stream.__anext__()` in audio mixer. I tried to ignore the RuntimeError of `gen.aclose` and add a break to the `_gen_wrapper` when play handle stopped, but the job process cannot fully exit if it breaks from the `_gen_wrapper` somehow...


```python
import asyncio
import logging

from dotenv import load_dotenv

from livekit.agents import (
    AudioConfig,
    BackgroundAudioPlayer,
    BuiltinAudioClip,
    JobContext,
[...]

PR #4197, opened 2025-12-08
fix race condition when stop background audio play handle

longcw:
fix https://github.com/livekit/agents/issues/3606, the issue was caused by a race condition when [`gen.aclose()`](https://github.com/livekit/agents/blob/livekit-agents@1.3.6/livekit-agents/livekit/agents/voice/background_audio.py#L372) is called during [`stream.__anext__()`](https://github.com/livekit/python-sdks/blob/rtc-v1.0.20/livekit-rtc/livekit/rtc/audio_mixer.py#L179).

~needs https://github.com/livekit/python-sdks/pull/537~ used a simpler solution that ignore the RuntimeError from `gen.aclose` and add a stopped flag

### DIFF
--- a/livekit-agents/livekit/agents/voice/background_audio.py
+++ b/livekit-agents/livekit/agents/voice/background_audio.py
@@ -377,9 +377,10 @@ class BackgroundAudioPlayer:
 
             await asyncio.sleep(0)
             if play_handle._stop_fut.done():
-                with contextlib.suppress(RuntimeError):
-                    await gen.aclose()
                 stopped = True
+                with contextlib.suppress(RuntimeError):
+                    # ignore error caused by race condition between aclose() and gen.__anext__()
+                    await gen.aclose()
 
     @log_exceptions(logger=logger)
     async def _run_mixer_task(self) -> None:


livekit-agents/livekit/agents/voice/background_audio.py
```diff
@@ -372,7 +377,9 @@ async def _gen_wrapper() -> AsyncGenerator[rtc.AudioFrame, None]:
 
             await asyncio.sleep(0)
             if play_handle._stop_fut.done():
-                await gen.aclose()
+                with contextlib.suppress(RuntimeError):
+                    await gen.aclose()
+                stopped = True
```
chenghao-mou: Nitpicking: should this be set before the `aclose` call?

livekit-agents/livekit/agents/voice/background_audio.py
```diff
@@ -344,8 +344,13 @@ async def _play_task(
             else:
                 sound = audio_frames_from_file(sound)
 
+        stopped = False
```
theomonnom, line 347: Do we even need the stopped flag if we just ignore the RuntimError?

longcw: Is that okay to just ignore the runtime error, the generator is still not closed? I am wondering should we also close the generator from audio file?

PR #4208, opened 2025-12-09
feat(google-tts): add prompt to normal synthesize for Gemini TTS

NXV5111:
## Summary

Add prompt support for Gemini TTS normal synthesize in ChunkedStream.

## Changes

- Add document reference for supported voices in Gemini TTS models.
- Include prompt to tts_input in normal synthesize.


davidzhao (approved):
lgtm

PR #4211, opened 2025-12-09
ensure playback_segments_count is consistent in the audio output chain

longcw:
when `next_in_chain` is set in `AudioOutput`, `await self.next_in_chain.capture_frame(frame)` should be called before `await super().capture_frame(frame)` (in case the capture_frame of next_in_chain is waiting for some conditions) to make sure the audio outputs in the chain have the same playback_segments_count.

this fixes the issue when the agent has a greeting message using session.say or session.generate_reply in Agent.on_enter (or just after session started), it will stuck when closing if no participant joined the room.


chenghao-mou (approved):
LGTM!

PR #4214, opened 2025-12-09
fix blocked send task in liveavatar plugin

tinalenguyen:


livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/avatar.py
```diff
@@ -126,6 +126,10 @@ def on_agent_state_changed(ev):
             if ev.new_state == "idle":
                 self.send_event({"type": "agent.stop_listening", "event_id": str(uuid.uuid4())})
 
+        @self._agent_session.on("close")
+        def on_agent_session_close(ev):
+            self._msg_ch.close()
```
chenghao-mou, line 131: I think we can close it at the end of `_send_task` or the final `finally:` block.

chenghao-mou: After reading the issue, I think the change makes sense.

galiwohlstadter:
Hello,
The fix in PR #4214 is present in 1.4.6 but the race condition still occurs.                                                                                                                                                                             
                                                                                                                                                                                                                                                             
 The on_agent_session_close handler closes _msg_ch, but _recv_task can still see the WebSocket close before _send_task sets closing = True. The error APIConnectionError: LiveAvatar connection closed unexpectedly still fires on every normal session  end.
                                                                                                                                                                                                                                                             The root cause is in _main_task: _send_task sets closing = True only after iterating through _msg_ch and before calling ws_conn.close(), but _recv_task checks closing when it receives the WebSocket close message — and that check can lose the race.  
   
Environment: livekit-plugins-liveavatar 1.4.6, Python 3.11, macOS

PR #4218, opened 2025-12-10
add interruption timeout to SpeechHandle

longcw:
add interruption timeout to SpeechHandle, cancel the tasks of the speech and mark it as done if it's not done before the timeout, to avoid any potential deadlock that waiting for the speech playout.

### DIFF
--- a/livekit-agents/livekit/agents/voice/speech_handle.py
+++ b/livekit-agents/livekit/agents/voice/speech_handle.py
@@ -187,8 +187,6 @@ class SpeechHandle:
                 )
                 for task in self._tasks:
                     task.cancel()
-                if self._generations:
-                    self._mark_generation_done()
                 self._mark_done()
 
             self._interrupt_timeout_handle = asyncio.get_event_loop().call_later(


livekit-agents/livekit/agents/voice/speech_handle.py
```diff
@@ -172,9 +177,24 @@ def _cancel(self) -> SpeechHandle:
...
 
+            def _on_timeout() -> None:
+                logger.error(
+                    "speech not done in time after interruption, cancelling the speech arbitrarily.",
+                    extra={"speech_id": self._id, "timeout": INTERRUPTION_TIMEOUT},
+                )
+                for task in self._tasks:
+                    task.cancel()
+                if self._generations:
+                    self._mark_generation_done()
```
chenghao-mou: nitpicking: we don't need these two lines, right? They are also called in `_mark_done`.

livekit-agents/livekit/agents/voice/speech_handle.py
```diff
@@ -172,9 +177,24 @@ def _cancel(self) -> SpeechHandle:
         if self.done():
             return self
 
-        with contextlib.suppress(asyncio.InvalidStateError):
+        if not self._interrupt_fut.done():
             self._interrupt_fut.set_result(None)
 
+            def _on_timeout() -> None:
+                logger.error(
```
chenghao-mou, line 184: Do we need to check the `done` state again here?

longcw: both `_mark_done` and the timeout handle are sync method, the `_on_timeout` is only called when `_mark_done` isn't called.

PR #4220, opened 2025-12-10
feat: add Cohere plugin for LiveKit Agents

darshankparmar:


livekit-plugins/livekit-plugins-cohere/livekit/plugins/cohere/models.py
```diff
@@ -0,0 +1,23 @@
...
+    "command-r-plus-08-2024",
+    "command-r-08-2024",
+    "command-r",
+    "command",
+    "command-nightly",
+    "command-light",
+    "command-light-nightly",
+]
+
+EmbeddingModels = Literal[
```
Hormold: I don't think that embedding models can be used in voice calls

Hormold:
Hello, and thank you for your PR!
In most cases, if the API is quite similar to OpenAI's, we push new providers as additional function to OpenAI (e.g., recent PRs regarding OpenRouter or OVH) plugin. Could you update it to just be part of openai plugin?

darshankparmar:
> Hello, and thank you for your PR! In most cases, if the API is quite similar to OpenAI's, we push new providers as additional function to OpenAI (e.g., recent PRs regarding OpenRouter or OVH) plugin. Could you update it to just be part of openai plugin?

Hi @Hormold thanks for your message! I will. 😄

darshankparmar:
> Hello, and thank you for your PR! In most cases, if the API is quite similar to OpenAI's, we push new providers as additional function to OpenAI (e.g., recent PRs regarding OpenRouter or OVH) plugin. Could you update it to just be part of openai plugin?

Hi @Hormold ,

Thank you for the feedback! I started implementing the Cohere integration as part of the OpenAI plugin following the OpenRouter pattern, but I discovered an important technical issue:

**Cohere doesn't currently provide OpenAI-compatible API endpoints.**

During testing, I get `405 Method Not Allowed` errors when trying to use Cohere's API with OpenAI's client format. This is different from providers like OpenRouter, DeepSeek, and others that specifically provide OpenAI-compatible endpoints.

The separate plugin I originally created works perfectly with Cohere's native API, but integrating into the OpenAI plugin would require Cohere to support OpenAI's API format first.

What would you prefer as the approach here?

Thanks for your guidance!

Hormold:
Hey! Tested the Cohere integration and found two issues that need fixing before this can work properly with voice agents.

First, Cohere API returns 400: message must be at least 1 token long when there's no user message in the chat context. This happens in voice agents when generate_reply(instructions="...") is called without user input (like in on_enter() to greet the user). OpenAI handles this fine but Cohere doesn't... looks like Cohere requires at least one user message to generate a response.

Second, tool calling breaks with 400: schema 'type' must be a string. Array 'type' is unsupported for this model. Cohere's OpenAI-compatible API seems to have stricter JSON schema requirements than OpenAI - it doesn't accept union/array types that LiveKit generates for function tools. Need to figure out what schema format Cohere actually expects and adapt the tool serialization.

darshankparmar:
Thanks for the feedback!

darshankparmar:
I took a look at the first issue. One idea: we could add a check in the chat method (Cohere-only) to see if there’s at least one user message in the context. If not, we either auto-inject a small placeholder user message (e.g. “Hello”) so Cohere is happy, or just throw an exception instead. Not sure which direction makes more sense here, but this would at least guarantee we don’t hit that 400 on empty contexts.

darshankparmar:
I dug into the second issue as well. Ended up fixing it by setting _strict_tool_schema=False for Cohere.

darshankparmar:
> I took a look at the first issue. One idea: we could add a check in the chat method (Cohere-only) to see if there’s at least one user message in the context. If not, we either auto-inject a small placeholder user message (e.g. “Hello”) so Cohere is happy, or just throw an exception instead. Not sure which direction makes more sense here, but this would at least guarantee we don’t hit that 400 on empty contexts.

@Hormold  can you confirm?

chenghao-mou:
Adding a placeholder message should be fine. This is what we did with Gemini Realtime:
Screenshot 2025-12-19 at 11 27 11

darshankparmar:
@Hormold pushed the Cohere issues fix, PR is ready for review.

Hormold:
Hey, I tested, and the PR looks good. One minor thing is a conflict here. Also, I encountered a couple of timeouts on Cohere responses.

Hormold:
Merge break imports. Could you please add ChatMessage to imports again?

darshankparmar:
> Hey, I tested, and the PR looks good. One minor thing is a conflict here. Also, I encountered a couple of timeouts on Cohere responses.

Thanks for testing! Regarding the timeouts - Cohere API can have high latency (25+ seconds TTFT) which may cause timeout errors in real-time applications. Consider:

1. For a small, fast model: `command-r7b-12-2024`
2. For general purpose use: `command-r-08-2024` 
3. For more advanced capabilities: `command-a-03-2025`
4. Increasing timeout values for production use
5. Setting appropriate max_completion_tokens to reduce response time

I also updated the latest Cohere Command models (text generation).

Hormold:
Thanks! Works great!

darshankparmar:
> Thanks! Works great!

Thanks for the feedback! Glad it's working well.
Any final changes needed or ready to merge? 🚀

carschandler:
Would be awesome to see this make it in! Command A+ is ranking super high on performance-to-latency ratio!

PR #4222, opened 2025-12-11
Fix #4188: Prevent LogQueueListener crash when logging RpcError

### LINKED ISSUE #4188: `LogQueueListener` crashes when logging `RpcError`s
### Bug Description

When performing an RPC from my agent to my client and I purposely close the connection, a RpcError will be raised at the agent side. **This is intended behaviour!** However, the newly introduced agent-insights logger cannot handle such errors:

```python
TypeError: RpcError.__init__() missing 1 required positional argument: 'message'
```

### Expected Behavior

Either: 
1. `RpcError` should be correctly pickled
2. or, ignore/drop this `LogRecord` such that the program does not crash

### Reproduction Steps

```python
import logging
from typing import override
import pickle
from livekit.rtc import RpcError

# This simulates the LogQueueListener receiving data
class LogHandler(logging.Handler):
    @override
    def emit(self, record: logging.LogRecord) -> None:
        # Data is send by some logging handler
        data = pickle.dumps(record)
        ...

        # The LogQueueListener will receive data and try to unpickle it
        pickle.loads(data)


# Setup
logging.getLogger().addHandler(LogHandler())

# This is fine...
logging.info("This is fine...")

# This is not fine...
logging.error(
    "This is crashes the program...",
    exc_info=RpcError(code=0, message="Test RPC error"),
)
```

This example will crash with the following error:
```python
TypeError: RpcError.__init__() missing 1 required positional argument: 'message'
```

### Operating System

Linux, macOS

### Models Used

_No response_

### Package Versions

```bash
livekit-agents==1.3.6
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
# This is a temporary fix!

# Inside the `LogQueueListener`, update `_monitor` method to:

def _monitor(self) -> None:
    while True:
        try:
            data = self._duplex.recv_bytes()
        except utils.aio.duplex_unix.DuplexClosed:
            break

        try:
            record = pickle.loads(data)
            if isinstance(record, logging.LogRecord):
                self.handle(record)
        except Exception:
            pass
```

### Additional Context

_No response_

_No response_

nithinyanna10:
- Add _safe_to_log() helper to convert RpcError objects to pickleable dicts
- Sanitize custom log record attributes before pickling
- Add regression tests to verify RpcError handling
- Prevents pickle.PicklingError when RpcError objects are in log extra fields

### DIFF
--- a/livekit-agents/livekit/agents/ipc/log_queue.py
+++ b/livekit-agents/livekit/agents/ipc/log_queue.py
@@ -14,7 +14,7 @@ from ..utils.aio import duplex_unix
 
 def _safe_to_log(obj: Any) -> Any:
     """Safely convert an object to a pickleable format.
-    
+
     Handles RpcError and other objects that may not be directly pickleable
     by converting them to a dictionary representation.
     """
@@ -138,7 +138,7 @@ class LogQueueHandler(logging.Handler):
                     "thread", "threadName", "exc_info", "exc_text", "stack_info",
                     "getMessage", "websocket"
                 }
-                
+
                 # Check custom attributes that might contain RpcError
                 for attr_name in dir(record):
                     if attr_name.startswith("_") or attr_name in standard_attrs:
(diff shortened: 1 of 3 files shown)


longcw:
it's already fixed in https://github.com/livekit/agents/pull/4185

PR #4223, opened 2025-12-11
disable interruptions for agent greeting

hiroshihorie:
Give iOS devices time to calibrate AEC

### DIFF
--- a/examples/voice_agents/basic_agent.py
+++ b/examples/voice_agents/basic_agent.py
@@ -39,7 +39,7 @@ class MyAgent(Agent):
     async def on_enter(self):
         # when the agent is added to the session, it'll generate a reply
         # according to its instructions
-        # Keep it uninterruptible so the client has time to calibrate AEC.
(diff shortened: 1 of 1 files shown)


davidzhao:
could you also update [basic_agent](https://github.com/livekit/agents/blob/main/examples/voice_agents/basic_agent.py#L42) and add a comment on why it should not allow interruption in the beginning

examples/voice_agents/basic_agent.py
```diff
@@ -39,7 +39,8 @@ def __init__(self) -> None:
     async def on_enter(self):
         # when the agent is added to the session, it'll generate a reply
         # according to its instructions
-        self.session.generate_reply()
+        # Keep it uninterruptible so the client has time to calibrate AEC.
```
davidzhao:
```suggestion
        # Keep it uninterruptible so the client has time to calibrate AEC (Acoustic Echo Cancellation).
```

davidzhao (approved):

PR #4228, opened 2025-12-11
AGT-2328: negative threshold in silero

chenghao-mou:
Allow speech to non-speech threshold in silero, same as the negative threshold in [silero](https://github.com/snakers4/silero-vad/blob/d5b52843f79c97810940534f3433e162470af83d/src/silero_vad/utils_vad.py#L272).


livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py
```diff
@@ -454,7 +458,9 @@ def _copy_speech_buffer() -> rtc.AudioFrame:
                     )
                 )
 
-                if p >= self._opts.activation_threshold:
+                if p >= self._opts.activation_threshold or (
+                    pub_speaking and p >= self._opts.deactivation_threshold
```
longcw: should we use max(0, activation_threshold - 0.15) and `pub_speaking and p > self._opts.deactivation_threshold`, otherwise when `deactivation_threshold` is smaller than 0 it will never be set to non speaking after speaking?

theomonnom (approved):
lgtm

PR #4232, opened 2025-12-11
fix(openai): migrate realtime STT to GA API

Hormold:
Migrates OpenAI Realtime STT from the deprecated Beta API to the GA API.

**Changes:**
- Remove deprecated `OpenAI-Beta: realtime=v1` header
- Update session config format from `transcription_session.update` to `session.update` with nested `audio.input` structure
- Add `RECOGNITION_USAGE` metrics emission for audio duration tracking (was missing in the original implementation)

## Context

OpenAI announced that the Realtime API Beta (`OpenAI-Beta: realtime=v1`) will be deprecated and removed on **February 27, 2026**. This PR ensures compatibility with the GA API.

Reference: https://platform.openai.com/docs/guides/realtime

## Tests

- [x] Tested with `use_realtime=True` in console mode
- [x] Verified transcription works correctly
- [x] Verified STT metrics are now emitted (audio_duration)
- [x] Tested with all config options: `language`, `prompt`, `noise_reduction_type`, `turn_detection`

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
@@ -362,9 +362,7 @@ class STT(stt.STT):
         }
 
         if self._opts.noise_reduction_type:
-            input_config["noise_reduction"] = {
-                "type": self._opts.noise_reduction_type
-            }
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
lgtm

Issue #3821, opened 2025-11-07
generate_reply timed out waiting for generation_created event in LiveKit Agents (Gemini Live 2.5 Flash with Vertex AI) randomly.


aankitroy:
### Bug Description

Description
When using Gemini Live 2.5 Flash models via Google Vertex AI with LiveKit Agents, the agent frequently throws the following error during response generation:

generate_reply timed out waiting for generation_created event
This occurs after calling generate_reply() on an active LiveKit LLM session, particularly when connected to a Vertex AI gemini-live-2.5-flash model through the google-vertexai provider.

### Expected Behavior

The LiveKit agent should receive the generation_created event from the Vertex AI Gemini model and begin streaming a response.

### Reproduction Steps

```bash
Steps to Reproduce
Configure a LiveKit Agent to use:
Provider: google-vertexai
Model: gemini-live-2.5-flash
Livekit Agent version: 1.2.17 
Start a standard session and attempt to generate replies multiple times.
Observe that:
In some runs, the agent streams replies normally.
In others, it hangs for several seconds and then times out with:
generate_reply timed out waiting for generation_created event
The error occurs even with stable network conditions and identical setup parameters.
```

### Operating System

Ubuntu 24

### Package Versions

```bash
LiveKit Agents version: 1.2.17

Google GenAI SDK version: google-genai >= 1.41.0

Python version: 3.13

Model: gemini-live-2.5-flash

Provider: google-vertexai
```

### Proposed Solution

```python

```

tinalenguyen:
Hi, I am unable to reproduce this error, could you provide a reproducible code sample? Does this usually happen within function tools?

edit: I was able to reproduce this error without Vertex AI, do you have configured access to the `gemini-live-2.5-flash` model? If not, then that may explain this error; the websocket connection is denied without proper access/permission for using the model

longcw:
should be fixed in #4237 and https://github.com/livekit/agents/pull/4273

parthroy:
any example of use generate_reply?

Issue #3560, opened 2025-10-02
generate_reply timed out waiting for generation_created event in LiveKit Agents


ydushyant3110-glitch:
I am encountering a timeout error in LiveKit Python Agents while calling generate_reply. The error message is:

text
livekit.agents.llm.realtime.RealtimeError: generate_reply timed out waiting for generation_created event.
This error occurs consistently even after implementing retry logic with asyncio.wait_for and increasing timeout duration.

Environment:

Python version: (e.g., 3.13.7)

LiveKit Agents version: (e.g., 2.22)

Operating System: (e.g., Windows 10 64-bit)

Network condition: (Stable/Behind Proxy/VPN, etc.)

Steps to Reproduce:

Use LiveKit AgentSession with Google RealtimeModel.

Start session and call generate_reply with instructions.

Timeout occurs waiting for generation_created event on backend.

What I have tried:

Increased timeout using asyncio.wait_for.

Added retry logic for generate_reply.

Verified network connectivity and no proxy issues.

Updated LiveKit Agents package to the latest version.

Checked backend API status.

Relevant Code Snippet:

python
async def entrypoint(ctx: agents.JobContext):
    session = AgentSession(preemptive_generation=True)
    current_ctx = session.history.items

    await session.start(
        room=ctx.room,
        agent=Assistant(chat_ctx=current_ctx),
        room_input_options=RoomInputOptions(
            noise_cancellation=noise_cancellation.BVC()
        ),
    )

    for _ in range(3):
        try:
            await asyncio.wait_for(
                session.generate_reply(instructions=Reply_prompts),
                timeout=30
            )
            break
        except asyncio.TimeoutError:
            await asyncio.sleep(5)

    conv_ctx = MemoryExtractor()
    await conv_ctx.run(current_ctx)
Request:
Kindly help identify the root cause of this timeout and suggest solutions or workarounds. Is there any known issue with the realtime reply generation backend or configuration needed?

abhinavsaxena2308:
#Hi Team,
I’m encountering the same issue where generate_reply times out waiting for the generation_created event when using a RealtimeModel. Even after increasing the timeout, adding retries, and ensuring the session is fully started, the event doesn’t seem to fire reliably.
Environment:
Python: 3.13.7
LiveKit Agents: 2.22
OS: Windows 10 64-bit
Network: Stable, no proxy/VPN


session = AgentSession(preemptive_generation=True)

await session.start(
    room=ctx.room,
[...]

Issue #4017, opened 2025-11-19
1.31.1 "generate_reply" error only with Gemini realtime


mercuryyy:
### Bug Description

Gemini realtime error on "generate_reply" 
Model = "gemini-2.5-flash-native-audio-preview-09-2025"

```
2025-11-19 01:36:41,358 - [5fe7fd69-9ad2-4f28] - iMash-Agent - agents - ERROR - Error in _realtime_reply_task
    01:36:41 ERROR  livekit.agents   Error in _realtime_reply_task                 {"pid": 8334, "job_id": "AJ_dbwMYUNR3t3z",    
                                     Traceback (most recent call last):            "room_id": "RM_U9HkfcjCGnPV"}                 
                                       File                                                                                      
                                     "/opt/homebrew/Caskroom/miniconda/base/lib/py                                               
                                     thon3.12/site-packages/livekit/agents/utils/l                                               
                                     og.py", line 16, in async_fn_logs                                                           
                                         return await fn(*args, **kwargs)                                                        
                                                ^^^^^^^^^^^^^^^^^^^^^^^^^                                                        
                                       File                                                                                      
                                     "/opt/homebrew/Caskroom/miniconda/base/lib/py                                               
                                     thon3.12/site-packages/livekit/agents/voice/a                                               
                                     gent_activity.py", line 2058, in                                                            
                                     _realtime_reply_task                                                                        
                                         generation_ev = await                                                                   
                                     self._rt_session.generate_reply(                                                            
                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^                                               
                                     ^^^^^^^^^^^^^                                                                               
                                     livekit.agents.llm.realtime.RealtimeError:                                                  
                                     generate_reply timed out waiting for                                                        
                                     generation_created event.                                                                   
2025-11-19 01:36:41,359 - [1630dee8] - iMash-Agent - agents - ERROR - Error in _realtime_reply_task
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.12/site-packages/livekit/agents/utils/log.py", line 16, in async_fn_logs
    return await fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.12/site-packages/livekit/agents/voice/agent_activity.py", line 2058, in _realtime_reply_task
    generation_ev = await self._rt_session.generate_reply(
[...]

PR #4237, opened 2025-12-12
fix generate_reply timeout for gemini

longcw:
fix https://github.com/livekit/agents/issues/3560, two cases:

- `generate_reply` may be interrupted by user speech or background noise, the interrupted response was ignored, causing a `received server content but no active generation.` warning and a timeout error.
- when only `user_input` is set for `generate_reply` and it's called right after `session.start`, the Gemini session may not created yet, so the user message is added to the chat_ctx and synced to the API during connection, in that case after https://github.com/livekit/agents/pull/3898, `turn_complete = True` is not sent to the API so no response created.

fix https://github.com/livekit/agents/issues/3821 and https://github.com/livekit/agents/issues/4017

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -589,8 +589,8 @@ class RealtimeSession(llm.RealtimeSession):
 
         # Gemini requires the last message to end with user's turn
         # so we need to add a placeholder user turn in order to trigger a new generation
+        turns = []
         if is_given(instructions):
-            turns = []
             turns.append(types.Content(parts=[types.Part(text=instructions)], role="model"))
         turns.append(types.Content(parts=[types.Part(text=".")], role="user"))
         self._send_client_event(types.LiveClientContent(turns=turns, turn_complete=True))


livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -411,11 +411,23 @@ async def _close_active_session(self) -> None:
...
                     self._active_session = None
 
-    def _mark_restart_needed(self) -> None:
+    def _mark_restart_needed(self, on_error: bool = False) -> None:
         if not self._session_should_close.is_set():
             self._session_should_close.set()
+
             # reset the msg_ch, do not send messages from previous session
-            self._msg_ch = utils.aio.Chan[ClientEvents]()
+            new_ch = utils.aio.Chan[ClientEvents]()
```
chenghao-mou: nitpicking: we don't need to declare a new variable since we don't use it. But fine to keep it if we want it with the commented `# new_ch.send_nowait(msg)`

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -411,11 +411,23 @@ async def _close_active_session(self) -> None:
...
+            if not on_error:
+                while not self._msg_ch.empty():
+                    msg = self._msg_ch.recv_nowait()
+                    if isinstance(msg, types.LiveClientContent) and msg.turn_complete is True:
+                        logger.warning(
+                            "discarding client content for turn completion, may cause generate_reply timeout",
+                            extra={"content": str(msg)},
+                        )
+                        # new_ch.send_nowait(msg)
```
chenghao-mou: QQ: is there a reason we don't want to forward this? 

I guess the `on_error` part means the input likely caused the issue and we don't want to repeat the error.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
longcw: removed it

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
longcw: I am not sure if that's related to the timeout error, adding a log for now so users can report if they see the issue again.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -415,10 +415,19 @@ async def _close_active_session(self) -> None:
...
 
-    def _mark_restart_needed(self) -> None:
+    def _mark_restart_needed(self, on_error: bool = False) -> None:
         if not self._session_should_close.is_set():
             self._session_should_close.set()
             # reset the msg_ch, do not send messages from previous session
+            if not on_error:
+                while not self._msg_ch.empty():
+                    msg = self._msg_ch.recv_nowait()
+                    if isinstance(msg, types.LiveClientContent) and msg.turn_complete is True:
```
davidzhao, line 425: so when the session is resumed, everything in chat context should be added again right? or maybe we should carry over user responses that have not yet been sent? wdyt?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -748,7 +757,7 @@ async def _send_task(self, session: AsyncSession) -> None:
                 if isinstance(msg, types.LiveClientContent):
                     await session.send_client_content(
                         turns=msg.turns,  # type: ignore
-                        turn_complete=msg.turn_complete or True,
+                        turn_complete=msg.turn_complete if msg.turn_complete is not None else True,
```
davidzhao, line 760: great catch!

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -583,8 +592,8 @@ def generate_reply(
         if is_given(instructions):
             turns = []
             turns.append(types.Content(parts=[types.Part(text=instructions)], role="model"))
-            turns.append(types.Content(parts=[types.Part(text=".")], role="user"))
-            self._send_client_event(types.LiveClientContent(turns=turns, turn_complete=True))
+        turns.append(types.Content(parts=[types.Part(text=".")], role="user"))
```
davidzhao, line 595: I think this might result in duplicate turns again? so when a text message is received, it will perform update_chat_context first, then call generate_reply with `instructions=None`.. in that case, we'd add the chat context but also the extra `.`?

longcw: I think the duplicate turns before is caused by `turn_complete=msg.turn_complete or True` that it didn't respect `turn_complete=False`, so both `update_chat_ctx` (with turn_complete=False) and `generate_reply` will create responses separately.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
longcw: > everything in chat context should be added again

the chat context will be synced when creating the connection. 

> maybe we should carry over user responses that have not yet been sent? 

the restart may be caused by `update_options, `update_tools` or `update_instructions`, ideally we shouldn't call these methods when a `generate_reply` is not done. I think it's fine to just add a warning in case they break the `generate_reply`, so that we know the reason when this happened.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
longcw: the duplicate turns should be fixed after `turn_complete` is fixed.

davidzhao: got it.. that makes sense.. we are always adding the message with turn_complete=False.. and counting on `generate_reply` to trigger generation.

davidzhao (approved):

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -583,8 +592,8 @@ def generate_reply(
         if is_given(instructions):
             turns = []
```
tinalenguyen:
```suggestion
        turns = []
        if is_given(instructions):
```
i get `UnboundLocalError: cannot access local variable 'turns' where it is not associated with a value` from line 595

PR #4248, opened 2025-12-14
feat(google): update default realtime model to gemini-2.5 12-2025

### LINKED ISSUE #4153: Client identification for Gemini API
### Feature Type

Nice to have

### Feature Description

Hey team, would it be possible to add the following client identification to the library to help us track usage origins better? 

`X-Goog-Api-Client`: API client identification. The value is a space-separated list of NAME "/" SEMVER strings, where the NAME should only contain lowercase letters, digits, and "-", and the SEMVER should be a semantic version string. For example: X-Goog-Api-Client: python/3.5.0 grpc-google-pubsub-v1/0.1.0-beta2 linux/2.7.0.

See system parameter docs here: https://docs.cloud.google.com/apis/docs/system-parameters 

```
Example: 
## Python:

from google import genai

client = genai.Client(
    api_key="...",
    http_options={
        "headers": {
            "x-goog-api-client": "my-sdk-py/1.23",
        }
    }
)

## JavaScript
import { GoogleGenAI } from '@google/genai';

const ai = new GoogleGenAI({
    apiKey: '...',
    httpOptions: {
        headers: {
            'x-goog-api-client': 'my-sdk-js/1.23'
        }
    }
});

## Direct API
curl https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash:generateContent?key=$GOOGLE_API_KEY \
    -H 'Content-Type: application/json'
    -H 'X-Goog-Api-Client: my-sdk/1.23'
    -d '{...}'
```

You can see an example implementation of that client identification in the SDK here: https://github.com/google/adk-python/blob/main/src/google/adk/models/google_llm.py#L262-L272

_No response_

### Additional Context

_No response_

davidzhao:
also fixed handling of reasoning output in transcription
added headers to identify client

closes #4153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10"]
+        python-version: ["3.9", "3.13"]
         
     steps:
       - uses: actions/checkout@v4
@@ -65,6 +65,7 @@ jobs:
 
       - name: Check Types
         run: uv run mypy --install-types --non-interactive
+              --untyped-calls-exclude=smithy_aws_core
               -p livekit.agents
               -p livekit.plugins.openai
               -p livekit.plugins.anthropic


longcw (approved):
lgtm

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
```diff
@@ -190,7 +190,7 @@ async def _run(self) -> None:
                 config_kwargs["aws_session_token"] = self._credentials.session_token
             else:
                 config_kwargs["aws_credentials_identity_resolver"] = (
-                    EnvironmentCredentialsResolver()
+                    EnvironmentCredentialsResolver(),
```
chenghao-mou, line 193: Sorry, is the one-value-to-a-tuple change intentional? @davidzhao

davidzhao: shoot.. it's not. great catch chenghao!

PR #4258, opened 2025-12-15
Add local dev commands for linking to rtc-sdk

lukasIO:
most importantly:

```
make help
Development Workflows:
  link-rtc             Link to local python-rtc (default FFI version)
  link-rtc-local       Build and link local rust SDK from source
  link-rtc-version     Install specific livekit version from PyPI (e.g., make link-rtc-version VERSION=0.18.0)
  unlink-rtc           Unlink local and restore PyPI version
  status               Show current linking status
  doctor               Check development environment health
```

### DIFF
--- a/makefile
+++ b/makefile
@@ -152,12 +152,22 @@ link-rtc-local: ## Build and link local rust SDK from source
 	cd $(RUST_SUBMODULE) && cargo build --release -p livekit-ffi; \
 	echo "$(CYAN)📝 Generating protobuf FFI protocol...$(RESET)"; \
 	cd $(PYTHON_RTC) && ./generate_proto.sh; \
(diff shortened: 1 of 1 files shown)


theomonnom:
we already have a makefile. should we remove it?

lukasIO:
@theomonnom should I instead add the local linking commands to the makefile?
justfile requires you to install just, which isn't great.
I don't have a good reason to be using justfile, it just felt like the most straight forward to deal with.

theomonnom:
> @theomonnom should I instead add the local linking commands to the makefile? justfile requires you to install just, which isn't great. I don't have a good reason to be using justfile, it just felt like the most straight forward to deal with.

I've never really used `just`, so I don't have strong opinions. My only feeling was that they both serve the same purpose.

lukasIO:
I've moved the helpers into the existing makefile

PR #4265, opened 2025-12-15
livekit-blingfire 1.1.0 & add python 3.14 support

theomonnom:


davidzhao (approved):

theomonnom:
@copilot add a test to livekit-blingfire for cibuildwheel, just importing the package is enough

from livekit import blingfire

PR #4274, opened 2025-12-16
Update Deepgram STT base URL to v2

relevance-worktrial-001:


davidzhao:
v1 and v2 APIs aren't the same.. you should use `deepgram.STTv2()` for the new API

PR #4288, opened 2025-12-17
fix(xai): a few more exports

davidzhao:


theomonnom (approved):
oh right, I forgot about the Plugin registry

Issue #4051, opened 2025-11-21
Pydantic Validation Issues for nullable tool fields


chasemcdo:
### Bug Description

When an agent tries to call a tool which has a nullable parameter but does not provide a value for said parameter it fails pydantic model validation.

Specifically the tools are defined as per the docs via:
```
from livekit.agents.llm import function_tool

async def tool_method(context: RunContext, non_nullable: str, nullable: str | None = None) -> Dict[str, Any]:
    # method cost
    return {"success": "ok"}

function_tool(tool_method, name="tool name", description="description")
``` 

### Expected Behavior

Not get a validation error and use default parameter

### Reproduction Steps

```bash
1. define tools as per the bug description
2. get agent to call tool without setting the nullable parameter
3. get validation error:

pydantic_core._pydantic_core.ValidationError: 1 validation error for ToolMethodArgs
nullable
  Field required [type=missing, input_value={'non_nullable': 'value'}, input_type=dict]

...
- Sample code snippet, or a GitHub Gist link -
```

### Operating System

MacOS 26.0.1 (25A362)

### Models Used

Deepgram Nova 3, Gemini 2.5 Flash, Cartesia Sonic 2

### Package Versions

```bash
livekit-agents >= 1.2.17
```

### Proposed Solution

```python

```

### Additional Context

This issue is presently blocking an update from LK version `1.2.14` and was originally observed on version `1.2.17` and continues to occur on `1.3.3`.

I've taken a glance through the source code and cannot identify any recent changes to this code related to this error, but this does not happen on the current version of `1.2.14` but as of `1.2.17` this happens very consistently.

longcw:
Can you share the full traceback or the example? I cannot reproduce it with the following tool call, also tried to remove the nullable var from the arguments manually in `llm_node`

```python
    @function_tool
    async def get_weather(self, location: str, date: str | None = None) -> str:
        """
        Called when the user asks about the weather.

        Args:
            location: The location to get the weather for
            date: The date to get the weather for, if user does not provide a date, ignore it.
        """
        logger.info(f"Getting weather for {location} for date {date}")
        return f"The weather in {location} is sunny {date or 'today'}."

    async def llm_node(self, chat_ctx: ChatContext, tools: list[FunctionTool], model_settings: ModelSettings):
        async for chunk in Agent.default.llm_node(self, chat_ctx, tools, model_settings):
            if isinstance(chunk, ChatChunk) and chunk.delta and chunk.delta.tool_calls:
[...]

PR #4290, opened 2025-12-17
fix dynamic FieldInfo for pydantic 2.12

longcw:
fix https://github.com/livekit/agents/issues/4051

changing the attr of FieldInfo dynamically is not supported since pydantic 2.12

ref: https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.FieldInfo
> The FieldInfo class is meant to expose information about a field in a Pydantic model or dataclass. FieldInfo instances shouldn't be instantiated directly, nor mutated.
>
> If you need to derive a new model from another one and are willing to alter FieldInfo instances, refer to this [dynamic model example](https://docs.pydantic.dev/2.12/examples/dynamic_models/).

### DIFF
--- a/livekit-agents/livekit/agents/llm/utils.py
+++ b/livekit-agents/livekit/agents/llm/utils.py
@@ -324,14 +324,23 @@ def function_arguments_to_pydantic_model(func: Callable[..., Any]) -> type[BaseM
             continue
 
         default_value = param.default if param.default is not param.empty else ...
+        field_info: FieldInfo | None = None
         field_attrs: dict[str, Any] = {}
 
         # Annotated[str, Field(description="...")]
         if get_origin(type_hint) is Annotated:
             annotated_args = get_args(type_hint)
             type_hint = annotated_args[0]
-            field_info = next((x for x in annotated_args[1:] if isinstance(x, FieldInfo)), Field())
-            field_attrs = field_info.asdict()["attributes"]
+            annotated_field = next(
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -324,14 +324,23 @@ def function_arguments_to_pydantic_model(func: Callable[..., Any]) -> type[BaseM
...
             annotated_args = get_args(type_hint)
             type_hint = annotated_args[0]
-            field_info = next((x for x in annotated_args[1:] if isinstance(x, FieldInfo)), Field())
-            field_attrs = field_info.asdict()["attributes"]
+            annotated_field = next(
+                (x for x in annotated_args[1:] if isinstance(x, FieldInfo)), None
+            )
+            if annotated_field and hasattr(annotated_field, "asdict"):
+                # `asdict` is available after pydantic 2.12
+                field_attrs = annotated_field.asdict()["attributes"]
```
chenghao-mou, line 339: QQ: should we use `field_info = annotated_field` here too?

longcw: no, it's not supported after pydantic 2.12 to modify fields like `field_info.default = None` so it has to be created from `Field(**attrs)`

theomonnom (approved):
lgtm! This is also the kind of things we should be able to quickly catch with our e2e testing project

PR #4298, opened 2025-12-17
Ensure makefile checks for livekit_lib_path

lukasIO:
Before the status and doctor commands would check for the presence of an `.env` file that wasn't guaranteed to be loaded when running a script. 
This change ensures the LIVEKIT_LIB_PATH is actually in the environment in order to report that a locally linked rust version is being used


chenghao-mou (approved):
lgtm

PR #4300, opened 2025-12-17
Allow for Cartesia TTS language auto-detection

yuyuma:
Currently the Cartesia TTS plugin sets default language to `en` and does not allow for language to be unset or set to `None`.

It's not 100% clear in the Cartesia docs, but they do allow for auto-detection of languages. Current behavior is causing undesirable accents for us in multi-lingual scenarios.
1. https://docs.cartesia.ai/api-reference/tts/websocket - note language is not a required field
2. Their playground specifically says to omit language for auto-detection
image

This PR allows for language to be set to `None` to enable auto-detection. I feel `None` should be the default, but left it as `en` for backward compatibility. The language option is set and passed through websockets as `None` rather than not set in case there is a need to `update_options` language setting between auto-detection vs constrained.


theomonnom (approved):
oh interesting, thanks!

PR #4306, opened 2025-12-18
re-export TurnDetection for xAI

tinalenguyen:
this would allow users to pass TurnDetection settings like so without importing from openai:

```
from livekit.plugins import xai

session = AgentSession(
    llm=xai.RealtimeModel(
        turn_detection=xai.TurnDetection(
            type="server_vad",
            threshold=0.5,
            prefix_padding_ms=300,
            silence_duration_ms=200,
            create_response=True,
            interrupt_response=True,
        )
    ),
)
```


theomonnom:
Sorry let's export it under `xai.realtime` instead

PR #4310, opened 2025-12-18
fix handoff to Realtime model with existing session context

davidzhao:
we didn't correctly filter out AgentHandoff items in the chat context

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1343,7 +1343,6 @@ class RealtimeSession(
         elif utils.is_given(audio_transcript):
             # sync the forwarded text to the remote chat ctx
             chat_ctx = self.chat_ctx.copy(
-                exclude_function_call=True,
                 exclude_handoff=True,
             )
             if (idx := chat_ctx.index_by_id(message_id)) is not None:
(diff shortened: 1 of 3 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1340,7 +1342,10 @@ def truncate(
             )
         elif utils.is_given(audio_transcript):
             # sync the forwarded text to the remote chat ctx
-            chat_ctx = self.chat_ctx.copy()
+            chat_ctx = self.chat_ctx.copy(
+                exclude_function_call=True,
```
longcw:
```suggestion
```
this will remove the function calls from the chat context?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model_beta.py
```diff
@@ -1170,7 +1171,10 @@ def truncate(
             )
         elif utils.is_given(audio_transcript):
             # sync the forwarded text to the remote chat ctx
-            chat_ctx = self.chat_ctx.copy()
+            chat_ctx = self.chat_ctx.copy(
+                exclude_function_call=True,
```
longcw:
```suggestion
```

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
davidzhao: yeah you are right we shouldn't need to do this, reverting

PR #4311, opened 2025-12-18
feat(gemini3) use low latency thinking_level by default for gemini 3 models

pushkar-nurix:
The default thinking_level is now "minimal" for Gemini 3 Flash models and "low" for other Gemini 3 models because [the API defaults to "high" when unspecified](https://ai.google.dev/gemini-api/docs/gemini-3)


davidzhao (approved):

davidzhao:
lgtm!

PR #4323, opened 2025-12-18
handle exceptions in task_results

tinalenguyen:
if `return_exceptions` is True, we add the error to the task results dict and continue. by default it is false and allows for users to handle custom exceptions

### DIFF
--- a/livekit-agents/livekit/agents/beta/workflows/task_group.py
+++ b/livekit-agents/livekit/agents/beta/workflows/task_group.py
@@ -44,7 +44,13 @@ class TaskGroup(AgentTask[TaskGroupResult]):
         return_exceptions: bool = False,
         chat_ctx: NotGivenOr[llm.ChatContext] = NOT_GIVEN,
     ):
-        """Creates a TaskGroup instance."""
+        """TaskGroup orchestrates a sequence of multiple AgentTasks. It also allows for users to regress to previous tasks if requested.
+
+        Args:
+            summarize_chat_ctx (bool): Whether or not to summarize the interactions within the TaskGroup into one message and merge the context. Defaults to True.
+            return_exceptions (bool): Whether or not to directly propagate an error. When set to True, the exception is added to the results dictionary and the sequence continues. Defaults to False.
+
+        """
         super().__init__(instructions="*empty*", chat_ctx=chat_ctx, llm=None)
 
         self._summarize_chat_ctx = summarize_chat_ctx


livekit-agents/livekit/agents/beta/workflows/task_group.py
```diff
@@ -84,8 +86,12 @@ async def on_enter(self) -> None:
                     task_stack.insert(0, task_id)
                 continue
             except Exception as e:
-                self.complete(e)
-                break
+                if self._return_exceptions:
```
miladmnasr, line 95: @tinalenguyen Very interesting.. just curious: what is the benefit of tracking the exception per task in the group? How does that affect the task group's function within the agent?

livekit-agents/livekit/agents/beta/workflows/task_group.py
```diff
@@ -41,12 +41,14 @@ def __init__(
         self,
         *,
         summarize_chat_ctx: bool = True,
+        return_exceptions: bool = False,
         chat_ctx: NotGivenOr[llm.ChatContext] = NOT_GIVEN,
     ):
         """Creates a TaskGroup instance."""
```
miladmnasr: `nit` I think `return_exceptions` warrants commentary regarding what it does and when it would be useful.

livekit-agents/livekit/agents/beta/workflows/task_group.py
tinalenguyen: This would allow for `TaskGroup` to continue with the sequence even if a task returns an exception and it can be logged and accessed in the results. I think this flexibility is beneficial for custom exceptions that may not be especially breaking but have significance (something like `ToolError`)

livekit-agents/livekit/agents/beta/workflows/task_group.py
tinalenguyen: Ah, I was going to update the docs site but I will update the docstring here as well. Thanks for reviewing and the feedback!

livekit-agents/livekit/agents/beta/workflows/task_group.py
miladmnasr: That makes a ton of sense, thanks for the explanation 🙏

livekit-agents/livekit/agents/beta/workflows/task_group.py
miladmnasr: Thank you for these changes and the quick turnaround!

PR #4328, opened 2025-12-19
stringify cartesia error to be pickleable

tinalenguyen:


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -523,7 +523,7 @@ async def _recv_task(ws: aiohttp.ClientWebSocketResponse, cartesia_context_id: s
         except Exception as e:
             logger.error(
```
theomonnom: Let's use logger.exception instead

PR #4332, opened 2025-12-19
Enable Soniox STT turn detection & metrics

### LINKED ISSUE #4034: Bug: Soniox STT Plugin Missing END_OF_SPEECH Events and RecognitionUsage Metrics
### Bug Description

# Bug Report: Soniox STT Plugin Missing END_OF_SPEECH Events and RecognitionUsage Metrics

## Description

The Soniox STT plugin has two critical bugs that prevent it from working properly with LiveKit's turn detection and usage tracking:

1. **Missing END_OF_SPEECH events**: The plugin receives the `<end>` token from Soniox but never emits `SpeechEventType.END_OF_SPEECH`, so LiveKit sessions can't close turns properly even though Soniox has finished the utterance.

2. **Missing RecognitionUsage metrics**: The plugin doesn't emit `RecognitionUsage` events, making it impossible to track audio duration for billing and analytics.

## Impact

- Applications cannot use `turn_detection_mode="stt"` with Soniox
- No way to track audio usage for billing/analytics
- Users miss out on Soniox's excellent semantic endpoint detection

## Environment

- Running in production on Breez voice agent platform
- Tested with LiveKit sessions using Soniox STT
- Wrapper code proves the fix works correctly

## Additional Context

This was identified in collaboration with Klemen (Soniox CEO) and Bojan from Soniox team. The fix follows the same pattern as other STT plugins (Deepgram) and is fully backwards compatible.

Happy to provide more details or submit a PR if helpful!

## Expected Behavior

1. When Soniox sends a FINAL_TRANSCRIPT (with `<end>` token), the plugin should emit a `SpeechEventType.END_OF_SPEECH` event
2. The plugin should track audio frame durations and periodically emit `RecognitionUsage` events (similar to Deepgram plugin)

### Reproduction Steps

```bash
## Reproduction

1. Create a LiveKit session with Soniox STT
2. Set `turn_detection_mode="stt"`
3. Observe that turns never close even though Soniox sends `<end>` token
4. Try to capture RecognitionUsage metrics - none are emitted
```

### Operating System

docker

### Models Used

soniox-tts

### Package Versions

```bash
livekit-plugins-soniox==1.3.1
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
## Proposed Fix

We've been running a wrapper in production that fixes both issues. The implementation is straightforward:

### Fix 1: Emit END_OF_SPEECH after FINAL_TRANSCRIPT


# In the event forwarding loop (_run method or similar)
async for event in self._inner_stream:
    self._event_ch.send_nowait(event)
    
    # Add this:
    if event.type == stt.SpeechEventType.FINAL_TRANSCRIPT:
        try:
            self._event_ch.send_nowait(
                stt.SpeechEvent(type=stt.SpeechEventType.END_OF_SPEECH)
            )
        except Exception as e:
            logger.warning(f"Failed to emit END_OF_SPEECH: {e}")


### Fix 2: Track and Emit RecognitionUsage Metrics


# Import
from livekit.plugins.deepgram._utils import PeriodicCollector

# In __init__
self._audio_duration_collector = PeriodicCollector(
    callback=self._on_audio_duration_report,
    duration=5.0,  # Report every 5 seconds
)

# Add callback method
def _on_audio_duration_report(self, duration: float) -> None:
    usage = stt.RecognitionUsage(audio_duration=duration)
    self._event_ch.send_nowait(
        stt.SpeechEvent(
            type=stt.SpeechEventType.RECOGNITION_USAGE,
            recognition_usage=usage
        )
    )

# In push_frame
def push_frame(self, frame: rtc.AudioFrame) -> None:
    # Track duration
    frame_duration = frame.samples_per_channel / frame.sample_rate
    self._audio_duration_collector.push(frame_duration)
    # ... rest of push_frame logic

# In aclose
    # Emit final metrics
    final_duration = self._audio_duration_collector.get_total()
    if final_duration > 0:
        self._on_audio_duration_report(final_duration)
    
    await self._audio_duration_collector.aclose()
    # ... rest of cleanup
```

### Additional Context

_No response_

_No response_

matejmarinko-soniox:
This PR addresses issues preventing the Soniox STT plugin from working correctly with LiveKit's `turn_detection_mode="stt"` and adds missing usage metrics.

Fixes #4034.

### End of speech event

Previously, the plugin received `<end>` tokens from Soniox but did not forward them to the LiveKit agent. This caused the agent to hang or rely on timeouts to close turns. Plugin now emits `SpeechEventType.END_OF_SPEECH` upon receiving an `<end>` token from Soniox.

### Usage metrics

Implemented `RecognitionUsage` event emission. The plugin now calculates and reports audio duration based on `total_audio_proc_ms` from the Soniox response.

### Other changes

- Updated the default model from `stt-rt-preview` to `stt-rt-v3`. The old model is already just an alias to the `stt-rt-v3` model.
- VAD refactor (Breaking change): Removed the vad parameter from __init__ and the internal VAD task. Manual finalization using VAD events was removed. The plugin now relies entirely on Soniox's native server-side endpoint detection. This simplifies the architecture and aligns behavior with other STT plugins (e.g., Deepgram, AssemblyAI).

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
@@ -356,8 +356,10 @@ class SpeechStream(stt.SpeechStream):
         # Language code sent by Soniox if language detection is enabled (e.g. "en", "de", "fr")
         final_transcript_language: str = ""
 
+        is_speaking = False
+
         def send_endpoint_transcript():
-            nonlocal final_transcript_buffer, final_transcript_language
+            nonlocal final_transcript_buffer, final_transcript_language, is_speaking
             if final_transcript_buffer:
                 event = stt.SpeechEvent(
                     type=SpeechEventType.FINAL_TRANSCRIPT,
@@ -371,8 +373,12 @@ class SpeechStream(stt.SpeechStream):
 
                 self._event_ch.send_nowait(stt.SpeechEvent(type=stt.SpeechEventType.END_OF_SPEECH))
 
+                # Reset buffers.
                 final_transcript_buffer = ""
                 final_transcript_language = ""
+
+                # Reset speaking state, so the next transcript will send START_OF_SPEECH again.
+                is_speaking = False
 
         # Method handles receiving messages from the Soniox Speech-to-Text API.
         while self._ws:
@@ -414,6 +420,15 @@ class SpeechStream(stt.SpeechStream):
                                         non_final_transcription_language = token.get("language")
 
                             if final_transcript_buffer or non_final_transcription:
+                                if not is_speaking:
+                                    # Send START_OF_SPEECH if this is the first transcript.
+                                    is_speaking = True
+                                    self._event_ch.send_nowait(
+                                        stt.SpeechEvent(
+                                            type=SpeechEventType.START_OF_SPEECH,
+                                        )
+                                    )
+
                                 event = stt.SpeechEvent(
                                     type=SpeechEventType.INTERIM_TRANSCRIPT,
                                     alternatives=[


livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
chenghao-mou, line 47: I think we can remove this `FINALIZE_MESSAGE` constant.

livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
```diff
@@ -17,7 +17,6 @@
 import asyncio
```
chenghao-mou, line 17: One thing that seems missing is the `START_OF_SPEECH` event. It would be great if we can add it here too.

livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
chenghao-mou: nitpicking: VAD is still mentioned here.

livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
matejmarinko-soniox: Soniox doesn't provide explicit `START_OF_SPEECH` events. Does it make sense to send start of speech event as soon as we receive `INTERIM_TRANSCRIPT` text?

chenghao-mou: Yeah, that's probably the best. That's how some other plugins work as well.

matejmarinko-soniox: Added the `START_OF_SPEECH` event as well. Thanks for the feedback!

PR #4337, opened 2025-12-19
[draft] async text mode

longcw:
to test it 
```bash
python examples/voice_agents/sms_agent.py sms-console [--sess-data-file ./session_data.db]
```


examples/voice_agents/sms_agent.py
```diff
@@ -0,0 +1,98 @@
...
+    session = AgentSession(llm="openai/gpt-4.1-mini")
+    if ctx.session_data:
+        await session.rehydrate(ctx.session_data)
+    else:
+        await session.start(agent=MyAgent(greet_on_enter=False))
+
+    result = await session.run(user_input=ctx.text)
+
+    await ctx.send_result(result)
+    await ctx.save_session(session)
```
chenghao-mou: Q: how do you feel about automatically saving with a context manager?

```python
async with ctx.resume(session=session, session_data=session_data, agent=...):
    ...
    
```

so users don't have to call ctx.save_session explicitly?

chenghao-mou: So for `rtc_session`, we have operations at the session level:
- session.generate_reply
- session.say
- session.run

It feels out of sync to me if we have to use `ctx` to `send_result`.  WDYT?

livekit-agents/livekit/agents/cli/cli.py
```diff
@@ -1006,8 +1017,34 @@ def _key_read(ch: str) -> None:
...
+                    session_data: bytes | None = None
+                    if os.path.isfile(sess_data_file):
+                        with open(sess_data_file, "rb") as f:
+                            session_data = f.read()
+
+                    text_context = TextMessageContext(text=text, session_data=session_data)
+                    await sms_handler(text_context)
+
+                    # serialize the state of the session
+                    if text_context.session_data:
```
chenghao-mou: QQ: how about we handle this inside ctx.save_session?

examples/voice_agents/sms_agent.py
```diff
@@ -0,0 +1,98 @@
...
+            instructions="Your name is Kelly. You would interact with users via voice."
+            "with that in mind keep your responses concise and to the point."
+            "do not use emojis, asterisks, markdown, or other special characters in your responses."
+            "You are curious and friendly, and have a sense of humor."
+            "you will speak english to the user",
+        )
+        self._greet_on_enter = greet_on_enter
+
+    async def on_enter(self):
+        if self._greet_on_enter:
```
chenghao-mou: Do we want to add a function called on_rehydrate?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -842,6 +844,37 @@ async def _aclose_impl(
     async def aclose(self) -> None:
         await self._aclose_impl(reason=CloseReason.USER_INITIATED)
 
+    def get_state(self) -> dict[str, Any]:
```
chenghao-mou: Should we standardize this with `__setstate__` and `__getstate__` functions?

examples/voice_agents/sms_agent.py
theomonnom: This is technically going to be the default Python def `__setstate__` method.

livekit-agents/livekit/agents/voice/agent_session.py
theomonnom: Fwiw, for the AgentSession, I think this method should stay private

longcw: I didn't use `__setstate__` and `__getstate__` because we cannot load the AgentSession using `pickle.loads(data)` directly if it's required to create the AgentSession in user's code.

examples/voice_agents/sms_agent.py
theomonnom: Agree it’s a bit odd to have this “out of sync” issue. I think the agent code shouldn’t be sending SMS outside the sms_handler tho. Just for separation of concerns.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -842,6 +844,37 @@ async def _aclose_impl(
...
+        tool_ctx = llm.ToolContext(self.tools)
+        return {
+            "tools": list(tool_ctx.function_tools.keys()),
+            "chat_ctx": self._chat_ctx.to_dict(
```
theomonnom: On the AgentSession, we should name it history
```suggestion
            "history": self._chat_ctx.to_dict(
```

livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -368,6 +369,41 @@ def _get_activity_or_raise(self) -> AgentActivity:
 
         return self._activity
 
+    def __getstate__(self) -> dict[str, Any]:
```
theomonnom: How should it looks like from the user perspective when they add their own states, should they override this function and call `super().__getstate__`?

examples/voice_agents/sms_agent.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+server = AgentServer()
+
+
+@server.sms_handler()
+async def sms_handler(ctx: TextMessageContext):
+    logger.info(f"SMS received: {ctx.text}")
+
+    session = AgentSession(llm="openai/gpt-4.1-mini")
```
Hormold: Can we let user to setup some config for sms handler here?
Something like that:
```
- `conversation_id` (required) — unique identifier for the conversation partner (Default to ctx.participant_number?)
- `retention_hours` (optional, default 168 = 7 days) — how long to store session
- `max_tokens` (optional) — limit on loaded context size
- `encryption_key` (optional) — for E2E encryption
```

Hormold: And maybe:
auto_save — almost always True. False only if developer wants conditional saving (e.g., don't save if user asked to "forget" instantly)

examples/voice_agents/sms_agent.py
```diff
@@ -0,0 +1,98 @@
...
+
+        logger.info(f"Looking up weather for {location}")
+
+        return "sunny with a temperature of 70 degrees."
+
+
+server = AgentServer()
+
+
+@server.sms_handler()
```
davidzhao: let's position this as a text_handler, instead of sms. getting text right would mean this is accessible on LK Cloud: `https://myapp.livekit.cloud/agents/name/handler`

jayeshp19 (commented):
Hey Thanks for working on text mode. I read through the PR and had two main concerns:

1. /text should probably require auth or signed validation. it can trigger stateful handler execution through the worker HTTP server.

2. should clients be able to send `session_state.snapshot` directly, given it can reach SessionStore rehydration/deserialization paths?

I think persistence should stay server-owned by default, with clients resuming via `session_id` or a token instead. wdyt?

PR #4343, opened 2025-12-20
durable scheduler

theomonnom:


livekit-agents/livekit/agents/durable.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+
+    def __str__(self) -> str:
+        if self.exc_message:
+            return f"{self.exc_type}: {self.exc_message}"
+
+        return self.exc_type
+
+
+class EffectCall:
```
davidzhao, line 45: nit on the name. it seems a bit difficult to understand the `effect` portion of the name. is this essentially a `DurableWrapperAsync`? that wraps an async function and making it compatible with DurableFunction?

PR #4344, opened 2025-12-20
add ProviderTool & support built-in tools for xai & gemini realtime

theomonnom:
```
class SearchAgent(Agent):
    def __init__(self):
        super().__init__(instructions="You are a helpful assistant", tools=[xai.realtime.XSearch()])
```

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -863,7 +863,7 @@ class AgentActivity(RecognitionHooks):
                     elif is_function_tool(tool):
                         info = get_function_info(tool)
 
-                    if not info or (info.flags & ToolFlag.IGNORE_ON_ENTER):
+                    if info and (info.flags & ToolFlag.IGNORE_ON_ENTER):
                         continue
 
                     filtered_tools.append(tool)
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -509,15 +509,11 @@ class RealtimeSession(llm.RealtimeSession):
     async def update_tools(
         self, tools: list[llm.FunctionTool | llm.RawFunctionTool | llm.ProviderTool]
     ) -> None:
-        if self._tools == tools:
+        tool_ctx = llm.ToolContext(tools)
+        if self._tools == tool_ctx:
             return
 
-        self._tools = llm.ToolContext(tools)
-        logger.info(f"Updated function tools: {self._tools.function_tools}")
-        if len(self._tools.function_tools) > 0:
-            logger.info(
-                f"first function tool: {llm.is_function_tool(list(self._tools.function_tools.values())[0])}"
-            )
+        self._tools = tool_ctx
         self._mark_restart_needed()
 
     @property


livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/realtime/realtime_model.py
```diff
@@ -61,3 +100,31 @@ def __init__(
...
+
+    def _create_tools_update_event(
+        self, tools: list[llm.FunctionTool | llm.RawFunctionTool | Toolset]
+    ) -> SessionUpdateEvent | dict:
+        event = super()._create_tools_update_event(tools)
+
+        # inject supported Toolset
+        xai_tools: list[dict] = []
+        for tool in tools:
+            print(tool)
```
longcw:
```suggestion
```

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -33,6 +34,10 @@
 from typing_extensions import NotRequired, Required, TypedDict, TypeGuard
 
 
+class Toolset(ABC):  # noqa: B024
```
davidzhao: naming-wise, is this supposed to represent model-side tools? that seems cleaner, to have a set of tools that are provided by each model (and serving roughly the same functionality). maybe we should call it `ModelTool`

for example, we'd want to have a `WebSearchTool`, which adds itself as different tools depending on model provider. (and throws an exception if a model used does not support this tool)

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1180,17 +1185,19 @@ def _is_content_empty(msg_id: str) -> bool:
...
+    async def update_tools(
+        self, tools: list[llm.FunctionTool | llm.RawFunctionTool | Toolset]
+    ) -> None:
         async with self._update_fnc_ctx_lock:
             ev = self._create_tools_update_event(tools)
             self.send_event(ev)
 
-            assert isinstance(ev.session, RealtimeSessionCreateRequest)
-            assert ev.session.tools is not None
+            # assert isinstance(ev.session, RealtimeSessionCreateRequest)
```
davidzhao: is this intentional?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1217,6 +1225,8 @@ def _create_tools_update_event(
                 tool_desc = tool_info.raw_schema
                 tool_desc.pop("meta", None)  # meta is not supported by OpenAI Realtime API
                 tool_desc["type"] = "function"  # internally tagged
+            elif isinstance(tool, Toolset):
```
davidzhao: rather than subclasses checking, I think this should be defined within the toolset itself.. just like how we have `to_provider_format` on the ChatContext.

then it could become:

```
class WebSearch(ModelTool):
    def __init__(self, *, optional = False)
        self._optional = optional

    def to_provider_format(self, format: str):
        if ... # provider supports it
            return {"type": "web_search"}
        if not self._optional:
            raise Exception(...)
        return None

File: livekit-agents/livekit/agents/llm/tool_context.py (hunk 2 above)
Comment (chenghao-mou) [reply livekit-agents/livekit/agents/llm/tool_context.py]: The name is a bit confusing for me with the "set" term. Looking at the examples (XSearch, WebSearch) here, it feels like just one tool, and not a set.

Maybe `XToolSet` that includes individual tools like `XSearch`, `XWebSearch`?

Regard naming: I am thinking about `PlatformTool`. `ModelTool` sounds a tad unnecessary because most tools are for the model.

---

And w.r.t unified Tool interface for common functionality, I can see a situation like this:

```
WebSearchTool(
openai_top_k=NotGiven,
...
x_top_k=NotGiven,
...
claude_top_k=NotGiven,
...
google_search_retrieval=types.GoogleSearchRetrieval(
        dynamic_retrieval_config=types.DynamicRetrievalConfig(
            mode=types.DynamicRetrievalConfigMode.MODE_DYNAMIC,
            dynamic_threshold=0.7 # Only search if confidence > 70%
        )
    )
)
```

Once they decide to add more functionality or parameters. Users might also get the wrong idea that you can use the OAI search tool in X. I would say maybe let's hold off abstracting them now until we see common patterns?

Comment (longcw) [reply livekit-agents/livekit/agents/llm/tool_context.py]: `Toolset` is also overlapping with the `ToolSet` here https://github.com/livekit/agents/pull/3296/changes#diff-a083ed44e719cc13c6c2b9c9f259d216e1c824a00757f4815d7ed6ad1c352c0eR94, maybe combine them into a single one?
```python
@runtime_checkable
class ToolSet(Protocol):
    def __livekit_tools__(self) -> Sequence[FunctionTool | RawFunctionTool]: ...

    def __livekit_model_tools(self) -> Sequence[dict]: ...
```

davidzhao: good point in terms of not over-abstracting this right now. I'll take this and continue with something like the current approach.

in terms of naming, I think the inheritance tree should be:

Tool
- FunctionTool
- RawTool
- ProviderTool
  - xai.XSearch
  - xai.WebSearch
  - google.WebSearch
  - ...

ToolSet (a group of tools)
  - works like a set, where multiple sets could be merged to form a new ToolSet
  - then all of our interfaces could accept Tool or ToolSet
  - you can mix a ProviderTool and a RawTool into the same ToolSet

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -851,9 +855,9 @@ def _generate_reply(
         # if tool has the IGNORE_ON_ENTER flag, every generate_reply inside on_enter will ignore it
         if on_enter_data := _OnEnterContextVar.get(None):
             if on_enter_data.agent == self._agent and on_enter_data.session == self._session:
-                filtered_tools = []
+                filtered_tools: list[llm.FunctionTool | llm.RawFunctionTool | llm.ProviderTool] = []
```
longcw, line 858: `ProviderTool` will be ignored silently in `on_enter`, is this expected?

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -97,7 +97,6 @@ async def _llm_inference_task(
     data.started_fut.set_result(None)
 
     text_ch, function_ch = data.text_ch, data.function_ch
-    tools = list(tool_ctx.function_tools.values())
```
longcw, line 100: it was intended to use a copy of tools as input of the `llm_node`, users can modify the content of the copy to change the tool for this turn.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -512,17 +506,19 @@ async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
...
-    async def update_tools(self, tools: list[llm.FunctionTool | llm.RawFunctionTool]) -> None:
-        new_declarations: list[types.FunctionDeclaration] = to_fnc_ctx(
-            tools, use_parameters_json_schema=False, tool_behavior=self._opts.tool_behavior
-        )
-        current_tool_names = {f.name for f in self._gemini_declarations}
-        new_tool_names = {f.name for f in new_declarations}
+    async def update_tools(
+        self, tools: list[llm.FunctionTool | llm.RawFunctionTool | llm.ProviderTool]
+    ) -> None:
+        if self._tools == tools:
```
longcw: `self._tools` is `ToolContext` and `tools` is a list, it's always False.

livekit-agents/livekit/agents/voice/agent_activity.py
davidzhao: good point, that's not the intent. provider tools cannot be filtered

livekit-agents/livekit/agents/voice/generation.py
davidzhao: `tool_ctx.all_tools` is a copy of the tools always. it seems this can be abstracted so that tool_context's tools are not accidentally mutable

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
davidzhao: yup, good catch! I meant to convert it to a ToolContext first.

livekit-agents/livekit/agents/voice/generation.py
longcw: changes on the `all_tools` property won't be kept, I think we intended to make `tools` of the `llm_node` mutable for this turn but it won't change the `tool_context` globally.

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -109,12 +108,12 @@ async def _llm_inference_task(
...
     )
 
-    llm_node = node(chat_ctx, tools, model_settings)
+    llm_node = node(chat_ctx, tool_ctx.all_tools, model_settings)
     if asyncio.iscoroutine(llm_node):
         llm_node = await llm_node
 
     # update the tool context after llm node
-    tool_ctx.update_tools(tools)
+    tool_ctx.update_tools(tool_ctx.all_tools)
```
tinalenguyen, line 116: quick question, are any changes made to `tool_ctx.all_tools` within the `llm_node` coroutine? otherwise, isn't this redundant?

PR #4352, opened 2025-12-21
fix: return error ToolOutput on unknown tools

sebitokazu:
## Description

When LLM calls a function that doesn't exist, we're only logging and swallowing the error. 
Now, we return an error ToolOutput with the list of available tools, so the LLM can call the correct one or mention the error to the user


theomonnom:
Thanks I slightly edited the PR in https://github.com/livekit/agents/pull/4374

Issue #2117, opened 2025-04-24
Is it possible to display agent transcript in sync with audio (instead of waiting for full transcription)?


marcinjackowski:
Hi LiveKit team,

I’m working with LiveKit Agent and would like to display the agent’s transcript text in real time, synchronized with the audio, as the agent is speaking.

I’ve tried listening to the conversation_item_added event, but it seems that this event only fires after the entire agent response (both audio and text) is completed. This results in the text being shown all at once, rather than incrementally.

Is there a way to receive partial transcript updates or otherwise display the agent’s spoken response dynamically, in sync with the audio?

Any guidance would be greatly appreciated!

Thanks in advance 🙌

longcw:
check the [document here](https://docs.livekit.io/agents/build/text/#transcriptions) for the audio synchronized transcription.

marcinjackowski:
Thanks for the quick response! I really appreciate it! Unfortunately, I still don’t have the answer I’m looking for.

What I’d like to achieve is getting real-time text on the backend side (agent), not on the frontend. On the frontend, I know it’s straightforward to print synchronized text as audio plays, but on the agent side, I’m really struggling to find the right way to do this.

I’d really appreciate it if you could share a short code snippet showing how to achieve that.

longcw:
If you really want to get that on the agent side, for now you can define a custom `TextOutput` and inject it into the room io
```python
from livekit.agents.voice.io import TextOutput


class MyTextOutput(TextOutput):
    def __init__(self, *, next_in_chain: TextOutput | None = None) -> None:
        super().__init__(next_in_chain=next_in_chain)

    async def capture_text(self, text: str) -> None:
        print(text, end="", flush=True)

    def flush(self) -> None:
        print("\n")

async def entrypoint(...):
    await session.start(
        ...,
        room_output_options=RoomOutputOptions(transcription_enabled=True),
    )

    session._room_io._agent_tr_output._sinks.append(MyTextOutput())
```

but pls note `session._room_io._agent_tr_output._sinks` is not a public api so it might be changed in the future. We are planning to allow the output to be chained easily for next versions.

I am also curious what is the use case of the synced agent transcription on the agent side.

Belgskinny:
`session._room_io._agent_tr_output._sinks` stopped working in the latest version. What is the equivalent now?
@longcw

Issue #4341, opened 2025-12-19
Is it possible to display agent transcript in sync with audio (instead of waiting for full transcription)?


Belgskinny:
### Bug Description

In the old version I just follow this guide here: https://github.com/livekit/agents/issues/2117#issuecomment-2829609821.
but `session._room_io._agent_tr_output._sinks` stopped working in the latest version. What is the equivalent now?

### Expected Behavior

`session._room_io._agent_tr_output._sinks` should work

### Reproduction Steps

```bash
1. see https://github.com/livekit/agents/issues/2117#issuecomment-2829609821
2.
3.
...
- Sample code snippet, or a GitHub Gist link -
```

### Operating System

macOS

### Package Versions

```bash
latest version
```

### Proposed Solution

```python

```

longcw:
created a fix https://github.com/livekit/agents/pull/4353 so you can add a custom text output to the chain

PR #4353, opened 2025-12-22
support next_in_chain for RoomIO text output

longcw:
close https://github.com/livekit/agents/issues/2117 and https://github.com/livekit/agents/issues/4341

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/types.py
+++ b/livekit-agents/livekit/agents/voice/room_io/types.py
@@ -104,7 +104,7 @@ class TextOutputOptions:
     """Speed factor of transcription synchronization with audio output.
     Only effective if `sync_transcription` is True."""
     next_in_chain: TextOutput | None = None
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/room_io/types.py
```diff
@@ -102,6 +103,8 @@ class TextOutputOptions:
     transcription_speed_factor: float = 1.0
     """Speed factor of transcription synchronization with audio output.
     Only effective if `sync_transcription` is True."""
+    next_in_chain: TextOutput | None = None
+    """The next text output in the chain for the agent. If provided, the agent's transcription will be passed to it."""
```
davidzhao, line 107: would this bypass the transcription output when set?

not sure about the naming for the param `next_in_chain`

longcw: no, it's the next node chained after the transcription output if provided, the output will be forwarded to it after the original transcription output node.

PR #4365, opened 2025-12-23
Add allowed_tools and transport_type parameters to MCPServerHTTP

wasaybaig201:
- Add transport_type parameter to explicitly specify 'sse' or 'streamable_http'
- Add allowed_tools parameter to filter which tools are available
- Maintain backward compatibility with URL-based transport detection


alisadiq-ai:
This feature would be very helpful as most consumers dont want all the tools to be exposed to the Agent in an MCP server.

theomonnom (approved):
Thanks!

PR #4366, opened 2025-12-23
(gemini realtime) check for vertexai for api version

tinalenguyen:
it appears that VertexAI models do not work with `api_version="v1alpha"`


davidzhao (approved):
lg

theomonnom (approved):

PR #4368, opened 2025-12-23
standardize Tool interface

longcw:
- add `Tool` as base class of `FunctionTool`, `RawFunctionTool` and `ProviderTool`
- add `ToolSet`
- `Agent` and `update_tools` accept `list[Tool | ToolSet]` as argument
- LLM and realtime plugins still use `list[Tool]` as argument, takes `tool_ctx.all_tools` as input

### DIFF
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -527,7 +527,7 @@ async def test_generate_reply() -> None:
     ],
 )
 async def test_preemptive_generation(preemptive_generation: bool, expected_latency: float) -> None:
-    speed = 5.0
+    speed = 2.0
     actions = FakeActions()
     actions.add_user_speech(0.5, 2.0, "Hello, how are you?", stt_delay=0.2)
     actions.add_llm("I'm doing great, thank you!", ttft=0.1, duration=0.3)


livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -14,31 +14,36 @@
...
+
+class ProviderTool(Tool):
+    """Tool that provided by the LLM provider."""
 
-# TODO: refactor Tool inheritance, all tools (FunctionTool, RawFunctionTool, ProviderTool) should inherit from Tool
-class ProviderTool(ABC):  # noqa: B024
     pass
 
 
+class ToolSet(ABC):
```
theomonnom:
```suggestion
class Toolset(ABC):
```

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -254,28 +288,35 @@ def empty(cls) -> ToolContext:
...
+        """A copy of all provider tools in the tool context."""
         return self._provider_tools
 
     @property
-    def all_tools(self) -> list[FunctionTool | RawFunctionTool | ProviderTool]:
-        tools: list[FunctionTool | RawFunctionTool | ProviderTool] = []
-        tools.extend(list(self._tools_map.values()))
+    def tool_sets(self) -> list[ToolSet]:
+        """A copy of all tool sets in the tool context."""
+        return self._tool_sets
```
theomonnom:
```suggestion
    def toolsets(self) -> list[ToolSet]:
        """A copy of all tool sets in the tool context."""
        return self._tool_sets
```

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -286,33 +327,73 @@ def __eq__(self, other: object) -> bool:
...
+        tool_behavior: _provider_format.google.TOOL_BEHAVIOR | None = None,
+    ) -> list[dict[str, Any]]: ...
+
+    @overload
+    def to_provider_format(self, format: Literal["aws"]) -> list[dict[str, Any]]: ...
+
+    @overload
+    def to_provider_format(self, format: Literal["anthropic"]) -> list[dict[str, Any]]: ...
+
+    def to_provider_format(
```
theomonnom: It isn't obvious that `to_provider_format` will not include any ProviderTool. Can we add a comment that this is only returning function_tools. 

Or maybe a different function name?


For more context, also see: https://live-kit.slack.com/archives/C07TWVC0W4A/p1766344944136839?thread_ts=1766343122.195189&cid=C07TWVC0W4A

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -254,28 +288,35 @@ def empty(cls) -> ToolContext:
...
+    def tool_sets(self) -> list[ToolSet]:
+        """A copy of all tool sets in the tool context."""
+        return self._tool_sets
+
+    @property
+    def all_tools(self) -> list[Tool]:
+        tools: list[Tool] = []
+        tools.extend(list(self._fnc_tools_map.values()))
         tools.extend(self._provider_tools)
         return tools
```
theomonnom: I don't think this should be a property, wdyt of
```suggestion
    def flatten(self) -> list[Tool]:
        tools: list[Tool] = []
        tools.extend(list(self._fnc_tools_map.values()))
        tools.extend(self._provider_tools)
        return tools
```

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -14,31 +14,36 @@
...
 
-# TODO: refactor Tool inheritance, all tools (FunctionTool, RawFunctionTool, ProviderTool) should inherit from Tool
-class ProviderTool(ABC):  # noqa: B024
     pass
 
 
+class ToolSet(ABC):
+    @abstractmethod
+    def get_tools(self) -> list[Tool]:
+        pass
```
theomonnom, line 46:
```suggestion
    @property
    @abstractmethod
    def tools(self) -> Sequence[Tool]:
        """Tools exposed by this toolset."""
        raise NotImplementedError
```

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -82,23 +87,32 @@ def __init__(self) -> None:
...
+    """Wrapper for a function decorated with @function_tool"""
 
-    def __call__(self, *args: Any, **kwargs: Any) -> Any: ...
+    def __init__(self, func: Callable[..., Any], info: _FunctionToolInfo) -> None:
+        self._func = func
+        self._info = info
+        functools.update_wrapper(self, func)
 
+    @property
+    def info(self) -> _FunctionToolInfo:
```
theomonnom: Should we now make _FunctionToolInfo public?

theomonnom (approved):
lgtm! this is way better!

livekit-agents/livekit/agents/llm/tool_context.py
davidzhao: I think `ToolSet` are two words so this capitalization seems more correct

theomonnom: Seems like other frameworks are using Toolset:
https://github.com/pydantic/pydantic-ai/blob/c5352758442d20d704220dbe54f3d4322096db21/pydantic_ai_slim/pydantic_ai/toolsets/abstract.py#L62

theomonnom: It seems cleaner TBH, e.g: ProviderToolSet vs ProviderToolset

But I'm definitely opinionated

davidzhao: fair.. Toolset def seems more "standard".. according to Grok: https://grok.com/share/bGVnYWN5_8f2a4ef7-16be-4cc9-be43-876badd917f4

livekit-agents/livekit/agents/llm/tool_context.py
longcw: yeah, ProviderTool is not included because in plugins we can access them via `tool_ctx.provider_tools` directly. maybe use a different method name like `parse_function_tools()`

livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -247,26 +239,27 @@ def chat(
 class LLMStream(llm.LLMStream):
     def __init__(
         self,
-        llm: LLM | llm.LLM,
+        llm_v: LLM | llm.LLM,
```
davidzhao, line 242: 👍

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -287,30 +287,29 @@ def copy(
         exclude_instructions: bool = False,
         exclude_empty_message: bool = False,
         exclude_handoff: bool = False,
-        tools: NotGivenOr[
-            Sequence[FunctionTool | RawFunctionTool | ProviderTool | str | Any]
-        ] = NOT_GIVEN,
+        tools: NotGivenOr[Sequence[Tool | Toolset | str | Any]] = NOT_GIVEN,
```
davidzhao: do we still want to allow `Any` here?

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -97,16 +93,19 @@ async def _llm_inference_task(
...
-    )
-    current_span.set_attribute(
-        trace_types.ATTR_FUNCTION_TOOLS, json.dumps(list(tool_ctx.function_tools.keys()))
+    current_span.set_attributes(
+        {
+            trace_types.ATTR_CHAT_CTX: json.dumps(
+                chat_ctx.to_dict(exclude_audio=True, exclude_image=True, exclude_timestamp=False)
+            ),
+            trace_types.ATTR_FUNCTION_TOOLS: list(tool_ctx.function_tools.keys()),
+            trace_types.ATTR_PROVIDER_TOOLS: [
```
davidzhao, line 104: nice

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -884,7 +881,7 @@ def chat(
         self,
         *,
         chat_ctx: ChatContext,
-        tools: list[FunctionTool | RawFunctionTool | ProviderTool] | None = None,
+        tools: list[llm.Tool] | None = None,
```
davidzhao, line 884: should Toolset be supported at the LLM level? or is this strictly tools?

davidzhao (approved):
great work! much cleaner

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
longcw: I think for now at the LLM and realtime model level, we can still use `Tool` for simplicity

PR #4374, opened 2025-12-23
fix silent generation on invalid AI functions

theomonnom:
based on https://github.com/livekit/agents/pull/4352
avoid failing silently when the LLM hallucinates

### DIFF
--- a/livekit-agents/livekit/agents/voice/generation.py
+++ b/livekit-agents/livekit/agents/voice/generation.py
@@ -476,6 +476,13 @@ async def _execute_tools_task(
                         "speech_id": speech_handle.id,
                     },
                 )
+                _tool_completed(
+                    make_tool_output(
+                        fnc_call=fnc_call,
+                        output=None,
+                        exception=RuntimeError(f"unknown AI function `{fnc_call.name}`"),
+                    )
+                )
                 continue
 
             if not is_function_tool(function_tool) and not is_raw_function_tool(function_tool):
--- a/livekit-agents/livekit/agents/voice/generation.py
+++ b/livekit-agents/livekit/agents/voice/generation.py
@@ -481,6 +481,7 @@ async def _execute_tools_task(
                         fnc_call=fnc_call,
                         output=None,
                         exception=RuntimeError(f"unknown AI function `{fnc_call.name}`"),
+                        include_error_details_in_llm_output=True,
                     )
                 )
                 continue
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -476,6 +476,13 @@ def _tool_completed(out: ToolExecutionOutput) -> None:
                         "speech_id": speech_handle.id,
                     },
                 )
+                _tool_completed(
```
davidzhao, line 479: currently xai is calling these tools even though they are supposed to be executed on the server side, I'll raise with them.. but just to note that this will likely break the xai tools

theomonnom: which tool did you test? XSearch or WebSearch?

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -476,6 +476,14 @@ def _tool_completed(out: ToolExecutionOutput) -> None:
...
+                    make_tool_output(
+                        fnc_call=fnc_call,
+                        output=None,
+                        exception=RuntimeError(f"unknown AI function `{fnc_call.name}`"),
```
Hormold, line 483: Maybe we should expose the list of available functions back to the LLM to help with the next step (optionally)?

livekit-agents/livekit/agents/voice/generation.py
Hormold: for example, web + x search request:
 
 ```
  14:32:53.218 WARNI… livekit.agents     unknown AI function `web_search_with_snippets` {"function": "web_search_with_snippets", "speech_id": "speech_c7ce525a9080", "room": "mock_room"}
    14:32:53.221 WARNI… livekit.agents     unknown AI function `x_keyword_search` {"function": "x_keyword_search", "speech_id": "speech_c7ce525a9080", "room": "mock_room"}
    14:32:53.223 WARNI… livekit.agents     unknown AI function `x_semantic_search` {"function": "x_semantic_search", "speech_id": "speech_c7ce525a9080", "room": "mock_room"}
```

Hormold: - x_semantic_search
- x_keyword_search
- x_user_search
- x_thread_fetch
- web_search_with_snippets
- browse_page

**here is full list**

livekit-agents/livekit/agents/voice/generation.py
theomonnom: It should already have it in context?

Issue #4339, opened 2025-12-19
Logging configuration for LiveKit Agents: avoid JSON stdout in production mode


dhansmair:
### Feature Type

Would make my life easier

### Feature Description

Hi there,

I’m exploring how logging in LiveKit Agents can be configured. In production mode, the worker outputs JSON-formatted logs to stdout. Our project would prefer plain-text (non-JSON) logging.

Context:
- I inspected the source and found that LiveKit attaches a separate handler to the root logger.
- To tailor logging for our needs, I would need to attach our own handler to the root logger and remove the JSON handler. However, the JSON handler is only attached when `cli.run_app` is called. This means it cannot be removed except from within the entrypoint function, in which case there will be a period during which logs are emitted in JSON format.

Questions:
- What is the recommended, clean, and “unhacky” way to solve this?
- The conventional approach for Python libraries is to maintain their own scoped loggers. What is the motivation for attaching the handler directly to the root logger, rather than to a library-specific logger such as "livekit" or "livekit.agents"?

Best regards,
David

PR #4382, opened 2025-12-24
add log_setup_fnc for cli.run_app

longcw:
close https://github.com/livekit/agents/issues/4339

for example
```python

def log_setup(logger: logging.Logger, log_level: int | str):
    logger.setLevel(log_level)
    handler = logging.StreamHandler(sys.stdout)
    handler.setFormatter(logging.Formatter("%(asctime)s - %(levelname)s - %(message)s"))
    logger.addHandler(handler)


if __name__ == "__main__":
    cli.run_app(server, log_setup_fnc=log_setup)
```

### DIFF
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -1262,12 +1262,7 @@ def _run_console(
         raise typer.Exit(code=1) from None
 
 
-def _run_worker(
-    server: AgentServer,
-    args: proto.CliArgs,
-    jupyter: bool = False,
-    log_setup_fnc: Callable[[logging.Logger, int | str], None] | None = None,
-) -> None:
+def _run_worker(server: AgentServer, args: proto.CliArgs, jupyter: bool = False) -> None:
     c: AgentsConsole | None = None
     if args.devmode:
         c = AgentsConsole.get_instance()  # colored logs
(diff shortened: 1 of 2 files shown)


longcw:
> ideally this isn't a CLI parameter, but something they could configure inside their agent?

users can remove the json handler in entrypoint but the original issue mentioned that 
> the JSON handler is only attached when cli.run_app is called. This means it cannot be removed except from within the entrypoint function, in which case there will be a period during which logs are emitted in JSON format.

maybe we can make it a callback of AgentServer or `cli.run_app` for user to customize rather than adding a CLI parameter.

davidzhao:
> > the JSON handler is only attached when cli.run_app is called. This means it cannot be removed except from within the entrypoint function, in which case there will be a period during which logs are emitted in JSON format.
> 
> maybe we can make it a callback of AgentServer or `cli.run_app` for user to customize rather than adding a CLI parameter.

exactly, entrypoint is a bit too late, the framework logger has already been used by that point.. IMO giving users control is the right path, instead of automatically configuring colored logger

theomonnom:
I'm wondering if we should just have a flag that disable any automatic logging configuration from the framework

davidzhao:
> I'm wondering if we should just have a flag that disable any automatic logging configuration from the framework

would that give users the ability to configure the logger as soon as the process is spun up? I think it's two related goals
- do not override the user's logging configuration
- letting them configure the logger very early on, before the framework itself starts emitting logs

longcw:
> I'm wondering if we should just have a flag that disable any automatic logging configuration from the framework

I tried to do that, but it seems it's not easy for user to setup the logger in dev mode when reload enabled without passing the fnc to the process?

Issue #4378, opened 2025-12-24
Add WebSocket (JSON and non-JSON) support to the Rime TTS LiveKit plugin with client-side protocol switching.


gokuljs:
### Feature Type

Would make my life easier

### Feature Description

**Description:**

Hey LiveKit team 👋,

I am planning to improve and extend the `livekit-rime-plugin` to support additional Rime transport protocols.

### Current state

At the moment, the Rime plugin supports:

* Rime TTS over HTTP

Rime itself supports additional protocols:

* WebSocket (JSON)
* WebSocket (non-JSON / binary)

### Goal

I would like to extend the existing Rime LiveKit plugin so it can support:

* HTTP
* WebSocket (JSON)
* WebSocket (non-JSON)

### Design question

Before starting implementation, I wanted to get feedback on the overall design approach.

One idea is to keep a **single Rime plugin** and allow users to select the protocol during initialization. For example:

* Add a `protocol` parameter when initializing the Rime TTS
* Supported values could be something like `http`, `ws_json`, `ws_binary`
* Inside `synthesize()`, route to the appropriate implementation and return the correct stream type based on the selected protocol

This would allow:

* A single, unified Rime plugin
* Minimal API surface changes for users
* Flexibility to switch protocols without changing plugins

### Question

Does this approach align with LiveKit’s plugin design philosophy, or would you recommend:

* Separate implementations internally with a shared interface
* Multiple plugins instead of one
* Any existing patterns in the codebase that would be better to follow here

Before moving forward, I would appreciate any guidance or recommendations on the preferred design.

Thanks!

### Workarounds / Alternatives

No workarounds

gokuljs:
#4389

PR #4389, opened 2025-12-25
feat(TTS): Add WebSocket (JSON and non-JSON) support to the Rime TTS LiveKit plugin with client-side protocol switching.

gokuljs:


gokuljs:
#4378

PR #4391, opened 2025-12-26
fix AttributeError in NVIDIA Riva STT

gau-nernst:
I'm getting `AttributeError` here

https://github.com/livekit/agents/blob/80861211bad66a584c7ba84e6cd0666d2485f372/livekit-plugins/livekit-plugins-nvidia/livekit/plugins/nvidia/stt.py#L289-L291

The safer way is to use `getattr()`, which I see that it is already used in a few lines earlier

https://github.com/livekit/agents/blob/80861211bad66a584c7ba84e6cd0666d2485f372/livekit-plugins/livekit-plugins-nvidia/livekit/plugins/nvidia/stt.py#L278-L279

Pinging @chenghao-mou for review since it looks like you made the last commit to the file. Thank you!

### DIFF
--- a/livekit-plugins/livekit-plugins-nvidia/livekit/plugins/nvidia/stt.py
+++ b/livekit-plugins/livekit-plugins-nvidia/livekit/plugins/nvidia/stt.py
@@ -286,9 +286,9 @@ class SpeechStream(stt.SpeechStream):
             text=transcript,
             words=[
                 TimedString(
-                    text=word.get("word", ""),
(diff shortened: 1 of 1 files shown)


davidzhao:
can you please share the full stacktrace? `word` should be a json dict.. so an `AttributeError` would indicate something more sinister.

gau-nernst:
I will check more detailed next monday, but iirc, word is a protobuf object.

davidzhao:
I see, you are right. this module isn't typed and that makes it a bit difficult to see the structure.

PR #4393, opened 2025-12-26
fix (gemini streaming tts): change default audio encoding and voice 

### LINKED ISSUE #4386: TTS not working when gemini-3.-flash-preview model is used, but it is working for 2.5-flash
### Bug Description

TTS not working when gemini-3.-flash-preview model is used, but it is working for 2.5-flash

### Expected Behavior

It should work

### Operating System

Ubuntu

### Models Used

_No response_

### Package Versions

```bash
livekit-agent=1.3.10
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

tinalenguyen:
closes #4386

- change default audio encoding to PCM
- add Gemini TTS models 
- update default voice logic for chirp 3 and gemini models

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
@@ -32,7 +32,7 @@ from livekit.agents.types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, NotGive
 from livekit.agents.utils import is_given
 
 from .log import logger
-from .models import Gender, SpeechLanguages
+from .models import GeminiTTSModels, Gender, SpeechLanguages
 
 NUM_CHANNELS = 1
 DEFAULT_LANGUAGE = "en-US"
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -141,7 +140,13 @@ def __init__(
                 voice_cloning_key=voice_cloning_key,
             )
         else:
-            voice_params.name = voice_name if is_given(voice_name) else DEFAULT_VOICE_NAME
+            voice_params.name = (
+                voice_name
+                if is_given(voice_name)
```
davidzhao: this is a bit hard to parse, could we break this out into more reasonable params? 

do they support different voice names between the models? then we need a explicit `model_name` check as well? (or comment indicating how the default is chosen)

tinalenguyen: seems like Chirp 3 is the only outlier that requires the prefix, i added a `model_name` check

davidzhao (approved):
lg

Issue #4325, opened 2025-12-19
min_endpointing_delay behaves differently in VAD vs STT turn detection mode


MonkeyLeeT:
### Bug Description

  In 1.3.7, endpointing waits until last_speaking_time + min_endpointing_delay.

- VAD mode: last_speaking_time is set during speech (VAD inference), so by VAD END_OF_SPEECH the delay may already
    be satisfied → effective delay ≈ max(VAD_min_silence, min_endpointing_delay).
- STT mode: last_speaking_time is set at STT END_OF_SPEECH, so the full min_endpointing_delay is always added after
    STT EOS → additive.

Docs imply a consistent “after EOS” delay, but behavior is mode‑dependent and EOU metrics in STT mode exclude STT
    silence which is perceived by user.

### Expected Behavior

  Either:

  - Consistent “after end‑of‑utterance signal” delay in all modes, or
  - Docs/metrics clearly state VAD = max(), STT = additive.

### Reproduction Steps

```bash
1. VAD mode: set VAD min_silence=2s, min_endpointing_delay=2s. Observe commit ≈ 2s after user stop, not 4s.
2. STT mode (with providers like AssemblyAI): set STT EOS silence=2s and min_endpointing_delay=2s. Observe total ≈ 4s (STT silence + endpointing), while EOU metric end_of_utterance_delay shows ~2s.
```

### Operating System

Linux

### Package Versions

```bash
livekit-agents=1.3.7
```

### Proposed Solution

```python
- Update docs to explain mode‑dependent behavior.
- Or set last_speaking_time at EOS in VAD mode for consistent additive behavior.
- Add a metric for “silence‑to‑commit” including STT EOS delay.
```

longcw:
`min_endpointing_delay` is the delay after the end of user speech, it's the same for both VAD and STT mode. the difference is with VAD we have a more accurate timing when the user stopped speaking (before VAD fires `END_OF_SPEECH` event).

> Docs imply a consistent “after EOS” delay

can you point me where this is mentioned in the doc?

MonkeyLeeT:
> `min_endpointing_delay` is the delay after the end of user speech, it's the same for both VAD and STT mode. the difference is with VAD we have a more accurate timing when the user stopped speaking (before VAD fires `END_OF_SPEECH` event).
> 
> > Docs imply a consistent “after EOS” delay
> 
> can you point me where this is mentioned in the doc?

[...]

PR #4396, opened 2025-12-26
Fixes #4388: Correct transcription_delay metric calculation in STT turn detec…

### LINKED ISSUE #4388: Incorrect transcription_delay when using STT turn detection mode
## Problem

  When using `turn_detection="stt"` (e.g. with Deepgram Flux), transcription_delay 
  is always ~0 because END_OF_SPEECH overwrites `_last_speaking_time`:

  ```python
  # audio_recognition.py:446-452
  elif ev.type == SpeechEventType.END_OF_SPEECH and self._turn_detection_mode == "stt":
      self._speaking = False
      self._user_turn_committed = True
      self._last_speaking_time = time.time()  # overwrites previous value
```
  Then the calculation becomes:
```python
  # line 594
  transcription_delay = last_final_transcript_time - last_speaking_time
  # Both are ~time.time(), so result is ~0
```
 
Also, if VAD is passed alongside STT turn detection, its timing gets discarded  by this overwrite. Seems inconsistent behavior when VAD is being used for calculating latency in normal turn detection modes (en or multi lingual model)

  Broader question

  Why is transcription_delay using VAD events and wall-clock time.time() for delay calculation?

  SpeechData already has audio timestamps:
  - end_time - when speech ended in the audio stream
  - words[].end_time - per-word timing
  - start_time_offset - audio stream reference

  These tell us exactly when speech occurred in the audio. The delay calculation
  could be:

  transcription_delay = now - (audio_stream_start + speech_end_in_audio)

  Note: The data is already there in SpeechData, just not being used.
  

### Expected Behavior

Ideally, transcription delay should be accurately calculated using speech data in cases of stt turn based detection

### Reproduction Steps
Use deepgram flux with turn detection mode as stt

```python

@server.rtc_session
async def my_agent(ctx: JobContext):
    # Logging setup
    # Add any other context you want in all log entries here
    ctx.log_context_fields = {
        "room": ctx.room.name,
    }
    stt = deepgram.STTv2()
    tts = elevenlabs.TTS(
        model="eleven_turbo_v2_5", voice_id="jqcCZkN6Knx8BJ5TBdYR", auto_mode=True
    )
    llm = anthropic.LLM(model="claude-sonnet-4-5", caching="ephemeral")

    # Set up a voice AI pipeline using OpenAI, Cartesia, AssemblyAI, and the LiveKit turn detector
    session = AgentSession(
        # Speech-to-text (STT) is your agent's ears, turning the user's speech into text that the LLM can understand
        # See all available models at https://docs.livekit.io/agents/models/stt/
        stt=stt,
        # A Large Language Model (LLM) is your agent's brain, processing user input and generating a response
        # See all available models at https://docs.livekit.io/agents/models/llm/
        llm=llm,
        # Text-to-speech (TTS) is your agent's voice, turning the LLM's text into speech that the user can hear
        # See all available models as well as voice selections at https://docs.livekit.io/agents/models/tts/
        tts=tts,
        # VAD and turn detection are used to determine when the user is speaking and when the agent should respond
        # See more at https://docs.livekit.io/agents/build/turns
        turn_detection=EnglishModel(),
        vad=ctx.proc.userdata["vad"],
        # allow the LLM to generate a response while waiting for the end of turn
        # See more at https://docs.livekit.io/agents/build/audio/#preemptive-generation
        preemptive_generation=True,
    )

    # Start the session, which initializes the voice pipeline and warms up the models
    await session.start(
        agent=Assistant(),
        room=ctx.room,
                    noise_cancellation=lambda params: noise_cancellation.BVCTelephony()
                    if params.participant.kind == rtc.ParticipantKind.PARTICIPANT_KIND_SIP
                    else noise_cancellation.BVC(),
            ),
        ),
        record=True
    )

    # Join the room and connect to the user
    await ctx.connect()

if __name__ == "__main__":
    cli.run_app(server)
```

You will notice the EOU Metrics will have 0 as transcription_delay
```
EOU metrics {"model_name": "en", "model_provider": "livekit", "end_of_utterance_delay": 0.81, "transcription_delay": 0.79}
```

### Operating System

MacOS

### Models Used

Deepgram flux

### Package Versions

```bash
livekit==1.0.23
livekit-agents==1.3.10
livekit-api==1.1.0
livekit-blingfire==1.1.0
livekit-plugins-anthropic==1.3.10
livekit-plugins-cartesia==1.3.10
livekit-plugins-deepgram==1.3.10
livekit-plugins-elevenlabs==1.3.10
livekit-plugins-google==1.3.10
livekit-plugins-noise-cancellation==0.2.5
livekit-plugins-openai==1.3.10
livekit-plugins-silero==1.3.10
livekit-plugins-turn-detector==1.3.10
livekit-protocol==1.1.1
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
Use Speech Data or VAD if provided as mentioned above (Ideally speech data as that is more accurate representation of how much audio transcription providers have processed so far)
```

### Additional Context

_No response_

_No response_

devbyteai:
## Summary

Fixes #4388

This PR fixes the incorrect `transcription_delay` metric calculation when using STT-based turn detection (e.g., Deepgram Flux).

---

## Problem

When using STT turn detection mode, the `transcription_delay` metric incorrectly shows ~0 seconds instead of reflecting the actual transcription latency.

**User-Reported Behavior:**
> "EOU metrics showing ~0.79 transcription_delay when should reflect actual processing time"

The metric should measure the time between when the user stopped speaking and when the transcript was received, but it was always returning near-zero values.

---

## Root Cause

In `audio_recognition.py`, the `transcription_delay` is calculated as:

```python
transcription_delay = max(last_final_transcript_time - last_speaking_time, 0)
```

The bug was in the STT END_OF_SPEECH handler (line 452), which **overwrote** `_last_speaking_time` with `time.time()`:

```python
    ...
    self._last_speaking_time = time.time()  # BUG: Overwrites the value!
```

**Event Timeline in STT Mode (Buggy):**
1. START_OF_SPEECH → `_last_speaking_time = time.time()` (correct)
2. FINAL_TRANSCRIPT → `_last_final_transcript_time = time.time()` (correct)
3. END_OF_SPEECH → `_last_speaking_time = time.time()` (BUG - overwrites!)

Since END_OF_SPEECH typically arrives shortly after FINAL_TRANSCRIPT in STT mode, both timestamps become nearly identical, resulting in `transcription_delay ≈ 0`.

---

## Solution

Remove the line that overwrites `_last_speaking_time` at END_OF_SPEECH in STT mode. The value was already correctly set at START_OF_SPEECH.

**Comparison with VAD Mode:**
VAD mode does NOT update `_last_speaking_time` at END_OF_SPEECH - it keeps the value from the last INFERENCE_DONE event. STT mode should follow the same pattern.

**After Fix:**
1. START_OF_SPEECH → `_last_speaking_time = time.time()` (preserved)
2. FINAL_TRANSCRIPT → `_last_final_transcript_time = time.time()`
3. END_OF_SPEECH → No overwrite

Result: `transcription_delay = last_final_transcript_time - last_speaking_time` now correctly represents the actual transcription latency.

---

## Testing

All 15 existing agent session tests pass:

```
tests/test_agent_session.py::test_events_and_metrics PASSED
tests/test_agent_session.py::test_tool_call PASSED
tests/test_agent_session.py::test_interruption[False-5.5] PASSED
tests/test_agent_session.py::test_interruption[True-5.5] PASSED
tests/test_agent_session.py::test_interruption_options PASSED
tests/test_agent_session.py::test_interruption_by_text_input PASSED
tests/test_agent_session.py::test_interruption_before_speaking[False-3.5] PASSED
tests/test_agent_session.py::test_interruption_before_speaking[True-3.5] PASSED
tests/test_agent_session.py::test_generate_reply PASSED
tests/test_agent_session.py::test_preemptive_generation[True-0.8] PASSED
tests/test_agent_session.py::test_preemptive_generation[False-1.1] PASSED
tests/test_agent_session.py::test_interrupt_during_on_user_turn_completed[False-0.0] PASSED
tests/test_agent_session.py::test_interrupt_during_on_user_turn_completed[False-2.0] PASSED
tests/test_agent_session.py::test_interrupt_during_on_user_turn_completed[True-0.0] PASSED
tests/test_agent_session.py::test_interrupt_during_on_user_turn_completed[True-2.0] PASSED

======================== 15 passed in 75.96s ========================
```

---

## Backward Compatibility

**No breaking changes** - This fix only corrects the metric calculation. The actual agent behavior (speech recognition, turn detection, interruption handling) is completely unchanged.

**Expected Impact:**
- Users with STT turn detection will now see accurate `transcription_delay` values in their metrics
- Dashboards showing this metric will now report correct latency (previously under-reported as ~0)

---

## Edge Cases Handled

1. **No VAD present** - Already handled at lines 376-382, falls back to STT timestamps
2. **Multiple speech segments** - START_OF_SPEECH updates `_last_speaking_time` for each new segment
3. **Preflight transcripts** - Also update `_last_final_transcript_time` correctly
4. **VAD mode unchanged** - Fix only affects STT turn detection mode

---

## Files Changed

**`livekit-agents/livekit/agents/voice/audio_recognition.py`**
- Removed the buggy `self._last_speaking_time = time.time()` line from END_OF_SPEECH handler
- Added explanatory comment documenting why we don't update the timestamp here

---

## Related Issues

- Issue #4325: min_endpointing_delay behavior differences between VAD and STT modes (related timing inconsistency)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of initial speech timestamps, resulting in more accurate latency measurements and more reliable end-of-turn detection during voice recognition.
  * Reduces premature or inconsistent end-of-utterance decisions, especially when voice activity detection is active.
  * No other user-visible behavior changes beyond improved timing and reliability.

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -373,7 +373,7 @@ class AudioRecognition:
             self._audio_preflight_transcript = ""
             self._final_transcript_received.set()
 
-            if not self._vad or self._last_speaking_time == 0:
+            if not self._vad or self._last_speaking_time is None:
                 # vad disabled, use stt timestamp
                 # TODO: this would screw up transcription latency metrics
                 # but we'll live with it for now.
@@ -425,7 +425,7 @@ class AudioRecognition:
             self._audio_preflight_transcript = (self._audio_transcript + " " + transcript).lstrip()
             self._audio_interim_transcript = transcript
 
-            if not self._vad or self._last_speaking_time == 0:
+            if not self._vad or self._last_speaking_time is None:
                 # vad disabled, use stt timestamp
                 self._last_speaking_time = time.time()
 
@@ -449,7 +449,8 @@ class AudioRecognition:
 
             self._speaking = False
             self._user_turn_committed = True
-            self._last_speaking_time = time.time()
+            if not self._vad or self._last_speaking_time is None:
+                self._last_speaking_time = time.time()
 
             chat_ctx = self._hooks.retrieve_chat_ctx().copy()
             self._run_eou_detection(chat_ctx)
--- a/uv.lock
+++ b/uv.lock
@@ -2358,14 +2358,12 @@ name = "livekit-plugins-mistralai"
 source = { editable = "livekit-plugins/livekit-plugins-mistralai" }
 dependencies = [
     { name = "livekit-agents" },
-    { name = "livekit-plugins-openai" },
     { name = "mistralai" },
 ]
 
 [package.metadata]
 requires-dist = [
     { name = "livekit-agents", editable = "livekit-agents" },
-    { name = "livekit-plugins-openai", editable = "livekit-plugins/livekit-plugins-openai" },
     { name = "mistralai", specifier = ">=1.9.11" },
 ]
 


davidzhao:
@codex review

davidzhao (commented):
@devbyteai please remove PR_DESCRIPTION.md from the commit

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -449,7 +449,9 @@ async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
 
             self._speaking = False
             self._user_turn_committed = True
-            self._last_speaking_time = time.time()
+            # Note: Don't update _last_speaking_time here - it was already set at START_OF_SPEECH.
+            # Updating it here would make transcription_delay calculation incorrect (~0) since
+            # END_OF_SPEECH typically arrives after FINAL_TRANSCRIPT in STT mode.
```
longcw: this is true, when in stt mode the `END_OF_SPEECH` may arrive right ~before~ after the `FINAL_TRANSCRIPT` so the transcription delay is incorrect.

> Remove the line that overwrites _last_speaking_time at END_OF_SPEECH in STT mode. The value was already correctly set at START_OF_SPEECH.

but I don't think using the time of `START_OF_SPEECH` as the `last_speaking_time` is the right way, it's the start time but not the last speaking time. similar to `FINAL_TRANSCRIPT` event, I think a workaround for now is to only update the last speaking time if VAD is disabled 

```suggestion
            if not self._vad or self._last_speaking_time == 0:
                self._last_speaking_time = time.time()
```

davidzhao (approved):
lg!

PR #4404, opened 2025-12-27
fix (mistral-ai): add flexibility for timestamps 

tinalenguyen:
`timestamp_granularities` and `language` can't be used in tandem. to use one, the other must be set to `None`

### DIFF
--- a/livekit-plugins/livekit-plugins-mistralai/livekit/plugins/mistralai/stt.py
+++ b/livekit-plugins/livekit-plugins-mistralai/livekit/plugins/mistralai/stt.py
@@ -130,7 +130,7 @@ class STT(stt.STT):
                 alternatives=[
                     stt.SpeechData(
                         text=resp.text,
-                        language=self._opts.language,
+                        language=self._opts.language if self._opts.language else "",
                         start_time=resp.segments[0].start if resp.segments else 0,
                         end_time=resp.segments[-1].end if resp.segments else 0,
                         words=[


chenghao-mou (approved):
Thanks for spotting that!

I tested it locally, and it returned timestamps. However, I don't think we can turn the flag on because it is still non-streaming, and therefore only relative to each VAD spans.

```
STT event: SpeechEvent(type=<SpeechEventType.FINAL_TRANSCRIPT: 'final_transcript'>, request_id='', alternatives=[SpeechData(language=None,        
                                           text=' Can you tell me a different story?', start_time=0.3, end_time=1.9, confidence=0.0, speaker_id=None, is_primary_speaker=None, words=[' Can  
                                           you tell me a different story?'])], recognition_usage=None) 
...
STT event: SpeechEvent(type=<SpeechEventType.FINAL_TRANSCRIPT: 'final_transcript'>, request_id='', alternatives=[SpeechData(language=None,        
                                           text=' You know.', start_time=0.3, end_time=0.7, confidence=0.0, speaker_id=None, is_primary_speaker=None, words=[' You know.'])],                
                                           recognition_usage=None) 
```

tinalenguyen:
@chenghao-mou Ah okay, thanks for the clarification! It's always the case that `aligned_transcript` is set if `streaming` is `True`, right?

ChenghaoMou:
> @chenghao-mou Ah okay, thanks for the clarification! It's always the case that `aligned_transcript` is set if `streaming` is `True`, right?

Yes. If it supports streaming and can return word or chunk timestamps, otherwise it should always be False.

Issue #4409, opened 2025-12-29
Broken streaming tts retries


jayeshp19:
### Bug Description

it looks like streaming TTS retries are broken. The retry loop reuses the same SynthesizeStream instance, but the _input_ch channel and plugin state are one-shot resources that get consumed during the first attempt.

As a result, retrying the same request sends no text or produces no segments, leading to consistent failures like APIError: no audio frames were pushed or segments mismatch.

related issues: https://github.com/livekit/agents/issues/4135 

### Expected Behavior

we should be able to replay events during retry

### Reproduction Steps

```bash
1. 
2.
3.
...
- Sample code snippet, or a GitHub Gist link -
```

### Operating System

Macos

### Package Versions

```bash
livekit-agents=1.3.10
```

### Proposed Solution

```python

```

longcw:
with the current version, I think the retry only works when the connection establishing failed. it's not changed since the very beginning so it's hard to say it's "broken". 

I am not sure if this behavior is by design, but there is one issue I can see if we allow input replaying is that it may generate repeated content after failed in the middle of generation. cc @theomonnom wdyt?

btw, I think it's a different issue if the TTS only fails after the connection created and input sent (maybe 11labs with the multi-context TTS?), we should fix that plugin specifically.

jayeshp19:
> I am not sure if this behavior is by design, but there is one issue I can see if we allow input replaying is that it may generate repeated content after failed in the middle of generation. 

added a guard to prevent duplicate speech if a failure happens mid-generation - https://github.com/livekit/agents/pull/4410/changes/6ab2b6c35ef2d6dd3b91ffbb1fabf2f802089987

PR #4410, opened 2025-12-29
fix streaming tts retries

jayeshp19:
fixes: https://github.com/livekit/agents/issues/4409

## Summary by CodeRabbit

* **Bug Fixes**
  * Smarter TTS retry handling with clearer recoverable vs non-recoverable outcomes and detection of missing audio to avoid silent failures.
  * Improved finalization and cleanup across providers to prevent hangs, resource leaks, and missed audio delivery.

* **Improvements**
  * Input buffering and replay so streamed text/audio survive retries.
  * Per-run scoped streaming resources for better concurrency isolation and safer connection handling.

### DIFF
--- a/livekit-agents/livekit/agents/tts/tts.py
+++ b/livekit-agents/livekit/agents/tts/tts.py
@@ -600,7 +600,8 @@ class SynthesizeStream(ABC):
         """Close ths stream immediately"""
         await aio.cancel_and_wait(self._task)
         self._event_ch.close()
-        self._input_ch.close()
+        with self._input_lock:
+            self._input_ch.close()
 
         if self._metrics_task is not None:
             await self._metrics_task


livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -542,13 +559,30 @@ def flush(self) -> None:
...
+        self._started_time = 0
+
+        ch = aio.Chan[Union[str, SynthesizeStream._FlushSentinel]]()
+        for ev in self._replay_events:
+            ch.send_nowait(ev)
+
+        if self._input_ended:
+            ch.close()
+
+        self._input_ch = ch
```
longcw: close the previous input_ch before replacing?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -709,10 +720,24 @@ async def _recv_loop(self) -> None:
...
                         ctx.timeout_timer.cancel()
 
                 if data.get("isFinal"):
+                    if not ctx.received_audio and not ctx.waiter.done():
+                        # ElevenLabs sometimes returns `isFinal` with an empty `audio` payload.
+                        # Treat that as a retryable failure and force a reconnect for the next attempt.
+                        ctx.waiter.set_exception(
+                            APIError("11labs stream ended without audio", retryable=True)
+                        )
+                        self.mark_non_current()
```
longcw, line 740: why `mark_non_current` is needed when one of the generation failed?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -709,10 +720,24 @@ async def _recv_loop(self) -> None:
                 if data.get("audio"):
                     b64data = base64.b64decode(data["audio"])
                     emitter.push(b64data)
+                    ctx.received_audio = True
                     if ctx.timeout_timer:
                         ctx.timeout_timer.cancel()
 
                 if data.get("isFinal"):
+                    if not ctx.received_audio and not ctx.waiter.done():
+                        # ElevenLabs sometimes returns `isFinal` with an empty `audio` payload.
```
longcw, line 734: is there a valid case that elevenlabs returns final without audio, like when the pushed text is empty?

livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
```diff
@@ -455,7 +455,7 @@ def __init__(self, *, tts: TTS, conn_options: APIConnectOptions):
         super().__init__(tts=tts, conn_options=conn_options)
         self._tts: TTS = tts
         self._opts = replace(tts._opts)
-        self._segments_ch = utils.aio.Chan[tokenize.SentenceStream]()
+        self._segments_ch: utils.aio.Chan[tokenize.SentenceStream] | None = None
```
longcw: why this `self._segments_ch` is reserved?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
jayeshp19: It just ensures the next attempt gets a fresh connection

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
jayeshp19: Yes, empty/whitespace input returns isFinal with audio: null. I added a sent_text guard so we only error when real text was sent

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -509,51 +527,81 @@ def _emit_metrics() -> None:
 
     def push_text(self, token: str) -> None:
         """Push some text to be synthesized"""
-        if not token or self._input_ch.closed:
+        if not token:
             return
+        with self._input_lock:
```
longcw, line 532: `push_text` is a sync method, any reason a lock is needed? I don't think we are going to push text from different threads, right?

PR #4412, opened 2025-12-30
Inference: Improved support for mid session TTS updates

adrian-cowham:


livekit-agents/livekit/agents/inference/tts.py
chenghao-mou, line 447: I think we need to 1. add `update_options` to the stream and 2. add stream opts update here so changes are propagated, similar to what we already do in stt.py.

adrian-cowham: If I recall correctly, our approach is *not* to update TTS params mid stream. This is how the other TTS plugins behave, correct me if I'm wrong.

chenghao-mou: I might have missed some context here. Just double-checked, you are right that we don't have any tts mid-stream updates, probably because they don't support that I think. (Could be a cool feature to support in Inference though)

But if we are not updating them mid-stream, then we don't really need to send these parameters every sentence, right? I would assume we only need to send it once during the WebSocket initialization when the stream is created. Or is this more of a keeping-it-redundant/stateless thing so it's easier to track in Inference?

adrian-cowham: Yeah, I thought that same thing but given that we're using a connection pool, the connection is reused. That is, a new connection to Inference isn't guaranteed.

chenghao-mou: Right.

We could do this:

```diff
--- a/livekit-agents/livekit/agents/inference/tts.py
+++ b/livekit-agents/livekit/agents/inference/tts.py
@@ -445,6 +445,9 @@ class TTS(tts.TTS):
         if is_given(extra_kwargs):
             self._opts.extra_kwargs.update(extra_kwargs)
 
+        if is_given(model) or is_given(voice) or is_given(language):
+            self._pool.invalidate()
+
```

but I am not sure if it's faster. Or we can abstract the first message into its own function and recall it in the `_sentence_stream_task` task. I will create a separate PR for refactoring, and approve this for now.

PR #4419, opened 2025-12-30
fix: return 503 health check when worker fails to connect to LiveKit

rusg77:
We faced a network issue in our k8s cluster where agents were unable to connect to LiveKit servers. After exhausting all retry attempts, the health check still returned 200, so pods remained running in a broken state.

This change returns 503 from the health check when connection retries are exhausted, allowing Kubernetes to restart the pods.


davidzhao (approved):
this makes sense.. thanks for the PR!

PR #4424, opened 2025-12-31
Add extra comments about Google model deprecation

chenghao-mou:
Tested the ambiguous ones: "gemini-2.0-flash-exp" and "gemini-live-2.5-flash-preview-native-audio". Both still work with the right setting, but are not mentioned in any official docs or change logs.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/api_proto.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/api_proto.py
@@ -5,17 +5,21 @@ from typing import Literal, Union
 
 from google.genai import types
 
+# Gemini API deprecations: https://ai.google.dev/gemini-api/docs/deprecations
+# Gemini API release notes with preview deprecations: https://ai.google.dev/gemini-api/docs/changelog
+# live models: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/live-api
+# VertexAI retirement: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions#retired-models
+# Additional references:
+# 1. https://github.com/kazunori279/adk-streaming-test/blob/main/test_report.md
 LiveAPIModels = Literal[
     # VertexAI models
-    "gemini-live-2.5-flash-native-audio",
-    "gemini-live-2.5-flash-preview-native-audio",
-    # deprecated vertexai models
-    "gemini-2.0-flash-exp",
-    "gemini-live-2.5-flash-preview-native-audio-09-2025",
+    "gemini-live-2.5-flash-native-audio",  # GA https://docs.cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-flash-live-api#live-2.5-flash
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/api_proto.py
```diff
@@ -5,17 +5,21 @@
...
 LiveAPIModels = Literal[
     # VertexAI models
-    "gemini-live-2.5-flash-native-audio",
-    "gemini-live-2.5-flash-preview-native-audio",
-    # deprecated vertexai models
-    "gemini-2.0-flash-exp",
-    "gemini-live-2.5-flash-preview-native-audio-09-2025",
+    "gemini-live-2.5-flash-native-audio",  # GA https://docs.cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-flash-live-api#live-2.5-flash
+    "gemini-live-2.5-flash-preview-native-audio-09-2025",  # Public preview https://docs.cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-flash-live-api#live-2.5-flash-preview
+    "gemini-live-2.5-flash-preview-native-audio",  # still works, possibly an alias, but not mentioned in any docs or changelog
```
davidzhao, line 18: I think we should remove from the list if it's not mentioned on their docs

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/api_proto.py
```diff
@@ -5,17 +5,21 @@
...
     # Gemini API models
-    "gemini-2.5-flash-native-audio-preview-12-2025",
-    # deprecated Gemini API models
-    "gemini-2.5-flash-native-audio-preview-09-2025",
+    "gemini-2.5-flash-native-audio-preview-12-2025",  # https://ai.google.dev/gemini-api/docs/models#gemini-2.5-flash-live
+    "gemini-2.5-flash-native-audio-preview-09-2025",  # https://ai.google.dev/gemini-api/docs/models#gemini-2.5-flash-live
+    "gemini-2.0-flash-exp",  # still works in Gemini API but not VertexAI
```
davidzhao, line 22: I've received deprecation notice for this one, but it's possible that was for VertexAI.

davidzhao (approved):

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/api_proto.py
chenghao-mou: It is probably fine to keep this one. It looks like an alias to the dated model following their naming convention, and their docs aren't always updated.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/api_proto.py
chenghao-mou: I can only find a deprecation notice from a third-party website:
Screenshot 2026-01-02 at 09 09 29

and apparently it's still very much alive on both platforms via [OpenRouter](https://openrouter.ai/google/gemini-2.0-flash-exp:free/performance)

We can still deprecate it on our side though.

PR #4425, opened 2025-12-31
Update STT tests and add batch recognition flag

chenghao-mou:
- Add batch recognition flag in STT capabilities
- Added manual workflow to test a PR/branch/revision
- Updated tests to support all STT vendors except two of them:
    - clova
	- spitch
	- rtzr

### DIFF
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -80,7 +80,7 @@ class STTCapabilities:
     interim_results: bool
     diarization: bool = False
     aligned_transcript: Literal["word", "chunk", False] = False
-    batch_recognition: bool = True
+    offline_recognize: bool = True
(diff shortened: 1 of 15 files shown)


chenghao-mou:
AWS's client seems to leak tasks, so it always fails now in our tests.

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -80,6 +80,8 @@ class STTCapabilities:
     interim_results: bool
     diarization: bool = False
     aligned_transcript: Literal["word", "chunk", False] = False
+    batch_recognition: bool = True
```
theomonnom:
```suggestion
    offline_recognition: bool = True
```

theomonnom: Or just offline

PR #4428, opened 2026-01-02
fix (gemini filesearch): require only filestore names

tinalenguyen:
error will be thrown if `file_search_store_names` is not provided, rest are optional


chenghao-mou (approved):
lgtm. It is interesting that they report the error with a 400 instead of a client-side one.

Issue #4432, opened 2026-01-02
Allow cancellation even when allow_interruptions is False


chenghao-mou:
### Feature Type

Nice to have

### Feature Description

This stems from the conversations in #4418 and #4413.

The core issue is that when we create a speech that has `allow_interruptions` set to False, we don't have a way to reliably stop or cancel it **before** it's played, even if the customer continues to speak after it's being scheduled. This also makes the cancellation of a tool call with `say/generate_reply` invoked indeterministic, esp since the introduction of #4102 

Scenario 1: (current behavior)
- user says something
- agent speech (`allow_interruptions=False`) scheduled
- user continues speaking, preventing the agent speech from being played
- agent speech plays when the user stops speaking, but new user input is skipped

In this case, the agent would appear missing input

Scenario 2: (old behavior)
- user says something
- agent speech (`allow_interruptions=False`) scheduled
- user continues speaking
- agent speech plays, overlapping with user's speech

Here, the agent would appear "impatient"

Scenario 3: (ideal behavior)
- user says something
- agent speech (`allow_interruptions=False`) scheduled
- user continues speaking, leading to cancellation of the scheduled agent speech
- new agent speech (`allow_interruptions=False`) scheduled and then played

### Workarounds / Alternatives

- restore old behavior by checking speech handle flags; or
- introduce new ways of speech cancellation

davidzhao:
IMO, allow_interruptions should only apply to user interruptions.. if the user speaks *before* the agent starts speaking, then we should not consider it interruptions.

also if the developer uses `interrupt()`, it should always be allowed..

longcw:
I think `Scenario 2: (old behavior)` makes more sense. when a developer schedule a `uninterruptible` speech I don't think they are expecting it will be cancelled by user speech in any case. If they really want it, they can call the `session.interrupt(force=True)` in a event callback when user state changed to `speaking`.

for the scenario 3, it's just postponed the speech, and if the speech was from `session.say`, the content won't be updated no matter what user said which feels like the agent ignored the user's input.

chenghao-mou:
I created a PR #4434 to restore the old one for now.

In the examples, scenario 1 is intended for the delaying case, and scenario 3 would have a new user transcript, thus a new "turn". 

Do we expect the same "expectations" when the speech is scheduled from both locations?
1. normal agent turns, internally calling generate_reply
2. user explicitly calls generate_reply or say

[...]

PR #4434, opened 2026-01-03
restore old behavior

chenghao-mou:
This temporarily restores the old behavior until we settle the discussion in #4432

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1616,19 +1616,18 @@ class AgentActivity(RecognitionHooks):
         )
         audio_output = self._session.output.audio if self._session.output.audio_enabled else None
 
-        wait_for_authorization = asyncio.ensure_future(speech_handle._wait_for_authorization())
         # See discussion in https://github.com/livekit/agents/issues/4432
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1617,7 +1617,12 @@ async def _tts_task_impl(
         audio_output = self._session.output.audio if self._session.output.audio_enabled else None
 
         wait_for_authorization = asyncio.ensure_future(speech_handle._wait_for_authorization())
-        wait_for_user_silence = asyncio.ensure_future(self._user_silence_event.wait())
+        # See discussion in https://github.com/livekit/agents/issues/4432
+        wait_for_user_silence = (
```
longcw: I would suggest to use a list of tasks instead of creating an empty one
```python
        authorization_tasks = [asyncio.ensure_future(speech_handle._wait_for_authorization())]
        if speech_handle.allow_interruptions:
            authorization_tasks.append(asyncio.ensure_future(self._user_silence_event.wait()))
        await speech_handle.wait_if_not_interrupted(authorization_tasks)
```

longcw (approved):
lgtm!

PR #4436, opened 2026-01-03
fix(voice): reset capturing flag in clear_buffer() to fix segment counting

martin-purplefish:
## Summary
- `AudioOutput.clear_buffer()` was not resetting the internal `__capturing` flag, unlike `flush()`
- This caused the playback segment counter to become desynchronized after interruptions
- Result: `wait_for_playout()` returns prematurely or spurious warnings about mismatched playback finished events

## Changes
- Add `self.__capturing = False` to base class `clear_buffer()` in `io.py`
- Add `super().clear_buffer()` calls to all implementations to invoke base class logic (matching the pattern used by `flush()`)

## Test plan
- [x] Verified the fix aligns with the docstring stating both `flush()` and `clear_buffer()` segment playback
- [x] Manual testing with voice agent interruptions

### DIFF
--- a/examples/Dockerfile-example
+++ b/examples/Dockerfile-example
@@ -27,6 +27,7 @@ RUN apt-get update && \
     gcc \
     g++ \
     python3-dev \
+    libglib2.0-0 \
     && rm -rf /var/lib/apt/lists/*
 
 USER appuser
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -179,6 +179,7 @@ class ConsoleAudioOutput(io.AudioOutput):
             self._flush_task = asyncio.create_task(self._wait_for_playout())
 
     def clear_buffer(self) -> None:
+        super().clear_buffer()
         with self._audio_lock:
             self._output_buf.clear()
             self._output_buf_empty.set()
--- a/livekit-agents/livekit/agents/inference/tts.py
+++ b/livekit-agents/livekit/agents/inference/tts.py
@@ -502,6 +502,14 @@ class SynthesizeStream(tts.SynthesizeStream):
             async for ev in sent_tokenizer_stream:
                 token_pkt = base_pkt.copy()
                 token_pkt["transcript"] = ev.token + " "
+                generation_config: dict[str, Any] = {}
+                if self._opts.voice:
+                    generation_config["voice"] = self._opts.voice
+                if self._opts.model:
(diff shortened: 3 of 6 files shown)


martin-purplefish:
This tracks with an issue we've been seeing where we get seemingly random ~5-10s delays on interruptions. I'm less familiar with this part of the code however, so while it looks correct to me (and I've tested locally), I'm not sure what the intent of clear_buffer() is.

livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
```diff
@@ -151,6 +151,7 @@ def flush(self) -> None:
         self._stream_writer = None
```
chenghao-mou, line 151: I think two more changes are missing:

- livekit-agents/livekit/agents/cli/cli.py with `ConsoleAudioOutput`
- livekit-agents/livekit/agents/voice/transcription/synchronizer.py with `_SyncedAudioOutput`

Both of them are `AudioOutput` subclasses, so we should include them.

martin-purplefish: Cool - should we change these from abstract methods to just base methods + overrides?

longcw:
do we really need this? we rest the flag when flush is called and it's always called for each segment (ensued by finally). I think it needs to be careful to check if the change has any side effects.

also, the capturing is stopped after the perform audio forwarding is done in agent activity, and after that the flush is called. Clear buffer can be called at any time when interrupting the agent, and when it's called the forwarding may haven't been stopped yet. it shouldn't rest the capturing flag from the design IMO.

martin-purplefish:
> do we really need this? we rest the flag when flush is called and it's always called for each segment (ensued by finally). I think it needs to be careful to check if the change has any side effects.

Fair concern. I think you're right - it looks like flush() is always called. But I'm reacting to the wait_for_playout check finished_count >= target which it feels like if flush wasn't called, wait_for_playout would return immediately the second time if I'm reading it right?

Alternatively, is this the wrong contract for clear_buffer()? "Clear the buffer, stopping playback immediately" - wouldn't we need to stop capturing?

And I'm a bit confused by the docstring on on_playback_finished - which seems to suggest we should stop capturing?

>         Developers building audio sinks must call this method when a playback/segment is finished.
>         Segments are segmented by calls to flush() or clear_buffer()

martin-purplefish:
In any case - let's close for now, we can revisit since it seems in practice flush() is called.

PR #4448, opened 2026-01-05
fix: misleading conn_options handling in StreamAdapters

darshankparmar:
StreamAdapter.stream(conn_options=…) previously ignored caller-provided options for the adapter’s own connection, always overriding them with hardcoded defaults.

### DIFF
--- a/livekit-agents/livekit/agents/stt/stream_adapter.py
+++ b/livekit-agents/livekit/agents/stt/stream_adapter.py
@@ -9,11 +9,6 @@ from ..types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, APIConnectOptions, N
 from ..vad import VAD, VADEventType
 from .stt import STT, RecognizeStream, SpeechEvent, SpeechEventType, STTCapabilities
 
-# already a retry mechanism in STT.recognize, don't retry in stream adapter
-DEFAULT_STREAM_ADAPTER_API_CONNECT_OPTIONS = APIConnectOptions(
-    max_retry=0, timeout=DEFAULT_API_CONNECT_OPTIONS.timeout
-)
-
 
 class StreamAdapter(STT):
     def __init__(self, *, stt: STT, vad: VAD) -> None:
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/stt/stream_adapter.py
```diff
@@ -9,11 +9,6 @@
 from ..vad import VAD, VADEventType
 from .stt import STT, RecognizeStream, SpeechEvent, SpeechEventType, STTCapabilities
 
-# already a retry mechanism in STT.recognize, don't retry in stream adapter
-DEFAULT_STREAM_ADAPTER_API_CONNECT_OPTIONS = APIConnectOptions(
```
longcw, line 13: like mentioned in the comment, `conn_options` of `stt.stream()` is passed to `stt.recognize` and there is already a retry mechanism in `stt.recognize`. there is no need to retry in stream adapter.

darshankparmar: agreed that retries should stay in `stt.recognize()` and not in the stream adapter.

The issue here isn’t retry duplication, but **API semantics**:
`stt.stream(conn_options=...)` accepts options that are silently overridden for the stream wrapper itself. This is surprising, since callers expect the provided options (e.g. timeout) to apply to the stream connection.

The fix keeps `max_retry=0` in the adapter, but either:

* respects caller-provided fields like `timeout`, or
* removes/renames `conn_options` to make the behavior explicit.

Happy to adjust based on preferred API direction.

longcw: the `conn_options` passed to `stt.stream` is respected in `stt.recognize ` which is where the actual connection made.

probably you missed the `self._wrapped_stt_conn_options = conn_options` and the `self._wrapped_stt_conn_options` is passed to `stt.recognize `. in this way the `timeout` is applied to the connection.

darshankparmar: Got it, thanks for pointing that out. you’re right, the `conn_options` passed to `stream()` are applied via `_wrapped_stt_conn_options` when calling `stt.recognize()`, so the timeout is respected.

My original concern was around API clarity rather than behavior. I’m okay to drop this change if you feel the current behavior is intended.

darshankparmar:
Closing since this is intended behavior and no code change is needed.

PR #4452, opened 2026-01-06
add connect CLI command

tinalenguyen:
`python agent.py connect --room ROOM_NAME --participant-identity PARTICIPANT_IDENTITY
`
- participant identity is optional
- connects to existing room if found, creates new room if not

### DIFF
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -1316,25 +1316,27 @@ def _run_worker(server: AgentServer, args: proto.CliArgs, jupyter: bool = False)
     @server.once("worker_started")
     def _simulate_job() -> None:
         async def simulate_job() -> None:
-            room_name = args.simulate_job.room
-            async with api.LiveKitAPI(args.url, args.api_key, args.api_secret) as lk_api:
-                room_request = api.ListRoomsRequest(names=[room_name])
-                active_room = await lk_api.room.list_rooms(room_request)
+            if args.simulate_job is not None:
+                room_name = args.simulate_job.room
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/cli/cli.py
```diff
@@ -1293,6 +1313,31 @@ async def _worker_run(worker: AgentServer) -> None:
...
+    def _simulate_job() -> None:
+        async def simulate_job() -> None:
+            if args.simulate_job is not None:
+                room_name = args.simulate_job.room
+                async with api.LiveKitAPI(args.url, args.api_key, args.api_secret) as lk_api:
+                    room_request = api.ListRoomsRequest(names=[room_name])
+                    active_room = await lk_api.room.list_rooms(room_request)
+
+                    if not active_room.rooms:
+                        room_info = await lk_api.room.create_room(
```
davidzhao, line 1326: is the explicit create request needed? I think we should just connect the agent directly to the room?

livekit-agents/livekit/agents/cli/cli.py
```diff
@@ -1522,6 +1570,59 @@ async def _run_loop() -> None:
             logger.warning("exiting forcefully")
             os._exit(1)
 
+    @app.command()
+    def connect(
```
davidzhao, line 1574: when we are running `connect`, it should not register itself as a worker. this is to ensure that one could use it as a simple command line connector, where the likely use case is to connect to an existing room (and if a room doesn't already exist.. autoconnect behavior should help to recreate it)

tinalenguyen: ah, good to know there is autoconnect behavior, also wasn't sure what the previous usage for `connect` looked like. i'll make a change

PR #4454, opened 2026-01-06
Add DeepL translation plugin for multilingual agent support

darshankparmar:
Add a DeepL translation plugin to enable real-time, bidirectional multilingual support for LiveKit agents. The plugin integrates seamlessly into the agent pipeline, acting as a translation layer between STT/LLM and LLM/TTS components.

## Supported Languages

[DeepL API documentation](https://developers.deepl.com/docs/getting-started/supported-languages)

## Dependencies

- `deepl>=1.27.0` : Python library for the DeepL API.


darshankparmar:
Hi @davidzhao 👋  
This PR is ready for review.  
Please let me know if you’d like any changes or additional tests.  
Thanks!

livekit-plugins/livekit-plugins-deepl/livekit/plugins/deepl/translator.py
```diff
@@ -0,0 +1,154 @@
...
+
+        if source_lang == target_lang:
+            return [text] if isinstance(text, str) else list(text)
+
+        # The deepl library's translate_text is synchronous.
+        # We use asyncio.to_thread to run it in a separate thread pool
+        # to avoid blocking the event loop.
+        try:
+            result = await asyncio.to_thread(
+                self._translator.translate_text,
```
longcw, line 119: I don't think it's worth to be a plugin if it's just a wrap to the deepl's SDK. users can call their SDK directly.

darshankparmar: Thanks for the feedback.

The main value of the plugin isn’t just wrapping the DeepL SDK, but adapting it to the LiveKit agent runtime:
- DeepL’s SDK is synchronous, while agents require async, non-blocking execution
- The plugin integrates translation directly into the agent lifecycle (e.g. user turns, LLM flow) without users wiring it themselves

That said, I’m open to reducing scope or repositioning it (e.g. example / optional integration) if that fits better with the project direction.

PR #4457, opened 2026-01-06
Adding model query param to the STT and TTS websocket connection string.

adrian-cowham:
This allows inference to preserve existing behavior when enforcing per model rates limits/quotas. That is, enforcement is done at time of connection.


chenghao-mou:
/test-stt

I had to try this command. It didn't work 🤷

But changes look good to me. QQ about quota: for fallbacks, the quota/limit is enforced inside Inference, right?

chenghao-mou (approved):

PR #4494, opened 2026-01-11
update examples to use LK Inference

davidzhao:
also added a synthesize method for TTS plugins that are streaming-only

### DIFF
--- a/examples/drive-thru/test_agent.py
+++ b/examples/drive-thru/test_agent.py
@@ -121,7 +121,10 @@ async def test_failure() -> None:
             await (
                 result.expect.next_event()
                 .is_message(role="assistant")
-                .judge(judge_llm, intent="should inform the user that an error occurred")
+                .judge(
+                    judge_llm,
+                    intent="should inform the user that something went wrong, it's ok to ask them to try again",
+                )
             )
 
             # leaving this commented, some LLMs may occasionally try to retry.


examples/voice_agents/basic_agent.py
```diff
@@ -83,13 +84,13 @@ async def entrypoint(ctx: JobContext):
     session = AgentSession(
         # Speech-to-text (STT) is your agent's ears, turning the user's speech into text that the LLM can understand
         # See all available models at https://docs.livekit.io/agents/models/stt/
-        stt="deepgram/nova-3",
+        stt=inference.STT("deepgram/nova-3", language="multi"),
         # A Large Language Model (LLM) is your agent's brain, processing user input and generating a response
         # See all available models at https://docs.livekit.io/agents/models/llm/
-        llm="openai/gpt-4.1-mini",
+        llm=inference.LLM("openai/gpt-4.1-mini"),
```
longcw, line 90: do we prefer `inference.LLM("openai/gpt-4.1-mini")` than `"openai/gpt-4.1-mini"`?

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -343,6 +358,103 @@ async def __aexit__(
...
+    """
+
+    def __init__(
+        self,
+        *,
+        tts: TTS,
+        input_text: str,
+        conn_options: APIConnectOptions,
+    ) -> None:
+        # Don't call super().__init__ yet - we'll handle initialization differently
```
longcw: why it needs a custom `__init__` method?

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -343,6 +358,103 @@ async def __aexit__(
...
+                await self._main_task()
+
+        self._synthesize_task = asyncio.create_task(
+            _traceable_main_task(), name="TTS._synthesize_task"
+        )
+        self._synthesize_task.add_done_callback(lambda _: self._event_ch.close())
+        self._tts_request_span: trace.Span | None = None
+
+    async def _main_task(self) -> None:
+        """Override main task to directly forward SynthesizedAudio events from stream"""
```
longcw: do we really need to directly forward the events? Use the `_run` method will make the implementation simpler
```python
class ChunkedStreamFromStreaming(ChunkedStream):
    _tts_request_span_name: ClassVar[str] = "tts_synthesize_from_streaming"

    def __init__(self, *, tts: TTS, input_text: str, conn_options: APIConnectOptions) -> None:
        super().__init__(tts=tts, input_text=input_text, conn_options=conn_options)
        if not self._tts.capabilities.streaming:
            raise ValueError("TTS does not support streaming")

    async def _run(self, output_emitter: AudioEmitter) -> None:
        output_emitter.initialize(
            request_id=shortuuid(),
            sample_rate=self._tts.sample_rate,
            num_channels=self._tts.num_channels,
            mime_type="audio/pcm",
            stream=False
        )
        async with self._tts.stream(
            conn_options=APIConnectOptions(max_retry=0, timeout=self._conn_options.timeout)
        ) as stream:
            stream.push_text(self._input_text)
            stream.end_input()
            async for ev in stream:
                output_emitter.push(ev.frame.data.tobytes())
                if timed_transcripts := ev.frame.userdata.get(USERDATA_TIMED_TRANSCRIPT):
                    output_emitter.push_timed_transcript(timed_transcripts)

        output_emitter.flush()
```

examples/voice_agents/basic_agent.py
davidzhao: yes. because using pure strings make it difficult to know how to pass in parameters, or where to look at references.

this way a user could cmd + click into the LLM definition to see what they can pass in

livekit-agents/livekit/agents/tts/tts.py
davidzhao: good catch.. initially I took a diff path, the current functions are fairly similar

livekit-agents/livekit/agents/tts/tts.py
davidzhao: yeah, this is simpler.. claude wrote the previous code.

longcw (approved):
lgtm!

Issue #4500, opened 2026-01-12
Race condition in _send_kill_signal: ValueError: process object is closed after SIGUSR1


rafallezanko:
### Bug Description

A race condition in `supervised_proc.py` causes a `ValueError: process object is closed` exception during process termination. The error occurs in `_send_kill_signal()` when the process closes itself after receiving SIGUSR1 but before `self._proc.kill()` is called.

Have a look for a stacktrace:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/asyncio/tasks.py", line 507, in wait_for
    return await fut
           ^^^^^^^^^
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/.venv/lib/python3.13/site-packages/livekit/agents/ipc/supervised_proc.py", line 252, in aclose
    await asyncio.wait_for(
        asyncio.shield(self._supervise_atask), timeout=self._opts.close_timeout
    )
  File "/usr/local/lib/python3.13/asyncio/tasks.py", line 506, in wait_for
    async with timeouts.timeout(timeout):
               ~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/local/lib/python3.13/asyncio/timeouts.py", line 116, in __aexit__
    raise TimeoutError from exc_val
TimeoutError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/.venv/lib/python3.13/site-packages/livekit/agents/worker.py", line 1261, in _handle_termination
    await proc.aclose()
  File "/app/.venv/lib/python3.13/site-packages/livekit/agents/ipc/supervised_proc.py", line 261, in aclose
    await self._send_kill_signal()
  File "/app/.venv/lib/python3.13/site-packages/livekit/agents/ipc/supervised_proc.py", line 317, in _send_kill_signal
    self._proc.kill()
    ~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/multiprocessing/process.py", line 139, in kill
    self._check_closed()
    ~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/multiprocessing/process.py", line 101, in _check_closed
    raise ValueError("process object is closed")
ValueError: process object is closed
```

### Expected Behavior

The `_send_kill_signal()` method should handle the case where the process closes between the `is_alive()` check and the `kill()` call without raising an exception.

### Reproduction Steps

```bash
[...]

PR #4501, opened 2026-01-12
#4500 Fix for Race condition in _send_kill_signal: ValueError: process object is closed after SIGUSR1

rafallezanko:
The commit solves issue #4500 -> passes ValueError exception when process is already killed/closed.


livekit-agents/livekit/agents/ipc/supervised_proc.py
```diff
@@ -305,7 +305,10 @@ async def _send_kill_signal(self) -> None:
 
         logger.info("killing process", extra=self.logging_extra())
         if sys.platform == "win32":
-            self._proc.terminate()
+            try:
+                self._proc.terminate()
```
chenghao-mou:
```suggestion
                if self._proc.is_alive():
                    self._proc.terminate()
```

livekit-agents/livekit/agents/ipc/supervised_proc.py
```diff
@@ -314,7 +317,10 @@ async def _send_kill_signal(self) -> None:
                     await asyncio.sleep(0.5)
                 except Exception:
                     pass
-            self._proc.kill()
+            try:
+                self._proc.kill()
```
chenghao-mou:
```suggestion
                if self._proc.is_alive():
                    self._proc.kill()
```

rafallezanko:
Right @chenghao-mou, thank you for your suggestion. Changes applied :)

chenghao-mou (approved):
LGTM. Thanks!

PR #4504, opened 2026-01-12
Support static context in integration with langchain

benlangfeld:
This is used to store data about an agent invocation which should be accessible to tools but not to the LLM, like API keys, user IDs etc.

See:
* https://docs.langchain.com/oss/python/langchain/context-engineering#runtime-context-6
* https://github.com/langchain-ai/langgraph/pull/5243

### DIFF
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -722,7 +722,9 @@ class FrequencyVisualizer:
 
         if self.show_shortcuts:
             for shortcut_key, desc in AUDIO_SHORTCUTS:
-                table.add_row(Text.assemble(("   ", ""), (shortcut_key, "dim bold"), (f"  {desc}", "dim")))
+                table.add_row(
+                    Text.assemble(("   ", ""), (shortcut_key, "dim bold"), (f"  {desc}", "dim"))
+                )
         else:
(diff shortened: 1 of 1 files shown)


davidzhao:
this looks good, could you roll back the uv.lock change and fix the types CI please?

benlangfeld:
> this looks good, could you roll back the uv.lock change and fix the types CI please?

I cannot roll back the uv.lock change, because this feature was introduced in langgraph 0.6.0.

I am trying to fix the type error, it's not clear to me why it's occurring right now. The dependency upgrade should have fixed it.

```
uv run python
Python 3.12.4 (main, Jul  1 2024, 18:16:41) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from typing import get_type_hints
>>> from langgraph.pregel.protocol import PregelProtocol
>>> get_type_hints(PregelProtocol.astream)['context']
typing.Optional[~ContextT]
```

The type error raised by mypy is wrong. I am not really a python developer, so I'd appreciate any help in figuring out why it might be happening.

bcdonadio:
@benlangfeld This should fix it: https://github.com/powerhome/agents/pull/3

benlangfeld:
@davidzhao This should be ready to go.

PR #4507, opened 2026-01-12
fix: reuse BytesIO buffer to avoid reallocation in StreamBuffer

darshankparmar:
This PR reduces memory churn in `StreamBuffer.read()` by reusing the existing `BytesIO` buffer instead of reallocating a new one after each read.

### DIFF
--- a/livekit-agents/livekit/agents/utils/codecs/decoder.py
+++ b/livekit-agents/livekit/agents/utils/codecs/decoder.py
@@ -98,7 +98,13 @@ class StreamBuffer:
                 if data:
                     # shrink the buffer to remove already-read data
                     remaining = self._buffer.read()
-                    self._buffer = io.BytesIO(remaining)
+
+                    # Reuse the same buffer instead of reallocating
+                    self._buffer.seek(0)
+                    self._buffer.truncate()
+                    self._buffer.write(remaining)
+                    self._buffer.seek(0)
+
                     return data
 
                 if self._eof:


livekit-agents/livekit/agents/utils/codecs/decoder.py
```diff
@@ -98,7 +98,13 @@ def read(self, size: int = -1) -> bytes:
                 if data:
                     # shrink the buffer to remove already-read data
                     remaining = self._buffer.read()
-                    self._buffer = io.BytesIO(remaining)
+
+                    # Reuse the same buffer instead of reallocating
+                    self._buffer.seek(0)
+                    self._buffer.truncate()
+                    self._buffer.write(remaining)
+                    self._buffer.seek(0)
```
davidzhao, line 106: I don't think this works.

would future bytes be written at 0? overwriting remaining? reading and writing are happening at the same position, which isn't what we want

darshankparmar: > I don't think this works.
> 
> would future bytes be written at 0? overwriting remaining? reading and writing are happening at the same position, which isn't what we want

Thanks for the review. After a deeper look, you’re right.

While the truncate-based approach reduces allocations, it relies on implicit BytesIO cursor behavior and isn’t robust enough for a concurrency primitive.
Alternative designs (e.g. ring buffers) would break PyAV’s file-like contract or require a larger redesign.

I’ll revert this change and keep the existing behavior for correctness.

PR #4513, opened 2026-01-13
feat(google): add warnings when system messages are dropped in Gemini realtime model

### LINKED ISSUE #4497: Context Injection via `update_chat_ctx()` Strips System Messages in Gemini RealtimeModel
### Bug Description

When using `google.realtime.RealtimeModel` with the LiveKit Agents framework, calling `update_chat_ctx()` to inject context mid-conversation **only works for user/assistant messages but silently drops all system/developer messages**. This prevents critical use cases like:

- Dynamic context injection (real-time data updates, sports scores, news)
- Proactive state management (user preferences, session state)

Injecting context as a **user message works**, but injecting as a **system message fails silently** (no error, the message simply disappears). There is a warning stating that removal is not supported, but appending system messages also appears not to be supported. 

### 1. System Messages Are Explicitly Filtered

In `agent_activity.py`, before forwarding context to the realtime session, instructions are removed:

```python
# livekit-agents/livekit/agents/voice/agent_activity.py:329-338
async def update_chat_ctx(
    self, chat_ctx: llm.ChatContext, *, exclude_invalid_function_calls: bool = True
) -> None:
    chat_ctx = chat_ctx.copy(tools=self.tools if exclude_invalid_function_calls else NOT_GIVEN)
    self._agent._chat_ctx = chat_ctx

        remove_instructions(chat_ctx)  # <-- System messages removed here!
        await self._rt_session.update_chat_ctx(chat_ctx)
```

### 2. Realtime Session Also Excludes Instructions

The Gemini realtime session strips instructions again:

```python
# livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py:473-511
async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
    chat_ctx = chat_ctx.copy(
        exclude_handoff=True,
        exclude_instructions=True,  # <-- Strips system/developer messages
        exclude_empty_message=True
    )
    
    # ... compute diff, then:
    
    if append_ctx.items:
        turns_dict, _ = append_ctx.copy(exclude_function_call=True).to_provider_format(
            format="google", inject_dummy_user_message=False
        )
        # ^ Note: extra_data (containing system messages) is discarded with `_`
        turns = [types.Content.model_validate(turn) for turn in turns_dict]
        if turns:
            self._send_client_event(types.LiveClientContent(turns=turns, turn_complete=False))
```

Why is the system message explicitly being filtered out? They are crucial for dynamic context injection and are critical for our use cases. This works perfectly fine for non realtime models. 

### Expected Behavior

Based on the LiveKit documentation and API design, `update_chat_ctx()` should allow injecting both user and system messages to update the model's context during an active conversation. System messages are critical for providing instructions and context that should influence model behavior without appearing as user input.

## Actual Behavior

1. ✅ **User/Assistant Messages**: Successfully injected and visible to the model
2. ❌ **System/Developer Messages**: Silently dropped, never reach the Gemini Live session
3. ⚠️ **Warning Log**: `"Gemini Live does not support removing messages"` appears (even though we're adding, not removing)

### Reproduction Steps

```bash
### Test Case 1: System Message Injection (FAILS)


from livekit.agents import Agent

class TestAgent(Agent):
    async def transcription_node(self, text, model_settings):
        async for delta in text:
            yield delta
        
        # Inject system message with secret keyword
        chat_ctx = self.chat_ctx.copy()
        chat_ctx.add_message(
            role="system",
            content="IMPORTANT: The secret keyword is BANANA"
        )
        await self.update_chat_ctx(chat_ctx)

# Run agent, connect, say "Hello"
# Then ask: "What is the secret keyword?"
# Result: Model responds "I don't know" ❌


### Test Case 2: User Message Injection (WORKS)


# Same setup, but inject as user message instead
chat_ctx.add_message(
    role="user",
    content="Remember: The secret keyword is BANANA"
)
await self.update_chat_ctx(chat_ctx)

# Ask: "What is the secret keyword?"
# Result: Model responds with "BANANA" ✅
```

### Operating System

macOS 15.6.1 (24G90)

### Models Used

gemini-2.5-flash-native-audio-preview-12-202515.6.1 (24G90)

### Package Versions

```bash
livekit-agents[anam,google]~=1.3
livekit-plugins-noise-cancellation>=0.2
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

pass the system messages instead of discarding it unless there is a clear reason why we are doing so. 

### Additional Context

This makes it impossible to build agents that require proactive state updates (e.g., "The user just logged in, here is their name", "The score of the game just changed"). We are forced to use Function Calling as a workaround, which is reactive (user must ask) rather than proactive.

_No response_

dhruvnigam93:
System/instructions messages added to chat_ctx were **silently** being dropped in Gemini realtime model since it does not support these roles. 

Add clear warning logs when system/developer messages are dropped by the Gemini Realtime API due to role constraints. The API only supports 'user' and 'model' roles.

Changes:
- Add _count_system_messages() helper method to count system/developer messages
- Add warning in update_chat_ctx() when system messages are present
- Add warning during session initialization when the initial context has system messages
- Warnings include model name, count of dropped messages, and guidance to use update_instructions()

Tests:
- Add tests suite in test_google_realtime_system_message_warnings.py
- Manual testing verified with real Gemini API

Fixes #4497


livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
longcw, line 505: I think it's better to check `append_ctx`, and it can be a single line so you don't need to add a method and the test case.
```python
        if any(
            item.type == "message" and item.role in ["system", "developer"]
            for item in append_ctx.items
        ):
            logger.warning("...")
```

dhruvnigam93: The messages are dropped much before this line inside the copy method. 
```python
    chat_ctx = chat_ctx.copy(
            exclude_handoff=True, exclude_instructions=True, exclude_empty_message=True
        )
``` 

To make this work I would have to copy with `exclude_instructions=False` and loop through all system messages too. I did not want to do that since it might have unintended consequences elsewhere. 

```python
 async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
        # Check for system/developer messages that will be dropped
        system_msg_count = self._count_system_messages(chat_ctx)
        if system_msg_count > 0:
            logger.warning(
                f"Gemini Realtime model '{self._opts.model}' does not support 'system' or "
                f"'developer' roles in chat history. Dropping {system_msg_count} system "
                f"message(s) from chat context. Gemini Realtime only supports 'user' and "
                f"'model' roles. Use update_instructions() to set system-level context instead."
            )

        chat_ctx = chat_ctx.copy(
            exclude_handoff=True, exclude_instructions=True, exclude_empty_message=True
        )
        async with self._session_lock:
            if not self._active_session:
                self._chat_ctx = chat_ctx
                return

        diff_ops = llm.utils.compute_chat_ctx_diff(self._chat_ctx, chat_ctx)

        if diff_ops.to_remove:
            logger.warning("Gemini Live does not support removing messages")

        append_ctx = llm.ChatContext.empty()
        for _, item_id in diff_ops.to_create:
            item = chat_ctx.get_by_id(item_id)
            if item:
                append_ctx.items.append(item)

```

dhruvnigam93: I can modify my approach to count the messages inline instead of defining a new function and remove tests if that is preferred.

longcw: yeah you are right, I think you can count them inline from the chat_ctx.

dhruvnigam93: @longcw Made this change minimal as per feedback. 
I've moved the counting to inline and removed the test file(i made sure I ran it during dev, and all checks passed).

longcw (approved):

longcw:
thanks, looks good to me!

PR #4516, opened 2026-01-13
fix frame capture order and add playback start callback in console mode

chenghao-mou:
- fix frame capture order since we have a `if self._interrupted_event.is_set() or self._pushed_duration == 0:` short-circuit check;
- add playback started callback in console so barge-in can still work;

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/_output.py
+++ b/livekit-agents/livekit/agents/voice/room_io/_output.py
@@ -99,8 +99,8 @@ class _ParticipantAudioOutput(io.AudioOutput):
             await self._flush_task
 
         for f in self._audio_bstream.push(frame.data):
+            self._audio_buf.send_nowait(f)
             self._pushed_duration += f.duration
-            await self._audio_buf.send(f)
 
     def flush(self) -> None:
         super().flush()


theomonnom (approved):

livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -99,8 +99,8 @@ async def capture_frame(self, frame: rtc.AudioFrame) -> None:
             await self._flush_task
 
         for f in self._audio_bstream.push(frame.data):
-            await self._audio_buf.send(f)
             self._pushed_duration += f.duration
+            await self._audio_buf.send(f)
```
longcw: since the `self._audio_buf` is unbounded, maybe we can use sync method `self._audio_buf.send_nowait(f)` then the order doesn't matter.

chenghao-mou: That could work. I think we already use that pattern in `flush`. I was assuming we have it async here to make sure frames are forwarded/played as soon as possible.

theomonnom: Right, send_nowait works too

chenghao-mou: Changed to the sync send.

Issue #4460, opened 2026-01-06
Support agent state of tool calling


MonkeyLeeT:
### Feature Type

Would make my life easier

### Feature Description

Given that tool calls have been more broadly adopted, it'd be nice to distinguish the agent state when it's doing tool-calling which is usually taking time / silent. That'd allow application side to be able to trigger proper animation/ux to smooth the experience for users.

### Workarounds / Alternatives

I can simulate this event handling by firing tool call event from server to worker side, but not as nice as natively supported by livekit (cloud).

PR #4518, opened 2026-01-13
add processing agentstate

tinalenguyen:
the flow for a tool call goes:
`listening` -> `thinking` -> `processing` -> `thinking` -> `speaking` -> `listening`

the first `thinking` state change is regarding making the tool call, and the second is for processing the tool call output

### DIFF
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -177,14 +177,14 @@ async def test_tool_call() -> None:
     check_timestamp(playback_finished_events[0].playback_position, 2.0, speed_factor=speed)
     check_timestamp(playback_finished_events[1].playback_position, 3.0, speed_factor=speed)
 
-    assert len(agent_state_events) == 6
+    assert len(agent_state_events) == 7
     assert agent_state_events[0].old_state == "initializing"
     assert agent_state_events[0].new_state == "listening"
     assert agent_state_events[1].new_state == "thinking"
-    assert agent_state_events[2].new_state == "speaking"
     assert (
-        agent_state_events[3].new_state == "thinking"
-    )  # from speaking to thinking when tool call is executed
+        agent_state_events[2].new_state == "processing"
+    )  # from thinking to processing when tool call is executed
+    assert agent_state_events[3].new_state == "thinking"
     check_timestamp(agent_state_events[3].created_at - t_origin, 5.5, speed_factor=speed)
     assert agent_state_events[4].new_state == "speaking"
     assert agent_state_events[5].new_state == "listening"


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1977,6 +1944,7 @@ def _tool_execution_started_cb(fnc_call: llm.FunctionCall) -> None:
             # reset the `created_at` to the start time of the tool execution
             fnc_call.created_at = time.time()
             speech_handle._item_added([fnc_call])
+            self._session._update_agent_state("processing")
```
longcw, line 1947: what if the tool call has a text message alongside, or there is a `session.say` in the tool call? the state may become `thinking -> speaking -> processing` (while agent is still speaking), or `thinking -> processing -> speaking` (while the function tool is running).

the main problem is the function call execution can be parallel with other states. I am not sure what is the original purpose of adding this state, but we had a `function_tools_executed` event, what if adding a `function_tools_started` event? does that solve the issue?

davidzhao:
@codex review

livekit-agents/livekit/agents/voice/agent_activity.py
MonkeyLeeT: Thanks for the comment. 

I mentioned in https://github.com/livekit/agents/issues/4460 that I can already fire that event from server to worker, and simulate that. So event-based handling isn't an issue.

The problem is that our client side also communicate with livekit cloud for agent state, and that state will still be thinking when tool is being used. Sure I could communicate between client and my backend/worker, but that's kinda circumventing the entire livekit agent state management.

tinalenguyen: @MonkeyLeeT What do you think of supporting something like a `ToolState`, which would switch between `executing` and `idle`? Perhaps this could be an `AgentSession` property? Let me know if this could address your use case!

MonkeyLeeT: That could work! As long as that's synced via livekit cloud so any client connecting to that can get this.

longcw: @MonkeyLeeT you can sync the tool state to client via `room.local_participant.set_attributes`, for example the `agent_state` is updated in https://github.com/livekit/agents/blob/livekit-agents@1.3.10/livekit-agents/livekit/agents/voice/room_io/room_io.py#L425-L429. 

I think you can track the tool state in the function tool itself and sync the state to the client via the `set_attributes` API.

MonkeyLeeT: Would client side get an event for attribute updated?

tinalenguyen: maybe  you can listen on `"participant_attributes_changed"`?

[docs](https://docs.livekit.io/transport/data/state/participant-attributes/)

MonkeyLeeT: @tinalenguyen Lemme make a try, but is there any other concern to support this as a state? It feels very natural to me.

Issue #4481, opened 2026-01-09
Support PCM and Opus Output Formats for ElevenLabs TTS


LuluW8071:
### Feature Type

Nice to have

### Feature Description

Currently, the LiveKit ElevenLabs TTS plugin only supports MP3 output. ElevenLabs TTS supports additional formats like PCM (e.g., 16000 Hz) and Opus (e.g., 48000 Hz, configurable bitrate). Adding support for these formats would improve real-time streaming and bandwidth efficiency.

- PCM is useful for low-latency real-time audio, ideal for WebRTC and LiveKit streaming.
- Opus offers better compression and quality at low bitrates compared to MP3

Proposed changes:

```py

from livekit.plugins.elevenlabs import TTS

tts_service = TTS(
    api_key=ELEVEN_LABS_API_KEY,
    model="eleven_flash_v2_5", 
    encoding="mp3_22050_32",        # Default (currently only supports mp3)
)

```
- Allow `encoding` to `pcm_16000`, `opus_48000_32`, or similar when creating the TTS instance. Reference for [output_format](https://elevenlabs.io/docs/api-reference/text-to-speech/convert?explorer=true)

### Additional Context

<img width="332" height="347" alt="Image" src="https://github.com/user-attachments/assets/107ee142-cbf3-4181-80d6-73dc8844e023" />

rafallezanko:
@LuluW8071 I've opened a PR for the feature request 😄 https://github.com/livekit/agents/pull/4525

LuluW8071:
@rafallezanko  Tysm :)

PR #4525, opened 2026-01-15
#4481 Added Opus and PCM encoding to ElevenLabs TTS

rafallezanko:
Added based on the document: https://elevenlabs.io/docs/api-reference/text-to-speech/convert#request.query.output_format

## Summary by CodeRabbit

* **New Features**
  * Expanded ElevenLabs text‑to‑speech audio format options: added an MP3 variant and multiple Opus and PCM formats for more output quality and compatibility choices.
  * Audio content type selection now adapts to the chosen encoding to ensure correct MIME handling for different formats.

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -53,6 +53,17 @@ def _sample_rate_from_format(output_format: TTSEncoding) -> int:
     return int(split[1])
 
 
+def _encoding_to_mimetype(encoding: TTSEncoding) -> str:
+    if encoding.startswith("mp3"):
+        return "audio/mp3"
+    elif encoding.startswith("opus"):
+        return "audio/opus"
+    elif encoding.startswith("pcm"):
+        return "audio/pcm"
+    else:
+        raise ValueError(f"Unsupported encoding: {encoding}")
+
+
 @dataclass
 class VoiceSettings:
     stability: float  # [0.0 - 1.0]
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/models.py
```diff
@@ -13,12 +13,25 @@
 
 TTSEncoding = Literal[
```
chenghao-mou, line 15: Small request: can you add the doc url here as a comment? This can help us verify/update the list faster.

rafallezanko: @chenghao-mou sure, added 👍

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/models.py
```diff
@@ -11,14 +11,28 @@
...
     "mp3_22050_32",
+    "mp3_24000_48",
     "mp3_44100",
     "mp3_44100_32",
     "mp3_44100_64",
     "mp3_44100_96",
     "mp3_44100_128",
     "mp3_44100_192",
+    "opus_48000_32",
```
chenghao-mou, line 24: Sorry I didn't realize this issue earlier:

rn, the code assumes `audio/mp3` so we need to update that as well. Otherwise it won't work with opus or pcm. (There are two places where `mime_type` are defined)

rafallezanko: Right, working on that. Thank you for the comment 👍

rafallezanko: @chenghao-mou changes have been applied. Great catch!

PR #4541, opened 2026-01-17
Chatterbox model support

plangary:
## Summary by CodeRabbit

* **New Features**
  * Added support for selecting Resemble TTS models ("chatterbox", "chatterbox-turbo") at initialization or via update; chosen model is sent for both standard and streaming synthesis.
* **Bug Fixes**
  * Model property now exposes the configured model value (or "unknown" if unset), reflecting runtime updates.

### DIFF
--- a/livekit-plugins/livekit-plugins-resemble/livekit/plugins/resemble/models.py
+++ b/livekit-plugins/livekit-plugins-resemble/livekit/plugins/resemble/models.py
@@ -6,4 +6,4 @@ class Precision(str, Enum):
     PCM_16 = "PCM_16"
 
 
-TTSModels = Literal["resemble", "chatterbox", "chatterbox-turbo"]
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-resemble/livekit/plugins/resemble/models.py
```diff
@@ -1,5 +1,9 @@
 from enum import Enum
+from typing import Literal
 
 
 class Precision(str, Enum):
     PCM_16 = "PCM_16"
+
+
+TTSModels = Literal["chatterbox", "chatterbox-turbo"]
```
davidzhao: we should include other resemble models. if no explicit name, we could call it `resemble`

that should be the default model as that's the current behavior..

PR #4546, opened 2026-01-17
Defensive fixes

chenghao-mou:
Going through detail.dev issue reports and putting some fixes together. Most of them are clean-ups or defensive fixes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Detect closed channels before reporting capacity errors
  * Recreate HTTP sessions if a previous session was closed
  * Skip processing when speech recognition responses contain no results

* **Improvements**
  * Register inference requests earlier and clean up if sending fails to avoid leaked pending work
  * Actively consume metric event streams instead of ignoring them
  * Close server resources concurrently during shutdown
  * Shield and explicitly cancel gathered tasks on interruption to prevent cascading cancellations

### DIFF
--- a/livekit-agents/livekit/agents/voice/speech_handle.py
+++ b/livekit-agents/livekit/agents/voice/speech_handle.py
@@ -169,13 +169,8 @@ class SpeechHandle:
     async def wait_if_not_interrupted(self, aw: list[asyncio.futures.Future[Any]]) -> None:
         # wrap each future in shield so we don't cancel them when we cancel the gather future
         gather_fut = asyncio.gather(*[asyncio.shield(fut) for fut in aw], return_exceptions=True)
-        _, pending = await asyncio.wait(
-            [
-                gather_fut,
-                self._interrupt_fut,
-            ],
-            return_when=asyncio.FIRST_COMPLETED,
-        )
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
these are solid fixes. lgtm

livekit-agents/livekit/agents/voice/speech_handle.py
```diff
@@ -172,8 +172,8 @@ async def wait_if_not_interrupted(self, aw: list[asyncio.futures.Future[Any]]) -
             self._interrupt_fut,
         ]
         _, pending = await asyncio.wait(fs, return_when=asyncio.FIRST_COMPLETED)
-        if pending:
-            await utils.aio.cancel_and_wait(*pending)
+        if cancellable := [fut for fut in pending if fut is not self._interrupt_fut]:
+            await utils.aio.cancel_and_wait(*cancellable)
```
longcw: I think we should let the caller to manage the awaitables instead of cancelling them whenever the speech is interrupted?

chenghao-mou: Good point. I think the intention was to cancel the `gather` only, not each `awaitable`. I will wrap them in `shield` instead.

PR #4563, opened 2026-01-20
improve EndCallTool

longcw:
1. close the job process and delete room after AgentSession closed
2. add `extra_description` and `on_tool_called`, `on_tool_completed`

## Summary by CodeRabbit

* **New Features**
  * Enhanced End Call tool with configurable end instructions, lifecycle callbacks, delayed shutdown handling, and optional room deletion.
  * Agents support an on-enter greeting and richer tool-called/tool-completed events for controlled goodbye/greeting flows.
  * Toolset now exposes tool-called and tool-completed event types for integrations.

* **Updates**
  * Session startup API simplified for a leaner flow.
  * Default TTS provider switched to cartesia/sonic-3.

### DIFF
--- a/livekit-agents/livekit/agents/beta/tools/end_call.py
+++ b/livekit-agents/livekit/agents/beta/tools/end_call.py
@@ -109,6 +109,11 @@ class EndCallTool(Toolset):
 
     def _on_session_close(self, ev: CloseEvent) -> None:
         """Close the job process when AgentSession is closed"""
+        if self._shutdown_session_task:
+            # cleanup
+            self._shutdown_session_task.cancel()
+            self._shutdown_session_task = None
+
         job_ctx = get_job_context()
 
         if self._delete_room:


examples/voice_agents/session_close_callback.py
```diff
@@ -25,26 +16,16 @@
...
 
-server = AgentServer()
-
 
 class MyAgent(Agent):
     def __init__(self):
         super().__init__(
             instructions="You are a helpful assistant.",
-            tools=[EndCallTool()],
+            tools=[EndCallTool(end_instructions="thanks the user for calling and goodbye")],
```
davidzhao:
```suggestion
            tools=[EndCallTool(end_instructions="thanks the user for calling and tell them goodbye")],
```

examples/voice_agents/session_close_callback.py
```diff
@@ -55,20 +36,11 @@ async def entrypoint(ctx: JobContext):
     session = AgentSession(
         stt="assemblyai/universal-streaming",
         llm="openai/gpt-4.1-mini",
-        tts="rime/arcana",
+        tts="elevenlabs",
```
davidzhao:
```suggestion
        tts="cartesia/sonic-3",
```

livekit-agents/livekit/agents/beta/tools/end_call.py
```diff
@@ -1,28 +1,87 @@
...
+        end_instructions: str
+        | Callable[[RunContext], Awaitable[None]]
+        | None = "say goodbye to the user",
+    ):
+        """
+        This tool allows the agent to end the call and disconnect from the room.
+
+        Args:
+            extra_instructions: Additional instructions to add to the end call tool.
+            delete_room: Whether to delete the room when the user ends the call.
```
davidzhao:
```suggestion
            delete_room: Whether to delete the room when the user ends the call. deleting the room disconnects all remote users, including SIP callers.
```

livekit-agents/livekit/agents/beta/tools/end_call.py
```diff
@@ -1,28 +1,87 @@
...
+"""
 
 
 class EndCallTool(Toolset):
-    @function_tool(name="end_call")
-    async def _end_call(self, ctx: RunContext) -> None:
+    def __init__(
+        self,
+        *,
+        extra_instructions: str = "",
```
davidzhao:
```suggestion
        extra_description: str = "",
```

livekit-agents/livekit/agents/beta/tools/end_call.py
```diff
@@ -1,28 +1,87 @@
+from collections.abc import Awaitable
+from typing import Callable
+
+from ...job import get_job_context
 from ...llm import Tool, Toolset, function_tool
 from ...log import logger
-from ...voice.events import RunContext
+from ...voice.events import CloseEvent, RunContext
+
+END_CALL_INSTRUCTIONS = """
```
davidzhao:
```suggestion
END_CALL_DESCRIPTION = """
```

livekit-agents/livekit/agents/beta/tools/end_call.py
```diff
@@ -1,28 +1,87 @@
...
+        | Callable[[RunContext], Awaitable[None]]
+        | None = "say goodbye to the user",
+    ):
+        """
+        This tool allows the agent to end the call and disconnect from the room.
+
+        Args:
+            extra_instructions: Additional instructions to add to the end call tool.
+            delete_room: Whether to delete the room when the user ends the call.
+            end_instructions: If a string is provided, it will be used as the instructions of
```
davidzhao: this is pretty neat, but slightly awkward phrasing.. I wouldn't expect `end_instructions=my_function` to behave that way..

maybe renaming to `on_end` but keeping this behavior? then it's not explicit that this is a set of instructions.

livekit-agents/livekit/agents/beta/tools/end_call.py
```diff
@@ -1,28 +1,135 @@
...
+            self._end_call,
+            name="end_call",
+            description=f"{END_CALL_DESCRIPTION}\n{extra_description}",
+        )
+        self._shutdown_session_task: asyncio.Task[None] | None = None
+
+    async def _end_call(self, ctx: RunContext) -> Any | None:
         logger.debug("end_call tool called")
-        ctx.session.shutdown()
+        llm_v = ctx.session.current_agent._get_activity_or_raise().llm
```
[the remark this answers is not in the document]
longcw: `ctx.session.llm` may not the current llm used

livekit-agents/livekit/agents/beta/tools/end_call.py
```diff
@@ -1,28 +1,127 @@
...
+        try:
+            speech_handle = await asyncio.wait_for(speech_created_fut, timeout=5.0)
+            await speech_handle
+        except asyncio.TimeoutError:
+            logger.warning("tool reply timed out, shutting down session")
+        finally:
+            ctx.session.off("speech_created", _on_speech_created)
+            ctx.session.shutdown()
+
+    def _on_session_close(self, ev: CloseEvent) -> None:
```
chenghao-mou, line 110: maybe we can clean up the `_shutdown_session_task` here?

livekit-agents/livekit/agents/beta/tools/end_call.py
```diff
@@ -1,28 +1,127 @@
...
+- The user clearly indicates they are done (e.g., “that’s all, bye”).
+- The agent determines the conversation is complete and should end.
+
+Do not call when:
+- The user asks to pause, hold, or transfer.
+- Intent is unclear.
+
+This is the final action the agent can take.
+Once called, no further interaction is possible with the user.
+Don't generate any other text or response when the tool is called.
```
chenghao-mou, line 24: This line might not be needed if we allow tool reply?

longcw:
@chenghao-mou 
> I think there might be a race condition when using a realtime model

I saw the same issue sometimes when using gemini realtime model, the tool reply is `> assistant: Let me check` without audio. It seems to be a bug of the model. I checked the message received from the server, it indeed only replied this text message then ended the turn. when you talk more before end, it can generate tool reply properly.

livekit-agents/livekit/agents/beta/tools/end_call.py
longcw: this is to prevent a text message alongside the tool call, the tool response shouldn't be affected.

davidzhao:
> @chenghao-mou
> 
> > I think there might be a race condition when using a realtime model
> 
> I saw the same issue sometimes when using gemini realtime model, the tool reply is `> assistant: Let me check` without audio. It seems to be a bug of the model. I checked the message received from the server, it indeed only replied this text message then ended the turn. when you talk more before end, it can generate tool reply properly.

should we try adding a bit of additional prompt in the description? I've seen similar behavior with previous gemini flash models and adding something like the following helped to suppress the extra response

```
do not generate a response before using tools.
```

PR #4571, opened 2026-01-20
AssemblyAI added EU streaming endpoint option

ftsef:
## Summary by CodeRabbit

* **New Features**
  * Added configurable AssemblyAI streaming endpoints (US default or EU) for live transcription connections.
* **Documentation**
  * Updated docs describing endpoint options and how to configure them during setup.

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -77,13 +77,13 @@ class STT(stt.STT):
         keyterms_prompt: NotGivenOr[list[str]] = NOT_GIVEN,
         http_session: aiohttp.ClientSession | None = None,
         buffer_size_seconds: float = 0.05,
-        endpoint_url: str = "streaming.assemblyai.com",
+        base_url: str = "wss://streaming.assemblyai.com",
     ):
         """
         Args:
-            endpoint_url: The AssemblyAI streaming endpoint hostname. Use the EU endpoint
-                (streaming.eu.assemblyai.com) for streaming in the EU. Defaults to
-                streaming.assemblyai.com.
+            base_url: The AssemblyAI streaming endpoint base URL. Use the EU endpoint
+                (wss://streaming.eu.assemblyai.com) for streaming in the EU. Defaults to
+                wss://streaming.assemblyai.com.
                 See https://www.assemblyai.com/docs/universal-streaming for more details.
         """
         super().__init__(
@@ -94,7 +94,7 @@ class STT(stt.STT):
                 offline_recognize=False,
             ),
         )
-        self._endpoint_url = endpoint_url
+        self._base_url = base_url
         assemblyai_api_key = api_key if is_given(api_key) else os.environ.get("ASSEMBLYAI_API_KEY")
         if not assemblyai_api_key:
             raise ValueError(
@@ -154,7 +154,7 @@ class STT(stt.STT):
             opts=config,
             api_key=self._api_key,
             http_session=self.session,
-            endpoint_url=self._endpoint_url,
+            base_url=self._base_url,
         )
         self._streams.add(stream)
         return stream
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -77,7 +77,15 @@ def __init__(
         keyterms_prompt: NotGivenOr[list[str]] = NOT_GIVEN,
         http_session: aiohttp.ClientSession | None = None,
         buffer_size_seconds: float = 0.05,
+        endpoint_url: Literal[
+            "wss://streaming.assemblyai.com/v3/ws", "wss://streaming.eu.assemblyai.com/v3/ws"
+        ] = "wss://streaming.assemblyai.com/v3/ws",
```
longcw:
```suggestion
        endpoint_url: str = "wss://streaming.assemblyai.com/v3/ws",
```

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -189,12 +199,14 @@ def __init__(
         conn_options: APIConnectOptions,
         api_key: str,
         http_session: aiohttp.ClientSession,
+        endpoint_url: str,
```
longcw: can you move this to the `conn_options`?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
ftsef: Now that I think about it a second time, using a string is way more flexible in case the url changes. Thanks for the hint

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
ftsef: i'm not quite sure what you mean. 
The `conn_options` is a shared object not in the AssemblyAI module. So i can't just add an other property here: https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/types.py#L54

Could you suggest how/what to change?
Thank you!

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -77,7 +77,13 @@ def __init__(
         keyterms_prompt: NotGivenOr[list[str]] = NOT_GIVEN,
         http_session: aiohttp.ClientSession | None = None,
         buffer_size_seconds: float = 0.05,
+        endpoint_url: str = "wss://streaming.assemblyai.com/v3/ws",
     ):
+        """
+        Args:
+            endpoint_url: The AssemblyAI streaming WebSocket endpoint URL. Use the EU endpoint
+                (wss://streaming.eu.assemblyai.com/v3/ws) for streaming in the EU. Defaults to US
+        """
```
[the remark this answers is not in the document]
ftsef: added the suggested improvement

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -77,7 +77,14 @@ def __init__(
         keyterms_prompt: NotGivenOr[list[str]] = NOT_GIVEN,
         http_session: aiohttp.ClientSession | None = None,
         buffer_size_seconds: float = 0.05,
+        endpoint_url: str = "wss://streaming.assemblyai.com/v3/ws",
```
davidzhao: do you want this to be a base_url instead? do you expect the `/v3/ws` parameter to be modifiable.

ftsef: Hey, thanks for your comment.

For me personally the given solution seems to be the most flexible. 
E.g. they provide a v4 oder just a v3-testing or v3-staging endpoint. This way we would've the possibility to change the target at any time.

I'll consult the assembly ai team for their opinion.

davidzhao: if there's v4 endpoint, I would imagine the plugin would need to be updated to work with it as well.

ftsef: you're totally right here. I've changed the variable to base_url and updated the description.

PR #4580, opened 2026-01-21
add session_end_timeout and gracefully cancel entrypoint on shutdown

theomonnom:


examples/frontdesk/frontdesk_agent.py
```diff
@@ -202,6 +202,8 @@ async def on_session_end(ctx: JobContext) -> None:
     else:
         ctx.tagger.fail(reason="Appointment was not booked")
```
chenghao-mou, line 203: Should this be part of a final evaluation from the judge group? Given all the evaluation results, an arbitrator could give the final verdict whether this is a success or not, and why it failed if possible.

PR #4601, opened 2026-01-23
fix: 11Labs Scribe v2 model not working with EOT prediction model

Ludobaka:
This PR fixes an issue where end-of-turn (EOT) detection fails when using ElevenLabs' Scribe V2 STT model with the `MultilingualModel` turn detector.

## Problem

When running an agent with the `scribe_v2` model and turn detection enabled via `MultilingualModel`, EOT detection does not work. The turn detector rejects the language code returned by the STT service:
```
12:08:15.241 DEBUG  livekit.agents     received user transcript
                                       {"user_transcript": "Hi, I'm just testing things.", "language": "eng", ...}
12:08:15.245 INFO   livekit.agents     Turn detector does not support language eng {"room": "console"}
```

## Root Cause

The EOT model expects the ISO 639-1 language code `en`, but ElevenLabs Scribe V2 returns the ISO 639-3 code `eng` (see [ElevenLabs API reference](https://elevenlabs.io/docs/api-reference/speech-to-text/convert#response.body.SpeechToTextChunkResponseModel.language_code)).

## Solution

Normalize the language code in `livekit-agents/livekit/agents/voice/audio_recognition.py` when setting `self._last_language`.

### Why this approach?

Handling normalization at the `audio_recognition.py` level (rather than in the ElevenLabs plugin) ensures consistent behavior across all STT providers that may return non-standard language codes.

Of course any recommendations are welcome and I'll be glad to improve this proposal.

## Summary by CodeRabbit

* **New Features**
  * Added a public language-code mapping and helper for ISO 639-3 → ISO 639-1 conversions.
* **Improvements**
  * Applied language-code normalization across speech-to-text paths so detected and streamed language labels are more consistent.
* **Chores**
  * No breaking changes; only new constants/helpers were added.


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -350,7 +350,8 @@ async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
             if not self._last_language or (
                 language and len(transcript) > MIN_LANGUAGE_DETECTION_LENGTH
             ):
-                self._last_language = language
+                normalized_language = "en" if language == "eng" else language
```
longcw: perhaps move this to the plugin of the 11labs?

Ludobaka: That was my initial thought as well, but I went with this approach for resilience: any other STT service that returns `eng` instead of `en` would be handled automatically. It also opens the way to use a normalizing function later in case other languages have the issue.

That said, I'm happy to move the normalization to the ElevenLabs plugin if you think that's the more appropriate place for it.

Ludobaka: @longcw Any conclusion?

As soon as I'll get a confirmation for one or the other approach I'll apply any needed change to the PR

chenghao-mou: Each vendor/plugin should take care of its conversion/translation to standard values. The duplicated code is acceptable at this level at this point. 

We can definitely optimize this later when this pattern becomes actually common in plugins.

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -60,6 +60,27 @@ class VADOptions(TypedDict, total=False):
...
+    "por": "pt",
+    "nld": "nl",
+    "zho": "zh",
+    "jpn": "ja",
+    "kor": "ko",
+    "ind": "id",
+    "tur": "tr",
+    "rus": "ru",
+    "hin": "hi",
+}
```
chenghao-mou: One small thing: we should include all the languages, and let the EOT decide if it is supported or not. This information might be used in other places/components where language is important (TTS/LLM). We can include a comment to the link as well https://elevenlabs.io/docs/overview/capabilities/speech-to-text#supported-languages

You can definitely create a separate `languages.py` to list all the values.

Ludobaka: As ISO 639-3 language codes represent a thousand of languages, can we consider using a Python library for this instead of handling every cases manually? I found [pycountry](https://github.com/pycountry/pycountry) that might do the job.

Another way would be to only consider Elevenlabs supported languages and manually implement the mapping.

Feel free to suggest any other library you prefer or ask for a manual dictionary implementation (either full ISO 639-3 or only the ones supported by 11Labs).

chenghao-mou: Sorry about the confusion, it's all the languages 11labs can support.

PR #4604, opened 2026-01-23
export session state as db and compute delta for sync

longcw:


livekit-agents/livekit/agents/utils/session_store.py
```diff
@@ -0,0 +1,899 @@
...
+        columns = [row[1] for row in cursor.execute(f"PRAGMA table_info({table_name})")]
+
+        # Read all rows
+        rows = []
+        for row in cursor.execute(f"SELECT * FROM {table_name}"):
+            rows.append(dict(zip(columns, row)))
+
+        return rows
+
+    def _compute_table_diff(
```
theomonnom: Do we need to compute the table diff ourselves? 
I thought sqlite3session could do it for us

theomonnom: https://sqlite.org/sessionintro.html

longcw: the session extension tracks the changes made to the table, it seems that we still need to compute the diff ourselves to update the db then it can generate the changeset. I'll test the idea later so we know more about it, right now it's still a draft.

longcw: nvm, it seems there is a session diff https://sqlite.org/session/sqlite3session_diff.html, will try to use it

longcw:
@davidzhao diff is used to compute the changeset to upload to the cloud.

davidzhao:
> @davidzhao diff is used to compute the changeset to upload to the cloud.

ah of course. 🤦🏻‍♂️

PR #4605, opened 2026-01-23
fix(liveavatar): handle AudioSegmentEnd to properly signal playback completion

ayush-ranjan-official:
## Summary

This PR fixes a bug in the LiveAvatar plugin where `AudioSegmentEnd` messages are ignored in the `_forward_audio()` function, causing the agent session to never receive playback completion signals.

## Problem

When using the LiveAvatar plugin, the agent session's state machine breaks:
- Session gets stuck cycling between `listening` → `thinking` → `listening`
- The `speaking` state is never reached
- `notify_playback_finished()` is never called for normal audio completion
- This causes the LLM to repeatedly generate responses (step repetition)

### Root Cause

The `_forward_audio()` function in `avatar.py` iterates over `self._audio_buffer` which yields two types:
1. `rtc.AudioFrame` - actual audio data
2. `AudioSegmentEnd` - marker indicating end of audio segment

However, **only `rtc.AudioFrame` is handled**. When `AudioSegmentEnd` is received, the loop continues without any action:

```
# Current buggy code
async def _forward_audio() -> None:
    async for audio_frame in self._audio_buffer:
        if isinstance(audio_frame, rtc.AudioFrame):  # Only this case handled!
            # ... process audio
            self._playback_position += resampled_frame.duration
        # AudioSegmentEnd is IGNORED - no else clause!
```

This means `notify_playback_finished()` is never called when audio naturally ends, only when interrupted (via `_on_clear_buffer`).

### Solution

Add an `elif` clause to handle `AudioSegmentEnd` and call `notify_playback_finished()`:

```
async def _forward_audio() -> None:
    async for audio_frame in self._audio_buffer:
        if isinstance(audio_frame, rtc.AudioFrame):
            # ... existing audio frame handling ...
        elif isinstance(audio_frame, AudioSegmentEnd):
            if self._audio_playing:
                    interrupted=False,
                )
                self._playback_position = 0.0
                self._audio_playing = False
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio playback handling to reliably detect when an audio segment finishes, notify playback completion, reset playback position, and stop playback.
  * Results in more accurate end-of-audio behavior and prevents lingering or resumed playback after a segment ends, improving call and media session reliability.

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4605">
  <picture>
    Open with Devin
  </picture>
</a>


davidzhao:
hi, could you rebase it form main?

ayush-ranjan-official:
rebased onto main as requested and resolved merge conflicts.

Issue #4026, opened 2025-11-19
Audio quality issue with sip outbound call


multib0t:
### Bug Description

Hi I have an outbound agent using sip trunk in twilio. I noticed some audio quality issues with tts. Specifically I get sort of fading parts of the words ie missing / fading letters and the same agent config when i use as inbound call is working perfectly. I tried different sip trunk provider wavix but the issue is still there. I tried different tts ie openai, fishaudio, deepgram, elevelabs, cartesia and same is happening for all. is this a samplerate issue or hd codec not enable  on cpaas provider? Any help would be appreciated.

### Expected Behavior

I would like the sound to at least as good as for inbound

### Reproduction Steps

```bash
initiate outbound call to my number
```

### Operating System

mac linux

### Package Versions

```bash
"livekit-agents[azure,deepgram,openai,silero,groq,cerebras,cartesia,elevenlabs,fishaudio]==1.2.18",
    "livekit-plugins-langchain>=1.1.6",
    "livekit-plugins-silero>=1.1.6",
    "livekit-plugins-noise-cancellation==0.2.5",
```

### Proposed Solution

```python

```

PR #4611, opened 2026-01-25
fix(google): improve error message for model/API mismatch in Realtime API

cdutr:
## Summary

This PR improves error messages when users accidentally configure a Gemini API model with VertexAI (or vice versa) in the Google Realtime plugin. Currently, users receive a cryptic WebSocket 1008 policy violation error that doesn't explain the root cause.

## Problem

When using the Google Realtime plugin with incorrect model/API combinations:
- Users get a confusing error: `websockets.exceptions.ConnectionClosedError: received 1008 (policy violation)`
- The error doesn't explain that the model name doesn't match the API being used
- Users spend significant time debugging WebSocket configurations instead of fixing the simple model name issue
- This is a common pain point (see issue #4026 and related reports)

### Example Scenario

A user tries to use VertexAI with a Gemini API model name:
```python
model = google.realtime.RealtimeModel(
    model="gemini-2.5-flash-native-audio-preview-09-2025",  # Gemini API model
    vertexai=True,  # But using VertexAI
)
```

**Current error:**
```
websockets.exceptions.ConnectionClosedError: received 1008 (policy violation)
```

**After this fix:**
```
Failed to connect to Gemini Live

Hint: Model 'gemini-2.5-flash-native-audio-preview-09-2025' appears to be a Gemini API model,
but VertexAI is enabled. VertexAI requires different model names (prefixed with 'gemini-live-').
Try using model='gemini-live-2.5-flash-preview-native-audio-09-2025' instead.
```

## Solution

Added a helper function `_get_model_mismatch_hint()` that:
1. Detects WebSocket 1008 policy violation errors
2. Checks if the model name pattern matches the API being used
3. Returns a helpful hint with the suggested correct model name
4. Only activates on actual 1008 errors (non-intrusive approach)

The hint is appended to both:
- Logger error messages
- Raised `APIConnectionError` exceptions

This provides immediate, actionable feedback without blocking any valid configurations or making assumptions about future model naming patterns.

## Testing

- Verified hint appears for Gemini API models + VertexAI
- Verified hint appears for VertexAI models + Gemini API
- Verified no hint for correct configurations
- Verified no hint for non-1008 errors
- Verified correct model suggestions for September and December 2025 models

Fixes https://github.com/livekit/livekit/issues/4026

## Summary by CodeRabbit

* **New Features**
  * Added model/API compatibility validation to detect mismatched model vs API choices.
* **Bug Fixes**
  * Improved error messages for connection failures with contextual hints for WebSocket policy violations and retry paths.
  * Error logs now suggest configuration adjustments to resolve detected model/API mismatches.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -43,57 +43,74 @@ DEFAULT_IMAGE_ENCODE_OPTIONS = images.EncodeOptions(
 
 lk_google_debug = int(os.getenv("LK_GOOGLE_DEBUG", 0))
 
+# Known VertexAI models for the Live API
+# See: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/live-api
+KNOWN_VERTEXAI_MODELS: frozenset[str] = frozenset(
+    {
+        "gemini-live-2.5-flash-native-audio",
+        "gemini-live-2.5-flash-preview-native-audio-09-2025",
+        "gemini-live-2.5-flash-preview-native-audio",
+    }
+)
 
-def _get_model_mismatch_hint(model: str, use_vertexai: bool, error_message: str) -> str | None:
+# Known Gemini API models for the Live API
+# See: https://ai.google.dev/gemini-api/docs/models#gemini-2.5-flash-live
+KNOWN_GEMINI_API_MODELS: frozenset[str] = frozenset(
+    {
+        "gemini-2.5-flash-native-audio-preview-12-2025",
+        "gemini-2.5-flash-native-audio-preview-09-2025",
+        "gemini-2.0-flash-exp",
+    }
+)
+
+
+def _validate_model_api_match(model: str, use_vertexai: bool) -> None:
     """
-    Generate a helpful hint when a WebSocket 1008 error suggests model/API mismatch.
+    Validate that the model name matches the API being used.
+
+    Raises ValueError if a known model is used with the wrong API configuration.
 
     Args:
         model: The model name being used
         use_vertexai: Whether VertexAI is enabled
+    """
+    if use_vertexai and model in KNOWN_GEMINI_API_MODELS:
+        raise ValueError(
+            f"Model '{model}' is a Gemini API model, but vertexai=True. "
+            f"Use a VertexAI model (e.g., 'gemini-live-2.5-flash-native-audio') "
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -44,6 +44,58 @@
...
+        A helpful hint string, or None if no mismatch detected
+    """
+    # Only provide hints for 1008 policy violations
+    if "1008" not in error_message and "policy violation" not in error_message.lower():
+        return None
+
+    is_vertexai_model = model.startswith("gemini-live-")
+    is_gemini_api_model = model.startswith(("gemini-2.", "gemini-1."))
+
+    if use_vertexai and is_gemini_api_model:
```
davidzhao: this logic doesn't seem very robust. we do not know how google is going to name their future models.

if you intend to do this, I would suggest:
- keeping a list known of vertex models and another list of gemini
- when specifically detect for mismatch, and throw an exception if a user is using vertext with gemini and vice versa

alternatively, translating the error code and log a clear message (this model isn't supported)

cdutr: Good, it makes sense. Will adapt and resubmit. Thanks for the review!

cdutr: Hi David, the first solutions was indeed too specific and made assumptions about future model naming patterns.

I refactored it to be more generic:

- Added `KNOWN_VERTEXAI_MODELS` and `KNOWN_GEMINI_API_MODELS` as explicit frozensets with the models we know today
- Created `_validate_model_api_match()` that checks if the model is in the wrong list and raises a `ValueError` right away in `__init__` - so devs get a clear error before even trying to connect
- Simplified the 1008 error hint to just be a generic message for unknown models that slip through

The idea is: if we know the model, fail fast with a clear message. If we don't know it, let it through and give a helpful hint if it fails with 1008.

I tried to optimize for developer experience, getting a `ValueError` at initialization is way better than debugging a cryptic WebSocket error later.

Let me know if you think it's overcomplicated or not generic enough. Happy to iterate on it!

This is my first time contributing to the project, so I picked a small issue to learn the process and get to know the maintainers and other contributors. Really appreciate your feedback, it helped me understand the codebase better. Looking forward to more reviews!

davidzhao (approved):
lgtm!

PR #4622, opened 2026-01-26
AGT-2474: add commit user turn support for realtime models

chenghao-mou:
This adds `commit_user_turn` support for realtime models:
- OpenAI realtime model will sends 3 messages recommended by the official doc: https://platform.openai.com/docs/guides/realtime-conversations#disable-vad
- Other realtime models will ignore this call

This allows users to use `turn_detection="manual"` with a realtime model.

## Summary by CodeRabbit

* **New Features**
  * Added an explicit "commit user turn" action for real-time agent sessions to finalize a user's turn.
  * OpenAI provider: committing a user turn also triggers response creation.
  * Google/AWS/Ultravox providers: the action exists but logs warnings or acts as a placeholder where unsupported.
  * Voice agent: when a real-time session is active, the realtime commit path is invoked before continuing audio processing.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -978,7 +978,6 @@ class AgentActivity(RecognitionHooks):
     def commit_user_turn(self, *, transcript_timeout: float, stt_flush_duration: float) -> None:
         if self._rt_session is not None:
             self._rt_session.commit_user_turn()
-            return
 
         assert self._audio_recognition is not None
         self._audio_recognition.commit_user_turn(


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1292,6 +1292,24 @@ def clear_audio(self) -> None:
...
+            )
+
+        self.commit_audio()
+        self.send_event(
+            ResponseCreateEvent(
+                type="response.create",
+                response=RealtimeResponseCreateParams(),
+            )
+        )
+        self.clear_audio()
```
longcw: why a `clear_audio` is needed here?

chenghao-mou: I know, right? It seems redundant, but it is required for OpenAI according to their doc:
Screenshot 2026-01-27 at 12 55 20

longcw: > Send  `input_audio_buffer.clear` before beginning a new user input.

I think it means you need to clear the buffer before next time you want to start a new user speech but not means it's required after `response.create` for this turn.

maybe it's similar to the `session.clear_user_turn` in the example @bml1g12 mentioned above
```pyton
# When user starts speaking
@ctx.room.local_participant.register_rpc_method("start_turn")
async def start_turn(data: rtc.RpcInvocationData):
    session.interrupt()  # Stop any current agent speech
    session.clear_user_turn()  # Clear any previous input
    session.input.set_audio_enabled(True)  # Start listening
```

chenghao-mou: <del>Yeah, I can put that call in the clear_user_turn part.</del>

Turns out we don't need this if we call `session.clear_user_turn`.

bml1g12: ah yes clear_user_turn already calls  clear_audio() under the hood it seems, so indeed when you start the new turn you probably want to clear audio, now when you end the turn - and indeed  that means probably not needed in this PR

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -976,6 +976,10 @@ def clear_user_turn(self) -> None:
             self._rt_session.clear_audio()
 
     def commit_user_turn(self, *, transcript_timeout: float, stt_flush_duration: float) -> None:
+        if self._rt_session is not None:
+            self._rt_session.commit_user_turn()
+            return
```
longcw: can we commit to both rt session and audio recognition?

chenghao-mou: Yeah, that's a good point. I don't see why we can't. I will skip the return here.

bml1g12: I think that would be relevant when using external STT with openai realtime model right? Actually we are currently exploring that, I.e. using manual turn taking with a discrete STT engine (not using openai server side whisper as STT) 

I wonder if we take this approach, if it's possible to ensure the local and remote chat contexts are consistent and also that we have a local chat context we can use to recover the real time session if there's any networking issues. 

It seems to me that when we commit the STT transcripts to local context (which maybe is what would happen if we commit the audio recognition in this code?) that we get a healthy looking local livekit chat context, but that somehow (presumably by livekit design) gets upserted automatically to the remote openai chat context, and I'm not sure if that's safe or not, given from OpenAI server side perspective they automatically add an empty user "content" message when server side STT is disabled (which I assume represents OpenAI's internal representation of those audio tokens, which is not human readable hence no "content")

This slack post explains in more detail [what I am refering to](https://livekit-users.slack.com/archives/C07FY8WHGPM/p1769699612801959?thread_ts=1769698438.162659&cid=C07FY8WHGPM)

It looks to me that maybe if using an external STT, we would ideally want to put the user STT transcriptions to local chat context (i.e. so it can be recovered if the server connection is broken) but there is no need to upset it to remot context (because OpenAI realtime API model has no need at all for user side STT transcriptions to my understanding, given it processes raw audio tokens). Is that possible in livekit I wonder though? I imagine typically livekit tries to keep local and remote contexts in sync, which creates a (theoretical) problem here when we have local STT not needed by remote model.  Although this is a theoretical problem, I should note I have not yet seen any adverse effects of this, just being cautious as it would be tough to spot if there was an issue I think

chenghao-mou: STT is essentially emitting the final transcript for the realtime mode in this case, and realtime model doesn't really process/respond to final transcripts (different from generate_reply which is direct user input)

bml1g12: I see, so yes I suppose no harm here in also committing the audio recognition - it does raise the question of whether discrete STT plus realtime model really makes sense though if it doesn't integrate with the chat context at all

Incidently, in my attempt to use it as local chat context I tried

```python
@session.on("user_input_transcribed")
def on_user_input_transcribed(transcript: UserInputTranscribedEvent) -> None:
     ....
     activity._rt_session.commit_audio() # let OpenAI create an empty message for this chunk of user input 
     manually_commit_user_message_to_local_ctx(session, transcript) # let livekit be aware of what has been said so far

async def manually_commit_user_message_to_local_ctx(
    session: AgentSession, transcript: UserInputTranscribedEvent
) -> None:
    chat_ctx = session.current_agent.chat_ctx
    chat_msg = chat_ctx.add_message(
        role="user",
        content=transcript.transcript,
        created_at=transcript.created_at,
    )
    await session.current_agent.update_chat_ctx(chat_ctx)
    session.emit("conversation_item_added", ConversationItemAddedEvent(item=chat_msg))
```
Which seems to work, but like I say might have some unforseen risks in terms of upserting to remote openai chat context

chenghao-mou: Yeah, we probably can help users commit the STT part into the chat context for real-time models in a different PR.

theomonnom:
is commit_user_turn equivalent to generate_reply without passing any argument?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -976,6 +976,9 @@ def clear_user_turn(self) -> None:
             self._rt_session.clear_audio()
 
     def commit_user_turn(self, *, transcript_timeout: float, stt_flush_duration: float) -> None:
+        if self._rt_session is not None:
+            self._rt_session.commit_user_turn()
```
theomonnom, line 980: Maybe this can be simplified to 
```suggestion
            self._rt_session.commit_audio()
            self._rt_session.generate_reply()
```

chenghao-mou: I think commit audio alone works and we don't really need any changes.

chenghao-mou: https://github.com/livekit/agents/pull/4663

PR #4627, opened 2026-01-26
plugins/google: http_options factory

fredvollmer:
VertexAI supports some request options which users may want to enable on retried requests, but not on every request. For example, the `X-Vertex-AI-LLM-Request-Type` header can be set to a value of "priority" to decrease the odds of receiving a 429. However, this options incurs 1.8x the standard cost, so it would be beneficial to enable it only on retries. I'm sure there are other scenarios as well, but this is the one my team is currently facing.

## Summary by CodeRabbit

* **New Features**
  * HTTP options can now be provided as a callable to customize request options (headers, timeouts) per retry attempt.
  * Public constructor accepts either static HTTP options or a per-attempt callable for dynamic request configuration.

* **Refactor**
  * Streaming retry logic now records the current attempt number to enable per-attempt behavior.

* **Tests**
  * Added comprehensive tests covering static and per-attempt HTTP options, retry behavior, streaming, and attempt-tracking.

### DIFF
--- a/tests/test_google_llm_http_options.py
+++ b/tests/test_google_llm_http_options.py
@@ -190,9 +190,7 @@ class TestHttpOptionsIntegration:
             )
             mock_client_cls.return_value = mock_client
 
-            static_options = types.HttpOptions(
-                timeout=5000, headers={"X-Custom": "static-value"}
-            )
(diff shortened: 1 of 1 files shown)


fredvollmer:
I now realize that a FallbackAdapter can largely handle this use case. I'll leave this open in case the maintainers feel there's a use for it, but feel free to close as well.

PR #4628, opened 2026-01-26
(taskgroup): support on_complete callback functions

tinalenguyen:
in the `add` function, you can now pass a callback that executes after the task completes. optionally taking in the task result as the only argument

## Summary by CodeRabbit

* **New Features**
  * Task groups now support per-task asynchronous completion callbacks. Callbacks are invoked immediately after each task finishes, receive the task identifier and result, and are awaited before the task group completes.
  * Ensures reliable per-task post-processing and makes it easier to react to individual task outcomes without altering overall task-group flow.

### DIFF
--- a/livekit-agents/livekit/agents/beta/workflows/task_group.py
+++ b/livekit-agents/livekit/agents/beta/workflows/task_group.py
@@ -31,6 +31,13 @@ class TaskGroupResult:
     task_results: dict[str, Any]
 
 
+@dataclass
+class TaskCompletedEvent:
+    agent_task: AgentTask
+    task_id: str
+    result: Any
+
+
 class _OutOfScopeError(ToolError):
     def __init__(self, target_task_ids: list) -> None:
         self.target_task_ids = target_task_ids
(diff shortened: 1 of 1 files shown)


theomonnom:
Should the on_complete callback be inside the TaskGroup constructor vs per-task?

tinalenguyen:
@theomonnom what if the user needed different callbacks for each task? or do you mean that the callbacks are passed all at once in the taskgroup constructor

davidzhao:
> Should the on_complete callback be inside the TaskGroup constructor vs per-task?

maybe `on_completed` and `on_task_completed`?

theomonnom:
yeah something like

`on_task_completed: Callable[[str, AgentTask], Any] # str being the task_id`

tinalenguyen:
would it feel unnatural to add the callbacks in the taskgroup constructor before the tasks themselves are added?

what if taskgroup has a method similar to `JobContext.add_shutdown_callback` to add them in bulk after the tasks are added and before taskgroup is awaited?

davidzhao:
> would it feel unnatural to add the callbacks in the taskgroup constructor before the tasks themselves are added?
> 
> what if taskgroup has a method similar to `JobContext.add_shutdown_callback` to add them in bulk after the tasks are added and before taskgroup is awaited?

I think this would be the right way.. you are basically treating this as a container.. i.e. here's my group and here's how I want to handle callbacks from the group.

then whatever items you are adding to the group should not matter.. the user is interacting with the group itself

tinalenguyen:
ahh okay, i see what you mean now, thank you for clarifying! i refactored accordingly

livekit-agents/livekit/agents/beta/workflows/task_group.py
```diff
@@ -43,22 +43,37 @@ def __init__(
         summarize_chat_ctx: bool = True,
         return_exceptions: bool = False,
         chat_ctx: NotGivenOr[llm.ChatContext] = NOT_GIVEN,
+        on_task_completed: Callable[[str, Any], Coroutine[None, None, None]] | None = None,
```
theomonnom: Should we pass an instance to the AgentTask as well?

PR #4632, opened 2026-01-27
fix(fallback_adapter): don't mark LLM unavailable on 499 client cancellation gemini 3

clykins90:
Body:
  ## Summary

  When using preemptive generation with Gemini 3, client-initiated request
  cancellations (HTTP 499) can occur frequently as users interrupt the agent. These cancellations don't
  indicate that the LLM service is unhealthy - they're normal behavior when speculative requests are no
  longer needed.

  Previously, any exception in the fallback adapter would mark the LLM as unavailable, triggering
  unnecessary fallback cascades even though the primary LLM was functioning correctly which returns 499 status codes when requests are cancelled mid-stream.

  ## Changes

  - Added check for `APIStatusError` with status code 499 before marking LLM as unavailable
  - Client cancellations are now handled gracefully without affecting LLM availability status

  ## Test plan

  - [x] Tested locally with preemptive generation enabled using Gemini 3
  - [x] Verified 499 errors no longer trigger "LLM unavailable" warnings
  - [x] Confirmed fallback adapter continues using primary LLM after 499 errors

  ## Before/After

  **Before fix:** 499 error → LLM marked unavailable → fallback cascade triggered
  **After fix:** 499 error → LLM remains available → no unnecessary fallback

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fallback LLM error handling to distinguish between client-initiated cancellations and actual service failures. Client cancellations no longer incorrectly mark the LLM provider as unavailable.

### DIFF
--- a/livekit-agents/livekit/agents/llm/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/fallback_adapter.py
@@ -7,7 +7,7 @@ from collections.abc import AsyncIterable
 from dataclasses import dataclass
 from typing import Any, ClassVar, Literal
 
-from .._exceptions import APIConnectionError, APIError
+from .._exceptions import APIConnectionError, APIError, APIStatusError
 from ..log import logger
 from ..types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, APIConnectOptions, NotGivenOr
 from .chat_context import ChatContext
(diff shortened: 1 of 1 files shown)


chenghao-mou:
Thanks for reporting the issue and opening the PR. However, I am going to close this in favor of #4637, a different approach that should solve this within the plugin. The difference here is that we haven't seen enough evidence to codify the 499 handling in the core codebase for all plugins, so it is better to have this logic only in google's plugin for now.

Issue #4641, opened 2026-01-28
Changing STT options using `update_options` doesn't change settings when using STT FallbackAdapter


xtreme-sameer-vohra:
### Bug Description

After the agent is initialized, updating the agent's STT options using `update_options` method to change STT model/language doesnt seem to have any impact.
Furthermore, the method returns successfully without any errors indicating any issues.

### Expected Behavior

If the call to `update_options` succeeds, then STT options of all underlying STT objects configured in the FallbackAdapter have successfully been updated

### Reproduction Steps

```bash
1. Initialize Agent with FallbackAdapter with STT Config (eg. model: `nova-3` and language:`multi`)
2. Subsequently, update the agent's STT configuration using `update_options` (eg. model: `nova-3` and language: `fr`) method
3. After the method returns, transcription is still configured with `multi` because agent transcribes English, Spanish, etc instead of just French.


----
Update STT options snippet

self.stt.update_options(
                language=dg_language_code,
                model=dg_properties.model,
                keywords=dg_properties.keywords,
                keyterms=dg_properties.keyterms,
            )
```

### Operating System

MacOS & Linux

### Models Used

Deepgram Nova 3

### Package Versions

```bash
livekit-agents[deepgram]=1.3.12
```

### Proposed Solution

```python

```

### Additional Context

FallBackAdapter doesn't have `update_options` method and there is no clean way to invoke the underlying STT object(s) `update_options` method

xtreme-sameer-vohra:
I tested a few previous versions `1.3.6` which also seems to have the issue.

However, version `1.2.11` seems to be working.

gokuljs:
#4658

davidzhao:
this is expected. because options with underlying models are not uniform, you'd need to update options on the STTs individually.

xtreme-sameer-vohra:
> this is expected. because options with underlying models are not uniform, you'd need to update options on the STTs individually.

ack. thx for the suggestion.

PR #4658, opened 2026-01-30
feat: add update_options method to FallbackAdapter for STT instances

gokuljs:
missing ```update_options``` method in STT ```FallbackAdapter``` that prevented runtime configuration changes from propagating to underlying STT providers. The STT FallbackAdapter class was missing the update_options method, causing calls to update STT settings (language, model, keywords, etc.) to fail silently when using the adapter. This affected all STT providers (Deepgram, OpenAI, Cartesia, Azure, etc.) when wrapped in the FallbackAdapter.

## Summary by CodeRabbit

* **New Features**
  * Added dynamic option updates for fallback speech-to-text so configuration changes propagate to all underlying recognizers at runtime.
* **Bug Fixes**
  * Safely handles providers that can't accept updated options by logging a warning instead of failing, preventing crashes or interruptions.

### DIFF
--- a/livekit-agents/livekit/agents/stt/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/stt/fallback_adapter.py
@@ -280,7 +280,7 @@ class FallbackAdapter(
                 except TypeError as e:
                     logger.warning(
                         f"Failed to update options for {stt.label}: {e}. "
-                        "Ensure kwargs are compatible with this provider's update_options signature."
+                        "Ensure options passed are compatible with this provider's update_options signature."
                     )
 
     def _on_metrics_collected(self, *args: Any, **kwargs: Any) -> None:


gokuljs:
#4641

chenghao-mou:
Thanks for the PR. I can see the need to update options using the fallback adapter, but I am not sure this is the right approach: for plugins that take similar options but expect totally different values (voice, model), errors won't be immediately thrown until during streaming.

An alternative is that you keep a reference to whatever STT you want to change and change it explicitly. Or you update each one of them with explicit calls.

For illustration:
```python
dg_stt, cartesia_stt = ..., ...
stt = FallbackAdapter([dg_stt, cartesia_stt])
...

def update_language(lan: str):
    dg_stt.update_options(language=lan)
    cartesia_stt.update_options(lan=lan)
```

gokuljs:
@chenghao-mou Thanks for the review! That makes total sense. The fact that errors would not be thrown until streaming would make testing difficult, and issues could start surfacing later in production. I think we can close this PR.

PR #4663, opened 2026-01-30
fix: commit user turn with STT and realtime

chenghao-mou:
Turns out using STT with a realtime model will trigger a response when committing a turn. I didn't notice the interruption when testing so wrongly assumed it was doing just fine.

Example to reproduce:

```python
    @session.on("agent_state_changed")
    def _on_agent_state_changed(ev: AgentStateChangedEvent):
        if ev.new_state == "speaking":
            logger.info("speaking")
            session.input.set_audio_enabled(False)
        elif ev.new_state == "listening":
            logger.info("listening")
            session.input.set_audio_enabled(True)
            session.clear_user_turn()

            async def commit_user_turn():
                nonlocal session
                logger.info("waiting 20 seconds to commit user turn")
                await asyncio.sleep(20)
                logger.info("committing user turn")
                session.commit_user_turn()
                logger.info("user turn committed")
                session.input.set_audio_enabled(False)

            task = asyncio.create_task(commit_user_turn())
            tasks.add(task)
            task.add_done_callback(tasks.discard)
        else:
            logger.info(f"agent state changed to {ev.new_state}")
```

The model will respond after 20 seconds

- [x] the STT transcripts will be committed into the chat context without triggering a duplicate response
- [x] the model will only respond once when committing a turn manually 

cc @bml1g12 

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the `commit_user_turn` method from the Realtime Session interface. This method was previously unused and non-functional across implementations and has been eliminated to simplify the API surface and reduce unnecessary complexity. Applications that reference this method require updating for compatibility.

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4663">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1017,7 +1017,9 @@ class AgentActivity(RecognitionHooks):
     def commit_user_turn(self, *, transcript_timeout: float, stt_flush_duration: float) -> None:
         skip_reply: bool = False
         if self._rt_session is not None:
-            # commit to generate a reply
+            # commit audio buffer and trigger response generation
+            # `skip_reply` prevents duplicate reply from _on_user_turn_completed
+            # but keeps flushing STT transcript into the chat context
             self._rt_session.commit_audio()
             self._session.generate_reply()
             skip_reply = True
@@ -1492,10 +1494,11 @@ class AgentActivity(RecognitionHooks):
                 return
 
             if self._rt_session is not None:
-                if info.skip_reply and info.new_transcript != "":
-                    # only add user message to chat context if reply should be skipped
-                    self._agent._chat_ctx.items.append(user_message)
-                    self._session._conversation_item_added(user_message)
+                if info.skip_reply:
+                    if info.new_transcript != "":
+                        # only add user message to chat context if reply should be skipped
+                        self._agent._chat_ctx.items.append(user_message)
+                        self._session._conversation_item_added(user_message)
                     return
                 self._rt_session.commit_audio()
 
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -32,6 +32,7 @@ _EOU_MAX_HISTORY_TURNS = 6
 @dataclass
 class _EndOfTurnInfo:
     skip_reply: bool
+    """If True, a reply was already triggered and should be skipped after end of turn detection."""
     new_transcript: str
     transcript_confidence: float
 


davidzhao (commented):
is this true when manual turn mode is used? 

I see that if it's not server vad, we have to manually create them: https://platform.openai.com/docs/api-reference/realtime-client-events/input_audio_buffer/commit

chenghao-mou:
> is this true when manual turn mode is used?
> 
> I see that if it's not server bad, we have to manually create them: [platform.openai.com/docs/api-reference/realtime-client-events/input_audio_buffer/commit](https://platform.openai.com/docs/api-reference/realtime-client-events/input_audio_buffer/commit)

I think I found the issue: when using with STT, it actually committed the transcripts to the model and therefore triggered a response. Without STT, it will not respond as expected.

davidzhao:
> > is this true when manual turn mode is used?
> > I see that if it's not server bad, we have to manually create them: [platform.openai.com/docs/api-reference/realtime-client-events/input_audio_buffer/commit](https://platform.openai.com/docs/api-reference/realtime-client-events/input_audio_buffer/commit)
> 
> I think I found the issue: when using with STT, it actually committed the transcripts to the model and therefore triggered a response. Without STT, it will not respond as expected.

got it, so the bug is: when STT is used with manual turn detection mode and realtime model, it should not trigger a response.

chenghao-mou:
> > > is this true when manual turn mode is used?
> > > I see that if it's not server bad, we have to manually create them: [platform.openai.com/docs/api-reference/realtime-client-events/input_audio_buffer/commit](https://platform.openai.com/docs/api-reference/realtime-client-events/input_audio_buffer/commit)
> > 
> > 
> > I think I found the issue: when using with STT, it actually committed the transcripts to the model and therefore triggered a response. Without STT, it will not respond as expected.
> 
> got it, so the bug is: when STT is used with manual turn detection mode and realtime model, it should not trigger a response.

Yep. But it is also kinda ambiguous what users want if they have an STT configured: do they want it only for the transcripts or actual model text input.

bml1g12:
In our use case we want STT with manual turn taking both as a way to know when it's safe to commit audio before a user turn has ended (as a user turn could be very long, we cannot wait until end of turn to commit) and for its improved transcripts post call versus openai server side transcription 

So I think ideally we'd have STT transcripts available and in the local chat context (so if there is a connection error we can restore the conversation on a fresh session) but the remote openai realtime chat context need not even be aware of it as it can function perfectly fine without a transcript of user utterances. STT also provides a neat signal to the user that the system is working during a long user turn (by allowing them to view their own transcription as they talk).

PR #4669, opened 2026-01-30
fix(liveavatar): emit playback_finished on AudioSegmentEnd

### LINKED ISSUE #4668: Bug Report: LiveAvatar playback_finished not emitted
### Bug Description

When the LiveAvatar plugin is enabled, RPCs that await `session.say(...)` never return after the speech completes. The same flow works as expected when LiveAvatar is not used. The hang occurs because the audio output chain does not emit a playback-finished signal on segment end, so the `SpeechHandle` never completes.

### Expected Behavior

The RPC should return after `session.say(...)` playback finishes, even when LiveAvatar is enabled.

### Reproduction Steps

```bash
1. Start an `AgentSession` and enable LiveAvatar (`AvatarSession.start(...)`) so the audio output is replaced with `QueueAudioOutput`.
2. Trigger an RPC handler that calls `await session.say(text)` and returns after it resolves.
3. Observe that the RPC never returns even though the audio finished playing.
```

### Operating System

Windows 11

### Models Used

- STT: `openai.STT.with_azure(model="gpt-4o-transcribe")` - LLM: `openai.LLM.with_azure(model="gpt-5.2")` - TTS: `openai.TTS.with_azure(model="gpt-4o-mini-tts", voice="marin")`

### Package Versions

```bash
- livekit-agents: 1.3.12
- livekit-plugins-liveavatar: 1.3.12
- livekit-plugins-noise-cancellation: 0.2.5
```

### Session/Room/Call IDs

_No response_

### Proposed Solution
`AvatarSession._main_task` should notify playback finished when it receives `AudioSegmentEnd` from `QueueAudioOutput`.

```python
# livekit/plugins/liveavatar/avatar.py
async def _forward_audio():
    async for audio_frame in self._audio_buffer:
        if isinstance(audio_frame, rtc.AudioFrame):
            ...
        else:  # AudioSegmentEnd
            if self._audio_playing:
                self._audio_playing = False
                    interrupted=False,
                )
                self._playback_position = 0.0
```

### Additional Context

`QueueAudioOutput.flush()` emits `AudioSegmentEnd`, but the LiveAvatar audio forwarding loop only streams audio frames. Without a playback-finished notification on segment end, `SpeechHandle.wait_for_playout()` does not complete.

_No response_

MSameerAbbas:
Fixes issue where session.say() hangs when LiveAvatar is enabled. The _forward_audio loop now handles AudioSegmentEnd markers by calling notify_playback_finished, matching the pattern in AvatarRunner.

fixes #4668

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio playback handling to properly manage completion of audio segments, reducing potential playback issues and ensuring correct state transitions.

### DIFF
--- a/livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/avatar.py
+++ b/livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/avatar.py
@@ -22,7 +22,7 @@ from livekit.agents import (
     utils,
 )
 from livekit.agents.utils import is_given
-from livekit.agents.voice.avatar import QueueAudioOutput
+from livekit.agents.voice.avatar import AudioSegmentEnd, QueueAudioOutput
 from livekit.agents.voice.room_io import ATTRIBUTE_PUBLISH_ON_BEHALF
 
 from .api import LiveAvatarAPI, LiveAvatarException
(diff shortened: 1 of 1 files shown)


MSameerAbbas:
@tinalenguyen @chenghao-mou pls review

PR #4670, opened 2026-01-30
add `lk.agent.name` attr & `wait_for_agent` utility 

theomonnom:


livekit-agents/livekit/agents/types.py
```diff
@@ -22,6 +22,12 @@
 ```
 """
 
+ATTRIBUTE_AGENT_NAME = "lk.agent.name"
```
davidzhao, line 25: when should we start changing it to `lk.dispatch_name`?

PR #4674, opened 2026-01-31
support wrapped tools with a warning message

longcw:
add backward compatibility and warning message if a function tool is wrapped.

for example, if a function tool is wrapped by a user defined wrapper, it will be converted to function tool internally with a warning message that it's supposed to wrap the function before convert it `function_tool`
```python
def log_tool(func):
    import functools

    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        logger.info(f"Calling tool {func.__name__}")
        result = func(*args, **kwargs)
        return result

    return wrapper


@log_tool
@function_tool
async def get_weather():
    ...
```

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4674">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tool update flow to correctly apply updates when no activity is ongoing.

* **Improvements**
  * Improved automatic normalization of wrapped callables so they behave like native tools, with warnings when normalization occurs.
  * Enhanced validation with clearer error messages for unsupported tool types; update flow now accepts both tool instances and normalizable objects.

* **Logging**
  * Added runtime info logs for tool invocations (tool name and received arguments).

### DIFF
--- a/livekit-agents/livekit/agents/llm/tool_context.py
+++ b/livekit-agents/livekit/agents/llm/tool_context.py
@@ -183,7 +183,6 @@ class _BaseFunctionTool(Tool, Generic[_InfoT, _P, _R]):
         return bound_tool
 
     def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _R:
-        logger.info(f"Calling tool {self._info.name} with args: {args} and kwargs: {kwargs}")
         if self._instance is not None:
             return self._func(self._instance, *args, **kwargs)
         return self._func(*args, **kwargs)


livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -303,11 +312,53 @@ def get_raw_function_info(f: RawFunctionTool) -> RawFunctionToolInfo:
...
+    if isinstance(tool, (FunctionTool, RawFunctionTool)):
+        return tool
+
+    resolved_tool: FunctionTool | RawFunctionTool | None = None
+    if (
+        hasattr(tool, "__wrapped__")  # automatically added by functools.wraps
+        and isinstance(tool.__wrapped__, (FunctionTool, RawFunctionTool))
+    ):
+        wrapped = tool.__wrapped__
+        resolved_tool = wrapped.__class__(tool, wrapped.info)  # type: ignore
```
chenghao-mou, line 329: Maybe this?
```suggestion
    if (
        hasattr(tool, "__wrapped__")  # automatically added by functools.wraps
    ):
        resolved_tool = _resolve_wrapped_tool(tool.__wrapped__)
```

longcw: this won't work as the `tool` is the callable user customized, we need to create a new `FunctionTool` that use the custom callable. using `_resolve_wrapped_tool` will recursively will ignore that.

chenghao-mou: Got it, so something like this if we really want to support nested functions:
```python
    if (
        hasattr(tool, "__wrapped__")  # automatically added by functools.wraps
        and isinstance(
            unwrapped := _resolve_wrapped_tool(tool.__wrapped__),
            (FunctionTool, RawFunctionTool),
        )
    ):
        resolved_tool = unwrapped.__class__(tool, unwrapped.info)  # type: ignore
```

but then we might have multiple warnings for the nested one. No strong opinion.

Issue #4486, opened 2026-01-10
Agent audio cuts off after one word - segment synchronizer warnings on every subsequent prompt


DexterW1:
### Bug Description

I'm experiencing an issue (occurs very rarely, but once it happens it persists) where the agent only says one word, then I see segment synchronizer warnings, and the audio cuts out. The agent is unable to complete any response after this point.

  Every time you reprompt the agent after this occurs, it will say one word and the rest gets cut off. The session becomes unusable until restarted.

  Note: The "Processing chunk" logs shown below are generated inside the tts_node (in my custom text processing stream).

### Expected Behavior

 The agent should speak the complete response without being cut off.

### Reproduction Steps

```bash
1. Run a voice agent with normal conversation flow
 2. Have back-and-forth conversation (user speaks shortly after agent finishes)
 3. At some point, the agent will get "stuck" - saying only one word per response.
 4. Every subsequent prompt results in the same behavior until session restart
```

### Operating System

macOS

### Models Used

Deepgram Nova2/ gpt-4o / gpt-4o-mini-tts

### Package Versions

```bash
"livekit-agents[cartesia,deepgram,elevenlabs,openai,silero,turn-detector]==1.3.6",
    "livekit-plugins-noise-cancellation==0.2.5",
```

### Proposed Solution

```python

```

### Additional Context

```
DEFAULT 2026-01-09T19:25:11.568075Z  Agent state: thinking -> listening
DEFAULT 2026-01-09T19:25:11.618958Z _SegmentSynchronizerImpl.pause called after close
DEFAULT 2026-01-09T19:25:11.619760Z _SegmentSynchronizerImpl.pause called after close
DEFAULT 2026-01-09T19:25:11.667113Z _SegmentSynchronizerImpl.pause called after close
DEFAULT 2026-01-09T19:25:11.669077Z _SegmentSynchronizerImpl.pause called after close
DEFAULT 2026-01-09T19:25:11.720997Z _SegmentSynchronizerImpl.pause called after close
DEFAULT 2026-01-09T19:25:11.771172Z _SegmentSynchronizerImpl.pause called after close
DEFAULT 2026-01-09T19:25:11.771279Z received user transcript
DEFAULT 2026-01-09T19:25:11.776787Z _SegmentSynchronizerImpl.resume called after close
DEFAULT 2026-01-09T19:25:11.970478Z  User state changed from speaking to listening
DEFAULT 2026-01-09T19:25:11.971063Z  User finished speaking event fired.
[...]

PR #4686, opened 2026-02-02
fix(transcription): prevent stale synchronizer impls (#4486)

furious-luke:
Issue #4486 has been giving us trouble for a little while, it resulted in our customers having their sessions stall early on without recovery.

The issue appeared to be related to pause/resume getting called when the previous synchronizer impl is closed; a race condition. The solution I've implemented tries to track the pause state so that even if the old impl is getting closed the system should pick up the pause value and apply it.

I've had this running in our system for a week or so and the issue has not reappeared yet.

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4686">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/livekit-agents/livekit/agents/voice/transcription/synchronizer.py
+++ b/livekit-agents/livekit/agents/voice/transcription/synchronizer.py
@@ -239,7 +239,13 @@ class _SegmentSynchronizerImpl:
             return
 
         if self._paused_wall_time is not None:
-            self._paused_duration += time.time() - self._paused_wall_time
+            if self._start_wall_time is not None:
+                # use max() to handle the case where pause() was called before the first audio
+                # frame was pushed (i.e., before _start_wall_time was set). This can happen when
+                # a new impl is created while the synchronizer is already paused.
+                self._paused_duration += time.time() - max(
+                    self._start_wall_time, self._paused_wall_time
+                )
             self._paused_wall_time = None
         self._output_enabled_ev.set()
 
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/transcription/synchronizer.py
```diff
@@ -469,13 +472,24 @@ def _on_attachment_changed(
...
+
+        # always create a new impl even if aclose() failed, to avoid leaving
+        # self._impl pointing to a closed impl which causes the agent to get stuck
         self._impl = _SegmentSynchronizerImpl(
             options=self._opts, next_in_chain=self._text_output._next_in_chain
         )
 
+        # apply the current pause state to the new impl
+        if self._paused:
+            self._impl.pause()
```
longcw, line 497: when pause the `impl`, it records the time it paused, and when it's resumed it's used to compute the duration the impl paused
```python
# in pause()
        if self._paused_wall_time is None:
            self._paused_wall_time = time.time()

# in resume()
        if self._paused_wall_time is not None:
            self._paused_duration += time.time() - self._paused_wall_time
            self._paused_wall_time = None
```

so if we pause the `impl` here once it's created, the `_paused_wall_time` will be earlier than `_start_wall_time` which is the time the first audio frame pushed to the synchronizer. maybe we should also change the `impl.resume()` to check the `start_wall_time`
```python
        if self._paused_wall_time is not None and self._start_wall_time is not None:
            self._paused_duration += time.time() - max(
                self._start_wall_time, self._paused_wall_time
            )
```

furious-luke: Yeah makes sense, I'll add the change.

furious-luke: I've updated the duration calculation, let me know if it doesn't match what you were thinking.

PR #4693, opened 2026-02-02
feat(azure): Add Azure Voice Live Realtime API support

### LINKED ISSUE #4716: livekit-plugins-azure doesn't support Azure Realtime API
### Feature Type

### Feature Description

 **Azure AI Voice Live Realtime API** (part of Azure AI Foundry) provides a unified, low-latency, speech-to-speech WebSocket interface.
But  `livekit-plugins-azure` package doesn't support Azure Realtime API

Use other livekit realtime plugins such as  OpenAI Realtime plugin.

### Additional Context

 **Azure Documentation:** [Voice Live API Overview](https://learn.microsoft.com/en-us/azure/ai-services/speech-service/voice-live)

007DXR:
## Description

This PR adds support for Azure Voice Live, enabling real-time speech-to-speech conversations through the Azure plugin.
Fix https://github.com/livekit/agents/issues/4716

### What's New

**Azure Voice Live Integration**
- New `RealtimeModel` class providing end-to-end speech-to-speech capabilities
- Full bidirectional audio streaming
- Server-side VAD (Voice Activity Detection) with configurable thresholds
- Automatic reconnection handling for connection resilience

### Features

- **Speech-to-Speech**: Direct audio input/output without separate STT/TTS pipeline
- **Function Calling**: Built-in tool use for agentic workflows  
- **Multilingual Support**: Works with Azure's multilingual neural voices (e.g., `en-US-AvaMultilingualNeural`)
- **Interruption Handling**: Graceful handling of user interruptions during responses
- **Metrics Collection**: Token usage and TTFT (Time to First Token) tracking
- **Debug Mode**: Optional audio saving per turn for debugging (`save_audio_per_turn=True`)

### Usage

```python
from livekit.plugins import azure

session = AgentSession(
    llm=azure.realtime.RealtimeModel(
        voice="en-US-AvaMultilingualNeural",
    )
)

await session.start(room=ctx.room, agent=Agent(instructions="You are helpful."))
```

### Environment Variables

```bash
export AZURE_VOICELIVE_ENDPOINT=https://<region>.api.cognitive.microsoft.com/
export AZURE_VOICELIVE_API_KEY=<your-speech-key>
```

### New Dependencies

- `azure-ai-voicelive[aiohttp]>=1.0.0`
- `azure-identity>=1.15.0`

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4693">
  <picture>
    Open with Devin
  </picture>
</a>


luzhangtina:
I'm excited to see Azure Voice Live Realtime API support coming to LiveKit agents! I've been evaluating Voice Live Realtime API for my app.

Based on my use case requirements, it would be great if the following capabilities could be considered:

1. BYOM Profile Support
My use case requires reliable function/tool calling with strict schema support, so I need to use models like gpt-5.1 or gpt-5.2 rather than the default gpt-realtime model. I chose Voice Live over the traditional STT→LLM→TTS pipeline due to its better latency performance, but I still need flexibility in model selection.

Per [Microsoft's BYOM documentation](https://learn.microsoft.com/en-us/azure/ai-services/speech-service/how-to-bring-your-own-model), the WebSocket URL supports a profile query parameter:

byom-azure-openai-realtime - For realtime models
byom-azure-openai-chat-completion - For chat completion models (gpt-4o, gpt-5.1, gpt-5.2, etc.)
Would it be possible to expose this profile selection?

2. Runtime Input Mode Switching
My app allows users to switch between voice and text input mid-conversation. Some way to update these session properties at runtime would be great:

modalities: ["text", "audio"] ↔ ["text"]
turn_detection: enabled ↔ disabled

3. Additional Session Configuration
It would also be helpful to have options for:
- Noise reduction: input_audio_noise_reduction
- Echo cancellation: input_audio_echo_cancellation
- TTS speed/rate control
- Semantic VAD types (azure_semantic_vad_multilingual, etc.)

007DXR:
> 1. BYOM Profile Support
>    My use case requires reliable function/tool calling with strict schema support, so I need to use models like gpt-5.1 or gpt-5.2 rather than the default gpt-realtime model. I chose Voice Live over the traditional STT→LLM→TTS pipeline due to its better latency performance, but I still need flexibility in model selection.
> 
> Per [Microsoft's BYOM documentation](https://learn.microsoft.com/en-us/azure/ai-services/speech-service/how-to-bring-your-own-model), the WebSocket URL supports a profile query parameter:
> 
> byom-azure-openai-realtime - For realtime models byom-azure-openai-chat-completion - For chat completion models (gpt-4o, gpt-5.1, gpt-5.2, etc.) Would it be possible to expose this profile selection?
>

thanks for your comments. this PR already allows user to specify different models. here I create a voice live agent using gpt-4o model

```python

session = AgentSession(
    llm=azure.realtime.RealtimeModel(
        endpoint=os.getenv("AZURE_VOICELIVE_ENDPOINT"),
        api_key=os.getenv("AZURE_VOICELIVE_API_KEY"),
        model=os.getenv("AZURE_VOICELIVE_MODEL", "gpt-4o"),
        voice=os.getenv("AZURE_VOICELIVE_VOICE", "en-US-AvaMultilingualNeural"),
        turn_detection=turn_detection,
        tool_choice="auto",  # Enable function calling
    )
)
```

you can see livekit-plugins\livekit-plugins-azure\README.md for more info.

007DXR:
> 3\. Runtime Input Mode Switching
>     My app allows users to switch between voice and text input mid-conversation. Some way to update these session properties at runtime would be great:
> 
> modalities: ["text", "audio"] ↔ ["text"] turn_detection: enabled ↔ disabled

 LiveKit agents framework provides built-in support for both voice and text input. RoomOptions for Mixed I/O: You can configure your session to accept both voice and text input using [RoomOptions]
```
 await session.start(
    agent=agent,
    room=ctx.room,
        audio_input=room_io.AudioInputOptions(sample_rate=24000, num_channels=1),
        audio_output=room_io.AudioOutputOptions(sample_rate=24000, num_channels=1),
        text_input=True,  # Enable text input
        text_output=True,  # Enable text output
    ),
)
```

here is a agent demo i run successfully
1. set LiveKit server
2. run this agent on dev mode
3.    Open https://meet.livekit.io/ and create a room. Your agent will join automatically.
```
"""
Voice and Text Realtime Agent

A multimodal agent that accepts both voice and text input, responding with voice and text output.
Uses Azure Voice Live API for low-latency speech-to-speech interactions.

Features:
- Voice input: Speak naturally, the agent listens and responds
- Text input: Type in the console or send via LiveKit TextStream to `lk.chat` topic
- Voice output: Agent responds with natural speech
- Text output: Agent responses printed to console and sent via `lk.transcription` topic
- Function calling: Weather lookup example

This is ideal for applications that need:
- Accessibility support (text fallback for hearing impaired)
- Quiet environments where voice isn't suitable
- Hybrid chat + voice interfaces

Environment Variables:
    AZURE_VOICELIVE_ENDPOINT - Azure Voice Live endpoint (wss://...)
    AZURE_VOICELIVE_API_KEY - Azure API key
    AZURE_VOICELIVE_MODEL - Azure model name (default: gpt-4o)
    AZURE_VOICELIVE_VOICE - Azure voice (default: en-US-AvaMultilingualNeural)
    LIVEKIT_URL - LiveKit server URL
    LIVEKIT_API_KEY - LiveKit API key
    LIVEKIT_API_SECRET - LiveKit API secret

Usage:
    python voice_and_text_realtime_agent.py dev
    Open https://meet.livekit.io/ and create a room. Your agent will join automatically.

Try:
 - Speak: "What's the weather like in Tokyo?"
 - send text via TextStream to `lk.chat` topic
"""

import asyncio
import logging
import os
import sys

from azure.ai.voicelive.models import AzureSemanticVadEn

from livekit.agents import (
    Agent,
    AgentServer,
    AgentSession,
    JobContext,
    RunContext,
    cli,
    room_io,
)
from livekit.agents.voice.events import ConversationItemAddedEvent
from livekit.plugins import azure

logger = logging.getLogger("voice-text-realtime-agent")
logger.setLevel(logging.INFO)

load_dotenv()


class VoiceAndTextAgent(Agent):
    """An agent that handles both voice and text interactions."""

    def __init__(self) -> None:
        super().__init__(
            instructions=(
                "You are a friendly, helpful assistant that can interact via voice and text. "
                "Keep your responses conversational and concise since users may be listening. "
                "Avoid using markdown, emojis, or special formatting in your responses. "
                "You can help with weather information and general questions."
            ),
        )

    async def on_enter(self):
        """Called when the agent becomes active in the session."""
        # Generate a greeting when the agent starts
        # allow_interruptions=False ensures the greeting completes
        self.session.generate_reply(
            instructions="Greet the user warmly and let them know they can speak or type to interact with you.",
            allow_interruptions=False,
        )

    @function_tool
    async def get_weather(
        self,
        context: RunContext,
        location: str,
    ) -> str:
        """Get the current weather for a location.

        Args:
            location: The city or location to get weather for (e.g., "Tokyo", "New York")
        """
        logger.info(f"Looking up weather for {location}")
        # In a real application, you would call a weather API here
        # This is a placeholder response
        return f"The weather in {location} is currently sunny with a temperature of 72°F (22°C)."


server = AgentServer()


@server.rtc_session()
    """Main entry point for the agent session."""
    logger.info(f"Starting voice and text realtime agent in room: {ctx.room.name}")

    # Configure Semantic VAD for intelligent turn detection
    turn_detection = AzureSemanticVadEn(
        threshold=0.5,  # Voice activity detection threshold (0.0-1.0)
        silence_duration_ms=500,  # Silence duration before turn ends
        prefix_padding_ms=300,  # Audio padding before speech
        speech_duration_ms=200,  # Minimum speech duration to trigger detection
        remove_filler_words=True,  # Remove filler words like "um", "uh"
    )

    # Create session with Azure Voice Live API
    session = AgentSession(
        llm=azure.realtime.RealtimeModel(
            endpoint=os.getenv("AZURE_VOICELIVE_ENDPOINT"),
            api_key=os.getenv("AZURE_VOICELIVE_API_KEY"),
            model=os.getenv("AZURE_VOICELIVE_MODEL", "gpt-4o"),
            voice=os.getenv("AZURE_VOICELIVE_VOICE", "en-US-AvaMultilingualNeural"),
            turn_detection=turn_detection,  # Use semantic VAD
            tool_choice="auto",
        )
    )

    # Start the session with both voice and text I/O enabled
    await session.start(
        agent=VoiceAndTextAgent(),
        room=ctx.room,
            # Voice I/O configuration
                sample_rate=24000,
                num_channels=1,
            ),
            audio_output=room_io.AudioOutputOptions(
                sample_rate=24000,
                num_channels=1,
            ),
            # Text I/O configuration
            # - Text input: Users can send text via TextStream to `lk.chat` topic
            # - Text output: Agent responses sent via TextStream to `lk.transcription` topic
            text_input=True,
            text_output=True,
        ),
    )


if __name__ == "__main__":
    cli.run_app(server)
```

007DXR:
>Additional Session Configuration
>     It would also be helpful to have options for:
> 
> * Noise reduction: input_audio_noise_reduction
> * Echo cancellation: input_audio_echo_cancellation
> * TTS speed/rate control
> * Semantic VAD types (azure_semantic_vad_multilingual, etc.)

Voice Live as having built-in noise reduction and echo cancellation (even when you don’t explicitly set options).
You can explicitly set Semantic VAD like this:
 ```
   # Configure Semantic VAD for intelligent turn detection
    # Options:
    # - AzureSemanticVad: Default semantic VAD (multilingual)
    # - AzureSemanticVadEn: English-only, optimized for English
    # - AzureSemanticVadMultilingual: Explicit multilingual support
    from azure.ai.voicelive.models import AzureSemanticVadEn

    turn_detection = AzureSemanticVadEn(
        threshold=0.5,  # Voice activity detection threshold (0.0-1.0)
        silence_duration_ms=500,  # Silence duration before turn ends
        prefix_padding_ms=300,  # Audio padding before speech
        speech_duration_ms=200,  # Minimum speech duration to trigger detection
        remove_filler_words=True,  # Remove filler words like "um", "uh"
    )
```

**Comment by luzhangtina** [comment]
> > 1. BYOM Profile Support
> >    My use case requires reliable function/tool calling with strict schema support, so I need to use models like gpt-5.1 or gpt-5.2 rather than the default gpt-realtime model. I chose Voice Live over the traditional STT→LLM→TTS pipeline due to its better latency performance, but I still need flexibility in model selection.
> > 
> > Per [Microsoft's BYOM documentation](https://learn.microsoft.com/en-us/azure/ai-services/speech-service/how-to-bring-your-own-model), the WebSocket URL supports a profile query parameter:
> > byom-azure-openai-realtime - For realtime models byom-azure-openai-chat-completion - For chat completion models (gpt-4o, gpt-5.1, gpt-5.2, etc.) Would it be possible to expose this profile selection?
> 
> thanks for your comments. this PR already allows user to specify different models. here I create a voice live agent using gpt-4o model
> 
> ```python
> session = AgentSession(
>     llm=azure.realtime.RealtimeModel(
>         endpoint=os.getenv("AZURE_VOICELIVE_ENDPOINT"),
>         api_key=os.getenv("AZURE_VOICELIVE_API_KEY"),
>         model=os.getenv("AZURE_VOICELIVE_MODEL", "gpt-4o"),
>         voice=os.getenv("AZURE_VOICELIVE_VOICE", "en-US-AvaMultilingualNeural"),
>         turn_detection=turn_detection,
>         tool_choice="auto",  # Enable function calling
>     )
> )
> ```
> 
> you can see livekit-plugins\livekit-plugins-azure\README.md for more info.

Thanks for the clarification! I tested it and confirmed that gpt-5.2 works without the profile parameter - Azure auto-detects the model type. Good to know!

**Comment by luzhangtina** [comment]
> > 3. Runtime Input Mode Switching
> > My app allows users to switch between voice and text input mid-conversation. Some way to update these session properties at runtime would be great:
> > modalities: ["text", "audio"] ↔ ["text"] turn_detection: enabled ↔ disabled
> 
> LiveKit agents framework provides built-in support for both voice and text input. RoomOptions for Mixed I/O: You can configure your session to accept both voice and text input using [RoomOptions]
> 
> ```
>  await session.start(
>     agent=agent,
>     room=ctx.room,
>     room_options=room_io.RoomOptions(
>         audio_input=room_io.AudioInputOptions(sample_rate=24000, num_channels=1),
>         audio_output=room_io.AudioOutputOptions(sample_rate=24000, num_channels=1),
>         text_input=True,  # Enable text input
>         text_output=True,  # Enable text output
>     ),
> )
> ```
> 
> here is a agent demo i run successfully
> 
> 1. set LiveKit server
> 2. run this agent on dev mode
> 3. Open https://meet.livekit.io/ and create a room. Your agent will join automatically.
> 
> ```
> """
> Voice and Text Realtime Agent
> 
> A multimodal agent that accepts both voice and text input, responding with voice and text output.
> Uses Azure Voice Live API for low-latency speech-to-speech interactions.
> 
> Features:
> - Voice input: Speak naturally, the agent listens and responds
> - Text input: Type in the console or send via LiveKit TextStream to `lk.chat` topic
> - Voice output: Agent responds with natural speech
> - Text output: Agent responses printed to console and sent via `lk.transcription` topic
> - Function calling: Weather lookup example
> 
> This is ideal for applications that need:
> - Accessibility support (text fallback for hearing impaired)
> - Quiet environments where voice isn't suitable
> - Hybrid chat + voice interfaces
> 
> Environment Variables:
>     AZURE_VOICELIVE_ENDPOINT - Azure Voice Live endpoint (wss://...)
>     AZURE_VOICELIVE_API_KEY - Azure API key
>     AZURE_VOICELIVE_MODEL - Azure model name (default: gpt-4o)
>     AZURE_VOICELIVE_VOICE - Azure voice (default: en-US-AvaMultilingualNeural)
>     LIVEKIT_URL - LiveKit server URL
>     LIVEKIT_API_KEY - LiveKit API key
>     LIVEKIT_API_SECRET - LiveKit API secret
> 
> Usage:
>     python voice_and_text_realtime_agent.py dev
>     Open https://meet.livekit.io/ and create a room. Your agent will join automatically.
> 
> Try:
>  - Speak: "What's the weather like in Tokyo?"
>  - send text via TextStream to `lk.chat` topic
> """
> 
> import asyncio
> import logging
> import os
> import sys
> 
> from azure.ai.voicelive.models import AzureSemanticVadEn
> from dotenv import load_dotenv
> 
> from livekit.agents import (
>     Agent,
>     AgentServer,
>     AgentSession,
>     JobContext,
>     RunContext,
>     cli,
>     room_io,
> )
> from livekit.agents.llm import function_tool
> from livekit.agents.voice.events import ConversationItemAddedEvent
> from livekit.plugins import azure
> 
> logger = logging.getLogger("voice-text-realtime-agent")
> logger.setLevel(logging.INFO)
> 
> load_dotenv()
> 
> 
> class VoiceAndTextAgent(Agent):
>     """An agent that handles both voice and text interactions."""
> 
>     def __init__(self) -> None:
>         super().__init__(
>             instructions=(
>                 "You are a friendly, helpful assistant that can interact via voice and text. "
>                 "Keep your responses conversational and concise since users may be listening. "
>                 "Avoid using markdown, emojis, or special formatting in your responses. "
>                 "You can help with weather information and general questions."
>             ),
>         )
> 
>     async def on_enter(self):
>         """Called when the agent becomes active in the session."""
>         # Generate a greeting when the agent starts
>         # allow_interruptions=False ensures the greeting completes
>         self.session.generate_reply(
>             instructions="Greet the user warmly and let them know they can speak or type to interact with you.",
>             allow_interruptions=False,
>         )
> 
>     @function_tool
>     async def get_weather(
>         self,
>         context: RunContext,
>         location: str,
>     ) -> str:
>         """Get the current weather for a location.
> 
>         Args:
>             location: The city or location to get weather for (e.g., "Tokyo", "New York")
>         """
>         logger.info(f"Looking up weather for {location}")
>         # In a real application, you would call a weather API here
>         # This is a placeholder response
>         return f"The weather in {location} is currently sunny with a temperature of 72°F (22°C)."
> 
> 
> server = AgentServer()
> 
> 
> @server.rtc_session()
> async def entrypoint(ctx: JobContext):
>     """Main entry point for the agent session."""
>     logger.info(f"Starting voice and text realtime agent in room: {ctx.room.name}")
> 
>     # Configure Semantic VAD for intelligent turn detection
>     turn_detection = AzureSemanticVadEn(
>         threshold=0.5,  # Voice activity detection threshold (0.0-1.0)
>         silence_duration_ms=500,  # Silence duration before turn ends
>         prefix_padding_ms=300,  # Audio padding before speech
>         speech_duration_ms=200,  # Minimum speech duration to trigger detection
>         remove_filler_words=True,  # Remove filler words like "um", "uh"
>     )
> 
>     # Create session with Azure Voice Live API
>     session = AgentSession(
>         llm=azure.realtime.RealtimeModel(
>             endpoint=os.getenv("AZURE_VOICELIVE_ENDPOINT"),
>             api_key=os.getenv("AZURE_VOICELIVE_API_KEY"),
>             model=os.getenv("AZURE_VOICELIVE_MODEL", "gpt-4o"),
>             voice=os.getenv("AZURE_VOICELIVE_VOICE", "en-US-AvaMultilingualNeural"),
>             turn_detection=turn_detection,  # Use semantic VAD
>             tool_choice="auto",
>         )
>     )
> 
>     # Start the session with both voice and text I/O enabled
>     await session.start(
>         agent=VoiceAndTextAgent(),
>         room=ctx.room,
>         room_options=room_io.RoomOptions(
>             # Voice I/O configuration
>             audio_input=room_io.AudioInputOptions(
>                 sample_rate=24000,
>                 num_channels=1,
>             ),
>             audio_output=room_io.AudioOutputOptions(
>                 sample_rate=24000,
>                 num_channels=1,
>             ),
>             # Text I/O configuration
>             # - Text input: Users can send text via TextStream to `lk.chat` topic
>             # - Text output: Agent responses sent via TextStream to `lk.transcription` topic
>             text_input=True,
>             text_output=True,
>         ),
>     )
> 
> 
> if __name__ == "__main__":
>     cli.run_app(server)
> ```

I appreciate the pointer to RoomOptions for mixed I/O. My use case is specifically about dynamic switching mid-session — users start in voice mode, switch to text (typing), then back to voice within the same session.

The reason I was looking at updating modalities mid-session is cost — Azure Voice Live charges per token, and audio output tokens are significantly more expensive than text tokens. When a user switches to text mode, I'd ideally want to tell Azure to only generate text output (no audio), to avoid paying for audio tokens that won't be used.

In my testing (I might be wrong here), it seems like Azure Voice Live doesn't support this:

session.update with modalities: ["text"] didn't seem to take effect — the response still showed modalities: ['audio', 'text']
Initializing with modalities: ["text"] from the start resulted in no output at all
turn_detection couldn't be re-enabled after being set to None (turn_detection_type_change_not_allowed)
I ended up working around it by controlling audio at the agent level (session.input/output.set_audio_enabled()), which works functionally but Azure still generates audio tokens server-side — so no cost savings on the Azure side unfortunately.

Have you seen different behavior? Would love to know if there's a supported way to switch to text-only output mid-session to save on audio token costs.

**Comment by theomonnom** [comment]
Could this be created on top of the openai plugin? Like we did for the xai plugin?

File: livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/realtime/utils.py [hunk 1]
```diff
@@ -0,0 +1,229 @@
...
+            parameters=raw_schema.get("parameters", {}),
+        )
+
+        logger.info(f"[TOOL_CONVERSION] Converted raw tool {tool.info.name}")
+        return azure_tool
+    else:
+        raise ValueError(f"Unsupported tool type: {type(tool)}")
+
+
+def livekit_item_to_azure_item(item: llm.ChatItem) -> dict[str, Any] | FunctionCallOutputItem:
```
luzhangtina: Bug: livekit_item_to_azure_item() returns raw dicts — breaks conversation.item.create() in text mode
I've been using this plugin and found that text input doesn't work when switching from voice to text mode. The root cause is in utils.py's livekit_item_to_azure_item().

The Problem
The function returns raw dicts for message and function_call item types. When these dicts are passed to the Azure SDK's conversation.item.create(), the SDK does this:

```
if not isinstance(item, ConversationRequestItem):
    item = ConversationRequestItem(**dict(item))
```

This unpacks the dict as kwargs into the base class ConversationRequestItem, which only accepts type and id. Subclass-specific fields like role and content cause:


TypeError: ConversationRequestItem.__init__() got an unexpected keyword argument 'role'
The error is silently caught by the except Exception in update_chat_ctx(), so the text message never reaches Azure — it just disappears.

The Fix
Return proper Azure SDK model instances instead of raw dicts:


```
def livekit_item_to_azure_item(item: llm.ChatItem) -> ConversationRequestItem:
    if item.type == "function_call_output":
        return FunctionCallOutputItem(call_id=item.call_id, output=item.output)

    if item.type == "function_call":
        return FunctionCallItem(
            call_id=item.call_id, name=item.name,
            arguments=item.arguments, id=item.id,
        )

    if item.type == "message":
        if item.role in ("system", "developer"):
            content_parts: list[MessageContentPart] = [
                InputTextContentPart(text=c) for c in item.content if isinstance(c, str)
            ]
            return SystemMessageItem(content=content_parts, id=item.id)

        if item.role == "assistant":
            content_parts: list[MessageContentPart] = [
                OutputTextContentPart(text=c) for c in item.content if isinstance(c, str)
            ]
            return AssistantMessageItem(content=content_parts, id=item.id)

        if item.role == "user":
            content_parts: list[MessageContentPart] = []
            for c in item.content:
                if isinstance(c, str):
                    content_parts.append(InputTextContentPart(text=c))
                elif isinstance(c, llm.AudioContent):
                    encoded_audio = base64.b64encode(
                        rtc.combine_audio_frames(c.frame).data
                    ).decode("utf-8")
                    content_parts.append(
                        InputAudioContentPart(audio=encoded_audio, transcript=c.transcript)
                    )
            return UserMessageItem(content=content_parts, id=item.id)
        raise ValueError(f"Unsupported role: {item.role}")
    raise ValueError(f"Unsupported item type: {item.type}")
```

These are all ConversationRequestItem subclasses, so they pass the SDK's isinstance check and skip the broken kwargs unpacking.

luzhangtina:
Language configuration missing — causes multi-language misidentification

Another issue I encountered during this integration testing where the Voice Live API incorrectly detects the user's language — responding in Japanese, German, French, or Chinese even when the user is speaking English only.

The root cause is that _configure_session doesn't set input_audio_transcription with a language parameter, so the API falls back to auto-detection which is unreliable. Additionally, AzureStandardVoice is created without a locale, leaving TTS language unconstrained.

Suggestion: Expose language as an optional parameter on RealtimeModel.__init__ so users can configure it per their use case — whether constraining to a single language or supporting multilingual sessions.

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/realtime/utils.py
007DXR: Thank you very much for pointing out this bug . I’ve  followed your advice and updated it.”

007DXR:
> Language configuration missing — causes multi-language misidentification
> 
> Another issue I encountered during this integration testing where the Voice Live API incorrectly detects the user's language — responding in Japanese, German, French, or Chinese even when the user is speaking English only.
> 
> The root cause is that _configure_session doesn't set input_audio_transcription with a language parameter, so the API falls back to auto-detection which is unreliable. Additionally, AzureStandardVoice is created without a locale, leaving TTS language unconstrained.
> 
> Suggestion: Expose language as an optional parameter on RealtimeModel.**init** so users can configure it per their use case — whether constraining to a single language or supporting multilingual sessions.

Thanks for your suggestion. I've added an input_audio_transcription parameter that configures language for both transcription and TTS. Here is an example:
```
# Configure input audio transcription with language constraint
# This helps prevent language misidentification
input_audio_transcription = AudioInputTranscriptionOptions(
    model="whisper-1",
    language="en-US",  # Constrain to English for reliable detection
)

session = AgentSession(
    llm=azure.realtime.RealtimeModel(
        endpoint=os.getenv("AZURE_VOICELIVE_ENDPOINT"),
        api_key=os.getenv("AZURE_VOICELIVE_API_KEY"),
        model=os.getenv("AZURE_VOICELIVE_MODEL", "gpt-4o"),
        voice=os.getenv("AZURE_VOICELIVE_VOICE", "en-US-AvaMultilingualNeural"),
        input_audio_transcription=input_audio_transcription,
        turn_detection=turn_detection,
        tool_choice="auto",  # Enable function calling
    )
)
```
Please see README for details.

007DXR:
> Could this be created on top of the openai plugin? Like we did for the xai plugin?

The current Azure plugin in livekit-plugins-azure is implementing Azure Voice Live, which cannot be built on top of the OpenAI plugin.
- Azure OpenAI Realtime - OpenAI's Realtime API hosted on Azure. This uses the same protocol as OpenAI and is already supported via livekit-plugins-openai.realtime.RealtimeModel.with_azure().

- Azure Voice Live - Azure's native voice AI service (azure.ai.voicelive). This is a completely different service with its own SDK and protocol.

007DXR:
> I ended up working around it by controlling audio at the agent level (session.input/output.set_audio_enabled()), which works functionally but Azure still generates audio tokens server-side — so no cost savings on the Azure side unfortunately.

Azure Voice Live supports generate text only. but there is a bug in  Azure realtime model plugin: the Azure realtime model plugin only handles RESPONSE_AUDIO_TRANSCRIPT_DELTA. When using modalities=["text"], the text is sent via RESPONSE_TEXT_DELTA, not RESPONSE_AUDIO_TRANSCRIPT_DELTA. I've fix this bug .
I create a text-only agent demo, which  ensures Azure Voice Live only generates text token, no audio token. you can try this agent on console mode: `uv run .\text_only.py console --text`
```
## This example demonstrates a text-only agent using direct text transport.
## Instead of using LiveKit's TextStream, this example shows how to:
## - Send text input via: `session.generate_reply(user_input="user's input text")`
## - Receive agent's response via `session.on("conversation_item_added", ev)`
## docs: https://docs.livekit.io/agents/build/events/#conversation_item_added

import asyncio
import logging
import os
import sys

from azure.ai.voicelive.models import AzureSemanticVadEn

from livekit.agents import (
    Agent,
    AgentServer,
    AgentSession,
    JobContext,
    RunContext,
    cli,
    room_io,
)
from livekit.agents.voice.events import ConversationItemAddedEvent, AgentStateChangedEvent
from livekit.plugins import azure

logger = logging.getLogger("voice-text-realtime-agent")
logger.setLevel(logging.INFO)

load_dotenv()

class MyAgent(Agent):
    def __init__(self) -> None:
        super().__init__(
        )

server = AgentServer()

@server.rtc_session()
    session = AgentSession(
        llm=azure.realtime.RealtimeModel(
            endpoint=os.getenv("AZURE_VOICELIVE_ENDPOINT"),
            api_key=os.getenv("AZURE_VOICELIVE_API_KEY"),
            model=os.getenv("AZURE_VOICELIVE_MODEL", "gpt-4o"),
            voice=os.getenv("AZURE_VOICELIVE_VOICE", "en-US-AvaMultilingualNeural"),
            # turn_detection=turn_detection,  # Use semantic VAD
            tool_choice="auto",
            modalities=["text"]
        )
    )


    # Wait for session to be ready before sending text input
    session_ready = asyncio.Event()

    @session.on("agent_state_changed")
    def on_agent_state_changed(event: AgentStateChangedEvent):
        # Session is ready when agent transitions from "initializing" to another state
        if event.old_state == "initializing" and event.new_state in ("listening", "idle"):
            session_ready.set()
           

    # Disable audio and transcript I/O (for pure text mode)
    session.input.audio = None
    session.output.audio = None
    session.output.transcription = None

    await session.start(
        agent=MyAgent(),
    )

    # Wait for the session to be ready (agent is no longer "initializing")
    await session_ready.wait()

    # Example: Send text input directly to the agent
    # You can call this from anywhere (e.g., HTTP endpoint, websocket, etc.)
    await session.generate_reply(user_input="Hello, how can you help me today?")


if __name__ == "__main__":
    cli.run_app(server)
```

monuminu:
any progress on this ?

luzhangtina:
> > I ended up working around it by controlling audio at the agent level (session.input/output.set_audio_enabled()), which works functionally but Azure still generates audio tokens server-side — so no cost savings on the Azure side unfortunately.
> 
> Azure Voice Live supports generate text only. but there is a bug in Azure realtime model plugin: the Azure realtime model plugin only handles RESPONSE_AUDIO_TRANSCRIPT_DELTA. When using modalities=["text"], the text is sent via RESPONSE_TEXT_DELTA, not RESPONSE_AUDIO_TRANSCRIPT_DELTA. I've fix this bug . I create a text-only agent demo, which ensures Azure Voice Live only generates text token, no audio token. you can try this agent on console mode: `uv run .\text_only.py console --text`
> 
> ```
> ## This example demonstrates a text-only agent using direct text transport.
> ## Instead of using LiveKit's TextStream, this example shows how to:
> ## - Send text input via: `session.generate_reply(user_input="user's input text")`
> ## - Receive agent's response via `session.on("conversation_item_added", ev)`
> ## docs: https://docs.livekit.io/agents/build/events/#conversation_item_added
> 
> import asyncio
> import logging
> import os
> import sys
> 
> from azure.ai.voicelive.models import AzureSemanticVadEn
> from dotenv import load_dotenv
> 
> from livekit.agents import (
>     Agent,
>     AgentServer,
>     AgentSession,
>     JobContext,
>     RunContext,
>     cli,
>     room_io,
> )
> from livekit.agents.llm import function_tool
> from livekit.agents.voice.events import ConversationItemAddedEvent, AgentStateChangedEvent
> from livekit.plugins import azure
> 
> logger = logging.getLogger("voice-text-realtime-agent")
> logger.setLevel(logging.INFO)
> 
> load_dotenv()
> 
> class MyAgent(Agent):
>     def __init__(self) -> None:
>         super().__init__(
>             instructions="You are a helpful assistant.",
>         )
> 
> server = AgentServer()
> 
> @server.rtc_session()
> async def entrypoint(ctx: JobContext):
>     session = AgentSession(
>         llm=azure.realtime.RealtimeModel(
>             endpoint=os.getenv("AZURE_VOICELIVE_ENDPOINT"),
>             api_key=os.getenv("AZURE_VOICELIVE_API_KEY"),
>             model=os.getenv("AZURE_VOICELIVE_MODEL", "gpt-4o"),
>             voice=os.getenv("AZURE_VOICELIVE_VOICE", "en-US-AvaMultilingualNeural"),
>             # turn_detection=turn_detection,  # Use semantic VAD
>             tool_choice="auto",
>             modalities=["text"]
>         )
>     )
> 
> 
>     # Wait for session to be ready before sending text input
>     session_ready = asyncio.Event()
> 
>     @session.on("agent_state_changed")
>     def on_agent_state_changed(event: AgentStateChangedEvent):
>         # Session is ready when agent transitions from "initializing" to another state
>         if event.old_state == "initializing" and event.new_state in ("listening", "idle"):
>             session_ready.set()
>            
> 
>     # Disable audio and transcript I/O (for pure text mode)
>     session.input.audio = None
>     session.output.audio = None
>     session.output.transcription = None
> 
>     await session.start(
>         agent=MyAgent(),
>     )
> 
>     # Wait for the session to be ready (agent is no longer "initializing")
>     await session_ready.wait()
> 
>     # Example: Send text input directly to the agent
>     # You can call this from anywhere (e.g., HTTP endpoint, websocket, etc.)
>     await session.generate_reply(user_input="Hello, how can you help me today?")
> 
> 
> if __name__ == "__main__":
>     cli.run_app(server)
> ```

I did a test, found following case:

Azure Voice Live Modality Test Results:

| Scenario | Azure Confirmed Modalities | Event Path | Audio Generated | Playback Heard |
|----------|---------------------------|------------|-----------------|----------------|
| Init `["text"]` only | `[TEXT]` | `RESPONSE_TEXT_DELTA` | No | No |
| Init `["text","audio"]` → mid-session `session.update(["text"])` | `[AUDIO, TEXT]` (ignored) | `RESPONSE_AUDIO_DELTA` + `RESPONSE_AUDIO_TRANSCRIPT_DELTA` | Yes | Yes |

From my test, it seems like mid-session modality switching via `session.update()` is cosmetic — Azure accepts but ignores it. Only init-time `modalities=["text"]` truly eliminates audio token generation. For runtime mode switching, I have to use `set_audio_enabled()` at the LiveKit agent level to make it work. Welcome for correction and suggestion.

007DXR:
Hi @tinalenguyen, just a gentle ping on this PR 😊
Please let me know if I should adjust the implementation.
Thanks!

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/realtime/realtime_model.py
```diff
@@ -0,0 +1,1250 @@
...
+        api_key: str | None = None,
+        use_default_credential: bool = False,
+        conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
+        save_audio_per_turn: bool = False,
+    ) -> None:
+        """
+        Initialize Azure Voice Live Realtime model.
+
+        Args:
+            endpoint: Azure Voice Live endpoint URL (wss://...). If None, reads from AZURE_VOICELIVE_ENDPOINT.
```
tinalenguyen: could we use `AZURE_VOICE_LIVE_ENDPOINT` instead? this goes for other variables in the plugin

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/realtime/realtime_model.py
```diff
@@ -0,0 +1,1250 @@
...
+            modalities: List of modalities to enable (default: ["text", "audio"]).
+            turn_detection: Turn detection configuration. Accepts ServerVad, AzureSemanticVad,
+                AzureSemanticVadEn, or AzureSemanticVadMultilingual (default: ServerVad with threshold=0.5).
+            tool_choice: Tool selection policy.
+            temperature: Sampling temperature (default: 0.8).
+            max_output_tokens: Maximum output tokens (default: 4096).
+            api_key: Azure API key. If None, reads from AZURE_VOICELIVE_API_KEY.
+            use_default_credential: Use DefaultAzureCredential for auth instead of API key.
+            conn_options: Connection retry and timeout options.
+            save_audio_per_turn: Save audio to file for each turn (default: False).
```
tinalenguyen: is this natively supported by azure voice live as an argument? if not, then i would exclude this and its respective logic. users can listen on `azure_server_event_received` to combine the received audio bytes into an audio file

tinalenguyen (commented):
hi, thank you so much for working on this PR! i left a few comments, could you also address the devin comments and typecheck?

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/realtime/realtime_model.py
007DXR: save_audio_per_turn is not a native Azure Voice Live API argument. I'll remove it along with its related logic.

tinalenguyen (commented):
when testing this out (i used the[ healthcare agent example](https://github.com/livekit/agents/blob/main/examples/healthcare/healthcare_agent.py)), i get this error consistently with every `generate_reply`:

`failed to generate a reply after tool execution: pending response discarded due to session reconnection`

i think it has to do with how the connection is handled here. does azure time out often? it seems the plugin expects many reconnects to happen

besides this bug, could you also resolve the type check?

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/realtime/realtime_model.py
```diff
@@ -0,0 +1,1265 @@
...
+        """
+        # Generate unique event ID to track this response
+        event_id = utils.shortuuid("response_create_")
+        fut = asyncio.Future[llm.GenerationCreatedEvent]()
+
+        # Store the future so we can resolve it when response.created arrives
+        self._response_created_futures[event_id] = fut
+
+        # Send response.create request with event_id in metadata
+        asyncio.create_task(self._run_generate_reply(event_id, instructions))
```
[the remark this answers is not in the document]
007DXR: Azure realtime plugin doesn't support generate_reply with tool_choice and tools now

007DXR:
When running the Healthcare Agent example with the Azure VoiceLive realtime plugin, the agent may enter an infinite loop. This Is NOT an Azure Plugin Bug but an architectural mismatch between the the Healthcare Agent design and how Azure VoiceLive manages sessions. 
May you choose another agent for testing? @tinalenguyen

007DXR:
> i get this error consistently with every `generate_reply`:
> 
> `failed to generate a reply after tool execution: pending response discarded due to session reconnection`
> 
> i think it has to do with how the connection is handled here. does azure time out often? it seems the plugin expects many reconnects to happen


## Why the Connection Drops During AgentTask Transitions

The core of the issue lies in how the session lifecycle is managed when the Agent moves from one task to another.


When the Agent switches tasks, the system checks if the new task is "equivalent" to the old one to see if it can reuse the existing connection.  In this specific implementation, **`is_equivalent` returns False**.


So the session cannot be reused. The system attempts to open a brand-new WebSocket connection for the next task.

In the context of the **Healthcare Agent**, this means that every single `AgentTask` transition triggers a full network cycle:
1.  **Disconnect:** The old WebSocket is killed.
2.  **Reconnect:** A new WebSocket is initialized.

### Conclusion
This Is NOT an Azure Plugin Bug but an architectural mismatch between the the Healthcare Agent design and how Azure VoiceLive manages sessions. 

May you choose another agent for testing?

monuminu:
Whats the latest update on this ?

007DXR:
> Whats the latest update on this ?

I recently fixed a few AI-suggested issues, but overall the code hasn’t changed much.

tinalenguyen:
hi @007DXR, my apologies for letting this slip! 

i think it would be better to connect to the client via a raw websocket versus the azure client package, i suspect that it would reduce latency and the issues i encountered

007DXR:
The Azure client package is actually the one officially recommended by Voice Live.
It essentially wraps the raw WebSocket connection.

gianpaj:
Would be great to merge this. I can see the model is GA now

image


https://github.com/MicrosoftDocs/azure-ai-docs/blob/main/articles/ai-services/speech-service/voice-live-webrtc.md

```
When initiating a WebRTC call session, use the voice-live/realtime/calls endpoint instead of voice-live/realtime. For example:

wss://<your-ai-foundry-resource-name>.services.ai.azure.com/voice-live/realtime/calls?api-version=2026-01-01-preview&model=gpt-realtime
```

BTW, I just tested in Firefox, and it sounds broken - but all the audio models do...

image

007DXR:
> 
> BTW, I just tested in Firefox, and it sounds broken - but all the audio models do...
> 


Thanks for the feedback! This issue seems unrelated to this PR.I'll report it to the relevant folks

livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/realtime/realtime_model.py
```diff
@@ -0,0 +1,1335 @@
...
+
+    @property
+    def model(self) -> str:
+        return self._opts.model
+
+    @property
+    def provider(self) -> str:
+        return "azure-voicelive"
+
+    def session(self) -> RealtimeSession:
```
tinalenguyen, line 258:
```suggestion
    def session(self, *, turn_detection_disabled: bool = False) -> RealtimeSession:
        # manual turn-taking is unsupported (can_disable_turn_detection=False)
```

livekit-plugins/livekit-plugins-azure/pyproject.toml
```diff
@@ -24,6 +24,8 @@ classifiers = [
 dependencies = [
     "livekit-agents[openai]>=1.8.2",
     "azure-cognitiveservices-speech>=1.48.2",
+    "azure-ai-voicelive[aiohttp]>=1.0.0",
+    "azure-identity>=1.15.0",
```
tinalenguyen, line 28: should we make these optional dependencies for the realtime module?

tinalenguyen (commented):
hi @007DXR, i tested the plugin and it works a lot better since released as GA. could you fix the ruff and type check tests and address the devin comments if applicable? i can merge your PR after and take over from there :) also left a few comments

tinalenguyen (approved):
thank you @007DXR for maintaining this PR, we really appreciate it!! i will take over from here

PR #4695, opened 2026-02-02
chore(docs): ditch the v0 docs and promote v1 docs to main path

rektdeckard:
We had previously removed links to old v0 reference documentation, but content was still hosted at `docs.livekit.io/reference/python` while v1 docs were at `docs.livekit.io/reference/python/v1`.

This Promotes v1 documentation to the base path (`docs.livekit.io/reference/python`) and removes logic for automated deployment of legacy docs.

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4695">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -37,7 +37,7 @@ jobs:
 
       - name: S3 Upload
         run: |
-          aws s3 sync docs/ s3://livekit-docs/python
+          aws s3 sync --delete docs/ s3://livekit-docs/python
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.DOCS_DEPLOY_AWS_ACCESS_KEY }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.DOCS_DEPLOY_AWS_API_SECRET }}


bcherry:
will v1 paths get automatically redirected on the backend?

rektdeckard:
@bcherry not automatically, that is done here: https://github.com/livekit/web/pull/3072

I ~will also make sure~ have already made sure on CloudFront side that between deploying this reference site and the main docs site, both forms of the URL are valid.

.github/workflows/publish-docs.yml
```diff
@@ -37,14 +37,7 @@ jobs:
...
         run: |
-          BRANCH="${GITHUB_REF#refs/heads/}"
-          # If branch is not 0.x, upload to /v1
-          if [[ $BRANCH != "0.x" ]]; then
-            aws s3 cp docs/ s3://livekit-docs/python/v1 --recursive
-          # Else, upload to root, preserving the v1 directory
-          else
-            aws s3 sync docs/ s3://livekit-docs/python --exclude "v1/*" --delete
-          fi
+          aws s3 sync docs/ s3://livekit-docs/python
```
davidzhao: do you need a --delete as well? to remove the older content

rektdeckard: I intentionally didn't add it at first, because there would be an awkward moment where, depending on whether the main docs site or this reference site is deployed first, either old links or new links without the `/v1/` path would break.

Instead I just copied the content so it exists in both places for now. As long as docs goes out before this workflow is run the first time, it's safe to delete older content.

Screenshot 2026-02-04 at 2 07 33 PM

davidzhao (approved):

PR #4707, opened 2026-02-04
fix(liveavatar): use server events to track avatar speaking state

MSameerAbbas:
Bug fixed: The original implementation set _audio_playing=False when AudioSegmentEnd was received (audio finished streaming to avatar), but the avatar may still be playing its buffer. This caused clear_buffer to skip sending agent.interrupt.

Fix: Use server events to track when the avatar is actually speaking.
- Replace single _audio_playing flag with two separate flags:
  - _audio_streaming: local state (sending frames to avatar)
  - _avatar_speaking: server state (avatar is actually playing)
- Add _on_server_event() to handle agent.speak_started/ended events
- Move notify_playback_finished() to be called on agent.speak_ended server event instead of on AudioSegmentEnd
- Send agent.speak_end on AudioSegmentEnd to signal end of audio stream

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4707">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/examples/avatar_agents/liveavatar/agent_worker.py
+++ b/examples/avatar_agents/liveavatar/agent_worker.py
@@ -24,7 +24,8 @@ async def entrypoint(ctx: JobContext):
         resume_false_interruption=False,
     )
 
-    liveavatar_avatar_id = os.getenv("LIVEAVATAR_AVATAR_ID")
+    # Default to a public male avatar if not specified
+    liveavatar_avatar_id = os.getenv("LIVEAVATAR_AVATAR_ID", "7b888024-f8c9-4205-95e1-78ce01497bda")
     avatar = liveavatar.AvatarSession(avatar_id=liveavatar_avatar_id)
     await avatar.start(session, room=ctx.room)
 


MSameerAbbas:
@tinalenguyen @chenghao-mou  I'm ngl I don't know if this should be merged tbh. It's probably not the prettiest solution. I will however let you know my motivation behind this. Tldr is that we send liveavatar the audio super quickly however the actual audio takes it's sweet time to complete. I figured this out because in my last PR, I added:
```python
                elif isinstance(audio_frame, AudioSegmentEnd):
                    if self._audio_playing:
                        self._audio_playing = False
                            interrupted=False,
                        )
                        self._playback_position = 0.0
```
which sets `self._audio_playing = False`. This however interferes with interruptions since this
```python
    def _on_clear_buffer(self) -> None:
        async def _handle_clear_buffer(audio_playing: bool) -> None:
            if audio_playing:
                    interrupted=True,
                )
                self.send_event({"type": "agent.interrupt", "event_id": str(uuid.uuid4())})
                self._playback_position = 0.0

        clear_buffer_task = asyncio.create_task(_handle_clear_buffer(self._audio_playing))
        self._tasks.add(clear_buffer_task)
        clear_buffer_task.add_done_callback(self._tasks.discard)
        self._audio_playing = False
```
checks to see if `self._audio_playing` is true. 

**Here's what happens**:

LiveKit sends all the audio frames to LiveAvatar nigh instantaneously, setting `self._audio_playing = False` the moment the last frame is sent. Any interruptions due to the user speaking after the first couple of milliseconds fail because of this - the avatar is still physically speaking, but our code thinks it's done.

I attempted to refactor the implementation to separate these two distinct states:

1. `_audio_streaming` - Tracks whether we're currently sending audio frames to the avatar (the fast part)
2. `_avatar_speaking` - Tracks whether the avatar is actually speaking (the slow part, based on server events)
The key change is that interruption logic now checks `_avatar_speaking` instead of the old `_audio_playing` flag. This state is set/unset based on actual WebSocket events from the LiveAvatar server:
- `agent.speak_started` -> `_avatar_speaking = True`
- `agent.speak_ended` -> `_avatar_speaking = False` (and only THEN do we call `notify_playback_finished`)
Instead of calling `notify_playback_finished` immediately on `AudioSegmentEnd` (when frames are done sending), we now wait for the `agent.speak_ended` event from the server (when the avatar actually finishes speaking). This ensures the interruption window stays open for the entire duration the avatar is speaking, not just the brief moment we're streaming frames.

livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/avatar.py
```diff
@@ -259,7 +300,13 @@ async def _send_task() -> None:
...
             while True:
                 msg = await ws_conn.receive()
-                if msg.type in (
+                if msg.type == aiohttp.WSMsgType.TEXT:
+                    try:
+                        event = json.loads(msg.data)
+                        self._on_server_event(event)
+                    except json.JSONDecodeError:
+                        logger.warning(f"Failed to parse server event: {msg.data}")
+                elif msg.type in (
```
chenghao-mou: nitpicking: we should exit early with the following pattern:

```python
if closing unexpectedly:
   raise exception
if not text message:
	raise exception

try:
	parsing
except:
   ...
```

MSameerAbbas:
Added a random male avatar fallback and switched to using early return (although I'm using continue rn for non-text msg types to be lenient. If you want I can raise an exception instead). You're probably right about the insta CLOSED message being a concurrency limit or some kind of liveavatar server side issue.

PR #4724, opened 2026-02-05
Fix LiveKit OpenAI Realtime plugin for proxy API compatibility

tg-bomze:
## Summary
- Patched `livekit-plugins-openai` realtime module to support non-direct OpenAI API providers (proxies like LiteLLM)
- Fixed WebSocket binary message handling and event name mapping between LiveKit and API
- Removed unsupported session parameters that caused API errors

## Changes
- Handle binary WebSocket messages (decode as UTF-8 JSON)
- Map `conversation.item.created` → `conversation.item.added`
- Map `response.audio.*` → `response.output_audio.*`
- Map `response.text.*` → `response.output_text.*`
- Remove unsupported `session.update` fields: `type`, `audio`, `max_output_tokens`
- Transform `output_modalities` → `modalities` (e.g., `output_text` → `text`)
- Transform content type `output_text` → `text` in `conversation.item.create`
- Ignore non-critical `ChanClosed` and `input_audio_buffer_commit_empty` errors

## Test plan
- [ ✔ ] Test voice call with direct OpenAI credentials
- [ ✔ ] Test voice call with proxy credentials
- [ ✔ ] Verify no errors in logs during conversation

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4724" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -746,6 +746,16 @@ class RealtimeSession(
                             by_alias=True, exclude_unset=True, exclude_defaults=False
                         )
 
+                    if ev.get("type") == "session.update" and "session" in ev:
+                        ev["session"].pop("type", None)
+                        ev["session"].pop("audio", None)
+                        ev["session"].pop("max_output_tokens", None)
+                        if "output_modalities" in ev["session"]:
+                            ev["session"]["modalities"] = [
+                                m.replace("output_", "")
+                                for m in ev["session"].pop("output_modalities")
+                            ]
+
                     self.emit("openai_client_event_queued", ev)
                     await ws_conn.send_str(json.dumps(ev))
             except Exception as e:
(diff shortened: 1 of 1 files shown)


davidzhao:
this is not the right. the realtime module is implementing OAI's v1 API. if LLM doesn't support it, you should be using the [beta API implementation](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model_beta.py)

tg-bomze:
> this is not the right. the realtime module is implementing OAI's v1 API. if LLM doesn't support it, you should be using the [beta API implementation](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model_beta.py)

Do you mean that I need to make my changes to realtime_model_beta.py? It's just that when I try to use Realtime Model with credentials from LiteLLM, I get an error (_livekit.agents.llm.realtime.RealtimeError: update_chat_ctx timed out_). It happens in both RealtimeModel and RealtimeModelBeta.

davidzhao:
`RealtimeModelBeta` should support that format. If it's not working, then you'd need to debug to see what LiteLLM is doing.. it's working well for OpenAI/Azure

PR #4730, opened 2026-02-06
gracefully stop AgentTask and parent agents when session close

longcw:
- clean up the old agent's activity when the session is closed during an AgentTask
- fix deadlock issue when session closes during tool execution but before waiting for AgentTask

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4730" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -786,7 +786,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
             while activity and isinstance(agent_task := activity.agent, AgentTask):
                 # notify AgentTask to complete and wait it to resume the parent agent
                 agent_task.cancel()
-                await agent_task.wait_for_inactive()
+                await agent_task._wait_for_inactive()
 
                 if old_agent := agent_task._old_agent:
                     activity = old_agent._activity
--- a/livekit-agents/livekit/agents/voice/agent.py
+++ b/livekit-agents/livekit/agents/voice/agent.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-import functools
 import time
 from collections.abc import AsyncGenerator, AsyncIterable, Coroutine, Generator
 from dataclasses import dataclass
@@ -701,9 +700,6 @@ class AgentTask(Agent, Generic[TaskResult_T]):
             return
         self.complete(ToolError(f"AgentTask {self.id} is cancelled"))
 
-    async def wait_for_inactive(self) -> None:
-        await self.__inactive_ev.wait()
-
     def complete(self, result: TaskResult_T | Exception) -> None:
         if self.__fut.done():
             raise RuntimeError(f"{self.__class__.__name__} is already done")
@@ -784,7 +780,6 @@ class AgentTask(Agent, Generic[TaskResult_T]):
             old_allow_interruptions = speech_handle.allow_interruptions
             speech_handle.allow_interruptions = False
 
-        self.__inactive_ev.clear()
         blocked_tasks = [current_task]
         if (
             old_activity._on_enter_task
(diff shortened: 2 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -638,6 +645,9 @@ async def _traceable_on_exit() -> None:
             await self._agent.on_exit()
 
         async with self._lock:
+            with contextlib.suppress(asyncio.InvalidStateError):
+                self._drain_fut.set_result(None)
+
```
chenghao-mou: should we just call `self._mark_draining()` here?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -784,35 +784,45 @@ async def _aclose_impl(
...
+                await activity.aclose()
 
-                # detach the inputs and outputs
-                self.input.audio = None
-                self.input.video = None
-                self.output.audio = None
-                self.output.transcription = None
+                # close the parent agent if the current one is an AgentTask
+                if isinstance(activity.agent, AgentTask) and activity.agent._old_agent:
+                    activity = activity.agent._old_agent._activity
```
chenghao-mou: Should we break explicitly here to avoid running checks on the same activity?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1070,12 +1080,12 @@ async def _update_activity(
...
                 if previous_activity == "close":
-                    await self._activity.drain()
-                    await self._activity.aclose()
+                    await activity.drain()
+                    await activity.aclose()
                 elif previous_activity == "pause":
-                    await self._activity.pause(blocked_tasks=blocked_tasks or [])
+                    await activity.pause(blocked_tasks=blocked_tasks or [])
 
             self._activity = self._next_activity
```
chenghao-mou, line 1098: does it matter if there is a concurrent `_aclose_impl` call?

longcw: this is a good question. 

between the old activity close/pause to the new activity start/resume is sync codes, the closing guard is added in between to allow to close/pause the old activity since it has a lock internally and it's safe to close an activity twice.

if the `session.aclose` is called during next activity start/resume, the `session._activity` is already changed to the next activity, the aclose impl will also try to close the next activity, and since the start/resume acquired lock first, it will be started/resumed then closed immediately.

btw, I moved the closing guard to before adding agent handoff items so it won't be recorded as an handoff.

chenghao-mou: One follow-up: we could still have double calls to `drain` though, right? Should we guard that as well?

longcw: that's a good point, updated to avoid duplicated calls to the `on_exit` in `drain`.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -782,35 +782,47 @@ async def _aclose_impl(
...
+                    self.output.audio = None
+                    self.output.transcription = None
+
+                    if (
+                        reason != CloseReason.ERROR
+                        and (audio_recognition := activity._audio_recognition) is not None
+                    ):
+                        # wait for the user transcript to be committed
+                        audio_recognition.commit_user_turn(
+                            audio_detached=True, transcript_timeout=2.0
```
theomonnom: Can we move this code after the loop?

longcw: `audio_recognition.commit_user_turn` needs to be run on the first activity

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1068,12 +1079,21 @@ async def _update_activity(
...
                 elif previous_activity == "pause":
-                    await self._activity.pause(blocked_tasks=blocked_tasks or [])
+                    await activity.pause(blocked_tasks=blocked_tasks or [])
+
+            if self._closing and new_activity == "start":
+                # disallow starting a new activity when the session is closing
+                logger.warning(
+                    f"session is closing, skipping {new_activity} activity of {self._next_activity.agent.id}",
+                )
+                self._next_activity = None
```
chenghao-mou, line 1094: nit: should we drain and close the new next activity first?

longcw: when `new_activity` is `start`, the next activity is created `self._next_activity = AgentActivity(agent, self)` but not started, so no need to close it?

chenghao-mou: It should be fine I guess, unless we later change the AgentActivity init to consume some resources.

PR #4746, opened 2026-02-06
fix: realtime agent handoff race conditions

randombet:
## Summary

- **Early return after handoff in realtime pipeline**: After `update_agent()` triggers activity transition, the old code continued using `self._rt_session` for chat context updates and reply generation. Since `update_agent()` closes the old realtime session, this caused errors and broken agent transfers. Added early `return` to skip post-handoff processing.
- **Resume paused agent activity on transfer back**: When transferring back to an agent that was previously active, its activity was paused rather than closed. The code always started a new activity, losing the paused state. Now checks for existing activity and resumes it instead.
- **Enable manual function calls for Gemini realtime**: Set `manual_function_calls=True` so the framework controls tool execution flow rather than Gemini auto-executing, preventing race conditions during agent handoff where tool responses could be sent to the wrong session.

## Context

These bugs were discovered while building a multi-agent example with Gemini Realtime API. Transferring from main agent to a specialist worked, but transferring back consistently failed because:

1. The handoff tool returned an `AgentHandoff`, triggering `update_agent()` which starts closing the old realtime session
2. But the realtime tool output handler continued past the handoff, trying to call `self._rt_session.update_chat_ctx()` on the closing session
3. This caused the transfer to hang or error out silently

## Test plan

- [ ] Verify agent transfer works bidirectionally (main → specialist → main) with Gemini realtime
- [ ] Verify STT-LLM-TTS pipeline agent transfer is unaffected (fix #1 only touches the realtime path)
- [ ] Verify single-agent Gemini realtime sessions work normally (no regression from `manual_function_calls=True`)

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4746" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


tinalenguyen:
Hi @randombet, I am unable to reproduce this. Could you provide a reproducible snippet? Also curious as to how you set up your handoff function

PR #4758, opened 2026-02-07
allow flexible recording options

davidzhao:
when starting a session, we now support record=RecordOptions() where audio, traces, logs, and transcript could be individually enabled or disabled.

---

<a href="https://livekit.devinenterprise.com/review/livekit/agents/pull/4758" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -476,7 +502,7 @@ async def start(
         # deprecated
         room_input_options: NotGivenOr[room_io.RoomInputOptions] = NOT_GIVEN,
         room_output_options: NotGivenOr[room_io.RoomOutputOptions] = NOT_GIVEN,
-        record: NotGivenOr[bool] = NOT_GIVEN,
+        record: NotGivenOr[bool | RecordingOptions] = NOT_GIVEN,
```
chenghao-mou: nit: should we move record up (above the `#deprecated` comment)?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -502,16 +528,26 @@ async def start(
...
+            if isinstance(record, RecordingOptions):
+                recording_options = record
+            elif isinstance(record, bool):
+                recording_options = RecordingOptions(
+                    audio=record, traces=record, logs=record, transcript=record
+                )
+            else:
+                recording_options = RecordingOptions(
+                    audio=False, traces=False, logs=False, transcript=False
+                )
```
chenghao-mou:
```suggestion
            else:
                recording_options = RecordingOptions(
                    audio=bool(record),
                    traces=bool(record),
                    logs=bool(record),
                    transcript=bool(record)
                )
```
since `bool(NOT_GIVEN)` is False.

tests/test_recording_options.py
```diff
@@ -0,0 +1,406 @@
...
+    report = _make_mock_report(
+        RecordingOptions(audio=True, traces=False, logs=False, transcript=False)
+    )
+    report.audio_recording_path = None
+    mock_http = _make_mock_http()
+
+    with _patch_upload_deps():
+        await _call_upload(report, http_session=mock_http)
+
+    mock_http.post.assert_called_once()
```
chenghao-mou: I think this fails because we exit early when: 
```python
if not recording_options.transcript and not has_audio:
        return
```

livekit-agents/livekit/agents/voice/agent_session.py
chenghao-mou:
```suggestion
            record: Whether to record the audio, transcripts, traces, or logs
```

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -63,6 +63,30 @@
     from .transcription.filters import TextTransforms
 
 
+@dataclass
+class RecordingOptions:
```
chenghao-mou: maybe we could have something like this: 

```python
    def __bool__(self) -> bool:
        return any(vars(self).values())
```

or `enabled` to be more explicit.

davidzhao: that's a good idea, I'll add this

livekit-agents/livekit/agents/ipc/proc_pool.py
```diff
@@ -89,7 +89,14 @@ async def start(self) -> None:
 
         if self._default_num_idle_processes > 0:
             # wait for the idle processes to be warmed up (by the main task)
-            await self._idle_ready.wait()
+            # use a timeout so start() doesn't block forever if initialization fails
+            try:
```
davidzhao, line 94: @theomonnom confirming that this should not block `.start` if processes take longer than init timeout

theomonnom: If initialization fails, other processes will try to be spawned again. So it shouldn't block forever even without a timeout

davidzhao: without this, [test_slow_initialization](https://github.com/livekit/agents/blob/dz/recording-options/tests/test_ipc.py#L270) could not start correctly... because I think each new process is hitting the timeout, so it was never able to reach capacity. wdty the right behavior should be?

tests/test_audio_decoder.py
```diff
@@ -17,6 +17,8 @@
 
 @pytest.mark.asyncio
 async def test_decode_and_transcribe():
+    if not os.environ.get("DEEPGRAM_API_KEY"):
+        pytest.skip("DEEPGRAM_API_KEY not set")
```
chenghao-mou:
```suggestion
@pytest.mark.asyncio
@pytest.mark.skipif(
    os.getenv("DEEPGRAM_API_KEY") is None,
    reason="DEEPGRAM_API_KEY not set",
)
async def test_decode_and_transcribe():
```

livekit-agents/livekit/agents/ipc/proc_pool.py
theomonnom: shouldn't  [await _wait_for_elements(start_q, num_idle_processes)](https://github.com/livekit/agents/blob/2e6f75c1d2b0b3613b7176634697a120731e4305/tests/test_ipc.py#L323C4-L323C58) already wait for those processes tho?

theomonnom: But I think you're right, we shouldn't wait here, can you ignore it for this PR? I can take over

theomonnom: It's used by the worker to make sure we don't register to LK Cloud until everything is ready

PR #4775, opened 2026-02-09
fix(elevenlabs): add validation warnings for out-of-range VoiceSettings

### LINKED ISSUE #4774: ElevenLabs VoiceSettings silently accepts out-of-range values — voice failed with no indication why
## Bug Description

Ran into this on 1.4.1 — had `speed` set to a value outside the valid range and spent some time figuring out why the voice failed. I found it pretty quickly since voice settings was the only thing I'd changed, but without that context it would've been a pain. The plugin doesn't validate or warn on any of the `VoiceSettings` parameters, so out-of-range values just get sent straight to the 11labs API which silently clamps them. No warning, no error, nothing in the logs — would be nice to improve the experience for future devs.

The Cartesia plugin already does this — `_check_generation_config()` logs warnings with `extra={}` for out-of-range speed/volume. The ElevenLabs plugin should do the same.

## Expected Behavior

A `logger.warning` when `VoiceSettings` params are outside their valid ranges:

- `stability`: 0.0 - 1.0
- `similarity_boost`: 0.0 - 1.0
- `style`: 0.0 - 1.0
- `speed`: 0.7 - 1.2

Should be warnings not exceptions — 11labs could change the ranges, but at least devs would have something in the logs to go on.

## Reproduction Steps

1. Set up ElevenLabs TTS with an out-of-range speed:
   ```python
   from livekit.plugins.elevenlabs import TTS, VoiceSettings

   tts = TTS(
       voice_settings=VoiceSettings(
           stability=0.5,
           similarity_boost=0.5,
           speed=0.5,  # below min of 0.7
       )
   )
   ```
2. Synthesize speech — voice failed
3. Check logs — nothing. No warning about the bad speed value
4. Same thing happens through `update_options()`

## Operating System

macOS Sequoia

## Models Used

ElevenLabs TTS (eleven_turbo_v2_5)

## Package Versions

```
livekit-plugins-elevenlabs==1.4.1
livekit-agents==1.4.1
```

## Proposed Solution

Add a `_validate_voice_settings()` function following the Cartesia `_check_generation_config()` pattern — `logger.warning()` with `extra={}` context for each param. Call it from `TTS.__init__()` and `TTS.update_options()` when voice_settings is provided.

Happy to put up a PR for this.

williamwarlick:
Closes #4774

## Summary

- Add `_validate_voice_settings()` that logs warnings when `VoiceSettings` params are outside their documented ElevenLabs API ranges — follows the same pattern as Cartesia's `_check_generation_config()`
- Call validation from both `TTS.__init__()` and `TTS.update_options()`
- Fix speed range comment: `[0.8 - 1.2]` -> `[0.7 - 1.2]` per ElevenLabs API docs

## Motivation

Out-of-range `VoiceSettings` values (e.g. `speed=0.5`) get sent straight to the 11labs API which silently clamps them. The voice fails and there's nothing in the logs to indicate why. The Cartesia plugin already handles this with `_check_generation_config()` — this brings the ElevenLabs plugin to the same level.

Warnings are non-blocking (`logger.warning` with `extra={}`) so values still pass through.

## Validated Ranges

| Parameter | Range |
|-----------|-------|
| `stability` | 0.0 - 1.0 |
| `similarity_boost` | 0.0 - 1.0 |
| `style` | 0.0 - 1.0 |
| `speed` | 0.7 - 1.2 |

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -69,7 +69,7 @@ class VoiceSettings:
     stability: float  # [0.0 - 1.0]
     similarity_boost: float  # [0.0 - 1.0]
     style: NotGivenOr[float] = NOT_GIVEN  # [0.0 - 1.0]
-    speed: NotGivenOr[float] = NOT_GIVEN  # [0.8 - 1.2]
+    speed: NotGivenOr[float] = NOT_GIVEN  # [0.7 - 1.2]
     use_speaker_boost: NotGivenOr[bool] = NOT_GIVEN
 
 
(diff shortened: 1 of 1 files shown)


longcw (approved):
looks good to me!

PR #4803, opened 2026-02-12
[inworld] use audio/pcm to leverage the fast AudioBytesStream path

ianbbqzy:
* strip wav header manually before pushing to emitter

### DIFF
--- a/livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
+++ b/livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
@@ -19,6 +19,7 @@ import base64
 import json
 import os
 import time
+import uuid
 import weakref
 from collections.abc import Callable
 from dataclasses import dataclass, field, replace
@@ -45,11 +46,14 @@ from livekit.agents.utils import is_given
 from livekit.agents.voice.io import TimedString
 
 from .log import logger
+from .version import __version__
+
+USER_AGENT = f"livekit-agents-py/{__version__}"
 
 DEFAULT_BIT_RATE = 64000
-DEFAULT_ENCODING = "OGG_OPUS"
-DEFAULT_MODEL = "inworld-tts-1"
-DEFAULT_SAMPLE_RATE = 48000
+DEFAULT_ENCODING = "LINEAR16"
+DEFAULT_MODEL = "inworld-tts-1.5-max"
+DEFAULT_SAMPLE_RATE = 24000
 DEFAULT_URL = "https://api.inworld.ai/"
 DEFAULT_WS_URL = "wss://api.inworld.ai/"
 DEFAULT_VOICE = "Ashley"
@@ -80,7 +84,13 @@ class _TTSOptions:
 
     @property
     def mime_type(self) -> str:
-        if self.encoding == "MP3":
+        if self.encoding == "LINEAR16":
+            # Use audio/pcm so the emitter takes the fast synchronous
+            # AudioByteStream path instead of the async AudioStreamDecoder.
+            # WAV headers from the server are stripped before pushing to the
+            # emitter (see _strip_wav_header).
+            return "audio/pcm"
+        elif self.encoding == "MP3":
             return "audio/mpeg"
         elif self.encoding == "OGG_OPUS":
             return "audio/ogg"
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
```diff
@@ -1109,6 +1119,18 @@ async def _send_task() -> None:
...
+def _strip_wav_header(data: bytes) -> bytes:
+    """Strip WAV header from audio data, returning raw PCM.
+
+    Inworld returns LINEAR16 audio wrapped in WAV containers. The emitter's
+    AudioByteStream fast-path requires raw PCM, so we strip the 44-byte
+    standard WAV header (RIFF + fmt + data chunk headers) when present.
+    """
+    if len(data) > 44 and data[:4] == b"RIFF":
+        return data[44:]
+    return data
```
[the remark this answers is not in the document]
ianbbqzy: For Inworld, we can assume it's fixed 44-bytes header

ianbbqzy:
cc @theomonnom  @longcw. Thanks!

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
```diff
@@ -47,9 +47,9 @@
 from .log import logger
 
 DEFAULT_BIT_RATE = 64000
-DEFAULT_ENCODING = "OGG_OPUS"
-DEFAULT_MODEL = "inworld-tts-1"
-DEFAULT_SAMPLE_RATE = 48000
+DEFAULT_ENCODING = "LINEAR16"
+DEFAULT_MODEL = "inworld-tts-1.5-max"
+DEFAULT_SAMPLE_RATE = 24000
```
davidzhao, line 56: this helps!

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
```diff
@@ -80,7 +80,13 @@ class _TTSOptions:
 
     @property
     def mime_type(self) -> str:
-        if self.encoding == "MP3":
+        if self.encoding == "LINEAR16":
+            # Use audio/pcm so the emitter takes the fast synchronous
+            # AudioByteStream path instead of the async AudioStreamDecoder.
+            # WAV headers from the server are stripped before pushing to the
+            # emitter (see _strip_wav_header).
+            return "audio/pcm"
```
davidzhao, line 92: if it's wav file sent back.. you should just pass back `audio/wav`, our decoder has a fast path for decoding wav files. 

this would be preferred rather than having multiple wav handling in the code base

ianbbqzy: I believe that's the existing behavior. When encoding is `LINEAR16`, it falls to the else branch of `audio/wav`. But I measured that to have ~30-40ms additional latency.

ianbbqzy: @tinalenguyen would you be able to advise here?

I tested with my own websocket benchmark script at https://github.com/inworld-ai/inworld-api-examples/tree/ian/livekit-integrations/integrations/livekit/python/benchmarks with 100+ iterations

summarized by AI, the reason is that

| | `audio/pcm` (raw) | `audio/wav` (decoder) |
|---|---|---|
| Threading | None — all in main loop | `ThreadPoolExecutor` + `StreamBuffer` with locks |
| Cross-thread hops | **0** | **2** (main→thread via StreamBuffer, thread→main via `call_soon_threadsafe`) |
| Event loop cycles to first frame | **1** | **3+** (push → thread wake → thread decode → call_soon_threadsafe → decode_task scheduled → decode_task runs) |
| AudioByteStream instances | 1 | **2** (one in `_decode_wav_loop`, one in `_decode_task`) |
| Buffer copies per read | 0 | `StreamBuffer.read()` creates new `BytesIO` + copies remaining bytes every read |

davidzhao: > When encoding is LINEAR16, it falls to the else branch of audio/wav. But I measured that to have ~30-40ms additional latency.

could you share the benchmark scripts that you had ran?

if AudioStreamDecoder is slow, we should optimize that instead. I still maintain that we should not be duplicating decoding logic within plugin code

ianbbqzy: Hi David, can you access this script https://github.com/inworld-ai/inworld-api-examples/tree/ian/livekit-integrations/integrations/livekit/python/benchmarks ? I have the instructions to run in the README, you should be able to checkout this feature branch in the submodule to see the difference.

We might also consider a new Audio encoding format `PCM` for headerless audio bytes to differentiate from the current encoding format on the server side

ianbbqzy: Hi @davidzhao  @tinalenguyen, curious if you were able to reproduce and if there is any plan for optimization of AudioStreamDecoder? I do think that would be ideal to speed up every other encoding formats besides WAV

davidzhao:
@ianbbqzy you should be able to base it on [this branch](https://github.com/livekit/agents/pull/4905), where it uses the optimized path for WAV files

PR #4806, opened 2026-02-12
fix: preserve MCP CallToolResult.meta field in tool output

IanSteno:
## Summary

- The MCP spec (since v1.10.0) supports a `meta` field on `CallToolResult` for returning sideband metadata from tool executions
- The SDK already passes `tool.meta` for tool **definitions** (line 85, 133-134 in mcp.py), but `tool_result.meta` is discarded when processing tool call **results**
- This preserves the `meta` field by including it in the returned JSON string

## Problem

When an MCP server returns a `CallToolResult` with a `meta` field:

```python
CallToolResult(
    content=[TextContent(type="text", text='{"results": [...]}')],
    meta={"notify": "data_updated", "data": {"affected_ids": [1, 2, 3]}}
)
```

The current `_tool_called` function only extracts `.content` and returns it as JSON, completely discarding `.meta`. This means consumers of tool results (e.g., event handlers listening to `FunctionToolsExecutedEvent`) cannot access any metadata the MCP server intended to communicate.

## Fix

After extracting content into `result`, check if `tool_result.meta` is present and merge it into the returned JSON:

```python
if getattr(tool_result, "meta", None) is not None:
    parsed = json.loads(result)
    if isinstance(parsed, dict):
        parsed["meta"] = tool_result.meta
    elif isinstance(parsed, list):
        parsed = {"content": parsed, "meta": tool_result.meta}
    result = json.dumps(parsed)
```

This is backward-compatible — if `meta` is `None` (the default), behavior is unchanged.

## Use Cases

The `meta` field enables MCP servers to return sideband data that consumers can intercept before it reaches the LLM:
- **Frontend notifications**: Signal UI updates (e.g., "requirements updated")
- **Citation tracking**: Source attribution metadata
- **Billing/metrics**: Token usage, cost data
- **Cache hints**: TTL or invalidation signals

## Test plan

- [x] Verified no behavior change when `tool_result.meta` is `None` (default case)
- [x] Verified single content item + meta produces correct JSON
- [x] Verified multiple content items + meta wraps in `{"content": [...], "meta": {...}}`
- [x] Verified error results still raise `ToolError` (meta check is after error handling)

### DIFF
--- a/livekit-agents/livekit/agents/llm/mcp.py
+++ b/livekit-agents/livekit/agents/llm/mcp.py
@@ -116,14 +116,25 @@ class MCPServer(ABC):
 
             # TODO(theomonnom): handle images & binary messages
             if len(tool_result.content) == 1:
-                return tool_result.content[0].model_dump_json()
+                result = tool_result.content[0].model_dump_json()
             elif len(tool_result.content) > 1:
-                return json.dumps([item.model_dump() for item in tool_result.content])
+                result = json.dumps([item.model_dump() for item in tool_result.content])
+            else:
+                raise ToolError(
+                    f"Tool '{name}' completed without producing a result. "
+                    "This might indicate an issue with internal processing."
+                )
 
-            raise ToolError(
-                f"Tool '{name}' completed without producing a result. "
-                "This might indicate an issue with internal processing."
-            )
+            # Preserve the meta field from CallToolResult if present
+            if getattr(tool_result, "meta", None) is not None:
+                parsed = json.loads(result) if isinstance(result, str) else result
+                if isinstance(parsed, dict):
+                    parsed["meta"] = tool_result.meta
+                elif isinstance(parsed, list):
+                    parsed = {"content": parsed, "meta": tool_result.meta}
+                result = json.dumps(parsed)
+
+            return result
 
         raw_schema = {
             "name": name,


IanSteno:
Quick refresh from my side on 2026-02-18: I revalidated this change against current `main` and it remains minimal (single-file MCP metadata preservation) with no local lint/format issues in the touched file.

Would appreciate a maintainer review when you have bandwidth.

livekit-agents/livekit/agents/llm/mcp.py
```diff
@@ -116,14 +116,25 @@ async def _tool_called(raw_arguments: dict[str, Any]) -> Any:
...
 
-            raise ToolError(
-                f"Tool '{name}' completed without producing a result. "
-                "This might indicate an issue with internal processing."
-            )
+            # Preserve the meta field from CallToolResult if present
+            if getattr(tool_result, "meta", None) is not None:
+                parsed = json.loads(result) if isinstance(result, str) else result
+                if isinstance(parsed, dict):
+                    parsed["meta"] = tool_result.meta
```
longcw, line 132: usually the metadata is not for LLM, instead of passing it output directly, I think it makes more sense to expose a `MCPToolResultResolver` that user can customize for reading the metadata. created a pr https://github.com/livekit/agents/pull/5046

longcw:
replaced by #5046

PR #4813, opened 2026-02-13
add AEC warmup to suppress false interruptions on first speech

longcw:
when `aec_warmup_duration` is set, it blocks interruptions for a few seconds after the agent starts speaking to allow client to calibrate AEC.

this only blocks the audio input when agent is speaking, and disable the interruption from audio input, `session.interrupt` still works.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -297,7 +297,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
             aec_warmup_duration (float, optional): The duration in seconds that the agent
                 will ignore user's audio interruptions after the agent starts speaking.
                 This is useful to prevent the agent from being interrupted by echo before AEC is ready.
-                Default ``None``.
+                Set to ``None`` to disable. Default ``3.0`` s.
             ivr_detection (bool): Whether to detect if the agent is interacting with an IVR system.
                 Default ``False``.
             conn_options (SessionConnectOptions, optional): Connection options for


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -246,6 +248,10 @@ def __init__(
                 can reduce response latency by overlapping model inference with user audio,
                 but may incur extra compute if the user interrupts or revises mid-utterance.
                 Defaults to ``False``.
+            echo_guard_duration (float, optional): The duration in seconds that the agent
+                will ignore user's audio interruptions after the agent starts speaking.
```
chenghao-mou, line 298: Q: Does this apply to both cases when the session starts:
1. Agent speaks first
2. Agent's first response (the user might speak first)

longcw: yes, it only considers agent speaking, no matter who speaking first.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -158,6 +159,7 @@ def __init__(
         use_tts_aligned_transcript: NotGivenOr[bool] = NOT_GIVEN,
         tts_text_transforms: NotGivenOr[Sequence[TextTransforms] | None] = NOT_GIVEN,
         preemptive_generation: bool = False,
+        echo_guard_duration: float | None = None,
```
theomonnom: I think we should do it by default.
Maybe we shouldn't even have an option for it, it seems like an issue everybody has

longcw: how long does it usually need to warm up the AEC?

chenghao-mou: I think having an option is fine? Or we could remove the option in the code but allow ENV var override?

Claude told me there is a way to check if RTC stats `echo_return_loss` converges, but we need this from the capture side. Sounds more like an overkill.

theomonnom: It's for telephony use cases. So we can't even get those stats. 
I think an avg of 4-5seconds doesn't seems to be that opinionated IMO

chenghao-mou: I believe we have a similar issue with WebRTC calls. In my test with the debug panel, overlapping speeches were often missed on the waveform in the beginning.

longcw: simplified the timer to a one-shot countdown that starts on first agent turn, and defaulted it to 3.0s.

PR #4819, opened 2026-02-14
(openai responses): update field names and image inputs

tinalenguyen:
no breaking changes, the documentation just displays different field names

https://developers.openai.com/api/reference/resources/responses#(resource)%20responses%20%3E%20(model)%20response_input_message_content_list%20%3E%20(schema)

### DIFF
--- a/livekit-agents/livekit/agents/llm/_provider_format/openai.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/openai.py
@@ -119,14 +119,14 @@ def _to_responses_image_content(image: llm.ImageContent) -> dict[str, Any]:
     img = llm.utils.serialize_image(image)
     if img.external_url:
         return {
-            "type": "image_url",
+            "type": "input_image",
             "image_url": img.external_url,
             "detail": img.inference_detail,
         }
     assert img.data_bytes is not None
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/llm/_provider_format/openai.py
```diff
@@ -153,31 +170,16 @@ def _to_responses_chat_item(msg: llm.ChatItem) -> dict[str, Any]:
...
-                    "id": msg.call_id,
-                    "type": "function",
-                    "function": {
-                        "name": msg.name,
-                        "arguments": msg.arguments,
-                    },
-                }
-            ],
-        }
-
```
theomonnom, line 180: is this removed on purpose?

davidzhao: this shouldn't be removed?

tinalenguyen: that part of the code is never accessed, `"function_call"` types are handled here in `to_responses_chat_ctx`: https://github.com/livekit/agents/blob/02e49a7125440fdb285827fdfb5dd24bfca87662/livekit-agents/livekit/agents/llm/_provider_format/openai.py#L131-L138

PR #4820, opened 2026-02-14
ensure proc pool spawns processes for waiting jobs under high load

theomonnom:


livekit-agents/livekit/agents/ipc/proc_pool.py
```diff
@@ -196,10 +202,11 @@ async def _main_task(self) -> None:
...
             while not self._closed:
                 current_pending = self._warmed_proc_queue.qsize() + len(self._spawn_tasks)
-                to_spawn = (
-                    min(self._target_idle_processes, self._default_num_idle_processes)
-                    - current_pending
+                target = max(
+                    min(self._target_idle_processes, self._default_num_idle_processes),
+                    self._jobs_waiting_for_process,
                 )
+                to_spawn = target - current_pending
```
davidzhao, line 209: could this be negative? then would range throw?

theomonnom: You will get an empty iterator

theomonnom:
```
>>> print(list(range(-20)))
[]
```

davidzhao (approved):

PR #4824, opened 2026-02-15
fix launch_job send failure leaving executor orphaned

theomonnom:
- also add retry to launch_job on IPC failures


livekit-agents/livekit/agents/ipc/proc_pool.py
```diff
@@ -99,27 +100,45 @@ async def aclose(self) -> None:
...
+
+        for attempt in range(MAX_ATTEMPTS):
+            proc = await _wait_for_proc()
+            try:
+                await proc.launch_job(info)
+                self.emit("process_job_launched", proc)
+                return
+            except Exception:
+                logger.warning(
+                    "failed to launch job on process, retrying with another",
```
davidzhao: on the last attempt, it should not log that it's retrying

PR #4827, opened 2026-02-15
chore(assemblyai): improve latency by default

davidzhao:
AssemblyAI has a higher server-side latency of 400ms before final transcripts are sent back. This is to avoid the transcript arriving in multiple chunks.

However, a higher latency by default also increases the amount of time the pipeline has to wait between turns.

Since we can handle transcripts arriving in chunks, reducing the defaults to optimize for latency


chenghao-mou (approved):
lgtm

PR #4828, opened 2026-02-15
fix: a few defensive fixes to guard for exceptions

davidzhao:


livekit-agents/livekit/agents/utils/connection_pool.py
```diff
@@ -65,8 +66,12 @@ async def _connect(self, timeout: float) -> T:
...
         """Drain and close all the connections queued for closing."""
-        for conn in list(self._to_close):
-            await self._maybe_close_connection(conn)
+        while self._to_close:
+            conn = self._to_close.pop()
+            try:
+                await self._maybe_close_connection(conn)
+            except Exception as e:
+                logger.warning(f"error closing connection: {conn}", exc_info=e)
         self._to_close.clear()
```
chenghao-mou: nit: do we still need this line after the while loop?

chenghao-mou (approved):
lgtm

livekit-agents/livekit/agents/utils/connection_pool.py
davidzhao: you are right, this isn't needed

PR #4829, opened 2026-02-15
feat(tokenize): add clause-level sentence tokenizer with multi-languangual.

SezginKahraman:
Add a new clause-level tokenizer that splits text at clause boundaries (commas, conjunctions, sentence endings) instead of only at sentence endings. This produces finer-grained segments for TTS, reducing perceived latency without sacrificing naturalness.

Key features:
- Multi-language support via LanguageProfile (English and Turkish built-in)
- Protected position detection (abbreviations, URLs, emails, quoted text, number formats, time formats)
- Conjunction-based clause splitting (including multi-word conjunctions)
- Drop-in replacement for basic.SentenceTokenizer
- Extensible: users can define custom LanguageProfile for any language

### DIFF
--- a/livekit-agents/livekit/agents/tokenize/_clause_sent.py
+++ b/livekit-agents/livekit/agents/tokenize/_clause_sent.py
@@ -429,7 +429,7 @@ def _find_split_positions(
     if profile.conjunctions:
         sorted_conjs = sorted(profile.conjunctions, key=len, reverse=True)
         conj_alts = "|".join(re.escape(c) for c in sorted_conjs)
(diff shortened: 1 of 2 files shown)


SezginKahraman:
pre-existing test errors.

davidzhao:
this isn't robust enough for us to include in the framework. When TTS doesn't have sufficient context, it would impact the quality of output that ends up getting synthesized. For example, splitting at conjugations doesn't seem like a good idea and will definitely change the tone of the generation.

you are welcome to use this approach for your own applications as a special tokenizer passed to the StreamAdapter

PR #4832, opened 2026-02-15
fix(job): fix lambda closure capture and init failure tracking

SezginKahraman:
- Fix race condition in participant task cleanup where lambda captured p.identity by reference instead of by value, causing the callback to use a potentially stale identity if the participant object changes
- Track job initialization failures in ThreadJobExecutor: log warnings for timeout/exception cases and set job status to FAILED instead of incorrectly reporting SUCCESS

### DIFF
--- a/livekit-agents/livekit/agents/job.py
+++ b/livekit-agents/livekit/agents/job.py
@@ -615,7 +615,7 @@ class JobContext:
             task = asyncio.create_task(coro(self, p), name=task_name)
             self._participant_tasks[(p.identity, coro)] = task
             task.add_done_callback(
-                lambda _, coro=coro: self._participant_tasks.pop((p.identity, coro))  # type: ignore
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/job.py
```diff
@@ -615,7 +615,7 @@ def _participant_available(self, p: rtc.RemoteParticipant) -> None:
             task = asyncio.create_task(coro(self, p), name=task_name)
             self._participant_tasks[(p.identity, coro)] = task
             task.add_done_callback(
-                lambda _, coro=coro: self._participant_tasks.pop((p.identity, coro))  # type: ignore
```
davidzhao, line 618: I don't think it's possible for a particular participant obj to change its identity.. why is this needed?

davidzhao (commented):
the job status change makes sense. could you run `make fix` to clean up the formatting issues?

PR #4833, opened 2026-02-15
fix: correctness and performance improvements

SezginKahraman:
- Remove deprecated loop parameter from asyncio.Future() in worker.py (deprecated since Python 3.8, removed in 3.10+)
- Guard against negative sleep duration in agent_activity speech playout delay calculation
- Simplify redundant volume formula: 10^(log10(v)) is just v
- Fix O(n²) chat context truncation: replace while/pop(0) with index-based slicing
- Fix "pasuing" → "pausing" typo in comment


davidzhao:
please do not open PRs without an accompanying issue. having issues with clear description would help us better understand what's going on.

davidzhao:
ChatContext truncate behavior has been fixed in https://github.com/livekit/agents/pull/4845

PR #4835, opened 2026-02-15
fix: STT fallback does not skip main_stream when recovering streams fail

davidzhao:


livekit-agents/livekit/agents/stt/fallback_adapter.py
```diff
@@ -291,22 +291,25 @@ async def _run(self) -> None:
...
             async for data in self._input_ch:
-                try:
-                    for stream in self._recovering_streams:
+                for stream in list(self._recovering_streams):
+                    try:
                         if isinstance(data, rtc.AudioFrame):
                             stream.push_frame(data)
                         elif isinstance(data, self._FlushSentinel):
                             stream.flush()
+                    except RuntimeError:
```
chenghao-mou: Should we also catch `Exception` here for `ChanClosed` or `ChanFull` exceptions?

tests/test_stt_fallback.py
chenghao-mou: we should rename the two`tts` occurrences too.

PR #4846, opened 2026-02-16
upload logs to server even when session fails to start

davidzhao:
if there's a crash earlier on in the session, we do not correctly upload logs to the server. this makes it difficult to troubleshoot.

we'll always upload crash logs but still honor user-specified record settings.

### DIFF
--- a/livekit-agents/livekit/agents/ipc/proc_pool.py
+++ b/livekit-agents/livekit/agents/ipc/proc_pool.py
@@ -217,6 +217,7 @@ class ProcPool(utils.EventEmitter[EventTypes]):
         if not initialized:
             self._executors.remove(proc)
             await proc.aclose()
+            self.emit("process_closed", proc)
             return
 
         monitor_task = asyncio.create_task(self._monitor_process_task(proc))
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -1193,10 +1193,10 @@ class AgentServer(utils.EventEmitter[EventTypes]):
             )
             if args.attributes:
                 availability_resp.availability.participant_attributes.update(args.attributes)
-            await self._queue_msg(availability_resp)
 
             wait_assignment = asyncio.Future[agent.JobAssignment]()
             self._pending_assignments[job_req.id] = wait_assignment
+            await self._queue_msg(availability_resp)
 
             # the job was accepted by the user, wait for the server assignment
             try:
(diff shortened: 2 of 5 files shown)


livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -503,7 +503,12 @@ def _shutdown_telemetry() -> None:
...
 
     if isinstance(logger_provider := get_logger_provider(), LoggerProvider):
-        # force_flush will cause deadlock when new logs from OTLPLogExporter are emitted
-        # logger_provider.force_flush()
-        logger.debug("shutting down telemetry logger provider")
+        # remove the OTLP LoggingHandler before flushing to avoid deadlock —
+        # force_flush triggers log export which emits new logs back through the handler
+        root = logging.getLogger()
+        for h in root.handlers[:]:
+            if isinstance(h, LoggingHandler):
```
theomonnom, line 521: Uh, this was very confusing to read. Would be better if it was otel_logger.LoggingHandler

livekit-agents/livekit/agents/job.py
theomonnom, line 1: would be better if this logic lives inside traces.py

where everything otel related happens

livekit-agents/livekit/agents/job.py
```diff
@@ -604,6 +669,7 @@ def init_recording(self, options: RecordingOptions) -> None:
                 enable_traces=options["traces"],
                 enable_logs=options["logs"],
             )
+            self._flush_early_log_buffer(replay=True)
```
theomonnom, line 664: Do we need this early log buffer only because `init_recording` is called too late? (can crash before).

Can you add a comment for that 🙏

theomonnom (approved):
otherwise lgtm

livekit-agents/livekit/agents/job.py
davidzhao: all recording related settings are after session.start().. before then everything is dropped

livekit-agents/livekit/agents/job.py
davidzhao: fair, let me move it

livekit-agents/livekit/agents/telemetry/traces.py
davidzhao: the comment already indicates it's talking about otel LoggingHandler, it's already used in https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/telemetry/traces.py#L217

PR #4851, opened 2026-02-16
fix: call speech_handle.add_done_callback even when task is done

### LINKED ISSUE #4849: [Detail Bug] SpeechHandle.add_done_callback ignores callbacks registered after completion, causing hangs
# Summary
- **Context**: `SpeechHandle` manages the lifecycle and callbacks for speech generation within the VoiceAgent framework.
- **Bug**: `SpeechHandle.add_done_callback` does not execute the provided callback if the `SpeechHandle` is already in a "done" state when the callback is registered.
- **Actual vs. expected**: Currently, the callback is only added to an internal set that is iterated once when the handle completes; if it's already complete, the iteration has already happened and the new callback is ignored. It should instead execute the callback immediately (or schedule it) if the handle is already done, matching the behavior of `asyncio.Future.add_done_callback`.
- **Impact**: This causes silent hangs in critical components like `AgentActivity.interrupt()` and `RunResult`, which wait for all speech handles to complete via these callbacks. If a speech handle completes before the callback is registered, the waiter will hang indefinitely.

# Code with bug
```python
    def add_done_callback(self, callback: Callable[[SpeechHandle], None]) -> None:
        self._done_callbacks.add(callback)  # <-- BUG 🔴 [Does not check if already done]
```

# Evidence

### Test script
```python
import asyncio
import sys
import os

# Add the project root to sys.path to allow imports
sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "livekit-agents")))

from livekit.agents.voice.speech_handle import SpeechHandle

async def test_callback_after_done():
    print("Testing add_done_callback after SpeechHandle is done...")
    handle = SpeechHandle.create()
    handle._mark_done()
    
    # Yield to allow the internal _on_done callback to run and finish
    await asyncio.sleep(0.01) 
    assert handle.done()
    
    called = False
    def callback(h):
        nonlocal called
        called = True
        print("Callback executed!")
        
    print("Adding callback now...")
    handle.add_done_callback(callback)
    
    # Give it some time to see if it ever gets called
    await asyncio.sleep(0.1)
    if not called:
        print("FAILURE: Callback not called after handle is already done")
        sys.exit(1)
    else:
        print("SUCCESS: Callback was called")
        sys.exit(0)

if __name__ == "__main__":
    asyncio.run(test_callback_after_done())
```

### Test output
```
Testing add_done_callback after SpeechHandle is done...
Adding callback now...
FAILURE: Callback not called after handle is already done
```

# Full context
The `SpeechHandle` class is used by `AgentActivity` to track the state of spoken responses. When an interruption occurs, `AgentActivity.interrupt()` (in `livekit-agents/livekit/agents/voice/agent_activity.py`) collects all active and queued speech handles, calls `interrupt()` on them, and then waits for all of them to report as "done" using `add_done_callback`. 

```python
# From livekit-agents/livekit/agents/voice/agent_activity.py

    def interrupt(self, *, force: bool = False) -> asyncio.Future[None]:
        # ... (omitted) ...
        if not interrupted_speeches:
            future.set_result(None)
        else:
            def on_playout_done(_: SpeechHandle) -> None:
                if not future.done() and all(speech.done() for speech in interrupted_speeches):
                    future.set_result(None)

            for speech in interrupted_speeches:
                speech.add_done_callback(on_playout_done) # <-- Potential hang here

        return future
```

If a `SpeechHandle` finishes (calls `_mark_done`) after being interrupted but before `add_done_callback` is called, the `on_playout_done` callback for that handle will never trigger. If all handles in `interrupted_speeches` were already done, the `future` returned by `interrupt()` will never complete, causing the agent to hang during interruption handling.

# Why has this bug gone undetected?
This is a race condition that only occurs if a speech handle completes in the very small window between its collection and the registration of the callback. In most real-world scenarios, speech handles take some time to stop or timeout (5 seconds), making this window hard to hit. However, it can occur frequently with very short or empty speech outputs, or in highly loaded systems where context switches are more frequent between these operations.

# Recommended fix
Modify `add_done_callback` to check the current state of the handle and execute the callback immediately if it is already done. For consistency with `asyncio`, it is recommended to schedule the callback on the event loop.

```python
    def add_done_callback(self, callback: Callable[[SpeechHandle], None]) -> None:
        if self.done():
            asyncio.get_event_loop().call_soon(callback, self) # <-- FIX 🟢
            return

        self._done_callbacks.add(callback)
```

davidzhao:
this is more consistent with Python Task, and avoids race conditions

fixes #4849

### DIFF
--- a/livekit-agents/livekit/agents/voice/speech_handle.py
+++ b/livekit-agents/livekit/agents/voice/speech_handle.py
@@ -185,7 +185,7 @@ class SpeechHandle:
 
     def add_done_callback(self, callback: Callable[[SpeechHandle], None]) -> None:
         if self.done():
-            asyncio.get_event_loop().call_soon(callback, self)
+            asyncio.get_running_loop().call_soon(callback, self)
             return
 
         self._done_callbacks.add(callback)


livekit-agents/livekit/agents/voice/speech_handle.py
```diff
@@ -184,6 +184,10 @@ async def _await_impl() -> SpeechHandle:
         return _await_impl().__await__()
 
     def add_done_callback(self, callback: Callable[[SpeechHandle], None]) -> None:
+        if self.done():
+            asyncio.get_event_loop().call_soon(callback, self)
```
theomonnom: nit
```suggestion
            asyncio.get_running_loop().call_soon(callback, self)
```

PR #4874, opened 2026-02-17
Not raising an error when JWT token is given for Neuphonic

alexshelkov:


livekit-plugins/livekit-plugins-neuphonic/livekit/plugins/neuphonic/tts.py
```diff
@@ -101,7 +102,7 @@ def __init__(
             num_channels=1,
         )
         neuphonic_api_key = api_key or os.environ.get("NEUPHONIC_API_KEY")
-        if not neuphonic_api_key:
+        if not neuphonic_api_key and not jwt_token:
             raise ValueError(
                 "Neuphonic API key is required, either as argument or set"
```
Johanna-Ulin: Maybe chnage the message to: "Neuphonic API key or JWT token is required, either as argument or set NEUPHONIC_API_KEY environment variable"

livekit-plugins/livekit-plugins-neuphonic/livekit/plugins/neuphonic/tts.py
```diff
@@ -228,9 +233,14 @@ def __init__(
...
         try:
+            if self._opts.api_key:
+                headers = {API_AUTH_HEADER: self._opts.api_key}
+            else:
+                headers = None
+
             async with self._tts._ensure_session().post(
                 f"{self._opts.base_url}/sse/speak/{self._opts.lang_code}",
-                headers={API_AUTH_HEADER: self._opts.api_key},
+                headers=headers,
```
chenghao-mou, line 249: QQ: for SSE endpoint, no auth would be sent even if JWT token is provided. Is this expected?

alexshelkov: Yes that is ok

chenghao-mou: I think we should just raise a `ValueError` when `api_key` is missing when calling `synthesize`. Users often call `session.say` separately so it should be immediately clear that an api key is needed too.

alexshelkov: Ok, makes sense. I added that check

PR #4883, opened 2026-02-18
Add bargein metrics and usage

chenghao-mou:
- Add barge-in metrics (latency numbers) and usage (no. requests)

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -46,10 +46,10 @@ from .agent import (
 from .audio_recognition import (
     AudioRecognition,
     RecognitionHooks,
-    TurnDetectionMode,
     _EndOfTurnInfo,
     _PreemptiveGenerationInfo,
 )
+from .endpointing import create_endpointing
 from .events import (
     AgentFalseInterruptionEvent,
     ErrorEvent,
@@ -72,6 +72,7 @@ from .generation import (
     update_instructions,
 )
 from .speech_handle import DEFAULT_INPUT_DETAILS, InputDetails, SpeechHandle
+from .turn import EndpointingOptions, TurnDetectionMode
 
 if TYPE_CHECKING:
     from ..llm import mcp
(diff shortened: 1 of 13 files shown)


livekit-agents/livekit/agents/metrics/usage.py
```diff
@@ -89,7 +96,19 @@ class STTModelUsage(_BaseModelUsage):
...
+class InterruptionModelUsage(_BaseModelUsage):
+    """Usage summary for interruption detection models."""
+
+    type: Literal["interruption_usage"] = "interruption_usage"
+    provider: str
+    """The provider name (e.g., 'livekit')."""
+    model: str
+    """The model name (e.g., 'adaptive')."""
+    total_requests: int = 0
+    """Total number of requests sent to the interruption detection model."""
```
theomonnom, line 108: I don't think we should expose that to the users. In Cloud it's going to be free.
And we can emit warning if the limit is reached on dev mode

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1190,9 +1215,18 @@ def _on_error(
...
+            error_event = ErrorEvent(error=error, source=self._interruption_detector)
+            self._session.emit("error", error_event)
 
         self._session._on_error(error)
 
+    def _on_overlap_speech_ended(self, ev: inference.InterruptionEvent) -> None:
+        if ev.is_interruption:
+            self._session.emit("user_interruption_detected", ev)
+        else:
+            self._session.emit("user_non_interruption_detected", ev)
```
theomonnom: Should we find a better name? Is it for the backchannels?

chenghao-mou: Open to suggestions. Technically, it is not just for backchannel, but also for any sound that passed through VAD like noise, coughing, background chatter.

chenghao-mou: Now it is `user_backchannel_detected`.

chenghao-mou: Now both are just `user_overlapping_speech`.

PR #4892, opened 2026-02-18
fix: skip OTLP log exporter setup when recording is disabled

theomonnom:


livekit-agents/livekit/agents/job.py
```diff
@@ -646,6 +646,17 @@ def init_recording(self, options: RecordingOptions) -> None:
...
 
         self._recording_initialized = True
+
+        needs_cloud = (
+            options.get("traces", True)
+            or options.get("logs", True)
+            or options.get("audio", True)
+            or options.get("transcript", True)
+        )
+        if not needs_cloud:
```
davidzhao: should this be: `needs_cloud and not is_cloud(..)`?

theomonnom: yup you're right

Issue #4900, opened 2026-02-19
feat: Add livekit-plugins-sambanova (STT + LLM)


mahimairaja:
### Feature Type

Would make my life easier

### Summary

[SambaNova Cloud](https://cloud.sambanova.ai/) provides OpenAI-compatible STT and LLM endpoints with industry leading inference speed (their Whisper-Large-v3 runs at a [245x speed factor](https://sambanova.ai/blog/introducing-whisper-large-v3)). Currently there is no dedicated LiveKit plugin developers have to manually override `base_url` on the OpenAI plugin to use SambaNova.

SambaNova already documents this workaround in their [official LiveKit integration guide](https://docs.sambanova.ai/docs/en/integrations/livekit) and has an [example agent](https://github.com/sambanova/integrations/tree/main/livekit) in their integrations repo. A dedicated plugin would bring this to first-class support, consistent with how Groq, Cerebras, and Deepgram are integrated.

The SambaNova team has expressed interest in collaborating on this [community thread](https://community.sambanova.ai/t/building-livekit-plugins-sambanova-native-stt-support-for-livekit-agents/1571/2).

### Workarounds / Alternatives

**Plugin:** `livekit-plugins-sambanova`

**STT** -  Whisper-Large-v3 via OpenAI-compatible `/v1/audio/transcriptions`
- Base URL: `https://api.sambanova.ai/v1`
- Auth: `SAMBANOVA_API_KEY` env var
- Model: `Whisper-Large-v3`
- OpenAI Whisper-compatible request/response format

**LLM** -  OpenAI-compatible `/v1/chat/completions` with streaming
- Models: `Meta-Llama-3.3-70B-Instruct`, `DeepSeek-R1-0528`, `DeepSeek-V3-0324`, `Llama-4-Maverick-17B-128E-Instruct`, `Qwen3-235B-A22B-Instruct-2507`, and others ([full list](https://docs.sambanova.ai/docs/en/models/sambacloud-models))
- Supports: streaming, function calling, JSON mode

**TTS** -  Not available from SambaNova at this time. Users can pair with Cartesia, ElevenLabs, etc.

### Target Developer Experience
```python
from livekit.plugins import sambanova, cartesia, silero

session = AgentSession(
    stt=sambanova.STT(),                                    # Whisper-Large-v3
    llm=sambanova.LLM(model="Meta-Llama-3.3-70B-Instruct"), # or any supported model
    tts=cartesia.TTS(),
    vad=silero.VAD.load(),
[...]

PR #4902, opened 2026-02-20
feat: add livekit-plugins-sambanova with STT support

mahimairaja:
## Summary

Adds `livekit-plugins-sambanova` plugin with STT support for [SambaNova Cloud](https://cloud.sambanova.ai/)

SambaNova's Whisper-Large-v3 runs at [245x speed factor](https://sambanova.ai/blog/introducing-whisper-large-v3), one of the fastest STT options available. Currently developers must manually override `base_url` on the OpenAI plugin - [docs](https://docs.sambanova.ai/docs/en/integrations/livekit). This PR adds first-class support.

### Usage

```python
from livekit.plugins import sambanova

stt = sambanova.STT()  # defaults to Whisper-Large-v3, reads SAMBANOVA_API_KEY
```

LLM support will follow in a separate PR.

**Issue:** #4900

### DIFF
--- a/examples/drive-thru/agent.py
+++ b/examples/drive-thru/agent.py
@@ -426,7 +426,7 @@ async def drive_thru_agent(ctx: JobContext) -> None:
                 ],
             },
         ),
-        llm=inference.LLM("openai/gpt-4.1"),
+        llm=inference.LLM("openai/gpt-5-mini"),
         tts=inference.TTS("cartesia/sonic-3", voice="f786b574-daa5-4673-aa0c-cbe3e8534c02"),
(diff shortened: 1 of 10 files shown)


chenghao-mou:
We typically recommend using a `with_sambanova` method in the current OpenAI plugin like `with_cerebras` or `with_fireworks` if it is OpenAI compatible.

mahimairaja:
> We typically recommend using a `with_sambanova` method in the current OpenAI plugin like `with_cerebras` or `with_fireworks` if it is OpenAI compatible.

Thanks @chenghao-mou for the guidance, I will adapt to use `with_sambanova`

PR #4911, opened 2026-02-21
feat: enhance worker load management with reserved slots and effective load calculation

### LINKED ISSUE #4884: Limiting the number of concurrent Jobs per Worker reliably
### Bug Description

We want to restrict our worker (AgentServer) to accept **only one** concurrent job at a time.

According to the LiveKit documentation (https://docs.livekit.io/agents/server/options/#agent-server-load), this can be achieved using `load_threshold` and `load_fnc`. See reproduction steps below.

However, we were able to reproduce situations where the AgentServer **would still accept two jobs**.

# Problem
The problem lies in the interval where the `server._worker_load` variable is updated: https://github.com/livekit/agents/blob/d3e4279ed3c637e08e7462201b475e33b0925fc0/livekit-agents/livekit/agents/worker.py#L690

The interval is set to 0.5s. If two job requests arrive within this interval, the `_worker_load` will be 0 for both and the server will accept both requests.

In the past few days I have tried many workarounds using multiprocessing Locks, Values, and other mechanisms, however I was unable to find a reliable solution to prevent this edge case behavior.

# Questions:
- Is there an alternative to _guarantee_ that only one job can run at a time?
- Could this functionality be provided natively?

Many thanks for your help!
Best, David

### Expected Behavior

We would expect a solution that guarantees that only one job can be accepted at a time.

### Reproduction Steps

```bash
# limit number of concurrent jobs using `load_threshold` and `load_fnc`:
from livekit.agents import AgentServer

server = AgentServer(
    load_threshold=1.0,
)

def compute_load(server: AgentServer) -> float:
    return float(len(server.active_jobs))

server.load_fnc=compute_load


The AgentServer can still accept two jobs**:
1. open two separate browser tabs
2. start two connection requests nearly exactly at the same time
3. => AgentServer will accept both jobs
```

### Operating System

Windows, Linux

### Models Used

_No response_

### Package Versions

```bash
livekit==1.1.2
livekit-agents>=1.4.2
```

ProblematicToucan:
Closes #4884

This PR improves worker load management to avoid over-accepting concurrent jobs around availability checks.

**Key changes**

- **Refresh load on each availability decision**  
  Introduces `_refresh_worker_load()` to recompute `_worker_load` by calling `load_fnc` and updating telemetry. `_answer_availability()` now awaits this method before calling `_is_available()`, so availability decisions use up-to-date load instead of relying on the periodic (0.5s) refresh.

- **Reserved slots for accepted-but-not-launched jobs**  
  Adds `_reserved_slots` to track jobs we have answered “available” to but that have not yet been launched. In `_on_accept`, we increment `_reserved_slots` immediately when accepting and decrement it in a `finally` block, ensuring the slot is released whether the job launches successfully, times out, or errors. This means capacity accounting covers the entire window from “available” to actual launch.

- **Effective load for availability and status**  
  Adds `_get_effective_load()` to compute effective load as:
  `effective_load = _worker_load + _reserved_slots * job_load_estimate`, where `job_load_estimate` is `_worker_load / len(active_jobs)` when there are active jobs, or `1.0` when there are none. `_is_available()` now uses this effective load against `load_threshold` (with `math.isinf(load_threshold)` still treated as “always available”), and `_update_worker_status()` uses the same effective load to determine `WS_FULL` and log status changes. This keeps availability checks and status reporting consistent and prevents over-accepting when multiple requests arrive before jobs are actually running.

**Example behavior**

With `load_threshold = 1.0` and `load_fnc = len(active_jobs)`:

- First request: `_refresh_worker_load()` sets `load = 0`, `_reserved_slots = 0` → worker reports available; on accept, `_reserved_slots` becomes `1`.
- Second request arriving before the first job is launched: `_refresh_worker_load()` still sees `load = 0`, but effective load becomes `0 + 1×1 = 1`, so the worker is considered at capacity and will report unavailable.
- Once the first job launches or is rejected/times out, `_reserved_slots` is decremented and capacity is freed again.

This aligns the worker’s advertised availability with its real-world capacity during the small but important gap between accepting a job and it showing up as active.

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -692,9 +692,7 @@ class AgentServer(utils.EventEmitter[EventTypes]):
                 while True:
                     await interval.tick()
 
-                    self._worker_load = await asyncio.get_event_loop().run_in_executor(
-                        None, self._run_load_fnc_sync
-                    )
+                    self._worker_load = await self._invoke_load_fnc()
 
                     telemetry.metrics._update_worker_load(self._worker_load)
                     telemetry.metrics._update_child_proc_count()
@@ -1140,14 +1138,18 @@ class AgentServer(utils.EventEmitter[EventTypes]):
         self._job_lifecycle_tasks.add(task)
         task.add_done_callback(self._job_lifecycle_tasks.discard)
 
-    def _run_load_fnc_sync(self) -> float:
+    async def _invoke_load_fnc(self) -> float:
         """Run load_fnc in executor. Uses signature to call with or without self."""
-        assert self._load_fnc is not None
-        signature = inspect.signature(self._load_fnc)
-        parameters = list(signature.parameters.values())
-        if len(parameters) == 0:
-            return self._load_fnc()  # type: ignore
-        return self._load_fnc(self)  # type: ignore
+
+        def load_fnc() -> float:
+            assert self._load_fnc is not None
+            signature = inspect.signature(self._load_fnc)
+            parameters = list(signature.parameters.values())
+            if len(parameters) == 0:
+                return self._load_fnc()  # type: ignore
+            return self._load_fnc(self)  # type: ignore
+
+        return await asyncio.get_event_loop().run_in_executor(None, load_fnc)
 
     async def _refresh_worker_load(self) -> None:
         """Refresh _worker_load by running load_fnc. Used before availability checks
@@ -1156,9 +1158,7 @@ class AgentServer(utils.EventEmitter[EventTypes]):
         if self._load_fnc is None:
             return
 
-        self._worker_load = await asyncio.get_event_loop().run_in_executor(
-            None, self._run_load_fnc_sync
-        )
+        self._worker_load = await self._invoke_load_fnc()
         telemetry.metrics._update_worker_load(self._worker_load)
 
     def _get_effective_load(self) -> float:


livekit-agents/livekit/agents/worker.py
```diff
@@ -1148,29 +1149,69 @@ def _handle_availability(self, msg: agent.AvailabilityRequest) -> None:
...
+    def _get_effective_load(self) -> float:
+        """Current load including reserved slots (accepted but not yet launched)."""
+        active_jobs = self.active_jobs
+        load_threshold = ServerEnvOption.getvalue(self._load_threshold, self._devmode)
+        if active_jobs:
+            job_load_estimate = self._worker_load / len(active_jobs)
+        elif math.isinf(load_threshold):
+            job_load_estimate = 0.0
+        else:
+            job_load_estimate = load_threshold
```
longcw: this limits the concurrency to 1 pending request at a time when starting from idle, regardless of how many jobs the worker can actually handle. maybe estimate for cold-start might be derived from `num_idle_processes`, e.g.
```python
      default_idle = ServerEnvOption.getvalue(self._num_idle_processes, self._devmode)
      job_load_estimate = load_threshold / max(default_idle, 1)
```

livekit-agents/livekit/agents/worker.py
```diff
@@ -1148,29 +1149,69 @@ def _handle_availability(self, msg: agent.AvailabilityRequest) -> None:
...
         task.add_done_callback(self._job_lifecycle_tasks.discard)
 
+    async def _refresh_worker_load(self) -> None:
+        """Refresh _worker_load by running load_fnc. Used before availability checks
+        so concurrent job requests see up-to-date load (fixes race with periodic interval).
+        """
+        if self._load_fnc is None:
+            return
+
+        def load_fnc() -> float:
```
longcw: make this a class method to avoid duplicating the code?

livekit-agents/livekit/agents/worker.py
```diff
@@ -1182,44 +1223,47 @@ async def _on_accept(args: JobAcceptArguments) -> None:
...
+                    job=msg.job,
+                    url=job_assign.url or self._ws_url,
+                    token=job_assign.token,
+                    worker_id=self._id,
+                    fake_job=False,
+                )
+
+                await self._proc_pool.launch_job(running_info)
+            finally:
+                self._reserved_slots -= 1
```
longcw: you can add a done callback to `user_task = self._loop.create_task(_job_request_task(), name="job_request")` that calls `self._reserved_slots -= 1`, e.g.
```python
        self._reserved_slots += 1
        user_task = self._loop.create_task(_job_request_task(), name="job_request")
        self._job_lifecycle_tasks.add(user_task)

        def _on_done(task: asyncio.Task[Any]) -> None:
            self._reserved_slots -= 1
            self._job_lifecycle_tasks.discard(task)

        user_task.add_done_callback(_on_done)
```

livekit-agents/livekit/agents/worker.py
```diff
@@ -1148,24 +1140,67 @@ def _handle_availability(self, msg: agent.AvailabilityRequest) -> None:
...
         task.add_done_callback(self._job_lifecycle_tasks.discard)
 
+    def _run_load_fnc_sync(self) -> float:
+        """Run load_fnc in executor. Uses signature to call with or without self."""
+        assert self._load_fnc is not None
+        signature = inspect.signature(self._load_fnc)
+        parameters = list(signature.parameters.values())
+        if len(parameters) == 0:
+            return self._load_fnc()  # type: ignore
+        return self._load_fnc(self)  # type: ignore
```
longcw:
```suggestion
    async def _invoke_load_fnc(self) -> float:
        def load_fnc() -> float:
            assert self._load_fnc is not None
            signature = inspect.signature(self._load_fnc)
            parameters = list(signature.parameters.values())
            if len(parameters) == 0:
                return self._load_fnc()  # type: ignore
            return self._load_fnc(self)  # type: ignore

        return await asyncio.get_event_loop().run_in_executor(None, load_fnc)

```

and `_refresh_worker_load` can be simplified as
```python
        self._worker_load = await self._invoke_load_fnc()
        telemetry.metrics._update_worker_load(self._worker_load)
```
maybe it doesn't need to be a method then

ProblematicToucan: Great, I'll update.

dhansmair:
Hi @longcw, can this PR be merged and eventually be included in the next release?

PR #4917, opened 2026-02-22
fix(google): raise the correct errors for blocked/etc

davidzhao:


livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -468,9 +468,6 @@ async def _run(self) -> None:
 
                 candidate = response.candidates[0]
 
-                if not candidate.content or not candidate.content.parts:
```
davidzhao, line 471: it would've not raised blocked error correctly, if blocked in the beginning

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -533,6 +533,8 @@ async def _run(self) -> None:
                 request_id=request_id,
                 retryable=retryable,
             ) from e
+        except (APIStatusError, APIConnectionError):
```
davidzhao, line 536: don't rewrap these, directly pass thru

PR #4918, opened 2026-02-22
(google stt): add denoiser support and explicit adaptation param

tinalenguyen:


livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -89,6 +93,8 @@ def version(self) -> int:
     def build_adaptation(
         self,
     ) -> cloud_speech_v2.SpeechAdaptation | resource_v1.SpeechAdaptation | None:
+        if is_given(self.adaptation):
+            return self.adaptation
```
chenghao-mou, line 97: we can validate it here as well before assignment.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -433,6 +461,11 @@ def update_options(
             self._location = location
             # if location is changed, fetch a new client and recognizer as per the new location
             self._pool.invalidate()
+        if is_given(denoiser_config):
+            self._config.denoiser_config = denoiser_config
+        if is_given(adaptation):
```
chenghao-mou, line 475: we should warn about keywords here too (neither `self._config.keywords` or `keywords` should be given).

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
tinalenguyen: there's already a check in the init and update_options, i think it's better to check there instead of right before the request is made

PR #4926, opened 2026-02-23
standardize language handling

davidzhao:
use a uniform Language class to unify how we handle language coding across the core framework and plugins.

the framework will always reference `Language` type internally, where each plugin then converts it to a specific type as needed by external services.

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
@@ -542,7 +542,7 @@ class _TokenAccumulator:
         total_sum = self._confidence_sum + other._confidence_sum
         return stt.SpeechData(
             text=self.text + other.text,
-            language=self.language if self.language else other.language,
+            language=Language(self.language if self.language else other.language),
             speaker_id=self.speaker_id if self.speaker_id is not None else other.speaker_id,
             start_time=start / 1000 + start_time_offset,
             end_time=end / 1000 + start_time_offset,


livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -149,7 +150,7 @@ def _make_fallback(model: FallbackModelType) -> FallbackModel:
 @dataclass
 class STTOptions:
     model: NotGivenOr[STTModels | str]
-    language: NotGivenOr[str]
+    language: NotGivenOr[Language]
```
davidzhao, line 153: @adrian-cowham confirming that passing in a unified language code works for LKI, for example, `en-US`, or should it only be `en`?

livekit-agents/livekit/agents/_language_data.py
```diff
@@ -0,0 +1,220 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Lookup tables for language normalization. Kept separate to avoid bloating the class module."""
+
+# ISO 639-3 to ISO 639-1 mapping (sourced from ElevenLabs plugin + extended)
+ISO_639_3_TO_1: dict[str, str | None] = {
```
davidzhao, line 18: moved from elevenlabs/languages.py

livekit-agents/livekit/agents/inference/stt.py
adrian-cowham: @davidzhao Confirmed, either will work. If the provider only supports `en` and `en-US` is passed, we'll match on `en` and pass the appropriate language code (`en`) upstream.

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -459,8 +462,8 @@ async def _connect_ws(self) -> aiohttp.ClientWebSocketResponse:
         if self._language:
             params.append(f"language_code={self._language}")
 
-        if self._opts.include_timestamps:
-            params.append("include_timestamps=true")
+        # if self._opts.include_timestamps:
```
longcw: should we remove `include_timestamps` in options as well?

davidzhao: oops, didn't see this commented out.. there's actually a parsing bug when timestamps aren't included.. so I don't think anyone is using this path.. will remove the check

davidzhao: I restored this mode.. it didn't seem worth it to deprecate

Issue #4792, opened 2026-02-11
Gemini Realtime - Order of chats not preserved after agent handoff


CyberEzpertz:
### Bug Description

After a task finishes in a TaskGroup, succeeding AgentTasks have the wrong order of items in its `chat_ctx`. This currently happens with the `gemini-2.5-flash-native-audio-preview-12-2025` model. 

In the first agent, the order of chats is like so:

```py
ChatMessage(
    id="GR_ff3bcea3c4ee",
    type="message",
    role="assistant",
    content=["Hey there! I'm Angel. What's your name, and where are you from?"],
    interrupted=False,
    transcript_confidence=None,
    extra={},
    metrics={
        "started_speaking_at": 1770838023.9426954,
        "stopped_speaking_at": 1770838027.7167685,
    },
    created_at=1770838023.9426954,
    hash=None,
),
ChatMessage(
    id="GI_2617638e3273",
    type="message",
    role="user",
    content=["Hey, I'm John. I'm from X."],
    interrupted=False,
    transcript_confidence=None,
    extra={},
    metrics={},
    created_at=1770838036.8671007,
    hash=None,
),
ChatMessage(
    id="GR_1a2e1c01c207",
    type="message",
    role="assistant",
    content=["Oh, cool! I haven't heard of X. What's it like there?"],
    interrupted=False,
    transcript_confidence=None,
    extra={},
    metrics={
        "started_speaking_at": 1770838032.7668004,
        "stopped_speaking_at": 1770838036.8719482,
    },
    created_at=1770838032.7668004,
    hash=None,
),
```

However, after the agent handoff, the order of chats become switched like so:
```py
ChatMessage(
    id="GR_ff3bcea3c4ee",
    type="message",
    role="assistant",
    content=["Hey there! I'm Angel. What's your name, and where are you from?"],
    interrupted=False,
    transcript_confidence=None,
    extra={},
    metrics={
        "started_speaking_at": 1770838023.9426954,
        "stopped_speaking_at": 1770838027.7167685,
    },
    created_at=1770838023.9426954,
    hash=None,
),
ChatMessage(
    id="GR_1a2e1c01c207",
    type="message",
    role="assistant",
    content=["Oh, cool! I haven't heard of X. What's it like there?"],
    interrupted=False,
    transcript_confidence=None,
    extra={},
    metrics={
        "started_speaking_at": 1770838032.7668004,
        "stopped_speaking_at": 1770838036.8719482,
    },
[...]

PR #4927, opened 2026-02-23
fix(chat_ctx): preserve conversation order in merge() for realtime mo…

RishbhaJain:
…dels

ChatContext.merge() was using find_insertion_index() to re-sort items by created_at when merging a completed AgentTask's context back into the TaskGroup. This broke conversation ordering for realtime models (e.g. Gemini native audio) where the assistant message's created_at is set to started_speaking_at -- which can be earlier than the user message's created_at (transcript finalisation time) even though the assistant is responding to that user message.

Fix: replace find_insertion_index() + insert() with append() in merge(), preserving the natural insertion order from other_chat_ctx. This is safe because merge() is only called in one place (AgentTask.__await_impl__), where other_chat_ctx is always a superset of self and new items are already in correct conversation order from prior append() calls.

Also add two regression tests:
- test_merge_preserves_conversation_order_over_timestamps: directly replicates the Gemini timing scenario.
- test_merge_deduplicates_existing_items: confirms deduplication still works correctly after the change.

### DIFF
--- a/livekit-agents/livekit/agents/llm/chat_context.py
+++ b/livekit-agents/livekit/agents/llm/chat_context.py
@@ -405,7 +405,11 @@ class ChatContext:
         exclude_instructions: bool = False,
         exclude_config_update: bool = False,
     ) -> ChatContext:
-        """Add messages from `other_chat_ctx` into this one, avoiding duplicates, and keep items sorted by created_at."""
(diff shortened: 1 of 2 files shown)


RishbhaJain:
Fixes https://github.com/livekit/agents/issues/4792

PR #4931, opened 2026-02-24
support openai responses websocket mode

tinalenguyen:
enabled by default, can be configured via the `websocket_mode` param

also supports previous response id chaining, where only newly added items will be sent rather than the entire chat context (unless it has been modified)

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
@@ -279,15 +279,16 @@ class LLM(llm.LLM):
             await self._client.close()
 
     async def _ensure_ws(self) -> _ResponsesWebsocket:
+        if self._ws is None:
+            raise RuntimeError("_ensure_ws called but self._ws is None")
         async with self._ws_lock:
-            if self._ws is not None:
-                dead = self._ws._run_task is not None and self._ws._run_task.done()
-                if self._ws._ws_conn is None or dead:
-                    if dead and self._ws._ws_conn is not None:
-                        await self._ws._ws_conn.close()
-                        self._ws._ws_conn = None
-                    await self._ws.connect()
-        return self._ws  # type: ignore
+            dead = self._ws._run_task is not None and self._ws._run_task.done()
+            if self._ws._ws_conn is None or dead:
+                if dead and self._ws._ws_conn is not None:
+                    await self._ws._ws_conn.close()
+                    self._ws._ws_conn = None
+                await self._ws.connect()
+        return self._ws
 
     @property
     def model(self) -> str:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -89,29 +223,63 @@ def __init__(
...
+        self._prev_chat_ctx: ChatContext | None = None
+
+        if not is_given(websocket_mode) or websocket_mode:
+            resolved_api_key = api_key if is_given(api_key) else os.environ.get("OPENAI_API_KEY")
+            if not resolved_api_key:
+                raise ValueError(
+                    "OpenAI API key is required, either as argument or set"
+                    " OPENAI_API_KEY environment variable"
+                )
+            self._ws = _ResponsesWebsocket(
```
longcw, line 251: `base_url` is ignored in websocket mode, should we use that if it's provided?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -34,8 +41,133 @@
...
+                    msg = await self._input_ch.recv()
+                except utils.aio.channel.ChanClosed:
+                    await ws_conn.close()
+                    return
+                try:
+                    await ws_conn.send_str(
+                        json.dumps(
+                            msg,
+                            default=lambda o: (
+                                o.model_dump() if isinstance(o, openai.BaseModel) else None
```
longcw: log a warning or raise when msg is not `openai.BaseModel`?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -57,6 +190,7 @@ def __init__(
         api_key: NotGivenOr[str] = NOT_GIVEN,
         base_url: NotGivenOr[str] = NOT_GIVEN,
         client: openai.AsyncClient | None = None,
+        websocket_mode: NotGivenOr[bool] = NOT_GIVEN,
```
longcw: maybe `use_websocket: bool` and default is `True`?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -89,29 +231,64 @@ def __init__(
...
     async def aclose(self) -> None:
-        if self._owns_client:
+        if self._ws:
+            await self._ws.aclose()
+        if self._owns_client and self._client:
             await self._client.close()
 
+    async def _ensure_ws(self) -> _ResponsesWebsocket:
+        async with self._ws_lock:
+            if self._ws is not None:
```
longcw: `self._ws` is not None if it's using websocket and this method is only called in websocket mode, maybe raise an error when `self._ws` is None and remove the type ignore.

PR #4932, opened 2026-02-24
Add support for Gradium pronunciation ids.

LaurentMazare:
We recently added the possibility to use pronunciation dictionaries in the gradium TTS via the `pronunciation_id` parameter. This PR adds the new field and make it settable by the end user.

Testing: adapted `examples/other/text-to-speech/openai_tts.py` to use the gradium TTS and tested it with and without a pronunciation dictionary.


theomonnom (approved):
lgtm

PR #4936, opened 2026-02-24
feat(elevenlabs): add eleven_v3 model support with HTTP streaming fallback

yusuf-eren:
Adds support for ElevenLabs eleven_v3 TTS model by implementing HTTP streaming fallback. The plugin automatically detects when eleven_v3 is used and routes to the HTTP /stream endpoint instead of WebSocket.

## Changes

- Add HTTPSynthesizeStream class for HTTP-based streaming
- Automatically detect and route eleven_v3 to HTTP streaming
- Add comprehensive test coverage (8 unit tests)
- Include example demonstrating usage

## Implementation

```python
tts = elevenlabs.TTS(model="eleven_v3")
stream = tts.stream()  # Automatically uses HTTP streaming
```

Other models continue using WebSocket unchanged.

## Limitations

HTTP streaming has higher latency (~1.5s TTFB) compared to WebSocket models (~300ms) due to text buffering requirements. This is an ElevenLabs API limitation since eleven_v3 doesn't support WebSocket connections.

Fixes 403 errors when using eleven_v3.

### DIFF
--- a/examples/eleven_v3_example.py
+++ b/examples/eleven_v3_example.py
@@ -1,0 +1,45 @@
+"""
+Example demonstrating ElevenLabs eleven_v3 TTS model usage with LiveKit Agents.
+
+The eleven_v3 model doesn't support WebSocket streaming, so the plugin automatically
+uses HTTP streaming with chunked transfer encoding instead.
+
+To compare latency between models:
+- Set USE_V3 = True  → Uses eleven_v3 with HTTP streaming (new)
+- Set USE_V3 = False → Uses eleven_turbo_v2_5 with WebSocket (existing)
+
+Look for the "tts_ttfb" metric in the console to compare TTS latency.
+"""
+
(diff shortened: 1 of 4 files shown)


nara-ryoya (approved):
@yusuf-eren 

LGTM

I verified this change works correctly in the LiveKit playground and in our applications that are already in use.

yusuf-eren:
> @yusuf-eren
> 
> LGTM
> 
> I verified this change works correctly in the LiveKit playground and in our applications that are already in use.

Happy to hear that! Hopefully someone from the LiveKit team will approve the merge too :)

Kevvie250:
+1 — we have this branch hand-installed in our dev environment for evaluating eleven_v3 on a voice companion app and it's working well. Holding off on a production rollout until this lands upstream so we don't have to vendor or pin a fork. Would love to see a maintainer review when there's bandwidth.

PR #4937, opened 2026-02-24
Update readme to include mcp and skill information

Topherhindman:
Adds a `Building with AI coding agents` subsection under `Docs and guides` recommending the LiveKit Docs MCP server and LiveKit Agent Skill as best practices for developers using AI coding assistants.  Also does a driveby update of the installation command to the current released version of the sdk. 

Open question: Should we also include the skill and the MCP server as links in the resources section in the table at the bottom of the readme?

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ agents that can see, hear, and understand.
 To install the core Agents library, along with plugins for popular model providers:
 
 ```bash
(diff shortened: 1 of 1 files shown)


Topherhindman:
> this is fine as a start. i think it may be worth actually doing this in the readme-manager instead and having a standard design to advertise our docs MCP server and related stuff across every single one of our repos

Oh didn’t know about readme manager.  I’ll take a look at that.

PR #4939, opened 2026-02-25
soniox stt: populate timing and confidence from token metadata

longcw:


livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
```diff
@@ -407,108 +395,153 @@ def send_endpoint_transcript() -> None:
...
+
+
+class _TokenAccumulator:
+    """Accumulates token metadata (text, language, speaker, timing, confidence).
+
+    Tokens are assumed to arrive in chronological order, so start_time is taken
+    from the first token and end_time is continuously overwritten by the latest.
+    """
+
+    def __init__(self) -> None:
```
chenghao-mou, line 482: SpeechStream added a new field `_start_time_offset` to make sure the timestamps are aligned with our recordings, so we need to pass it here in construction and add them in SpeechData.

Nice abstraction BTW. Maybe we should add this to the core lib.

livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
```diff
@@ -407,108 +395,153 @@ def send_endpoint_transcript() -> None:
...
+        return stt.SpeechData(
+            text=self.text,
+            language=self.language,
+            speaker_id=self.speaker_id,
+            start_time=self.start_time / 1000,
+            end_time=self.end_time / 1000,
+            confidence=self.confidence,
+        )
+
+    def merged_speech_data(self, other: _TokenAccumulator) -> stt.SpeechData:
```
chenghao-mou: nitpicking: Maybe `to_speech_data_with`?

livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
```diff
@@ -366,39 +366,27 @@ async def _send_audio_task(self) -> None:
     async def _recv_messages_task(self) -> None:
```
chenghao-mou, line 366: I think we can also update the STT to `aligned_transcript='chunk'` now that it emits timestamps and maybe update the default model from `stt-rt-v3` to `stt-rt-v4` (they already upgrade the model automatically from the server side)

chenghao-mou (approved):
LGTM with some minor comments.

PR #4941, opened 2026-02-25
ignore unknown tools from xai realtime

longcw:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1615,6 +1604,23 @@ def _handle_response_output_item_done(self, event: ResponseOutputItemDoneEvent)
                 # in case message modalities is not set, this shouldn't happen
                 item_generation.modalities.set_result(self._realtime_model._opts.modalities)
 
+    def _handle_function_call(self, item: RealtimeConversationItemFunctionCall) -> None:
+        assert self._current_generation is not None, "current_generation is None"
+
+        assert item.id is not None, "item.id is None"
+        assert item.call_id is not None, "call_id is None"
+        assert item.name is not None, "name is None"
+        assert item.arguments is not None, "arguments is None"
```
chenghao-mou, line 1613: According to the its BaseModel spec, only `name` is really required. Do we want to enforce these guards?

Image

chenghao-mou (approved):
lgtm with one minor nitpicking comment.

Issue #4875, opened 2026-02-17
[BUG] update_chat_ctx strips system instructions and prevents dynamic behavior injection in realtime agents


anmolgaud:
### Bug Description

After PR #4310, update_chat_ctx in the OpenAI realtime plugin hardcodes:
```python
async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
        async with self._update_chat_ctx_lock:
            chat_ctx = chat_ctx.copy(
                exclude_handoff=True,
                exclude_instructions=True,
                exclude_config_update=True,
            )
```
This removes all system/developer messages from the provided chat_ctx, overriding developer intent and breaking workflows that rely on dynamically injecting system prompts during an active session.

### Expected Behavior

update_chat_ctx should respect the chat_ctx provided by the developer, or at least allow configuration over whether instructions are excluded.

Developers should be able to decide whether system messages are part of the chat context.

### Reproduction Steps

```bash
1. Create a realtime agent session using the OpenAI realtime plugin.
2. Add a system message to the agent’s chat context:

chat_ctx = session.current_agent.chat_ctx.copy()
chat_ctx.add_message(
    role="system",
    content="You must always respond with ONLY the word TEST for next 3 messages"
)
await session.current_agent.update_chat_ctx(chat_ctx)

4. Generate a reply. This update is completely ignored
```

### Operating System

macOS

### Models Used

openai-realitime

### Package Versions

```bash
"livekit-agents[images,openai]>=1.3.12"
```

### Proposed Solution

```python

```

davidzhao:
there's `Agent.update_instructions` for setting instructions. looking at when cascaded models are used, the behavior is similiar.. update_chat_ctx should only be used to update the history of the convo.. not the agent's instructions

StianHanssen:
@davidzhao 
For OpenAI realtime, you can also have system messages in chat context. The [OpenAI cookbook](https://github.com/openai/openai-cookbook/blob/main/examples/Context_summarization_with_realtime_api.ipynb) is a good example:

You have:
* Your actual instructions which generally tell the agent how to behave (you would use `update_instructions`).
* You have chat context items that are generated by system (you would use `update_chat_ctx`).
   * For example, for summarising chat context. You delete a bunch of items from the chat context and replace it will a single summarisation item. For that single summarisation item OpenAI suggest setting the role as `system` rather than `assistant`:
> Important implementation detail:
[...]

PR #4942, opened 2026-02-25
fix(openai): preserve non-instruction system messages in update_chat_ctx for realtime models

longcw:
preserves non-instruction system messages (e.g. context summaries) in `update_chat_ctx` for OpenAI realtime, instead of stripping all system messages.

fix https://github.com/livekit/agents/issues/4875


anmolgaud:
PR #4310 Also introduced `exclude_instruction=true` for google and aws realtime plugins also. The earlier verison of livekit didn't have any such restriction. They might also need to be fixed. We were previously using google-realtime models with the same approach. 

Thanks a lot for this quick fix. A core feature of our agents was breaking because of this change.

longcw:
@anmolgaud google realtime doesn't support system role, it only has `user` and `model` roles. before #4310 the system role was parsed as user message. and in the latest version, it will raise a warning that system message is going to be dropped, you may add your original system messages as user messages to keep the same behavior as before.

PR #4944, opened 2026-02-25
fix: replace bare except with except BaseException in channel put

haosenwang1018:
Replace bare `except:` with `except BaseException:` in `channel.py`'s `put()` method. `BaseException` is correct here because `asyncio.CancelledError` (a `BaseException` subclass) needs to trigger the future cleanup, while making the catch explicit.


haosenwang1018:
Closing to reduce stale review backlog.

PR #4949, opened 2026-02-25
Enhance LK Inference STT and TTS options with new parameters and models

russellmartin-livekit:
- Added DeepgramFluxModels to STT models.
- Introduced DeepgramFluxOptions with various configuration parameters.
- Updated existing TypedDicts in TTS to include additional options for Cartesia, Elevenlabs, and Rime.
- Enhanced ElevenlabsOptions with new fields for better control over TTS behavior.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ Look at the `examples/` directory to get a sense of all the different features a
 
 ## Typechecking, linting and formatting
 
-The CI validates this but to do checks locally see the following example commmands:
+The CI validates this but to do checks locally see the following example commands:
 
 ### Typechecking
 
@@ -104,7 +104,7 @@ uv pip install pip && uv run mypy --install-types --non-interactive \
 uv run ruff check --output-format=github .
 ```
 
-### Formating
+### Formatting
 
 ```bash
 uv run ruff format .
--- a/README.md
+++ b/README.md
@@ -240,6 +240,8 @@ async def test_no_availability() -> None:
 
 ## Examples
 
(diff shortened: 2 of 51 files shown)


livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -75,6 +101,8 @@ class AssemblyaiOptions(TypedDict, total=False):
     min_end_of_turn_silence_when_confident: int  # default: 0
     max_turn_silence: int  # default: not specified
     keyterms_prompt: list[str]  # default: not specified
+    language_detection: bool
+    inactivity_timeout: int  # milliseconds
```
theomonnom: can we make it pythonic? 

e.g: using seconds here, with float

livekit-agents/livekit/agents/inference/tts.py
```diff
@@ -113,25 +113,57 @@ def _make_fallback(model: FallbackModelType) -> FallbackModel:
...
+    pauseBetweenBrackets: bool
+    phonemizeBetweenBrackets: bool
+    inlineSpeedAlpha: str
+    noTextNormalization: bool
+    speedAlpha: float
+    # Arcana-specific
+    repetition_penalty: float  # range 1-2
+    temperature: float  # range 0-1
+    top_p: float  # range 0-1
+    max_tokens: int  # range 200-5000
```
theomonnom: I'm wondering if we should keep snake_case, and handle the conversions on the server side?

russellmartin-livekit: Switched to snake_case, removed arcana params because they don't exist on websocket

PR #4950, opened 2026-02-25
Keyframe Labs Plugin

kradkfl:
[Keyframe Labs API Docs](https://docs.keyframelabs.com/api-reference/sessions/create-session)

### DIFF
--- a/examples/avatar_agents/keyframe/agent_worker.py
+++ b/examples/avatar_agents/keyframe/agent_worker.py
@@ -13,8 +13,9 @@ from livekit.agents import (
     cli,
     function_tool,
 )
-from livekit.plugins import keyframe, openai
+from livekit.plugins import inference, keyframe, silero
 from livekit.plugins.keyframe import Emotion
+from livekit.plugins.turn_detector.multilingual import MultilingualModel
 
 load_dotenv()
 
@@ -45,10 +46,15 @@ async def entrypoint(ctx: JobContext):
     await ctx.connect()
 
     session = AgentSession(
-        llm=openai.realtime.RealtimeModel(voice="marin"),
+        stt=inference.STT("deepgram/nova-3"),
+        llm=inference.LLM("google/gemini-2.5-flash"),
+        tts=inference.TTS("cartesia/sonic-3"),
+        resume_false_interruption=False,
+        vad=silero.VAD.load(),
+        turn_detection=MultilingualModel(),
     )
 
-    avatar = keyframe.AvatarSession(persona_slug="public:lyra_persona-1.5-live")
+    avatar = keyframe.AvatarSession(persona_slug="public:cosmo_persona-1.5-live")
     await avatar.start(session, room=ctx.room)
 
     await session.start(


livekit-plugins/livekit-plugins-keyframe/livekit/plugins/keyframe/types.py
```diff
@@ -0,0 +1,21 @@
+from dataclasses import dataclass
+from typing import Literal
+
+Emotion = Literal["neutral", "happy", "sad", "angry"]
+
+
+@dataclass
```
tinalenguyen: was this meant to be passed to the `AvatarSession`? i don't see it used anywhere

livekit-plugins/livekit-plugins-keyframe/livekit/plugins/keyframe/avatar.py
```diff
@@ -0,0 +1,168 @@
...
+        expression and demeanor.
+
+        Args:
+            emotion: The emotion to express ('neutral', 'happy', 'sad', 'angry').
+        """
+        if self._room is None:
+            logger.warning("set_emotion() called before start()")
+            return
+
+        await self._room.local_participant.publish_data(
```
tinalenguyen, line 167: just wondering, are there no plans to support setting the emotion via an API call? i think that would be more ideal if possible

livekit-plugins/livekit-plugins-keyframe/livekit/plugins/keyframe/avatar.py
```diff
@@ -0,0 +1,168 @@
...
+            else _AVATAR_AGENT_IDENTITY
+        )
+        self._avatar_participant_name = (
+            avatar_participant_name
+            if utils.is_given(avatar_participant_name)
+            else _AVATAR_AGENT_NAME
+        )
+
+        # Resolve persona
+        self._persona_id = persona_id if utils.is_given(persona_id) else None
```
tinalenguyen: nit: i would add the check `if utils.is_given(persona_id) and utils.is_given(persona_slug)` before setting the variables

examples/avatar_agents/keyframe/README.md
```diff
@@ -0,0 +1,44 @@
...
+You can change the avatar's facial expression at runtime using `set_emotion()`:
+
+```python
+await avatar.set_emotion("happy")  # "neutral", "happy", "sad", "angry"
+```
+
+To let the LLM control the avatar's expression, register it as a tool:
+
+```python
+@session.tool()
```
tinalenguyen:
```suggestion
@function_tool()
```

livekit-plugins/livekit-plugins-keyframe/livekit/plugins/keyframe/version.py
tinalenguyen:
```suggestion
__version__ = "1.4.3"
```

livekit-plugins/livekit-plugins-keyframe/livekit/plugins/keyframe/avatar.py
kradkfl: Do you mean by way of a REST call to `api.keyframelabs.com`? 

If so, the latency would be pretty high. Additionally, the avatar session itself on the backend isn't listening to any central server calls once connected to the room, it's just listening to data channels (which is what we use here).

Are you imagining that the user of the avatar plugin wouldn't want to access the underlying avatar object and call this function themselves? Or that accessing the underlying avatar isn't ergonomic?

livekit-plugins/livekit-plugins-keyframe/livekit/plugins/keyframe/types.py
kradkfl: Ah left-over. Given that it's a one or the other situation for the arguments, I didn't actually find much value in grouping them, and it made for a weirder interface.

kradkfl: Removed in 1f72fa3.

livekit-plugins/livekit-plugins-keyframe/livekit/plugins/keyframe/avatar.py
kradkfl: Fixed in 1f72fa3.

examples/avatar_agents/keyframe/README.md
kradkfl: Fixed in 1f72fa3.

livekit-plugins/livekit-plugins-keyframe/livekit/plugins/keyframe/version.py
kradkfl: Fixed in 1f72fa3.

livekit-plugins/livekit-plugins-keyframe/livekit/plugins/keyframe/avatar.py
tinalenguyen: ah yes, i was just curious how users would update settings mid-session in other scenarios/use cases, but the latency using the data channels is much better as you said. iirc most of our other providers don't allow for dynamic updates like this, so this would be a new feature (very cool to see the changes in real-time!)

tinalenguyen (approved):
looks good to me! small ask - could you update the examples to use this setup instead for the `AgentSession`:

```
session = AgentSession(
        stt=inference.STT("deepgram/nova-3"),
        llm=inference.LLM("google/gemini-2.5-flash"),
        tts=inference.TTS("cartesia/sonic-3"),
        resume_false_interruption=False,
    )
```
i noticed that this setup called `set_emotion` more often as well, so it really seems like the avatar is reacting throughout the conversation. i didn't notice much of a difference in latency either

PR #4955, opened 2026-02-26
feat: support multimodal tool outputs (text + image)

### LINKED ISSUE #4893: feat: support ImageContent in tool return value
### Feature Type

### Feature Description

Currently, `@function_tool` functions only support primitive return types (`str`, `int`, `float`, `bool`, `None`). Returning `ImageContent` fails validation silently `fnc_call_out` is set to `None` and the LLM receives no tool response.

This makes it impossible to implement tools that return visual content to the LLM (e.g. a `read_whiteboard()` or `take_screenshot()` tool).

Several providers already support multimodal content in tool results at the API level (Anthropic `tool_result`, Gemini `FunctionResponse`), so this could be wired up for supported backends. OpenAI's agents-python SDK also supported this feature https://github.com/openai/openai-agents-python/pull/1898

If this is an existing feature, please let me know. If this is on the road map or the community would like to support this feaure, I could take it or help with it.

Manually inject `ImageContent` into `chat_ctx` inside the tool body and return a plain string description:

```python
@function_tool
async def read_screenshot(self, context: RunContext) -> str:
    b64 = base64.b64encode(get_screenshot()).decode()
    context.session.chat_ctx.add_message(
        role="user",
        content=["Current screenshot:", ImageContent(image=f"data:image/png;base64,{b64}")],
    )
    return "screenshot added to context."
```

This workaround makes the content visible to the LLM but it should originally being passed as tool result/function response instead of a user message.

### Additional Context

_No response_

tpirc3:
## Summary
- extend `FunctionCallOutput.output` to support `ImageContent` and `list[str | ImageContent]`
- add shared normalization/splitting/text-fallback helpers for tool outputs
- implement provider-specific handling for tool-result images:
  - native: OpenAI Responses, Anthropic, Google, AWS standard LLM path
  - fallback to text placeholder for unsupported paths (OpenAI chat default, Mistral, realtime variants)
- add optional OpenAI chat-completions flag `supports_tool_image_output` to support Qwen-compatible providers that accept tool image content
- fix telemetry/span/type assumptions by converting multimodal outputs to text where string-only sinks are required

## Why
Closes #4893

## Tests
- `make check`
- `uv run pytest tests/test_tool_output_multimodal.py tests/test_chat_ctx.py tests/test_tools.py -q`

### DIFF
--- a/livekit-agents/livekit/agents/llm/chat_context.py
+++ b/livekit-agents/livekit/agents/llm/chat_context.py
@@ -466,7 +466,10 @@ class ChatContext:
                 if isinstance(item.output, ImageContent):
                     item.output = ""
                 elif isinstance(item.output, list):
-                    filtered = [c for c in item.output if isinstance(c, str)]
+                    filtered: list[ToolOutputContent] = []
+                    for content in item.output:
+                        if isinstance(content, str):
+                            filtered.append(content)
                     item.output = filtered if filtered else ""
 
             items.append(item)
--- a/livekit-agents/livekit/agents/voice/generation.py
+++ b/livekit-agents/livekit/agents/voice/generation.py
@@ -628,7 +628,8 @@ async def _execute_tools_task(
 
                     if fnc_call_out := output.fnc_call_out:
                         current_span.set_attribute(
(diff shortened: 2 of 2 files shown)


tpirc3:
Sorry for the confusion with these two PRs.

PR #4961, opened 2026-02-26
feat(stt): add support for AssemblyAI u3-rt-pro model and mid-session updates

russellmartin-livekit:
https://github.com/livekit/agent-gateway/pull/319
https://github.com/livekit/e2e/pull/908
https://github.com/livekit/cloud-config/pull/8602
https://github.com/livekit/agents/pull/4961

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -419,6 +419,7 @@ class STT(stt.STT):
     ) -> STTOptions:
         """Create a sanitized copy of options with language override if provided."""
         options = replace(self._opts)
+        options.extra_kwargs = dict(options.extra_kwargs)
 
         if is_given(language):
             options.language = Language(language)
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -437,6 +439,21 @@ def __init__(
         self._reconnect_event = asyncio.Event()
         self._speaking = False
         self._speech_duration: float = 0
+        self._ws: aiohttp.ClientWebSocketResponse | None = None
+
+    async def update_session(self, *, extra_kwargs: dict[str, Any]) -> None:
```
chenghao-mou: Adding the support makes sense, but this function is not called anywhere. Do we expect more changes to come?

russellmartin-livekit: When u3-rt-pro is enabled in gateway this will allow mid-session updates by users (will also be able to do it for deepgram-flux when its updated in gateway)

There is an example here https://github.com/livekit/e2e/pull/908/changes#diff-be337c20cd29f5768663fa4227a69096f86c10817d41be3a47aa9b949619ae37R299

davidzhao: shouldn't this just be `stt.update_options`?  where it attempts to update the session if supported.

we should be creating the same API across the board.. versus having a special API only for specific models

PR #4962, opened 2026-02-26
feat(stt): expose turn_detection mode to STT streams

gsharp-aai:
## Summary

- Thread the `turn_detection` mode (`"stt"`, `"vad"`, `"realtime_llm"`, `"manual"`) from `AudioRecognition` through `ModelSettings` to `RecognizeStream`
- Add a `turn_detection` property on `RecognizeStream` (following the existing `start_time_offset` pattern) so any STT plugin can read the active turn detection mode
- Add `turn_detection` field to `ModelSettings` and populate it in `AudioRecognition._stt_task`

## Motivation

STT plugins currently have no way to know what `turn_detection` mode the user configured at the Agent/AgentSession level. This information stops at `AudioRecognition` and is never passed to the STT stream.

Exposing this value enables plugins to adapt their default behavior based on the turn detection strategy — for example, adjusting end-of-turn parameter defaults when a 3rd-party turn detector is handling turn detection.

## Changes

- **`ModelSettings`** (`voice/agent.py`): Added `turn_detection: str | None = None` field
- **`AudioRecognition._stt_task`** (`voice/audio_recognition.py`): Populate `turn_detection` from `self._turn_detection_mode` when constructing `ModelSettings`
- **`RecognizeStream`** (`stt/stt.py`): Added `turn_detection` property/setter (defaults to `None`)
- **`Agent.default.stt_node`** (`voice/agent.py`): Set `stream.turn_detection` from `model_settings.turn_detection`

## Non-breaking

All changes use default values (`None`), so existing plugins, custom `stt_node` overrides, and `ModelSettings()` calls continue to work without modification.

### DIFF
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -7,7 +7,10 @@ from collections.abc import AsyncIterable, AsyncIterator
 from dataclasses import dataclass, field
 from enum import Enum, unique
 from types import TracebackType
-from typing import Generic, Literal, TypeVar
+from typing import TYPE_CHECKING, Generic, Literal, TypeVar
+
+if TYPE_CHECKING:
+    from ..voice.audio_recognition import TurnDetectionMode
 
 from pydantic import BaseModel, ConfigDict, Field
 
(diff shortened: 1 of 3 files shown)


gsharp-aai:
This PR is ready for review ✅ 

The latest commit updates the `turn_detection` type from `str | None` to `TurnDetectionMode | None` so plugins can distinguish between no turn detection configured (None) vs. a custom turn detector model being used (e.g. EnglishModel instance). Prior to that commit, those two scenarios were indistinguishable.

PR #4964, opened 2026-02-26
feat: add Backboard.io LLM plugin (persistent memory + RAG)

KJFromMicromonic:


KJFromMicromonic:
Friendly follow-up on this PR. CI is green and I’ve addressed prior feedback.
Could a maintainer confirm preferred path: merge as-is, request changes, or keep as external plugin?
If helpful, I can split this into smaller PRs and own ongoing maintenance/docs for Backboard.io integration.

nirdoshi:
CI is green, are there any additional tests or benchmarks that reviewers would like to see before merging?

Happy to add more coverage if needed.

ferhimedamine:
Persistent memory combined with retrieval-augmented context is the right architecture for voice agents that need to remember past conversations.

The key design choice is whether the memory layer does retrieval-augmented generation (RAG) inline or just provides context that the agent can use. In our experience, providing context via memory recall is cleaner — the agent receives relevant past memories as additional context, then decides how to use them. This keeps the memory system focused on storage/retrieval and leaves reasoning to the LLM.

For voice agents specifically, you also want semantic compression at store time (extract key facts from verbose transcripts rather than storing raw audio transcripts), and low-latency retrieval (sub-50ms) since the user is waiting in real-time. Our memory server handles both with vector + BM25 hybrid search over compressed memories.

Self-hosted: https://github.com/Dakera-AI/dakera-deploy

PR #4965, opened 2026-02-27
feat(assemblyai): add u3-rt-pro model plus mid-stream updates, SpeechStarted, and ForceEndpoint support

gsharp-aai:
## Summary

Adds Universal-3-Pro (`u3-rt-pro`) model support to the AssemblyAI streaming plugin with several improvements to the existing streaming implementation.

### New model
- Add `u3-rt-pro` to the supported `model` literals
- Accept deprecated `u3-pro` name with a warning, remapped to `u3-rt-pro`
- Add `prompt` parameter (u3-rt-pro only) for custom transcription instructions, validated at init
- Default `language_detection` to `True` for `u3-rt-pro` model (previously only defaulted for `multilingual` models)
- Default `min_end_of_turn_silence_when_confident` and `max_turn_silence` to **100ms** for `u3-rt-pro` for optimal out-of-the-box performance/latency across most LiveKit configurations. This provides finals quickly for third-party turn detection models while still working well with built-in turn detection.
  - If a user sets min without setting max, max defaults to match min rather than its API default of 1000ms. Both parameters are fully overridable. For reference, the AssemblyAI API defaults are `min=100ms` and `max=1000ms.` We will clearly document the plugin's defaults and how to override them.

### Speaker diarization support
- Add `speaker_labels` parameter (`bool`) to enable speaker diarization on the streaming connection
- Add `max_speakers` parameter (`int`, 1-10) to set the maximum number of speakers when diarization is enabled
- Both are connection-level parameters sent as WebSocket query params at connect time (not updatable mid-stream via `UpdateConfiguration`)

### Mid-stream configuration updates
- Replace reconnect-based update_options() with in-place UpdateConfiguration websocket messages (previously, updating options would tear down and restart the entire websocket connection)
- Queue-based approach (asyncio.Queue) for thread-safe sync-to-async communication, with a dedicated coroutine that sends config messages immediately and independently of audio flow
- Supported fields: `prompt`, `keyterms_prompt`, `max_turn_silence`, `min_end_of_turn_silence_when_confident`, `end_of_turn_confidence_threshold`, `vad_threshold`
- Add `keyterms_prompt` to `update_options()` (previously only available at connection time)

### Rename `min_end_of_turn_silence_when_confident` → `min_turn_silence`

The AssemblyAI API accepts both names via `AliasChoices`. This promotes `min_turn_silence` as the primary parameter — shorter and consistent with `max_turn_silence`. The old name is still accepted but logs a deprecation warning. If both are provided, `min_turn_silence` takes precedence.

### New websocket message support
- **`SpeechStarted`**: Handle new server event, mapped to `SpeechEventType.START_OF_SPEECH` for barge-in detection
- **`ForceEndpoint`**: Add `force_endpoint()` method to immediately finalize the current turn via `{"type": "ForceEndpoint"}`

### Fixes
- Set `interim_results=True` in `STTCapabilities` (was incorrectly `False` despite emitting `INTERIM_TRANSCRIPT` events)
- Fix `send_config_task` shutdown hang by separating it from `asyncio.gather` so it is cancelled in `finally` instead of blocking graceful shutdown

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -54,7 +54,7 @@ class STTOptions:
     ] = "universal-streaming-english"
     language_detection: NotGivenOr[bool] = NOT_GIVEN
     end_of_turn_confidence_threshold: NotGivenOr[float] = NOT_GIVEN
-    min_end_of_turn_silence_when_confident: NotGivenOr[int] = NOT_GIVEN
+    min_turn_silence: NotGivenOr[int] = NOT_GIVEN
     max_turn_silence: NotGivenOr[int] = NOT_GIVEN
     format_turns: NotGivenOr[bool] = NOT_GIVEN
     keyterms_prompt: NotGivenOr[list[str]] = NOT_GIVEN
@@ -79,7 +79,7 @@ class STT(stt.STT):
         ] = "universal-streaming-english",
         language_detection: NotGivenOr[bool] = NOT_GIVEN,
         end_of_turn_confidence_threshold: NotGivenOr[float] = NOT_GIVEN,
-        min_end_of_turn_silence_when_confident: NotGivenOr[int] = NOT_GIVEN,
+        min_turn_silence: NotGivenOr[int] = NOT_GIVEN,
         max_turn_silence: NotGivenOr[int] = NOT_GIVEN,
         format_turns: NotGivenOr[bool] = NOT_GIVEN,
         keyterms_prompt: NotGivenOr[list[str]] = NOT_GIVEN,
@@ -90,6 +90,8 @@ class STT(stt.STT):
         http_session: aiohttp.ClientSession | None = None,
         buffer_size_seconds: float = 0.05,
         base_url: str = "wss://streaming.assemblyai.com",
+        # Deprecated — use min_turn_silence instead
+        min_end_of_turn_silence_when_confident: NotGivenOr[int] = NOT_GIVEN,
     ):
         """
         Args:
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -320,37 +398,49 @@ async def recv_task(ws: aiohttp.ClientWebSocketResponse) -> None:
...
             finally:
-                if ws is not None:
-                    await ws.close()
+                await utils.aio.gracefully_cancel(config_task, *tasks)
+        finally:
+            if ws is not None:
+                await ws.close()
 
     async def _connect_ws(self) -> aiohttp.ClientWebSocketResponse:
+        # u3-rt-pro defaults: min=100, max=min (so both 100 unless overridden)
```
davidzhao, line 424: why does u3-rt-pro work differently here?

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -320,37 +398,49 @@ async def recv_task(ws: aiohttp.ClientWebSocketResponse) -> None:
...
+                self._opts.min_turn_silence if is_given(self._opts.min_turn_silence) else 100
+            )
+            max_silence = (
+                self._opts.max_turn_silence
+                if is_given(self._opts.max_turn_silence)
+                else min_silence
+            )
+        else:
+            min_silence = (
+                self._opts.min_turn_silence if is_given(self._opts.min_turn_silence) else None
```
davidzhao, line 438: setting it to None would lead to `min_turn_silence: null` to be sent to your end. is that desired/supported?

if not, I would recommend only setting these fields when given

gsharp-aai: None values are filtered out (via [filtered_config](https://github.com/gsharp-aai/agents/blob/663c963fa4072457e882e5bac75b1724b5246738/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py#L478-L482)) before building the query string, so they're never sent to the API.

PR #4967, opened 2026-02-27
feat(stt): add keyterms parameter in Elevenlabs STT plugin

Arjun-A-I:
# Summary

- Add keyterms parameter to ElevenLabs STT plugin for batch recognition, allowing users to bias transcription towards specific words or phrases (e.g. product names, technical terms)
- Supported on Scribe v2 models only (not realtime streaming), matching the ElevenLabs API specification
- Also exposed via update_options() for runtime updates

Related Elevenlabs API Docs :
- https://elevenlabs.io/docs/api-reference/speech-to-text/convert#request.body.keyterms.keyterms


Arjun-A-I:
@theomonnom can you take a look

PR #4971, opened 2026-02-28
fix(sarvam-tts): add keepalive ping and stale connection detection to prevent audible silence

sanskar1104srivastava:
Problem

In production IVR systems, the Sarvam TTS plugin causes audible silence in two scenarios:

1. Between agent turns — Sarvam server closes idle WebSocket connections, and the pool returns a stale connection on the next turn.
2. At the start of a new call — the pool reuses a WebSocket connection from a previous ended call that Sarvam has already closed server-side.

Both result in ClientConnectionResetError: Cannot write to closing transport, causing a 0.5s silence gap while the plugin retries with a fresh connection.

Fix

1. Keepalive ping — sends {"type": "ping"} every 10 seconds while the WebSocket is idle between agent turns, as per the official Sarvam WebSocket API spec. This prevents Sarvam server from closing the connection due to inactivity.

2. Stale connection check — checks ws.closed and ws.close_code before attempting to write. This detects dead pool connections immediately and forces a fresh connection instead of failing mid-write.

Additional fixes based on Sarvam API documentation

- pitch validation range corrected from -20/+20 to -0.75/+0.75
- pace validation range for bulbul:v2 corrected from 0.5-2.0 to 0.3-3.0
- speech_sample_rate now sent as string, matching the API schema which defines it as a string enum
- default sample rate for bulbul:v3-beta corrected from 22050 to 24000 Hz
- missing output_audio_codec config parameter added with full validation

Testing

Validated in a production IVR system handling real concurrent calls. The ClientConnectionResetError no longer occurs after applying both fixes.

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
@@ -333,6 +333,10 @@ class TTS(tts.TTS):
         http_session: aiohttp.ClientSession | None = None,
         send_completion_event: bool = True,
     ) -> None:
+        # Fix: correct default sample rate for v3/v3-beta before passing to parent
+        if speech_sample_rate == 22050 and (model == "bulbul:v3-beta" or model == "bulbul:v3"):
+            speech_sample_rate = 24000
+
         super().__init__(
             capabilities=tts.TTSCapabilities(streaming=True),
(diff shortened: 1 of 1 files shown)


sanskar1104srivastava:
Hi @theomonnom @davidzhao would appreciate a review when you get a chance. 
This fixes a production issue causing audible silence in IVR systems using 
the Sarvam TTS plugin. All CI checks are passing.

PR #4980, opened 2026-03-03
Add Phonic Plugin to LiveKit agents

qionghuang6:
Implements Phonic Plugin in Livekit Agents Python SDK. 
(Original LiveKit Agents JS implementation: https://github.com/Phonic-Co/livekit-agents-js/tree/main/plugins/phonic)

## Changes
- Implements `RealtimeSession` and `RealtimeModel`, supporting STS and tool calls
- Uses Phonic Python SDK
- Adds example to `examples/voice-agents`

## Testing
- Tested in Agent playground, including tool use.

## Demo video
- https://drive.google.com/file/d/1-7iP-g1HrRmRHAcjZf_kE9mRR3A5gCH_/view?usp=sharing

### DIFF
--- a/livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
@@ -551,6 +551,7 @@ class RealtimeSession(llm.RealtimeSession):
             )
 
         if not gen.text_ch.closed:
+            gen.text_ch.send_nowait("")
             gen.text_ch.close()
         if not gen.audio_ch.closed:
             gen.audio_ch.close()


livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -0,0 +1,684 @@
...
+    def push_audio(self, frame: rtc.AudioFrame) -> None:
+        if self._session_should_close.is_set() or not self._ready_to_start or not self._socket:
+            return
+
+        for f in self._resample_audio(frame):
+            for nf in self._bstream.write(f.data.tobytes()):
+                b64_audio = base64.b64encode(nf.data.tobytes()).decode("utf-8")
+                asyncio.create_task(
+                    self._socket.send_audio_chunk(AudioChunkPayload(audio=b64_audio))
+                )
```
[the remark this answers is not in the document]
qionghuang6: Changed to a queue-ing system using LK channels

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -0,0 +1,688 @@
...
+                    )
+                    sent = True
+
+        if not sent:
+            logger.warning(
+                "update_chat_ctx called but no new tool call outputs to send. "
+                "Phonic does not support general chat context updates."
+            )
+        else:
+            self._start_new_assistant_turn()
```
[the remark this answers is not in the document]
qionghuang6: No, this is intentional. We want to restart generation if a tool call result is sent. (the assistant isn't considered to be done speaking when tool calls are sent, so `assistant_started_speaking` is never sent, we need to restart the generation manually)

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -0,0 +1,688 @@
...
+                "project": self._opts.project,
+                "welcome_message": self._opts.welcome_message,
+                "generate_welcome_message": self._opts.generate_welcome_message,
+                "system_prompt": self._opts.instructions,
+                "voice_id": self._opts.voice,
+                "input_format": "pcm_44100",
+                "output_format": "pcm_44100",
+                "recognized_languages": self._opts.languages,
+                "audio_speed": self._opts.audio_speed,
+                "tools": tools_payload,
```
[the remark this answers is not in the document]
qionghuang6: Hmm, I suppose this is correct if you used a phonic agent. Fixing!

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -0,0 +1,688 @@
...
+        return self._tools.copy()
+
+    async def update_instructions(self, instructions: str) -> None:
+        if self._config_sent:
+            logger.warning(
+                "update_instructions called after config was already sent. "
+                "Phonic does not support updating instructions mid-session."
+            )
+            return
+        self._opts.instructions = instructions
```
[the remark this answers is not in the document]
qionghuang6: This is the pattern that the Google realtime and other realtime models use.

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -0,0 +1,688 @@
...
+        self._instructions_ready.set()
+
+    async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
```
tinalenguyen, line 261: at the start of the session, `update_chat_ctx` is called with a fresh chat context and i see this log:

`WARNI… livekit.…ns.phonic update_chat_ctx called but no new tool call outputs to send. Phonic does not support general chat context updates. `

perhaps we could add a check if the session just started?

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -0,0 +1,688 @@
...
+    def commit_audio(self) -> None:
+        logger.warning("commit_audio is not supported by the Phonic realtime model.")
+
+    def clear_audio(self) -> None:
+        logger.warning("clear_audio is not supported by the Phonic realtime model.")
+
+    def interrupt(self) -> None:
+        logger.warning(
+            "interrupt() is not supported by Phonic realtime model. "
+            "User interruptions are automatically handled by Phonic."
```
tinalenguyen: we call `interrupt()` when input speech is detected ([relevant lines](https://github.com/livekit/agents/blob/8343db123e6186dfbbed20879bf2b70055382435/livekit-agents/livekit/agents/voice/agent_activity.py#L1174-L1185)), so even if there is no ongoing speech to interrupt this log appears. adding this check should filter out those instances
```suggestion
    def interrupt(self) -> None:
        if self._current_generation:
            logger.warning(
                "interrupt() is not supported by Phonic realtime model. "
                "User interruptions are automatically handled by Phonic."
            )
```

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -0,0 +1,691 @@
...
+    def _handle_audio_chunk(self, message: AudioChunkResponsePayload) -> None:
+        # In Phonic, audio chunks can come in when assistant isn't explicitly active.
+        # We start a generation if text is present to align with the framework pattern.
+        if self._current_generation is None and message.text:
+            logger.debug("Starting new generation due to text in audio chunk")
+            self._start_new_assistant_turn()
+
+        gen = self._current_generation
+        if gen is None:
+            return
```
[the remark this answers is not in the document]
qionghuang6: This is intentional. Phonic can send silent audio chunks (or audio chunks mixed in with background noise), but we want to ignore these chunks in our integration with the Livekit agent SDKs.

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
qionghuang6: Yup, just added a check to log this only if the config message has already been sent to Phonic.

qionghuang6:
Thanks @tinalenguyen . I made some quick changes. I also just reset `uv.lock` to what it is on `main`. Just to confirm, we don't want to update `uv.lock` here even though we changed `pyproject.toml`?

Also, I think the ruff CI is having some issues due to GitHub being partially down? We might need to re-run it.  https://github.com/livekit/agents/actions/runs/22639600339/job/65611716048?pr=4980

tinalenguyen:
The `uv.lock` file is usually auto-updated from version releases, and looks like the ruff tests passed! 

Just tested again, I consistently see this log when a tool is called: 
`_SegmentSynchronizerImpl.playback_finished called before text/audio input is done {"text_done": false, "audio_done": true}
`

I'm checking if this is related to any plugin changes, I don't recall seeing that initially

tinalenguyen:
@qionghuang6 I think marking the generation as done upon `"assistant_finished_speaking"` when a tool is called causes this behavior. The tool call starts a new generation with no text and audio, the expected behavior is that it's included in the current generation as well

Is there an event similar to `response.done` to mark the end of a turn with a function call?

qionghuang6:
EDIT: What I said below assumes the assistant speaks before calling the tool. In that case, we don't get the warning, but the warning happens when the assistant doesn't say anything before calling the tool, so nothing makes its way into the text channel, and `text_done` doesn't become `true` by the time that generation is closed. 

@tinalenguyen Are you saying that when a tool is called, it is in it's own generation without audio or text? I think the current implementation uses the same generation as the existing audio and text. (Unless `if self._current_generation is None:` is not satisfied)

`assistant_finished_speaking` is our version of `response.done`, it's just that the turn is not considered to be done (i.e. the assistant has not finished speaking) until it has received the results of the tool call and finished speaking after receiving the results of the tool call.

The current flow of events when a tool call happens is:
`assistant_started_speaking` -> New generation begins
`audio_chunk` 
...
`audio_chunk` 
`tool_call` -> Tool call put in function channel, we call `_close_current_generation`, current generation closes.
...
After the tool is executed, Livekit Agents calls `update_chat_ctx` which calls `_start_new_assistant_turn`, starting a new generation.
Agent continues speaking:
`audio_chunk`
...
`assistant_finished_speaking`. 

For turns without tool calls, it is:
`assistant_started_speaking` -> New generation begins
`audio_chunk` 
...
`audio_chunk` 
`assistant_finished_speaking`

I tried logging when `assistant_finished_speaking` is received, and it seems that I am seeing this `_SegmentSynchronizerImpl` warning prior to the `assistant_finished_speaking` event. 

Continuing to look into this.

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -0,0 +1,692 @@
...
+        if gen.output_text:
+            self._chat_ctx.add_message(
+                role="assistant",
+                content=gen.output_text,
+                id=gen.response_id,
+                interrupted=interrupted,
+            )
+
+        if not gen.text_ch.closed:
+            gen.text_ch.send_nowait("")
```
qionghuang6, line 554: I investigated more deeply, and I think the reason we were seeing
```
_SegmentSynchronizerImpl.playback_finished called before text/audio input is done {"text_done": false, "audio_done": true}
```

is because of this chain of events:

User says -> "Please turn on light A 5 for me"
Phonic -> `assistant_started_speaking` (Plugin -> starts a new generation)
Phonic -> `tool_call` (Plugin -> closes current generation, but no text chunks was put in the text_ch by the time it closes, so )

(We need to close the current generation so that tools actually get sent back to the plugin through `update_chat_ctx)

However, because no text ever made it into the text, channel, the `text_done` never becomes true. 

A quick fix for this is to just put an empty string in the text channel, so that the Text/Audio Synchronizer can change the state from `false` to `true`.

tinalenguyen: Thank you for the detailed overview! Just to confirm, the tool call results are received after the `"assistant_started_speaking"` event and that's why there's something like an in-between empty generation?

qionghuang6: Yup, the `assistant_started_speaking` fires before the `tool_call`, because `assistant_started_speaking` is basically our version of saying a turn has started, and tool calls need to happen within an assistant turn.

tinalenguyen: Ahh okay, I see now! Thanks again

PR #4983, opened 2026-03-03
[Inworld] Flush to drain decoder on every audio chunk from server

ianbbqzy:
The **emitter's non-PCM path still routes through `_decode_task`**, which on `FlushSegment` does `audio_decoder.end_input()` + `await decode_atask` + `audio_decoder = None`. This means:

1. **Without per-chunk `flush()`**: All audio bytes accumulate in the decoder. The final `output_emitter.flush()` at line 1137 is the only `FlushSegment`. TTFB = time to receive **all** audio from server, not just the first chunk.

2. **With per-chunk `flush()`**: Each chunk triggers `FlushSegment` → decoder drains → `_flush_frame()` → `SynthesizedAudio` is emitted. TTFB = time to first chunk from server (correct).

This is an Inworld-specific issue because Inworld's HTTP API returns **one JSON-line per audio chunk** (each with its own WAV payload), unlike other providers which returns a single continuous raw PCM byte stream. For Inworld, each JSON line *is* a discrete audio segment, and delaying the flush until the end defeats the purpose of streaming.

====

Also updated timestamps parsing logic to always add a trailing space regardless if it's the end of a chunk or not. Because in Inworld case, end of a chunk could just be end of a phrase rather than end of a sentence

### DIFF
--- a/livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
+++ b/livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
@@ -51,9 +51,9 @@ from .version import __version__
 USER_AGENT = f"livekit-agents-py/{__version__}"
 
 DEFAULT_BIT_RATE = 64000
-DEFAULT_ENCODING = "OGG_OPUS"
-DEFAULT_MODEL = "inworld-tts-1"
-DEFAULT_SAMPLE_RATE = 48000
+DEFAULT_ENCODING = "LINEAR16"
+DEFAULT_MODEL = "inworld-tts-1.5-max"
+DEFAULT_SAMPLE_RATE = 24000
 DEFAULT_URL = "https://api.inworld.ai/"
 DEFAULT_WS_URL = "wss://api.inworld.ai/"
 DEFAULT_VOICE = "Ashley"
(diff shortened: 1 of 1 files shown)


ianbbqzy:
@tinalenguyen @davidzhao PTAL, Thanks!

tinalenguyen:
Hi @ianbbqzy, thanks for the PR! Few notes and Q's:
- Could you update the TTS docstring with the new defaults?
- If each chunk is already flushed, might be worth removing the outer flush [here](https://github.com/livekit/agents/blob/9a49e199bc56406855603e9b5df319ad84291350/livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py#L1136)

> Also updated timestamps parsing logic to always add a trailing space regardless if it's the end of a chunk or not. Because in Inworld case, end of a chunk could just be end of a phrase rather than end of a sentence

Could you elaborate more on this? I also noticed that the last punctuation is not present, not sure if that was pre-existing behavior

ianbbqzy:
Hi @tinalenguyen 

Fixed the two issues you raised. Thanks!

Regarding timestamps, Inworld TTS currently returns word timestamps without punctuations. Previously I thought a space is not needed at the end of sentence, but the fact is the sentence chunking logic doesn't really affect how timestamps are generated. and hence space should be added to every  timestamp result to prevent  transcripts not having spaces between words.

PR #4984, opened 2026-03-03
Pass through additional params to LemonSlice when using the LemonSlice Avatar

jp-lemon:
This will open the door for us to handle special one-off params when necessary.


tinalenguyen (approved):
thanks, looks good to me!

PR #4985, opened 2026-03-03
(openai responses): add websocket connection pool

tinalenguyen:
when there are two parallel streams, restart and send the full context on the next request
each request will use its own websocket connection (edit: and each WS connection is independent of response IDs)

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
@@ -116,12 +116,10 @@ class _ResponsesWebsocket:
                     aiohttp.WSMsgType.CLOSE,
                     aiohttp.WSMsgType.CLOSING,
                 ):
-                    close_code = raw_msg.data if isinstance(raw_msg.data, int) else -1
-                    close_reason = raw_msg.extra or "connection closed unexpectedly"
                     raise APIStatusError(
-                        f"OpenAI Responses WebSocket closed: {close_reason}",
-                        status_code=close_code,
-                        retryable=False,
+                        "OpenAI Responses WebSocket connection closed unexpectedly",
+                        status_code=raw_msg.data or -1,
+                        body=f"{raw_msg.data=} {raw_msg.extra=}",
                     )
                 if raw_msg.type != aiohttp.WSMsgType.TEXT:
                     continue


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -41,140 +40,88 @@
...
-        self._ws_conn: aiohttp.ClientWebSocketResponse | None = None
         self._session: aiohttp.ClientSession | None = None
-        self._input_ch = utils.aio.Chan[dict]()
-        self._output_queue: collections.deque[utils.aio.Chan[dict]] = collections.deque()
-        self._run_task: asyncio.Task | None = None
+
+        self._pool = utils.ConnectionPool[aiohttp.ClientWebSocketResponse](
+            connect_cb=self._create_ws_conn,
+            close_cb=self._close_ws,
+            max_session_duration=3600,
```
chenghao-mou, line 61: But we can't reuse the same WS connection for different conversations, right? Even with `store=True`, the server has to rehydrate the chat history if the ws connection expects a different `previous_response_id`.

tinalenguyen: I could have worded my comment better, but each websocket connection is independent of `previous_response_id`

chenghao-mou: But the server-side in-memory cache for each connection seems very dependent:

> On an active WebSocket connection, the service keeps one previous-response state in a connection-local in-memory cache (the most recent response). Continuing from that most recent response is fast because the service can reuse connection-local state. Because the previous-response state is retained only in memory and is not written to disk, you can use WebSocket mode in a way that is compatible with store=false and Zero Data Retention (ZDR).
> 
> If a previous_response_id is not in the in-memory cache, behavior depends on whether you store responses:
> 
> With store=true, the service may hydrate older response IDs from persisted state when available. Continuation can still work, but it usually loses the in-memory latency benefit.
> With store=false (including ZDR), there is no persisted fallback. If the ID is uncached, the request returns previous_response_not_found.

Do we see `previous_response_not_found` when `store=false` if you reuse the same connection for two conversations?

tinalenguyen: Ah we don't use `previous_response_id` when `store=False`, we just send the entire context in that case ([relevant line](https://github.com/livekit/agents/blob/8343db123e6186dfbbed20879bf2b70055382435/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py#L347))

chenghao-mou: Gotcha, that makes sense.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -41,140 +40,88 @@
...
         self._base_url = base_url if base_url else OPENAI_RESPONSES_WS_URL
 
-        self._ws_conn: aiohttp.ClientWebSocketResponse | None = None
         self._session: aiohttp.ClientSession | None = None
-        self._input_ch = utils.aio.Chan[dict]()
-        self._output_queue: collections.deque[utils.aio.Chan[dict]] = collections.deque()
-        self._run_task: asyncio.Task | None = None
+
+        self._pool = utils.ConnectionPool[aiohttp.ClientWebSocketResponse](
+            connect_cb=self._create_ws_conn,
```
chenghao-mou: nit: I think we can just call it `_create_ws`, since the connection part is implied.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -41,140 +40,88 @@
...
+        await self._pool.aclose()
 
-                    def _default(o: object) -> object:
-                        if isinstance(o, openai.BaseModel):
-                            return o.model_dump()
-                        raise TypeError(f"unexpected type {type(o)}")
+    async def send_request(self, msg: dict) -> AsyncGenerator[dict, None]:
+        def _default(o: object) -> object:
+            if isinstance(o, openai.BaseModel):
+                return o.model_dump()
```
chenghao-mou: nit: `model_dump(mode='json')`

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -41,140 +40,88 @@
...
-                    return
-                try:
+    async def aclose(self) -> None:
+        await self._pool.aclose()
 
-                    def _default(o: object) -> object:
-                        if isinstance(o, openai.BaseModel):
-                            return o.model_dump()
-                        raise TypeError(f"unexpected type {type(o)}")
+    async def send_request(self, msg: dict) -> AsyncGenerator[dict, None]:
```
chenghao-mou: nit: the name seems misleading in a sense that it sends the request but also receives the responses. Maybe `process_request`?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -41,140 +40,88 @@
...
+                raise APIConnectionError("failed to send request over WebSocket") from e
 
-        @utils.log_exceptions(logger=logger)
-        async def _recv_task() -> None:
             while True:
-                msg = await ws_conn.receive()
-                if msg.type in (
+                raw_msg = await ws.receive()
+                if raw_msg.type == aiohttp.WSMsgType.ERROR:
+                    raise APIConnectionError("OpenAI Responses WebSocket error") from raw_msg.data
```
chenghao-mou: I think we should use `APIStatusError` for errors that arrive after the connection has been set up, with proper error code propagation.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -41,140 +40,88 @@
...
+                if raw_msg.type in (
                     aiohttp.WSMsgType.CLOSED,
                     aiohttp.WSMsgType.CLOSE,
                     aiohttp.WSMsgType.CLOSING,
                 ):
-                    if not self._output_queue:  # if there are no more pending requests
-                        return
                     raise APIConnectionError(
```
chenghao-mou: Same here about the error class.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -389,6 +330,19 @@ def __init__(
...
+        if self._llm._active_streams > 0:
+            self._llm._parallel_generation = True
+        self._llm._active_streams += 1
+        try:
+            await self._run_impl()
+        finally:
+            self._llm._active_streams -= 1
+            if self._llm._active_streams == 0 and self._llm._parallel_generation:
+                self._llm._prev_resp_id = ""
+                self._llm._prev_chat_ctx = None
```
chenghao-mou, line 348: we still need to reset these two even when `self._llm._parallel_generation` is False, right?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -465,7 +418,7 @@ async def _run(self) -> None:
     def _parse_ws_event(self, event: dict) -> ResponseStreamEvent | None:
         event_type = event.get("type", "")
         if event_type == "error":
-            return ResponseErrorEvent.model_validate(event)
+            return ResponseErrorEvent.model_validate({**event, **event.get("error", {})})
```
chenghao-mou: This will override the `type` field because both the event and the error have it, but only the even-level should be allowed: https://developers.openai.com/api/docs/guides/websocket-mode/#errors-to-handle

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
tinalenguyen: when `self._llm._parallel_generation` is `False` and generations are serial, `prev_resp_id` and `prev_chat_ctx` are maintained in [these lines](https://github.com/livekit/agents/blob/7027f6f620fb850cab3b845bb7362d8ea335d51a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py#L521-L522)

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
tinalenguyen: how about `generate_response`?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -106,14 +106,22 @@ def _default(o: object) -> object:
...
                 ):
-                    raise APIConnectionError(
-                        message="OpenAI Responses WebSocket connection closed unexpectedly"
+                    close_code = raw_msg.data if isinstance(raw_msg.data, int) else -1
+                    close_reason = raw_msg.extra or "connection closed unexpectedly"
+                    raise APIStatusError(
+                        f"OpenAI Responses WebSocket closed: {close_reason}",
+                        status_code=close_code,
+                        retryable=False,
                     )
```
chenghao-mou, line 123: raw_msg has both .data and .extra we can leverage:

```python
APIStatusError(
                        "AssemblyAI connection closed unexpectedly",
                        status_code=ws.close_code or -1,
                        body=f"{msg.data=} {msg.extra=}",
                    )
```

PR #4987, opened 2026-03-03
add modality-aware Instructions with audio/text variants

longcw:


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -33,6 +33,121 @@
...
+
+    @property
+    def text(self) -> str:
+        """The text variant of the instructions.
+
+        Falls back to the audio variant when no text variant was provided.
+        """
+        return self._text_variant if self._text_variant is not None else self._audio_variant
+
+    def for_modality(self, modality: Literal["audio", "text"]) -> Instructions:
```
chenghao-mou: nit: should we use `to_modality`, similar to our `to_provider_format`?

davidzhao: the `for_` prefix seems unusual. `as_` or `to_` are a bit more pythonian.

longcw: renamed to `as_modality`

Issue #4969, opened 2026-02-27
llm.ChatMessage MetricsReport.started_speaking_at misalignment


hudson-worden:
### Bug Description

We are seeing an issue where the `start_speaking_at` within `llm.ChatMessage` is misaligned with what we'd expect. Below is a visualization from our telemetry tooling. We're using the livekit opentelemetry traces with custom spans associated with `llm.ChatMessage` to visualize them along side everything else.

We created such spans with the following code within a `conversation_item_added` session hook

```
from opentelemetry import trace

...

def _on_conversation_item_added(self, msg_ev: ConversationItemAddedEvent) -> None:
    if msg_ev.item.type == "message":
        tracer = trace.get_tracer(__name__)
        item_copy = msg_ev.item.copy()
        started_at = item_copy.metrics.get("started_speaking_at")
        stopped_at = item_copy.metrics.get("stopped_speaking_at")
        if started_at is not None and stopped_at is not None:
            start_ns = int(started_at * 1_000_000_000)
            end_ns = int(stopped_at * 1_000_000_000)

            span_name = "chat_item_message_assistant" if item_copy.role == "assistant" else "chat_item_message_user"
            if item_copy.interrupted:
                span_name += "_interrupted"
            span = tracer.start_span(
                span_name,
                start_time=start_ns,
            )
            span.set_attribute("item_role", item_copy.role)
            span.set_attribute("item_id", item_copy.id)
            span.set_attribute("item_type", item_copy.type)
            span.set_attribute("item_content", item_copy.content)
            span.set_attribute("item_interrupted", item_copy.interrupted)
            span.end(end_time=end_ns)
```

<img width="1056" height="868" alt="Image" src="https://github.com/user-attachments/assets/c08d3ba8-a96d-4bbb-88ee-47588da44862" />

### Expected Behavior

What we'd expect is that the `started_speaking_at` attribute be aligned with the start of the `user_speaking` span. The above picture shows the dotted lines of what we'd expect.

### Reproduction Steps

```bash
Create a custom span like shown above. Before initiating a reply to the agent make a bunch of noise ahead of time. This is to trip what I believe is Voice Activity Detection.
```

### Operating System

amazon linux

### Models Used

stt: deepgram nova-3,  llm: open ai gpt-4o, tts: elevenlabs eleven_turbo_v2

### Package Versions

```bash
livekit-agents==1.4.3
```

### Proposed Solution

```python

```

### Screenshots and Recordings

[...]

PR #4988, opened 2026-03-03
Fix VAD eou detection is exiting early

hudson-worden:
Issue:
There seems to be a case where _stt is defined and there's no transcript yet, but the` _turn_detection_mode` is set to "vad". VAD doesn't need a transcript for eou detection, so the consequence is we're exiting earlier than necessary. And the consequence of that is the `started_speaking_at` is older than it should be b/c it's never reset [here](https://github.com/livekit/agents/blob/8343db123e6186dfbbed20879bf2b70055382435/livekit-agents/livekit/agents/voice/audio_recognition.py#L657)

Here's what impact the bug has. See the context below on how I obtained this.

image

It's showing that we invoked the [skipped branch](https://github.com/hudson-worden/livekit-agents/pull/1/changes/BASE..8066c2e9731f4d4290753dc4a0b6cca9f159742a#diff-6ebfa91a7614432d570af64493a27554b30cc7cf93875abab1ae52aa354397caR590)

Context: 
[Related issue
](https://github.com/livekit/agents/issues/4969)I've been looking at ways to improve [these two attributes](https://github.com/livekit/agents/blob/8343db123e6186dfbbed20879bf2b70055382435/livekit-agents/livekit/agents/llm/chat_context.py#L111-L112) with `MetricsReport` (`started_speaking_at` and `stopped_speaking_at`). I've been auditing discrepancies between recordings and where these values are set by using some custom otel tracing to visualize the `started_speaking_at` and `stopped_speaking_at` of the MetricsReport on the conversation items.

pseudo-code
```
from opentelemetry import trace

...

def _on_conversation_item_added(self, msg_ev: ConversationItemAddedEvent) -> None:
    if msg_ev.item.type == "message":
        tracer = trace.get_tracer(__name__)
        item_copy = msg_ev.item.copy()
        started_at = item_copy.metrics.get("started_speaking_at")
        stopped_at = item_copy.metrics.get("stopped_speaking_at")
        if started_at is not None and stopped_at is not None:
            start_ns = int(started_at * 1_000_000_000)
            end_ns = int(stopped_at * 1_000_000_000)

            span_name = "chat_item_message_assistant" if item_copy.role == "assistant" else "chat_item_message_user"
            if item_copy.interrupted:
                span_name += "_interrupted"
            span = tracer.start_span(
                span_name,
                start_time=start_ns,
            )
            span.set_attribute("item_role", item_copy.role)
            span.set_attribute("item_id", item_copy.id)
            span.set_attribute("item_type", item_copy.type)
            span.set_attribute("item_content", item_copy.content)
            span.set_attribute("item_interrupted", item_copy.interrupted)
            span.end(end_time=end_ns)
```

I'm also visualizing key decision points using custom spans in [livekit](https://github.com/hudson-worden/livekit-agents/pull/1/changes).


hudson-worden:
we're currently working on addressing the test failures

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -522,12 +522,14 @@ async def _on_vad_event(self, ev: vad.VADEvent) -> None:
...
 
     def _run_eou_detection(self, chat_ctx: llm.ChatContext, skip_reply: bool = False) -> None:
-        if self._stt and not self._audio_transcript and self._turn_detection_mode != "manual":
+        if self._stt and not self._audio_transcript and self._turn_detection_mode == "stt":
             # stt enabled but no transcript yet
             return
 
         chat_ctx = chat_ctx.copy()
-        chat_ctx.add_message(role="user", content=self._audio_transcript)
+        if self._audio_transcript:
```
hudson-worden: I believe `chat_ctx` only really ends up in observability / otel traces so that's the impact here

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1453,8 +1453,8 @@ def on_end_of_turn(self, info: _EndOfTurnInfo) -> bool:
                 extra={"user_input": info.new_transcript},
             )
 
-            if self._session._closing:
-                # add user input to chat context
+            if self._session._closing and info.new_transcript:
```
hudson-worden: This is what actually fixed the tests.

livekit-agents/livekit/agents/voice/agent_activity.py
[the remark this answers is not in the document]
hudson-worden: happy to add it, but need some more context from a reviewer if this is necessary

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -522,12 +522,14 @@ async def _on_vad_event(self, ev: vad.VADEvent) -> None:
                 self._run_eou_detection(chat_ctx)
 
     def _run_eou_detection(self, chat_ctx: llm.ChatContext, skip_reply: bool = False) -> None:
-        if self._stt and not self._audio_transcript and self._turn_detection_mode != "manual":
+        if self._stt and not self._audio_transcript and self._turn_detection_mode == "stt":
```
davidzhao: we have to wait for transcript because our turn detector utilizes text input to help determine end of turn.

hudson-worden: Hi @davidzhao, thanks for the reply are you referring to [this](https://docs.livekit.io/agents/logic/turns/turn-detector/)?

hudson-worden: If that's the case, then I think [based on this](https://github.com/livekit/agents/blob/5b23800d54bb79d95de19615fbe299c4dec970f3/livekit-agents/livekit/agents/voice/audio_recognition.py#L121-L124) I can make a change

davidzhao: yes, you can see it [here](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/voice/audio_recognition.py#L545), where it performs a EOU inference depending on recent chat context.

hudson-worden: cool then I think [tightening up the guard should work](https://github.com/livekit/agents/pull/4988/changes/c43f3f30b8cb35e83d2e2b2a08c72a70b6c0b365)!

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -522,12 +522,22 @@ async def _on_vad_event(self, ev: vad.VADEvent) -> None:
...
-        if self._stt and not self._audio_transcript and self._turn_detection_mode != "manual":
+        if (
+            self._stt
+            and not self._audio_transcript
+            and (
+                # if turn detection is based on stt
+                # OR if a turn detector is provided (e.g the MultilingualModel)
+                self._turn_detection_mode == "stt" or self._turn_detector is not None
+            )
+        ):
```
davidzhao: this isn't quite right.. I would recommend writing it like this for readability:

```python
if self._stt and not self.audio_transcript:
    if self._turn_detection_mode == "stt":
        return
    if self._turn_detection_mode == "vad" and self._turn_detector is None:
        return
```

hudson-worden: I updated the if statement to be more readable, though it's different than the way you have it

https://github.com/livekit/agents/pull/4988/commits/a25deeeae250c7b8edc8ea346ac5e3a5cc354573


```
  if self._stt and not self._audio_transcript:
      if self._turn_detection_mode == "stt":
          # stt enabled but no transcript yet
          return
      if self._turn_detector is not None:
          # a turn detector like (MultilingualModel) is provided but no transcript yet
          return
```
in this case you mentioned

```
if self._turn_detection_mode == "vad" and self._turn_detector is None:
    return
```

That is the case that we do want to pass through since no transcript is needed in that case. Unless I'm mistaken?

davidzhao: with manual mode, it should always proceed immediately. so either explicitly checking for `vad`, or doing `!= "manual"` like it was doing before

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -521,13 +524,30 @@ async def _on_vad_event(self, ev: vad.VADEvent) -> None:
...
+            match self._turn_detection_mode:
+                case "stt" | "realtime_llm" | None:
+                    return True
+                case "manual" | "vad":
+                    return False
+                case _:
+                    # If not specified then we assume it requires transcript
+                    return True
+        else:
+            return False
```
hudson-worden, line 541: Ok made this into it's own function to enumerate all of the possible values

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -521,13 +524,30 @@ async def _on_vad_event(self, ev: vad.VADEvent) -> None:
...
 
+    def _eou_requires_transcript(self) -> bool:
+        if self._stt:
+            # while we aren't checking _turn_detector here,
+            #   _turn_detector and _turn_detection_mode are mutually exclusive (such that if one is provided, the other must be None)
+            # e.g. if _turn_detector is provided, _turn_detection_mode is None, and vice versa
+            match self._turn_detection_mode:
+                case "stt" | "realtime_llm" | None:
+                    return True
+                case "manual" | "vad":
```
hudson-worden, line 535: The addition of `vad` here is the fix

livekit-agents/livekit/agents/voice/audio_recognition.py
hudson-worden: Before everything except for manual was essentially going through 
the first case statement.

chenghao-mou:
Maybe I am missing something.

The current code skips EOU for VAD EOS without changing start speaking time, because STT transcript for that VAD speech might arrive later. When it arrives later, we call the same EOU function with the new transcript, at which time, it will be committed with the same start speaking time and reset correctly.

hudson-worden:
[I'm currently testing a few hypotheses after this comment](https://github.com/livekit/agents/pull/4988#issuecomment-4007320506). 

I still think something is up, but it may be b/c of something different than what I'm proposing here. I'm thinking if we went ahead with allowing VAD to pass through and do EOU detection (like this PR), we'd run into situations where STT would have provided a more complete `_audio_transcript` after VAD declared END_OF_SPEECH?

I'm thinking the issue may be related to [this](https://github.com/livekit/agents/issues/4969#issuecomment-3985769231) instead.

hudson-worden:
@chenghao-mou @davidzhao - Thanks for the review. This may have been misguided so I'll go ahead and close this PR.

I've opened a [separate PR with the fix for this particular issue.](https://github.com/livekit/agents/pull/5027)

PR #4993, opened 2026-03-03
Generate and upload markdown docs

Topherhindman:
## Summary
- Add script that converts pdoc3 HTML API reference to markdown for LLM/agent consumption via the docs MCP server 
- Add tests covering path mapping, link rewriting, page conversion, validation, and link integrity
- Add beautifulsoup4 and markdownify to docs dependency group
- Update CI to convert HTML docs to markdown, upload to s3://livekit-docs/python-md, and gate on `--check-links`

### Why not use pdoc3's built-in markdown output?
pdoc3 has two markdown modes, but neither produces clean per-module files suitable for MCP serving:
- pdoc (no flags) — Writes per-module .md files, but uses Markdown-Extra syntax (definition lists with :, setext === headings) instead of CommonMark, sub-module links are plain text instead of relative file links, Pydantic docstring cross-references like [FieldInfo][pydantic.fields.FieldInfo] render as broken reference links, and raw `<code>` tags appear in places instead of backticks.
- pdoc `--pdf` — Full signature fidelity but concatenates everything into a single stdout stream with Pandoc metadata (geometry: margin=1in), anchor-based internal links (#livekit.agents.llm), and no per-module file output.

Either path would still require a transformation script to produce clean CommonMark files with working relative links. Since CI already builds and publishes the HTML docs, converting from the structured HTML DOM turned out to be more reliable than post-processing markdown text — HTML elements like `<dt>`, `<dd>`, `<code>`, and `<a href>` are unambiguous to parse, whereas patching definition-list syntax and setext headings in raw markdown is fragile.

The converter also adds link integrity validation (`--check-links`) that gates CI, catching broken internal links before they reach the MCP server.

### DIFF
--- a/.github/convert_html_docs.py
+++ b/.github/convert_html_docs.py
@@ -13,7 +13,7 @@ Path mapping (canonical, no ambiguity):
     livekit/agents/vad.html   -> livekit/agents/vad.md
 
 Usage:
-    uv run python scripts/convert_html_docs.py INPUT_DIR OUTPUT_DIR [options]
+    uv run python .github/convert_html_docs.py INPUT_DIR OUTPUT_DIR [options]
       --validate-only    Only validate existing markdown against HTML
       --no-validate      Skip validation after conversion
       --verbose / -v     Show per-file validation details
--- a/tests/test_convert_html_docs.py
+++ b/tests/test_convert_html_docs.py
@@ -1,21 +1,30 @@
-"""Tests for scripts/convert_html_docs.py — pdoc3 HTML to markdown converter."""
+"""Tests for .github/convert_html_docs.py — pdoc3 HTML to markdown converter."""
 
 from __future__ import annotations
 
+import importlib
+import sys
 from pathlib import Path
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


bcherry:
is there a preview of the output somewhere?

Topherhindman:
> is there a preview of the output somewhere?

shared via slack

tests/test_convert_html_docs.py
theomonnom, line 1: Are the tests really needed?

.github/convert_html_docs.py
theomonnom, line 1: Can we move that inside .github instead of scripts

Topherhindman: Definitely

tests/test_convert_html_docs.py
Topherhindman: Given its custom logic, I feel like they are good to have, and so we know if there is a change made that the docs logic doesn't handle properly.  That being said, if we aren't too concerned about it, we could leave them out.  I probably lean more towards keeping them than not, but ultimately am willing to defer to you on it

Issue #4989, opened 2026-03-03
Tool mutations ignored when llm_node is overridden as async generator


timothyaufdermauer:
### Bug Description

When llm_node is overridden as an async generator, tool mutations made to the tools list parameter inside the generator body are invisible to the framework's tool execution validation.

_llm_inference_task in generation.py calls tool_ctx.update_tools(tools) immediately after creating the generator object but before the generator body has started executing. This means tool_ctx._fnc_tools_map (used to validate incoming tool calls from the LLM at line 482) is built from the pre-mutation tools list.

The result: the LLM correctly receives updated tool schemas and calls the new tool, but the framework rejects it with "unknown AI function" because tool_ctx.function_tools was never updated.

### Expected Behavior

Tool mutations made to the tools list inside llm_node should be reflected in tool_ctx.function_tools so the framework can execute tools that the LLM was told about. The update_tools call should happen after the generator body has had a chance to run its pre-yield logic, not before.

### Reproduction Steps
Override llm_node as an async generator
Inside the generator body (before yielding), add a new tool to the tools list via tools[:] = new_tools_list
Call super().llm_node() to send the updated tools to the LLM
The LLM sees the new tool, calls it
Framework logs WARNING unknown AI function 'new_tool_name' and silently drops the call
Relevant framework code in generation.py:

```
# line 95:  tools = tool_ctx.flatten()          # creates flat list
# line 116: llm_node = node(chat_ctx, tools, model_settings)  # generator created, body NOT executed
# line 122: tool_ctx.update_tools(tools)        # builds _fnc_tools_map from PRE-MUTATION list
# ...
# line ~134: async for chunk in llm_node:       # NOW generator body runs, tools[:] mutation happens
# ...
# line 482: tool_ctx.function_tools.get(fnc_call.name)  # lookup against STALE map → None
```

### Operating System

Windows 11

### Models Used

Anthropic Claude Sonnet 4.6

### Package Versions

```bash
livekit-agents==1.6.1
```

### Proposed Solution

```python

```

### Additional Context

The framework comment at line 121-122 says: "store any updated tools, to ensure subsequent tool calls in the same turn (nested calls) are using the newer tools" — this confirms the intent was to support tool mutations in llm_node, but the async generator timing defeats it. The update_tools call fires before `async for` drives the generator, so any mutations in the generator's pre-yield body are missed.

PR #4994, opened 2026-03-04
fix: resync tool context when tools are mutated inside llm_node

longcw:
fix https://github.com/livekit/agents/issues/4989


davidzhao (approved):
lg

PR #4998, opened 2026-03-04
update error message when TTS is not set

longcw:


longcw:
`session.output.set_audio_enabled(False)` is not room option, it also applied for console mode. I think it should be fine the RoomOptions is ignored when using console mode.

chenghao-mou:
> `session.output.set_audio_enabled(False)` is not room option, it also applied for console mode. I think it should be fine the RoomOptions is ignored when using console mode.

I think I know the cause: I have to put `session.output.set_audio_enabled(False)` before `session. start`, and disabling audio output in room options has no effect on the session output itself.

PR #4999, opened 2026-03-04
fix(google): prevent function_call text from leaking to TTS output

BkSouX:
## Summary                                                                                                                       
  - When Gemini returns a `Part` containing both `function_call` and `text`, the text (typically the function name) was passed as
  `content` in the `ChatChunk`, causing the TTS to speak it aloud                                                                  
  - Set `content=None` when a `function_call` is present in `_parse_part`, matching the behavior of the OpenAI plugin
  - Added unit tests for `_parse_part` covering function_call with/without text, text-only, and empty text cases


tothonx:
> good catch. this makes sense.
> 
> could you rename the test to: `test_plugin_google_llm.py` ?

done

PR #5003, opened 2026-03-04
fix(assemblyai): use cumulative words text for PREFLIGHT_TRANSCRIPT

### LINKED ISSUE #4779: AssemblyAI STT: transcription text jumps/regresses mid-segment
### Bug Description

When using AssemblyAISTT, transcription text on the client occasionally "jumps", the last speech bubble's text shrinks to just a few words, then reappears in full on the next update. STT and voice interaction work correctly; the issue is purely visual.

Example sequence on a single segment ID (all final: false):

1. "I need you to look beyond the eyes and be honest with me" — correct, growing
2. "And be honest with me." — text regresses to just the tail end
3. "I need you to look beyond the eyes and be honest with me, always" — full text restored

### Expected Behavior

Transcription text for a segment should only grow or be replaced by the final transcript. never shrink mid-utterance.

### Reproduction Steps

1. Set up a voice agent with AssemblyAISTT (streaming mode)
2. Connect a frontend client that renders TranscriptionReceived segments by ID
3. Speak continuously for 10+ seconds
4. Observe the last speech bubble's text occasionally shrinking then restoring

### Operating System

Linux (also observed on iOS/Safari client)

### Models Used

AssemblyAI universal-streaming-multilingual

### Package Versions

- livekit-agents[assemblyai] == 1.4.1
- livekit-client (JS) latest

### Session/Room/Call IDs

_No response_

### Proposed Solution

AssemblyAI Turn messages contain both a words array (cumulative) and an utterance field (chunk-based). It appears the plugin emits INTERIM_TRANSCRIPT from cumulative words and PREFLIGHT_TRANSCRIPT from the chunk-based utterance. AudioRecognition routes both through on_interim_transcript, and since the transcription uses replacement mode (is_delta_stream=False), the PREFLIGHT's chunk text overwrites the INTERIM's cumulative text on the same segment ID.

A length guard in the framework (skip updates where text shrinks for the same segment) would fix this generically. We're currently applying this workaround on the client side.

### Additional Context

This is independent of format_turns,  the utterance field is always chunk-based regardless of that setting.

_No response_

weiguangli-io:
## Summary

Fixes #4779

- **Root cause**: The `PREFLIGHT_TRANSCRIPT` event was emitting the chunk-based `utterance` field as its text, while `INTERIM_TRANSCRIPT` used the cumulative `words` array. Both events flow through `on_interim_transcript` in the framework and are rendered in replacement mode (`is_delta_stream=False`), so the shorter chunk text from PREFLIGHT would overwrite the longer cumulative text on the same segment ID, causing visible text regression/jumping on the client.

- **Fix**: Use `interim_text` (built from the cumulative `words` array) for the `PREFLIGHT_TRANSCRIPT` text field, matching the behavior of `INTERIM_TRANSCRIPT`. Also added a guard to require `timed_words` to be non-empty before entering the utterance block, since `interim_text` is only defined within the `timed_words` scope.

## Test plan

- [ ] Verify `ruff check` and `ruff format` pass (confirmed locally)
- [ ] Manual test: set up a voice agent with `AssemblyAISTT` in streaming mode, speak continuously for 10+ seconds, and confirm the transcription text on the client only grows or is replaced by the final transcript (never shrinks mid-utterance)
- [ ] Verify existing `test_plugin_assemblyai_stt.py` tests pass

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -540,7 +540,7 @@ class SpeechStream(stt.SpeechStream):
             )
             self._event_ch.send_nowait(interim_event)
 
-        if utterance:
+        if utterance and timed_words:
             if self._last_preflight_start_time == 0.0:
                 self._last_preflight_start_time = start_time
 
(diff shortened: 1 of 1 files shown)


weiguangli-io:
Friendly ping — CI is green. Would appreciate a review when you get a chance. Thanks!

PR #5004, opened 2026-03-04
fix(google): split text exceeding 5000-byte API limit in TTS synthesize

### LINKED ISSUE #4762: Google TTS fails with  “input.text or input.ssml longer than 5000 bytes” even for short utterances (LiveKit Agents)
### Bug Description

When using LiveKit Agents with the Google TTS plugin, speech synthesis fails with a
400 INVALID_ARGUMENT error stating that `input.text` or `input.ssml` exceeds 5000 bytes.

This error occurs even when the assistant output is a very short sentence (well under 5000 bytes). and it was breaking the session.

The issue seems unrelated to actual text length and may be caused by SSML handling,
internal buffering, retry concatenation, or chunking behavior in the Google TTS plugin.

### Expected Behavior

Short assistant responses (a few words or sentences) should synthesize successfully
using Google TTS without hitting the 5000-byte limit.

The TTS plugin should:
- Properly chunk or stream text
- Avoid concatenating previous failed attempts
- Respect Google TTS size limits internally

### Reproduction Steps

```bash
#the tts configuration iam using was:
        tts=google.TTS(
            voice_name="te-IN-Chirp3-HD-Erinome",  # Use Standard voice
            language="te-IN",
            gender="female",
            credentials_file=os.getenv("GOOGLE_CREDENTIALS_PATH"),        
            use_streaming=False,
            enable_ssml=True,
            audio_encoding=texttospeech.AudioEncoding.LINEAR16,  # Must be PCM
            sample_rate=24000,
            speaking_rate=1.0,
            volume_gain_db=3.0,
        ),
```

### Operating System

Windows11 

### Models Used

gemini cloud tts

### Package Versions

```bash
livekit-agents[google]~=1.3
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

weiguangli-io:
## Summary

Fixes #4762

Google Cloud TTS `synthesize_speech` API rejects `input.text` or `input.ssml` longer than 5000 bytes. When using non-streaming mode (`use_streaming=False`), the Google TTS plugin sent the entire input text in a single API request without any byte-length validation. This caused `400 INVALID_ARGUMENT` errors even for moderately sized text, especially with multi-byte scripts (e.g., Telugu where each character encodes to 3 bytes in UTF-8).

### Changes

- Add automatic text chunking in `ChunkedStream._run()` that splits input text into chunks that each fit within the 5000-byte API limit
- Text is split hierarchically: first at sentence boundaries (`.!?` etc.), then at word boundaries, and as a last resort at character boundaries
- SSML wrapper overhead (`<speak></speak>` = 15 bytes) is accounted for when `enable_ssml=True`
- Refactor `_build_ssml()` to a static method and extract `_build_synthesis_input()` for cleaner code
- Audio responses from multiple chunks are pushed sequentially to the output emitter, producing seamless audio output

### How it works

For text within the limit (the common case), behavior is identical to before -- a single API call is made. When text exceeds 5000 bytes, the new `_get_text_chunks()` method splits it into safe-sized pieces, each of which is synthesized independently and the audio is concatenated.

## Test plan

- [x] Verified `_split_text_by_bytes` logic with ASCII text, Telugu (multi-byte) text, and SSML overhead edge cases
- [x] Verified `ruff check` and `ruff format` pass cleanly
- [ ] Integration test with Google Cloud TTS using Telugu language and `enable_ssml=True` configuration from the issue

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
@@ -48,7 +48,6 @@ DEFAULT_LANGUAGE = "en-US"
 DEFAULT_GENDER = "neutral"
 # Google Cloud TTS API limit: input.text or input.ssml must not exceed 5000 bytes
 GOOGLE_TTS_MAX_INPUT_BYTES = 5000
-_word_tokenizer = tokenize.basic.WordTokenizer(ignore_punctuation=False)
 
 
 @dataclass
@@ -305,10 +304,9 @@ class ChunkedStream(tts.ChunkedStream):
 
         Google Cloud TTS API rejects input.text or input.ssml longer than
         5000 bytes. Uses the configured sentence tokenizer (blingfire by
-        default) for boundary detection, with a word tokenizer fallback for
-        sentences that individually exceed the byte limit. Structured content
-        (SSML, markup) is returned as-is because naive sentence splitting
-        would break tag structure.
+        default) for boundary detection. Structured content (SSML, markup)
+        is returned as-is because naive sentence splitting would break tag
+        structure.
         """
         if self._opts.enable_ssml or self._opts.use_markup:
             return [self._input_text]
@@ -509,8 +507,9 @@ class SynthesizeStream(tts.SynthesizeStream):
 def _split_sentences_by_bytes(sentences: list[str], max_bytes: int) -> list[str]:
     """Merge tokenized sentences into chunks that fit within a byte limit.
 
-    When a single sentence exceeds *max_bytes*, it is further split at word
-    boundaries using the module-level ``_word_tokenizer``.
+    If a single sentence exceeds *max_bytes* it is included as-is and a warning
+    is logged.  Splitting mid-sentence would generally produce incorrect
+    intonation from the TTS engine.
     """
     chunks: list[str] = []
     current = ""
@@ -525,19 +524,13 @@ def _split_sentences_by_bytes(sentences: list[str], max_bytes: int) -> list[str]
             if current:
                 chunks.append(current)
             if len(sentence.encode("utf-8")) > max_bytes:
-                # split oversized sentence at word boundaries
-                words = _word_tokenizer.tokenize(text=sentence)
-                current = ""
-                for word in words:
-                    word_candidate = (current + " " + word) if current else word
-                    if len(word_candidate.encode("utf-8")) <= max_bytes:
-                        current = word_candidate
-                    else:
-                        if current:
-                            chunks.append(current)
-                        current = word
-            else:
-                current = sentence
+                logger.warning(
+                    "single sentence exceeds Google TTS byte limit (%d bytes), "
+                    "synthesis may fail: %.100s...",
+                    len(sentence.encode("utf-8")),
+                    sentence,
+                )
+            current = sentence
 
     if current:
         chunks.append(current)
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
@@ -48,6 +48,7 @@ DEFAULT_LANGUAGE = "en-US"
 DEFAULT_GENDER = "neutral"
 # Google Cloud TTS API limit: input.text or input.ssml must not exceed 5000 bytes
 GOOGLE_TTS_MAX_INPUT_BYTES = 5000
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -293,42 +297,53 @@ def __init__(self, *, tts: TTS, input_text: str, conn_options: APIConnectOptions
...
+        Google Cloud TTS API rejects input.text or input.ssml longer than
+        5000 bytes. This method splits the input accordingly, accounting for
+        SSML wrapper overhead when SSML is enabled.
+        """
+        if self._opts.enable_ssml:
+            max_bytes = GOOGLE_TTS_MAX_INPUT_BYTES - SSML_WRAPPER_OVERHEAD
+        else:
+            max_bytes = GOOGLE_TTS_MAX_INPUT_BYTES
+
+        return _split_text_by_bytes(self._input_text, max_bytes)
```
davidzhao: a tokenizer should be used instead of hand crafting one. see: https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py#L399 as an example

we already bundle blingfire for tokenization

davidzhao (changes requested):
nice catch, let's update the tokenization strategy and we can merge it

livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -477,3 +502,86 @@ def _encoding_to_mimetype(encoding: texttospeech.AudioEncoding) -> str:
...
+    sentences = re.split(r"(?<=[.!?。！？])\s*", text)
+    chunks: list[str] = []
+    current = ""
+
+    for sentence in sentences:
+        if not sentence:
+            continue
+        candidate = current + sentence if current else sentence
+        if len(candidate.encode("utf-8")) <= max_bytes:
+            current = candidate
```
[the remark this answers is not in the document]
weiguangli-io: Updated — the previous commit already switched to the blingfire `SentenceTokenizer` as you suggested. This new commit also fixes a secondary issue: SSML and markup content is now returned as-is (not split) since the sentence tokenizer is unaware of XML tag structure and would produce malformed chunks. Plain text splitting continues to use the blingfire tokenizer with word/char fallbacks.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -293,52 +296,99 @@ def __init__(self, *, tts: TTS, input_text: str, conn_options: APIConnectOptions
...
+    def _build_ssml(text: str) -> str:
+        return f"<speak>{text}</speak>"
+
+    def _get_text_chunks(self) -> list[str]:
+        """Split input text into chunks that respect Google TTS byte limits.
+
+        Google Cloud TTS API rejects input.text or input.ssml longer than
+        5000 bytes. Uses the configured sentence tokenizer (blingfire by
+        default) for boundary detection, with a word tokenizer fallback for
+        sentences that individually exceed the byte limit. Structured content
```
davidzhao: I don't think we need a word tokenizer fallback here. do you see a realistic use case where a sentence is more than 5000 chars?

synthesizing part of a sentence will generally not produce the right intonations or prosody

weiguangli-io: Good point — a single sentence realistically won't exceed 5000 bytes. I'll remove the word tokenizer fallback and keep it simple with just sentence-level splitting. Will push the update shortly.

weiguangli-io:
Word tokenizer fallback removed in commit 846392f. Now only using sentence-level splitting as requested. If a single sentence exceeds 5000 bytes, it's sent as-is with a warning logged.

weiguangli-io:
Hi @davidzhao, I've updated the tokenization strategy per your feedback. The word tokenizer fallback has been removed, and now we only use sentence-level splitting via blingfire. If a single sentence exceeds 5000 bytes, we log a warning and send it as-is (to preserve intonation and prosody). Could you please review the changes?

weiguangli-io:
@davidzhao Hi! Just a friendly follow-up — I've addressed all the feedback from your review (removed the word tokenizer, simplified the splitting logic). Would you have a chance to take another look? Thanks!

PR #5005, opened 2026-03-04
fix: strip SSML tags from conversation messages

### LINKED ISSUE #4802: "<break" tag leaks into conversation message when used in .say
### Bug Description

When I add `<break>` tag to the text argument from .say.
The agent says the message without mentioning the `<break>` tag.
But the conversation message contains the `<break>` tag.

### Expected Behavior

When agent uses `session.say` with a text that includes `<break>`, the `<break>` should not be present in the conversation message.

### Reproduction Steps

```bash
Steps to reproduce the behavior:
1. one client is connected to the room (in addition to agent)
2. clients and agent exchange messages
3. when agent uses `session.say` with a text that includes `<break ...`, the `<break ...` tag leaks into the message
```

### Operating System

MacOS, Linux

### Models Used

Azure-Speech, Cartesia, OpenAI

### Package Versions

```bash
**Server**
- Version: [0.x.x] (unsure what this is)
- Environment: EKS
- Dependencies: azure-speech, cartesia, gpt-4.1

**Client**
- SDK: python
- Version:
  - livekit = "1.0.20"
  - livekit-agents = {version = "1.3.6", extras = ["turn-detector", "silero", "cartesia", "azure"]}
  - livekit-api = "1.1.0"
  - livekit-plugins-noise-cancellation = "0.2.5"
  - livekit-plugins-openai = "1.3.6"
  - livekit-protocol = "1.1.1"
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

weiguangli-io:
## Summary

Fixes #4802

When using `session.say()` with SSML-tagged text (e.g. `<break time="1s"/>`), the TTS engine correctly processes the tags — it pauses as expected and does not read them aloud. However, the raw SSML tags leak into the conversation messages displayed to the user, which is not the expected UX.

This PR strips SSML tags before they reach the conversation transcript:

- **Default `transcription_node`** (`agent.py`): strips SSML tags from plain-string chunks while preserving `TimedString` pass-through (TTS-aligned transcription already excludes SSML). Buffers incomplete tags that may span streaming chunks.
- **`strip_ssml()` helper** (`filters.py`): synchronous utility for stripping SSML tags from a string.
- **`filter_ssml` async transform** (`filters.py`): streaming-aware SSML filter registered as `"filter_ssml"` in `TextTransforms`, following the same pattern as `filter_markdown` and `filter_emoji`.

## Test plan

- [x] Added `test_ssml_filter` parameterized across 8 chunk sizes (1, 2, 3, 5, 7, 11, 50, 200) to verify correct stripping regardless of how tags are split across stream boundaries
- [x] Added `test_ssml_filter_no_tags` — plain text passes through unchanged
- [x] Added `test_ssml_filter_only_tags` — all-SSML input produces empty output
- [x] Added `test_ssml_filter_self_closing_variants` — handles `<break time="1s"/>` and `<break time="500ms" />` formats
- [x] Added `test_strip_ssml` — unit tests for the synchronous helper

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent.py
+++ b/livekit-agents/livekit/agents/voice/agent.py
@@ -464,9 +464,40 @@ class Agent:
         async def transcription_node(
             agent: Agent, text: AsyncIterable[str | TimedString], model_settings: ModelSettings
         ) -> AsyncGenerator[str | TimedString, None]:
-            """Default implementation for `Agent.transcription_node`"""
+            """Default implementation for `Agent.transcription_node`
+
+            Strips SSML tags (e.g. ``<break time="1s"/>``) so they do not leak
+            into conversation messages shown to the user.
+            """
+            from .io import TimedString as _TimedString
+            from .transcription.filters import strip_ssml
+
+            buffer = ""
+
             async for delta in text:
-                yield delta
+                if isinstance(delta, _TimedString):
+                    # TimedString comes from TTS-aligned transcription which
+                    # has already been processed by the TTS engine, so SSML
+                    # tags should not be present.  Yield as-is to preserve
+                    # timing metadata.
+                    yield delta
+                    continue
+
+                # Buffer plain strings to handle SSML tags that may span chunks
+                buffer += delta
+                last_open = buffer.rfind("<")
+                if last_open != -1 and ">" not in buffer[last_open:]:
+                    # Potentially incomplete tag - yield text before it
+                    complete = buffer[:last_open]
+                    if complete:
+                        yield strip_ssml(complete)
+                    buffer = buffer[last_open:]
+                else:
+                    yield strip_ssml(buffer)
+                    buffer = ""
+
+            if buffer:
+                yield strip_ssml(buffer)
 
         @staticmethod
         async def realtime_audio_output_node(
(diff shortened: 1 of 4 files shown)


weiguangli-io:
Friendly ping — CI is green. Would appreciate a review when you get a chance. Thanks!

livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -464,9 +464,40 @@ async def _forward_input() -> None:
...
+                buffer += delta
+                last_open = buffer.rfind("<")
+                if last_open != -1 and ">" not in buffer[last_open:]:
+                    # Potentially incomplete tag - yield text before it
+                    complete = buffer[:last_open]
+                    if complete:
+                        yield strip_ssml(complete)
+                    buffer = buffer[last_open:]
+                else:
+                    yield strip_ssml(buffer)
```
longcw, line 496: modify the `transcription_node` is not optimal since users may overwrite it with a customized node. I think it's good to have a built-in `filter_ssml`, and ppl want to enable it can customize the `transcription_node`, or like `tts_text_transforms`, we can add a `transcription_transforms` that applies the transforms before the node.

marioplumbeer: ping @weiguangli-io ❤️

PR #5006, opened 2026-03-04
fix(elevenlabs): restore chunk_length_schedule in WS init payload

IanSteno:
## Summary
- restore `generation_config.chunk_length_schedule` in ElevenLabs multi-context websocket init packets
- keep existing default `auto_mode=True` for normal usage, but default to `auto_mode=False` when `chunk_length_schedule` is explicitly provided so the schedule is not implicitly disabled
- refactor init packet construction into a helper for deterministic unit coverage

## Why
`chunk_length_schedule` has been accepted/stored by the plugin but not forwarded in current multi-context websocket flow, making it a no-op in practice.

## Changes
- `livekit/plugins/elevenlabs/tts.py`
  - add `_build_context_init_packet(...)` to include:
    - `text`
    - `voice_settings`
    - `context_id`
    - `generation_config.chunk_length_schedule` when provided
    - `pronunciation_dictionary_locators` when provided
  - use this helper in `_Connection._send_loop()` init path
  - set constructor defaulting to `auto_mode = not is_given(chunk_length_schedule)` when `auto_mode` is not explicitly provided

## Tests
Added `tests/test_plugin_elevenlabs_tts.py` covering:
- auto_mode defaults with and without explicit chunk schedule
- explicit `auto_mode` override precedence
- init packet includes `generation_config` when schedule is set
- init packet omits `generation_config` when schedule is unset
- pronunciation dictionary locators are preserved in init packet

Run locally:
- `uv run pytest tests/test_plugin_elevenlabs_tts.py -q`
- `uv run ruff check livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py tests/test_plugin_elevenlabs_tts.py`

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -501,7 +501,9 @@ class _TTSOptions:
 
 def _build_context_init_packet(opts: _TTSOptions, *, context_id: str) -> dict[str, Any]:
     voice_settings = (
-        _strip_nones(dataclasses.asdict(opts.voice_settings)) if is_given(opts.voice_settings) else {}
+        _strip_nones(dataclasses.asdict(opts.voice_settings))
+        if is_given(opts.voice_settings)
+        else {}
     )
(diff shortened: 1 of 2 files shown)


tinalenguyen (approved):
tested it and it works, thank you!

PR #5016, opened 2026-03-05
add trunk_config to WarmTransferTask for SIP endpoint transfers

longcw:


livekit-agents/livekit/agents/beta/workflows/warm_transfer.py
```diff
@@ -69,7 +69,8 @@ def __init__(
         target_phone_number: str,
         *,
         hold_audio: NotGivenOr[AudioSource | AudioConfig | list[AudioConfig] | None] = NOT_GIVEN,
-        sip_trunk_id: NotGivenOr[str] = NOT_GIVEN,
+        trunk_config: NotGivenOr[api.SIPOutboundConfig] = NOT_GIVEN,
```
davidzhao: I'm not the biggest fan of the `trunk` terminology.. because a user might just want to dial out to another sip address. we were talking about changing this in the CreateSIPParticipant API as well.

wdyt about simplifying it like the following:
- rename sip_number to sip_user (it's the same thing, user is the more technically correct term)
- map to deprecated sip_number field
- allow user to pass in: sip_trunk_id or sip_connection: SIPConnection

SIPConnection would contain:
- address
- username
- password
- destination_country: str

davidzhao: another idea.. thinking out loud.. this could be made even more generic.. where we could work with our Whatsapp and Twilio connectors too.. if we refactored out the dial out portion into a function.. then it could be
 
- `SIPDialAction(sip_user, address, username, password, destination_country)`
- `SIPTrunkDialAction(sip_user, trunk_id)`
- `TwilioConnectorDial(twilio_api_key, to_number, etc)`
- `WhatsappConnectorDial(...)`

longcw: is there a reference for the Whatsapp and Twilio connectors?

longcw: should we merge this pr first if these connectors are not ready yet?

davidzhao: yeah lets not wait for connectors.

livekit-agents/livekit/agents/beta/workflows/warm_transfer.py
```diff
@@ -66,10 +66,11 @@ class WarmTransferResult:
 class WarmTransferTask(AgentTask[WarmTransferResult]):
     def __init__(
         self,
-        target_phone_number: str,
+        sip_call_to: NotGivenOr[str] = NOT_GIVEN,
```
davidzhao, line 69: can we add docstring on usage? we should
- document format of `sip_call_to` (i.e. tel:+..., sip:...)
- how to use sip_connection

Issue #5018, opened 2026-03-05
`commit_user_turn()` is not awaitable, making it hard to ensure STT completion before LLM inference in manual turn detection


nara-ryoya:
### Feature Type

I cannot use LiveKit without it

### Feature Description

When using `turn_detection="manual"`, `AgentSession.commit_user_turn()` is a synchronous fire-and-forget method. Because it cannot be awaited, there is no reliable way to ensure that STT has fully completed before the pipeline transitions to LLM inference.

## Current behavior

`commit_user_turn()` internally spawns an async task and returns immediately to the caller:

```python
# audio_recognition.py
def commit_user_turn(self, ...) -> None:
    async def _commit_user_turn() -> None:
        # Step 1: wait for final transcript (up to transcript_timeout seconds)
        await asyncio.wait_for(
            self._final_transcript_received.wait(),
            timeout=transcript_timeout,
        )
        # Step 2: run EOU detection → eventually reaches llm_node
        self._run_eou_detection(chat_ctx)

    self._commit_user_turn_atask = asyncio.create_task(_commit_user_turn())
    # returns immediately ↑
```

The `transcript_timeout` parameter controls how long Task 1 waits *internally* for the final STT transcript, but since the method is not awaitable, the caller has no way to block until that wait completes. As a result, any logic the caller places after `commit_user_turn()` runs concurrently with — not after — the STT completion and LLM transition.

## Why this matters for manual turn detection

With `turn_detection="manual"`, the developer has explicit control over when the user turn ends. A common pattern is:

1. Stop audio input
2. **Wait for STT to finish transcribing**
3. Transition to LLM inference (e.g., send a "thinking" state update to the client)

Step 3 logically depends on step 2, but because `commit_user_turn()` is fire-and-forget, there is no supported way to sequence these steps. The caller cannot know when the STT final transcript has been received and the pipeline has actually handed off to the LLM.

## Expected behavior

`commit_user_turn()` should return an awaitable so that callers can wait for the STT-to-LLM transition to complete:

```python
# Proposed
await session.commit_user_turn(transcript_timeout=30.0)
# At this point, STT is done and LLM inference has started
do_something_after_llm_starts()
```

### Workarounds / Alternatives

Until this is addressed, two workarounds might be available (we do not verified they work well):

### Option 1: Await the internal task directly

You can access the internal `_commit_user_turn_atask` directly and await it to block until the STT flush and transcript wait have completed:

```python
[...]

PR #5019, opened 2026-03-05
make commit_user_turn() return a Future with the audio transcript

longcw:
fix https://github.com/livekit/agents/issues/5018

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1029,7 +1029,9 @@ class AgentActivity(RecognitionHooks):
         if self._rt_session is not None:
             self._rt_session.clear_audio()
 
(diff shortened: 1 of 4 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1029,7 +1029,9 @@ def clear_user_turn(self) -> None:
         if self._rt_session is not None:
             self._rt_session.clear_audio()
 
-    def commit_user_turn(self, *, transcript_timeout: float, stt_flush_duration: float) -> None:
+    def commit_user_turn(
+        self, *, transcript_timeout: float, stt_flush_duration: float
+    ) -> asyncio.Future[str]:
```
theomonnom, line 1034: Should we create a CommitUserTurnResult dataclass?

PR #5024, opened 2026-03-06
add execution status to FunctionCall and example for history compaction with openai realtime

longcw:
fix https://github.com/livekit/agents/pull/5021

### DIFF
--- a/examples/voice_agents/realtime_compact_history.py
+++ b/examples/voice_agents/realtime_compact_history.py
@@ -1,0 +1,94 @@
+import logging
+
+from dotenv import load_dotenv
+
+from livekit.agents import AgentServer, JobContext, cli, llm
+from livekit.agents.voice import Agent, AgentSession
+from livekit.plugins import openai
+
+logger = logging.getLogger("realtime-compact-history")
+logger.setLevel(logging.INFO)
+
+load_dotenv()
+
+
+class MyAgent(Agent):
+    def __init__(self) -> None:
+        super().__init__(
+            instructions="You are a helpful assistant.",
+            llm=openai.realtime.RealtimeModel(),
+        )
+        self.summarizer_llm = openai.LLM(model="gpt-4.1")
+
+    @llm.function_tool
+    async def compact_history(self) -> str:
+        """Called when the user asks to compact the chat history.
+        This will compact the chat history into a short, faithful summary.
+        """
+
+        logger.info("compacting chat history")
+
+        # summarize the chat history using the LLM
+        source_text = "\n".join(
+            f"{item.role}: {(item.text_content or '').strip()}"
+            for item in self.chat_ctx.items
+            if item.type == "message" and item.role in {"user", "assistant"}
+        ).strip()
+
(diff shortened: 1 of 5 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model_beta.py
```diff
@@ -1416,14 +1416,24 @@ def _handle_response_output_item_done(self, event: ResponseOutputItemDoneEvent)
...
+            ):
+                fnc_call = remote_item.item
+            else:
+                # fallback to create a new function call
+                logger.warning(
+                    "function call not found in remote chat ctx, creating a new one",
+                    extra={"item_id": item.id},
+                )
+                fnc_call = llm.FunctionCall(
+                    id=item.id,
```
chenghao-mou, line 1430: nit: use `item_id` for consistency

examples/voice_agents/realtime_compact_history.py
```diff
@@ -0,0 +1,94 @@
...
+            role="assistant",
+            content=f"[history summary]\n{summary}",
+            created_at=last_updated_at,
+        )
+
+        # keep new items and incomplete function calls for realtime session
+        tail_items: list[llm.ChatItem] = []
+        for item in self.realtime_llm_session.chat_ctx.items:
+            if item.created_at > last_updated_at or (
+                item.type == "function_call" and item.status != "completed"
```
longcw, line 72: it's not a good ux to ask user to care about this, but I am not sure if it's a good idea to add the guard of inflight tool call into the `update_chat_ctx`. maybe create a new api for `truncate` operation for realtime session?

examples/voice_agents/realtime_compact_history.py
```diff
@@ -0,0 +1,94 @@
...
+            ):
+                tail_items.append(item)
+        if tail_items:
+            compacted_history.items.extend(tail_items)
+            logger.info(f"keeping {len(tail_items)} items from realtime session")
+
+        await self.update_chat_ctx(compacted_history)
```
chenghao-mou, line 79: It seems that the original system message is missing. I tried this example, and the agent started speaking Spanish after compaction.

chenghao-mou (approved):
lgtm. Two nits.

examples/voice_agents/realtime_compact_history.py
StianHanssen: To me, the biggest concern is that if you are unlucky with timing and another chat context update happens at the same time as a new function call arrives, you can see get the problem.

If we had a different method for sending precise context updates using the `_update_chat_ctx_lock`, that could be a solution to the problem as well. So long as the example also provides a way to update local contest without `update_chat_ctx`.

examples/voice_agents/realtime_compact_history.py
longcw: probably it's bc the instructions didn't specify the response language.

PR #5027, opened 2026-03-06
Set _speech_start_time when VAD START_OF_SPEECH activates

hudson-worden:
This image attempts to convey the issue. I have plotted where the ChatMessage was (determined by [MetricsReport](https://github.com/livekit/agents/blob/89f94d36db6ec289729b1280bd904b0981cba498/livekit-agents/livekit/agents/llm/chat_context.py#L111-L112) and the box that shows it is designated across the whole span) vs where I believe it should be (on the right).

Screenshot 2026-03-06 at 10 01 29 AM

Basically it should align with this custom span I added that represents the vad start of speech.

image

I believe the reason this is happening is b/c we're setting `_speech_start_time` during [INFERENCE_DONE](https://github.com/livekit/agents/blob/89f94d36db6ec289729b1280bd904b0981cba498/livekit-agents/livekit/agents/voice/audio_recognition.py#L523).  When VAD picks up speech ([even though it doesn't match up with the activation](https://github.com/livekit/agents/blob/89f94d36db6ec289729b1280bd904b0981cba498/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/vad.py#L472-L474)) `_speech_start_time` is set. Instead it should be reset when a START_OF_SPEECH event occurs. This would align it with the activation threshold and also a few other places

[here](https://github.com/livekit/agents/blob/89f94d36db6ec289729b1280bd904b0981cba498/livekit-agents/livekit/agents/voice/agent_activity.py#L1292)
and [here](https://github.com/livekit/agents/blob/89f94d36db6ec289729b1280bd904b0981cba498/livekit-agents/livekit/agents/voice/audio_recognition.py#L510).

We're not removing the INFERENCE_DONE branch b/c it would handle the case where no other START_OF_SPEECH event comes. I'm open to removing that too though.

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -535,6 +535,7 @@ class AudioRecognition:
             with trace.use_span(self._ensure_user_turn_span()):
                 self._hooks.on_end_of_speech(ev)
 
+            self._vad_speech_started = False
             self._speaking = False
 
             if self._vad_base_turn_detection or (


hudson-worden:
This is the related [issue thread on this](https://github.com/livekit/agents/issues/4969#issuecomment-3986571535)

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -489,9 +489,9 @@ async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
     @utils.log_exceptions(logger=logger)
     async def _on_vad_event(self, ev: vad.VADEvent) -> None:
         if ev.type == vad.VADEventType.START_OF_SPEECH:
-            with trace.use_span(
-                self._ensure_user_turn_span(start_time=time.time() - ev.speech_duration)
-            ):
+            speech_start_time = time.time() - ev.speech_duration
+            self._speech_start_time = speech_start_time
```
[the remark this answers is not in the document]
hudson-worden: hm this is a good point

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -148,6 +148,8 @@ def __init__(
         self._user_turn_span: trace.Span | None = None
         self._closing = asyncio.Event()
 
+        self._vad_speech_was_started: bool = False
+
```
chenghao-mou, line 152: nit: we can drop `was`

hudson-worden: @chenghao-mou done!

chenghao-mou:
This is another issue related to this: https://github.com/livekit/agents/pull/5083. Will wait for that to be merged first.

livekit-agents/livekit/agents/voice/audio_recognition.py
chenghao-mou: @hudson-worden I just merged #5083 where the timestamps are only reset if there is no new speech after the EOU detection. Could you put a `_vad_speech_started` flag reset here as well with them (you might need to merge main first)?

hudson-worden: let me take a look

hudson-worden: @chenghao-mou [let me know if this is what you're thinking](https://github.com/livekit/agents/pull/5027/changes/e8e723f378145b0e0376cdbc3f3fb37f07e9a6f9). If it is I want to confirm our code still works with that.

hudson-worden: wasn't sure if you wanted me to remove the other one or not

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -535,7 +535,6 @@ async def _on_vad_event(self, ev: vad.VADEvent) -> None:
             with trace.use_span(self._ensure_user_turn_span()):
                 self._hooks.on_end_of_speech(ev)
 
-            self._vad_speech_started = False
             self._speaking = False
```
chenghao-mou, line 539: We can keep this one.

hudson-worden: @chenghao-mou should be updated

PR #5028, opened 2026-03-06
Strip parallel_tool_calls from extra_kwargs when no tools provided

ChadyMoukel:
## Summary

When no tools are provided to the LLM stream, `parallel_tool_calls` is now stripped from `extra_kwargs` alongside the existing `tool_choice` removal. Sending `parallel_tool_calls` without any tools can cause API errors, so this ensures both tool-related parameters are cleaned up consistently.


ChadyMoukel:
Up ?

PR #5031, opened 2026-03-06
healthcare example

tinalenguyen:


livekit-agents/livekit/agents/beta/workflows/address.py
chenghao-mou, line 99: We can use `textwrap.dedent` to make these instructions prettier:

```python
       confirm_address_instruction = (
            "Call `confirm_address` after the user confirmed the address is correct."
            if require_confirmation is not False
            else ""
        )

        super().__init__(
            instructions=dedent(
                f"""
                You are only a single step in a broader system, responsible solely for capturing an address.
                You will be handling addresses from any country. Expect that users will say address in different formats with fields filled like:
                - 'street_address': '450 SOUTH MAIN ST', 'unit_number': 'FLOOR 2', 'locality': 'SALT LAKE CITY UT 84101', 'country': 'UNITED STATES',
                - 'street_address': '123 MAPLE STREET', 'unit_number': 'APARTMENT 10', 'locality': 'OTTAWA ON K1A 0B1', 'country': 'CANADA',
                - 'street_address': 'GUOMAO JIE 3 HAO, CHAOYANG QU', 'unit_number': 'GUOMAO DA SHA 18 LOU 101 SHI', 'locality': 'BEIJING SHI 100000', 'country': 'CHINA',
                - 'street_address': "5 RUE DE L'ANCIENNE COMÉDIE", 'unit_number': 'APP C4', 'locality': '75006 PARIS', 'country': 'FRANCE',
                - 'street_address': 'PLOT 10, NEHRU ROAD', 'unit_number': 'OFFICE 403, 4TH FLOOR', 'locality': 'VILE PARLE (E), MUMBAI MAHARASHTRA 400099', 'country': 'INDIA',
                Normalize common spoken patterns silently:
                - Convert words like 'dash' and 'apostrophe' into symbols: `-`, `'`.
                - Convert spelled out numbers like 'six' and 'seven' into numerals: `6`, `7`.
                - Recognize patterns where users speak their address field followed by spelling: e.g., 'guomao g u o m a o'.
                - Filter out filler words or hesitations.
                - Recognize when there may be accents on certain letters if explicitly said or common in the location specified. Be sure to verify the correct accents if existent.
                Don't mention corrections. Treat inputs as possibly imperfect but fix them silently.
                Call `update_address` at the first opportunity whenever you form a new hypothesis about the address. (before asking any questions or providing any answers.)
                Don't invent new addresses, stick strictly to what the user said.
                {confirm_address_instruction}
                When reading a numerical ordinal suffix (st, nd, rd, th), the number must be verbally expanded into its full, correctly pronounced word form.
                Do not read the number and the suffix letters separately.
                Confirm postal codes by reading them out digit-by-digit as a sequence of single numbers. Do not read them as cardinal numbers.
                For example, read 90210 as 'nine zero two one zero.'
                Avoid using bullet points and parenthese in any responses.
                Spell out the address letter-by-letter when applicable, such as street names and provinces, especially when the user spells it out initially.
                If the address is unclear or invalid, or it takes too much back-and-forth, prompt for it in parts in this order: street address, unit number if applicable, locality, and country.
                Ignore unrelated input and avoid going off-topic. Do not generate markdown, greetings, or unnecessary commentary.
                Always explicitly invoke a tool when applicable. Do not simulate tool usage, no real action is taken unless the tool is explicitly called.{extra_instructions}
                """
            ).strip(),
```

livekit-agents/livekit/agents/beta/workflows/name.py
```diff
@@ -0,0 +1,232 @@
...
+
+    async def on_enter(self) -> None:
+        self.session.generate_reply(
+            instructions=f"Ask the user for their name, follow this order '{self._name_format}' but do not mention the format."
+        )
+
+    @function_tool()
+    async def update_name(
+        self,
+        first_name: str,
```
chenghao-mou: This makes the frist name required, different from the init function.

livekit-agents/livekit/agents/beta/workflows/credit_card.py
```diff
@@ -0,0 +1,480 @@
...
+    async def on_enter(self) -> None:
+        await self.session.generate_reply(
+            instructions="Ask for the user's credit card number.",
+        )
+
+    @function_tool()
+    async def record_card_number(
+        self,
+        context: RunContext,
+        card_number: int,
```
chenghao-mou: I still think we should use str for better semantics. You can easily extract the useful part from the given str: 

```python
>>> card_number = "ab1234 fsfs3456"
>>> ''.join([d for d in card_number if d.isdigit()])
'12343456'
```

examples/healthcare/healthcare_agent.py
```diff
@@ -0,0 +1,735 @@
...
+                # in the case that the user creates a new profile or restarts, the recorded session profile is cleared
+                self.session.userdata.profile = {}
+
+        if event.task_id == "get_dob_task":
+            existing_record = self._database.get_patient_by_name_and_dob(
+                self._pending_name, event.result.date_of_birth
+            )
+            if existing_record:
+                logger.info(
+                    f"Found existing patient profile for {self._pending_name}, with the details: {existing_record}"
```
chenghao-mou: Let's avoid PIIs in logs, even for examples.

livekit-agents/livekit/agents/beta/workflows/credit_card.py
```diff
@@ -0,0 +1,480 @@
...
+                return None
+
+            confirm_tool = self._build_confirm_tool(security_code=stripped)
+            current_tools = [t for t in self.tools if t.id != "confirm_security_code"]
+            current_tools.append(confirm_tool)
+            await self.update_tools(current_tools)
+
+            return (
+                "The security code has been updated.\n"
+                "Do not repeat the security code back to the user, ask them to repeat themselves.\n"
```
chenghao-mou, line 349: This is no confirmation prompt for this one.

livekit-agents/livekit/agents/beta/workflows/credit_card.py
```diff
@@ -0,0 +1,480 @@
...
+            confirm_tool = self._build_confirm_tool(
+                expiration_month=expiration_month, expiration_year=expiration_year
+            )
+            current_tools = [t for t in self.tools if t.id != "confirm_expiration_date"]
+            current_tools.append(confirm_tool)
+            await self.update_tools(current_tools)
+
+            return (
+                "The expiration date has been updated.\n"
+                "Do not repeat the expiration date back to the user, ask them to repeat themselves.\n"
```
chenghao-mou, line 457: No confirmation prompt here either.

theomonnom:
Can we use the new Instructions class?

livekit-agents/livekit/agents/beta/workflows/credit_card.py
tinalenguyen: By confirmation prompt, do you mean a line like` "Call 'confirm_security_code' once the user confirms, do not call it preemptively"`?

livekit-agents/livekit/agents/beta/workflows/credit_card.py
tinalenguyen: Hmm do you think it would provide a better user experience? Open to changing it to str, just curious about your thoughts

livekit-agents/livekit/agents/beta/workflows/credit_card.py
chenghao-mou: Yep. It didn't call the confirm function even though I told it several times.

livekit-agents/livekit/agents/beta/workflows/credit_card.py
chenghao-mou: One common pattern is the first digit, used for [Major industry identifier](https://en.wikipedia.org/wiki/ISO/IEC_7812#Major_industry_identifier), which could be zero in some cases:

CleanShot 2026-03-10 at 11 47 15@2x

Even from the LLM perspective, they should be more accustomed to `str` for this type of data from their training. `int` is "surprising" for both humans and LLMs in that regard. 

This also makes porting it to other languages safer (e.g. js might not be able to use MAX_SAFE_INTEGER to represent the card number)

tinalenguyen: Ahh okay, that makes sense. Thank you for the explanation!

livekit-agents/livekit/agents/beta/workflows/address.py
tinalenguyen: swapped this block of text for `Instructions()`

theomonnom:
~Can we add the pdf to Git LFS?~ I guess it's tiny here

PR #5034, opened 2026-03-07
fix(deepgram): include word confidence for stt v2 alternatives

inickt:
I noticed Deepgram STT v2 word confidence wasn't set on the word level, even though it is used to set the overall combined confidence.

Tested running locally with a STT node.


longcw (approved):
lgtm!

davidzhao (approved):

Issue #4869, opened 2026-02-17
[Bug]: Significant Memory Regression (~1GB jump) in v1.3.12 and Massive Spike with Turn Detection


OmarSalah26:
### Bug Description
After upgrading our LiveKit Python stack from `v1.2.8` to `v1.3.12`, we observed a significant increase in the memory footprint of our agent processes in our production OpenShift environment. 

Even without enabling new features, the baseline memory usage has increased by **166%**. Furthermore, enabling the `turn-detector` plugin causes the memory to spike to **3500MB** before a session even begins, which is significantly higher than the documented requirements.

### **Environment**
*   **Orchestration:** OpenShift (Production)
*   **Python Version:** 3.11

### **Version Comparison**

| Package | Old Version | New Version |
| :--- | :--- | :--- |
| `livekit` | 1.0.12 | 1.0.23 |
| `livekit-agents` | 1.2.8 | 1.3.12 |
| `livekit-api` | 1.0.5 | 1.1.0 |
| `livekit-plugins-google` | 1.2.8 | 1.4.1 |
| `livekit-plugins-silero` | 1.2.8 | 1.3.12 |
| `livekit-plugins-turn-detector` | 1.2.8 (Non-functional) | 1.3.12 |

### **Memory Usage Breakdown**

*   **Old Version (v1.2.8):** ~600 MB (Stable baseline)
*   **New Version (v1.3.12) - Turn Detection DISABLED:** ~1600 MB 
*   **New Version (v1.3.12) - Turn Detection ENABLED:** ~3500 MB (Measured before session start)

### **MY Questions**
*   Are there new default background processes (such as a shared inference server or OTEL tracing) introduced between 1.2.x and 1.3.x that account for the 1GB baseline increase?
*   Why does the Turn Detector plugin consume ~1.9GB when documented as <500MB? 
*   Is there a recommended configuration for OpenShift/K8s to limit this pre-allocation or pre-warming memory?

### **Observed Behavior**
1.  **Baseline Increase:** Simply upgrading the core packages results in a 1GB increase in memory usage. In the previous version (1.2.8), the memory footprint was manageable at 600MB.
2.  **Turn Detector Overhead:** The `livekit-plugins-turn-detector` documentation states the model requires `<500 MB` of RAM. However, enabling it in v1.3.12 adds nearly **1.9 GB** of additional overhead (jumping from 1600MB to 3500MB). 
[...]

PR #5037, opened 2026-03-08
fix(turn-detector): reduce ONNX memory via single-thread sequential execution

giulio-leone:
## Summary

Addresses part of #4869 — reduces turn-detector plugin memory footprint by switching ONNX Runtime session to single-thread sequential execution, matching the silero VAD plugin's proven memory-efficient configuration.

## Root Cause Analysis

The turn-detector ONNX session uses dynamic multi-threaded execution (`intra_op_num_threads = ceil(cpu_count/2)`, up to 4), while the silero VAD plugin uses fixed single-thread sequential execution. Each ONNX Runtime thread allocates its own memory buffer (~100-200MB), so 2-4 threads = 200-800MB of unnecessary overhead per loaded model.

### Comparison: Turn-Detector vs Silero VAD

| Setting | Turn-Detector (before) | Silero VAD | Turn-Detector (after) |
|---------|----------------------|------------|----------------------|
| `intra_op_num_threads` | `max(1, min(ceil(cpu_count/2), 4))` | `1` | `1` |
| `inter_op_num_threads` | `1` | `1` | `1` |
| `execution_mode` | (default: parallel) | `ORT_SEQUENTIAL` | `ORT_SEQUENTIAL` |
| `allow_spinning` | (default: enabled) | `"0"` (disabled) | `"0"` (disabled) |

Reference: `livekit-plugins-silero/livekit/plugins/silero/onnx_model.py` lines 36-41

## Why Sequential is Better Here

The end-of-utterance ONNX model is **lightweight** — it runs a small transformer on ≤128 tokens. Multi-threaded execution provides negligible latency improvement for such a small model, while consuming 200-800MB of additional memory for thread-local buffers. The silero VAD plugin already validates this approach in production.

## Changes

**`livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/base.py`:**
- Set `intra_op_num_threads = 1` (was: dynamic 2-4)
- Set `execution_mode = ORT_SEQUENTIAL` (was: default parallel)
- Disable thread spinning (`session.intra_op.allow_spinning = "0"`)
- Remove unused `math` and `hw` imports

## Expected Impact

- **Memory reduction**: ~200-400MB per loaded model (English + Multilingual = 400-800MB total savings)
- **Latency impact**: Negligible — the EOU model processes ≤128 tokens, well within single-thread performance limits
- **No behavioral changes**: Same model, same predictions, just lower memory footprint

## Additional Context

The full 1GB baseline regression reported in #4869 likely has additional causes (newer transformers library overhead, larger multilingual model v0.4.1-intl vs v0.3.1-intl, framework `inference` module). This PR addresses the most impactful and surgically fixable component.

### DIFF
--- a/livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/base.py
+++ b/livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/base.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import asyncio
 import json
 import logging
-import math
 import re
 import time
 import unicodedata
@@ -16,7 +15,6 @@ from livekit.agents import LanguageCode, Plugin, llm
 from livekit.agents.inference_runner import _InferenceRunner
 from livekit.agents.ipc.inference_executor import InferenceExecutor
 from livekit.agents.job import get_job_context
-from livekit.agents.utils import hw
 
 from .log import logger
 from .models import HG_MODEL, MODEL_REVISIONS, ONNX_FILENAME, EOUModelType
@@ -122,10 +120,11 @@ class _EUORunnerBase(_InferenceRunner):
                 local_files_only=True,
             )
             sess_options = ort.SessionOptions()
-            sess_options.intra_op_num_threads = max(
-                1, min(math.ceil(hw.get_cpu_monitor().cpu_count()) // 2, 4)
-            )
+            sess_options.add_session_config_entry("session.intra_op.allow_spinning", "0")
+            sess_options.add_session_config_entry("session.inter_op.allow_spinning", "0")
             sess_options.inter_op_num_threads = 1
+            sess_options.intra_op_num_threads = 1
+            sess_options.execution_mode = ort.ExecutionMode.ORT_SEQUENTIAL
             sess_options.add_session_config_entry("session.dynamic_block_base", "4")
             self._session = ort.InferenceSession(
                 local_path_onnx, providers=["CPUExecutionProvider"], sess_options=sess_options


giulio-leone:
I have read the CLA Document and I hereby sign the CLA

theomonnom:
How does it impact latency for concurrent inference requests from different job processes?

giulio-leone:
Friendly ping — CI is green, tests pass, rebased on latest. Ready for review whenever convenient. Happy to address any feedback. 🙏

giulio-leone:
@theomonnom Good question. The change constrains ONNX to single-thread sequential execution (`ORT_SEQUENTIAL` + `intra_op_num_threads=1`), which reduces per-inference memory from ~200MB to ~50MB.

**Latency impact:**
- **Single request**: negligible — the turn-detection model is tiny (~4M params), single-thread inference takes <5ms
- **Concurrent processes**: each process has its own ONNX session, so no cross-process contention. Within a single process, inferences would serialize — but turn detection is a single-stream pipeline so this doesn't apply in practice

Net effect: same latency, ~75% less memory per agent process.

theomonnom:
> Concurrent processes: each process has its own ONNX session, so no cross-process contention. Within a single process, inferences would serialize — but turn detection is a single-stream pipeline so this doesn't apply in practice

This isn't true, they all share an inference process

giulio-leone:
You're right — I should have been more precise. Since all jobs share the same inference process, the single-thread constraint does affect concurrent inference requests: they serialize through ONNX's session lock.

However, the latency impact is bounded:
- Each turn-detection inference is very lightweight (~1-5ms for a small ONNX model)
- Single-thread execution eliminates the memory overhead of thread-local state per core
- The memory savings (from not allocating per-thread arenas) benefit the shared process, especially under high job counts

The trade-off is: slightly higher tail latency for concurrent inferences in exchange for significantly lower per-session memory footprint. For turn detection specifically (small model, fast inference), the serialization cost is small relative to the memory saving.

Happy to add a configuration option to let users choose the thread count if they prefer latency over memory for their deployment.

PR #5049, opened 2026-03-09
fix(google): avoid session restart on update_instructions, use mid-session client content

D-zigi:
## Summary

`RealtimeSession.update_instructions()` triggered a full WebSocket reconnect via `_mark_restart_needed()` whenever instructions were updated. This interrupted the active conversation and added unnecessary latency.

This PR replaces the reconnect with a mid-session `LiveClientContent` event using `role="model"`, consistent with how the initial chat context is prefilled at connect time in `_main_task`.

## Changes

- `update_instructions()` now sends a `LiveClientContent` event via `_send_client_event()` when an active session exists
- Falls back to `_mark_restart_needed()` only when there is no active session

## Notes

- `role="system"` in `LiveClientContent` turns is silently ignored by the Gemini Live API despite documentation suggesting otherwise — verified by tracing the full send path to the WebSocket
- `role="model"` works reliably and matches the pattern used for context prefill on connect

## Testing

Verified manually that `update_instructions()` called mid-session correctly updates the model's behavior without reconnecting.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -558,12 +558,14 @@ class RealtimeSession(llm.RealtimeSession):
             logger.debug("Updating instructions mid-session")
             self._send_client_event(
                 types.LiveClientContent(
-                    turns=[types.Content(
-                        parts=[types.Part(text=instructions)],
-                        # Vertex AI ignores role=None or role="system" and only works with role="model".
-                        # Gemini Live API (non-Vertex) errors on role="system"; role=None works as system role.
-                        role="model" if self._opts.vertexai else None,
-                    )],
+                    turns=[
+                        types.Content(
+                            parts=[types.Part(text=instructions)],
+                            # Vertex AI ignores role=None or role="system" and only works with role="model".
+                            # Gemini Live API (non-Vertex) errors on role="system"; role=None works as system role.
+                            role="model" if self._opts.vertexai else None,
+                        )
+                    ],
                     turn_complete=False,
                 )
             )


livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -547,7 +547,21 @@ def update_options(
...
+                    return
+
+            # Active session exists — send mid-session system instruction update (no reconnect needed)
+            logger.warning(f"Updating instructions: {instructions}")
+            self._send_client_event(
+                types.LiveClientContent(
+                    turns=[types.Content(role="model", parts=[types.Part(text=instructions)])],
+                    turn_complete=False,
+                )
+            )
```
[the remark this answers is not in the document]
D-zigi: Good catch on the semantic mismatch. However, this is intentional - `role="system"` was attempted first but the Gemini Live API silently ignores it entirely (verified by logging the raw WebSocket payload — the message reaches the API correctly but has no effect on model behavior).
> `role="model"` was chosen because it's the same pattern `_main_task` uses when prefilling the chat context on connect (L771-783) — the initial instructions are also sent as a `role="model"` turn, and the model correctly treats them as behavioral guidance rather than prior utterances.

> The silent dropping of `role="system"` in `LiveClientContent` is likely a bug on Google's side - their docs claim it's supported but it isn't in practice. Until that's fixed, `role="model"` is the only reliable way to send instruction updates mid-session without a full reconnect.

tinalenguyen:
Hi @D-zigi, thank you for the PR! The docs you linked in the issue pertain to VertexAI, so I think` role="system"` would work for that but not the Gemini Live API. Could you guard this by checking `self._opts.vertexai` and see if updating instructions works in VertexAI?

D-zigi:
@tinalenguyen Thanks for the context! I tested both paths:

- **Vertex AI**: `role="system"` and `role=None` are silently ignored (no error, but the model doesn't follow the updated instructions). `role="model"` works correctly.
- **Gemini Live API (non-Vertex)**: `role="system"` throws an error. `role=None` works correctly.

I verified by logging the raw WebSocket payload - the messages reach the API correctly in both cases, so the issue is server-side behavior, not the SDK.

Updated the PR to handle both cases:
```python
role="model" if self._opts.vertexai else None

**Review by tinalenguyen — APPROVED** [review]
lgtm, thanks again! could you also resolve the ruff error?

**Comment by D-zigi** [comment]
Thanks! all issues resolved.

File: livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py [hunk 2]
```diff
@@ -547,7 +547,28 @@ def update_options(
...
+            # Active session exists — send mid-session system instruction update (no reconnect needed)
+            logger.debug("Updating instructions mid-session")
+            self._send_client_event(
+                types.LiveClientContent(
+                    turns=[
+                        types.Content(
+                            parts=[types.Part(text=instructions)],
+                            # Vertex AI ignores role=None or role="system" and only works with role="model".
+                            # Gemini Live API (non-Vertex) errors on role="system"; role=None works as system role.
+                            role="model" if self._opts.vertexai else None,
```
longcw, line 566: if vertexai doesn't support the new system role, should we keep the old behavior that restarts the session for vertexai?

D-zigi: A session restart is safe but adds latency and interrupts the active conversation.

`role="model"` works reliably on Vertex AI - the model correctly adopts the updated 
instructions, similar to how `role="system"` works on the Gemini Live API. The semantic 
difference (model's "prior output" vs a system directive) doesn't seem to matter in 
practice for instruction following.

This same pattern is already used in `generate_reply()` - instructions are passed as 
`role="model"` turns there too. So this PR is consistent with existing behavior.

If/when Vertex AI adds proper support for `role="system"` in `LiveClientContent`, 
the role can be updated without any other changes.

What I'd still consider refactoring is the way the instructions are being updated across all providers, because google does not support chat context replacement update_instructions cannot replace the existing instructions - it is for adding new.

longcw: > google does not support chat context replacement update_instructions cannot replace the existing instructions - it is for adding new

yep this is exactly what I am concerning, the original `update_instructions` replaces the instructions with a new one, while with this change even the model works well with the `role="model"`, it won't forget the old instructions, right? this may cause a behavior change.

D-zigi: Exactly!
The Gemini Live API gets its instructions updated the same way with `role="system"` - it does not forget the old instructions, it adds the new instructions.
But in my experience of live calls speed is more important than accuracy here, 4-5 seconds downtime of the model makes the call experience much worse. I guess the only way is to report this behavior to google for them to fix which I'm already on. Will be happy if we could collaborate on that.

In addition we could figure a new argument/function specifically for google realtime to either replace the instructions(reconnect) or to append - same way i did, as google expects.

D-zigi: @longcw Is there a way to contact you?

longcw: you can find me from the livekit community slack or [longch1024@gmail.com](mailto:longch1024@gmail.com)

PR #5055, opened 2026-03-09
(responses llm): override provider property and set use_websocket to False for wrappers

tinalenguyen:
also if a client is given, then set `use_websocket=False`


davidzhao (approved):
lg

PR #5056, opened 2026-03-09
feat(google): support universal credential types in STT and TTS credentials_file

rafallezanko:
Previously, the Google STT and TTS plugins loaded file-based credentials using from_service_account_file, which only supports service account JSON. This prevented users from using other credential types such as Workload Identity
  Federation (WIF), Google Workload Identity (GWI), or external account credentials.

  Changes:
  - Replaced TextToSpeechAsyncClient.from_service_account_file (TTS) and SpeechAsyncClient.from_service_account_file (STT) with google.auth.load_credentials_from_file, which handles all credential types recognized by the google-auth
  library (service accounts, WIF external accounts, GWI, authorized user credentials, etc.)
  - The loaded credentials object is passed directly to the client constructor
  - credentials_info (dict-based) and the default ADC path are unchanged
  - Added # type: ignore[no-untyped-call] to both call sites since google-auth lacks type stubs

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
@@ -216,6 +216,7 @@ class STT(stt.STT):
         self._location = location
         self._credentials_info = credentials_info
         self._credentials_file = credentials_file
+        self._project_id: str | None = None
 
         if not is_given(credentials_file) and not is_given(credentials_info):
             try:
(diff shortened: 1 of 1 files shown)


rafallezanko:
Hi @chenghao-mou , can you try to have a look at the change please? :)

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -278,9 +279,11 @@ async def _create_client(self, timeout: float) -> SpeechAsyncClientV2 | SpeechAs
                 self._credentials_info, client_options=client_options
             )
         elif is_given(self._credentials_file):
-            client = client_cls.from_service_account_file(
-                self._credentials_file, client_options=client_options
+            credentials, _ = google.auth.load_credentials_from_file(  # type: ignore[no-untyped-call]
```
chenghao-mou: we should save this project_id so we can reuse it later in _get_recognizer, otherwise, it might default back to the default credential.

rafallezanko: Great idea, thanks! I've committed a fix.

PR #5066, opened 2026-03-10
feat(core): allow skip_reply when calling commit_user_turn

### LINKED ISSUE #5026: Feature Request: commit_user_turn() without automatic response generation
### Feature Type

Would make my life easier

### Feature Description

This is a feature request for https://community.livekit.io/t/how-to-commit-user-turn-without-generating-a-response/433

- We are using manual turn taking (https://docs.livekit.io/agents/logic/turns/#manual)
- We are using cascaded models (STT/LLM/TTS)
- Currently when commit_user_turn() is called it always triggers a response from the model automatically 

We need to force the user to give up their turn and get the agent to “say” something specific - so in this case, we _do not want to trigger the agent to say anything_ automatically when we end the user turn and call commit_user_turn().

We need to call commit_user_turn() though because we need to commit the user's latest speech from the STT (i.e. buffered transcript) into the livekit local state so the LLM is aware of it before the user next ends their turn. i.e. we do not wish to discard the users latest speech whenever the agent interrupts their turn. 

Ideally we could do something like `commit_user_turn(skip_reply=False)`

We could manually commit it to local state i.e. consume the transcript creation event and manually add it to the local context maybe like

```
@self._session.on("user_input_transcribed")
         ...
        agent = session.current_agent
        agent._chat_ctx.items.append(chat_msg)
        session._conversation_item_added(chat_msg)
```
like this, but the problem would be I do not know how to “flush” the STT to get the latest available transcript as input to this.  

Another idea would be to do

```
commit_user_turn()
session.interupt()
generate agent speech
```

But as it says here [Search | DeepWiki](https://deepwiki.com/search/if-i-call-in-python-sdk-commit_7c5b385d-d4b7-41f7-b80c-b8630a046f19?mode=fast) I think that would not be bulletproof either as its possible to interrupt before the LLM request starts making the interrupt a no-op

### Additional Context

_No response_

chenghao-mou:
Expose skip_reply at session level for both realtime and cascading agents when calling commit_user_turn.

closes #5026

This can be tested with a RPC:

```
    @ctx.room.local_participant.register_rpc_method("commit_user_turn")
    async def on_commit_user_turn(data: rtc.RpcInvocationData) -> str:
        skip_reply = data.payload == "skip_reply"
        logger.info(f"RPC commit_user_turn called, skip_reply={skip_reply}")
        fut = session.commit_user_turn(skip_reply=skip_reply)
        transcript = await fut
        logger.info(f"Committed transcript: {transcript}")
        return transcript
```

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1032,15 +1032,15 @@ class AgentActivity(RecognitionHooks):
             self._rt_session.clear_audio()
 
     def commit_user_turn(
-        self, *, transcript_timeout: float, stt_flush_duration: float
+        self, *, transcript_timeout: float, stt_flush_duration: float, skip_reply: bool = False
     ) -> asyncio.Future[str]:
-        skip_reply: bool = False
         if self._rt_session is not None:
-            # commit audio buffer and trigger response generation
+            # commit audio buffer and conditionally trigger response generation
+            self._rt_session.commit_audio()
+            if not skip_reply:
+                self._session.generate_reply()
             # `skip_reply` prevents duplicate reply from _on_user_turn_completed
             # but keeps flushing STT transcript into the chat context
-            self._rt_session.commit_audio()
-            self._session.generate_reply()
             skip_reply = True
 
         assert self._audio_recognition is not None
@@ -1530,6 +1530,12 @@ class AgentActivity(RecognitionHooks):
                         self._session._conversation_item_added(user_message)
                     return
                 self._rt_session.commit_audio()
+
+        if info.skip_reply:
+            if info.new_transcript != "":
+                self._agent._chat_ctx.items.append(user_message)
+                self._session._conversation_item_added(user_message)
+            return
 
         if (current_speech := self._current_speech) is not None:
             if not current_speech.allow_interruptions:
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1082,7 +1082,11 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         self._activity.clear_user_turn()
 
     def commit_user_turn(
-        self, *, transcript_timeout: float = 2.0, stt_flush_duration: float = 2.0
+        self,
+        *,
+        transcript_timeout: float = 2.0,
+        stt_flush_duration: float = 2.0,
+        skip_reply: bool = False,
     ) -> asyncio.Future[str]:
         """Commit the user turn and generate a reply.
 
@@ -1095,6 +1099,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
                 Increase this value if the STT is slow to respond.
             stt_flush_duration (float, optional): The duration of the silence to be appended to the STT
                 to flush the buffer and generate the final transcript.
+            skip_reply (bool, optional): Whether to skip the reply generation after committing the user turn.
 
         Returns:
             asyncio.Future[str]: A future that resolves with the audio transcript.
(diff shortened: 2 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1032,15 +1032,15 @@ def clear_user_turn(self) -> None:
...
-        self, *, transcript_timeout: float, stt_flush_duration: float
+        self, *, transcript_timeout: float, stt_flush_duration: float, skip_reply: bool = False
     ) -> asyncio.Future[str]:
-        skip_reply: bool = False
         if self._rt_session is not None:
-            # commit audio buffer and trigger response generation
+            # commit audio buffer and conditionally trigger response generation
+            self._rt_session.commit_audio()
+            if not skip_reply:
+                self._session.generate_reply()
```
longcw, line 1041: should we also skip the reply of realtime model if `skip_reply` is False?

chenghao-mou: We set `skip_reply` to `True` after this line so downstream realtime reply is always skipped at [line](https://github.com/livekit/agents/blob/5c483d12396c2d96a9b57a6c1b9bb4fb310dabc7/livekit-agents/livekit/agents/voice/agent_activity.py#L1526). This is the only place where we allow a response.

longcw: no I mean we also skip the generate reply of realtime model if it's False?

chenghao-mou: Are you saying we should always skip the reply for the realtime model? I am not sure I understand why, though.

longcw: yes, I am not sure if users set `skip_reply=False`, are they expecting a response from realtime model?

chenghao-mou: The default behavior of `commit_user_turn` before this PR is committing and then triggering a response for both realtime and cascading agents. Previously, this `skip_reply` flag was only used to prevent _duplicate_ realtime responses and flush the STT transcripts. Now it is exposed to control the replying behaviour for both consistently, so I think to your question, my answer would be yes.

longcw: previously `skip_reply` is just an internal flag to ensure only one response is generated. the `commit_user_turn` will always generate a reply.

when you add a `skip_reply` option to this method, it's weird to me that it's not respected for the realtime model. why do you think it's expected?

longcw: could it be something like this?
```python
    def commit_user_turn(
        self, *, transcript_timeout: float, stt_flush_duration: float, skip_reply: bool = False
    ) -> asyncio.Future[str]:
            # commit audio buffer and trigger response generation
            # `skip_reply` prevents duplicate reply from _on_user_turn_completed
            # but keeps flushing STT transcript into the chat context
            self._rt_session.commit_audio()
            if not skip_reply:
                self._session.generate_reply()
                skip_reply = True

        assert self._audio_recognition is not None
        return self._audio_recognition.commit_user_turn(
            audio_detached=not self._session.input.audio_enabled,
            transcript_timeout=transcript_timeout,
            stt_flush_duration=stt_flush_duration,
            skip_reply=skip_reply,
        )

```

chenghao-mou: Not sure I understand why it is not respecting the flag. The code you are suggesting is semantically the same as this PR:

CleanShot 2026-03-10 at 14 51 42@2x

skip_reply is set to True after the if block.

longcw: never mind... I guess what I want to ask is "skip the reply of realtime model if skip_reply is True"...  the double negative....

PR #5068, opened 2026-03-10
feat(core): add configure_logging() API for user controlled log setup

eelkeh:
Add a public `configure_logging()` function in `livekit.agents.log` that allows for custom handlers, formatters, an option for JSON mode, and log levels before the CLI starts. The CLI detects pre-existing logging configuration and skips its own setup, preventing double-handler issues.

This is especially useful for using JSON formatted log entries while in dev mode, retaining live reload (which is currently not possible) for log aggregators and/or optimized logs for coding agents.


eelkeh:
Let me know if you need a different approach to tackle this, happy to dive in and redo it or make any changes!

PR #5079, opened 2026-03-11
fix(types): replace TypeGuard with TypeIs in is_given for bidirectional narrowing

longcw:


livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -264,12 +264,12 @@ def update_options(
         if is_given(language):
             self._opts.language = LanguageCode(language) if language else None
         if is_given(voice):
-            self._opts.voice = cast(str | list[float], voice)
+            self._opts.voice = voice
         if is_given(speed):
             self._opts.speed = cast(TTSVoiceSpeed | float, speed)
```
chenghao-mou, line 269: nit: one more to go.

livekit-agents/livekit/agents/utils/misc.py
```diff
@@ -4,9 +4,11 @@
 import re
 import time
 import uuid
-from typing import TypeGuard, TypeVar
+from typing import TypeVar
 from urllib.parse import urlparse
 
+from typing_extensions import TypeIs
```
chenghao-mou, line 10: nit: I just learned this when googling it: TypeIs is available in typing since 3.13, so we could do this:

```python
try:
    from typing import TypeIs  # 3.13+
except ImportError:
    from typing_extensions import TypeIs
```

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
longcw: somehow this one doesn't work without `cast`...

livekit-agents/livekit/agents/utils/misc.py
longcw: in `typing_extensions` it's defined as following, it should be fine to always import from `typing_extensions`
```python
# 3.13+
if hasattr(typing, 'TypeIs'):
    TypeIs = typing.TypeIs
# <=3.12
else:
    ...
```

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
chenghao-mou: Maybe a mypy issue? ty warned me about redundant cast.

longcw: I think so... do you think if we can switch to `ty`?

chenghao-mou: It is getting better, but still somewhat noisy (334 issues in the livekit-agents folder). Maybe we need a different set of rules for it. I will look into it. It is definitely faster than mypy.

PR #5080, opened 2026-03-11
fix: include `null` in enum array for nullable enum schemas

### LINKED ISSUE #5078: Bug Report: `_ensure_strict_json_schema` drops `null` from `enum` arrays in nullable enum fields
## Bug Description

When using `response_format` with a Pydantic model that has a nullable enum field (e.g. `Literal["a", "b"] | None`), the `_ensure_strict_json_schema` function in `livekit.agents.llm._strict` simplifies the `anyOf` union but fails to add `null` to the `enum` array. This makes it structurally impossible for the model to output `null` for that field under constrained decoding.

Pydantic generates this schema for `Literal["male", "female"] | None`:

```json
{
  "anyOf": [
    { "enum": ["male", "female"], "type": "string" },
    { "type": "null" }
  ]
}
```

After `_ensure_strict_json_schema` (lines 150-171 of `_strict.py`), this becomes:

```json
{
  "enum": ["male", "female"],
  "type": ["string", "null"]
}
```

The `type` array correctly includes `"null"`, but the `enum` array still only contains `["male", "female"]`. In JSON Schema, a value must satisfy **both** `type` and `enum` constraints. Since `null` is not in the `enum`, the model is forced to pick a concrete value even when `null` is the correct answer.

Fields without an `enum` (e.g. `bool | None`, `int | None`) are unaffected because their schema uses only `type: ["boolean", "null"]`, which correctly permits `null`.

## Expected Behavior

The simplified schema should include `null` in the `enum` array:

```json
{
  "enum": ["male", "female", null],
  "type": ["string", "null"]
}
```

This allows the constrained decoder to output `null` as a valid option alongside the enum values.

## Reproduction Steps

```python
from typing import Literal
from pydantic import BaseModel, Field
from livekit.agents.llm._strict import to_strict_json_schema
import json

class MyModel(BaseModel):
    # bool | None -- works correctly (no enum)
    flag: bool | None = Field(None)

    # Literal enum | None -- null gets dropped from enum
    status: Literal["active", "inactive"] | None = Field(None)

schema = to_strict_json_schema(MyModel)
print(json.dumps(schema, indent=2))

# Output for 'flag':    {"type": ["boolean", "null"]}         <-- correct, null is valid
# Output for 'status':  {"enum": ["active", "inactive"],      <-- BUG: null missing from enum
#                         "type": ["string", "null"]}
```

## Operating System

Windows 11 (10.0.26200)

## Models Used

OpenAI GPT-4.1 (via `livekit-plugins-openai`), using `response_format` with a Pydantic model for structured output.

## Package Versions

```
livekit-agents==1.4.4
pydantic==2.12.5
Python 3.13.11
```

## Proposed Solution

In `_strict.py`, the nullable union simplification block (lines 150-171) should append `None` to the `enum` array when present:

```python
# simplify nullable unions ("anyOf" or "oneOf")
for union_key in ("anyOf", "oneOf"):
    variants = json_schema.get(union_key)
    if is_list(variants) and len(variants) == 2 and {"type": "null"} in variants:
        non_null = next(
            (item for item in variants if item != {"type": "null"}),
            None,
        )
        assert is_dict(non_null)

        if "type" not in non_null:
            continue

        t = non_null["type"]
        if isinstance(t, str):
            non_null["type"] = [t, "null"]

        # --- FIX: propagate null into the enum array ---
        if "enum" in non_null:
            non_null["enum"].append(None)

        merged = {k: v for k, v in json_schema.items() if k not in ("anyOf", "oneOf")}
        merged.update(non_null)
        json_schema = merged
        break
```

## Additional Context

This is particularly impactful for agentic workflows that use `response_format` for structured extraction. When the model cannot output `null` for an enum field, it hallucinates a concrete value instead -- even when the system prompt explicitly instructs it to output null. The prompt cannot override a structural constraint imposed by constrained decoding.

MSameerAbbas:
## Summary

Fixes #5078.

- `_ensure_strict_json_schema` collapsed nullable unions (e.g. `Literal["a","b"] | None`) into a single schema but only added `"null"` to the `type` array, not the `enum` array. JSON Schema requires a value to satisfy **both** constraints, so the model could never output `null` for these fields -- it would hallucinate a concrete value instead.
- Refactored the nullable-union simplification block to make both `type` and `enum` constraints nullable in parallel, and simplified the merge with dict union (`|`).
- Added tests covering nullable enum, nullable non-enum, and non-nullable enum fields.

## Test plan

- [x] `pytest tests/test_strict_json_schema.py` -- 3 tests pass
- [x] `ruff check` and `ruff format --check` -- clean


MSameerAbbas:
@tinalenguyen @chenghao-mou could you review this?

MSameerAbbas:
moved tests to [test_tools.py](https://github.com/livekit/agents/blob/livekit-agents@1.3.12/tests/test_tools.py)

PR #5083, opened 2026-03-11
fix(eou): only reset speech/speaking time when no new speech

chenghao-mou:
Only resets user speech-related timestamps when there is no new speech when committing a turn, by comparing the captured `last_speaking_time` and the latest `self._last_speaking_time`


chenghao-mou:
> lgtm! one question: any chance the race conditions happen to the `_last_final_transcript_time` as well?

If there is a new final transcript, we will call the _run_eou_detection, which cancels the old _end_of_turn_task. So it should be fine.

PR #5087, opened 2026-03-11
add FileContent and ToolOutput

### LINKED ISSUE #4893: feat: support ImageContent in tool return value

tinalenguyen:
- backward compatible, does not break existing user code assuming that tool outputs are `str`
- add more depth to function tool outputs, allowing for `ImageContent` and `FileContent` to be passed along with strings, like so:

```
    @function_tool
    async def get_chart(self, context: RunContext):
        """Returns a chart image and checkup report when the user asks for a chart or visual."""
        with open("mock_checkup_report.pdf", "rb") as f:
            report_data = f.read()
        img_content = ImageContent(image="https://my.clevelandclinic.org/-/scassets/images/org/health/articles/10881-vital-signs")
        file_content = FileContent(name="mock_checkup_report.pdf", data=report_data, mime_type="application/pdf"),
        return "Here is the normal adult vital chart:", img_content, "And the patient report file:", file_content
            
```


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -312,7 +346,7 @@ class FunctionCallOutput(BaseModel):
     type: Literal["function_call_output"] = Field(default="function_call_output")
     name: str = Field(default="")
     call_id: str
-    output: str
+    output: ToolOutput
```
theomonnom, line 460: Can't this just be?
`ToolOutput = ImageContent | FileContent | str`

tinalenguyen: what if users wanted to pass multiple items from a function tool

livekit-agents/livekit/agents/llm/_provider_format/anthropic.py
```diff
@@ -64,13 +64,19 @@ def to_chat_ctx(
...
+                    parsed = json.loads(msg.output.text_contents)
+                    result_content = (
+                        parsed
+                        if isinstance(parsed, list)
+                        else [{"type": "text", "text": msg.output.text_contents}]
+                    )
+                except (json.JSONDecodeError, TypeError):
+                    result_content = [{"type": "text", "text": msg.output.text_contents}]
+            for img in msg.output.image_contents:
+                result_content.append(_to_image_content(img))
```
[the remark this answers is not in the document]
tinalenguyen: anthropic accepts `None` as tool result content

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -218,6 +220,148 @@ class AudioContent(BaseModel):
...
+    type: Literal["file_content"] = Field(default="file_content")
+    name: str
+    """Display name of the file."""
+    data: str | bytes
+    """File data — str for text, bytes for binary (e.g. PDF). Bytes are base64-encoded in JSON."""
+    mime_type: str
+    """MIME type, e.g. 'text/plain', 'application/pdf'."""
+
+
+class ToolOutput(str):
```
theomonnom, line 235: I'm not sure if this should be a `str`.
This doesn't feel right

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -218,6 +220,148 @@ class AudioContent(BaseModel):
...
+                        mime = img.mime_type or "image/jpeg"
+                        d["image"] = (
+                            f"data:{mime};base64,{base64.b64encode(img.data_bytes).decode('utf-8')}"
+                        )
+                    elif img.external_url:
+                        d["image"] = img.external_url
+                result.append(d)
+            elif isinstance(item, FileContent):
+                result.append(item.model_dump(mode="json"))
+        return result
```
theomonnom, line 342: All of that will not be needed if we don't use str?

livekit-agents/livekit/agents/llm/chat_context.py
tinalenguyen: ideally it isn't, but what about existing user code assuming that the output is a string?

```
 for item in agent_session.chat_ctx.items:
      if isinstance(item, FunctionCallOutput) and item.name == "book_appointment":
          if  item.output == "confirmed": ...
```
wouldn't this fail if ToolOutput was a list or BaseModel?

aolast:
Quick bump here. Any update on the remaining API direction for `ToolOutput` / `FunctionCallOutput.output`?

It looks like this is mostly waiting on deciding whether `output` should stay string-compatible for existing user code. Once that is settled, the branch can be updated against `main` and CI rerun.

aolast:
A bit more context on why I’m asking for an update here: for our use case, having image content come back as part of the tool output is the clean path.

If we try to work around this by injecting `ImageContent` into the chat context from inside the tool function, that is not quite equivalent. To make the image available to the model in the same turn, we would need extra orchestration around the current tool-response flow, e.g. stop/interrupt the current generation and call `generate_reply` again with the updated context, or otherwise force a second reply pass. So the workaround is possible, but it is a fairly special-case control-flow hack rather than a normal tool result.

That’s why I’m interested in whether there is an updated preferred direction for `ToolOutput` / `FunctionCallOutput.output` here.

PR #5089, opened 2026-03-11
(google realtime): replace deprecated mediaChunks

tinalenguyen:
`mediaChunks` is deprecated, use `audio`, `video` , and `text` instead

[reference](https://ai.google.dev/api/live#bidigeneratecontentrealtimeinput)

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -925,7 +925,9 @@ class RealtimeSession(llm.RealtimeSession):
                         types.LiveClientRealtimeInput,
                     ),
                 ):
-                    if not isinstance(msg, types.LiveClientRealtimeInput) or not (msg.audio or msg.video or msg.text):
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -906,9 +904,12 @@ async def _send_task(self, session: AsyncSession) -> None:
...
                 elif isinstance(msg, types.LiveClientRealtimeInput):
-                    if msg.media_chunks:
-                        for media_chunk in msg.media_chunks:
-                            await session.send_realtime_input(media=media_chunk)
+                    if msg.audio:
+                        await session.send_realtime_input(audio=msg.audio)
+                    elif msg.video:
+                        await session.send_realtime_input(video=msg.video)
+                    elif msg.text:
+                        await session.send_realtime_input(text=msg.text)
```
theomonnom, line 912: Can they be sent at the same time? What's the best thing to do here?

tinalenguyen: we can only send one argument per call or else google throws an error

Issue #4123, opened 2025-11-29
Persistent STT stream


superMDguy:
### Feature Type

Would make my life easier

### Feature Description

Currently, the default `stt_node` creates a new stream after each agent handoff. It would be nice if there was an option to re-use the same stream to reduce handoff latency. I've also noticed (specifically with Deepgram) that STT hallucinations are more common when the stream is first created, so having a persistent stream would reduce those issues

### Workarounds / Alternatives

Currently working on using a custom `stt_node` that persists the stream to session userdata.

PR #5093, opened 2026-03-12
feat(voice): reuse STT connection across agent handoffs

longcw:
close https://github.com/livekit/agents/issues/4123


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -508,7 +508,23 @@ async def _traceable_on_enter() -> None:
...
+        """Return True if new_activity can reuse this activity's AudioRecognition.
+
+        Requires the same STT instance and the same stt_node implementation
+        """
+        if self._audio_recognition is None:
+            return False
+        if self.stt is None or new_activity.stt is None:
+            return False
+        return (
+            type(self.agent).stt_node is type(new_activity.agent).stt_node
```
toubatbrian: Should we also check for tts, vad, as well as other configurations?

longcw: `vad` is always updated, the tts and llm streams are created per turn so no need to reuse.

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -451,6 +517,23 @@ def update_vad(self, vad: vad.VAD | None) -> None:
             self._interruption_detection is not None and self._vad is not None
         )
 
+    async def detach_stt(self) -> _STTPipeline | None:
+        """Detach the STT pipeline for handoff to another AudioRecognition.
+
+        Returns the pipeline (pump task + channels) without stopping it.
+        The caller is responsible for passing it to the new AudioRecognition
+        via update_stt(..., pipeline=pipeline).
```
chenghao-mou: nit: via start first, right?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1203,43 +1203,58 @@ async def _update_activity(
...
-                )
-                self._next_activity = None
-                self._activity = None
-                return
+            # detach STT pipeline from old activity for potential reuse in the new one
+            _reuse_pipeline = (
+                await self._activity._detach_stt_pipeline_if_reusable(self._next_activity)
+                if self._activity is not None
+                else None
+            )
```
chenghao-mou, line 1221: Q: if this is detached before the new one is ready (until start/resume in the new activity), the audio frames are silently dropped during the window, right? 

I guess this is pre-existing before the PR. At least it is better now when we don't have to restart the STT connection.

longcw: yes, the audio frames are dropped during the handoff.

chenghao-mou (approved):
lgtm. Tested locally with agent handoffs. Do you think we should add something to track the handoff time?

Issue #5092, opened 2026-03-12
openai.responses.LLM breaks when user interrupts during active tool call (400: No tool output found)


nader-bytesai:
### Bug Description

When using `openai.responses.LLM()` and a user speaks while a long-running tool call is still executing, the next LLM request includes the `function_call` in the chat context but **no corresponding `function_call_output`** (because the tool hasn't returned yet). The Responses API rejects this with a 400 error:

```
No tool output found for function call call_tqhMiOw00xHdJm3aGQzOtBra.
```

**This permanently breaks the agent for the remainder of the session** — every subsequent LLM request sends the same malformed context, and the agent becomes completely unresponsive. There is no recovery.

This does **not** happen with `openai.LLM()` (Chat Completions API) — only with `openai.responses.LLM()` (Responses API). It reproduces with both `use_websocket=True` and `use_websocket=False`.

### Additional issue: WebSocket mode masks the real error

When `use_websocket=True` (default), the 400 error from OpenAI is returned as a WebSocket error event missing the `sequence_number` field required by the `ResponseErrorEvent` pydantic model. This causes a `ValidationError` in `_parse_ws_event` (line 442), which gets caught by the generic `except Exception` and re-raised as a generic `APIConnectionError("Connection error.")` — completely masking the actual 400 from OpenAI. Setting `use_websocket=False` reveals the real error.

### Expected Behavior

The framework should not send an LLM request with an incomplete tool call in the chat context. Either:
1. Skip/defer LLM inference while a tool call is actively executing
2. Strip the incomplete tool call from the chat context before sending
3. Add a placeholder tool output

### Reproduction Steps

1. Create the minimal agent below with `openai.responses.LLM()` and a tool that takes ~5 seconds
2. Start the agent and connect via phone/browser
3. Say "Can you send me the menu?" to trigger the tool
4. While the tool is executing (~5s), say anything (e.g., "Hello? Did you send it?")
5. The agent breaks with repeated 400 errors and never recovers

### Full Reproduction File

```python
import logging
import asyncio
import os
from dotenv import load_dotenv
from livekit import rtc
from livekit.agents import (
    Agent,
    AgentServer,
    AgentSession,
    JobContext,
    JobProcess,
    RunContext,
    cli,
    function_tool,
    inference,
    room_io,
)
from livekit.plugins import (
    noise_cancellation,
    openai,
    silero,
)
from livekit.plugins.turn_detector.english import EnglishModel

logger = logging.getLogger()
[...]

PR #5094, opened 2026-03-12
fix(openai): only reuse previous_response_id when pending tool calls are completed

longcw:
fix https://github.com/livekit/agents/issues/5092


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
```diff
@@ -325,6 +328,15 @@ def chat(
...
         )
 
+    def _pending_tool_calls_completed(self, items: list[ChatItem]) -> bool:
+        if not self._pending_tool_calls:
+            return True
+
+        completed_tool_calls = {
+            item.call_id for item in items if item.type == "function_call_output"
+        }
+        return all(call_id in completed_tool_calls for call_id in self._pending_tool_calls)
```
chenghao-mou, line 338: Do we want to remove completed ones from the set?

longcw: `pending_tool_calls` is bound with `response_id`, so perhaps keep it and reset when `response_id` changed.

PR #5095, opened 2026-03-12
fix(stt): reset VAD when STT sends EOT

chenghao-mou:
This allows VAD to start interrupting if STT sends a premature EOT.

When turn_detection is set to STT, an EOT from STT will force to change user state from speaking to listening, while VAD internal state is unchanged. This would skip any path for VAD-based interruption (including adaptive interruption) unless VAD picks up a new SOS/EOS.

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -483,7 +483,11 @@ class AudioRecognition:
             with trace.use_span(self._ensure_user_turn_span()):
                 self._hooks.on_end_of_speech(None)
 
+            # STT EOT changes user state from speaking to listening without updating VAD internal states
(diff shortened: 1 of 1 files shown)


theomonnom:
What is a "reset" doing? Can you provide more context on why this is needed?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -483,6 +483,14 @@ async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
             with trace.use_span(self._ensure_user_turn_span()):
                 self._hooks.on_end_of_speech(None)
 
+            # reset VAD so that incorrect end of turn from STT can be corrected by VAD interruption
```
theomonnom, line 489: Can you add a comment that it is related to the fact that vad is waiting some amount of "speech" before triggering an event

PR #5104, opened 2026-03-13
feat(voice): add callable TextTransforms support with built-in replace transform

longcw:


examples/voice_agents/basic_agent.py
```diff
@@ -103,6 +104,11 @@ async def entrypoint(ctx: JobContext):
         false_interruption_timeout=1.0,
         # blocks interruptions for a few seconds after the agent starts speaking to allow client to calibrate AEC
         aec_warmup_duration=3.0,
+        tts_text_transforms=[
+            "filter_emoji",
+            "filter_markdown",
+            text_transforms.replace({"LiveKit": "<<ˈ|l|aɪ|v>> <<ˈ|k|ɪ|t>>"}),
+        ],
```
chenghao-mou, line 111: Should this be part of TTS so that they could support different transformations?

longcw: perhaps using a custom `tts_node` if they want different tranaforms per agent, or we can add agent specific tts_text_transforms option

livekit-agents/livekit/agents/voice/transcription/filters.py
davidzhao, line 1: should these be in transcription? or a new package? Since these transform should only apply to the TTS node, and not the transcripts that we generate & collect.

longcw: maybe it fits here, the `transcription` module handles agent output transcription — both post-processing (audio/transcript sync) and pre-processing (text transforms before TTS).

Issue #1399, opened 2025-01-21
Feature Request: Allow Selective Track Subscription for VoiceAssistant


boussaidev:
Description: Currently, the VoiceAssistant automatically subscribes to all participant audio tracks when instantiated, even when AutoSubscribe.SUBSCRIBE_NONE is used. This behavior prevents the ability to selectively subscribe to specific tracks and ignore others.

In use cases like the WebRTC web app, there is a need for the voice assistant to subscribe only to a specific participant's audio track, while ensuring that the other participant cannot hear the audio. The track should have a unique identifier (e.g., a custom name) so that the agent can subscribe to it while other participants' audio is ignored.

Proposed Solution: It would be beneficial if the VoiceAssistant instance could support more granular control over track subscriptions. Specifically, adding a feature to allow the agent to subscribe to specific tracks based on a unique identifier, while ignoring other tracks.

This would allow better flexibility in managing track subscriptions and would be especially useful for applications where private communication with the agent is required without broadcasting audio to other participants.

Use Case:

In a WebRTC web app, a user can interact with the voice assistant while ensuring the other participant does not hear the audio.
The agent can subscribe to a participant's specific audio track using its unique name or identifier, while other tracks are ignored.
Benefits:

Enhanced control over which audio tracks the agent subscribes to.
Improved privacy and user experience in scenarios with multiple participants.
Would love to see this functionality added to the VoiceAssistant!

boussaidev:
@dsa 
@nfma

longcw:
Thanks for the suggestion. The current agent will automatically subscribe to the first participant to join the room. An upcoming major release will allow subscribing to any specific participant and switching the audio input to the agent 🚀

PR #5106, opened 2026-03-13
fix(personaplex): fix connection and audio streaming

ShayneP:
## Summary

Fixes the PersonaPlex realtime plugin so it can successfully connect and stream audio with the PersonaPlex server. Based on @milanperovic's work in #1399.

- **Pin `sphn>=0.1.4,<0.2`** — The PersonaPlex server requires sphn `<0.2`. The 0.1.x and 0.2.x versions use incompatible wire formats and different APIs (`append_pcm()`/`read_bytes()` two-step in 0.1.x vs single return value in 0.2.x). Using the wrong version caused the server's `opus_reader.read_pcm()` to return `None`, crashing the server-side opus decode loop.
- **Fix audio frame size (1920 samples, was 2400)** — sphn 0.1.x requires valid Opus frame sizes (120, 240, 480, 960, 1920, 2880). The previous value of `SAMPLE_RATE // 10 = 2400` is not a valid Opus frame size and raises a `ValueError`.
- **Wait for handshake before sending audio** — The PersonaPlex server runs an `is_alive()` check during system prompt processing (~2-9s) that consumes incoming WS messages without feeding them to the Opus decoder. Audio sent during this phase is silently dropped. The official JS client only starts streaming mic audio after the handshake arrives. This change matches that behavior — audio frames buffer in the send channel during prompt processing and flush immediately when the handshake arrives.
- **Handle close code 1000 gracefully** — When the server closes the connection normally (code 1000), finalize any active generation and reconnect cleanly instead of treating it as a retriable error with exponential backoff.
- **Flush stale audio on reconnect** — Close and replace `_msg_ch` between connections so Opus-encoded frames from a previous session don't leak into a new connection with fresh codec state.
- **Use sphn 0.1.x API** — Switch `_encode_and_send` and `_handle_audio_data` to the two-step sphn 0.1.x API.


ShayneP:
Closing, fixed in #4660

PR #5114, opened 2026-03-16
fix(realtime): sync remote items to local chat_ctx with placeholders to prevent in-flight deletion

longcw:
fix https://github.com/livekit/agents/pull/5021

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1175,7 +1175,7 @@ class AgentActivity(RecognitionHooks):
         # so they don't need placeholders.
         last_item_id = local_chat_ctx.items[-1].id if local_chat_ctx.items else None
         if ev.previous_item_id is None or ev.previous_item_id == last_item_id:
-            local_chat_ctx.items.append(ev.item)
+            local_chat_ctx.items.append(ev.item.model_copy())
 
     def _on_error(
         self, error: llm.LLMError | stt.STTError | tts.TTSError | llm.RealtimeModelError


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1161,6 +1163,20 @@ def _on_metrics_collected(
...
+        if local_chat_ctx.get_by_id(ev.item.id) is not None:
+            return
+
+        # only add placeholders for server-initiated items (responses, function calls),
+        # which always append at the end of the conversation. client-initiated items
+        # (from update_chat_ctx) already exist in _agent._chat_ctx and go local→remote,
+        # so they don't need placeholders.
+        last_item_id = local_chat_ctx.items[-1].id if local_chat_ctx.items else None
+        if ev.previous_item_id is None or ev.previous_item_id == last_item_id:
+            local_chat_ctx.items.append(ev.item.model_copy())
```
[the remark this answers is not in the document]
longcw: The server still has that item, the next server-initiated item would have previous_item_id pointing to this message, which wouldn't match last_item_id in _on_remote_item_added, so the next placeholder would be silently dropped.

so maybe we should allow empty item for realtime session.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1161,6 +1163,20 @@ def _on_metrics_collected(
...
+        local_chat_ctx = self._agent._chat_ctx
+        if local_chat_ctx.get_by_id(ev.item.id) is not None:
+            return
+
+        # only add placeholders for server-initiated items (responses, function calls),
+        # which always append at the end of the conversation. client-initiated items
+        # (from update_chat_ctx) already exist in _agent._chat_ctx and go local→remote,
+        # so they don't need placeholders.
+        last_item_id = local_chat_ctx.items[-1].id if local_chat_ctx.items else None
+        if ev.previous_item_id is None or ev.previous_item_id == last_item_id:
```
chenghao-mou, line 1177: QQ: would parallel tool calls have the same `previous_item_id`? I wonder if we should check for ev.previous_item_id in item_ids instead.

livekit-agents/livekit/agents/voice/agent_activity.py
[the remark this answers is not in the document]
chenghao-mou: I think empty message is fine? If it is interrupted w/o any text forwarded or is an audio-only message, the remote message should be "empty" as well, right?

livekit-agents/livekit/agents/voice/agent_activity.py
StianHanssen: gpt-realtime can do "parallel" tool calls by doing multiple tool calls concurrently on one turn, but they have a sequential order on arrival from what I understand.

When I saw this line, I have to admit it stood out as a risk, but I can't think of a realistic scenario that causes an issue.

I thought we could be on the safe side, and log a warning in the else condition, just to catch if there is anything we have missed.

StianHanssen:
Nice work on this, I can't see any clear issues when reading over! I will load in the branch into my test suit tomorrow morning for some checks, but I suspect it will pass

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: yeah, I think the function calls will come in sequential with different previous item id, each `ConversationItemAdded` event carries a `ConversationItem` (either a message or a function call) and a `previous_item_id`.

```python
class ConversationItemAdded(BaseModel):
    """Sent by the server when an Item is added to the default Conversation.

    This can happen in several cases:
    - When the client sends a `conversation.item.create` event.
    - When the input audio buffer is committed. In this case the item will be a user message containing the audio from the buffer.
    - When the model is generating a Response. In this case the `conversation.item.added` event will be sent when the model starts generating a specific Item, and thus it will not yet have any content (and `status` will be `in_progress`).

    The event will include the full content of the Item (except when model is generating a Response) except for audio data, which can be retrieved separately with a `conversation.item.retrieve` event if necessary.
    """

    event_id: str
    """The unique ID of the server event."""

    item: ConversationItem
    """A single item within a Realtime conversation."""

    type: Literal["conversation.item.added"]
    """The event type, must be `conversation.item.added`."""

    previous_item_id: Optional[str] = None
    """The ID of the item that precedes this one, if any.

    This is used to maintain ordering when items are inserted.
    """



ConversationItem: TypeAlias = Annotated[
    Union[
        RealtimeConversationItemSystemMessage,
        RealtimeConversationItemUserMessage,
        RealtimeConversationItemAssistantMessage,
        RealtimeConversationItemFunctionCall,
        RealtimeConversationItemFunctionCallOutput,
        RealtimeMcpApprovalResponse,
        RealtimeMcpListTools,
        RealtimeMcpToolCall,
        RealtimeMcpApprovalRequest,
    ],
    PropertyInfo(discriminator="type"),
]
```

StianHanssen (approved):
@longcw I tested the solution both with the test suit and manual tests following patterns I know produce the issue. I was unable to trigger the issue I have been observing.

As noted in discussions above, I would encourage a defensive warning log to capture if we miss anything with the placeholder not finding the previous ID to be the most recent in context.

That said, it seems to work nicely! I approve of this PR!

theomonnom (approved):

Issue #5115, opened 2026-03-16
Mistral plugin is broken with the latest mistralai 2.0 release


csanz91:
### Bug Description

Mistral recently released version 2.0 of their Python SDK. This release includes several breaking changes that cause the current LiveKit Mistral plugin to fail. The plugin needs to be updated to support the new API client and method signatures introduced in v2.0.

### Expected Behavior

The llm inference should work without crashing

### Reproduction Steps

```bash
1. uv add "livekit-agents[mistralai]~=1.4"
2. Run the mistral example

The code will fail with:
ImportError: cannot import name 'ChatCompletionStreamRequestMessagesTypedDict' from 'mistralai' (unknown location)
```

### Operating System

macOS

### Models Used

mistral-large-2512

### Package Versions

```bash
livekit-plugins-mistralai==1.4.5
livekit==1.1.2
livekit-agents==1.4.5
```

### Proposed Solution

```python
We should either lock the mistralai package to mistralai<2.0.0 or refactor the plugin to make it work with the latest library.
```

Pauldevillers:
Hi @davidzhao, I upgraded Mistral plugging with our new SDKV2

https://github.com/livekit/agents/pull/5163

PR #5116, opened 2026-03-16
Restrict mistralai dependecy to use v1 sdk

csanz91:
Mistral library introduced the version 2 of its python library with breaking changes that are currently not supported.

This fixes: #5115


davidzhao:
thanks!

PR #5121, opened 2026-03-16
Improve gpt-realtime token counting/Fix cost tracking via OTEL spans 

bml1g12:
**Context**: I realised that our gpt-realtime cost estimates shown in LangFuse (based on token counts) were almost double what we were being billed by OpenAI and investigated why

**Cause**: The cause was that OpenAI reports total token count _inclusive_ of cached tokens. LangFuse can only calculate costs by multiplying a span attribute by a $ price, which means currently it is impossible to calculate the correct price as for that we would need the uncached token count as a raw attribute. This is a known limitation currently: https://github.com/langfuse/langfuse/issues/10592

**Proposed Solution in this PR:** We should specify an attribute which is the uncached token count which can optionally be used in cost calculation post processsing.

The problem is that prior to this PR `input_text_tokens` attribute was inclusive of the cached token count, meaning it got double counted in the eventual cost calculation. So this PR makes two changes:

- Fix: Set `input_text_tokens` and `input_audio_tokens` to mean the uncached portion of text tokens, which is how LangFuse actually interprets it
- Improvement: Split input_cached_tokens into input_cached_text_tokens and input_cached_audio_tokens as gpt-realtime distinguishes these in their metrics (albeit currently priced the same)

**Example of usage**

LangFuse themselves provide model<>cost definition like the below screenshot:

image

Or users can make their own, but after this PR, the `input_text_tokens` correctly represents the input _uncached_  tokens, whereas before this PR it incorrectly included cached tokens. 

Note: I focus on LangFuse here, but this PR would likely be useful for any provider that does cost calculations via a single operation: taking token counts and multipying it by the cost, so I think its generally useful for even non-langfuse users.

### DIFF
--- a/livekit-agents/livekit/agents/telemetry/trace_types.py
+++ b/livekit-agents/livekit/agents/telemetry/trace_types.py
@@ -93,3 +93,4 @@ ATTR_EXCEPTION_MESSAGE = "exception.message"
 
 # Platform-specific attributes
 ATTR_LANGFUSE_COMPLETION_START_TIME = "langfuse.observation.completion_start_time"
+ATTR_LANGFUSE_USAGE_DETAILS = "langfuse.observation.usage_details"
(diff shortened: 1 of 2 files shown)


bml1g12:
lets hold off merging this until langfuse reply here https://github.com/langfuse/langfuse/issues/10592#issuecomment-4098119872

bml1g12:
Replaced by https://github.com/livekit/agents/pull/5447

PR #5128, opened 2026-03-17
feat(llm): add ChatContext.remove() helper

### LINKED ISSUE #5085: Add remove_by_id helper to ChatContext
### Feature Type

Would make my life easier

## Problem

`ChatContext` exposes helpers like:

- `get_by_id()`
- `index_by_id()`

but it does not provide a public way to remove an item by id.

As a result, users often need to manipulate the private `_items` list directly:

```python
if idx is not None:
    ctx._items.pop(idx)
````

Accessing `_items` directly relies on internal implementation details and may
break if the internal structure changes.

---

## Use case

When building conversational agents that dynamically modify the prompt
(e.g. graph-based dialogue orchestration), it is sometimes necessary to
replace or remove a specific item in the context (for example a dynamically
updated system instruction message).

Without a public helper, this requires mutating the private `_items`
attribute.

---

## Proposal

Add a helper method:

```python
def remove_by_id(self, item_id: str) -> ChatItem | None:
```

Behavior:

* removes the **first item** matching the given id
* returns the removed `ChatItem`
* returns `None` if no item with that id exists

Returning the removed item follows common Python patterns like
`dict.pop()` and `list.pop()` and allows callers to inspect or reuse
the removed item if needed.

This would also align nicely with the existing helpers:

* `get_by_id`
* `index_by_id`

---

## Contribution

Since `ChatContext` already exposes `insert()` , `add_message()` `get_by_id()` and `index_by_id()`, a `remove_by_id()` helper would fit naturally alongside them. I'd be happy to open a PR implementing it if this approach makes sense.

Currently the workaround is to use `index_by_id()` and then remove the item
directly from the private `_items` list:

```python
if idx is not None:
    ctx._items.pop(idx)
````

This works, but it requires accessing the private `_items` attribute,
which relies on internal implementation details and could break if the
internal structure changes in the future.

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by 3eid**
## Summary

Adds a public `remove()` method to `ChatContext` for removing items by ID or instance, avoiding the need to access the private `_items` list directly.

## Problem

`ChatContext` exposes helpers like `get_by_id()` and `index_by_id()` but provides no public way to remove an item. As a result, users are forced to manipulate the private `_items` attribute directly:

```python
if idx is not None:
    ctx._items.pop(idx)
```

This relies on internal implementation details and may break if the internal structure changes.

## Fix

Adds a `remove()` method that accepts a single `ChatItem`, an item ID (`str`), or a sequence of either — consistent with the suggestion from @tinalenguyen in the linked issue. Returns a list of the removed `ChatItem`s (empty list if none were found), following the same pattern as `dict.pop()` and aligning with the existing `insert()` helper.

```python
# Remove by ID
ctx.remove(item_id)

# Remove by ChatItem instance
ctx.remove(item)

# Remove multiple at once
ctx.remove([item_id_1, item_2])
```

## Testing

Added 7 unit tests to `tests/test_chat_ctx.py`:
- `test_remove_by_id` — removes a single item by ID
- `test_remove_by_item` — removes a single item by `ChatItem` instance
- `test_remove_sequence` — removes a sequence of IDs
- `test_remove_sequence_of_items` — removes a sequence of `ChatItem` instances
- `test_remove_sequence_mixed` — removes a mix of `ChatItem` instances and ID strings
- `test_remove_sequence_partial_match` — skips nonexistent IDs silently, removes only matches
- `test_remove_nonexistent_returns_empty` — returns `[]` when ID is not found

Closes #5085


3eid:
I’ll close this and reopen a clean PR with a better approach.

PR #5131, opened 2026-03-17
Add `remove()` method to ChatContext for flexible item removal

3eid:
**Summary**
Adds a public `remove()` method to `ChatContext` for removing items by ID, by `ChatItem` instance, or a sequence of either. This avoids the need to access the private `_items` list directly.

**Problem**
`ChatContext` already provides helpers like `get_by_id()` and `index_by_id()`, but no public way exists to remove an item. Users currently must manipulate `_items` directly:

```python
if idx is not None:
    ctx._items.pop(idx)
```

Accessing `_items` relies on internal implementation details and can break if the structure changes.

**Fix**
Adds a `remove()` method that:

* Accepts a single `ChatItem`, an item ID (`str`), or a sequence of either.
* Returns a list of removed items (empty if none were found).

This follows the suggestion from @tinalenguyen to allow more flexible usage and aligns with existing helpers like `insert()`.

**Usage Examples**

```python
# Remove by ID
ctx.remove(item_id)

# Remove by ChatItem instance
ctx.remove(item)

# Remove multiple at once
ctx.remove([item_id_1, item_2])
```

**Testing**
Added comprehensive unit tests covering:

* Single removal by ID or `ChatItem`
* Removal of sequences (IDs, instances, or mixed)
* Partial matches and nonexistent IDs
* Removal from empty context
* Special cases like `AgentConfigUpdate` items

**Closes** #5085

### DIFF
--- a/livekit-agents/livekit/agents/llm/chat_context.py
+++ b/livekit-agents/livekit/agents/llm/chat_context.py
@@ -408,6 +408,16 @@ class ChatContext:
             idx = self.find_insertion_index(created_at=_item.created_at)
             self._items.insert(idx, _item)
 
+    def remove(self, item: ChatItem | str) -> None:
+        """Remove the first item from the chat context by ChatItem or item ID.
+
+        Raises ValueError if the item/ID is not found.
+        """
+        idx = self.index_by_id(item.id if not isinstance(item, str) else item)
+        if idx is None:
+            raise ValueError(f"Item not found: {item!r}")
+        self._items.pop(idx)
+
     def get_by_id(self, item_id: str) -> ChatItem | None:
         return next((item for item in self.items if item.id == item_id), None)
 
--- a/tests/test_chat_ctx.py
+++ b/tests/test_chat_ctx.py
@@ -327,6 +327,29 @@ def test_truncate_multiple_instructions():
     assert ctx.items[0].content == ["first"]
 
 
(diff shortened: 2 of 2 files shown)


theomonnom:
Hey, this is great, but can we have a similar API to list.remove? Like https://docs.python.org/3/tutorial/datastructures.html#more-on-lists

3eid:
> Hey, this is great, but can we have a similar API to list.remove? Like https://docs.python.org/3/tutorial/datastructures.html#more-on-lists

Good idea.

To stay close to `list.remove`, I’ll make `remove()` operate on a single item and raise if not found.

For the input, I’m leaning toward supporting both `ChatItem` and `str` (ID), since IDs are already a first-class way to reference items in ChatContext (e.g. get_by_id, index_by_id).

Internally it would just match by item.id in both cases.

Let me know if you’d prefer restricting it to ChatItem only.

3eid:
> Hey, this is great, but can we have a similar API to list.remove? Like https://docs.python.org/3/tutorial/datastructures.html#more-on-lists

Hey @theomonnom , just following up.

I've updated the method to align with list.remove() semantics (raises ValueError if not found, single item only). 
Let me know if this looks good or if you'd like any changes!

longcw (approved):
looks good to me!

PR #5134, opened 2026-03-18
fix: strip empty {} entries from anyOf/oneOf in strict JSON schema (#5130)

### LINKED ISSUE #5130: `_ensure_strict_json_schema` should handle empty `{}` entries in `anyOf`/`oneOf`
### Bug Description

When a Pydantic model field is typed as `Union[Literal["a", "b"], Any]` (common pattern in codegen clients like Fern for forward-compatible "open enums"), `model_json_schema()` produces an anyOf containing a bare `{}`:

```
"anyOf": [
    {"enum": ["a", "b"], "type": "string"},
    {},
    {"type": "null"}
]
```

`_ensure_strict_json_schema` recurses into this but doesn't strip or transform the `{}` entry. OpenAI's strict mode rejects schemas containing `{}` because it lacks a type field, so any `FunctionTool` whose signature includes a model with a `Union[..., Any]` field fails at the API level

### Expected Behavior

`_ensure_strict_json_schema` should strip bare `{}` entries from `anyOf`/`oneOf` arrays. `{}` is the JSON Schema identity element for `anyOf` (matches anything), so removing it is semantically neutral and makes the schema strict-mode compatible

### Reproduction Steps

```bash
import typing
import json
from pydantic import BaseModel
from livekit.agents.llm import _strict

# Simulates a codegen'd "open enum" (e.g. Fern Python SDK)
OpenEnum = typing.Union[typing.Literal["a", "b"], typing.Any]

class Inner(BaseModel):
    value: str
    preference: typing.Optional[OpenEnum] = None

class Outer(BaseModel):
    items: typing.List[Inner]

schema = _strict.to_strict_json_schema(Outer)
print(json.dumps(schema, indent=2))

# The "preference" field's anyOf still contains {}:
# "anyOf": [{"enum": ["a", "b"], "type": "string"}, {}, {"type": "null"}]
# OpenAI strict mode rejects this
```

### Operating System

macOS Sequoia (also reproducible on Linux)

### Models Used

Any OpenAI model with strict function calling (tested with gpt-4.1-mini)

### Package Versions

```bash
livekit-agents==1.6.2
pydantic==2.11.1
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
# In _strict.py, before the existing anyOf/oneOf simplification block (~line 150),
# strip empty schema objects:

for union_key in ("anyOf", "oneOf"):
    variants = json_schema.get(union_key)
    if is_list(variants):
        filtered = [v for v in variants if v != {}]
        if len(filtered) < len(variants):
            if len(filtered) == 1:
                json_schema.update(filtered[0])
                json_schema.pop(union_key, None)
            else:
                json_schema[union_key] = filtered
```

### Additional Context

This affects anyone using Fern, Speakeasy, or similar API client generators that emit `Union[Literal[...], Any]` as an "open enum" pattern for forward compatibility. The `Any` arm is intentional on their side (graceful handling of unknown enum values from the server), but it poisons downstream JSON Schema generation

_No response_

guoyangzhen:
## Summary

Fixes #5130

When a Pydantic model field is typed as `Union[Literal["a", "b"], Any]` (common pattern in codegen clients like Fern for forward-compatible "open enums"), `model_json_schema()` produces an `anyOf` containing a bare `{}` entry:

```json
"anyOf": [
    {"enum": ["a", "b"], "type": "string"},
    {},
    {"type": "null"}
]
```

OpenAI's strict mode rejects schemas containing `{}` because it lacks a type field. This causes any `FunctionTool` whose signature includes a model with a `Union[..., Any]` field to fail at the API level.

## Root Cause

`_ensure_strict_json_schema` recurses into `anyOf`/`oneOf` variants but doesn't strip or transform empty `{}` entries. Pydantic generates these for `Any` types in unions.

## Fix

Strip empty schema objects (`{}`) from `anyOf` and `oneOf` arrays before recursing. `{}` is the JSON Schema identity element for `anyOf` (matches anything), so removing it is semantically neutral and makes the schema strict-mode compatible.

## Changes

- **`livekit-agents/livekit/agents/llm/_strict.py`**: Filter out `{}` entries before processing `anyOf`/`oneOf` variants
- **`tests/test_tools.py`**: Add `TestEmptySchemaStripping` test class with test cases for:
  - Single model with `Union[Literal, Any]` field
  - Nested model containing open enum model
  - Validation that output schema contains no empty objects

## Verification

```python
from pydantic import BaseModel
from typing import Literal, Union, Any
from livekit.agents.llm._strict import to_strict_json_schema

class OpenEnum(BaseModel):
    preference: Union[Literal["a", "b"], Any, None] = None

schema = to_strict_json_schema(OpenEnum)
# Before fix: anyOf contains {} at index 1 → OpenAI rejects
# After fix: {} stripped → schema is strict-mode compatible
```

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -1,5 +1,7 @@
 import enum
-from typing import Literal
+import json
+import typing
+from typing import Any, Literal, Union
 
 import pytest
 from pydantic import BaseModel, Field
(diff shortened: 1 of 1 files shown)


theomonnom:
Hey, lgtm, can you fix the ruff CI?

theomonnom:
Continued here: https://github.com/livekit/agents/pull/5137

PR #5139, opened 2026-03-18
fix(11labs): add empty keepalive message and remove final duplicates

### LINKED ISSUE #4609: ElevenLabs STT stream does not reconnect after mid-stream WebSocket disconnection
### Bug Description

When using `elevenlabs.STT` (version `1.3.6`) in streaming mode, if the WebSocket connection drops mid-stream (e.g., due to network instability or ElevenLabs service issues), the `SpeechStream` raises `APIStatusError` with `retryable=True` but does not attempt to reconnect. The stream effectively dies, and my transcription becomes non-English characters. I also see LiveKit start repeating the same sentence (I am not sure this repeat behavior is related to stream issue).

Current Behavior
------

1. `SpeechStream._run()` establishes WebSocket connection via `_connect_ws()`
2. `recv_task` monitors incoming messages
3. When WebSocket closes unexpectedly, `recv_task` raises: 
    ```
    APIStatusError: ElevenLabs STT connection closed unexpectedly (status_code=-1, request_id=None, body=None, retryable=True)
    ```
4. Exception propagates through `task.result()` and breaks out of the `while True` loop
5. Stream ends without reconnection attempt
6. AgentSession receives error with `recoverable=True`

### Expected Behavior

Either:

- Plugin-level: `SpeechStream` should catch connection errors and attempt reconnection (similar to how `_reconnect_event` triggers reconnection for `update_options`)

- Framework-level: `AgentSession` should recreate STT streams when receiving recoverable STT errors

Relevant Code
--------------
In `stt.py`, the `_run` method's reconnection logic only triggers on `_reconnect_event`:

```
async def _run(self) -> None:
    while True:
        try:
            ws = await self._connect_ws()
            # ... tasks created ...
            
            for task in done:
                if task != wait_reconnect_task:
                    task.result()  # <-- Exception raised here breaks out of loop
            
            if wait_reconnect_task not in done:
                break  # <-- Exits loop on error
            
            self._reconnect_event.clear()  # <-- Only reconnects if event was set
```

### Reproduction Steps
1. Configure `AgentSession` with `elevenlabs.STT`, set `realtime` to `True` to use `Scribe V2`
2. Start a voice conversation
3. Just talk with agent for a few minutes in LiveKit playground
4. Observe error in logs with `retryable=True`, and my transcription becomes non-English characters.

This is how I create a session with ElevenLabs STT:
```python


from livekit.plugins import elevenlabs,


params = {"use_realtime": True}
stt = elevenlabs.STT(**params)
...
session = AgentSession[ParticipantData](
            userdata=participant_data,
            stt=stt,    # scribe v2
            llm=llm,   # OpenAI "gpt-5" model
            tts=tts,    # eleven_multilingual_v2
)
```

### Operating System

MacOS

### Models Used

Scribe V2

### Package Versions

```bash
livekit-agents~=1.3.6
livekit-api~=1.1.0
livekit-plugins-elevenlabs~=1.3.6

Python: 3.11
```

Image

chenghao-mou:
- 11labs disconnects when no audio is sent for about 10s;
- It also sends committed_transcript_with_timestamps even when include_timestamps is set to False;

This fixes #4609. Thanks to @tonydialpad for the detailed reporting.

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
@@ -357,8 +357,18 @@ class SpeechStream(stt.SpeechStream):
         async def keepalive_task(ws: aiohttp.ClientWebSocketResponse) -> None:
             try:
                 while True:
-                    await ws.ping()
-                    await asyncio.sleep(30)
+                    # scribe_v2_realtime model requires a keepalive message instead of a ping
+                    await asyncio.sleep(10)
+                    await ws.send_str(
+                        json.dumps(
+                            {
+                                "message_type": "input_audio_chunk",
+                                "audio_base_64": "",
+                                "commit": False,
+                                "sample_rate": self._opts.sample_rate,
+                            }
+                        )
+                    )
             except Exception:
                 return
 
@@ -565,10 +575,10 @@ class SpeechStream(stt.SpeechStream):
                 )
                 self._event_ch.send_nowait(interim_event)
 
-        # 11labs sends both when include_timestamps is True
-        elif (
-            message_type == "committed_transcript" and not self._opts.include_timestamps
-        ) or message_type == "committed_transcript_with_timestamps":
+        # 11labs sends both when include_timestamps is True or when the model is scribe_v2_realtime :(
+        elif (message_type == "committed_transcript" and not self._opts.include_timestamps) or (
+            message_type == "committed_transcript_with_timestamps" and self._opts.include_timestamps
+        ):
             # Final committed transcripts - these are sent to the LLM/TTS layer in LiveKit agents
             # and trigger agent responses (unlike partial transcripts which are UI-only)
             if text:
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -357,8 +357,18 @@ async def _run(self) -> None:
...
                 while True:
-                    await ws.ping()
-                    await asyncio.sleep(30)
+                    # scribe_v2_realtime model requires a keepalive message instead of a ping
+                    await asyncio.sleep(10)
+                    await ws.send_str(
+                        json.dumps(
+                            {
+                                "message_type": "input_audio_chunk",
+                                "audio_base_64": "",
```
theomonnom, line 366: is empty keepalive?

chenghao-mou: I don't see any errors after having this, so I think so.

PR #5142, opened 2026-03-18
fix(bargein): properly fallback to VAD when there is an error

chenghao-mou:
- Added LIVEKIT_DEV_MODE env so that bargein is disabled by default for _self-hosted_ production env
- Added two utility functions for env checking: is_hosted and is_dev_mode
- Fallback to VAD when barge in hits an unretryable error or timeout (700ms) in both HTTP and WS streams
- Clean up some more event names

### DIFF
--- a/livekit-agents/livekit/agents/inference/interruption.py
+++ b/livekit-agents/livekit/agents/inference/interruption.py
@@ -1140,7 +1140,14 @@ class InterruptionWebSocketStream(InterruptionStreamBase):
         ) as e:
             if isinstance(e, aiohttp.ClientResponseError) and e.status == 429:
                 raise APIStatusError(
-                    "LiveKit Adaptive Interruption quota exceeded", status_code=e.status
+                    "LiveKit Adaptive Interruption quota exceeded",
+                    status_code=e.status,
+                    retryable=False,
+                ) from e
+            elif isinstance(e, asyncio.TimeoutError):
+                raise APIConnectionError(
+                    "failed to connect to LiveKit Adaptive Interruption: timeout",
+                    retryable=False,
                 ) from e
             raise APIConnectionError("failed to connect to LiveKit Adaptive Interruption") from e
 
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/cli/cli.py
```diff
@@ -1663,6 +1663,11 @@ class LogLevel(str, enum.Enum):
 def _build_cli(server: AgentServer) -> typer.Typer:
     app = typer.Typer(rich_markup_mode="rich")
 
+    @app.callback(invoke_without_command=True)
+    def _set_dev_mode(ctx: typer.Context) -> None:
+        if ctx.invoked_subcommand in ("console", "dev", "connect"):
```
theomonnom:
```suggestion
        if ctx.invoked_subcommand in ("console", "dev"):
```

livekit-agents/livekit/agents/inference/interruption.py
```diff
@@ -713,10 +715,7 @@ def update_options(
         if is_given(min_interruption_duration):
             self._opts.min_frames = math.ceil(min_interruption_duration * _FRAMES_PER_SECOND)
 
-    @log_exceptions(logger=logger)
     async def _run(self) -> None:
-
-        @log_exceptions(logger=logger)
```
theomonnom, line 719: Are we sure we're logging logs?

chenghao-mou: It is caught in main_task, so we don't log it multiple times.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -3169,6 +3193,14 @@ def _resolve_interruption_detection(self) -> inference.AdaptiveInterruptionDetec
         ):
             return None
 
+        if (
```
davidzhao, line 3203: I think this logic disables interruption handling even on LK Cloud?

it should also check to see if it's running on LKC (using `REMOTE_EOT_URL` seems reasonable)

chenghao-mou: Yes. Forgot to push the commit. It's checking that env now with a new function utils.is_hosted()

PR #5147, opened 2026-03-18
Fix Phonic `generate_reply` to resolve with the current `GenerationCreatedEvent`

qionghuang6:
Previously, the Phonic plugin's `generate_reply` immediately creates a dummy generation and resolves the returned future before the request immediately,

This means that when we do `await session.generate_reply(instructions="Greet the user, asking about their day.")`, this would actually finish at the wrong time, since with the other plugins, this resolves when the reply speech actually finishes (or if it were interrupted by user speech).

This change makes it rather than making a dummy generation, we resolve the `generate_reply` future the next time `assistant_started_speaking` is received from Phonic and we make a new generation. 

We can repro this behavior when doing something like:
```
    logger.info(f"Generating reply, {time.time()}")
    await session.generate_reply(
        instructions="Greet the user, asking about their day.",
    )
    logger.info(f"Reply generated, {time.time()}")
```

## Please see the video demo for a repro:
https://screen.studio/share/00Yuk0vN?state=uploading

### DIFF
--- a/livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
@@ -443,6 +443,10 @@ class RealtimeSession(llm.RealtimeSession):
 
         self._close_current_generation(interrupted=False)
 
+        if self._pending_generate_reply_fut and not self._pending_generate_reply_fut.done():
+            self._pending_generate_reply_fut.cancel()
+            self._pending_generate_reply_fut = None
+
         if self._generate_reply_task and not self._generate_reply_task.done():
             await utils.aio.cancel_and_wait(self._generate_reply_task)
 


theomonnom:
If we send multiple generate_reply, are we sure the future is picking the right one? It isn't really synchronized with the server?

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -432,6 +443,10 @@ async def aclose(self) -> None:
 
         self._close_current_generation(interrupted=False)
 
+        if self._pending_generate_reply_fut and not self._pending_generate_reply_fut.done():
+            self._pending_generate_reply_fut.cancel()
```
theomonnom, line 447: E.g the future here is cancelled, but nothing is cancelled on the serverside right?

qionghuang6: Phonic currently doesn't explicitly associate any audio chunks with a certain `generate_reply` message, but any `generate_reply` message received over the WebSocket will cease any existing audio generation and begin a new one, so the next `assistant_started_speaking` message will correspond to that of the latest `generate_reply` message. 

I think in this case I'm just cancelling the future so that it doesn't finish with an exception.

Issue #4402, opened 2025-12-26
Feature Request: Add Original Language Detection to Gladia STT Plugin


mowtschan:
### Summary
The Gladia STT plugin currently exposes `target_language` from translation data but does not expose the `original_language` (source language) that was detected/used during transcription. This information would be valuable for multilingual applications and language-aware workflows. 

### Current Behavior
The plugin currently handles translation data and sets the target language:
```python
target_language = translation_data.get("target_language", "")
language = translated_utterance.get("language", target_language)
...
if translated_text and language:
    speech_data = stt.SpeechData(
        language=language,  # Use the target language
        ...
```

The `SpeechData.language` field is already being used to store the `target_language` for translated text. However, the `original_language` (the language that was actually spoken/detected) is not being captured or exposed anywhere.

### Problem
When translation is enabled, we lose information about what language the user was actually speaking. The `language` field in `SpeechData` contains the target language (what the text was translated to), but there's no way to access the original/source language (what was actually spoken).
**This becomes critical when using multiple `translation_target_languages`.** 
According to the documentation, when multiple target languages are specified, the plugin emits a separate transcription event for each language. Without the original language information, it becomes impossible to distinguish which translations came from which source language.

### Proposed Solution
Add support for capturing and exposing the `original_language` from Gladia's transcription/translation response.

Store in `speaker_id` field
Since `SpeechData.language` is already taken by the target language, the `speaker_id` field could potentially be used to store the original language information:

```python
...
original_language = translation_data.get("original_language", "")
...
if translated_text and language:
    speech_data = stt.SpeechData(
        ...
        speaker_id=original_language
        ...
```
**Concerns**: This feels like a misuse of the `speaker_id` field ;-(

tinalenguyen:
Hi, thanks for the detailed request! I'm able to retrieve the originally spoken language in `SpeechData` when iterating through the `stt_node`. Could you share how you were trying to access it for more context?

For reference, I override the STT node like so:
```
async def stt_node(self, audio: AsyncIterable[rtc.AudioFrame], model_settings: ModelSettings) -> AsyncIterable[stt.SpeechEvent]:
        async def event_stream():
            async for event in Agent.default.stt_node(self, audio, model_settings):
                print(event)
                yield event

        return event_stream()
```

And I get a stream of `SpeechEvent`, with the language in the `alternatives` field being the detected spoken one and not the target: 
[...]

PR #5148, opened 2026-03-18
(gladia & soniox): add translation support

### LINKED ISSUE #4943: Add Soniox Real-Time Translation Support to livekit-plugins-soniox
### Feature Type

## Description

The Soniox Speech-to-Text API supports real-time translation (both one-way and two-way) as a first-class feature on their WebSocket streaming endpoint. However, the LiveKit Soniox plugin (`livekit-plugins-soniox`) does not expose this capability at all. There are three specific gaps in the current implementation:

### 1. `STTOptions` has no `translation` field

The `STTOptions` dataclass defines configuration for model, language hints, speaker diarization, etc., but has no field for the `translation` config object that the Soniox API accepts. Users cannot pass translation parameters through the plugin.

**Current `STTOptions`:**

```python
@dataclass
class STTOptions:
    model: str = "stt-rt-v3"
    language_hints: list[str] | None = None
    language_hints_strict: bool = False
    context: ContextObject | str | None = None
    num_channels: int = 1
    sample_rate: int = 16000
    client_reference_id: str | None = None
    # ❌ No translation field
```

### 2. `_connect_ws` does not include `translation` in the WebSocket config

The `SpeechStream._connect_ws` method builds a config dict and sends it as the first WebSocket message. It includes all other `STTOptions` fields but never adds a `translation` key. Even if a user somehow attached a `translation` field to the options, it would be silently dropped.

**Current config dict in `_connect_ws`:**

```python
config: dict[str, Any] = {
    "api_key": self._stt._api_key,
    "model": self._stt._params.model,
    "audio_format": "pcm_s16le",
    "num_channels": self._stt._params.num_channels or 1,
    "enable_endpoint_detection": True,
    "sample_rate": self._stt._params.sample_rate,
    "language_hints": self._stt._params.language_hints,
    "language_hints_strict": self._stt._params.language_hints_strict,
    "context": context_value,
    "enable_speaker_diarization": self._stt._params.enable_speaker_diarization,
    "enable_language_identification": self._stt._params.enable_language_identification,
    "client_reference_id": self._stt._params.client_reference_id,
    # ❌ No translation key
}
```

### 3. `_recv_messages_task` ignores `translation_status` on tokens

When translation is enabled, the Soniox API returns tokens with a `translation_status` field that can be `"none"`, `"original"`, or `"translation"`. It also includes `language` and `source_language` metadata on each token. The current `_recv_messages_task` method **completely ignores** this field and concatenates all tokens (original + translation) into a single `final_transcript_buffer`. This produces garbled output where the original text and its translation are merged together into one string.

**Current token processing (line-by-line):**

```python
for token in tokens:
    if token["is_final"]:
        if is_end_token(token):
            send_endpoint_transcript()
            ...
        else:
            # ❌ Blindly appends ALL tokens regardless of translation_status
            final_transcript_buffer += token["text"]
            ...
    else:
        # ❌ Same issue for interim tokens
        non_final_transcription += token["text"]
```

**What happens in practice:** If a user says "Good morning" and the target translation is German, the Soniox API returns tokens like:

```
{text: "Good", translation_status: "original", language: "en"}
{text: " morning", translation_status: "original", language: "en"}
{text: "Guten", translation_status: "translation", language: "de", source_language: "en"}
{text: " Morgen", translation_status: "translation", language: "de", source_language: "en"}
```

The plugin currently produces: `"Good morning Guten Morgen"` — a single merged `FINAL_TRANSCRIPT` event. There is no way for the consumer to distinguish original text from translated text.

---

## Soniox Translation API Reference

The Soniox WebSocket API accepts a `translation` object in the initial config message:

**One-way** (translate all speech into a single target language):

```json
{
  "translation": {
    "type": "one_way",
    "target_language": "fr"
  }
}
```

**Two-way** (translate back and forth between two specified languages):

```json
{
  "translation": {
    "type": "two_way",
    "language_a": "en",
    "language_b": "de"
  }
}
```

Each token in the response includes:

| Field                | Description                                                    |
| -------------------- | -------------------------------------------------------------- |
| `text`               | Token text                                                     |
| `translation_status` | `"none"` (not translated), `"original"` (spoken), `"translation"` (translated) |
| `language`           | Language of the token (e.g. `"en"`, `"de"`)                    |
| `source_language`    | Original language (only present on `"translation"` tokens)     |

Full documentation: https://soniox.com/docs/stt/rt/real-time-translation

---

### Step 1: Add a `TranslationConfig` dataclass and a `translation` field to `STTOptions`

```python
@dataclass
class TranslationConfig:
    """Configuration for Soniox real-time translation.

    See: https://soniox.com/docs/stt/rt/real-time-translation
    """
    type: str  # "one_way" or "two_way"
    target_language: str | None = None   # required for one_way
    language_a: str | None = None        # required for two_way
    language_b: str | None = None        # required for two_way


@dataclass
class STTOptions:
    model: str = "stt-rt-v3"
    language_hints: list[str] | None = None
    language_hints_strict: bool = False
    context: ContextObject | str | None = None
    num_channels: int = 1
    sample_rate: int = 16000
    client_reference_id: str | None = None
    translation: TranslationConfig | None = None  # ✅ NEW
```

### Step 2: Include `translation` in the WebSocket config in `_connect_ws`

After building the existing config dict, add:

```python
# In _connect_ws, after building the config dict:
if self._stt._params.translation is not None:
    config["translation"] = asdict(self._stt._params.translation)
```

This uses the existing `asdict` import that is already at the top of the file.

### Step 3: Update `_recv_messages_task` to separate original and translation tokens

This is the most impactful change. When translation is enabled, `_recv_messages_task` must bucket tokens by their `translation_status` and emit separate `SpeechEvent` objects so consumers can distinguish original text from translated text.

**Option A — Emit two separate `FINAL_TRANSCRIPT` events per endpoint (one for original, one for translation):**

Each `SpeechEvent.alternatives` already supports multiple `SpeechData` entries. The primary alternative (index 0) could be the original text, and a second alternative (index 1) could be the translation:

```python
for token in tokens:
    status = token.get("translation_status", "none")
    if token["is_final"]:
        if is_end_token(token):
            send_endpoint_transcript()
        elif status in ("original", "none"):
            final_original_buffer += token["text"]
            # ... capture language, speaker_id as before
        elif status == "translation":
            final_translation_buffer += token["text"]
            final_translation_language = token.get("language", "")
    else:
        if status in ("original", "none"):
            non_final_original += token["text"]
        elif status == "translation":
            non_final_translation += token["text"]
```

Then in `send_endpoint_transcript()`:

```python
def send_endpoint_transcript():
    alternatives = []
    if final_original_buffer:
        alternatives.append(
            stt.SpeechData(
                text=final_original_buffer,
                language=final_original_language,
                speaker_id=final_speaker_id,
            )
        )
    if final_translation_buffer:
        alternatives.append(
            stt.SpeechData(
                text=final_translation_buffer,
                language=final_translation_language,
            )
        )
    if alternatives:
        event = stt.SpeechEvent(
            type=SpeechEventType.FINAL_TRANSCRIPT,
            alternatives=alternatives,
        )
        self._event_ch.send_nowait(event)
        self._event_ch.send_nowait(
            stt.SpeechEvent(type=SpeechEventType.END_OF_SPEECH)
        )
    # ... reset buffers
```

This way, consumers can check `len(event.alternatives)` and know that `alternatives[0]` is always the original and `alternatives[1]` (when present) is the translation. The `language` field on each `SpeechData` tells you what language it is in.

**Option B — Use a dedicated event type or metadata field:**

If the `alternatives` list approach is not desirable (since `alternatives` traditionally means "N-best hypotheses"), consider adding a `translation` field to `SpeechEvent` or `SpeechData`:

```python
@dataclass
class SpeechData:
    text: str
    language: str
    speaker_id: str | None = None
    source_language: str | None = None         # ✅ NEW: populated for translations
    translation_status: str | None = None      # ✅ NEW: "original", "translation", "none"
```

### Step 4: Export `TranslationConfig` from `__init__.py`

```python
from .stt import STT, STTOptions, ContextObject, TranslationConfig

__all__ = [
    "STT",
    "STTOptions",
    "ContextObject",
    "TranslationConfig",
]
```

---

## Usage After Implementation

```python
from livekit.plugins.soniox import STT, STTOptions, TranslationConfig

stt = STT(
    params=STTOptions(
        model="stt-rt-v4",
        language_hints=["en", "de"],
        enable_language_identification=True,
        translation=TranslationConfig(
            type="two_way",
            language_a="en",
            language_b="de",
        ),
    ),
)

async with stt.stream() as stream:
    # ... push audio frames ...
    async for event in stream:
        if event.type == SpeechEventType.FINAL_TRANSCRIPT:
            original = event.alternatives[0]  # spoken text
            if len(event.alternatives) > 1:
                translated = event.alternatives[1]  # translated text
                print(f"[{original.language}] {original.text}")
                print(f"[{translated.language}] {translated.text}")
```

tinalenguyen:
add `input_language` and `input_text` to `SpeechData` and add translation support for gladia and soniox

closes #4943 and #4402

to test, iterate through STT node:
```
async def stt_node(
        self, audio: AsyncIterable, model_settings: ModelSettings
    ) -> AsyncIterable[stt_module.SpeechEvent]:
        async for event in Agent.default.stt_node(self, audio, model_settings):
            if isinstance(event, stt_module.SpeechEvent) and event.alternatives:
                alt = event.alternatives[0]
                if alt.input_language:
                    logger.info(
                        f"[STT translation] input_language={alt.input_language}, "
                        f"language={alt.language}, "
                        f"input_text={alt.input_text!r}, text={alt.text!r}"
                    )
            yield event
```

### DIFF
--- a/livekit-plugins/livekit-plugins-gladia/livekit/plugins/gladia/stt.py
+++ b/livekit-plugins/livekit-plugins-gladia/livekit/plugins/gladia/stt.py
@@ -1119,7 +1119,7 @@ class SpeechStream(stt.SpeechStream):
                         source_languages=[LanguageCode(original_language)]
                         if original_language
                         else None,
-                        source_texts=[original_text] if original_text else None,
+                        source_texts=[original_text or ""] if original_language else None,
                         start_time=translated_utterance.get("start", 0) + self.start_time_offset,
                         end_time=translated_utterance.get("end", 0) + self.start_time_offset,
                         confidence=translated_utterance.get("confidence", 1.0),


livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -60,10 +60,21 @@ class SpeechData:
     speaker_id: str | None = None
     is_primary_speaker: bool | None = None
     words: list[TimedString] | None = None
+    input_language: LanguageCode | None = None
+    """the detected/input language spoken by the user. populated by STT services that support translation,
+    where `language` holds the target language and `input_language` holds the original spoken language"""
+    input_text: str | None = None
```
chenghao-mou: nit: borrowing terms from machine translation terminology, we should name them `source_language`, and `source_text`.

livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
```diff
@@ -388,10 +424,14 @@ def send_endpoint_transcript() -> None:
 
                 # Reset buffers.
                 final.reset()
+                final_original.reset()
 
                 # Reset speaking state, so the next transcript will send START_OF_SPEECH again.
                 is_speaking = False
+            else:
+                final_original.reset()
```
chenghao-mou, line 438: I don't think we need to reset this here

livekit-agents/livekit/agents/stt/stt.py
tinalenguyen: I renamed to `source_languages` and `source_texts`, for the polyglots

PR #5149, opened 2026-03-18
increase generate_reply timeout & remove RealtimeModelBeta

theomonnom:


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/__init__.py
```diff
@@ -1,11 +1,8 @@
 from .realtime_model import RealtimeModel, RealtimeSession
-from .realtime_model_beta import RealtimeModelBeta, RealtimeSessionBeta
 
 __all__ = [
     "RealtimeSession",
     "RealtimeModel",
-    "RealtimeModelBeta",
```
longcw, line 7: Should we mention this in the release notes?

Issue #5103, opened 2026-03-13
Inworld: Context not found


joshiayush:
### Bug Description

Hi,

I’m trying to implement Inworld tts in my agent, a week ago it was working perfectly as a POC but now when added to staging it’s showing the following error:

Inworld error: Context d6b3587d74a7 is not found

I’m attaching the full traceback here:

```
agent-1  |     09:02:56.123 WARNI… livekit.agents     failed to synthesize speech:
agent-1  |                                            Inworld error: Context 8a7fc93930c9
agent-1  |                                            not found, retrying in 2.0s
agent-1  |                                          {"tts":
agent-1  | "livekit.plugins.inworld.tts.TTS", "attempt": 2, "streamed": true, "interview":
agent-1  | "69a9692736b631a8c43e634f", "pid": 286, "job_id": "AJ_3CYyjWscBLGP", "room_id":
agent-1  | "RM_nsZxNH44h28H"}
agent-1  |     09:02:58.440 WARNI… livekit.agents     failed to synthesize speech:
agent-1  |                                            Inworld error: Context 722d505d52ff
agent-1  |                                            not found, retrying in 2.0s
agent-1  |                                          {"tts":
agent-1  | "livekit.plugins.inworld.tts.TTS", "attempt": 3, "streamed": true, "interview":
agent-1  | "69a9692736b631a8c43e634f", "pid": 286, "job_id": "AJ_3CYyjWscBLGP", "room_id":
agent-1  | "RM_nsZxNH44h28H"}
agent-1  |     09:02:59.728 WARNI… livekit.agents     process memory usage is high
agent-1  |                                          {"memory_usage_mb": 702.38671875,
agent-1  | "memory_warn_mb": 500, "memory_limit_mb": 0, "pid": 286, "job_id":
agent-1  | "AJ_3CYyjWscBLGP", "room_id": "RM_nsZxNH44h28H"}
agent-1  |     09:03:00.758 ERROR… livekit.agents     Error in _tts_inference_task
agent-1  |                                            Traceback (most recent call last):
agent-1  |                                              File
agent-1  |                                            "/app/.venv/lib/python3.13/site-pa…
agent-1  |                                            line 631, in __anext__
agent-1  |                                                val = await
agent-1  |                                            self._event_aiter.__anext__()
agent-1  |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^…
agent-1  |                                            StopAsyncIteration
agent-1  |
agent-1  |                                            During handling of the above
[...]

PR #5153, opened 2026-03-18
Fix(inworld): "Context not found" errors caused by invalid enum parameter types

ianbbqzy:
Passing non-string values (e.g. `text_normalization=True` instead of `"ON"`) silently produced malformed WebSocket messages that the API gateway rejected, while subsequent messages for the same context passed through — resulting in "Context not found" errors with no actionable logging.

**Changes:**
- Add runtime validation for `encoding`, `timestamp_type`, `text_normalization`, and `timestamp_transport_strategy` — raises `ValueError` immediately on invalid values
- Handle gateway-level error responses (`{"error": ...}`) in the WebSocket receive loop instead of silently dropping them
- Add logging for context lifecycle and error diagnostics
- Change default temperature to 1.0

Fixes https://github.com/livekit/agents/issues/5103


ianbbqzy:
The validation might be a bit paranoid. logging of the error message might be sufficient for developers to understand what's wrong

PR #5163, opened 2026-03-19
feat(mistralai): upgrade to SDK v2

Pauldevillers:
## Summary
  - Upgrade mistralai dependency from >=1.9.11,<2.0.0 to >=2.0.0
  - Update import paths for SDK v2 module reorganization
  - Rename ChatCompletionStreamRequestMessagesTypedDict to singular form

  No runtime API changes. Reverses the v1 lock from #5116.

  ## Test plan
  - [x] ruff format + ruff check pass
  - [x] mypy strict passes (0 errors in 6 files)
  - [x] Plugin imports cleanly
  - [x] Tested with live agent in console mode (Mistral LLM + STT working)

### DIFF
--- a/livekit-plugins/livekit-plugins-mistralai/pyproject.toml
+++ b/livekit-plugins/livekit-plugins-mistralai/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "hatchling.build"
 [project]
 name = "livekit-plugins-mistralai"
 dynamic = ["version"]
-description = "LiveKit Agents Plugin for services from AWS"
+description = "LiveKit Agents Plugin for services from Mistral AI"
 readme = "README.md"
 license = "Apache-2.0"
 requires-python = ">=3.10.0"


davidzhao:
@Pauldevillers should we update the default models here for LLM to `"ministral-8b-latest`?

Pauldevillers:
Hi @davidzhao, will recommend to choose latest version of 8B model with `ministral-8b-latest`, will follow up if/when we believe there is a better model for this use.

livekit-plugins/livekit-plugins-mistralai/livekit/plugins/mistralai/llm.py
davidzhao:
```suggestion
        model: str | ChatModels = "ministral-8b-latest",
```

livekit-plugins/livekit-plugins-mistralai/pyproject.toml
davidzhao:
```suggestion
description = "LiveKit Agents Plugin for services from Mistral AI"
```

PR #5168, opened 2026-03-19
fix azure openai realtime support & add realtime models tests

theomonnom:


sarthakgoyal23:
Hi @theomonnom After updating to 1.5.0 version where RealtimeBeta was removed we are facing this error

**body=RealtimeError(message="Unknown parameter: 'session.type'.**

With the changes in this PR will this be fixed?

Issue #5177, opened 2026-03-20
Participant spoken attribution always set to agent


Uldiniad:
### Bug Description

Spoken input from participants is attributed to the agent in chat. Text input is correctly attributed to the participant

### Expected Behavior

Spoken input from participants is attributed to the correct participant.

### Reproduction Steps

```bash
1. Configure an agent that transcribes user speech and publishes transcriptions 
2. On the frontend, receive transcriptions
3. Observe participant identity is always the agent's identity
```

### Operating System

Windows 11 (WSL 2)

### Models Used

AWS

### Package Versions

```bash
livekit==1.0.25
livekit-agents[aws]==1.5.0
livekit-plugins-noise-cancellation==0.2.5
livekit-plugins-openai==1.5.0
livekit-plugins-silero==1.5.0
livekit-plugins-turn-detector==1.5.0

livekit-client==2.17.3
@livekit/components-core==0.12.13
@livekit/components-react==2.9.20
@livekit/components-styles==1.2.0
```

### Session/Room/Call IDs

LiveKit Cloud

### Proposed Solution

```python

```

Uldiniad:
@cosmastech, any updates on this? know who to ping to have them take a look?

longcw:
@Uldiniad could you share a minimal example that can reproduce the issue?

Uldiniad:
I'm not seeing the issue after updating dependencies to the latest version. I'll reopen if I see the issue again

Uldiniad:
> I'm not seeing the issue after updating dependencies to the latest version. I'll reopen if I see the issue again

nvm, I was not testing properly. it's still present with latest version of dependencies too. but I see I cannot reopen the issue. @longcw mind doing that for me please?

Uldiniad:
it's only happening when e2ee is enabled

PR #5190, opened 2026-03-21
Store participant identity in _represented_by

cosmastech:
To resolve https://github.com/livekit/agents/issues/5177


Uldiniad:
I don't understand why the test is failing. That said, anything we can do to get this unblocked and reviewed?

PR #5191, opened 2026-03-21
Sarvam's WebSocket API decoder format mismatch

DharanHaarish:
## Summary
Fix mime_type mismatch in Sarvam TTS WebSocket streaming path. The `SynthesizeStream` class was declaring `audio/wav` but Sarvam's WebSocket API actually returns MP3 audio, causing the LiveKit WAV decoder to fail on MPEG bytes.

## Changes
- Changed `mime_type` from `"audio/wav"` to `"audio/mp3"` in `SynthesizeStream._run()`

## Notes
- The HTTP REST path (`ChunkedStream`) correctly handles WAV format
- The WebSocket path was incorrectly assuming the same format without verification
- This aligns with how other plugins (AWS, Resemble) handle MP3 streams


tinalenguyen:
hi, thank you for the pr! we will include the fix in https://github.com/livekit/agents/pull/5209

PR #5194, opened 2026-03-23
fix(tests): update test_recording to use renamed observability_url parameter

Lyt060814:
## Problem

6 tests in `tests/test_recording.py` fail on main because `_setup_cloud_tracer()` and `_upload_session_report()` were refactored to accept `observability_url` instead of `cloud_hostname`, but the tests still use the old parameter name:

```
TypeError: _setup_cloud_tracer() got an unexpected keyword argument 'cloud_hostname'
TypeError: _upload_session_report() got an unexpected keyword argument 'cloud_hostname'
```

Failing tests:
- `test_upload_returns_early_when_none`
- `test_upload_transcript_only`
- `test_upload_session_report_sent_without_transcript`
- `test_upload_audio_only_no_file`
- `test_upload_evaluations_emitted_without_logs`
- `test_setup_cloud_tracer_logger_provider_always_created`

## Fix

Update both call sites in `test_recording.py` to use `observability_url` instead of `cloud_hostname`.

### DIFF
--- a/tests/test_recording.py
+++ b/tests/test_recording.py
@@ -151,7 +151,7 @@ async def _call_upload(
     """Call _upload_session_report with sensible defaults."""
     await _upload_session_report(
         agent_name="test-agent",
-        cloud_hostname="test.livekit.cloud",
+        observability_url="test.livekit.cloud",
         report=report,
         tagger=tagger or _make_mock_tagger(),
         http_session=http_session or _make_mock_http(),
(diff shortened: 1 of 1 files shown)


Lyt060814:
Closing — upstream already addressed this more robustly with the `_observability_endpoint_arg()` helper in d4079b6 / e65ccadf. The dynamic signature check is a better approach than my static rename. 🎉

PR #5196, opened 2026-03-23
fix: expose Chirp 3 google STT endpoint sensitivity

karlsonlee-livekit:
Adds support for [Chirp 3's endpointing_sensitivity setting](https://docs.cloud.google.com/speech-to-text/docs/models/chirp-3#adjust_endpointing_sensitivity), which controls the trade-off between latency and accuracy when detecting end-of-speech.


karlsonlee-livekit:
image
This is the benchmark script that I ran to understand this a bit more

PR #5201, opened 2026-03-23
Adding xAI Grok llm support for inference

russellmartin-livekit:
https://github.com/livekit/agent-gateway/pull/478

### DIFF
--- a/livekit-agents/livekit/agents/inference/llm.py
+++ b/livekit-agents/livekit/agents/inference/llm.py
@@ -61,6 +61,7 @@ _UNSUPPORTED_PARAMS: dict[str, set[str]] = {
     "gpt-5": _REASONING_UNSUPPORTED_PARAMS,
     "grok-4-1-fast-reasoning": _XAI_REASONING_UNSUPPORTED_PARAMS,
     "grok-4.20-0309-reasoning": _XAI_REASONING_UNSUPPORTED_PARAMS,
+    "grok-4.20-multi-agent": _XAI_REASONING_UNSUPPORTED_PARAMS,
 }
 
 # models that don't support reasoning_effort when function tools are present
@@ -125,6 +126,7 @@ XAIModels = Literal[
     "xai/grok-4-1-fast-reasoning",
     "xai/grok-4.20-0309-non-reasoning",
     "xai/grok-4.20-0309-reasoning",
+    "xai/grok-4.20-multi-agent-0309",
 ]
 
 LLMModels = OpenAIModels | GoogleModels | KimiModels | DeepSeekModels | XAIModels


livekit-agents/livekit/agents/llm/_provider_format/openai.py
```diff
@@ -76,9 +74,8 @@ def _to_chat_item(msg: llm.ChatItem) -> dict[str, Any]:
                 "arguments": msg.arguments,
             },
         }
-        # Include provider-specific extra content (e.g., Google thought signatures)
-        if msg.extra.get("google"):
-            tc["extra_content"] = {"google": msg.extra["google"]}
+        if msg.extra:
+            tc["extra_content"] = msg.extra
```
theomonnom: Should we separate the providers tho? In the `extra_content`?

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -176,6 +177,9 @@ async def _llm_inference_task(
                         data.generated_functions.append(fnc_call)
                         function_ch.send_nowait(fnc_call)
 
+                if chunk.delta.extra:
+                    data.generated_extra.update(chunk.delta.extra)
```
theomonnom, line 181: They push extra for every chunk? How does it work?

russellmartin-livekit: It should only come on the last chunk (when `extra != none`), but it might come part of the way through on a tool call

livekit-agents/livekit/agents/llm/_provider_format/openai.py
russellmartin-livekit: I think keeping it generic here would be better, that way it doesn't need to change if we add more. They're already separated underneath `msg.extra`= 
```
{"google": {"thought_signature": "base64..."}}   # from Gemini
{"xai": {"encrypted_reasoning": "base64..."}}     # from Grok
```

theomonnom: oh the gateway is already separating them?

russellmartin-livekit: yeah it does it on the gateway side, does that make sense? Can do it either way

theomonnom: This function is also used by other plugins, so it really depends on what is the default behavior of Google and xAI

russellmartin-livekit: What do you mean by default behavior?

theomonnom: Adrian did what I meant here https://github.com/livekit/agents/pull/5374

theomonnom: The code inside livekit-agents/livekit/agents/llm/_provider_format/openai.py is used by the gateway but also by our plugins. So we can't assume extra_content is always going to be separated by providers

russellmartin-livekit: I see, also changed it for tool calls

davidzhao:
@russellmartin-livekit could you also handle JS?

PR #5202, opened 2026-03-23
(nvidia personaplex): detect EOT via volume thresholds

tinalenguyen:
default behavior: if 500ms of audio passes without meeting the minimum volume threshold, the agent turn is considered complete

### DIFF
--- a/livekit-plugins/livekit-plugins-nvidia/livekit/plugins/nvidia/experimental/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-nvidia/livekit/plugins/nvidia/experimental/realtime/realtime_model.py
@@ -43,6 +43,8 @@ DEFAULT_SILENCE_THRESHOLD_MS = 500
 MAX_RETRY_DELAY = 30.0
 INITIAL_RETRY_DELAY = 1.0
 
+_DEFAULT_SPEECH_VOLUME_THRESHOLD = 0.001
+
 
 @dataclass
 class _PersonaplexOptions:
@@ -51,6 +53,7 @@ class _PersonaplexOptions:
     text_prompt: str
     seed: int | None
     silence_threshold_ms: int
+    speech_volume_threshold: float
     use_ssl: bool = False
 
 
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-nvidia/livekit/plugins/nvidia/experimental/realtime/realtime_model.py
```diff
@@ -531,6 +537,10 @@ def _handle_audio_data(self, opus_payload: bytes) -> None:
             if pcm_float is None or len(pcm_float) == 0:
                 return
 
+            peak = float(np.abs(pcm_float).max())
+            if peak < self._opts.speech_volume_threshold:
+                return
```
theomonnom, line 542: How big is this frame, don't we need a moving avg?

theomonnom: They're streaming 500ms by 500ms?

tinalenguyen: oh the 500ms is the timeout, it streams normally

i think the frame is 80 ms. i don't think we need a moving average if the frames are consistently almost silent

theomonnom: but isn't 80ms too short? what if it's just a pause between two words?

tinalenguyen: when i tested it, pauses between words and sentences didn't trigger the timeout, should we increase the timeout to be safe?

one thing to note is that for interruptions, the model replies so quickly that the interruption is included in the original turn, though i don't see a way around this right now

milanperovic: This makes sense! The single-frame peak check works because the silence timeout is the real guard — you'd need ~10 consecutive filler frames (~800ms) before EOT triggers, so a brief pause between words won't cause a false positive. Increasing the default timeout to 800ms sounds like a good safety margin to me.

PR #5210, opened 2026-03-24
fix: re-generate LLM response when chat context becomes stale

longcw:
re-generate LLM/TTS if chat context changed between generation start and authorization, fix the case when `generate_reply` is called without awaiting or interrupting the ongoing speech, e.g.
```python
        session.generate_reply(user_input="count 1 2 3")
        # ... speech scheculed and not finished yet
        session.generate_reply(user_input="continue to count to 10")
```

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -63,6 +63,7 @@ from .events import (
 from .generation import (
     ToolExecutionOutput,
     _AudioOutput,
+    _LLMGenerationData,
     _TextOutput,
     _TTSGenerationData,
     apply_instructions_modality,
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2149,76 +2162,74 @@ async def _pipeline_reply_task_impl(
...
-        if new_message is not None and speech_handle.scheduled:
-            self._agent._chat_ctx.insert(new_message)
-            self._session._conversation_item_added(new_message)
-            user_metrics = new_message.metrics
+        pipeline_gen = await _start_generation(chat_ctx)
 
+        wait_for_scheduled = asyncio.ensure_future(speech_handle._wait_for_scheduled())
+        await speech_handle.wait_if_not_interrupted([wait_for_scheduled])
         if speech_handle.interrupted:
             current_span.set_attribute(trace_types.ATTR_SPEECH_INTERRUPTED, True)
```
chenghao-mou, line 2230: Any reason the user message is not inserted anymore here?

PR #5214, opened 2026-03-24
feat(soniox): expose max_endpoint_delay_ms option

pstrav:
## Summary
- Add configurable `max_endpoint_delay_ms` to `STTOptions` for tuning endpoint detection latency
- The Soniox API supports 500–3000ms (default 2000ms server-side), but the option was not previously exposed
- Add `__post_init__` validation to ensure the value is within the supported range
- Only sent to the API when explicitly set, preserving existing default behavior

This is useful for voice AI agents where the default 2000ms delay is too slow for responsive turn-taking.

See: https://soniox.com/docs/stt/rt/endpoint-detection

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
@@ -119,8 +119,17 @@ class STTOptions:
     enable_speaker_diarization: bool = False
     enable_language_identification: bool = True
 
+    max_endpoint_delay_ms: int = 500
+    """Maximum delay in milliseconds between speech cessation and endpoint detection.
+    Range: 500–3000.
+    See: https://soniox.com/docs/stt/rt/endpoint-detection"""
+
     client_reference_id: str | None = None
     translation: TranslationConfig | None = None
+
+    def __post_init__(self) -> None:
+        if not (500 <= self.max_endpoint_delay_ms <= 3000):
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
```diff
@@ -119,9 +119,20 @@ class STTOptions:
     enable_speaker_diarization: bool = False
     enable_language_identification: bool = True
 
+    max_endpoint_delay_ms: int | None = None
```
davidzhao: thanks for pointing this out.

I think we should handle it differently in the plugin since LK handles end of turn detection natively. the defaults for plugins should be to return finalized text chunks as quickly as possible.

with that in mind, should we set the delay to 200ms by default here?

pstrav: @davidzhao no problem. Do you mean 2000ms? That's the Soniox default. If we wanted to go lower, the lowest we can do is 500ms. Updating for now to 2000ms but happy to move it to a different value.

davidzhao: let's keep it as low as possible. 2s would be extremely long to finalize a turn; especially if the model makes a mistake

pstrav: @davidzhao all done

PR #5215, opened 2026-03-24
fix(stt): preserve aligned transcript in fallback adapter

BhaskarKapri07:
## Summary

This fixes a bug where `stt.FallbackAdapter` drops `STTCapabilities.aligned_transcript` when rebuilding its merged capabilities.

As a result, downstream code sees `aligned_transcript=False` even when all wrapped STTs support aligned transcripts, which disables adaptive interruption handling.

## What changed

- Preserve a merged `aligned_transcript` capability in `stt.FallbackAdapter`
- Downgrade mixed `"word"` / `"chunk"` providers to `"chunk"` as the common
denominator
- Added regression coverage for:
  - all `"word"` -> `"word"`
  - mixed `"word"` / `"chunk"` -> `"chunk"`
  - any `False` -> `False`

## Why this belongs in `FallbackAdapter`

The individual STT plugins already report their own `aligned_transcript` capability correctly. The information was being lost when `FallbackAdapter` reconstructed `STTCapabilities`.

### DIFF
--- a/livekit-agents/livekit/agents/stt/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/stt/fallback_adapter.py
@@ -68,11 +68,20 @@ class FallbackAdapter(
                 StreamAdapter(stt=t, vad=vad) if not t.capabilities.streaming else t for t in stt
             ]
 
+        aligned_vals = [t.capabilities.aligned_transcript for t in stt]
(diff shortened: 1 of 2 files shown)


BhaskarKapri07:
closing because of https://github.com/livekit/agents/pull/5237

PR #5217, opened 2026-03-24
Fix mcp session teardown

### LINKED ISSUE #5212: AgentSession.aclose() leaves the event loop hot when adding MCPs
### Bug Description

I’m seeing a cleanup issue when using MCPs over HTTP streamable.

After a session finishes and await session.aclose() returns, the process stays busy and CPU remains high. If I remove MCP, the process goes idle normally.

I can reproduce this with both:
mcp_servers=[MCPServerHTTP(...)]
tools=[MCPToolset(...)]

Actual result:
- CPU stays around 20%+ in my repro
- loop._ready keeps one callback alive: CancelScope._deliver_cancellation()
- there are also 2 lingering async generators from the MCP HTTP transport

If I run the same flow without MCP, CPU drops back to idle and there are no lingering ready callbacks / async generators.
If run with num_idle_processes=0, the leak is hidden because the whole child is tear down.

What I found while debugging:

MCP setup and teardown seem to happen across different asyncio tasks.
The failure looks related to AnyIO cancel-scope ownership in the MCP HTTP transport / client session lifecycle.
When I manage the MCP server manually in one task (initialize() -> use tools -> aclose() in the same task), the problem goes away

### Expected Behavior

When I close a session that used MCP servers, the loop._ready should clean all the callbacks and the usage of CPU should go back to ~0%

### Reproduction Steps

```bash
import asyncio
import time

from livekit.agents.llm.mcp import MCPServerHTTP, MCPToolset

MCP_URL = "https://docs.livekit.io/mcp"
MODEL = "gpt-5.4-nano"

def cpu_sample(prev):
    now_wall = time.monotonic()
    now_cpu = time.process_time()
    if prev is None:
        return (now_wall, now_cpu, None)
    wall = now_wall - prev[0]
    cpu = now_cpu - prev[1]
    pct = (cpu / wall) * 100 if wall > 0 else None
    return (now_wall, now_cpu, pct)


def dump_loop(loop, label):
    ready = list(getattr(loop, "_ready", []))
    asyncgens = list(getattr(loop, "_asyncgens", []))

    print(f"[{label}] ready={len(ready)} asyncgens={len(asyncgens)}")

    for i, handle in enumerate(ready[:3]):
        print("  ready_handle", i, repr(handle))

    for i, ag in enumerate(asyncgens[:5]):
        code_name = getattr(getattr(ag, "ag_code", None), "co_name", None)
        frame = getattr(ag, "ag_frame", None)
        if frame is not None:
            where = f"{frame.f_code.co_filename}:{frame.f_lineno}"
        else:
            where = "no_frame"
        print("  asyncgen", i, code_name, where)


async def main():
    llm = openai.LLM(
        model="gpt-5.4-nano",
        reasoning_effort="none",
    )

    server = MCPServerHTTP(MCP_URL)

    toolset = MCPToolset(id="mcp_test", mcp_server=server)
    session = AgentSession(llm=llm)
    agent = Agent(
        tools=[toolset],
    )

    await session.start(agent=agent, record=False)

    await session.run(user_input="Hello!")

    await session.aclose()
    session.shutdown()

    loop = asyncio.get_running_loop()
    prev = None

    # Here I can see something like this:
    # [idle_1] ready=1 asyncgens=2
    # ready_handle 0 <Handle CancelScope._deliver_cancellation()>
    # asyncgen 0 streamable_http_client no_frame
    # asyncgen 1 _streamable_http_with_client no_frame
    # idle_cpu_pct 21.71
    for i in range(5):
        prev = cpu_sample(prev)
        if prev[2] is not None:
            print("idle_cpu_pct", round(prev[2], 2))
        dump_loop(loop, f"idle_{i}")
        await asyncio.sleep(1)


asyncio.run(main())
```

### Operating System

macOS 15.7.4

### Models Used

_No response_

### Package Versions

```bash
livekit==1.1.3
livekit-agents==1.5.1
mcp==1.26.0
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
```

### Additional Context

_No response_

_No response_

csanz91:
Keep MCP client enter and exit on the same dedicated lifecycle task so AnyIO cancel scopes are closed by their owner task.

This fixes #5212

### DIFF
--- a/livekit-agents/livekit/agents/llm/mcp.py
+++ b/livekit-agents/livekit/agents/llm/mcp.py
@@ -102,12 +102,12 @@ class MCPServer(ABC):
         async with self._lifecycle_lock:
             if self._lifecycle_task is None or self._lifecycle_task.done():
                 loop = asyncio.get_running_loop()
-                ready_fut = loop.create_future()
+                new_ready_fut: asyncio.Future[None] = loop.create_future()
                 close_event = asyncio.Event()
-                self._ready_fut = ready_fut
+                self._ready_fut = new_ready_fut
                 self._close_event = close_event
                 self._lifecycle_task = asyncio.create_task(
-                    self._run_client_lifecycle(ready_fut=ready_fut, close_event=close_event),
+                    self._run_client_lifecycle(ready_fut=new_ready_fut, close_event=close_event),
                     name=f"{type(self).__name__}._run_client_lifecycle",
                 )
 


longcw:
fixed in https://github.com/livekit/agents/pull/5223

PR #5223, opened 2026-03-25
fix: ensure MCP client enter/exit run in the same task

longcw:
fix https://github.com/livekit/agents/issues/5212

### DIFF
--- a/livekit-agents/livekit/agents/llm/mcp.py
+++ b/livekit-agents/livekit/agents/llm/mcp.py
@@ -90,6 +90,7 @@ class MCPServer(ABC):
 
         self._client_task: asyncio.Task[None] | None = None
         self._closing_ev = asyncio.Event()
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/llm/mcp.py
```diff
@@ -96,22 +99,43 @@ def invalidate_cache(self) -> None:
         self._cache_dirty = True
 
     async def initialize(self) -> None:
+        if self._client_task and not self._client_task.done():
+            logger.warning("MCPServer is already initializing")
+            return
```
chenghao-mou, line 107: QQ: for concurrent initialize calls, it would return immediately without being ready, should we wait for the ready future here too?

longcw: this makes sense, I'll add it.

PR #5228, opened 2026-03-25
(azure openai realtime): remove api version query param

tinalenguyen:
[source](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/realtime-audio-preview-api-migration-guide#websocket-connection-returns-401-unauthorized)

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -679,7 +679,6 @@ class RealtimeModel(llm.RealtimeModel):
 def process_base_url(
     url: str,
     model: str,
-    is_azure: bool = False,
     azure_deployment: str | None = None,
 ) -> str:
     if url.startswith("http"):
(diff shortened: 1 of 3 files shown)


davidzhao (commented):
do both OAI and Azure work the same way wrt `api-version`? do you think re: future proof, we should keep it?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -412,6 +412,12 @@ def __init__(
...
 
+        if api_version is not None:
+            logger.warning(
+                "api_version is deprecated for Azure OpenAI Realtime GA endpoint and will be ignored. "
+                "The GA endpoint does not accept an api-version query parameter."
+            )
+
         is_azure = (
             api_version is not None or entra_token is not None or azure_deployment is not None
         )
```
chenghao-mou, line 423: We might need to update this check as well.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
chenghao-mou, line 233: We can drop this from the options since it is not read anymore.

PR #5232, opened 2026-03-26
feat: Add D-ID avatar plugin

osimhi213:
- Adds `livekit-plugins-did` — a D-ID virtual avatar plugin for LiveKit Agents.
- Plugin follows the same structure and conventions as the Tavus and Anam avatar plugins
- Includes example agent at `examples/avatar_agents/did/`
- Configurable audio sample rate via `AudioConfig` dataclass
- [example video](https://www.loom.com/share/ea996db9da7947748cc4def7035dd4d7)


tinalenguyen:
hi, thank you for the PR! could you sign the CLA and add the plugin to [livekit-agents/pyproject.toml](https://github.com/livekit/agents/blob/main/livekit-agents/pyproject.toml) and [pyproject.toml](https://github.com/livekit/agents/blob/main/pyproject.toml)

livekit-plugins/livekit-plugins-did/livekit/plugins/did/avatar.py
```diff
@@ -0,0 +1,126 @@
...
+        avatar_participant_identity: NotGivenOr[str] = NOT_GIVEN,
+        avatar_participant_name: NotGivenOr[str] = NOT_GIVEN,
+        conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
+    ) -> None:
+        self._http_session: aiohttp.ClientSession | None = None
+        self._conn_options = conn_options
+        self._agent_id = agent_id
+        self._audio_config = audio_config or AudioConfig()
+        self.session_id: str | None = None
+
```
tinalenguyen, line 48: we should throw an informative error if the `agent_id` isn't given here

livekit-plugins/livekit-plugins-did/livekit/plugins/did/api.py
```diff
@@ -0,0 +1,92 @@
...
+    ) -> str:
+        """Dispatch a D-ID avatar worker into the room.
+
+        Returns the session id.
+        """
+        payload: dict[str, Any] = {
+            "transport": transport,
+            "audio_config": audio_config,
+        }
+        response_data = await self._post(f"v2/agents/{agent_id}/sessions/join", payload)
```
tinalenguyen, line 56: are only v2 avatars supported for the plugin? it might be worth making a note in the readme

tinalenguyen (commented):
i left a few comments, also some feedback i have is that it's a bit tricky to find public avatar IDs to use with the plugin. i would recommend making it easier to find on the dashboard and also having a `DEFAULT_AGENT_ID` here as well so users can try it right away

livekit-plugins/livekit-plugins-did/livekit/plugins/did/avatar.py
osimhi213: typing is not enough IYO?

livekit-plugins/livekit-plugins-did/livekit/plugins/did/api.py
osimhi213: yes, will add a note on compatible avatars in the README.

osimhi213: I added a note and an example to the readme.

osimhi213:
> i left a few comments, also some feedback i have is that it's a bit tricky to find public avatar IDs to use with the plugin. i would recommend making it easier to find on the dashboard and also having a `DEFAULT_AGENT_ID` here as well so users can try it right away

I’ll ask our PM whether we want to share a public agent ID for plugin use.
Update: We are working on that but it's not a blocker.

livekit-plugins/livekit-plugins-did/livekit/plugins/did/avatar.py
tinalenguyen: the intention was to give context on how to create an agent id, but i see that there are instructions in the README now

tinalenguyen (approved):
lgtm, thanks for iterating!

Issue #5226, opened 2026-03-25
Error "byte stream handler for topic 'lk.agent.session' already set" if multiple agent sessions created since v1.5.0


deepmindru-afk:
### Bug Description

Tryin to start modified [https://github.com/livekit/agents/blob/main/examples/other/transcription/multi-user-transcriber.py](multiuser_transcriber) since v1.5.0 when second user connects to room "error byte stream handler for topic 'lk.agent.session' already set" occurs.

### Expected Behavior

same as in versions <= v1.5.0==rc2

### Reproduction Steps

```bash
async def _start_session(self, participant: rtc.RemoteParticipant) -> AgentSession:
        if participant.identity in self._sessions:
            return self._sessions[participant.identity]

        session = AgentSession(
            vad=self.ctx.proc.userdata["vad"],
        )
        await session.start(
            agent=Transcriber(
                participant_identity=participant.identity,
            ),
            room=self.ctx.room,
            room_options=room_io.RoomOptions(
                audio_input=True,
                text_output=True,
                audio_output=False,
                participant_identity=participant.identity,
                # text input is not supported for multiple room participants
                # if needed, register the text stream handler by yourself
                # and route the text to different sessions based on the participant identity
                text_input=False,
            ),
        )
        return session
```

### Operating System

Debian

### Models Used

Deepgram

### Package Versions

```bash
livekit-agents==1.5.1
```

### Proposed Solution

maybe same as in [https://github.com/livekit/agents/issues/2034](bug)

sixbangs:
We're running into this too - there are a lot of good reasons to support multiple concurrent instances of `AgentSession`, so hopefully it gets fixed soon. Here's a workaround:

```
    _original_register = room.register_byte_stream_handler

    def _idempotent_register(topic, handler):
        try:
            _original_register(topic, handler)
        except ValueError:
            pass

    room.register_byte_stream_handler = _idempotent_register
```

deepmindru-afk:
seems like patch for [https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/voice/remote_session.py](remote_session) needed: it tries to create byte_stream handler already existed for any new session. maybe try-except will work, but not sure for now if any other problems won't occur.

also there is no any documented cases of Session/TCP transport use in >=v.1.5.0.

PR #5241, opened 2026-03-27
fix: only start session host when it's primary session

longcw:
fix https://github.com/livekit/agents/issues/5226

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -605,6 +605,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
             self._started_at = time.time()
 
             # configure observability first
+            record_is_given = is_given(record)
             job_ctx: JobContext | None = None
             try:
                 # defer to server-side setting for recording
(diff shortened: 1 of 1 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -618,9 +618,20 @@ async def start(
...
 
+            is_primary = True
             if job_ctx:
                 job_ctx.init_recording(self._recording_options)
 
+                if job_ctx._primary_agent_session is None or job_ctx._primary_agent_session is self:
+                    job_ctx._primary_agent_session = self
+                else:
+                    is_primary = False
+                    if any(self._recording_options.values()):
```
davidzhao, line 629: I don't think we should raise the error if the user didn't pass in an explicit `record=True`

if users are leaving it to defaults, and the server is instructing it to record, we shouldn't raise here

longcw: This was added from the beginning, when using multiple AgentSession, it needs to be explicitly set to `record=False`. Do you want to automatically disable recording if it's not specified and the session isn't the primary one (the first one started)?

davidzhao: > Do you want to automatically disable recording if it's not specified and the session isn't the primary one (the first one started)

yeah I think that'd be better than failing the session. it seems strange that the default value would mean it's crashing.

PR #5246, opened 2026-03-27
feat: voxtral tts integration

fabitokki:
Integrate new voxtrall tts in livekit-plugins-mistralai

#5247


mahimairaja:
@fabitokki streaming is failing

fabitokki:
@mahimairaja apparently I used the wrong model names, which is weird cause it's the one from the doc: https://docs.mistral.ai/models/voxtral-tts-26-03

Anyway, thanks for noticing

fabitokki:
I am closing my PR since it is duplicated

Here is the official one: https://github.com/livekit/agents/pull/5245

Issue #5234, opened 2026-03-26
google: support for gemini-3.1-flash-live-preview


tg-bomze:
### Feature Type

Nice to have

### Feature Description

Today, Google [presented](https://blog.google/innovation-and-ai/technology/developers-tools/build-with-gemini-3-1-flash-live/) a new real-time speech-to-speech model: `gemini-3.1-flash-live-preview`
This model is not yet available in the `livekit-plugins-google` plugin. As far as I understand, TEXT (modality) output is also not yet available for this model.

smittie2000:
if you self host and run your own agent you can monkey patch it as follows for 3.1 to test this fixes the 1007 Request contains an invalid argument. This is quick test i did and got better stability on my tool calling. Opus did most of it so this is to test only 

https://ai.google.dev/gemini-api/docs/models/gemini-3.1-flash-live-preview#migrating

The migration guide on model card has more information

```python
# ============================================================================
# MONKEY-PATCH: Gemini 3.1 Flash Live — send_client_content compatibility
# ----------------------------------------------------------------------------
# Gemini 3.1 rejects ALL send_client_content calls (the old way of sending
# text turns). The LiveKit plugin's generate_reply() uses send_client_content
# internally to trigger the model to speak (e.g., greeting on_enter).
#
# This patch swaps generate_reply() to use send_realtime_input() instead,
# which 3.1 accepts. Only activates for models with "3.1" in the name —
# 2.5 models continue to use the original unpatched method.
#
# TODO(livekit-update): REMOVE this entire block when livekit-plugins-google
# ships native Gemini 3.1 support. Track: https://github.com/livekit/agents
# ============================================================================
try:
    from livekit.plugins.google.realtime.realtime_api import RealtimeSession as _GeminiRS
    from livekit.agents.types import NOT_GIVEN as _NOT_GIVEN
    from livekit.agents.utils import is_given as _is_given

    _original_generate_reply = _GeminiRS.generate_reply

    def _patched_generate_reply(self, *, instructions=_NOT_GIVEN):
        # Non-3.1 models: use original (send_client_content works fine)
        if "3.1" not in getattr(getattr(self, "_opts", None), "model", ""):
            return _original_generate_reply(self, instructions=instructions)

        # --- 3.1 path: replicate generate_reply but with send_realtime_input ---
        if self._pending_generation_fut and not self._pending_generation_fut.done():
            self._pending_generation_fut.cancel("Superseded by new generate_reply call")

[...]

PR #5251, opened 2026-03-27
fix(google realtime): support gemini-3.1-flash-live-preview

Hormold:
Adds working support for gemini-3.1-flash-live-preview.

Gemini 3.1 changed how send_client_content works. It's now only allowed for initial history seeding (with history_config.initial_history_in_client_content=true).

After the first model turn, all text input must go through send_realtime_input. This broke generate_reply, and any mid-session LiveClientContent (from update_instructions, update_chat_ctx, etc.) gets rejected with a 1007 error.

What this PR does:

- Introduces RESTRICTED_CLIENT_CONTENT_MODELS set in api_proto.py to identify models with this limitation - generate_reply() uses LiveClientRealtimeInput(text=...) instead of LiveClientContent for these models
- _send_task drops LiveClientContent events mid-session to prevent 1007 crashes. With reconnect_on_update=True, it triggers a reconnect instead so updates take effect via the new session's setup config
- _build_connect_config adds history_config.initial_history_in_client_content=True so initial history seeding still works on connect/reconnect
- _handle_server_content skips empty server_content that 3.1 sends (only usage_metadata, no actual content) - without this the first generation's audio gets dropped, thx #5234 
- Adds reconnect_on_update parameter to RealtimeModel — opt-in reconnect when update_instructions/update_chat_ctx is called on restricted models. Off by default since it interrupts the session.

Tested locally: greeting via generate_reply, bidirectional audio, function tool calling, update_instructions and update_chat_ctx with reconnect all work.

Known limitation: mid-session update_instructions and update_chat_ctx require a reconnect on 3.1 - there's no way around this since the model simply doesn't accept send_client_content after the first turn. Google's migration guide confirms this is by design.
image

Ref: https://ai.google.dev/gemini-api/docs/models/gemini-3.1-flash-live-preview#migrating
Related: #5234

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -927,14 +927,20 @@ class RealtimeSession(llm.RealtimeSession):
                     # Otherwise, silently drop to prevent 1007 errors.
                     if self._opts.model in RESTRICTED_CLIENT_CONTENT_MODELS:
                         if self._opts.reconnect_on_update:
-                            logger.debug("restricted model: reconnecting to apply LiveClientContent update")
+                            logger.debug(
+                                "restricted model: reconnecting to apply LiveClientContent update"
+                            )
                             self._mark_restart_needed()
                         else:
-                            logger.debug("restricted model: dropping LiveClientContent (not supported mid-session)")
+                            logger.debug(
+                                "restricted model: dropping LiveClientContent (not supported mid-session)"
+                            )
                     else:
                         await session.send_client_content(
                             turns=msg.turns,  # type: ignore
-                            turn_complete=msg.turn_complete if msg.turn_complete is not None else True,
+                            turn_complete=msg.turn_complete
+                            if msg.turn_complete is not None
+                            else True,
                         )
                 elif isinstance(msg, types.LiveClientToolResponse) and msg.function_responses:
                     await session.send_tool_response(function_responses=msg.function_responses)
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -25,7 +25,12 @@ from livekit.agents.types import (
     NotGivenOr,
 )
 from livekit.agents.utils import audio as audio_utils, images, is_given
-from livekit.plugins.google.realtime.api_proto import ClientEvents, LiveAPIModels, Voice
+from livekit.plugins.google.realtime.api_proto import (
+    RESTRICTED_CLIENT_CONTENT_MODELS,
+    ClientEvents,
+    LiveAPIModels,
+    Voice,
+)
 
 from ..log import logger
 from ..utils import create_tools_config, get_tool_results_for_realtime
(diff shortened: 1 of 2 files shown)


JiwaniZakir (commented):
In `generate_reply`, the restricted-model branch sends `instructions` directly as a `LiveClientRealtimeInput` text prompt, but the original non-restricted path framed those same instructions as a `role="model"` content turn followed by a `role="user"` placeholder turn. These have meaningfully different semantics — the instructions were intended to prime the model's perspective, not arrive as user speech — so the behavior diverges silently for callers passing `instructions` to `generate_reply` on a restricted model.

In `_send_task`, when `reconnect_on_update=True` triggers `_mark_restart_needed()` upon receiving a `LiveClientContent`, the content in `msg.turns` is discarded. The reconnect re-seeds history via `system_instruction`/`initial_history`, but any turn data specific to that particular `LiveClientContent` message (e.g., a one-off model-role instruction) is lost without any log warning, making this a silent data loss path that could be hard to debug.

The `RESTRICTED_CLIENT_CONTENT_MODELS` membership check is now scattered across at least three call sites (`generate_reply`, `_send_task`, and implicitly `_build_connect_config` per the comment). Centralizing this behind a small helper method like `_is_restricted_client_content_model()` would reduce the risk of a future model being added to the frozenset but missing one of the branching locations.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -218,6 +224,7 @@ def __init__(
         conn_options: APIConnectOptions = DEFAULT_API_CONNECT_OPTIONS,
         http_options: NotGivenOr[types.HttpOptions] = NOT_GIVEN,
         thinking_config: NotGivenOr[types.ThinkingConfig] = NOT_GIVEN,
+        reconnect_on_update: bool = False,
```
davidzhao, line 227: should this be a user specified option? when do you want this to be True unless the model requires it?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -701,12 +713,19 @@ def generate_reply(
...
-            turns.append(types.Content(parts=[types.Part(text=instructions)], role="model"))
-        turns.append(types.Content(parts=[types.Part(text=".")], role="user"))
-        self._send_client_event(types.LiveClientContent(turns=turns, turn_complete=True))
+        # so we need to add a placeholder user turn in order to trigger a new generation.
+        # Restricted client_content models (e.g. gemini-3.1-flash-live-preview) reject send_client_content
+        # after the first model turn and require send_realtime_input for text input.
+        # See: https://ai.google.dev/gemini-api/docs/models/gemini-3.1-flash-live-preview#migrating
+        if self._opts.model in RESTRICTED_CLIENT_CONTENT_MODELS:
+            prompt = instructions if is_given(instructions) else "."
+            self._send_client_event(types.LiveClientRealtimeInput(text=prompt))
```
davidzhao, line 722: this isn't going to work. realtime input is coming from the end user, but generate_reply instructions needs to be coming from the model itsel

davidzhao (changes requested):
I don't think this type of workaround is a good idea. let's engage the deepmind folks to understand the best path forward

PR #5261, opened 2026-03-28
feat(upliftai): add support for phrase replacement config id

zaidqureshi2:
We support a config ID users can create in our api and use that for to help pronunciation in their inference. Adding support for it in livekit plugin as well.


zaidqureshi2:
the CI/ruff check is failing on pre existing aws failures, i didn't check them in

i see https://github.com/livekit/agents/pull/5245 already addresses it as well.

PR #5271, opened 2026-03-30
fix: wait_for_participant waits until participant is fully active

davidzhao:
requires livekit v1.1.4


chenghao-mou (commented):
should we update the pyproject and uv lock files?

davidzhao:
> should we update the pyproject and uv lock files?

yup, waiting for python sdk release before doing so..

chenghao-mou (approved):

IngLP:
hi @davidzhao , this breaks my implementation! I was using wait_for_participant to setup the agent according to participant metadata, and this used to work because I had time to do so, while the telephone was ringing. Now I am losing that precious time to do the setup.

davidzhao:
@IngLP can you clarify what actually happens in your use case? this should introduce less than 500ms of latency to wait_for_participant in the typical case.

IngLP:
@davidzhao maybe that's true for incoming calls, but for outgoing calls, it's not. I had plenty of time while the other extension was ringing, now wait_for_participant only unlocks when the callee has actually answered.

davidzhao:
@IngLP can you share a code snippet of what you are doing? and describe the current behavior that you are observing?

IngLP:
Before:
* Outgoing call launched by backend with CreateSIPParticipant, agent config written in sip participant metadata
* The call starts and it's ringing
* Agent starts, waits for participant, reads metadata, setups agent and tools
* Call is still ringing
* Contact answers and everything is ready

Now:
* Contact answers and everything has still to be set up, including recording, which cuts part of the call.

davidzhao:
@IngLP please share example code snippet.. that'll make it easier for us to see what's going on. if you are calling specific APIs, include the actual calls with the parameters

IngLP:
@davidzhao it's just like this:
```
@server.rtc_session()
    [other setup code]

    logger.info(f"connecting to room {ctx.room.name}")

    try:
    except RuntimeError as exc:
        if _is_expected_participant_wait_disconnect(exc):
            logger.info(
                "Room disconnected before participant joined; ending job without agent startup"
            )
            return
        raise
    
    [setup agent according to participant metadata, other code]

    agent_session = AgentSessionExt(
        **components,
    )
    await agent_session.start_session(agent, participant, room=ctx.room)

```

davidzhao:
@IngLP where do you perform the outbound dial?

IngLP:
@davidzhao in the backend, a separate deployment from the agents runner.

decafdennis:
This change also hangs the `wait_for_participant` call in console mode, because `create_mock_room` does not mock out the participant state and so the participant never becomes active.

PR #5279, opened 2026-03-30
fix(recorder): guard against empty agent speech frames

chenghao-mou:
`buf` might contain an empty frame from split, so this PR guards against spurious buffer writes.


longcw (approved):
looks good to me! nit: instead of checking the sum of durations, should we filter out zero-duration frames in the buffer before `if buf`?

chenghao-mou:
> looks good to me! nit: instead of checking the sum of durations, should we filter out zero-duration frames in the buffer before `if buf`?

good point. Updated the code to filter buf instead.

PR #5283, opened 2026-03-30
feat(stt): add speaker diarization support to STT interface and proxy

russellmartin-livekit:
Related gateway change: https://github.com/livekit/agent-gateway/pull/557

### Changes in `agents`
- **Options**: Updates `DeepgramOptions` and `AssemblyaiOptions` to explicitly type diarization flags (`diarize` and `speaker_labels`).
- **Capabilities**: Modifies the `STT` inference wrapper to dynamically set the `diarization` capability based on the provided `extra_kwargs` during initialization and `update_options`.
- **Data Models**: Adds `speaker_id` to `TimedString` and populates it from the inference proxy's response.
- **Testing**: Adds tests to verify diarization capability detection.

Fixes AGT-2608

Slack thread: https://live-kit.slack.com/archives/C06TN33TV44/p1772573869144129?thread_ts=1771977322.899519&cid=C06TN33TV44

https://claude.ai/code/session_01VRKQuBXiq8BHKr9AiJ6uEw

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -721,6 +721,7 @@ class AudioRecognition:
             ):
                 self._last_language = language
 
+            self._final_transcript_received.set()
             if not transcript:
                 return
 
@@ -745,7 +746,6 @@ class AudioRecognition:
             transcript_changed = self._audio_transcript != self._audio_preflight_transcript
             self._audio_interim_transcript = ""
             self._audio_preflight_transcript = ""
-            self._final_transcript_received.set()
 
             if not self._vad or self._last_speaking_time is None:
                 # vad disabled, use stt timestamp


livekit-agents/livekit/agents/inference/stt.py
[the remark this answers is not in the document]
theomonnom: This seems like a valid concern?

tests/test_inference_stt_fallback.py
```diff
@@ -257,3 +257,51 @@ def test_connect_options_full_custom(self):
...
+        assert stt.capabilities.diarization is False
+        stt.update_options(extra={"diarize": True})
+        assert stt.capabilities.diarization is True
+
+    def test_update_options_disables_diarization(self):
+        """update_options can disable diarization by setting params to False."""
+        stt = _make_stt(extra_kwargs={"diarize": True})
+        assert stt.capabilities.diarization is True
+        stt.update_options(extra={"diarize": False})
+        assert stt.capabilities.diarization is False
```
theomonnom: Those tests aren't really useful?

russellmartin-livekit: yeah its overkill, removed most of them

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -688,12 +663,19 @@ def _process_transcript(self, data: dict, is_final: bool) -> None:
...
+            if word_speakers:
+                speaker = Counter(word_speakers).most_common(1)[0][0]
+
         speech_data = stt.SpeechData(
             language=language,
             start_time=self.start_time_offset + data.get("start", 0),
             end_time=self.start_time_offset + data.get("start", 0) + data.get("duration", 0),
             confidence=data.get("confidence", 1.0),
             text=text,
+            speaker_id=f"S{speaker}" if speaker is not None else None,
```
theomonnom: What's tricky with diarization in our inference API. Is that we need a way for the speaker_id to be consistent across every provider. 

Maybe some of them aren't even int, and str?

russellmartin-livekit: I could standardize the inference side or I could just pass through whatever we get from the provider

theomonnom: I think the gateway should standardize it. 
It would be OK if it was extra, but everything inside the core "inference" API should be the same across every provider

russellmartin-livekit: Agree, my latest changes standardizes it on the gateway size

russellmartin-livekit: Standardize on gateway side to return string ints so it stays compatible with plugins that require strings. Also added support for word level diarization

PR #5285, opened 2026-03-31
(phonic) Update languages fields

qionghuang6:
Phonic recently updated it's languages fields to use a default language with additional languages.  This changes our plugin to use the new fields.

### DIFF
--- a/livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
@@ -169,6 +169,10 @@ class RealtimeModel(llm.RealtimeModel):
             and not is_given(default_language)
             and not is_given(additional_languages)
         ):
+            logger.warning(
+                "The `languages` parameter is deprecated; use `default_language` and `additional_languages` instead. When both are omitted, "
+                "`languages[0]` is the default language and `languages[1:]` are additional languages."
+            )
             if languages:
                 default_language = languages[0]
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -100,7 +102,9 @@ def __init__(
         welcome_message: NotGivenOr[str | None] = NOT_GIVEN,
         generate_welcome_message: NotGivenOr[bool] = NOT_GIVEN,
         project: NotGivenOr[str | None] = NOT_GIVEN,
-        languages: NotGivenOr[list[str]] = NOT_GIVEN,
```
davidzhao, line 103: same question re: backwards compatibility on patch versions

qionghuang6: Yup, let me update that.

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -152,14 +164,26 @@ def __init__(
                 "set PHONIC_API_KEY environment variable."
             )
 
+        if (
+            is_given(languages)
+            and not is_given(default_language)
+            and not is_given(additional_languages)
+        ):
```
tinalenguyen, line 171: i would also add a` logger.warning` here with the note from the docstring

PR #2058, opened 2025-04-21
fix generate_reply instructions 

theomonnom:

PR #5287, opened 2026-03-31
append generate_reply instructions as system msg and convert it to user msg if unsupported

longcw:
- `generate_reply` now appends turn instructions as a system message
- Add `convert_mid_conversation_instructions` in provider format utils — converts mid-conversation system messages to user messages with `<instructions>` tags for Anthropic, Google, and AWS. OpenAI/Mistral unchanged (native support).

this improves the reliability for `generate_reply` with instructions when there is a long chat context, especially for gpt-4.1 and gpt-4.1-mini. 

testing results:
[mid_conversation_instructions_20260401_163916.html](https://github.com/user-attachments/files/26425603/mid_conversation_instructions_20260401_163916.html)

### DIFF
--- a/tests/fake_llm.py
+++ b/tests/fake_llm.py
@@ -121,8 +121,12 @@ class FakeLLMStream(LLMStream):
         item = items[-1]
 
         # for user message and generate_reply(instructions=...)
-        if item.type == "message" and item.role in ("user", "system"):
-            return item.text_content
+        if (
+            item.type == "message"
+            and item.role in ("user", "system")
+            and (text := item.text_content)
+        ):
+            return text
 
         # if the last item is a function call output, use the tool output
         if item.type == "function_call_output":


livekit-agents/livekit/agents/llm/_provider_format/utils.py
```diff
@@ -6,6 +6,49 @@
 from livekit.agents import llm
 from livekit.agents.log import logger
 
+_DEFAULT_INLINE_INSTRUCTIONS_TEMPLATE = "<instructions>\n{content}\n</instructions>"
+
+
+def convert_mid_conversation_instructions(
+    chat_ctx: llm.ChatContext,
+    *,
+    role: llm.ChatRole = "user",
```
davidzhao, line 15: should this be converted to a user message? or assistant? does it achieve the same result because it's wrapped in an instruction template?

since `to_provider_format` is already handling cases where it cannot end as assistant, it seems easier to reason with

longcw: the model might treat assistant messages as things it's already said, so perhaps user message is a better option, tho during my testing they have similar results.

cchen-dialpad:
@longcw I have a question about the `generate_reply()`'s `instructions` argument. The doc suggests it behaves differently for Realtime and Pipeline model: https://docs.livekit.io/agents/multimodality/audio/#how-instructions-interact-with-session-level-instructions I didn't track the documentation history, but I think this explicit call out is a recent change. 

In your changes under this PR, I saw you removed this comment in the following screenshot. The removed comment has been there for about a year (https://github.com/livekit/agents/pull/2058), and it suggested the behavior should be the same for Realtime and Pipeline model. So I'm a bit confused, do you know since which version this behavioral difference was introduced? 

image

longcw:
@cchen-dialpad I think the behavior is still the `instructions` of `generate_reply` are extra instructions, for both pipeline and realtime agents, I'll double check with the team.

cchen-dialpad:
> @cchen-dialpad I think the behavior is still the `instructions` of `generate_reply` are extra instructions, for both pipeline and realtime agents, I'll double check with the team.

Thanks. I did some testing with a minimal agent (v1.4.3), and Realtime (OpenAI) agent's behavior is different from Pipeline agent. Realtime agent's system prompt does not seem to be included by default in `generate_reply()`, while Pipeline agent can respond with knowledge from both system prompt and generate_reply()'s instructions.

jp-paper:
Hello! I am happy to have reviewed the release notes of `1.5.2` because this would have caused a breaking change in our implementation, as we are relying on any `role=system` messages in the `chat_ctx` to be appended to the initial system instructions.

PR #5288, opened 2026-03-31
fix(voice): make function call history preservation configurable in AgentTask

### LINKED ISSUE #5284: AgentTask hardcodes `exclude_function_call=True` when merging chat context back to parent agent, causing loss of tool call history
### Bug Description

When an AgentTask completes, the `__await_impl` finally block merges the sub-task's `chat_ctx` back into the parent agent's `chat_ctx` with a hardcoded `exclude_function_call=True`:

https://github.com/livekit/agents/blob/0ed8e6b9b04971b88d9fcecf099dcdf2f94d11d4/livekit-agents/livekit/agents/voice/agent.py#L893-L899

```python
merged_chat_ctx = old_agent.chat_ctx.merge(
    self.chat_ctx, exclude_function_call=True, exclude_instructions=True
)
# set the chat_ctx directly, [session._update_activity](cci:1://file:///Users/tony/src/python-core/.venv/lib/python3.11/site-packages/livekit/agents/voice/agent_session.py:1107:4-1181:63) will sync it to the rt_session if needed
old_agent._chat_ctx.items[:] = merged_chat_ctx.items
```
This unconditionally strips all function_call and function_call_output items from the sub-task's context when merging back to the parent. There is no parameter or hook to preserve them.

### Impact
* Tool call history is permanently lost when an AgentTask completes. Any data retrieved via tool calls during the sub-task (e.g., MCP tool results, API responses, database lookups) is removed from the parent agent's context.
* The parent's LLM session is synced with the stripped context during `resume() → _start_session()`, before the caller has any opportunity to intervene. For RealtimeModel, this means `rt_session.update_chat_ctx()` is called with a context missing all function call history.
* No workaround within the framework: the merge directly mutates `old_agent._chat_ctx.items[:]`, bypassing `update_chat_ctx`. Callers can manually call `update_chat_ctx` after await task returns, but by that point the Realtime session has already been synced with the filtered context.

### Expected Behavior

The merge behavior should be configurable. Users should be able to preserve function call history from sub-tasks when it's needed for context continuity (e.g., when users can switch between tasks and resume previous ones).

### Reproduction Steps

1. Create an Agent (parent) with tool-calling capabilities
2. Create an AgentTask (sub-task) that executes tool calls during its lifecycle (e.g., MCP tools, function tools that return data)
3. Complete the sub-task via self.complete(result)
4. Observe the parent agent's chat_ctx after the task returns — all function_call and function_call_output items from the sub-task are missing

```
class MySubTask(AgentTask[str]):
    async def on_enter(self):
        # Tool calls happen during sub-task lifecycle
        # e.g., MCP tools fetch data, function tools execute actions
        ...
        self.complete("done")

# In parent agent's function tool:
result = await MySubTask(instructions="...", chat_ctx=self.chat_ctx.copy(), tools=[...])
# At this point, parent's chat_ctx has lost all function_call/function_call_output
# items from the sub-task
```

### Operating System

macOS 15, all Linux platforms

### Models Used

_No response_

### Package Versions

```bash
livekit-agents==1.4.3 (Commit: 0ed8e6b9b04971b88d9fcecf099dcdf2f94d11d4)
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

Add a parameter to AgentTask (or complete()) to control the merge behavior:

```python
class AgentTask(Agent, Generic[TaskResult_T]):
    def __init__(self, *, merge_exclude_function_call: bool = True, ...):
        self._merge_exclude_function_call = merge_exclude_function_call
        ...
```

Then in `__await_impl`'s finally block:

```python
merged_chat_ctx = old_agent.chat_ctx.merge(
    self.chat_ctx,
    exclude_function_call=self._merge_exclude_function_call,
    exclude_instructions=True,
)
```
This preserves backward compatibility (default True) while allowing users to opt in to preserving function call history.

### Additional Context

_No response_

_No response_

GopalGB:
## Summary
- Adds `preserve_function_call_history` parameter to `AgentTask.__init__()` (default `False` for backward compatibility)
- When `True`, function call and function call output items from the sub-task's chat context are retained when merging back to the parent agent
- Previously, `exclude_function_call=True` was hardcoded, unconditionally stripping all tool call history

## Root Cause
In `AgentTask.__await_impl`, the merge call:
```python
old_agent.chat_ctx.merge(self.chat_ctx, exclude_function_call=True, ...)
```
unconditionally strips all function_call items. Users who need tool call results for context continuity (e.g., MCP tool responses, API lookups) had no way to preserve them.

## Test Plan
- [x] Default behavior unchanged (`preserve_function_call_history=False`)
- [x] New parameter correctly inverts `exclude_function_call` flag
- [x] Backward compatible — no changes to existing call sites

Fixes #5284

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent.py
+++ b/livekit-agents/livekit/agents/voice/agent.py
@@ -700,6 +700,7 @@ class AgentTask(Agent, Generic[TaskResult_T]):
         llm: NotGivenOr[llm.LLM | llm.RealtimeModel | None] = NOT_GIVEN,
         tts: NotGivenOr[tts.TTS | None] = NOT_GIVEN,
         mcp_servers: NotGivenOr[list[mcp.MCPServer] | None] = NOT_GIVEN,
+        preserve_function_call_history: bool = False,
         # deprecated
         turn_detection: NotGivenOr[TurnDetectionMode | None] = NOT_GIVEN,
         allow_interruptions: NotGivenOr[bool] = NOT_GIVEN,
@@ -733,6 +734,7 @@ class AgentTask(Agent, Generic[TaskResult_T]):
         self.__fut = asyncio.Future[TaskResult_T]()
         self.__inactive_ev = asyncio.Event()
         self.__inactive_ev.set()  # set when the agent is not awaited or activity is closed
+        self._preserve_function_call_history = preserve_function_call_history
 
         self._old_agent: Agent | None = None
 
@@ -892,7 +894,9 @@ class AgentTask(Agent, Generic[TaskResult_T]):
                     run_state._watch_handle(speech_handle)
 
                 merged_chat_ctx = old_agent.chat_ctx.merge(
-                    self.chat_ctx, exclude_function_call=True, exclude_instructions=True
+                    self.chat_ctx,
+                    exclude_function_call=not self._preserve_function_call_history,
+                    exclude_instructions=True,
                 )
                 # set the chat_ctx directly, `session._update_activity` will sync it to the rt_session if needed
                 old_agent._chat_ctx.items[:] = merged_chat_ctx.items


GopalGB:
I have read the CLA Document and I hereby sign the CLA

PR #5292, opened 2026-03-31
fix(core): reduce TTS output buffering latency

theomonnom:
## Summary

Reduces time-to-first-audio (TTFA) in the TTS output pipeline by tackling three compounding buffering layers.

Assuming a TTS provider generating at 2x realtime, the framework buffering overhead drops from **~200ms** to **~30ms** — saving **~170ms** before the end user hears the first audio.

### Before

```
t=0ms:    TTS starts generating
t=100ms:  200ms of audio accumulated → AudioByteStream emits first frame
          → stored in last_frame, nothing sent downstream
t=200ms:  next 200ms of audio ready → second frame triggers sending the first
          → RoomIO re-chunks to 50ms, first audio on wire

Framework TTFA overhead ≈ 200ms
```

### After

```
t=0ms:    TTS starts generating
t=10ms:   20ms of audio ready → first progressive frame emitted
          → split: 10ms sent immediately, 10ms tail held
          → RoomIO buffers 10ms (needs 20ms for first progressive frame)
t=30ms:   next 40ms of audio ready → combined with 10ms tail, 40ms sent
          → RoomIO has 50ms total, emits 20ms progressive first frame on wire

Framework TTFA overhead ≈ 30ms
```

## Changes

- **`AudioByteStream` progressive mode** — new `progressive=True` parameter. First frame is 20ms, each subsequent frame doubles until reaching the target size (`20 → 40 → 80 → 160 → 200ms`). `clear()` resets the progression so TTFA stays low after interruptions. Backward-compatible, defaults to `False`.

- **`AudioEmitter` 10ms tail coalescing** — previously held back an entire frame (200ms) just to know whether it was the last one (`is_final`). Now sends everything except the last 10ms immediately, holds only a 10ms tail. On segment end, flushes the tail with `is_final=True`.

- **RoomIO output** — switched from fixed 50ms re-chunking to progressive mode (`20 → 50ms`). Resets on `clear_buffer()` after interruptions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,6 @@ livekit-plugins-phonic = { workspace = true }
 [tool.uv.workspace]
 members = ["livekit-plugins/*", "livekit-agents"]
 exclude = [
-    "livekit-plugins/community",
     "livekit-plugins/livekit-blingfire",
     "livekit-plugins/livekit-blockguard",
     "livekit-plugins/livekit-durable",


livekit-agents/livekit/agents/tts/tts.py
longcw: since the first a few frames are tiny, we should avoid flush on them
```suggestion
            if flush_if_delayed and sent_duration > 0.2:
```

pyproject.toml
```diff
@@ -67,6 +67,7 @@ livekit-plugins-phonic = { workspace = true }
 [tool.uv.workspace]
 members = ["livekit-plugins/*", "livekit-agents"]
 exclude = [
+    "livekit-plugins/community",
```
davidzhao: remove this?

PR #5293, opened 2026-04-01
(phonic): support realtimemodel say()

tinalenguyen:


qionghuang6:
Just played around with this. Seems to be working!

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -216,3 +216,13 @@ async def aclose(self) -> None: ...
     def start_user_activity(self) -> None:
         """notifies the model that user activity has started"""
         pass
+
+    def say(
+        self,
+        text: str | AsyncIterable[str],
+        *,
+        allow_interruptions: NotGivenOr[bool] = NOT_GIVEN,
```
longcw: should we really expose `allow_interruptions` to this api? we don't have it in `generate_reply`, and disallow `allow_interruptions=False` for realtime session with server side VAD.

even the the api support it in server side VAD, but how can they know the audio playout is finished in agent and re-allow interruption?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -945,6 +938,20 @@ def say(
...
+        if (
+            not is_given(audio)
+            and not self.tts
+            and self._rt_session is None
+            and self._session.output.audio
+            and self._session.output.audio_enabled
+        ):
+            model_info = (
+                "a RealtimeSession that implements say()"
+                if isinstance(self.llm, llm.RealtimeModel)
```
longcw: `llm` cannot be `RealtimeModel` when `rt_session` is None? so maybe you need to add a capability to the RealtimeModel for `say`.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2574,6 +2593,47 @@ def _tool_execution_completed_cb(out: ToolExecutionOutput) -> None:
                 self._agent._chat_ctx.insert(tool_messages)
                 self._session._tool_items_added(tool_messages)
 
+    @utils.log_exceptions(logger=logger)
+    async def _realtime_say_task(
```
longcw: question: should we reuse `_realtime_reply_task`?
```python
    async def _realtime_reply_task(
        self,
        *,
        speech_handle: SpeechHandle,
        model_settings: ModelSettings,
        user_input: str | None = None,
        instructions: str | None = None,
        text: str | AsyncIterable[str] | None = None,
    ) -> None:
```
and we check only one of the user_input and the text can be set.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2574,6 +2593,47 @@ def _tool_execution_completed_cb(out: ToolExecutionOutput) -> None:
...
+            )
+        except NotImplementedError:
+            logger.error(
+                "say() is not implemented for %s; use a TTS model instead",
+                self._rt_session.realtime_model.provider,
+            )
+            return
+        except llm.RealtimeError as e:
+            logger.error("failed to say text: %s", str(e))
+            self._session._update_agent_state("listening")
```
longcw: this is not needed?

theomonnom:
I'm not sure about this, isn't Phonic just using a TTS underneath? This seems like a very specialized method for them?

livekit-agents/livekit/agents/llm/realtime.py
tinalenguyen: ah i see, that's a good point. thanks for pointing that out

livekit-agents/livekit/agents/voice/agent_activity.py
tinalenguyen: great idea! i updated to reuse it

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1044,6 +1037,20 @@ def say(
...
 
+        if (
+            not is_given(audio)
+            and not self.tts
+            and not (isinstance(self.llm, llm.RealtimeModel) and self.llm.capabilities.supports_say)
+            and self._session.output.audio
+            and self._session.output.audio_enabled
+        ):
+            model_info = (
+                "a RealtimeSession that implements say()"
```
longcw: nit: maybe we mention it's trying to generate speech from text without a TTS model or a RealtimeSession that supports say()? so they can add a TTS model for other realtime models to fix this. and we can keep this check on the top

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1025,17 +1025,10 @@ def say(
...
-            and not self.tts
-            and self._session.output.audio
-            and self._session.output.audio_enabled
-        ):
-            raise RuntimeError("trying to generate speech from text without a TTS model")
-
         if (
             isinstance(self.llm, llm.RealtimeModel)
             and self.llm.capabilities.turn_detection
+            and self.tts
```
longcw: why `and self.tts` is needed here?

longcw (commented):
overall looks good to me. something nit:

livekit-agents/livekit/agents/voice/agent_activity.py
tinalenguyen: ah was left over from previous commit

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1044,6 +1036,18 @@ def say(
...
+            not is_given(audio)
+            and not self.tts
+            and not (isinstance(self.llm, llm.RealtimeModel) and self.llm.capabilities.supports_say)
+            and self._session.output.audio
+            and self._session.output.audio_enabled
+        ):
+            raise RuntimeError(
+                "trying to generate speech from text without a TTS model or a RealtimeSession that supports say(); "
+                "add a TTS model to AgentSession to enable say()"
+            )
```
longcw, line 1038: should this part be still on the top of the say method?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1054,18 +1058,29 @@ def say(
             SpeechCreatedEvent(speech_handle=handle, user_initiated=True, source="say"),
         )
 
-        task = self._create_speech_task(
-            self._tts_task(
+        if self._rt_session is not None and not is_given(audio) and not self.tts:
```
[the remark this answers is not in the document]
longcw: > When audio output is disabled (text-only mode), the validation passes and the routing enters the realtime path, calling `self._rt_session.say(text)` which raises NotImplementedError

I think this issue makes sense. originally we allow `session.say` when audio output is disable and no tts is set for text output.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1054,18 +1061,29 @@ def say(
...
-                audio=audio or None,
-                add_to_chat_ctx=add_to_chat_ctx,
-                model_settings=ModelSettings(),
-            ),
-            speech_handle=handle,
-            name="AgentActivity.tts_say",
-        )
-        task.add_done_callback(self._on_pipeline_reply_done)
+                name="AgentActivity.realtime_say",
+            )
```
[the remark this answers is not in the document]
longcw: maybe add a warning when `add_to_chat_ctx` is False for the rt_session.say?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2746,6 +2761,20 @@ async def _realtime_reply_task(
         if speech_handle.interrupted:
             await utils.aio.cancel_and_wait(*authorization_tasks)
 
+        if text is not None:
+            try:
+                generation_ev = await self._rt_session.say(text)
+            except llm.RealtimeError as e:
+                logger.error("failed to say text: %s", str(e))
+                return
```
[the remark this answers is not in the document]
longcw: this also makes sense?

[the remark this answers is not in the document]
tinalenguyen: wouldn't the state have stayed at `"listening"`? is there somewhere upstream for `say() `where it's set to `"thinking"`

i also thought that was what you meant by your comment https://github.com/livekit/agents/pull/5293#discussion_r3025522308

livekit-agents/livekit/agents/voice/agent_activity.py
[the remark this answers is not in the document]
tinalenguyen: i added a warning!

PR #5298, opened 2026-04-01
feat(rime): add mistv3 model support

mcullan:
## Summary

Adds support for the `mistv3` model to `livekit-plugins-rime`.

`mistv3` is a major update to the Mist TTS engine with typical TTFB well below 100ms. It shares the Mist family's pronunciation control and customization features, with a few differences from `mistv2`:

- Drops `reduce_latency` (not supported)
- Default speaker: `cove`

### Changes

- **`models.py`** — adds `"mistv3"` to `TTSModels` Literal; adds `DefaultMistv3Voice`
- **`tts.py`** — adds `_Mistv3Options` dataclass, wires `mistv3` through `__init__`, `update_options`, and `ChunkedStream._run`; adds `MISTV3_MODEL_TIMEOUT = 30`

### Usage

```python
from livekit.plugins import rime

tts = rime.TTS(
    model="mistv3",
    speaker="peak",
)

# With optional Mist params
tts = rime.TTS(
    model="mistv3",
    speaker="peak",
    speed_alpha=1.2,
    phonemize_between_brackets=True,
)
```

## Test plan
- [x] `TTS(model="mistv3")` instantiates correctly, defaults speaker to `"cove"`
- [x] Mist-family params (`speed_alpha`, `pause_between_brackets`, `phonemize_between_brackets`) are passed through to the API payload
- [x] Existing `arcana` and `mistv2` behavior is unchanged

### DIFF
--- a/livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
+++ b/livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
@@ -77,7 +77,7 @@ NUM_CHANNELS = 1
 
 
 def _is_mist_model(model: TTSModels | str) -> bool:
-    return model in ("mistv2", "mistv3")
+    return "mist" in model
 
 
 def _timeout_for_model(model: TTSModels | str) -> int:


tinalenguyen:
hi, thank you for the PR! it seems repetitive to create another class for the Mistv3 options when the only difference is `reduce_latency`, could we use one _MistOptions and use it for both models? same with the timeout and default voice

mcullan:
Hi @tinalenguyen , thanks for the review/suggestions! I made the requested changes, consolidated logic for mistv2 and mistv3. Let me know if there's anything else I can do on this one!

livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
```diff
@@ -76,6 +76,16 @@ class _Mistv2Options:
 NUM_CHANNELS = 1
 
 
+def _is_mist_model(model: TTSModels | str) -> bool:
+    return model in ("mistv2", "mistv3")
```
tinalenguyen: nit: i would change this to `if "mist" in model` to apply to potential future mist models

tinalenguyen (approved):
thanks for iterating, lgtm!

livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
mcullan: fair enough, changed + pushed!

PR #5301, opened 2026-04-01
(taskgroups): set llm to NOT_GIVEN

tinalenguyen:
when set to None, it does not match the previous activity LLM and creates a new connection


theomonnom (approved):
Nice, can you add a comment, it isn't obvious

PR #5303, opened 2026-04-01
(realtimesession): add reset method

tinalenguyen:
providers can implement their own logic for reset if applicable by overriding this method


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -705,26 +705,13 @@ async def _setup_toolset(toolset: llm.Toolset) -> None:
 
             remove_instructions(self._agent._chat_ctx)
 
-            # skip the update if the session is reused and no mid-session update is supported
-            # this means the content is the same as the previous session
             capabilities = self.llm.capabilities
-            if not rt_reused or capabilities.mid_session_instructions_update:
```
longcw, line 711: move this out of agent activity is weird, the assumption is when `rt_reused` is True, either the session supports update the instructions or chat context in the middle, or the content is not changed. this assumption is checked in `AgentActiviy._detach_reusable_resources`.

longcw:
I'll merge this one and modify it on my pr. I guess the use case is they want to send the configuration (instructions, chat context, tools) in a single message, instead of right now in three messages.

PR #5304, opened 2026-04-02
(phonic) Add `min_words_to_interrupt` to Phonic plugin options

qionghuang6:
We released a new setting to Phonic today, and customer that wants to use it is using Livekit agents


tinalenguyen (approved):
lgtm!

PR #5309, opened 2026-04-02
Security: Unvalidated OpenAI base URL override can leak credentials

tuanaiseo:
## Problem

`get_base_url` returns caller/env-provided URLs without validation. If an attacker can influence this value, requests may be sent to attacker-controlled or internal hosts, potentially leaking bearer tokens and enabling SSRF-style access.

**Severity**: `high`
**File**: `livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py`

## Solution

Validate `base_url` against an allowlist of trusted hosts, require `https`, and block loopback/private-network destinations unless explicitly allowed for trusted deployments.

## Changes

- `livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/utils.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
+import ipaddress
 import os
 from collections.abc import Awaitable, Callable
+from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


davidzhao:
> If an attacker can influence this value, requests may be sent to attacker-controlled or internal hosts

this doesn't make sense.. the whole point is to allow developer to point to custom hostnames. an attacker is the developer in this case.

PR #5312, opened 2026-04-02
feat(smallestai): add Pulse STT with real-time streaming and batch transcription

harshitajain165:
## Summary

This PR adds speech-to-text support to the existing `livekit-plugins-smallestai` package via the Smallest AI Pulse STT API, complementing the Lightning TTS integration that already exists.

- **Streaming** (`SpeechStream`): real-time transcription over WebSocket with interim and final transcripts, ~64ms TTFT
- **Batch** (`_recognize_impl`): pre-recorded transcription via HTTP POST
- **Word-level timestamps**: per-word `start`/`end`/`confidence` included by default (`word_timestamps=True`)
- **Speaker diarization**: opt-in via `diarize=True`
- **Configurable end-of-utterance timeout**: `eou_timeout_ms` (100–10,000ms, default 800ms)

## Implementation notes

- Follows the same patterns as other STT plugins in the repo
- API field names (`transcript`, `is_final`, `is_last`, `finalize` message) verified against [docs.smallest.ai](https://docs.smallest.ai)
- `START_OF_SPEECH` is inferred from the first non-empty transcript since the Pulse API does not emit a dedicated speech-start event

## Test plan

- [x] `test_recognize[livekit.plugins.smallestai]` passes
- [x] `test_stream[livekit.plugins.smallestai]` passes
- [x] `ruff format` and `ruff check` pass
- [x] `mypy --strict` passes

### DIFF
--- a/livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/stt.py
+++ b/livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/stt.py
@@ -45,7 +45,7 @@ from .log import logger
 from .models import STTEncoding, STTModels
 
 NUM_CHANNELS = 1
-# Base URL for the Smallest AI Waves API.
+# Base URL for the Smallest AI API.
 # Streaming: wss://api.smallest.ai/waves/v1/{model}/get_text
 # Batch:     https://api.smallest.ai/waves/v1/{model}/get_text
 SMALLEST_STT_BASE_URL = "https://api.smallest.ai/waves/v1"
@@ -133,7 +133,7 @@ class STT(stt.STT):
             api_key: Smallest AI API key. Falls back to the SMALLEST_API_KEY
                 environment variable if not provided.
             http_session: An existing aiohttp ClientSession to reuse.
-            base_url: Override the default Waves API base URL.
+            base_url: Override the default API base URL.
         """
         super().__init__(
             capabilities=stt.STTCapabilities(
--- a/livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/stt.py
+++ b/livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/stt.py
@@ -439,9 +439,7 @@ class SpeechStream(stt.SpeechStream):
             "eou_timeout_ms": self._opts.eou_timeout_ms,
         }
         ws_url = (
-            self._opts.base_url.replace("https://", "wss://", 1).replace(
-                "http://", "ws://", 1
-            )
+            self._opts.base_url.replace("https://", "wss://", 1).replace("http://", "ws://", 1)
             + f"/{self._opts.model}/get_text"
             + f"?{urlencode(params)}"
         )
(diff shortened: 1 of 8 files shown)


tinalenguyen (commented):
hi, thank you for the PR! i have a few notes, could you:
- address all of the devin comments, especially the one regarding interim transcripts
- remove smallest ai from the test files, as we do not have a smallestai api key for testing as of yet
- sign the CLA if possible

harshitajain165:
Hey @tinalenguyen 

Thanks for the comment. I'm addressing the devin comments, removing smallest ai from test files and signing the CLA. Will keep you posted once all are done

harshitajain165:
recheck

harshitajain165:
Hey @tinalenguyen 

The devin comments have been incorporated, smallest ai has been removed from test files and I have signed the CLA too. Please feel free to re-review/take this forward.

tinalenguyen:
@harshitajain165 Thank you for iterating on the feedback! 

For the STT, I printed out the received events and it does seem that interim results are emitted. Is there a setting to pass to the API or does the API always send interim results? If that is always the case, I would set `interim_results` to `True` and disregard the devin comment. I also noticed that for the final transcript, there are often spaces in the beginning. Is that expected? 

Also, when testing the TTS, I keep facing this error:
`failed to synthesize speech: message='Bad Request (400)', status_code=400, retryable=False, retrying in 2.0s
`

livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/stt.py
```diff
@@ -0,0 +1,624 @@
...
+    def __init__(
+        self,
+        *,
+        model: STTModels | str = "pulse",
+        language: str = "en",
+        sample_rate: int = 16000,
+        encoding: STTEncoding | str = "linear16",
+        word_timestamps: bool = True,
+        diarize: bool = False,
+        eou_timeout_ms: int = 800,
```
theomonnom: Should this be 0?

With our end-of-turn detection model, we should prioritize minimizing latency to receive transcripts.

harshitajain165: Thanks for catching this! Changed the default to 0 and also updated the code to omit the parameter from the WebSocket query string entirely when it's 0, since sending 0 causes the API to reject the connection.

harshitajain165:
> @harshitajain165 Thank you for iterating on the feedback!
> 
> For the STT, I printed out the received events and it does seem that interim results are emitted. Is there a setting to pass to the API or does the API always send interim results? If that is always the case, I would set `interim_results` to `True` and disregard the devin comment. I also noticed that for the final transcript, there are often spaces in the beginning. Is that expected?
> 
> Also, when testing the TTS, I keep facing this error: `failed to synthesize speech: message='Bad Request (400)', status_code=400, retryable=False, retrying in 2.0s `

Thanks for testing this @tinalenguyen! Here's an update on each point:                    
                                                                       
- Interim results — The Pulse API always sends interim transcripts; there's no client-side flag to control it. I've updated STTCapabilities to interim_results=True to correctly reflect that.                             
                                                                      
- Leading spaces in final transcripts — Yes, this is expected streaming behaviour. The model uses look-ahead to confirm punctuation, which means it arrives attached to the next segment.
 
- TTS 400 Bad Request — This was caused since the default voice was _emily_, which is only valid for lightning-v2 and not for lightning-v3.1, changed the default to _sophia_. This should be fixed now.

harshitajain165:
Hey @tinalenguyen @theomonnom 

Re-requesting review, the changes have been made. Thanks

tinalenguyen (approved):
lgtm, thank you!!

PR #5320, opened 2026-04-03
fix xAI realtime update chat ctx

longcw:
this pr fixed issues
- `conversation.item.added` event always has the previous_item_id as None, causing the latest item being inserted at the top
- `conversation.item.deleted` event has item_id empty, causing "failed to delete item" error, in this fix we assume it's the first item in to queue
- input audio transcript was duplicated in the rt_session's chat item's content

### DIFF
--- a/livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/realtime/realtime_model.py
@@ -1,7 +1,5 @@
-import asyncio
 import os
 import time
-from collections import OrderedDict
 from typing import Any
 
 import aiohttp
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/realtime/realtime_model.py
```diff
@@ -87,6 +96,9 @@ def __init__(self, realtime_model: RealtimeModel) -> None:
         self._xai_model: RealtimeModel = realtime_model
         self._session_connected_at: float = 0.0
 
+        # keep the order of item deletion futures
+        self._item_delete_future = OrderedDict[str, asyncio.Future]()
```
chenghao-mou: nit: we don't this override for Python 3.7+

chenghao-mou (approved):

PR #5323, opened 2026-04-03
feat(azure-stt): Possibility to change segmentation options during a call

rafallezanko:


rafallezanko:
Hi @chenghao-mou , I think it is a pretty straightforward change, so can you have a look? 😄 Thank you in advance!

PR #5328, opened 2026-04-03
Feature - Configurable session close transcript timeout

bml1g12:
### Summary

Adds **`session_close_transcript_timeout`** to **`AgentSession`** / **`AgentSessionOptions`**, defaulting to **`DEFAULT_COMMIT_USER_TURN_TRANSCRIPT_TIMEOUT`** (same module-level default as **`commit_user_turn`**’s **`transcript_timeout`**). On session close, after inputs are detached, the pipeline commits the user turn with **`audio_detached=True`** and uses this value for the final-transcript wait instead of a hardcoded timeout.

Introduces **`DEFAULT_COMMIT_USER_TURN_TRANSCRIPT_TIMEOUT`** in **`audio_recognition`** (`2.0` seconds) as the single default for “wait for final transcript” on **`AgentSession`**. 
### Motivation

In **push-to-talk** flows, users often record **long stretches of speech** before a turn is committed, so STT can take **longer than a short default** to return a final transcript (similar to [`examples/voice_agents/push_to_talk.py`](examples/voice_agents/push_to_talk.py), where **`transcript_timeout`** may need to be raised above the default). If the session ends while that final transcript is still in flight, we risk **closing before the transcript lands**. Exposing **`session_close_transcript_timeout`** lets apps **match close-time behavior** to STT latency so teardown still **captures the last user text** when appropriate.

### How to test

- Instantiate **`AgentSession`** with a higher **`session_close_transcript_timeout`**, run a push-to-talk-style session with a long utterance, end the session, and confirm the final transcript shows up in context/events as expected..
- **`uv run ruff check`** and relevant tests under **`livekit-agents`** (e.g. **`tests/test_agent_session.py`**) as you usually do for this package.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -44,7 +44,6 @@ from ._utils import _set_participant_attributes
 from .agent import Agent, AgentTask
 from .agent_activity import AgentActivity, _ReusableResources
 from .amd import AMD
-from .audio_recognition import DEFAULT_COMMIT_USER_TURN_TRANSCRIPT_TIMEOUT
 from .events import (
     AgentEvent,
     AgentState,
@@ -142,7 +141,7 @@ class AgentSessionOptions:
     tts_text_transforms: Sequence[TextTransforms] | None
     ivr_detection: bool
     aec_warmup_duration: float | None
-    session_close_transcript_timeout: float = DEFAULT_COMMIT_USER_TURN_TRANSCRIPT_TIMEOUT
+    session_close_transcript_timeout: float
 
     @property
     def endpointing(self) -> EndpointingOptions:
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -141,6 +142,7 @@ class AgentSessionOptions:
     tts_text_transforms: Sequence[TextTransforms] | None
     ivr_detection: bool
     aec_warmup_duration: float | None
+    session_close_transcript_timeout: float = DEFAULT_COMMIT_USER_TURN_TRANSCRIPT_TIMEOUT
```
longcw:
```suggestion
    session_close_transcript_timeout: float
```

longcw (approved):
lgtm! nit: for simplicity maybe we can use the literal 2.0 as the default without introducing a global var. The default timeout for session close and `commit_user_turn` could be different.

livekit-agents/livekit/agents/voice/agent_session.py
bml1g12: OK, Thanks I have made that change

PR #5337, opened 2026-04-04
feat(inference): add debug/identification headers to inference requests

adrian-cowham:
## Summary
- Add `User-Agent`, `X-LiveKit-Room-ID`, and `X-LiveKit-Job-ID` headers to all inference module HTTP and WebSocket requests
- Define header constants and `get_inference_headers()` builder in `inference/_utils.py`
- Update all 4 call sites: `tts.py`, `stt.py`, `llm.py`, `interruption.py` (both HTTP POST and WebSocket)
- Headers enable server-side debugging, request tracing, and usage attribution
- Room/Job headers are only included when running inside a job context (gracefully omitted in console mode or tests)

## Test plan
- [x] `make check` passes (format, lint, type-check)
- [x] Unit tests pass (107 passed, 1 skipped)
- [ ] Manual: run agent in console mode → headers include only `User-Agent`
- [ ] Manual: run agent connected to a room → headers include all three fields

### DIFF
--- a/livekit-agents/livekit/agents/inference/interruption.py
+++ b/livekit-agents/livekit/agents/inference/interruption.py
@@ -49,6 +49,7 @@ from ._utils import (
     STAGING_INFERENCE_URL,
     create_access_token,
     get_default_inference_url,
+    get_inference_headers,
 )
 
 SAMPLE_RATE = 16000
@@ -1126,6 +1127,7 @@ class InterruptionWebSocketStream(InterruptionStreamBase):
         if base_url.startswith(("http://", "https://")):
             base_url = base_url.replace("http", "ws", 1)
         headers = {
+            **get_inference_headers(),
             "Authorization": f"Bearer {create_access_token(self._opts.api_key, self._opts.api_secret)}",
         }
         try:


livekit-agents/livekit/agents/inference/interruption.py
```diff
@@ -766,6 +767,7 @@ async def predict(self, waveform: np.ndarray) -> InterruptionResponse:
             async with self._session.post(
                 url=f"{self._opts.base_url}/bargein?threshold={self._opts.threshold}&min_frames={int(self._opts.min_frames)}&created_at={int(created_at)}",
                 headers={
+                    **get_inference_headers(),
```
chenghao-mou: I don't think we need this here since this is for Infserve.

livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -345,6 +345,9 @@ async def _run(self) -> None:
                 headers = self._extra_kwargs.setdefault("extra_headers", {})
                 headers["X-LiveKit-Inference-Provider"] = self._provider
 
+            extra_headers = self._extra_kwargs.setdefault("extra_headers", {})
+            extra_headers.update(get_inference_headers())
```
chenghao-mou: nit: maybe we can merge 344-346 into this as well:

```python
extra_headers = self._extra_kwargs.setdefault("extra_headers", {})
extra_headers.update(get_inference_headers())
if self._provider:
    extra_headers["X-LiveKit-Inference-Provider"] = self._provider
```

livekit-agents/livekit/agents/inference/interruption.py
adrian-cowham: Done — removed `get_inference_headers()` from both the HTTP POST and WebSocket paths in `interruption.py`, since bargein goes to Infserve.

livekit-agents/livekit/agents/inference/llm.py
adrian-cowham: Done — consolidated into a single `setdefault` call with inference headers first, then provider header conditionally.

Issue #5260, opened 2026-03-28
generate_reply times out on first dialogue with gemini-3.1-flash-live-preview — works fine with gemini-2.5-flash-


acharan-tech-200037:
### Bug Description

When using gemini-3.1-flash-live-preview as the realtime model, the agent fails to generate the first reply immediately after joining the room. It times out waiting for the generation_created event. The exact same code with gemini-2.5-flash works without any issues.

### Expected Behavior

Agent should generate and speak the first reply immediately after joining the room, same as it does with gemini-2.5-flash

### Reproduction Steps

```bash
1.Create an AgentSession with google.realtime.RealtimeModel using model gemini-3.1-flash-live-preview
2.Start the session with session.start()
3.Call session.generate_reply() for the opening dialogue

```

### Minimal Reproducible Example
```bash
from livekit import agents
from livekit.plugins import google, silero
from livekit.agents import AgentSession

@server.rtc_session(agent_name="my-agent")
async def my_agent(ctx: agents.JobContext):
    session = AgentSession(
        llm=google.realtime.RealtimeModel(
            model="gemini-3.1-flash-live-preview",  # ❌ fails
            voice="Puck",
        ),
        vad=silero.VAD.load(),
    )

    await session.start(ctx.room, participant=ctx.room.remote_participants)
    await session.generate_reply(
        instructions="Greet the user."
    )
```

### Operating System

Windows11

### Models Used

gemini-3.1-flash-live-preview

### Package Versions

```bash
livekit-agents==1.5.1
```

### Logs
```bash
ERROR… livekit.…ns.google  error in receive task: 1007 None.
                             Request contains an invalid argument.

websockets.exceptions.ConnectionClosedError:
  received 1007 (invalid frame payload data) Request contains an invalid argument.;
  then sent 1007 (invalid frame payload data) Request contains an invalid argument.

google.genai.errors.APIError: 1007 None. Request contains an invalid argument.

ERROR… livekit.agents  failed to generate a reply:
                        generate_reply timed out waiting for
                        generation_created event.

WARNI… livekit.…ns.google  received server content but no active generation.
```

oEdyb:
## Root cause analysis

Confirmed on `livekit-plugins-google==1.5.1` + `google-genai==1.68.0`.

### What's happening

`RealtimeSession.generate_reply()` (line 715-716 in `realtime_api.py`) uses `LiveClientContent` (i.e. `send_client_content`) to inject a placeholder `"."` user turn:

```python
[...]

PR #5344, opened 2026-04-05
fix(google): route tool responses by capability, not model name

leonobitech:
## Summary

Fixes tool calling with `gemini-3.1-flash-live-preview` where the agent stops responding after `function_tool` execution.

### Root cause

`update_chat_ctx` was completely disabled for Gemini 3.1 via an early return, because `send_client_content` is not supported mid-session on 3.1+. However, this also silently dropped **tool responses** (`send_tool_response`), which are a separate API method and still work on 3.1.

The model never received the tool result, timed out after ~7-12s, and sent `LiveServerToolCallCancellation`.

### Timeline from production logs

```
t=14.231  user stops speaking
t=14.265  TOOL:START — executes in 5ms
t=14.265  TOOL:END — returns result
t=14.265  update_chat_ctx → early return (not compatible) ← tool response dropped
          ... 7 seconds of silence ...
t=21.581  user speaks again
t=23.071  server cancelled tool calls
```

### Fix

Instead of hardcoding `if model == "gemini-3.1-flash-live-preview"`, this PR adds a `supports_client_content` capability to `RealtimeCapabilities` and routes by capability:

- **Tool responses** → always sent via `send_tool_response` regardless of model
- **Chat turns** → sent via `send_client_content` only when the capability is `True`, skipped otherwise
- **No model name checks** in routing logic — fully capability-driven

### Why capability-based

Google separated their API into distinct methods (`send_client_content`, `send_realtime_input`, `send_tool_response`), each with different compatibility across model versions. The transport layer should route based on what the model supports, not based on model name strings. Future models only need to declare their capabilities.

### Changes

| File | Change |
|------|--------|
| `livekit-agents/.../llm/realtime.py` | Add `supports_client_content: bool = True` to `RealtimeCapabilities` |
| `livekit-plugins-google/.../realtime_api.py` | Set capability based on model version, refactor `update_chat_ctx` to route by capability |

### Testing

Tested in production with `gemini-3.1-flash-live-preview` + `livekit-agents==1.5.1`:
- Function tools execute and Gemini continues speaking with the results ✅
- No `server cancelled tool calls` warnings ✅
- Data tracks sent from tools arrive at the frontend correctly ✅
- Regular conversation (without tools) unaffected ✅

Related: #5260

### DIFF
--- a/livekit-agents/livekit/agents/llm/realtime.py
+++ b/livekit-agents/livekit/agents/llm/realtime.py
@@ -72,6 +72,10 @@ class RealtimeCapabilities:
     """Whether the tools can be updated mid-session"""
     per_response_tool_choice: bool = False
     """Whether the tool and tool choice can be specified per response"""
+    supports_client_content: bool = True
+    """Whether the model supports send_client_content for mid-session chat context
+    updates. When False, chat turns should be sent via send_realtime_input and tool
+    responses via send_tool_response independently."""
 
 
 class RealtimeError(Exception):
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -267,6 +267,12 @@ class RealtimeModel(llm.RealtimeModel):
         if not is_given(output_audio_transcription):
             output_audio_transcription = types.AudioTranscriptionConfig()
 
+        if not is_given(model):
+            if vertexai:
+                model = "gemini-live-2.5-flash-native-audio"
+            else:
+                model = "gemini-2.5-flash-native-audio-preview-12-2025"
+
         server_turn_detection = True
         if (
             is_given(realtime_input_config)
@@ -275,6 +281,10 @@ class RealtimeModel(llm.RealtimeModel):
         ):
             server_turn_detection = False
         modalities = modalities if is_given(modalities) else [types.Modality.AUDIO]
+
+        # Gemini 3.1+ does not support send_client_content mid-session.
+        # Tool responses and chat turns must use their own dedicated API methods.
+        supports_client_content = "3.1" not in model
 
         super().__init__(
             capabilities=llm.RealtimeCapabilities(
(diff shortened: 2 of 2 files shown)


tinalenguyen:
Hi, thank you for creating this PR! The discrepancy in sending content between models has been raised and may be resolved in the near future. With this in mind, I am not sure if adding a capability field is the best way to resolve this. 

Since the capability flag is determined by the models containing "3.1" anyway, perhaps we can still filter by that and add a comment?

FMFigueroa:
Hey @tinalenguyen, thanks for the review!                                                                                                            
                                                                                                                                                                                                        I'd like to clarify something — `send_tool_response` has never been broken on Google's side. It works correctly on all model versions, including 3.1. It's a completely separate API method from `send_client_content`.                                                                                   
                                                                                                                                                       
The issue is that LiveKit's early return in `update_chat_ctx` blocks both methods, even though only `send_client_content` is unsupported on 3.1. Tool responses get silently dropped, and the model cancels the call after ~7 seconds of silence.                                                         
                                                                                                                                                       
What exactly would Google need to resolve here? The three methods (`send_client_content`, `send_tool_response`, `send_realtime_input`) are independent by design — they just shouldn't be gated by the same condition.                                                                          
                                                                                                                                                                                                 Happy to simplify the implementation approach — but the core issue remains: tool responses should never be blocked by a `send_client_content` guard.

tinalenguyen:
hi @FMFigueroa and @leonobitech, thank you very much for the helpful context! the discrepancy i mentioned was regarding `send_client_content`, which seems to be an API regression for the newer 3.1 model. you are right that `send_tool_response` is a separate method still supported by all models.

that said, i think the best way to approach this is by denoting mutability in `RealtimeCapabilities` rather than `send_client_content` compatibility. i put up #5413, please feel free to take a look and let me know your thoughts!

leonobitech:
Closing in favor of #5413 — better approach scoping the fix to the plugin level instead of touching core. Thanks @tinalenguyen for the collaboration!

tinalenguyen:
glad to work with you :)

PR #5349, opened 2026-04-06
fix: align WebSocket client with xAI streaming TTS

RishiikeshBalaji:
# xAI TTS WebSocket protocol fix

Summary
-------
This change updates the bundled xAI TTS plugin to match the current xAI streaming Text‑to‑Speech WebSocket contract. The previous implementation used a legacy/on‑vendor wire format which caused HTTP 400 handshake failures and resulted in "no audio frames were pushed" at runtime.

What changed
------------
- Build the WebSocket URL with required query parameters (language, voice, codec, sample_rate) when opening the socket instead of relying on a post‑connect "config" frame.
- Stop sending the legacy `config` JSON frame.
- Replace outbound frames from `text_chunk`/is_last → `text.delta` + `text.done`.
- Parse inbound frames according to the current spec: `audio.delta` (base64 in `delta`) and `audio.done`. Handle `error` frames.
- Propagate `APIStatusError` and `APIConnectionError` so callers receive meaningful errors instead of generic connection exceptions.
- Ensure each synthesize stream uses its own options (voice/language) when connecting.

Motivation / rationale
---------------------
The xAI docs require configuration at connection time (query params) and define a delta/done streaming contract. The previous code used a different on‑the‑wire protocol which led to handshake rejections (HTTP 400) and mismatched event parsing, so audio never reached the LiveKit emitter.

Testing performed
-----------------
- `make check` (format, ruff lint, mypy) — passed locally.
- Manual verification: built the WebSocket URL and validated the new message flow locally (no external API calls in CI).

Notes for release
-----------------
- This is a non‑breaking change for callers of the LiveKit TTS interface, but it changes the internal plugin behavior for xAI streaming sessions.
- If your deployment still fails after this change, include the `livekit-plugins-xai` version, the voice/language used, and the exact `APIStatusError` / `APIConnectionError` messages when filing a bug.
- The `.venv` modification noted during earlier investigation is not committed; the source change is in this repo under `livekit-plugins-xai`.


theomonnom:
superseded by https://github.com/livekit/agents/pull/5350

Issue #5351, opened 2026-04-06
feat: livekit-plugins-hf local HuggingFace LLM plugin with KV cache compression


mahimairaja:
### Feature Type

Would make my life easier

## Summary

Proposing a LiveKit Agents SDK LLM plugin (`livekit-plugins-hf`) that runs HuggingFace transformer models locally on GPU, with optional TurboQuant-GPU KV cache compression for ~5x memory reduction per concurrent session.

## Motivation

[OpenRTC](https://github.com/mahimairaja/openrtc-python) pools multiple voice agents in a single worker, but all current LLM providers are cloud APIs (OpenAI, Groq, Anthropic). A local inference plugin would:

- Eliminate per-token API costs for high-volume deployments
- Keep data on-prem for privacy-sensitive use cases
- Enable KV cache compression via TurboQuant-GPU allowing ~5x more concurrent sessions per GPU (e.g., 6 → 31 on a 24GB GPU with a 7B model)

## Feasibility (Validated)

The LiveKit Agents SDK `LLM` base class fully supports this:

- **Custom generation loop**: `LLMStream._run()` is abstract plugin controls the entire generation, just pushes `ChatChunk` objects to `self._event_ch`
- **Shared model instance**: `LLM` is stateless; each `chat()` call creates an independent `LLMStream` with its own channel/task  a single `LLM` holding model weights can serve concurrent sessions
- **Token streaming**: Push `ChatChunk(delta=ChoiceDelta(content="token"))` per generated token; final chunk includes `CompletionUsage`

## Proposed Scope

- Subclass `LLM` and `LLMStream` from `livekit.agents.llm`
- Load any HuggingFace `AutoModelForCausalLM` in `__init__` / `prewarm()`
- Implement `_run()` to run `model.generate()` and stream tokens via `_event_ch`
- Optional TurboQuant-GPU integration to compress `past_key_values` between generation steps
- Auto-fallback to PyTorch when cuTile GPU kernels aren't available

## References
- TurboQuant-GPU: [`DevTechJr/turboquant-gpu`](https://github.com/DevTechJr/turboquant-gpu)

PR #5352, opened 2026-04-06
Feat: hf llm plugin

mahimairaja:
# What does this PR does?

- Adds new plugin to load any llm from hugging face models
- Adds turbo quant optimization for local models

#5351


theomonnom:
Thanks! Can you address the Devin PR reviews?

Also do you think there is a way to avoid downloading the model at runtime? (vs the `download-files` CLI)

PR #5355, opened 2026-04-06
feat: add Runway Characters avatar plugin

robinandeer:
### Summary

Adds `livekit-plugins-runway` for [Runway Characters](https://dev.runwayml.com/) avatar integration.

- Plugin: `livekit-plugins-runway` with `AvatarSession`
- Example agent: `examples/avatar_agents/runway/`
- Follows the same pattern as Tavus/Bey avatar plugins

[Runway Characters API docs](https://dev.runwayml.com/docs/characters/overview)


livekit-plugins/livekit-plugins-runway/livekit/plugins/runway/avatar.py
```diff
@@ -0,0 +1,183 @@
...
+        await self._create_session(livekit_url, livekit_token, room.name)
+
+        agent_session.output.audio = DataStreamAudioOutput(
+            room=room,
+            destination_identity=self._avatar_participant_identity,
+            wait_remote_track=rtc.TrackKind.KIND_VIDEO,
+            sample_rate=SAMPLE_RATE,
+        )
+
+    async def _create_session(self, livekit_url: str, livekit_token: str, room_name: str) -> None:
```
tinalenguyen, line 126: i'm wondering if we need `assert` or `isinstance` to check types, we check in `__init__` before calling this method

tinalenguyen (commented):
Thank you for the PR, I can't wait to try it! I left a comment, could you also:
- remove `requirements.txt`
- add the plugin to the outer [pyproject.toml](https://github.com/rimelabs/livekit-agents/blob/main/pyproject.toml) and the agents[ pyproject.toml](https://github.com/rimelabs/livekit-agents/blob/main/livekit-agents/pyproject.toml)
- bump the version to latest, this makes it easier to update
- remove the `uv.lock` changes

livekit-plugins/livekit-plugins-runway/livekit/plugins/runway/avatar.py
robinandeer: Good call — removed the asserts. Restructured to build the `avatar` dict eagerly in `__init__` (right after the validation) so mypy can see the types are already narrowed. No runtime checks needed in `_create_session` anymore.

PR #5394, opened 2026-04-09
fix(openai): prepend session instructions in realtime generate_reply

longcw:
## Summary
- In OpenAI realtime API, per-response `instructions` completely replace session-level instructions. This means when `generate_reply(instructions=...)` is called, the agent loses its system prompt.
- This fix prepends the session-level instructions to the per-response instructions so `generate_reply` behaves consistently with pipeline agents.
- Addresses https://github.com/livekit/agents/pull/5287#issuecomment-4207711957


davidzhao (approved):
lgtm

cchen-dialpad:
@longcw thanks for the quick fix!

PR #5405, opened 2026-04-10
fix(tests): update drive thru instructions

chenghao-mou:


davidzhao (approved):
thanks for fixing this

PR #5430, opened 2026-04-13
feat(sarvam): expose max_session_duration as TTS constructor parameter

### LINKED ISSUE #5268: [sarvam] Expose max_session_duration as a constructor parameter to control WebSocket connection reuse
### Feature Type

Would make my life easier

### Feature Description

Currently the WebSocket connection pool has max_session_duration hardcoded to 3600 seconds, meaning connections are aggressively reused across TTS requests. Setting this to 0 (no reuse) results in a fresh WebSocket session per request, which noticeably improves TTS audio quality  (likely because stale session state or cached packet context on the Sarvam server side is avoided)

The only current workaround is reaching into a private attribute after construction:

```python
tts._pool._max_session_duration = 0
```

This is fragile and not part of the public API.

## Proposed Solution
Add a max_session_duration parameter to the TTS constructor:

```python
OurTTS = sarvam.TTS(
    target_language_code="hi-IN",
    model="bulbul:v3",
    speaker="ritu",
    max_session_duration=0,  # fresh connection per request
)
```

And pass it through to the pool:
```python
self._pool = utils.ConnectionPool(
    ...
    max_session_duration=max_session_duration,
)
```

### Additional Context

This was discovered while investigating the WAV codec bug (see issue #5267 ). 

During debugging, max_session_duration was manually set to 0 as a side effect, and TTS output quality was subjectively noticeably better. Speech sounded cleaner with less audio artifacting between utterances.

It is possible that the Sarvam WebSocket server maintains some internal state or packet buffer across a reused connection that negatively affects subsequent requests. A fresh connection appears to reset this.

Worth asking: should max_session_duration default to 0 for the Sarvam plugin specifically, rather than inheriting the 3600 default that may be more appropriate for other providers? Connection overhead for a TTS request is likely negligible compared to the quality tradeoff.

KeWang0622:
## Summary

Closes #5268

The Sarvam TTS plugin's WebSocket connection pool hardcodes `max_session_duration` to 3600 seconds, making it impossible to control connection reuse without reaching into private attributes (`tts._pool._max_session_duration = 0`).

This PR exposes `max_session_duration` as a public constructor parameter on `sarvam.TTS`:

```python
tts = sarvam.TTS(
    target_language_code="hi-IN",
    model="bulbul:v3",
    speaker="ritu",
    max_session_duration=0,  # fresh connection per request
)
```

**Changes:**
- Add `max_session_duration` field to `SarvamTTSOptions` dataclass (default: `3600`)
- Add `max_session_duration` parameter to `TTS.__init__()` with validation (`>= 0`)
- Pass the value through to `utils.ConnectionPool` instead of the hardcoded `3600`
- Update docstrings for both the dataclass and the constructor

**Behavior:**
- `max_session_duration=3600` (default) — existing behavior, connections reused for up to 1 hour
- `max_session_duration=0` — fresh WebSocket per request, no reuse (may improve audio quality as noted in the issue)
- `max_session_duration=N` — custom reuse window in seconds

No breaking changes — the default matches the previous hardcoded value.

## Test plan

- [ ] Verify `sarvam.TTS(target_language_code="hi-IN", max_session_duration=0)` creates a fresh connection per request
- [ ] Verify default behavior (no `max_session_duration` arg) is unchanged
- [ ] Verify `max_session_duration=-1` raises `ValueError`
- [ ] `ruff check` and `ruff format` pass

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
@@ -295,6 +295,7 @@ class SarvamTTSOptions:
         base_url: API endpoint URL
         ws_url: WebSocket endpoint URL
         word_tokenizer: Tokenizer for processing text
+        max_session_duration: Maximum WebSocket session reuse duration in seconds (0 = fresh per request)
     """
 
     target_language_code: SarvamTTSLanguages | str  # BCP-47 for supported Indian languages
@@ -318,6 +319,7 @@ class SarvamTTSOptions:
     word_tokenizer: tokenize.tokenizer.SentenceTokenizer | None = None
     send_completion_event: bool = True
     output_audio_codec: str = "mp3"
+    max_session_duration: float = 3600
 
 
 class TTS(tts.TTS):
@@ -347,6 +349,10 @@ class TTS(tts.TTS):
         ws_url: WebSocket endpoint URL
         http_session: Optional aiohttp session to use
         output_audio_codec: Optionally choose the output codec format (mp3)
+        max_session_duration: Maximum WebSocket connection reuse duration in seconds.
+            Connections are recycled after this period. Set to 0 for a fresh
+            connection per request, which may improve audio quality with some
+            providers. Defaults to 3600 (1 hour).
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


KeWang0622:
CLA signed via https://cla-assistant.io/livekit/agents?pullRequest=5430

PR #5432, opened 2026-04-13
docs: clarify RunResult testing surface

### LINKED ISSUE #5410: Feature Request: document a small testing-result surface around `voice.testing.RunResult.events`
### Feature Type

Would make my life easier

### Feature Description

We just shipped a small sample on the Assay side that consumes a frozen artifact derived from LiveKit Agents' documented `voice.testing.RunResult` surface:

https://github.com/Rul1an/assay/tree/main/examples/livekit-runresult-evidence

The goal is narrow: consume the smallest honest testing/result surface without turning it into telemetry, session analytics, room-state export, transcript export, or raw audio export.

Right now the smallest shape that seems to work is roughly:

- `RunResult.events` as the primary seam
- `outcome`, `timestamp`, `task_label`
- optional final output only when it is naturally present
- only typed events such as `message`, `function_call`, `function_call_output`, and `agent_handoff`

What would help is one of these:

- document that this is roughly the recommended small testing-result surface for external consumers
- point to a thinner official testing/result surface if one already exists
- expose a smaller typed helper/export on top of `RunResult` for this use case

That would make it much easier to build deterministic external consumers on top of `voice.testing` without depending on broader runtime or observability semantics.

Our current workaround is to keep a docs-backed frozen artifact on our side and intentionally ignore telemetry, room/session details, transcripts, and audio payloads:

https://github.com/Rul1an/assay/tree/main/examples/livekit-runresult-evidence

That works for a sample, but it means we are inferring the right seam rather than following an explicitly recommended one.

### Additional Context

Related testing/result issues that seem adjacent:

- #4662
- #3149
- #3162

This request is not asking for:

- session analytics
- room metrics
- transcript export
- raw audio export

It is only about a small testing-result-first seam for external consumers.

nightcityblade:
Fixes #5410

## Summary
- document `RunResult.events` as the primary minimal testing-facing surface
- clarify that events stay typed and chronological for external consumers
- note that `final_output` is optional and best treated as secondary

## Testing
- python3 -m compileall livekit-agents/livekit/agents/voice/run_result.py livekit-agents/livekit/agents/voice/agent_session.py

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -536,6 +536,16 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         input_modality: Literal["text", "audio"] = "text",
         output_type: type[Run_T] | None = None,
     ) -> RunResult[Run_T]:
+        """Run a single user turn and capture its testing result surface.
+
+        The returned `RunResult` is meant for tests and deterministic external
+        consumers. In most cases, `RunResult.events` should be treated as the primary
+        seam: it records typed events in chronological order and is usually sufficient
+        without depending on broader runtime or observability surfaces.
+
+        `RunResult.final_output` can be useful when an agent naturally produces one,
+        but it is optional and should be treated as a secondary convenience.
+        """
         if self._global_run_state is not None and not self._global_run_state.done():
             raise RuntimeError("nested runs are not supported")
 
--- a/livekit-agents/livekit/agents/voice/run_result.py
+++ b/livekit-agents/livekit/agents/voice/run_result.py
@@ -79,7 +79,16 @@ class RunResult(Generic[Run_T]):
 
     @property
     def events(self) -> list[RunEvent]:
-        """List of all recorded events generated during the run."""
+        """List of recorded events generated during the run.
+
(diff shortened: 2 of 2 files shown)


theomonnom:
lgtm! but AgentSession.run is not only for testing. It can be used to build text apps

nightcityblade:
Good call, thanks. I agree `AgentSession.run()` is broader than a testing-only entrypoint. I opened a narrower follow-up docs PR around the small testing surface on `RunResult.events` instead of framing `run()` itself as test-only.

theomonnom:
Why not updating the same PR?

nightcityblade:
I split it out because the original comment changed the premise of this PR: the issue was not with `AgentSession.run()` itself, but with how narrowly I wanted to document the testing surface around `RunResult.events`. Keeping that as a small follow-up made the diff easier to review and avoided rewriting this PR into a different scope. If you’d rather I fold the narrower doc change back into this PR instead, I can do that.

nightcityblade:
That makes sense. I split it initially to keep this PR from drifting after the original framing changed, but if you would rather consolidate it here I can still reopen the docs branch and fold the narrower `RunResult.events` wording into this PR instead of leaving it as a separate follow-up. Happy to do whichever is easier for you to review.

PR #5433, opened 2026-04-13
Fix FrameProcessor lifecycle for selector based noise cancellation

Topherhindman:
## Summary

This PR fixes the lifecycle for `FrameProcessor`-based input noise cancellation when `AudioInputOptions.noise_cancellation` is provided as a `NoiseCancellationSelector`.

Before this change, if a selector returned a `FrameProcessor` (for example `ai_coustics`), the processor was passed to `rtc.AudioStream.from_track()` but was not stored as the active input processor. As a result, it never received `_on_stream_info_updated()` or `_on_credentials_updated()`, which caused processors that depend on that configuration to run uninitialized.

While fixing that, this PR also addresses two related lifecycle gaps:
- closed processors could remain referenced after teardown and still receive later lifecycle calls
- active processors were not deterministically closed on final input shutdown

## Root cause

`_ParticipantAudioInputStream.__init__()` only initialized `self._processor` when `noise_cancellation` was passed directly as a `FrameProcessor`.

When `noise_cancellation` was a selector, the resolved value was only known later in `_create_stream()`, but that resolved processor was not stored on the input stream. This broke the path in `_on_track_available()` that injects stream info and credentials into the active processor.

## Changes

- store selector-returned `FrameProcessor`s as the active processor in `_create_stream()`
- clear stale processor references in `_close_stream()`
- clear stale processor state when a selector returns a non-`FrameProcessor` value
- close the active processor during `aclose()` for deterministic shutdown

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/_input.py
+++ b/livekit-agents/livekit/agents/voice/room_io/_input.py
@@ -287,6 +287,9 @@ class _ParticipantAudioInputStream(_ParticipantInputStream[rtc.AudioFrame], Audi
             else self._noise_cancellation
         )
 
+        if isinstance(noise_cancellation, rtc.FrameProcessor):
+            self._processor = noise_cancellation
+
         return rtc.AudioStream.from_track(
             track=track,
             sample_rate=self._sample_rate,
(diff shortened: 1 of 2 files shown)


longcw (approved):
looks good to me!

PR #5437, opened 2026-04-13
Fix Missing user message metrics (MetricsReport) due to early returns in _user_turn_completed_task and no initialization in on_end_of_turn

hudson-worden:
There appears to be a few early returns where `self._session._conversation_item_added(user_message)` is called before adding available `MetricsReport` values to the chat item. `info` is available at the beginning of the function so there's no need to delay adding the related `MetricsReport` values. This ensures that the metrics are available in those early returns as well. The only caveat is that `on_user_turn_completed_delay` necessarily needs to wait until after it's computed which is why it's left where it is.

Separately on_end_of_turn also is missing initialization of the metrics_report attribute even though it should have access to the same information, so I've added the attribute in that path as well.

The consequence is we're losing information that's already readily available.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1877,6 +1877,27 @@ class AgentActivity(RecognitionHooks):
             transcript_confidence=info.transcript_confidence,
         )
 
+        metrics_report: llm.MetricsReport = {}
+        if self.stt:
+            metrics_report["stt_metadata"] = {
+                "model_name": self.stt.model,
+                "model_provider": self.stt.provider,
+            }
+        if info.started_speaking_at is not None:
+            metrics_report["started_speaking_at"] = info.started_speaking_at
+
+        if info.stopped_speaking_at is not None:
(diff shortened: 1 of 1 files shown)


uv.lock
```diff
@@ -2524,6 +2525,22 @@ dependencies = [
 [package.metadata]
 requires-dist = [{ name = "livekit-agents", editable = "livekit-agents" }]
 
+[[package]]
```
hudson-worden: Sorry this is unnecessary let me remove.

hudson-worden:
first commit moves the metricsreport initialization.
second commit extracts a common function and uses it where it was before and also in on_end_of_turn.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1953,29 +1960,6 @@ async def _user_turn_completed_task(
                 self._session._conversation_item_added(user_message)
             return
 
-        metrics_report: llm.MetricsReport = {}
```
hudson-worden, line 1956: Being down here, we lose the opportunity to attach metrics for early returns above

davidzhao (approved):
lg, thanks for the PR!

PR #5442, opened 2026-04-14
feat: add HandoffMetrics for measuring agent handoff timing

longcw:
## Summary

- Add `HandoffMetrics` to measure agent handoff cost, emitted via `metrics_collected` and attached to `AgentHandoff.metrics`
- Tracks orchestration timing: `drain_duration`, `new_activity_duration`, `on_enter_duration`, and total `duration`
- Reports whether STT pipeline and realtime session were reused across the handoff (`stt_reused`, `realtime_session_reused`)

## Limitation

`HandoffMetrics` does **not** include the WebSocket connection time for new realtime sessions or STT streams. When resources are not reused, the connection is established asynchronously in a background task. This connection time is already reported separately via `RealtimeModelMetrics.acquire_time` and `STTMetrics.acquire_time` through the existing `metrics_collected` pipeline.

### DIFF
--- a/livekit-agents/livekit/agents/metrics/base.py
+++ b/livekit-agents/livekit/agents/metrics/base.py
@@ -181,6 +181,26 @@ class InterruptionMetrics(_BaseMetrics):
     metadata: Metadata | None = None
 
 
+class HandoffMetrics(_BaseMetrics):
+    type: Literal["handoff_metrics"] = "handoff_metrics"
+    timestamp: float
+    """The timestamp when the handoff started."""
+    duration: float
+    """Total handoff duration in seconds (drain + close/pause + start/resume)."""
+    drain_duration: float
+    """Time in seconds to drain the old activity (on_exit + pending speech tasks)."""
+    new_activity_duration: float
+    """Time in seconds to start or resume the new activity."""
+    on_enter_duration: float
+    """Time in seconds for the new agent's on_enter callback to complete."""
+    stt_reused: bool = False
+    """Whether the STT pipeline was reused from the previous agent."""
+    realtime_session_reused: bool = False
+    """Whether the realtime session was reused from the previous agent."""
+    old_agent_id: str | None = None
+    new_agent_id: str | None = None
+
+
 AgentMetrics = (
     STTMetrics
     | LLMMetrics
@@ -189,4 +209,5 @@ AgentMetrics = (
     | EOUMetrics
     | RealtimeModelMetrics
     | InterruptionMetrics
+    | HandoffMetrics
 )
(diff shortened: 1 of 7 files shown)


livekit-agents/livekit/agents/metrics/base.py
```diff
@@ -181,6 +181,26 @@ class InterruptionMetrics(_BaseMetrics):
...
+    new_activity_duration: float
+    """Time in seconds to start or resume the new activity."""
+    on_enter_duration: float
+    """Time in seconds for the new agent's on_enter callback to complete."""
+    stt_reused: bool = False
+    """Whether the STT pipeline was reused from the previous agent."""
+    realtime_session_reused: bool = False
+    """Whether the realtime session was reused from the previous agent."""
+    old_agent_id: str | None = None
+    new_agent_id: str | None = None
```
chenghao-mou, line 201: nit: Do we need these two if we have them in AgentHandoff already?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1342,19 +1358,41 @@ async def _update_activity(
...
+            drain_duration=drain_duration,
+            new_activity_duration=new_activity_duration,
+            on_enter_duration=on_enter_duration,
+            stt_reused=stt_reused,
+            realtime_session_reused=rt_session_reused,
+            old_agent_id=handoff_item.old_agent_id,
+            new_agent_id=handoff_item.new_agent_id,
+        )
+        handoff_item.metrics = handoff_metrics
+        self.emit("metrics_collected", MetricsCollectedEvent(metrics=handoff_metrics))
```
chenghao-mou, line 1395: Having it at the session-level makes sense, but we did just deprecate the session-level `metrics_collected` event. Should we put it somewhere else?

longcw: yeah that's an issue, I think one option is to put it in `AgentHandoff` and then emit via `self.emit("conversation_item_added", ConversationItemAddedEvent(item=handoff_item))`, but the metrics come after the handoff emitted.

chenghao-mou: Should we delay emitting "conversation_item_added" until the handoff metrics are available?

longcw: this may cause the handoff event emit after the reply from `on_enter`?

chenghao-mou: I see. Semantically, handoff happens before any reply in `on_enter`. Then maybe we should exclude new activity start/resume and on_enter from handoff metrics. Instead, we can have more generic Agent/AgentTask-level metrics measuring those.

longcw: yeah I agree, we shouldn't include on_enter in the handoff metrics.

longcw:
couldn't find a way to expose the STT connection time from the stt_node

Issue #5443, opened 2026-04-14
Support otel metrics for agent server


Bnowako:
### Feature Type

Would make my life easier

### Feature Description

Currently agent servers only export prometheus metrics, would be nice to support also otel

Bnowako:
What do you think @longcw?

PR #5445, opened 2026-04-14
agent server otel metrics

Bnowako:
Complementary otel setup for Agent Server metrics. 

It's enabled when either  `  "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT" | "OTEL_EXPORTER_OTLP_ENDPOINT"` env var is set

The only thing I did not understood quite is multiproc prometheus setup, as agent server has only 1 process and the load_task is running there. Would appreciate help there!

Lmk if we should default to livekit o11y url

related issue: https://github.com/livekit/agents/issues/5443

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -755,7 +755,10 @@ class AgentServer(utils.EventEmitter[EventTypes]):
                     self._worker_load = await self._invoke_load_fnc()
 
                     telemetry.metrics._update_worker_load(self._worker_load)
-                    if self._prometheus_multiproc_dir or telemetry.metrics.worker_observability_metrics_enabled():
+                    if (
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/worker.py
```diff
@@ -753,7 +755,10 @@ async def _load_task() -> None:
                     self._worker_load = await self._invoke_load_fnc()
 
                     telemetry.metrics._update_worker_load(self._worker_load)
-                    if self._prometheus_multiproc_dir:
+                    if (
```
Bnowako, line 758: This check could be confusing, I don't like it but not sure how to improve it for now

Im wondering If we even need multiproc setup, as we are collecting AgentServer metrics which is just 1 process, if yes potentially otel setup needs change now.

PR #5346, opened 2026-04-05
feat: add service_tier parameter to Responses API LLM

piyush-gambhir:
## Summary

The Chat Completions LLM (`openai.LLM`) already supports the `service_tier` parameter for configuring priority/flex/default processing per-request. The Responses API LLM (`openai.responses.LLM`) is missing this parameter despite the OpenAI Responses API [supporting it](https://platform.openai.com/docs/api-reference/responses/create).

This PR adds `service_tier` to the Responses LLM for parity.

## Changes

**`livekit-plugins/livekit-plugins-openai/.../responses/llm.py`** (1 file):
- Add `service_tier: NotGivenOr[str]` to `_LLMOptions`
- Add `service_tier` parameter to `LLM.__init__()`
- Pass `service_tier` through in `chat()` via extra kwargs

## Usage

\`\`\`python
from livekit.plugins.openai import responses

llm = responses.LLM(
    model="gpt-5.4",
    service_tier="priority",  # now supported
)
\`\`\`

## Backward Compatible

- Defaults to \`NOT_GIVEN\` — no impact on existing code
- Matches the existing pattern used by the Chat Completions LLM (\`NotGivenOr[str]\`)

PR #5449, opened 2026-04-14
feat(openai): expose max_output_tokens on Responses API LLM

piyush-gambhir:
## Summary

Adds `max_output_tokens` as a first-class constructor argument on the OpenAI, Azure, and xAI Responses API LLM wrappers — mirroring how `service_tier` is handled today (#5346).

The underlying `/v1/responses` endpoint accepts `max_output_tokens` natively (see `openai.types.responses.response_create_params`), but the plugins previously required callers to reach for `extra_kwargs` on every `chat()` call to set it. This PR makes it a plain constructor arg so a bound can be set once at LLM construction time.

### Changes
- `livekit-plugins-openai/.../responses/llm.py`: add `max_output_tokens: NotGivenOr[int]` to `_LLMOptions` and the `LLM.__init__` signature; forward it into `extra` so it reaches both the HTTP (`client.responses.create`) and WebSocket (`response.create`) paths.
- `livekit-plugins-azure/.../responses/llm.py`: thread the new arg through to `openai.responses.LLM.__init__`.
- `livekit-plugins-xai/.../responses/llm.py`: thread the new arg through to `openai.responses.LLM.__init__`.

### Usage

```python

llm = openai.responses.LLM(
    model="gpt-4.1",
    max_output_tokens=1024,
)
```

## Motivation

All sibling chat-completions LLMs (`openai.LLM`, `anthropic.LLM`, `google.LLM`, `aws.LLM`, `mistralai.LLM`, `groq.LLM`) already expose a first-class `max_completion_tokens` / `max_tokens` / `max_output_tokens` argument. The Responses variants were the only remaining path that forced callers into `extra_kwargs`. This brings them in line.

## Test plan

- [x] `ast.parse` passes on all three modified files
- [x] `make check` (format + lint + type-check)
- [x] Smoke test: instantiate `openai.responses.LLM(model="gpt-4.1", max_output_tokens=64)` and verify the response honors the cap in both `use_websocket=True` and `use_websocket=False` modes

### DIFF
--- a/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/responses/llm.py
+++ b/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/responses/llm.py
@@ -32,6 +32,7 @@ class LLM(openai.responses.LLM):
         tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
         timeout: httpx.Timeout | None = None,
         reasoning: NotGivenOr[Reasoning] = NOT_GIVEN,
+        max_output_tokens: NotGivenOr[int] = NOT_GIVEN,
     ) -> None:
         """
         This automatically infers the following arguments from their corresponding environment variables if they are not provided:
@@ -68,6 +69,7 @@ class LLM(openai.responses.LLM):
             parallel_tool_calls=parallel_tool_calls,
             tool_choice=tool_choice,
             reasoning=reasoning,
+            max_output_tokens=max_output_tokens,
         )
         self._azure_client = azure_client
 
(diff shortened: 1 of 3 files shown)


piyush-gambhir:
Smoke test (live OpenAI API, `gpt-4.1`):

- **HTTP path** ✅: `LLM(model='gpt-4.1', use_websocket=False, max_output_tokens=16)` — 500-word-story prompt returned truncated output (~12–16 tokens, e.g. *"Once upon a time, in the misty blue mountains of Elderglow,"*). Confirms `max_output_tokens=16` is forwarded to `client.responses.create(...)` and honored by the API.
- **WebSocket path** ✅: `stream._extra_kwargs['max_output_tokens'] = 16` before send — the param is spread into the `response.create` payload dict via the same `**self._extra_kwargs` path that carries every other option (`temperature`, `service_tier`, etc.), so wiring is symmetric with HTTP.

`make check` passes locally (only pre-existing warnings, none in changed files).

tinalenguyen (approved):
lgtm, thank you!

PR #5450, opened 2026-04-14
fix(llm) - Fix Raw Schema for Google Gemini

fgoessl:
# Bug Description
Function tools work correctly with Google Gemini if they are declared using the `@function_tool` decorator. 

If they are supplied as a Raw Schema however, only the tool name and the tool description are visible to the model and the parameters are empty. As a result the tool cannot be properly used by the model.

# Steps to Reproduce 
1. Define a tool using a raw schema as described here https://docs.livekit.io/agents/logic/tools/definition/#creating-tools-from-raw-schema. 
2. Use: `livekit-agents==1.5.2` and `gemini-live-2.5-flash-native-audio`. 
3. Observe the tool calls and/or directly ask the model on what tools it has access to and what parameters they have.

# Expected Behavior
The parameters are available to the model. The behaviour of decorated tools and tools generated from raw schemata should match.

# Actual Behaviour
Only the tool name and description are available to the model.

I would be happy to add some tests if needed. Just let me know how they are handled here.


longcw:
thanks for creating the pr! and yeah it's the Gemini Live doesn't support `parameters_json_schema` (see https://github.com/googleapis/python-genai/issues/1147). I create a separate pr to fix this for the live api https://github.com/livekit/agents/pull/5560

PR #5453, opened 2026-04-15
fix(liveavatar): wait for connected state and chunk audio before sending

dyi1:
This PR makes two changes:

1. Gate audio forwarding and keep-alive on the WebSocket session.state_updated connected event to prevent silently dropped messages. 
2. Batch audio frames into 600ms initial / 1s subsequent chunks to reduce WebSocket message overhead.


tinalenguyen (approved):
lgtm, thank you!

PR #5455, opened 2026-04-15
fix(room-io): preserve audio frame processors across stream transitions

### LINKED ISSUE #5448: AudioInputOptions.noise_cancellation: FrameProcessor killed by premature _close() in _close_stream() — processes zero frames
When using a custom `rtc.FrameProcessor` (e.g., DTLN noise suppression from[ livekit-plugins-dtln](https://github.com/aloware/livekit-plugins-dtln)) via `AudioInputOptions.noise_cancellation,` the processor is **permanently killed before it processes any audio frames.**          

`_close_stream() `in `_ParticipantInputStream` unconditionally calls `_processor._close()` on every track subscription — including the **first** one, when there is no previous stream to clean 
  up. This destroys the processor before `rtc.AudioStream` gets a chance to use it.

This affects any `rtc.FrameProcessor` passed via noise_cancellation, not just DTLN. Native `rtc.NoiseCancellationOptions `(Krisp/BVC) is unaffected because it's handled at the C/Rust  layer inside `AudioStream.from_track()`. 

### Expected Behavior

**Expected behavior**                                                                                                                                                                     
   
  `noise_cancellation` FrameProcessor should process every incoming audio frame for the duration of the call. `_close()` should only be called during final cleanup (`aclose()`), not during  
  stream transitions.                                       

  **Actual behavior**                                                                                                                                                                       
   
  dtln.summary  {"frames_processed": 0, "strength": 0.8, "enabled": false}                                                                                                              
                                                            
  The processor processes `zero frames` and is permanently disabled.

### Reproduction Steps

```bash
from livekit.plugins import dtln                                                                                                                                                      from livekit.agents import room_io                        
                                                                                                                                                                                        
await session.start(                                      
      agent=my_agent,
      room=ctx.room,                                                                                                                                                                    
              noise_cancellation=dtln.noise_suppression(strength=0.8),
          ),                              
      ),
  )                                                                                                                                                                                     
   
  # After call ends, check:                                                                                                                                                             
  # dtln.summary will show frames_processed=0, enabled=false
```

### Operating System

Linux

### Models Used

OpenAI Realtime Model

### Package Versions

```bash
livekit-agents: 1.5.2                                   
livekit-rtc: 0.22.x                                                                                                                                                                 
livekit-plugins-dtln: 0.2.2
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
Move `_processor._close()` out of `_close_stream()` (which runs on every stream transition) and into `aclose()` (which runs only during final shutdown)
```

### Additional Context

_No response_

_No response_

nightcityblade:
Fixes #5448

## Summary
- stop `_ParticipantInputStream._close_stream()` from closing the active frame processor during ordinary stream transitions
- close replaced per-track processors inside `_ParticipantAudioInputStream._create_stream()` and close orphaned processors when a track disappears without replacement
- add a regression test covering a direct `rtc.FrameProcessor` passed via `noise_cancellation` so it survives the initial subscription and still closes during final teardown

## Testing
- `.venv/bin/python -m pytest tests/test_room_io_noise_cancellation.py tests/test_room_io_teardown.py`
- `.venv/bin/ruff check livekit-agents/livekit/agents/voice/room_io/_input.py tests/test_room_io_noise_cancellation.py tests/test_room_io_teardown.py`

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/_input.py
+++ b/livekit-agents/livekit/agents/voice/room_io/_input.py
@@ -306,15 +306,28 @@ class _ParticipantAudioInputStream(_ParticipantInputStream[rtc.AudioFrame], Audi
             frame_size_ms=self._frame_size_ms,
         )
 
+    def _close_selector_processor(self) -> None:
+        if isinstance(self._noise_cancellation, rtc.FrameProcessor):
+            return
+        if self._processor is not None:
+            self._processor._close()
+            self._processor = None
+
+    @override
+    def set_participant(self, participant: rtc.RemoteParticipant | str | None) -> None:
+        previous_processor = self._processor
+        super().set_participant(participant)
+        if self._stream is None and self._processor is previous_processor:
+            self._close_selector_processor()
+
     @override
     def _on_track_unavailable(
         self, publication: rtc.RemoteTrackPublication, participant: rtc.RemoteParticipant
     ) -> None:
         current_processor = self._processor
         super()._on_track_unavailable(publication, participant)
-        if self._stream is None and self._processor is current_processor and current_processor is not None:
-            current_processor._close()
-            self._processor = None
+        if self._stream is None and self._processor is current_processor:
+            self._close_selector_processor()
 
     @override
     async def _forward_task(
--- a/livekit-agents/livekit/agents/voice/room_io/_input.py
+++ b/livekit-agents/livekit/agents/voice/room_io/_input.py
@@ -175,9 +175,6 @@ class _ParticipantInputStream(Generic[T], ABC):
             self._tasks.add(task)
             self._stream = None
             self._publication = None
-        if self._processor:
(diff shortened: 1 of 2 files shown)


nightcityblade:
Thanks, good catch. I restored the selector-created processor cleanup for participant changes that do not immediately create a replacement stream, including  and switching to a participant with no available track yet. I also added regression coverage for both paths and reran the targeted room I/O noise cancellation tests.

longcw:
thanks for the pr! If you don't mind I created a separate one with some cleanup and a different impl in https://github.com/livekit/agents/pull/5467

nightcityblade:
Thanks for the heads-up, and thanks for putting together the follow-up PR with the cleanup and alternate implementation. That sounds good to me. Since you've got #5467 moving forward, I'm happy to defer to that one.

PR #5458, opened 2026-04-15
xai stt

tinalenguyen:


livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/types.py
```diff
@@ -1,5 +1,35 @@
 from typing import Literal
 
+STTAudioEncoding = Literal["pcm", "mulaw", "alaw"]
```
theomonnom: We only have to support pcm, you can remove this type

theomonnom (approved):

livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/types.py
tinalenguyen: oh as in we do not support the other ones or that we should only use pcm

PR #5454, opened 2026-04-15
Rename e2ee to encryption in JobContext.connect

longcw:
## Summary
- Rename `e2ee` parameter to `encryption` in `JobContext.connect()` with backward-compatible deprecation via `@deprecate_params`

PR #5469, opened 2026-04-16
fix(examples): update e2ee.py to use encryption kwarg and env var

aryeila:
## Summary
- Rename `e2ee=` to `encryption=` on `ctx.connect` to match the renamed kwarg from #5454
- Load shared key from `LIVEKIT_E2EE_KEY` env var instead of hardcoding it

Resolves DOCS-1278

### DIFF
--- a/examples/primitives/e2ee.py
+++ b/examples/primitives/e2ee.py
@@ -1,4 +1,5 @@
 import logging
+import os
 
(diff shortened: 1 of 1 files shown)


theomonnom (approved):

davidzhao (approved):
thanks @aryeila !

PR #5476, opened 2026-04-17
feat(assemblyai): log connection lifecycle, silence, and session correlators

dlange-aai:
## Summary

Extends the DEBUG logging added in #5146.

- **WebSocket lifecycle** — `debug` before `ws_connect` (model + base_url; URL intentionally excluded since the query string contains customer `prompt` / `keyterms_prompt`) and after with the HTTP status.
- **Unexpected close** — `warning` with `session_id`, `close_code`, `msg.data`, `msg.extra` logged before the `APIStatusError` is raised, so the close code is visible even if the exception gets swallowed downstream.
- **Recv silence** — `warning` when no message arrives from AAI for 15s, and re-emitted every 15s while silence continues (with the actual elapsed duration). Makes the "interims keep arriving but no finals" vs "AAI went quiet" distinction explicit.
- **Session correlators** — `session=...` added to every existing debug/error/exception log in `_process_stream_event` and `recv_task` so logs can be cross-referenced with AAI-side logs.
- **Unhandled message types** — `debug` log for unknown `type` values, which were previously silently dropped.

No impact at default log levels — all new output is `debug` except the two `warning`-level anomaly signals.

## Test plan

- [x] `make format lint type-check` pass
- [x] Verify in a running agent session that:
  - `connecting to AssemblyAI model=... base_url=...` appears once at connect
  - `AssemblyAI WebSocket connected status=...` appears once
  - Existing interim/preflight/final logs now include `session=<id>`
  - Closing the WS server-side produces the unexpected-close warning

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -19,6 +19,7 @@ import asyncio
 import dataclasses
 import json
 import os
+import time
 import weakref
 from dataclasses import dataclass
 from typing import Literal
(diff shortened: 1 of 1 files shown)


tinalenguyen (approved):
lgtm, thank you!!

PR #5481, opened 2026-04-18
fix: strip markdown emphasis adjacent to punctuation

carschandler:
The `filter_markdown` inline regexes for `**bold**` and `*italic*` used `(?<!\S)` / `(?!\S)` boundaries, rejecting punctuation as a valid boundary. As a result LLM output like `**Frankie**!`, `**date**,`, and `(**1**)` passed through untouched and reached TTS as raw markdown.

Replace the asterisk patterns with `\w`/`*`-based boundaries (already used correctly by the underscore patterns) and add guards against whitespace-adjacent inner text so `a ** b` style arithmetic is still preserved.

Add parametrized regression tests across common punctuation characters and chunk sizes, plus false-positive guards for `x**2`, `*.py`, and arithmetic expressions.

Addresses #5482

### DIFF
--- a/.github/next-release/changeset-filter-markdown-punctuation.md
+++ b/.github/next-release/changeset-filter-markdown-punctuation.md
@@ -1,0 +1,5 @@
+---
+"livekit-agents": patch
+---
+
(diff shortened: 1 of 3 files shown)


longcw (approved):
thanks for the fix!

PR #5485, opened 2026-04-19
(aws realtime): add expiry check for cached credentials

### LINKED ISSUE #5484: [Detail Bug] Bedrock realtime voice sessions reuse expired AWS credentials, causing auth failures
# Summary
- **Context**: The `Boto3CredentialsResolver` class provides AWS credential caching for the Nova Sonic realtime voice model integration with AWS Bedrock.
- **Bug**: The `get_identity()` method returns cached credentials unconditionally without checking if they have expired, even though `_cached_expiry` is calculated and stored.
- **Actual vs. expected**: Cached credentials are returned even after they expire; credentials should be refreshed when expired.
- **Impact**: AWS API requests may fail with authentication errors when credentials expire mid-session or between sessions.

# Code with bug
```python
async def get_identity(self, **kwargs: Any) -> AWSCredentialsIdentity:
    """Asynchronously resolve AWS credentials.
    ...
    """
    # Return cached credentials if available
    # Session recycling will close the connection and get fresh credentials before these expire
    if self._cached_identity:
        return self._cached_identity  # <-- BUG: No expiry check!

    try:
        logger.debug("[CREDS] Attempting to load AWS credentials")
        credentials = self.session.get_credentials()
        ...
        # Cache the identity and expiry
        self._cached_identity = identity
        if expiry_time:
            # Session will restart 3 minutes before expiration
            self._cached_expiry = expiry_time.timestamp() - 180  # <-- Set but never used
```

# Evidence

## Evidence 1: `_cached_expiry` is computed but never used for validation

The field `_cached_expiry` is set at lines 253 and 260 but is **never read** to validate cached credentials. The only usage is logging:

| Line | Code | Purpose |
|------|------|---------|
| 200 | `self._cached_expiry: float \| None = None` | Declaration |
| 253 | `self._cached_expiry = expiry_time.timestamp() - 180` | Assignment |
| 256 | `f"expiry_time={expiry_time}, restart_before={self._cached_expiry}"` | Logging only |
| 260 | `self._cached_expiry = None` | Assignment |

**The `_cached_expiry` value is never compared against `time.time()` to determine cache validity.** This is dead code for validation purposes.

### Claim: "Session recycling creates a fresh client with fresh credentials"

**This is FALSE.** Here is the complete code trace:

### `_graceful_session_recycle()` (lines 681-748):
1. **Line 734**: `self._bedrock_client = None` — clears only the client reference
2. **Line 746**: Calls `await self.initialize_streams(is_restart=True)`

### `initialize_streams()` (lines 850-949):
- **Lines 861-863**: Checks if `_bedrock_client` is None and calls `_initialize_client()`:
```python
if not self._bedrock_client:
    logger.info("Creating Bedrock client")
    self._initialize_client()
```

### `_initialize_client()` (lines 564-574):
- **Line 569**: Gets the **same global singleton** credentials resolver:
```python
aws_credentials_identity_resolver=_get_credentials_resolver(),
```

### `_get_credentials_resolver()` (lines 102-110):
```python
global _shared_credentials_resolver
if _shared_credentials_resolver is None:
    _shared_credentials_resolver = Boto3CredentialsResolver()
return _shared_credentials_resolver  # <-- Returns SAME singleton every time
```

### `Boto3CredentialsResolver.get_identity()` (lines 202-263):
```python
if self._cached_identity:
    return self._cached_identity  # <-- Returns stale cached value
```

**Conclusion**: When `_graceful_session_recycle()` creates a new `BedrockRuntimeClient`, it uses the **exact same cached credentials** from the singleton `Boto3CredentialsResolver`. No cache invalidation occurs.

### Claim: "AWS SDK's SigV4AuthScheme checks the expiration field"

**This is INCORRECT.** The `IdentityResolver` interface (from `smithy_core.aio.interfaces.identity`) is called by the auth scheme to obtain credentials. The auth scheme's responsibility is to:
1. Call `get_identity()` to obtain credentials
2. Sign the request using those credentials

The AWS SDK does **not** independently check if the `expiration` field on `AWSCredentialsIdentity` has passed. It trusts the `IdentityResolver` to return valid credentials. The `expiration` field is informational only — it's used for logging or caching decisions by the resolver itself.

This is the standard pattern for identity resolvers across AWS SDKs: the resolver is responsible for credential freshness, not the consumer.

## Evidence 4: Timer delays can push recycle past expiry

The `_session_recycle_timer()` (lines 620-674) has multiple wait phases that can delay recycling:

```python
# Phase 1: Wait for assistant to finish speaking (NO TIMEOUT)
while not self._audio_end_turn_received:
    await asyncio.sleep(0.1)  # Unbounded wait

# Phase 2: Wait for audio to stop (NO UPPER BOUND)
while True:
    await asyncio.sleep(0.1)
    if self._last_audio_output_time == last_audio_time:
        await asyncio.sleep(0.9)
        if self._last_audio_output_time == last_audio_time:
            break  # Only exits if 1s silence
    else:
        last_audio_time = self._last_audio_output_time  # Reset timer
```

If the assistant is in the middle of a long response, the timer fires but recycling is delayed indefinitely. During this delay, credentials can expire.

## Evidence 5: Duration calculation allows exceeding expiry

Lines 590-600:
```python
time_until_cred_expiry = credential_expiry - now - CREDENTIAL_EXPIRY_BUFFER_SECONDS
duration = min(MAX_SESSION_DURATION_SECONDS, time_until_cred_expiry)

if duration < 30:
    duration = max(duration, 10)  # At least 10 seconds
```

If credentials expire in 5 seconds, `duration` becomes `min(360, 5-180) = -175`, then `max(-175, 10) = 10`. The session will try to run for 10 seconds with credentials that expire in 5 seconds.

# Why has this bug gone undetected?

1. **Short sessions + long-lived credentials**: The 6-minute default session often completes before temporary credentials (typically 15min-1hr) expire.

2. **Static credentials in testing**: Developers using static AWS keys (`AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`) never hit this, as static credentials have no expiry.

3. **No unit tests**: The AWS plugin has no test directory for this credential logic.

4. **Race condition**: The bug requires credentials to expire during an active session, which is timing-dependent.

5. **Silent failures**: Auth failures from expired credentials may be misattributed to network issues or other transient errors.

# Recommended fix

Add an expiry check in `get_identity()`:

```python
async def get_identity(self, **kwargs: Any) -> AWSCredentialsIdentity:
    now = time.time()
    # Check if cached credentials are still valid
    if self._cached_identity and (self._cached_expiry is None or now < self._cached_expiry):
        return self._cached_identity
    
    # Credentials expired or not cached - fetch fresh credentials
    self._cached_identity = None
    self._cached_expiry = None
    ...
```

Alternatively, add a cache invalidation method and call it before session recycling:

```python
def invalidate_cache(self) -> None:
    self._cached_identity = None
    self._cached_expiry = None
```

Then in `_graceful_session_recycle()`:
```python
_get_credentials_resolver().invalidate_cache()
```

FINAL_SUBMISSION

# History

This bug was introduced in commit 63cfb3a (@kachenjr, 2025-12-15, PR #4250). The commit added proactive session recycling to handle AWS credential expiry, but in doing so, it replaced the proper expiry check in `get_identity()` with an unconditional early return. The original code at lines 185-188 checked `if time.time() < self._cached_expiry` before returning cached credentials. The "fix" removed this expiry validation, assuming session recycling would always refresh credentials before expiry. However, the singleton credentials resolver has no cache invalidation method, and the session recycling logic does not clear the cached credentials—it only creates a new BedrockRuntimeClient. This leaves stale credentials persisting indefinitely in the singleton cache when recycling fails, is delayed, or credentials expire between recycling attempts.

tinalenguyen:
#5484

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
@@ -213,10 +213,16 @@ class Boto3CredentialsResolver(IdentityResolver):  # type: ignore[misc]
         Raises:
             ValueError: If no credentials could be found by boto3.
         """
-        # Return cached credentials if available
-        # Session recycling will close the connection and get fresh credentials before these expire
-        if self._cached_identity:
+        # Return cached credentials if available and not expired
+        current_time = time.time()
+        if self._cached_identity and (
+            self._cached_expiry is None or current_time < self._cached_expiry
+        ):
             return self._cached_identity
+
+        # Credentials expired or not cached - reset so fresh ones are fetched below
+        self._cached_identity = None
+        self._cached_expiry = None
 
         try:
             logger.debug("[CREDS] Attempting to load AWS credentials")


davidzhao (approved):
lg

PR #5492, opened 2026-04-20
feat(voice): add UserTurnLimitOptions to interrupt long user speech

longcw:
## Summary

Adds a configurable limit on how long a user can speak before the agent is given a chance to interrupt. Accumulates word count and wall-clock duration across consecutive user turns.

When a threshold is crossed, the framework fires `UserTurnExceededEvent` and invokes `Agent.on_user_turn_exceeded`. The default implementation calls `session.generate_reply(..., allow_interruptions=False, tool_choice="none")` with instructions to politely cut in with a short reply. Users can override the hook for custom behavior.

### API

```python
AgentSession(
    turn_handling={
        "user_turn_limit": {"max_words": 100, "max_duration": 30.0},
    },
)
```

Both thresholds default to `None` (disabled). Set either or both.

### Coordination

`AgentActivity._user_turn_exceeded_task` waits for the normal EOU-triggered response to resolve before firing the hook — if the agent reaches `speaking` on its own, the callback is skipped.


livekit-agents/livekit/agents/voice/events.py
```diff
@@ -219,6 +219,20 @@ class SpeechCreatedEvent(BaseModel):
...
+    transcript: str
+    """Transcript from the current (uncommitted) user turn only.
+    Previous turns in the accumulation window are already in the chat context."""
+    accumulated_transcript: str
+    """Full transcript since the start of user speaking."""
+    accumulated_word_count: int
+    """Total word count since the start of user speaking."""
+    duration: float
+    """Duration of the user turn in seconds."""
+    created_at: float = Field(default_factory=time.time)
```
theomonnom, line 244: If we're committing the turn to the chat_context, should we just give a reference to llm.ChatMessage?

livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -254,6 +255,26 @@ async def on_user_turn_completed(
         """
         pass
 
+    async def on_user_turn_exceeded(self, ev: UserTurnExceededEvent) -> None:
```
theomonnom, line 263: Do you think it should be a new overridable, vs a new event inside the AgentSession

livekit-agents/livekit/agents/voice/events.py
longcw: yeah that sounds good

livekit-agents/livekit/agents/voice/agent.py
longcw: it needs to be a overridable if we want to have a default behavior like now.

theomonnom (approved):

PR #5495, opened 2026-04-20
Custom attributes support in Avatar

Darshak03:


theomonnom:
I’m wondering if this is necessary, since you can directly use the agent participant’s attributes instead of the avatar’s?

livekit-plugins/livekit-plugins-anam/livekit/plugins/anam/avatar.py
```diff
@@ -93,8 +102,7 @@ async def start(
             .with_identity(self._avatar_participant_identity)
             .with_name(self._avatar_participant_name)
             .with_grants(api.VideoGrants(room_join=True, room=room.name))
-            # allow the avatar agent to publish audio and video on behalf of your local agent
-            .with_attributes({ATTRIBUTE_PUBLISH_ON_BEHALF: local_participant_identity})
```
longcw, line 97: I don't think we should allow customizing this attribute, internally we are using this attribute the detect if a participant is representing another participant.

PR #5514, opened 2026-04-21
voice: add opt-in audio turn detector hook

piyush-gambhir:
## Summary

This PR adds an opt-in extension point for audio-native turn detectors in `livekit-agents`.

Today, custom turn detectors operate on chat or transcript context only. This change adds:
- `AudioTurnContext`, a public container for buffered current-turn audio plus transcript and chat context
- `AudioTurnDetector`, a public protocol for audio-native end-of-turn detectors
- current-turn PCM buffering in `AudioRecognition`
- audio-detector invocation in the existing end-of-utterance path before endpointing and commit
- a focused example agent and unit coverage for the new hook

## Motivation

Some turn detectors operate directly on audio rather than text. Those detectors cannot currently plug into the existing custom turn-detection API because only `ChatContext` is exposed to custom detectors.

This PR adds the minimal generic hook needed to support that class of detector without introducing any provider-specific dependency.

## Scope

- no default behavior change
- existing text/chat-context turn detectors continue to work unchanged
- no built-in vendor or model integration is added in this PR
- audio is exposed as buffered current-turn frames; model-specific preprocessing remains the responsibility of the detector implementation

## Validation

- `make check`
- `uv run pytest tests/test_agent_session.py tests/test_audio_recognition_handoff.py tests/test_audio_turn_detection.py -q`
- `uv run python -m py_compile examples/voice_agents/audio_turn_detector.py`
- `uv run examples/voice_agents/audio_turn_detector.py --help`

## Notes

This draft intentionally keeps the change generic and provider-agnostic. A follow-up plugin PR can implement a concrete detector on top of this hook.

This is also intended as a narrower, lower-risk framing of audio-based turn detection than the earlier feature request in #3094.

### DIFF
--- a/examples/voice_agents/README.md
+++ b/examples/voice_agents/README.md
@@ -23,6 +23,7 @@ session = AgentSession(
 ### 🚀 Getting Started
 
 - [`basic_agent.py`](./basic_agent.py) - A fundamental voice agent using LiveKit Inference with metrics collection
+- [`audio_turn_detector.py`](./audio_turn_detector.py) - Using a custom audio-native turn detector with buffered current-turn PCM
 
 ### 🛠️ Tool Integration & Function Calling
 
--- a/examples/voice_agents/audio_turn_detector.py
+++ b/examples/voice_agents/audio_turn_detector.py
@@ -1,0 +1,113 @@
+import logging
+
+from dotenv import load_dotenv
+
+from livekit.agents import (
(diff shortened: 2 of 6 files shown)


davidzhao:
thanks for the suggestion. a native solution is coming, including new interfaces to support it. closing this in favor of #4722

PR #5522, opened 2026-04-22
fix: clear _hist buffer in MovingAverage.reset() to prevent stale averages

### LINKED ISSUE #5521: Bug: MovingAverage.reset() does not clear internal history buffer, causing incorrect averages after reset
## Bug Description

`MovingAverage.reset()` in `livekit-agents/livekit/agents/utils/moving_average.py` resets `_count` and `_sum` to zero but **does not clear the `_hist` ring-buffer**. This causes stale values from before the reset to corrupt the moving average once the window fills up again.

## Root Cause

```python
def reset(self) -> None:
    self._count = 0
    self._sum = 0
    # BUG: self._hist is not cleared!
```

After `reset()`, as `add_sample()` is called:
- `_count` increments from 0, `_sum` is rebuilt correctly *until* the window is full.
- Once `_count > len(self._hist)`, the code subtracts `self._hist[index]` from `_sum` to evict the oldest entry. But `_hist[index]` still contains the pre-reset value, so the sum is incorrectly reduced by that stale value.

## Fix

Clear `_hist` in `reset()`:

```python
def reset(self) -> None:
    self._hist = [0.0] * len(self._hist)
    self._count = 0
    self._sum = 0.0
```

## Impact

`MovingAverage` is used in metrics tracking across the agents framework. Incorrect averages after a reset can lead to skewed performance metrics and incorrect adaptive behaviour.

kuishou68:
Closes #5521

## Problem
`MovingAverage.reset()` resets `_count` and `_sum` to zero but does not clear the `_hist` ring-buffer. After a reset, once the window fills again, the eviction path subtracts stale pre-reset values from `_sum`, producing incorrect moving averages.

## Fix
Clear `_hist` in `reset()` by replacing with a zero-initialized list of the same length.

## Changes
- `livekit-agents/livekit/agents/utils/moving_average.py`: added `self._hist = [0.0] * len(self._hist)` as first line of `reset()`

### DIFF
--- a/livekit-agents/livekit/agents/utils/moving_average.py
+++ b/livekit-agents/livekit/agents/utils/moving_average.py
@@ -21,6 +21,7 @@ class MovingAverage:
         return self._sum / self.size()
 
     def reset(self) -> None:
+        self._hist = [0.0] * len(self._hist)
         self._count = 0
         self._sum = 0
 


livekit-agents/livekit/agents/utils/moving_average.py
```diff
@@ -21,6 +21,7 @@ def get_avg(self) -> float:
         return self._sum / self.size()
 
     def reset(self) -> None:
+        self._hist = [0.0] * len(self._hist)
```
longcw, line 24: the hist is used only when `count > len(hist)`, that means all values in hist were overwritten already. but I think it's good to also reset it here.

longcw (approved):

longcw:
could you sign the CLA so we can merge it?

kuishou68:
> could you sign the CLA so we can merge it?

OK

PR #5511, opened 2026-04-21
feat(avatar): add playback_started RPC for remote avatar workers

longcw:
## Summary

- Adds `lk.playback_started` RPC so remote avatar workers (e.g. `AvatarRunner`) can notify the agent session when playback actually begins, instead of approximating it the instant a frame is queued.
- Introduces a `wait_playback_start` opt-in on `DataStreamAudioOutput` / `QueueAudioOutput`. When false (default), behavior matches before: `on_playback_started` fires eagerly on first captured frame. When true, the event is deferred until the remote sends the RPC.
- If the avatar service is built on `AvatarRunner`, just upgrade `livekit-agents` and set `wait_playback_start=True` in the plugin will enable this feature.

PR #5524, opened 2026-04-22
feat(metrics): add playback_latency metric

longcw:
## Summary

Adds a new `playback_latency` metric on assistant `ChatMessage`s, measuring the time between the first audio frame being forwarded and playback actually starting.

This metric is typically near-zero in normal pipelines, and becomes significant when a remote avatar worker is in the chain. Requires playback_started RPC call from the avatar to measure this latency #5511


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -291,6 +291,13 @@ class MetricsReport(TypedDict, total=False):
     Assistant `ChatMessage` only
     """
 
+    playback_start_latency: float
+    """Latency between the first audio frame being forwarded and playback starting on the client.
+    Typically very small; can be significant when a remote avatar worker is in the chain.
+
+    Assistant `ChatMessage` only
+    """
```
theomonnom, line 302: Makes sense! Is it included inside the e2e_latency?

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -291,6 +291,13 @@ class MetricsReport(TypedDict, total=False):
     Assistant `ChatMessage` only
     """
 
+    playback_start_latency: float
```
theomonnom: Should we find a better name? Maybe just `playback_latency`?

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -291,6 +291,13 @@ class MetricsReport(TypedDict, total=False):
...
+    """Latency between the first audio frame being forwarded and playback starting on the client.
```
theomonnom: `Latency between the first audio frame being forwarded and playback starting on the client`

It isn't really accurate tho? We don't know when the client is playing the audio (+ network latency?) 
It's true for avatars. But we should maybe clarify

livekit-agents/livekit/agents/llm/chat_context.py
longcw: yes, it's included inside the e2e_latency

livekit-agents/livekit/agents/llm/chat_context.py
longcw: yeah, it didn't include the network latency, updated the docstring.

theomonnom (approved):

toubatbrian:

Classification: **core runtime improvement** (new `playback_latency` metric on assistant `ChatMessage`s in `MetricsReport`, plus tracking of `started_forwarding_at` in the audio forwarding task). No Python-specific dependencies — eligible for port.

Verified that no existing PR or branch in `livekit/agents-js` already ports this. Proceeding to port.

---

PR #5529, opened 2026-04-22
fix(anthropic): raise default httpx read timeout for streaming; add configurable timeout param

### LINKED ISSUE #5508: Anthropic plugin: default 5s httpx timeout too aggressive for adaptive-thinking / large-context models
## Summary

The Anthropic LLM plugin creates its httpx client with a hardcoded 5-second timeout:

```python
# livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py, around line 121
self._client = client or anthropic.AsyncClient(
    api_key=anthropic_api_key,
    base_url=base_url if is_given(base_url) else None,
    http_client=httpx.AsyncClient(
        timeout=5.0,    # ← per-chunk read timeout
        follow_redirects=True,
        limits=httpx.Limits(...),
    ),
)
```

This is the per-chunk read timeout on the streaming SSE response, and it's routinely exceeded in normal usage.

## When it fires

- **Claude Sonnet 4.6 / Opus 4.6 with adaptive thinking** — thinking phases of 10-30s before the first chunk are normal and expected behavior from the model
- **Large chat context** (≥20% of 200k tokens) — first-chunk latency scales with input size
- **Anthropic API latency spikes** — status-page-reported incidents (e.g. Anthropic's `Sonnet 4.5 error spike` on 2026-04-20) make the API slower across all users for a window
- **Network variability** — even minor TCP retransmit delays between chunks add up

## Symptom

When the 5s timeout fires, the plugin raises `APIConnectionError('Connection error.', retryable=...)` at `llm.py` line ~335 (where `raise APIConnectionError(retryable=retryable) from e` is called).

Traceback chain:
```
httpcore.ReadTimeout
  → httpx.ReadTimeout
  → anthropic._streaming.aiter_bytes (streaming reader)
  → livekit.plugins.anthropic.llm.LLM._run
  → livekit.agents.voice.generation._llm_inference_task (caller)
  → APIConnectionError('Connection error.', retryable=...)
```

If `retryable=False`, the session dies. If `retryable=True`, the SDK retries and re-hits the same timeout, cascading into a loop. In voice pipelines this also starves downstream: STT connections can time out because no audio is being consumed while the LLM path is stuck in retry loops.

Haiku users on simple short-context prompts don't typically see this because first-chunk latency is sub-second for that workload. The problem is silent until someone uses thinking-enabled models or heavy context.

## Workaround

Pass a custom `anthropic.AsyncClient` with a longer httpx read timeout via the `client=` constructor parameter. This is now possible thanks to #4129 being fixed — the plugin correctly honors the passed client.

```python
import httpx
import anthropic
from livekit.plugins.anthropic import LLM

def patient_client():
    return anthropic.AsyncClient(
        api_key=os.getenv("ANTHROPIC_API_KEY"),
        http_client=httpx.AsyncClient(
            timeout=httpx.Timeout(
                connect=5.0,
                read=60.0,      # ← generous for thinking
                write=10.0,
                pool=5.0,
            ),
            follow_redirects=True,
            limits=httpx.Limits(
                max_connections=1000,
                max_keepalive_connections=100,
                keepalive_expiry=120,
            ),
        ),
    )

llm = LLM(model=\"claude-sonnet-4-6\", client=patient_client())
```

Works cleanly on 1.5.4 and 1.5.5 in my testing.

## Suggested fix

Two options, either would resolve this for the default path without requiring every user to override:

1. **Bump the default to ≥30s** (preferably 60s given adaptive thinking's range). Preserves the fail-fast property for true network outages while no longer clipping legitimate thinking turns.

2. **Split the httpx Timeout into separate connect/read values** in the plugin default — keep `connect` tight (5s) so genuine connection failures surface quickly, but allow `read` to be generous (30-60s) so mid-stream pauses during thinking don't kill the call.

Option 2 is probably the better shape since it preserves both goals.

## Context

Found while running a self-hosted LiveKit voice agent with Sonnet 4.6 and adaptive thinking on non-trivial context. The 5s timeout made Sonnet sessions fail within a few turns; Haiku was also affected when context grew or during Anthropic latency spikes.

Versions checked: `livekit-agents` and `livekit-plugins-anthropic` at 1.5.4 and 1.5.5 — both have the same hardcoded 5.0 value.

Related: #4129 (client= parameter honored — closed) is what makes the workaround possible.

SuperMarioYL:
## Summary

Fixes #5508.

The Anthropic LLM plugin used `httpx.AsyncClient(timeout=5.0)`, which sets **all** httpx sub-timeouts — including the per-chunk SSE read timeout — to 5 seconds. Claude's adaptive-thinking phases routinely pause for 10–30 s before emitting the first content chunk, so the 5 s read timeout fires during normal usage and raises `APIConnectionError`, killing voice sessions mid-turn.

### Changes

- **Default split timeout**: `httpx.Timeout(5.0, read=30.0)` — connect stays at 5 s (genuine TCP failures surface fast) while the per-chunk read window is 30 s (covers standard thinking budgets with headroom).
- **New `timeout` constructor parameter**: `timeout: httpx.Timeout | None = None` lets callers supply a custom timeout without constructing a whole `anthropic.AsyncClient` — e.g. `httpx.Timeout(5.0, read=60.0)` for extended thinking or very large contexts. Aligns with the pattern already used by the OpenAI plugin.
- **Six unit tests** in `tests/test_plugin_anthropic.py` cover the default split, tight connect value, custom timeout pass-through, and `client=` precedence over `timeout=`.

### Before / after

```python
# Before — all sub-timeouts at 5 s, kills thinking phases
httpx.AsyncClient(timeout=5.0, ...)

# After — tight connect, generous read
httpx.AsyncClient(timeout=timeout or httpx.Timeout(5.0, read=30.0), ...)
```

### Usage (new parameter)

```python
# Default (30 s read) — sufficient for most models
llm = anthropic.LLM(model="claude-sonnet-4-6")

# Extended thinking or very large contexts
llm = anthropic.LLM(
    model="claude-opus-4-6",
    timeout=httpx.Timeout(5.0, read=60.0),
)
```

## Test plan

- [x] `uv run pytest tests/test_plugin_anthropic.py -v` — 6/6 pass
- [x] `uv run ruff check` — no errors
- [x] `uv run ruff format --check` — no changes needed

### DIFF
--- a/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
+++ b/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
@@ -75,6 +75,7 @@ class LLM(llm.LLM):
         parallel_tool_calls: NotGivenOr[bool] = NOT_GIVEN,
         tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
         caching: NotGivenOr[Literal["ephemeral"]] = NOT_GIVEN,
+        timeout: httpx.Timeout | None = None,
         _strict_tool_schema: bool = True,
     ) -> None:
         """
@@ -88,6 +89,12 @@ class LLM(llm.LLM):
         base_url (str, optional): The base URL for the Anthropic API. Defaults to None.
         user (str, optional): The user for the Anthropic API. Defaults to None.
         client (anthropic.AsyncClient | None): The Anthropic client to use. Defaults to None.
+        timeout (httpx.Timeout | None): HTTP timeout configuration for the underlying httpx client.
+            Defaults to ``httpx.Timeout(5.0, read=30.0)``, which keeps a tight connect timeout
+            while allowing up to 30 s between streamed chunks — long enough for Claude's
+            adaptive-thinking phases without masking genuine network stalls.
+            Pass a custom ``httpx.Timeout`` to override (e.g. ``httpx.Timeout(5.0, read=60.0)``
+            for very large contexts or extended thinking budgets).
         temperature (float, optional): The temperature for the Anthropic API. Defaults to None.
         parallel_tool_calls (bool, optional): Whether to parallelize tool calls. Defaults to None.
         tool_choice (ToolChoice, optional): The tool choice for the Anthropic API. Defaults to "auto".
@@ -118,7 +125,7 @@ class LLM(llm.LLM):
             api_key=anthropic_api_key,
             base_url=base_url if is_given(base_url) else None,
             http_client=httpx.AsyncClient(
-                timeout=5.0,
+                timeout=timeout or httpx.Timeout(5.0, read=30.0),
                 follow_redirects=True,
                 limits=httpx.Limits(
                     max_connections=1000,
--- a/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
+++ b/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
@@ -75,7 +75,7 @@ class LLM(llm.LLM):
         parallel_tool_calls: NotGivenOr[bool] = NOT_GIVEN,
         tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
         caching: NotGivenOr[Literal["ephemeral"]] = NOT_GIVEN,
-        timeout: httpx.Timeout | None = None,
+        timeout: NotGivenOr[httpx.Timeout] = NOT_GIVEN
         _strict_tool_schema: bool = True,
     ) -> None:
         """
--- a/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
+++ b/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
@@ -75,7 +75,7 @@ class LLM(llm.LLM):
         parallel_tool_calls: NotGivenOr[bool] = NOT_GIVEN,
         tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
         caching: NotGivenOr[Literal["ephemeral"]] = NOT_GIVEN,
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         parallel_tool_calls: NotGivenOr[bool] = NOT_GIVEN,
         tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
         caching: NotGivenOr[Literal["ephemeral"]] = NOT_GIVEN,
+        timeout: httpx.Timeout | None = None,
```
longcw: maybe use NOT_GIVEN
```suggestion
        timeout: NotGivenOr[httpx.Timeout] = NOT_GIVEN
```

longcw (approved):

gdoermann:
This doesn't touch _awaitable_anthropic_stream (line 278), the if not self._anthropic_stream: self._anthropic_stream = await ... block (lines 295-299), or the exception path. So the deep bug is still latent — but with the trigger frequency gone, it should rarely fire in practice.

livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         parallel_tool_calls: NotGivenOr[bool] = NOT_GIVEN,
         tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
         caching: NotGivenOr[Literal["ephemeral"]] = NOT_GIVEN,
+        timeout: NotGivenOr[httpx.Timeout] = NOT_GIVEN
```
longcw:
```suggestion
        timeout: NotGivenOr[httpx.Timeout] = NOT_GIVEN,
```

PR #5531, opened 2026-04-23
fix(openai): forward session.update on RealtimeModel.update_options

### LINKED ISSUE #5530: RealtimeModel.update_options() does not send session.update to OpenAI (race condition with _opts mutation)
## Summary

`RealtimeModel.update_options()` mutates the model-level `_opts` **before** forwarding to active sessions, causing `RealtimeSession.update_options()` to see no diff and skip sending `session.update` to OpenAI.

## Environment

- `livekit-agents`: 1.5.x
- `livekit-plugins-openai`: 1.5.x
- Model: `gpt-realtime-1.5`

## Steps to Reproduce

1. Start an OpenAI Realtime session via LiveKit
2. Call `agentSession._llm.update_options(speed=1.5)` (or any other option like `voice`, `turn_detection`, etc.)
3. Observe that no `session.update` event is sent to OpenAI
4. The change has no audible/observable effect

## Root Cause

In [`realtime_model.py`](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py):

**`RealtimeModel.update_options()`** (around line 649):
```python
# Mutates _opts FIRST
if is_given(speed):
    self._opts.speed = speed

# THEN forwards to sessions
for sess in self._sessions:
    sess.update_options(speed=speed, ...)
```

**`RealtimeSession.update_options()`** (around line 1207):
```python
if is_given(speed):
    if self._realtime_model._opts.speed != speed:  # Always False! Already mutated above
        audio_output.speed = speed
        has_audio_config = True
    self._realtime_model._opts.speed = speed
```

The session compares against the model's `_opts`, which was already mutated by the model-level call. The diff always sees "no change" and skips sending the update.

## Affected Options

All options using this pattern:
- `speed`
- `voice`
- `turn_detection`
- `input_audio_transcription`
- `input_audio_noise_reduction`
- `tool_choice`
- `tracing`
- `max_response_output_tokens`

## Workaround

Call `RealtimeSession.update_options()` directly on active sessions:

```python
for sess in llm._sessions:
    sess.update_options(speed=1.5)
```

This works because the session-level call hasn't had its comparison poisoned by prior mutation.

## Suggested Fix

In `RealtimeModel.update_options()`, forward to sessions **before** mutating `_opts`:

```python
def update_options(self, *, speed=NOT_GIVEN, ...):
    # Forward to sessions FIRST (while _opts still has old values)
    for sess in self._sessions:
        sess.update_options(speed=speed, ...)
    
    # THEN mutate model opts
    if is_given(speed):
        self._opts.speed = speed
    ...
```

## Additional Context

- The workaround requires accessing `llm._sessions`, which is a private attribute
- There's no public API to enumerate active sessions from `AgentSession`
- Initial session setup works correctly (only live updates are broken)

## Related Issues

- #3466 (similar `update_options` / `session.update` issues with `noise_reduction`)
- #3445 (voice changing unexpectedly, related to session update handling)

longcw:
## Summary

`RealtimeSession.update_options()` compared against the shared `RealtimeModel._opts`, but `RealtimeModel.update_options()` mutated that same dict before forwarding to sessions. The diff always saw "no change" and no `session.update` was sent to OpenAI.

Give each `RealtimeSession` its own `_opts` copy so the per-session diff is independent of the model-level state and of any other sessions sharing the same model.

Fixes #5530

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -9,7 +9,7 @@ import os
 import time
 import weakref
 from collections.abc import Iterator
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from typing import Any, Literal, overload
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
@@ -794,7 +794,7 @@ class RealtimeSession(
         super().__init__(realtime_model)
         self._realtime_model: RealtimeModel = realtime_model
         # per-session copy of opts so update_options can diff against session's own state
-        self._opts = copy.copy(realtime_model._opts)
+        self._opts = replace(realtime_model._opts)
         self._tools = llm.ToolContext.empty()
         self._msg_ch = utils.aio.Chan[RealtimeClientEvent | dict[str, Any]]()
         self._input_resampler: rtc.AudioResampler | None = None
@@ -826,12 +826,12 @@ class RealtimeSession(
     @utils.log_exceptions(logger=logger)
     async def _main_task(self) -> None:
         num_retries: int = 0
-        max_retries = self._realtime_model._opts.conn_options.max_retry
+        max_retries = self._opts.conn_options.max_retry
 
         async def _reconnect() -> None:
             logger.debug(
                 f"reconnecting to {self._realtime_model._provider_label}",
-                extra={"max_session_duration": self._realtime_model._opts.max_session_duration},
+                extra={"max_session_duration": self._opts.max_session_duration},
             )
 
             events: list[RealtimeClientEvent | dict[str, Any]] = []
@@ -864,10 +864,7 @@ class RealtimeSession(
                             by_alias=True, exclude_unset=True, exclude_defaults=False
                         )
 
-                    if (
-                        self._realtime_model._opts.is_azure
-                        and self._realtime_model._opts.api_version
-                    ):
+                    if self._opts.is_azure and self._opts.api_version:
                         _normalize_azure_client_event(ev)
 
                     self.emit("openai_client_event_queued", ev)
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1163,8 +1165,8 @@ def _wrap_session_update(
     def _create_session_update_event(self) -> SessionUpdateEvent | dict[str, Any]:
         audio_format = realtime.realtime_audio_formats.AudioPCM(rate=SAMPLE_RATE, type="audio/pcm")
         # they do not support both text and audio modalities, it'll respond in audio + transcript
-        modality = "audio" if "audio" in self._realtime_model._opts.modalities else "text"
-        opts = self._realtime_model._opts
+        modality = "audio" if "audio" in self._opts.modalities else "text"
```
chenghao-mou, line 1160: nit: There are several `self._realtime_model._opts.modalities` usage in other paths in this file, should we standardize them?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -793,6 +793,8 @@ class RealtimeSession(
     def __init__(self, realtime_model: RealtimeModel) -> None:
         super().__init__(realtime_model)
         self._realtime_model: RealtimeModel = realtime_model
+        # per-session copy of opts so update_options can diff against session's own state
+        self._opts = copy.copy(realtime_model._opts)
```
chenghao-mou: nit: should we use `dataclasses.replace(realtime_model._opts)`

chenghao-mou (approved):
Nice catch! Two nits.

toubatbrian:

Detected classification: **plugin fix** (OpenAI Realtime). This PR fixes a per-session state diff bug in `RealtimeSession.update_options()`, which is plugin-level behavior that also applies to the JS port. Initiating a port to [livekit/agents-js](https://github.com/livekit/agents-js) now.

---

PR #5535, opened 2026-04-23
fix(voice): pause output when user starts speaking during thinking

### LINKED ISSUE #5509: Agent transitions `thinking → speaking` after user has already transitioned to `speaking` (stale reply plays over new user turn)
### Summary

We are reliably reproducing a race on `livekit-agents ~= 1.5.2`: the agent enters `speaking` and begins TTS playout even though the `UserStateChangedEvent` for `listening → speaking` has already fired milliseconds earlier. The stale reply gets played over the user's new turn, causing both sides to talk at once and typically resulting in a bad interruption/EOT resolution.

We shipped a workaround on our side that works, but we believe the SDK should guard this internally — once the session has observed the user transitioning into `speaking`, it shouldn't let a reply that is still in `thinking` (no audio emitted yet) promote to `speaking`.

The bug is visible when turn detection is driven by both VAD and STT.

### Reproduction — real call logs

In this call the LLM generation for the previous user turn had just started when the user began a new turn. The SDK still flipped the agent to `speaking`:

```
02:43:58.700  Agent state changed: listening → thinking
02:43:59.094  VAD speech detected (triggering_probability=0.90)
02:43:59.095  User state changed: away → speaking
02:43:59.144  VAD waiting for EOS (probability=0.92, min_silence_required=1.10s)
02:43:59.264  Agent state changed: thinking → speaking   ← should not happen
02:43:59.299  ElevenLabs TTS request completed
```

Between the user going `speaking` (02:43:59.095) and the agent going `speaking` (02:43:59.264) there is ~170 ms — plenty of time for the SDK to observe the user state transition and cancel the pending generation.

A full chat history + traces + logs bundle from one of these calls is available on request — we didn't want to paste customer data in a public issue.

### Our workaround

In our `user_state_changed` handler, when the new state is `speaking` and `session.agent_state == "thinking"`, we call `session.interrupt()` to cancel the in-flight generation before it can promote to `speaking`. The committed user turn's `on_end_of_turn` path then regenerates naturally against the new transcript.

Guardrails we had to add:

- Skip when `current_speech.num_steps > 1` — otherwise we cancel the follow-up generation that runs after a tool call (e.g. agent handoff).
- Skip when `current_speech.allow_interruptions` is `False` — `SpeechHandle.interrupt(force=False)` raises `RuntimeError` for non-interruptible speeches (our initial greeting, mandatory compliance messages, etc.).
- Wrap `session.interrupt()` in `try/except RuntimeError` as defence in depth for non-interruptible speeches sitting in the scheduling queue.

### How often we see it

We added a metric (`agent.thinking_interrupted_by_user`) that increments every time our workaround fires. Across **72 completed calls** with **~51s average duration**, the metric fired **127 times** — so this is not a rare edge case, it fires multiple times per call in some conversations. Since shipping the workaround we haven't seen the bug in call reviews.

### Why we think this belongs in the SDK

- Every agent hitting this race has to reimplement the same `user_state_changed` listener with the same guardrails (tool-call follow-ups, non-interruptible speeches). That logic is easy to get wrong.
- The SDK already knows the user's state and the agent's state — it is in a much better position than application code to break the tie atomically.
- Semantically, if the user is already in `speaking` when a reply is still in `thinking` (nothing audible yet), promoting it to `speaking` is never the right call.

### Suggested fix direction

Inside the session scheduler, before allowing a speech to transition from `thinking → speaking`, check whether the user is currently in `speaking` state. If yes, cancel the pending speech (same semantics as `session.interrupt()` does today) rather than starting TTS playout. Respect `allow_interruptions=False` and the tool-call follow-up case the same way we had to externally.

If maintainers agree with this direction, happy to open a PR.

### Versions

```
livekit-agents ~= 1.5.2 (with deepgram/soniox/silero/turn-detector extras)
livekit-api >= 1.1.0, < 2
```

longcw:
## Summary

Fixes livekit/agents#5509. When the user starts a new turn while the agent is in `thinking` (LLM generating, TTS audio not yet flowing), the stale reply could still reach `speaking` and play over the new user turn. This pauses the pausable audio output on VAD SOS before any TTS frame hits the transport, so the stale reply neither plays nor promotes to `speaking`.

## Changes

**`livekit-agents/livekit/agents/voice/agent_activity.py`**
- New `on_start_of_speech` path: when the agent is not `speaking`, pauses the output if it supports pause. Uses `timeout=0` so a brief false-positive VAD resumes immediately on VAD EOS; `_interrupt_by_audio_activity` upgrades to the real timeout if VAD confirms active speech.
- `_update_paused_speech` preserves the `agent_state` captured at first pause across repeat calls
- `_PausedSpeechInfo` carries `handle + agent_state + timeout` so resume restores the exact state (e.g. `thinking`, not `speaking`) that was active when the pause began.

**Test helpers (`tests/fake_*.py`)**
- `FakeAudioOutput(can_pause=True)` supports pause now
- `FakeUserSpeech` with empty `transcript` fires VAD SOS/EOS only (no STT events) — simulates sub-`min_duration` noise.

## Known limitation

- With timeout=0 on the new pause, if the user's speech is shorter than min_interruption_duration but STT still produces a final transcript that arrives after VAD EOS, the paused agent speech briefly resumes (timer fires on EOS) and then gets interrupted when the STT transcript lands — so the user hears a short snippet of the stale reply before it cuts off.


miguelmoralai:
Nice approach — pausing sidesteps the idle risk entirely, no backstop needed.

One thought on the known limitation (short utterance → brief audio blip before cut-off): the resume on VAD EOS could be gated on whether the current STT interim has non-empty content. Empty interim → blip was noise, safe to resume. Non-empty interim → real speech, skip resume and let the interrupt path handle it. That's the heuristic we landed on for our own workaround and it eliminated that race cleanly.

chenghao-mou (approved):
lgtm. It worked well when I spoke during its thinking state.

toubatbrian:

Tracking: this core-runtime fix (voice/agent_activity.py — pause output when user starts speaking during `thinking`) qualifies for porting. A corresponding PR will be opened in `livekit/agents-js` shortly.

---

longcw:
@miguelmoralai 
>  the resume on VAD EOS could be gated on whether the current STT interim has non-empty content. 

thanks for the advise, yeah there is already an interruption logic on interim transcript, if non-empty interim transcript received, the timeout for resume will be updated to the default value (e.g. 2 seconds).

PR #5548, opened 2026-04-24
feat(xai): support model selection in realtime, default to grok-voice-think-fast-1.0

Hormold:
x.ai deprecated grok-voice-fast-1.0 and recommends grok-voice-think-fast-1.0 as the new flagship voice model, selectable via the model query param on the realtime websocket (see https://docs.x.ai/developers/model-capabilities/audio/voice-agent). This patch adds a model kwarg to xai.realtime.RealtimeModel, defaults it to grok-voice-think-fast-1.0, and exposes a GrokRealtimeModels literal with both the new and legacy ids; the parent openai realtime client already forwards model as a ws query param so the resulting url matches the x.ai docs exactly.


theomonnom (approved):

galigutta:
when is this going to be available in a release?

StanislavSava:
Is there a plan to migrate the python plugin to this model as well?

PR #5554, opened 2026-04-25
Clear AgentSession close task after shutdown

jayeshp19:


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1399,13 +1404,10 @@ def _on_error(
...
             )
 
-        def on_close_done(_: asyncio.Task[None]) -> None:
-            self._closing_task = None
-
         self._closing_task = asyncio.create_task(
             self._aclose_impl(error=error, reason=CloseReason.ERROR)
         )
-        self._closing_task.add_done_callback(on_close_done)
+        self._closing_task.add_done_callback(self._clear_closing_task)
```
longcw: let's move the on_close_done to `_close_soon` and call `_close_soon` here? `_clear_closing_task` doesn't need to be a class method.

PR #5556, opened 2026-04-26
fix(google): pause audio input during synchronous tool execution on t…

vedevpatel:
Gemini 3.1 live model

Gemini 3.1 forces synchronous tool calling, which means the model blocks until tool responses arrive. The plugin's _send_task was constantly forwarding microphone audio while tools executed, which caused the server to think of incoming audio as a new turn and cancel the pending tool call after ~12s. This caused duplicate tool execution with already-resolved call_ids as well as corrupted conversation state.

1. Adds a _tool_call_pending flag (for Gemini 3.1 only) that drops push_audio frames from the moment a toolCall is received until send_tool_response is flushed. 
2. Also clears the flag on tool_call_cancellation so the session never stalls. No behavior change for Gemini 2.5 models.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -495,6 +495,11 @@ class RealtimeSession(llm.RealtimeSession):
         self._session_lock = asyncio.Lock()
         self._num_retries = 0
 
+        # true while synchronous tool call is in flight for 3.1 only
+        # Audio frames dropped here to prevent server from thinking incoming audio is a
+        # new turn and cancelling the pending tool call
+        self._tool_call_pending = False
+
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -1294,6 +1303,9 @@ def _handle_tool_calls(self, tool_call: types.LiveServerToolCall) -> None:
                 )
             )
         self._mark_current_generation_done()
+        if "3.1" in self._opts.model:
```
longcw: first of all, I don't think dropping the audio when there is in-flight tool call is a right solution, but I am wondering why here it's only applied to 3.1?

fyi, we support gemini `NON_BLOCKING` tool call via `tool_behavior` option, you may check that option instead of using the model name? or even making this configurable when the tool behavior is blocking?

PR #5559, opened 2026-04-27
feat(mcp): allow updating headers on MCPServerHTTP

### LINKED ISSUE #5542: Ability to update the authorization header of an MCP
### Feature Type

I cannot use LiveKit Agent MCP functionality without it

### Feature Description

We have an MCP that sits in front of our cloud APIs. The MCP requires bearer token which is passed on to the cloud APIs. These tokens are short lived, in the range of minutes. 

The authorization header can be set with mcp.MCPServerHTTP constructor. However, there is no way to update the header later on. Our workaround is to add a new MCP which A) we are not sure if there is service interruption in between, B) We use lots of tokens re-initializing the MCP. 

It should be pretty easy to offer the ability to update the Authorization header. 

What do you think? Do you have other concerns regarding updating the header? 

_No response_

### Additional Context

_No response_

longcw:
## Summary
- Expose `MCPServerHTTP.headers` as a property with a setter that propagates updates to the active `httpx.AsyncClient`.

Closes #5542

## Usage
```python
mcp_server = MCPServerHTTP(url=..., headers={"Authorization": f"Bearer {token}"})
await mcp_server.initialize()

# Later, when the token rotates:
mcp_server.headers = {"Authorization": f"Bearer {new_token}"}
```

### DIFF
--- a/livekit-agents/livekit/agents/llm/mcp.py
+++ b/livekit-agents/livekit/agents/llm/mcp.py
@@ -262,7 +262,7 @@ class MCPServerHTTP(MCPServer):
             tool_result_resolver=tool_result_resolver,
         )
         self.url = url
-        self.headers = headers
+        self._headers = headers or {}
         self._timeout = timeout
         self._sse_read_timeout = sse_read_timeout
         self._allowed_tools = set(allowed_tools) if allowed_tools else None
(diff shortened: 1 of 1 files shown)


theomonnom (approved):
Lg! but let's deprecate MCPServer in favor of MCPToolset anyyway

longcw:
> let's deprecate MCPServer in favor of MCPToolset anyyway


I think we'll deprecate MCPServer for AgentSession but we still need to pass it to the MCPToolset right?

toubatbrian:

---

**Triage result: skipping port**

This PR adds a `headers` property setter on `MCPServerHTTP` that propagates updates to the live `httpx.AsyncClient`. The change is conceptually portable (an MCP runtime/plugin enhancement), but `agents-js` does not yet have an MCP server implementation — there is no equivalent `MCPServerHTTP` class, no `agents/src/llm/mcp.ts`, and no MCP-related exports in `agents/src/llm/index.ts`.

Porting this incremental enhancement in isolation isn't actionable without first porting the underlying MCP server infrastructure (transport selection between SSE / streamable HTTP, client lifecycle, tool listing/invocation, etc.), which is well beyond the scope of a single PR port.

No `agents-js` PR will be opened for this change. When MCP support lands in `agents-js`, the headers-setter behavior should be implemented at that time (the JS equivalent would expose `headers` as a getter/setter and update the underlying `fetch`/transport options on mutation).

cc @toubatbrian @livekit/agent-devs

---

PR #5560, opened 2026-04-27
fix(gemini live): use parameters instead of parameters_json_schema for raw schema function tools

longcw:
## Summary

- Gemini Live doesn't accept `parameters_json_schema` on `FunctionDeclaration`, so this PR switches the realtime session to the `parameters` field (simplified via `_GeminiJsonSchema`).
- The regular Gemini LLM keeps using `parameters_json_schema` to preserve the richer JSON Schema features it supports.

Closes #5450

Ref: https://github.com/googleapis/python-genai/issues/1147

### DIFF
--- a/livekit-agents/livekit/agents/llm/_provider_format/google.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/google.py
@@ -121,6 +121,7 @@ def to_fnc_ctx(
     tool_ctx: llm.ToolContext,
     *,
     tool_behavior: TOOL_BEHAVIOR | None = None,
+    use_parameters_json_schema: bool = True,
 ) -> list[dict[str, Any]]:
     tools: list[dict[str, Any]] = []
     for tool in tool_ctx.function_tools.values():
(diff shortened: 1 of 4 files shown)


tinalenguyen (approved):

toubatbrian:

Heuristic classification: this PR touches the Google plugin's Gemini Live realtime integration (plugin work, not Python-only deps), so it is in scope for porting. A draft PR will be opened against `livekit/agents-js` shortly.

---

toubatbrian:
Update from the automated routine: after inspecting `livekit/agents-js`, **no port is being opened** for this PR.

Reasoning:
- This fix targets *raw-schema* function tools (`info.raw_schema` -> `FunctionDeclaration`) being routed through `parameters_json_schema`, which Gemini Live rejects. agents-js has no `rawSchema` / raw-schema tool concept yet — function tools always go through `llm.toJsonSchema(tool.parameters, false)`.
- The JS Google plugin's `toFunctionDeclarations` (`plugins/google/src/utils.ts`) already emits the simplified `parameters` field via `convertJSONSchemaToOpenAPISchema(...)`, which is the same destination Python's Gemini Live is now switching to. Both regular `LLM` and `RealtimeModel` in agents-js consume that helper, so Gemini Live in JS was never affected by the `parameters_json_schema` regression this PR fixes.

If a future PR introduces raw-schema tools to agents-js (or splits the regular Gemini LLM onto `parameters_json_schema` for richer JSON Schema features), this same special-case will need to be ported then. For now there's nothing actionable on the JS side.

cc @toubatbrian

---

PR #5561, opened 2026-04-27
Update codec config, MIME mapping, and interruption handling

dhruvladia-sarvam:
- Always send output_audio_codec in WebSocket config (was only in
  model-specific branches, silently omitted for unmatched models)
- Add _codec_to_mime_type() mapping so linear16/mulaw/alaw correctly
  resolve to audio/pcm instead of invalid MIME types
- Expand ALLOWED_OUTPUT_AUDIO_CODECS to include linear16, mulaw, alaw
- Guard send_task and recv_task against ConnectionResetError during
  barge-in so transport-closing errors are suppressed, not re-raised
- Add input_sent_event coordination between send/recv tasks to prevent
  recv from starting before config is sent
- Simplify aclose() to delegate to base class, avoiding race with
  ConnectionPool cleanup

### DIFF
--- a/livekit-agents/livekit/agents/utils/codecs/decoder.py
+++ b/livekit-agents/livekit/agents/utils/codecs/decoder.py
@@ -59,6 +59,8 @@ def _mime_to_av_format(mime: str | None) -> str | None:
         "audio/ogg": "ogg",
         "audio/webm": "webm",
         "audio/mp4": "mp4",
+        "audio/mulaw": "mulaw",
+        "audio/alaw": "alaw",
     }
     return _TABLE.get(mime)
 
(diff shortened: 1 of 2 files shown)


tinalenguyen:
hi, thanks for the PR! i tried it and got this log: `TTS API error: Input parameters has to be a valid dictionary`

dhruvladia-sarvam:
> i tried it and got this log: `TTS API error: Input parameters has to be a valid dictionary`

You must've received it for bulbul:v2, right?

tinalenguyen:
Yes, the default model

dhruvladia-sarvam:
@tinalenguyen please re-review

dhruvladia-sarvam:
@tinalenguyen requesting a re-review

tinalenguyen:
hi @dhruvladia-sarvam, i still hit the same error when trying it out

tinalenguyen (approved):
lgtm, thanks!!

PR #5562, opened 2026-04-27
fix(smallestai): use close_stream signal to properly terminate STT session

harshitajain165:
## Summary

- Replaces `{"type": "finalize"}` with `{"type": "close_stream"}` when the STT input channel closes
- `finalize` only forces `is_final` on pending audio but keeps the WebSocket session open, causing `recv_task` to hang indefinitely waiting for `is_last=true` that never arrives
- `close_stream` (added in Waves v4.0.0) flushes remaining audio, emits final transcripts, and triggers `is_last=true` so the session terminates cleanly

## Test plan

- [ ] Run a streaming STT session and verify the WebSocket closes cleanly after audio ends (no hang in `recv_task`)
- [ ] Verify final transcript is still delivered before close
- [ ] Confirm `{"type": "finalize"}` mid-session behaviour is unaffected (not used in this path)

### DIFF
--- a/livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/stt.py
+++ b/livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/stt.py
@@ -280,9 +280,10 @@ class STT(stt.STT):
 
 
 class SpeechStream(stt.SpeechStream):
-    # Tells the server to flush remaining audio and close the session gracefully.
-    # After sending this, the server will respond with a final message (is_last=True).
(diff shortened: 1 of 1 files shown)


harshitajain165:
`{"type": "finalize"}` only forces is_final but keeps the session open, causing recv_task to hang waiting for is_last: true indefinitely. Replaced with `{"type": "close_stream"} `(added in Waves v4.0.0) which properly flushes, delivers final transcripts, and closes the session. 

@tinalenguyen @theomonnom Could you please hep with review?

tinalenguyen (approved):
thanks for the fix!

PR #5568, opened 2026-04-27
Hotfix; Updated default Avatar ID

hari-truviz:
Old avatar was retried and was replaced with new avatars.


tinalenguyen (approved):
thank you! 😸

Issue #5566, opened 2026-04-27
OpenAI Realtime: response.error mid-flight causes future to hang indefinitely (TODO at realtime_model.py:2005)


cphoward:
## TL;DR

OpenAI Realtime: response.error mid-flight causes future to hang indefinitely (TODO at realtime_model.py:2005)

## Summary

When the OpenAI Realtime substrate sends a `response.error` event mid-flight (during a generation), the future associated with that response is never resolved — it hangs indefinitely. Callers awaiting the future block forever.

The source code at [`livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py:2005`](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py) acknowledges this with a TODO comment.

## Reproduction

```python
from livekit.plugins.openai.realtime import RealtimeModel

session = RealtimeModel(model="gpt-4o-realtime-preview-2025-06-03").session()
fut = session.generate_reply(instructions="...")
# If OpenAI sends response.error mid-stream (e.g., due to substrate
# rate limit, content policy violation, or similar), the future
# `fut` will never resolve.
try:
    result = await asyncio.wait_for(fut, timeout=30.0)
except asyncio.TimeoutError:
    # This is the only way to detect the hang from the caller side
    print("future hung; substrate sent response.error and we didn't handle it")
```

## Source code reference

Around line 2005 in `realtime_model.py`, in `_handle_response_error` (or equivalent):

```python
def _handle_response_error(self, event: ResponseErrorEvent) -> None:
    # TODO: handle response.error mid-flight
    # Currently the response future hangs indefinitely.
    ...
```

(Exact TODO text and surrounding code may differ in current main; the line number reference is from the version this issue was authored against.)

## Proposed fix

When `response.error` arrives mid-flight, the handler should:

1. Identify the affected response by `event.response_id`
2. Pop the future from `_response_created_futures` (or equivalent state)
3. Resolve the future with `RealtimeError(event.error.message)` or similar
4. Clean up any associated state (e.g., the in-flight `_ResponseGeneration` entry)

```python
def _handle_response_error(self, event: ResponseErrorEvent) -> None:
    """Handle response.error: reject the associated future and clean up."""
    response_id = event.response_id  # may need to derive from event shape
    error_msg = event.error.message if event.error else "unknown error"

    # Reject any pending response_created future (matches by client_event_id
[...]

Issue #5565, opened 2026-04-27
RealtimeCapabilities has 6 undocumented fields + 1 misleading docstring


cphoward:
## TL;DR

RealtimeCapabilities has 6 undocumented fields + 1 misleading docstring

## Summary

The `RealtimeCapabilities` dataclass in `livekit-agents/livekit/agents/llm/realtime.py` has 11 fields. **6 of them have no docstrings**, and the `supports_say` field has a docstring that's misleading post-`add_to_chat_ctx` parameter introduction.

This is a small documentation-hygiene issue but matters because plugin authors implementing new Realtime backends use this dataclass to declare their capabilities; ambiguous semantics produce wrong `True`/`False` decisions.

## Undocumented fields

The following 6 fields lack docstrings (all in `livekit-agents/livekit/agents/llm/realtime.py`, around lines 60-75):

1. `message_truncation: bool`
2. `turn_detection: bool`
3. `user_transcription: bool`
4. `auto_tool_reply_generation: bool`
5. `audio_output: bool`
6. `manual_function_calls: bool`

## Misleading docstring

`supports_say` currently says: `"Whether the model supports session.say()"`.

This is misleading because `RealtimeModel` implementations that declare `supports_say=True` typically only support `say()` when `add_to_chat_ctx=True`. Plugin authors reading this docstring may incorrectly infer they should declare `supports_say=True` even when they cannot honor `add_to_chat_ctx=False`.

(Recommendation: amend the `supports_say` docstring AND add a sibling `ephemeral_say` field for the isolation capability — that work is happening in a separate PR.)

## Proposed docstrings

Verified against actual usage in `agent_activity.py` (line numbers from main):

```python
@dataclass
class RealtimeCapabilities:
    message_truncation: bool
    """Whether the substrate supports per-message truncation via a
    truncate() call (e.g., OpenAI's conversation.item.truncate). Used
    during user interruption to cut off in-flight audio responses by
    message_id."""

    turn_detection: bool
    """Whether the substrate provides server-side voice activity
    detection (VAD) for turn-taking. When True, AgentActivity relies on
    the substrate's turn detection events; when False, AgentActivity
    must orchestrate turns client-side."""

    user_transcription: bool
    """Whether the substrate provides server-side transcription of user
    audio. When True, AgentActivity skips client-side STT processing to
    avoid duplicate transcripts."""

    auto_tool_reply_generation: bool
    """Whether the substrate automatically generates a new response
    after tool execution results are submitted. When False,
[...]

Issue #5564, opened 2026-04-27
RealtimeSession.interrupt() is a no-op without response_id under concurrent in-flight responses (OpenAI plugin)


cphoward:
## TL;DR

RealtimeSession.interrupt() is a no-op when called without response_id under concurrent in-flight responses (OpenAI plugin)

## Summary

`RealtimeSession.interrupt()` in the OpenAI Realtime plugin sends a `ResponseCancelEvent(type="response.cancel")` without specifying `response_id`. Per empirical OpenAI Realtime API behavior, **a cancel event without `response_id` is a no-op when more than one response is in flight** — it does not cancel "all" responses; it cancels nothing.

Today this is harmless because the plugin's single-slot `_current_generation` model only allows one in-flight response at a time, so "the only in-flight response" is unambiguous and the substrate accepts the cancel.

This becomes broken under any future change that enables concurrent in-flight responses on a single session (e.g., a refactor of `_current_generation` to a multi-instance state container, which is needed to correctly represent the substrate's existing concurrent-OOB-response capability per [Research notes](#empirical-evidence) below).

## Reproduction (current state — no concurrency, so this is latent)

```python
from livekit.plugins.openai.realtime import RealtimeModel

session = RealtimeModel(model="gpt-4o-realtime-preview-2025-06-03").session()
# Today: only one response.create can be in flight at a time
# (single-slot _current_generation), so interrupt() works correctly.
fut = session.generate_reply(instructions="Speak slowly: aaaaaa")
await asyncio.sleep(0.5)
session.interrupt()  # cancels the only in-flight response — works fine
```

## Reproduction under hypothetical concurrent-response refactor (the latent bug)

```python
# Hypothetical: dict-based _generations replacing single-slot
fut_a = await session._fire_oob("Long response A " + " ".join(["a"] * 100))
fut_b = await session._fire_oob("Long response B " + " ".join(["b"] * 100))
await asyncio.sleep(0.5)
# Both responses now in flight (substrate supports this for OOB)
session.interrupt()  # NO-OP: cancels neither response
# Both responses continue to completion; user hears unwanted audio
```

## Empirical evidence

I ran probes against the OpenAI Realtime API (`gpt-4o-realtime-preview-2025-06-03`) to characterize cancel semantics:

- **Cancel WITH `response_id` while concurrent responses are in flight**: cancels the targeted response cleanly. Other in-flight responses continue.
[...]

PR #5569, opened 2026-04-27
feat(realtime/openai): add ephemeral say() via response.create(conversation: "none")

cphoward:
### Summary

Adds an ephemeral path to `RealtimeSession.say()` for the OpenAI Realtime plugin
via `response.create(conversation: "none")`: text rendered with
`add_to_chat_ctx=False` is heard by the user but does NOT enter the agent's
reasoning context, server-side OR local.

This closes the JS PR #1193 parity gap and addresses the silent-degradation
regression in that review.

### How to review

Suggested reading path (≈30 min):
1. `livekit-agents/livekit/agents/llm/realtime.py` — capability shape (`ephemeral_say` field) and abstract `say()` signature, ~30 lines.
2. `livekit-plugins/livekit-plugins-openai/.../realtime/realtime_model.py` — `say()` body and the orphan filter in `_handle_response_created`, ~140 lines.
3. `livekit-agents/livekit/agents/voice/agent_activity.py` — dispatcher capability check, plumbing, and the local-context gate at `_realtime_generation_task_impl`, ~30 lines.
4. `tests/test_realtime/test_realtime_say_integration.py` — the property contract for the feature; the unit test files are proofs of the same properties at finer granularity.

### Cross-plugin scope

Only the OpenAI plugin gains the substrate implementation. Phonic accepts but
ignores the kwarg (structurally exempt — substrate is strictly turn-based per
docs.phonic.co; no out-of-band response primitive). Google Gemini Live, AWS
Nova Sonic, Ultravox, NVIDIA all declare `supports_say=False` today; this PR
doesn't change that.

A new `RealtimeCapabilities.ephemeral_say: bool = False` field is added.
Plugins set it to `True` when their substrate provides an out-of-band response
primitive that does not enter conversation state. The OpenAI plugin declares
`ephemeral_say=True`. All other plugins default to `False`.

### Backward compatibility — caller side (deprecation cycle)

In this release, when a caller invokes `say(text, add_to_chat_ctx=False)` on a
plugin without `ephemeral_say`, we emit `DeprecationWarning` and continue with
the existing silent-degrade behavior. A future release will replace the warning
with `NotImplementedError`. This preserves backward compat for any callers
depending on the silent-degrade contract while loudly surfacing the capability
gap. Open to shipping the immediate-raise version if you'd prefer.

### Backward compatibility — plugin authors

The abstract `RealtimeSession.say()` signature gains a keyword-only
`add_to_chat_ctx: bool = True` parameter, and the dispatcher unconditionally
passes the kwarg via `_rt_session.say(text, add_to_chat_ctx=...)`. This is a
**breaking change for any third-party plugin that BOTH declares
`supports_say=True` AND has overridden `say()` with the original signature
`(self, text)`** — those overrides will hit `TypeError: say() got an unexpected
keyword argument 'add_to_chat_ctx'`.

In-tree plugins are updated:
- OpenAI plugin's `say()` accepts and honors the kwarg.
- Phonic accepts but ignores (signature-only shim; substrate is structurally
  exempt, dispatcher emits `DeprecationWarning` upstream).
- Gemini Live, AWS Nova Sonic, Ultravox, NVIDIA all declare `supports_say=False`
  today and the dispatcher never reaches them.

If preserving signature-level backward-compat for out-of-tree plugin authors is
preferred, the dispatcher could `try/except TypeError` and fall back to the
old call shape. Happy to add the shim — flagging this so you can choose
deliberately.

### Tests prove the isolation contract

Each property below is falsifiable: if the assertion fails, the named test fails. Tests run against `gpt-4o-realtime-preview-2025-06-03` for integration paths and network-free for the local-context-gate and orphan-filter behavioral tests.

| Property | Test |
|----------|------|
| Audibility — `say(text)` produces audio frames | `test_openai_realtime_say_audio_renders` (integration) |
| Server-side isolation — `say(secret, add_to_chat_ctx=False)` followed by `generate_reply` cannot retrieve `secret` | `test_openai_realtime_say_isolation_no_leak` (integration) |
| Substrate state isolation — substrate `chat_ctx` does not contain `secret` after isolated `say()` | `test_openai_realtime_say_isolation_no_remote_chat_ctx_leak` (integration) |
| Wire-format metadata — outbound `response.create` carries `metadata.client_event_id` | `test_openai_realtime_say_emits_metadata` (integration) |
| Local-context gate — `add_to_chat_ctx=False` does NOT populate `agent._chat_ctx`; `add_to_chat_ctx=True` DOES (positive+negative control) | `test_openai_realtime_say_isolation_no_local_leak` (behavioral, network-free) |
| Orphan filter — late `response.created` for a popped future is discarded; `_current_generation` stays None; `response.cancel(response_id=…)` is sent | `test_orphaned_response_after_timeout_filtered` (behavioral, network-free) |
| Dispatcher kwarg flow — `AgentActivity.say(add_to_chat_ctx=False)` actually forwards `False` to `_realtime_reply_task` | `test_agent_activity_say_realtime_dispatches_with_add_to_chat_ctx` (unit, frame-locals inspection) |
| Dispatcher capability warning — calling on a plugin without `ephemeral_say` emits `DeprecationWarning` | `test_agent_activity_say_realtime_capability_warns` (unit) |

### Late-event check

Builds on merged PR #2125's `InvalidStateError` handling at the
future-resolution layer. This PR adds an audio-playout layer check: when
`response.created` arrives carrying a `metadata.client_event_id` that we
previously issued but is no longer in `_response_created_futures` (because the
client-side timeout already popped it), we discard the late arrival and send a
defensive `response.cancel(response_id=...)`.

The check reuses the existing `_response_created_futures` dict — no new
lifecycle state is introduced. A metadata-presence guard ensures
server-VAD-initiated responses (no client-issued metadata) flow through
normally.

Two adjacent event handlers (`_handle_response_output_item_added`,
`_handle_response_content_part_added`) gain a defensive
`if self._current_generation is None: return` guard mirroring the existing
pattern in `_handle_response_done`. This protects the cancel-arrival race
window during which late events for an orphaned response could otherwise hit a
None-assertion crash.

### Empirical verification of the substrate contract

The `conversation: "none"` semantic and `metadata.client_event_id` round-trip
are both verified end-to-end by the integration tests in
`tests/test_realtime/test_realtime_say_integration.py` (6 tests total — 4 against
the real OpenAI Realtime API, requires `OPENAI_API_KEY`; 2 network-free
behavioral tests for the local-context gate and orphan filter that run on every
PR even from forks).

The local-context-gate test uses a positive+negative control structure: it
first verifies that `add_to_chat_ctx=True` populates the agent's chat context
(positive control — would catch a regression where local insertion silently
stops working), then verifies `add_to_chat_ctx=False` does NOT. Without the
positive control, the negative-only assertion could pass even if the entire
local-insertion path was broken.

The dispatcher kwarg-flow test inspects the captured `_realtime_reply_task`
coroutine's frame locals to assert that `AgentActivity.say(add_to_chat_ctx=False)`
actually forwards `False` through the call chain — not just that the realtime
path is dispatched.

### Scope note (deferred work, separately tracked)

This PR is intentionally focused. The following adjacent improvements are
deferred to standalone PRs to keep this review narrow:

- **Dict refactor of `_current_generation`**: see #1988. The slot pattern is
  preserved in this PR. The dominant call path through `AgentActivity.say()`
  serializes via `_speech_q` so concurrent OOB `say()` collisions are not
  exercised. Direct `_rt_session.say()` callers firing OOB responses in tight
  succession may collide; documented as known limitation. I plan to file the
  dict refactor as a follow-up PR motivated by #1988.
- **`update_options` race fix**: see #5530.
- **Capability docstring backfill (5 other undocumented fields)**: see #5565;
  will be a tiny standalone `docs:` PR.
- **`interrupt()` cancel-all-under-concurrency** (latent today, active under
  future dict refactor): see #5564.
- **`response.error` mid-flight future hang** (TODO at
  `realtime_model.py:2005`): see #5566.

### Scope note (`generate_reply()`)

This contribution is scoped to `say()`. A follow-up PR will add
`add_to_chat_ctx` to `generate_reply()` using the same `ephemeral_say` flag
pattern (use case A semantic only).

### CI note

Integration tests skip on fork PRs (no `OPENAI_API_KEY`). I've verified the
test suite locally against `gpt-4o-realtime-preview-2025-06-03` and will paste
the test output as a comment on this PR for inspection.

---

## File diff summary (across all 8 commits)

| File | Change |
|------|--------|
| `livekit-agents/livekit/agents/llm/realtime.py` | new `ephemeral_say` capability field; abstract `RealtimeSession.say()` signature gains `add_to_chat_ctx` kwarg |
| `livekit-agents/livekit/agents/voice/agent_activity.py` | dispatcher capability check (`DeprecationWarning`); plumbing of `add_to_chat_ctx` through 3 internal methods; local chat_ctx upsert gated on `add_to_chat_ctx` |
| `livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py` | OpenAI plugin declares `supports_say=True, ephemeral_say=True`; `say()` body using `response.create(conversation: "none")`; late-event check; defensive guards in two downstream handlers |
| `livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py` | Phonic `say()` accepts `add_to_chat_ctx` kwarg (signature compat; substrate is structurally exempt) |
| `tests/test_realtime_capabilities.py` | unit tests: capability advertisement (OpenAI + Phonic), Phonic signature shim, OpenAI `say()` wire format |
| `tests/test_agent_activity_say.py` | unit tests: dispatcher capability warning, kwarg-flow verification (frame-locals inspection of dispatched coroutine) |
| `tests/test_realtime/test_realtime_say_integration.py` | integration + behavioral tests: 4 against the real OpenAI Realtime API (audibility, server-side isolation, substrate-state isolation, wire-format metadata), plus 3 network-free behavioral tests (local-context gate with positive+negative control; orphan filter discards late `response.created`; metadata-presence guard bypasses the orphan filter for server-VAD-initiated responses) |

## Commit log

```
53b64313 test(realtime): assert orphan filter bypasses on metadata=None (server-VAD)
a9a3e2ac test(realtime): behavioral integration tests for ephemeral say()
cd500116 test(realtime): integration tests for ephemeral say()
579df8f7 feat(realtime): gate local chat_ctx upsert on add_to_chat_ctx
1ff15adf feat(realtime): dispatcher capability check + add_to_chat_ctx plumbing
92418578 feat(realtime/openai): implement say() with conversation: "none" + orphan filter
a97d4283 feat(phonic): accept add_to_chat_ctx kwarg on say() for signature compat
e7947d30 feat(realtime): add ephemeral_say capability flag
```

If reviewers prefer a single squashed commit, the maintainer is welcome to
squash on merge — the per-phase commit chain is for development hygiene and
isolated rollback.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -8,7 +8,7 @@ import json
 import os
 import time
 import weakref
-from collections.abc import Iterator
+from collections.abc import AsyncIterable, Iterator
 from dataclasses import dataclass, replace
 from typing import Any, Literal, overload
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
@@ -1537,6 +1537,99 @@ class RealtimeSession(
         fut.add_done_callback(lambda _: handle.cancel())
         return fut
 
+    def say(
+        self,
+        text: str | AsyncIterable[str],
+        *,
+        add_to_chat_ctx: bool = True,
+    ) -> asyncio.Future[llm.GenerationCreatedEvent]:
+        """Render text to user via OpenAI Realtime API.
+
+        Uses response.create(conversation: "none") for
+        add_to_chat_ctx=False with text wrapped in an assistant
+        message in the input field. The OpenAI Realtime API's
+        documented behavior is that conversation: "none" responses
+        do not enter the server-side conversation state.
+
+        On client-side 10s timeout, the future is rejected and
+        popped from _response_created_futures (existing behavior).
+        The orphan filter in _handle_response_created detects
+        late-arriving response.created by checking
+        missing-from-_response_created_futures.
+        """
+        if isinstance(text, str):
+            return self._say_with_text(text, add_to_chat_ctx=add_to_chat_ctx)
+
+        async def _materialize_and_say() -> llm.GenerationCreatedEvent:
+            full_text = ""
+            async for chunk in text:
+                full_text += chunk
+            return await self._say_with_text(full_text, add_to_chat_ctx=add_to_chat_ctx)
+
+        # Wrap async materialization in a Future to match the abstract base contract
+        outer_fut: asyncio.Future[llm.GenerationCreatedEvent] = asyncio.Future()
+
+        async def _runner() -> None:
+            try:
+                inner = await _materialize_and_say()
+                if not outer_fut.done():
+                    outer_fut.set_result(inner)
+            except BaseException as e:
+                if not outer_fut.done():
+                    outer_fut.set_exception(e)
+
+        asyncio.ensure_future(_runner())
+        return outer_fut
+
+    def _say_with_text(
+        self, full_text: str, *, add_to_chat_ctx: bool
+    ) -> asyncio.Future[llm.GenerationCreatedEvent]:
+        event_id = utils.shortuuid("response_create_say_")
+        fut: asyncio.Future[llm.GenerationCreatedEvent] = asyncio.Future()
+        self._response_created_futures[event_id] = fut
+
+        params = RealtimeResponseCreateParams(
+            input=[
+                realtime.RealtimeConversationItemAssistantMessage(
+                    type="message",
+                    role="assistant",
+                    content=[
+                        realtime.realtime_conversation_item_assistant_message.Content(
+                            type="output_text",
+                            text=full_text,
+                        )
+                    ],
+                )
+            ],
+            metadata={"client_event_id": event_id},
+        )
+        if not add_to_chat_ctx:
+            params.conversation = "none"
+
+        self.send_event(
+            ResponseCreateEvent(type="response.create", event_id=event_id, response=params)
+        )
+
+        def _on_timeout() -> None:
+            self._response_created_futures.pop(event_id, None)
+            if fut and not fut.done():
+                fut.set_exception(llm.RealtimeError("say() timed out."))
+                try:
+                    # response_id intentionally omitted: response.created has not
+                    # arrived yet at timeout, so the server-assigned id is unknown.
+                    # Best-effort cancel. The orphan filter
+                    # (_handle_response_created) sends a second cancel WITH
+                    # response_id when the late response.created arrives, giving
+                    # full coverage. The OpenAI substrate treats duplicate cancels
+                    # as no-ops.
+                    self.send_event(ResponseCancelEvent(type="response.cancel"))
+                except Exception:
+                    logger.debug("response.cancel failed on say() timeout")
+
+        handle = asyncio.get_event_loop().call_later(10.0, _on_timeout)
+        fut.add_done_callback(lambda _: handle.cancel())
+        return fut
+
     @property
     def has_active_generation(self) -> bool:
         return self._current_generation is not None or len(self._response_created_futures) > 0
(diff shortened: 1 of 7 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1535,6 +1537,99 @@ def _on_timeout() -> None:
...
+                )
+            ],
+            metadata={"client_event_id": event_id},
+        )
+        if not add_to_chat_ctx:
+            params.conversation = "none"
+
+        self.send_event(
+            ResponseCreateEvent(type="response.create", event_id=event_id, response=params)
+        )
```
longcw, line 1611: have you tested that if openai realtime will "say" this message instead of using it as context?

longcw: I just tested and it seems it's not supported by openai realtime session.

cphoward: You are right, my sincerest apologies. I wrote an made test validity error when experimenting via OpenAI's SDK. I only confirmed audio was returned, not that the audio was what I expected. Closing this PR.

PR #5583, opened 2026-04-28
feat(openai): expose verbosity in Responses LLM

AlessandroElyos:
Adds a `verbosity` constructor kwarg to the OpenAI Responses LLM, mirroring the chat-completions plugin and matching the pattern of the recently-added `service_tier` (#5341).

Note: in the Responses API `verbosity` lives under `text.verbosity` (alongside `text.format` for structured outputs), so the change merges into the `text` object rather than overwriting it.

### DIFF
--- a/.github/next-release/changeset-openai-responses-verbosity.md
+++ b/.github/next-release/changeset-openai-responses-verbosity.md
@@ -1,5 +1,0 @@
----
-"livekit-plugins-openai": patch
----
-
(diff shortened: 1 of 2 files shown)


tinalenguyen:
hi, thank you for the PR! could you remove the changeset?

AlessandroElyos:
hi @tinalenguyen, done thanks

tinalenguyen (approved):
thank you, lgtm!

PR #5585, opened 2026-04-28
fix(agents): persist _speech_start_time across intra-turn VAD bursts

AlessandroElyos:
Within a single user turn, VAD can fire multiple `START_OF_SPEECH`/`END_OF_SPEECH` cycles separated by short silences (e.g. the user says "Hello." then pauses briefly before continuing). End-of-turn detection is decoupled from VAD — a turn is only considered ended once `_bounce_eou_task` runs and clears the per-turn state.

`_vad_speech_started` was being reset to `False` in the `END_OF_SPEECH` branch of `_on_vad_event`. The next `START_OF_SPEECH` within the same turn would therefore overwrite `_speech_start_time` with the latest burst's start, losing the original turn-start timestamp. This propagates to `started_speaking_at` on the EOT metrics report, which downstream consumers (recording-to-transcript alignment, analytics) rely on to know when the user actually began speaking.

The fix removes the `_vad_speech_started = False` reset from the `END_OF_SPEECH` branch. The flag is now cleared only by the EOT cleanup in `_bounce_eou_task` — which is already the lifecycle owner of `_speech_start_time` — so both fields are governed symmetrically by turn boundaries rather than burst boundaries.

`_vad_speech_started` is read at exactly one site (the SOS guard) and is also defensively reset in the VAD task `finally` block on handoff/teardown, so removing the EOS reset has no other side effects.

Adds `tests/test_speech_start_time_persistence.py` with three cases following the existing `test_audio_recognition_aclose.py` style:
- `test_first_sos_sets_speech_start_time` — sanity check.
- `test_speech_start_time_persists_across_intra_turn_bursts` — the bug reproducer; fails before the fix, passes after.
- `test_eos_does_not_clear_vad_speech_started` — encodes the new invariant.

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -1084,10 +1084,6 @@ class AudioRecognition:
             _bounce_eou_task(
                 self._last_speaking_time,
                 self._last_final_transcript_time,
-                # Use the turn-level start (set when the user_turn_span was
-                # created on first SOS) so `started_speaking_at` on the EOT
-                # metrics reflects when the user *first* started speaking
-                # within this turn — not the latest VAD burst.
                 self._user_turn_start,
             )
         )
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -912,6 +912,7 @@ class AudioRecognition:
             with trace.use_span(self._ensure_user_turn_span()):
                 self._hooks.on_end_of_speech(ev)
 
+            self._vad_speech_started = False
             self._speaking = False
 
             if self._vad_base_turn_detection or (
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -911,7 +911,6 @@ class AudioRecognition:
             with trace.use_span(self._ensure_user_turn_span()):
                 self._hooks.on_end_of_speech(ev)
 
-            self._vad_speech_started = False
             self._speaking = False
 
             if self._vad_base_turn_detection or (
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -185,6 +185,7 @@ class AudioRecognition:
         self._agent_speaking: bool = False
 
         self._user_turn_span: trace.Span | None = None
+        self._user_turn_start: float | None = None
         self._stt_request_ids: list[str] = []
         self._closing = asyncio.Event()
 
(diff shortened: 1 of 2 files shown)


AlessandroElyos:
The failing `test_interruption_before_speaking[True-3.5]` looks like a pre-existing flake unrelated to this PR — it's a timing-sensitive assertion (5× speed factor, 0.75s tolerance on `check_timestamp`). I ran it 3× locally with the fix applied and it passed each time. Happy to fix it if it can help

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -911,7 +911,6 @@ async def _on_vad_event(self, ev: vad.VADEvent) -> None:
             with trace.use_span(self._ensure_user_turn_span()):
                 self._hooks.on_end_of_speech(ev)
 
-            self._vad_speech_started = False
```
longcw, line 914: this makes the `_speech_start_time` always the turn start time, may break the tracing that we create `user_speaking` spans using this timestamp as the start of the span ([ref](https://github.com/livekit/agents/blob/613a119aae9cfb4992e2f9a6e62052623e9de937/livekit-agents/livekit/agents/voice/agent_activity.py#L1658)).

using the start time of the `user_turn_span` could be a better solution if the goal is to make `started_speaking_at` in the EOT info  the first time user started speaking. we already have `_ensure_user_turn_span` that create the span only when there is no active user turn span, and the span is ended only after EOT committed.

the OTEL span is designed for write-only, so maybe add a `_user_turn_start` alongside the user_turn_span.

AlessandroElyos: thanks for the feedback @longcw.

I reverted the change on `_vad_speech_started`.

I also added a `_user_turn_start` variable as you suggested that should reflect the actual turn boundaries.

Let me know your thoughts.

AlessandroElyos: @longcw did you get the chance to look at the new version?

longcw (approved):

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1170,6 +1172,9 @@ def _ensure_user_turn_span(self, start_time: float | None = None) -> trace.Span:
         start_time_ns = int(start_time * 1_000_000_000) if start_time else None
         self._user_turn_span = tracer.start_span("user_turn", start_time=start_time_ns)
 
+        if start_time is not None and self._user_turn_start is None:
+            self._user_turn_start = start_time
```
chenghao-mou, line 1176: In `start_span`, the `start_time` parameter has a default value of current time in nano seconds if None is provided. So I think it is safe to assign `start_time or time.time()` here to be consistent (or you can override `start_time` to `time.time()` first before `start_time_ns` is calculated).

chenghao-mou: I'll create a new PR.

PR #5586, opened 2026-04-28
fix(mistral): use conversations API statelessly

TheCodingCvrlo:
## Summary

Makes the Mistral LLM plugin stateless — every call now uses `start_stream_async()` with the full chat context instead of tracking `conversation_id` and appending deltas via `append_stream_async()`.

**Usage is unchanged.** All public APIs, constructor parameters, and provider tool support remain identical.

## Problem

The Mistral [Conversations API](https://docs.mistral.ai/api/endpoint/beta/conversations) locks a conversation while a request is in flight (enforced server-side with a 90-second lock window). When a user interrupts the agent mid-speech, hence trying to perform a concurrent update:

1. LiveKit cancels the in-flight LLM task, disconnecting the HTTP stream
2. The server-side lock on that `conversation_id` persists (up to 90s)
3. New user speech triggers a new LLM call with `append_stream_async()` on the same `conversation_id`
4. The API returns **HTTP 409** — *"Conversation is being updated, concurrent requests are not supported"*
5. All retries fail because the retry budget (~4s) is far shorter than the 90s lock

This made interruption reliably break multi-turn conversations.

### Stateful tracking removed

The previous implementation maintained state across turns on the `LLM` instance:

| State | Purpose | Removed |
|---|---|---|
| `_conversation_id` | Reuse server-side conversation across calls | `LLM.__init__`, `LLMStream._run` |
| `_prev_chat_ctx` | Diff new context against previous to send only deltas | `LLM.__init__`, `LLM.chat` |
| `_pending_tool_calls` | Verify all function calls have results before appending | `LLM.__init__`, `LLM._pending_tool_calls_completed` |
| `_received_conversation_id` | Capture `conversation_id` from `ResponseStartedEvent` | `LLMStream.__init__`, `LLMStream._parse_event` |
| `full_chat_ctx` | Snapshot full context for post-response state update | `LLMStream.__init__` parameter |

On `update_options(model=...)`, all state was also reset — this is no longer needed.

### `LLM.chat()` simplification

Previously, `chat()` performed context diffing to decide between `start` and `append`:

- Compare `chat_ctx.items[:n]` against `_prev_chat_ctx` using `ChatContext.is_equivalent()`
- If the context extends the previous one and all pending tool calls have results → `append_stream_async()` with only new entries
- Otherwise → fall back to `start_stream_async()` with full context

Now it always passes the full `chat_ctx` directly to `LLMStream`.

### `LLMStream._run()` simplification

The `start` vs `append` branch is removed. Every call uses:

```python
await self._client.beta.conversations.start_stream_async(
    inputs=entries,
    model=self._model,
    instructions=extra_data.instructions,
    **start_kwargs,
    **self._extra_kwargs,
)
```

The post-stream state writeback (`_conversation_id`, `_prev_chat_ctx`, `_pending_tool_calls`) is also removed.

### What is unchanged

- Provider tool support (`WebSearch`, `CodeInterpreter`, `DocumentLibrary`)
- Streaming event handling (`FunctionCallEvent`, `ToolExecution*Event`, etc.)
- All `CompletionArgs` parameters (`temperature`, `top_p`, `tool_choice`, etc.)
- Provider format converter (`to_conversations_ctx()`)
- `LLM` constructor and `update_options()` signatures

### DIFF
--- a/livekit-plugins/livekit-plugins-mistralai/livekit/plugins/mistralai/llm.py
+++ b/livekit-plugins/livekit-plugins-mistralai/livekit/plugins/mistralai/llm.py
@@ -8,7 +8,6 @@ from livekit.agents import APIConnectionError, APIStatusError, APITimeoutError, 
 from livekit.agents.llm import (
     ChatChunk,
     ChatContext,
-    ChatItem,
     ToolChoice,
 )
 from livekit.agents.types import (
--- a/livekit-plugins/livekit-plugins-mistralai/livekit/plugins/mistralai/llm.py
+++ b/livekit-plugins/livekit-plugins-mistralai/livekit/plugins/mistralai/llm.py
@@ -113,9 +113,6 @@ class LLM(llm.LLM):
         if not client and not mistral_api_key:
             raise ValueError("Mistral AI API key is required. Set MISTRAL_API_KEY or pass api_key")
         self._client = client or Mistral(api_key=mistral_api_key)
-        self._conversation_id: str | None = None
-        self._prev_chat_ctx: ChatContext | None = None
-        self._pending_tool_calls: set[str] = set()
 
     @property
     def model(self) -> str:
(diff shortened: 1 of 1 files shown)


tinalenguyen (approved):
lgtm, thanks for the fix!

PR #5587, opened 2026-04-28
Add diagnostics core and plugin hooks

shizhigu:
## Summary
- Add structured diagnostics models, JSON serialization, exit-code handling, and Rich table rendering.
- Add side-effect-free plugin hooks for static diagnostic metadata and optional deep checks.
- Cover strict mode, JSON stability, and plugin diagnostic failure handling.

This does not add a CLI command yet. It only introduces the reusable diagnostics core and the backwards-compatible plugin extension points that later changes can build on.

## Tests
- `uv run pytest tests/test_plugin_diagnostics.py -q`
- `make check`

PR #5588, opened 2026-04-28
Add doctor CLI and startup preflight

shizhigu:
## Summary
- Add `python agent.py doctor` with table and JSON output.
- Support `--online`, `--deep`, and `--strict` without creating rooms, calling providers, downloading files, or mutating local state by default.
- Reuse the diagnostics core for startup preflight in `console`, `dev`, `start`, and `connect`; only fatal results block startup.
- Apply CLI LiveKit URL/key/secret options to the server before diagnostics and startup.

Stacked on #5587. This is a draft until #5587 lands; GitHub shows the earlier diagnostics-core diff because upstream PRs cannot use a fork branch as their base.

## Tests
- `uv run pytest tests/test_cli_doctor.py tests/test_cli_log_level.py tests/test_drain_timeout.py -q`
- `make check`

### DIFF
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -39,6 +39,14 @@ from livekit import api, rtc
 
 from .. import llm
 from .._exceptions import CLIError
+from ..diagnostics import (
+    DiagnosticContext,
+    DiagnosticSeverity,
+    build_diagnostics_table,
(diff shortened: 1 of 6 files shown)


shizhigu:
Closing this draft split PR in favor of the combined first-run diagnostics PR. Keeping the CLI and preflight changes with the diagnostics core should make the behavior easier to review end to end.

PR #5591, opened 2026-04-28
Fix the memory leak of Google STT when no audio input.

### LINKED ISSUE #5590: Google STT has a memory leak when an agent session is running with no audio input from the user
### Bug Description

I am seeing a memory leak in the agent; the memory consumption increases steadily over the time with idle status (no input, no output, just stays open the webRTC session).  During the debugging, I found that this memory leak is gone after I disabled STT component from the agent session setup, I replaced the STT component to deepgram, and also I enabled the audio input.

### Expected Behavior

Memory consumption isn't increasing during the idle status.

### Reproduction Steps

```bash
1. start an agent with Google STT locally
2. start an RTC session, set up a connection with the agent from ai playground
3. do not enable audio/video, stays the connection open
4. figure out the PID of the agent handling the RTC session
5. run `watch /proc/<PID>/status` and monitor the memory consumption
```

### Operating System

Linux

### Models Used

Google chirp v3

### Package Versions

```bash
livekit==1.5.4
livekit-agents=1.5.4
livekit-plugins-google=1.5.4
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
It seems stt.py isn't cleaning the input stream frames when timeout (which will happen when there are no audio input).  It'll re-create the Google API client object and re-connect again, but the old client object with related objects remain with the input frames.  That would be the cause of the memory leak.

So fixing the handling of the input stream would be the solution.  I created a patch with helps from Claude Code and confirmed it removed the memory leak on my local environment.
```

### Additional Context

_No response_

Debugged with tracemalloc; taking snapshots every 5 minutes and making the comparison.  The top 10 list are like:
```
/lib/python3.11/functools.py:58                                          0.28 MB
google/api_core/gapic_v1/method_async.py:52                              0.17 MB
google/api_core/grpc_helpers_async.py:164                                0.13 MB
<frozen abc>:123                                                          0.12 MB
grpc/aio/_channel.py:496                                                  0.09 MB
google/api_core/client_info.py:108                                        0.07 MB
grpc/_common.py:71                                                        0.06 MB
pydantic/main.py:253                                                      0.06 MB
/lib/python3.11/threading.py:265                                          0.06 MB
/lib/python3.11/json/decoder.py:353                                       0.05 MB
```

That indicates that client objects / API handling / connection remains.

jmuk:
When the RTC session is idle without any audio input, Google STT re-create the client object and re-connect periodically. However some old resources actually remain as they're not cancelled properly, causing memory leak.

This PR adds the clean-up of such scenarios.  With this change, I confirmed my experience of the memory leak is gone.

Fixes #5590

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
@@ -15,13 +15,16 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import dataclasses
 import time
 import weakref
-from collections.abc import AsyncGenerator, AsyncIterable, Callable
+from collections.abc import AsyncGenerator, AsyncIterable, Callable, Coroutine
 from dataclasses import dataclass
 from datetime import timedelta
-from typing import cast, get_args
+from typing import Any, cast, get_args
+
+from grpc.aio import StreamStreamCall
 
 import google.auth
 from google.api_core.client_options import ClientOptions
@@ -748,14 +751,28 @@ class SpeechStream(stt.SpeechStream):
             None,
         ]:
             nonlocal audio_pushed
+            stop_task = asyncio.create_task(should_stop.wait())
             try:
                 yield self._build_init_request(client)
 
-                async for frame in self._input_ch:
-                    # when the stream is aborted due to reconnect, this input_generator
-                    # needs to stop consuming frames
-                    # when the generator stops, the previous gRPC stream will close
-                    if should_stop.is_set():
+                input_iter = aiter(self._input_ch)
+                while True:
+                    # Race the next-frame await against should_stop so this generator
+                    # can exit even when no audio is flowing. Without this, on reconnect
+                    # the generator stays parked on _input_ch and pins the previous
+                    # gRPC streaming call, leaking it across iterations.
+                    frame_task = asyncio.create_task(
+                        cast(Coroutine[rtc.AudioFrame, Any, Any], anext(input_iter))
+                    )
+                    done, _ = await asyncio.wait(
+                        [frame_task, stop_task], return_when=asyncio.FIRST_COMPLETED
+                    )
+                    if stop_task in done:
+                        frame_task.cancel()
+                        return
+                    try:
+                        frame = frame_task.result()
+                    except StopAsyncIteration:
                         return
 
                     if isinstance(frame, rtc.AudioFrame):
@@ -765,6 +782,8 @@ class SpeechStream(stt.SpeechStream):
 
             except Exception:
                 logger.exception("an error occurred while streaming input to google STT")
+            finally:
+                stop_task.cancel()
 
         async def process_stream(
             client: SpeechAsyncClientV2 | SpeechAsyncClientV1,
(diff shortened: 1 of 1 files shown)


zelidrag-arbo (approved):

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -748,14 +751,28 @@ async def input_generator(
...
-                    if should_stop.is_set():
+                input_iter = aiter(self._input_ch)
+                while True:
+                    # Race the next-frame await against should_stop so this generator
+                    # can exit even when no audio is flowing. Without this, on reconnect
+                    # the generator stays parked on _input_ch and pins the previous
+                    # gRPC streaming call, leaking it across iterations.
+                    frame_task = asyncio.create_task(
+                        cast(Coroutine[rtc.AudioFrame, Any, Any], anext(input_iter))
+                    )
```
longcw: nit
```suggestion
                    frame_task = asyncio.create_task(
                        self._input_ch.recv()
                    )
```

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -748,14 +751,28 @@ async def input_generator(
...
+                    done, _ = await asyncio.wait(
+                        [frame_task, stop_task], return_when=asyncio.FIRST_COMPLETED
+                    )
+                    if stop_task in done:
+                        frame_task.cancel()
+                        return
+                    try:
+                        frame = frame_task.result()
+                    except StopAsyncIteration:
```
longcw: use `ChanClosed` if using the `input_ch.recv` on above

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -896,6 +921,10 @@ async def process_stream(
                 if e.code == 409:
                     if audio_pushed:
                         logger.debug("stream timed out, restarting.")
+                    else:
+                        # No audio has been pushed — back off so we don't tight-loop
+                        # against Google's server-side 5-minute idle timeout.
+                        await asyncio.sleep(5.0)
```
longcw: should we await sleep here? what if user audio comes in during the sleep?

longcw (commented):
thanks for the pr! overall looks good to me, my only concern is about the `await asyncio.sleep(5.0)` after timeout

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
jmuk: Indeed, didn't think much about such cases; removed.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
jmuk: Done

jmuk (commented):
Thanks for the reviews!

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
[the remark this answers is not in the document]
jmuk: this is an intended change

longcw (approved):

PR #5593, opened 2026-04-29
docs(examples), test: fix first-run docs and timing flake

shizhigu:
## Summary

This keeps the examples first-run docs accurate and folds in the small AgentSession timing-test stabilization that was already opened as #5597.

The docs part fixes stale example paths and expands `examples/.env.example` so a new user can see the common LiveKit/provider variables without hunting through every example. The test part fixes the repeated CI failure this PR hit twice in `tests/test_agent_session.py::test_tool_call`: wall-clock event emission can be a little late under CI scheduler load, but duration/playback-position checks stay strict.

I folded #5597 into this PR so reviewers do not have to reason about a docs PR blocked by a separate test-only PR.

## What changed

- Expanded `examples/.env.example` with commented, opt-in variables for common example categories.
- Fixed stale README paths in avatar and telephony examples.
- Removed voice-agent README entries for files that are no longer present.
- Updated the examples README Python version from 3.9 to 3.10.
- Split deterministic timestamp checks from wall-clock event timestamp checks in `tests/test_agent_session.py`.
- Kept early wall-clock assertions explicit while allowing a small extra late margin for event emission under CI load.

## Testing

- `make check`
- `PYTHONPATH="$PWD" uv run ruff check tests/test_agent_session.py`
- `PYTHONPATH="$PWD" uv run ruff format --check tests/test_agent_session.py`
- `PYTHONPATH="$PWD" uv run pytest tests/test_agent_session.py::test_generate_reply`
- Earlier targeted run on this branch: `PYTHONPATH="$PWD" uv run pytest tests/test_agent_session.py::test_tool_call tests/test_agent_session.py::test_interruption_before_speaking tests/test_agent_session.py::test_generate_reply tests/test_agent_session.py::test_preemptive_generation -q`
  - `7 passed`
- Earlier full file run on this branch: `PYTHONPATH="$PWD" uv run pytest tests/test_agent_session.py -q`
  - `20 passed`
- `git diff --check`

### DIFF
--- a/examples/.env.example
+++ b/examples/.env.example
@@ -1,5 +1,5 @@
-# Required for most examples. Everything below this first block is optional and
-# only needed by specific examples.
+# LiveKit connection. Required for examples that connect to a LiveKit room.
+# Everything below this first block is optional and only needed by specific examples.
 LIVEKIT_URL="wss://your-project.livekit.cloud"
 LIVEKIT_API_KEY="your_livekit_api_key"
 LIVEKIT_API_SECRET="your_livekit_api_secret"
--- a/examples/.env.example
+++ b/examples/.env.example
@@ -1,3 +1,90 @@
-LIVEKIT_API_SECRET="<your livekit api secret>"
-LIVEKIT_API_KEY="<your livekit api key>"
-LIVEKIT_URL="<your livekit ws url>"
+# Required for most examples. Everything below this first block is optional and
+# only needed by specific examples.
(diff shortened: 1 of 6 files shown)


shizhigu:
The current unit-tests failure is unrelated to this docs-only diff. It is the existing `tests/test_agent_session.py::test_tool_call` timing flake that also showed up on prior runs. I opened #5597 with a focused test fix so this PR does not need to carry unrelated test changes.

PR #5594, opened 2026-04-29
fix: clear stale paused speech state across generation steps

longcw:
## Summary

Alternative fix for the bug described in #5533.

When a `SpeechHandle` advances through multiple generation steps within a single turn (e.g. a silent tool-call step followed by a tool-reply step), the false-interruption `_paused_speech` state captured during an earlier step could leak into the next step. 

This fix resets `_paused_speech` — together with the matching false-interruption timer and the paused audio output — at the scheduling-loop boundary in `_scheduling_task`, right after each generation step finishes. The next step that runs on the same `SpeechHandle` then records fresh paused state with the correct `agent_state`.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1362,6 +1362,15 @@ class AgentActivity(RecognitionHooks):
                         continue
                 speech._authorize_generation()
                 await speech._wait_for_generation()
+
+                if self._paused_speech and self._paused_speech.handle is self._current_speech:
+                    # clear paused speech after generation done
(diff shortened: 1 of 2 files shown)


jayeshp19 (approved):

davidzhao (approved):

toubatbrian:

This PR (`fix: clear stale paused speech state across generation steps`) was identified as a **core runtime fix** to `_scheduling_task` in `agent_activity.py`. The base `resume_false_interruption` feature has already been ported to agents-js (see livekit/agents-js#1320), so the leak fix can be applied there too. A follow-up PR against `livekit/agents-js` will be opened shortly.

---

PR #5596, opened 2026-04-29
docs(examples): expand env template

shizhigu:
## Summary
- Expand `examples/.env.example` beyond the three LiveKit credentials.
- Add commented placeholders for provider, avatar, tracing, SIP, and primitive-example variables used by examples.
- Clarify that LiveKit Inference examples only need LiveKit Cloud credentials, while direct provider plugins need provider keys.

## Tests
- `uv run python - <<'PY' ... dotenv_values('examples/.env.example') ... PY`
- `make check`
- `git diff --check`
- Claude reviewed the docs-only diff and found no incorrect env var names.

### DIFF
--- a/examples/.env.example
+++ b/examples/.env.example
@@ -1,3 +1,90 @@
-LIVEKIT_API_SECRET="<your livekit api secret>"
-LIVEKIT_API_KEY="<your livekit api key>"
-LIVEKIT_URL="<your livekit ws url>"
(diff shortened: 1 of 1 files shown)


shizhigu:
Closing this standalone env-template PR because the change is now folded into #5593. Keeping the examples first-run docs fixes together should be easier to review than two tiny docs PRs.

PR #5597, opened 2026-04-29
test(agents): stabilize agent session timing assertions

shizhigu:
## Summary
- Split wall-clock event timestamp checks from deterministic duration and playback-position checks.
- Allow a small extra late margin for `created_at` event emissions under CI scheduler load while keeping early events strict.
- Preserve the wider early tolerance that `test_generate_reply` already used for its late conversation events.

## Tests
- `uv run ruff check tests/test_agent_session.py`
- `uv run ruff format --check tests/test_agent_session.py`
- `uv run pytest tests/test_agent_session.py::test_tool_call tests/test_agent_session.py::test_interruption_before_speaking tests/test_agent_session.py::test_generate_reply tests/test_agent_session.py::test_preemptive_generation -q`
- `uv run pytest tests/test_agent_session.py -q`
- `git diff --check`

Claude reviewed the diff after the tolerance split and found no blocking issues.

### DIFF
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -68,6 +68,8 @@ class MyAgent(Agent):
 
 
 SESSION_TIMEOUT = 60.0
+EVENT_TIMESTAMP_EARLY_TOLERANCE = 0.75
+EVENT_TIMESTAMP_LATE_TOLERANCE = 1.25
 
 
 async def test_events_and_metrics() -> None:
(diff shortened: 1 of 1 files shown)


shizhigu:
Closing this separate test-only PR because the same stabilization commit is now folded into #5593, which was repeatedly hitting the same AgentSession timing assertion in CI. Keeping the docs fixes and the CI-unblocker together avoids making reviewers coordinate two dependent PRs.

PR #5608, opened 2026-04-30
feat(voice): scope http_context to AgentSession when no job context

longcw:
## Summary

When AgentSession runs **outside** a job process (tests, scripts, ad-hoc usage) there is no process-level http session bound to the event loop, so `utils.http_context.http_session()` raises and STT/TTS that rely on it can't connect. AgentSession now creates the http session ctx in `start()` (when `get_job_context()` is None and no outer ctx is already set) and tears it down in `aclose()`, so the same plugins that work inside a job process also work standalone with an AgentSession.

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -588,6 +588,7 @@ class SpeechStream(stt.SpeechStream):
     async def _run(self) -> None:
         """Main loop for streaming transcription."""
         closing_ws = False
+        http_session = self._stt._ensure_session()
 
         @utils.log_exceptions(logger=logger)
         async def send_task(ws: aiohttp.ClientWebSocketResponse) -> None:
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -632,7 +632,7 @@ async def recv_task(ws: aiohttp.ClientWebSocketResponse) -> None:
                     aiohttp.WSMsgType.CLOSE,
                     aiohttp.WSMsgType.CLOSING,
                 ):
-                    if closing_ws or self._session.closed:
```
chenghao-mou, line 635: I believe we have the same pattern in other files as well.

longcw: added this back with a local http_session var

chenghao-mou: Should we fix all "or self._session.closed:" usage?

longcw:
in favor of https://github.com/livekit/agents/pull/5709

PR #5612, opened 2026-04-30
fix: end Runway realtime sessions on shutdown

robinandeer:
End Runway realtime sessions through the normal LiveKit data-channel lifecycle when the LiveKit job shuts down, so conversations complete cleanly instead of being marked cancelled.

Verify:
- `uv run ruff format livekit-plugins/livekit-plugins-runway/livekit/plugins/runway/avatar.py`
- `uv run ruff check livekit-plugins/livekit-plugins-runway/livekit/plugins/runway/avatar.py`
- `make type-check`


tinalenguyen (approved):
thank you!

PR #5613, opened 2026-04-30
(examples revamp): add example walkthroughs

tinalenguyen:


examples/frontdesk/README.md
```diff
@@ -3,3 +3,40 @@
...
+```bash
+ST_abc123 - Saturday, January 1, 2000 at 14:00 PDT (in 5 days)
+```
+
+The slots are also cached as a lookup table for `schedule_appointment`. 
+
+https://github.com/livekit/agents/blob/8283a5a5c9863a07bcf030ee90e8ab780e1e569b/examples/frontdesk/frontdesk_agent.py#L184
+
+
+This prevents the LLM from passing a hallucinated answer.
```
theomonnom: Can we add details on how we're using ToolError to prevent the LLM for hallucinating? Or at least allow self-corrections.

examples/survey/README.md
```diff
@@ -1,5 +1,58 @@
...
+
+After the candidate answers one of the questions, `self._check_completion()` is called to check if all 3 fields (`”strengths”`, `“weaknesses”`, `“work_style”`) have been collected. If so, then `BehavioralTask` is marked as complete. If not, then the agent will continue prompting for the rest of the answers.
+
+In practice, this would ensure variability among candidates’ experiences.
+
+### Closing out
+Once the interview is concluded and TaskGroup is completed, we extract the summary message (the last inserted message):
+
+```python
+summary = self.chat_ctx.items[-1]
```
theomonnom, line 38: btw not related to the readme, but I'm wondering if we should cleanup this pattern.

theomonnom (approved):

examples/survey/README.md
tinalenguyen: yeah.. i feel like working with `chat_ctx` is not intuitive

PR #5629, opened 2026-05-02
feat(fishaudio): use websocket API for faster inference

davidzhao:
reducing TTFB from ~400ms to ~230ms

### DIFF
--- a/livekit-plugins/livekit-plugins-fishaudio/livekit/plugins/fishaudio/tts.py
+++ b/livekit-plugins/livekit-plugins-fishaudio/livekit/plugins/fishaudio/tts.py
@@ -86,9 +86,7 @@ class TTS(tts.TTS):
             voice_id (NotGivenOr[str]): Voice model ID. Fish Audio's API refers to this
                 as ``reference_id``; it's the same value either way.
             output_format (OutputFormat): Audio output format. Defaults to ``"wav"``.
-            sample_rate (int): Audio sample rate in Hz. Defaults to 48000 for ``"opus"``
-                (the only rate Fish supports for opus) and 24000 for ``"pcm"``/``"wav"``/
-                ``"mp3"``. Passing a non-48000 rate with ``"opus"`` raises ``ValueError``.
+            sample_rate (int): Audio sample rate in Hz.
             base_url (NotGivenOr[str]): Custom base URL. Defaults to ``https://api.fish.audio``.
             latency_mode (LatencyMode): Streaming latency mode. ``"normal"``, ``"balanced"``,
                 or ``"low"``. Defaults to ``"balanced"``.
--- a/livekit-plugins/livekit-plugins-fishaudio/livekit/plugins/fishaudio/tts.py
+++ b/livekit-plugins/livekit-plugins-fishaudio/livekit/plugins/fishaudio/tts.py
@@ -25,7 +25,7 @@ from .log import logger
 from .models import LatencyMode, OutputFormat, TTSModels
 from .version import __version__
 
-DEFAULT_MODEL: TTSModels = "s1"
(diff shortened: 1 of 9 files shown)


livekit-plugins/livekit-plugins-fishaudio/livekit/plugins/fishaudio/tts.py
```diff
@@ -3,426 +3,430 @@
...
         self,
         *,
         api_key: NotGivenOr[str] = NOT_GIVEN,
-        model: Backends = DEFAULT_MODEL,
-        reference_id: NotGivenOr[str] = DEFAULT_REFERENCE_ID,
-        output_format: OutputFormat = "pcm",
-        sample_rate: int = 24000,
-        num_channels: int = 1,
+        model: TTSModels | str = DEFAULT_MODEL,
+        voice_id: NotGivenOr[str] = DEFAULT_VOICE_ID,
```
tinalenguyen, line 68: wouldn't the rename from `reference_id` to `voice_id` be a breaking change? and also removing `num_channels`

davidzhao: yes.. but this isn't widely used.. the current version is pretty limited.

livekit-plugins/livekit-plugins-fishaudio/livekit/plugins/fishaudio/tts.py
```diff
@@ -3,426 +3,430 @@
...
 )
 from livekit.agents.types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, NotGivenOr
 from livekit.agents.utils import is_given
 
 from .log import logger
-from .models import LatencyMode, OutputFormat
+from .models import LatencyMode, OutputFormat, TTSModels
+from .version import __version__
+
+DEFAULT_MODEL: TTSModels = "s1"
```
Kilerd: I think we can use s2-pro as default model which has better performance

theomonnom (approved):

toubatbrian:

Note: this PR contains a mix of plugin-specific (fishaudio), Python-specific (replacing the `fish-audio-sdk` dependency), and core runtime changes. Since agents-js does not currently have a fishaudio plugin and the dependency swap is Python-specific, only the core runtime improvements will be ported:

- `tts/tts.py`: don't tear the audio decoder down on a mid-stream `FlushSegment` (preserves stateful codec parser state for WAV/OGG/MP3).
- `utils/codecs/decoder.py`: gracefully handle EOF before any audio could be decoded.

---

toubatbrian:
Update from the routine: after inspecting agents-js, none of the changes in this PR have a direct port path, so no JS PR will be opened.

- **fishaudio plugin rewrite (HTTP+msgpack websocket)** — agents-js does not have a fishaudio plugin, so there is nothing to update.
- **`fish-audio-sdk` → `msgpack` dependency swap** — Python-specific, skipped per the routine's rules.
- **`tts/tts.py` mid-stream `FlushSegment` decoder fix** and **`utils/codecs/decoder.py` `EOFError` handling** — these patch the Python `AudioEmitter._decode_task` path which wraps a stateful FFmpeg-based streaming decoder. agents-js has no equivalent: every TTS plugin (cartesia, elevenlabs, deepgram, baseten, inworld, etc.) emits raw PCM directly via `AudioByteStream`, and FFmpeg is only used in `agents/src/audio.ts` for one-shot file playback and in `recorder_io.ts` for output encoding — neither has a `flush()`-driven decoder-teardown path or the EOF-before-any-output failure mode that the Python fixes target.

cc @toubatbrian

---

PR #5632, opened 2026-05-02
chore(tests): update tests to use Inference whenever possible AGT-2304

chenghao-mou:
Seems like only one spot is missed. Other plugin-focused tests are unchanged.

### DIFF
--- a/tests/test_audio_decoder.py
+++ b/tests/test_audio_decoder.py
@@ -8,9 +8,9 @@ from concurrent.futures import ThreadPoolExecutor
 import aiohttp
 import pytest
 
+from livekit.agents import inference
 from livekit.agents.stt import SpeechEventType
 from livekit.agents.utils.codecs import AudioStreamDecoder, StreamBuffer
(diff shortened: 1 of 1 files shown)


tests/test_audio_decoder.py
```diff
@@ -19,8 +19,8 @@
 
 @pytest.mark.asyncio
 @pytest.mark.skipif(
-    os.getenv("DEEPGRAM_API_KEY") is None,
-    reason="DEEPGRAM_API_KEY not set",
+    os.getenv("LIVEKIT_API_KEY") is None or os.getenv("LIVEKIT_API_SECRET") is None,
```
davidzhao: we also need to check [is_cloud](https://github.com/livekit/agents/blob/204e70a11b4775ad06b10e6c6cc3a7afd0a4fb17/livekit-agents/livekit/agents/utils/misc.py#L40)

davidzhao (approved):

theomonnom (approved):

PR #5634, opened 2026-05-03
feat(cli): add --reload-dir to dev for extra watch directories

korya:
## Problem

`dev`-mode auto-reload only watches the entrypoint, `livekit.agents`, and editable plugins. Edits to *other* packages the agent imports (shared utility code mounted editable in monorepos) are silently ignored — the worker keeps holding stale module objects with no warning.

## Solution

Add repeatable `--reload-dir` (and `LIVEKIT_RELOAD_DIRS`, `os.pathsep`-split) to `dev`. Resolved paths are unioned with the auto-discovered list and passed to the existing `watchfiles.arun_process`. Auto-discovery is unchanged; omitting the flag preserves identical behavior.

```mermaid
flowchart LR
    F["--reload-dir / LIVEKIT_RELOAD_DIRS"] --> CA[CliArgs.reload_dirs]
    AD[_find_watchable_paths<br/>main + livekit.agents + plugins] --> U[union]
    CA --> U
    U --> AP["watchfiles.arun_process(*paths,<br/>watch_filter=PythonFilter())"]
```

Naming mirrors uvicorn's `--reload-dir` so muscle memory transfers.

### Notes worth flagging

- Missing path fails fast (exit 1, clear error). Silent skip would re-create the gap this PR closes.
- `PythonFilter` is retained, so `.git/`, `.venv/`, `__pycache__/`, `node_modules/`, and non-`.py` files don't trigger reloads — even inside user-supplied dirs.
- The flag does **not** split on commas (`A,B` → single path). Only the env var splits on `os.pathsep`. Help text calls this out.
- No allowlist gating (`/`, `~`, paths above `cwd`): hard-rejecting outside `cwd` would break the very monorepo workflow this is for, and `PythonFilter` already drops the dangerous dirs.

## Test plan

`tests/test_reload_dir.py` — 16 scenarios:

- [x] Defaults, repeated flag, env-var split, flag-overrides-env
- [x] Relative + `~` paths, symlink resolution
- [x] Missing path (flag + env) → exit 1
- [x] `CliArgs` pickle round-trip
- [x] iTerm2 carve-out preserved (with and without `--reload-dir`)
- [x] `WatchServer.run()` unions auto + user paths, keeps `PythonFilter`
- [x] `PythonFilter` drops `.git/HEAD`, `.venv/*.py`, etc.
- [x] `tests/test_cli_log_level.py` (25 cases) still passes

Manual: `dev --help` shows the flag + env-var; `dev --reload-dir /tmp/extra` logs `Watching /tmp/extra (user)`; bad path → exit 1.

Changeset: `.github/next-release/changeset-reload-dir.md` (livekit-agents minor).

### DIFF
--- a/.github/next-release/changeset-reload-dir.md
+++ b/.github/next-release/changeset-reload-dir.md
@@ -1,0 +1,8 @@
+---
+"livekit-agents": minor
+---
+
+Add `--reload-dir` (repeatable) and `LIVEKIT_RELOAD_DIRS` to `dev` so the watcher
+restarts the worker on edits inside additional, user-supplied directories.
+Auto-discovery of the entrypoint, `livekit.agents`, and editable plugins is
+unchanged; the new flag is purely additive.
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -1816,7 +1816,32 @@ def _build_cli(server: AgentServer) -> typer.Typer:
                 envvar="LIVEKIT_API_SECRET",
             ),
         ] = None,
+        reload_dir: Annotated[
+            list[pathlib.Path] | None,  # noqa: UP007
+            typer.Option(
+                "--reload-dir",
+                envvar="LIVEKIT_RELOAD_DIRS",
+                help=(
+                    "Additional directory to watch for .py changes (repeatable). "
+                    "Watched recursively; avoid huge trees. "
+                    "Env var LIVEKIT_RELOAD_DIRS is split on os.pathsep."
+                ),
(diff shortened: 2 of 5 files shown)


theomonnom:
Hey, sorry for the late reply, this is great.

For 1.7.0, we're going to move the CLI on the Go side, I'll take a look at adding a reload-dir there as well

korya:
> Hey, sorry for the late reply, this is great.
> 
> For 1.7.0, we're going to move the CLI on the Go side, I'll take a look at adding a reload-dir there as well

No worries. Thanks for the update!

PR #5637, opened 2026-05-04
fix(amd): reset timer for late stt transcript

chenghao-mou:
Reset the short-speech human path timer when we receive any transcript inside the human silence window, changing it to the long speech path for LLM verification. Previously, transcript must be available before the end of speech to be triggered.

Benchmarked the changes against our eval dataset and no sign of regression.


theomonnom (approved):

toubatbrian:
> [!NOTE]
> This is an automated Claude Code Routine created by @toubatbrian. Right now it is in experimentation stage.

This PR looks like a core runtime AMD fix (silence-timer state machine for late STT transcripts). The automation will start porting it into `agents-js` automatically. The port will land on branch `claude/quirky-galileo-B4wih` and tag @toubatbrian + @livekit/agent-devs for review.

---

PR #5644, opened 2026-05-05
feat(stt): forward session VAD events to STT plugins

sam-s10s:
Currently STT providers do not recieve any external VAD events (e.g. from Silero VAD).

The proposal is to add STT.on_vad_event() hook and have AudioRecognition forward VAD events to the active STT instance, enabling plugins to react to session-level VAD (e.g. finalize on END_OF_SPEECH for externally- driven turn detection modes).

Example code:
```
class STT(stt.STT):

    ...

    def on_vad_event(self, ev: vad.VADEvent) -> None:
        """Auto-finalize when the session VAD reports end of speech.

        Only acts when running in EXTERNAL turn detection mode — other modes
        either delegate end-of-utterance handling to the Speechmatics service
        (ADAPTIVE, SMART_TURN, FIXED) or expect the caller to manage turns
        explicitly.
        """
        if ev.type != vad.VADEventType.END_OF_SPEECH:
            return
        if self._stt_options.turn_detection_mode != TurnDetectionMode.EXTERNAL:
            return
        self.finalize()

    ...

```

### DIFF
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -7,7 +7,10 @@ from collections.abc import AsyncIterable, AsyncIterator
 from dataclasses import dataclass, field
 from enum import Enum, unique
 from types import TracebackType
-from typing import Any, Generic, Literal, TypeVar
+from typing import TYPE_CHECKING, Any, Generic, Literal, TypeVar
+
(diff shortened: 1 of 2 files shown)


theomonnom:
Any reason why you need that? Is it required for some STT?
You can also create your own VAD instance inside your STT impl (like stt.StreamAdapter)

PR #5650, opened 2026-05-05
fix(amd): fix negative zero in amd delay calculation

chenghao-mou:


davidzhao (approved):

toubatbrian:
This is an automated Claude Code Routine created by @toubatbrian. Right now it is in experimentation stage. The automation evaluates merged PRs in `livekit/agents` and ports relevant core runtime / plugin changes into `livekit/agents-js`.

**Decision for this PR: Skip port.**

This PR fixes a negative-zero edge case in the `delay` field calculation when `_speech_ended_at` is `None` (replacing `time.time() - (self._speech_ended_at or time.time())` with `(time.time() - self._speech_ended_at) if self._speech_ended_at else 0.0`).

The agents-js AMD module (`agents/src/voice/amd.ts`) does not currently expose this metric:
- `AMDResult` in JS has only `{ category, transcript, reason, rawResponse, isMachine }` — no `delay` or `speech_duration` fields.
- The JS classifier tracks `speechStartedAt` but not `speechEndedAt`, so the negative-zero arithmetic this PR fixes has no analog there.

Since the buggy code path doesn't exist in agents-js, no port is needed. If/when the JS AMD adopts the `delay`/`speech_duration` metrics, the fix here should be applied at that time.

---

PR #5656, opened 2026-05-06
fix(async-toolset): use underscore in synthetic call_id

### LINKED ISSUE #5652: [Bug] AsyncToolset generates invalid OpenAI Realtime call_ids by using /update_* and /finished
### Bug Description

When using `AsyncToolset` with `livekit-plugins-openai` Realtime, async tool calls can complete successfully internally, but the final async tool result is not delivered back to the agent/user.

The OpenAI Realtime API rejects the injected tool output because the generated `item.call_id` contains `/`, which does not match the required call ID pattern:

```text
^[a-zA-Z0-9_-]+$
```

### Expected Behavior

`AsyncToolset` should generate synthetic function call IDs that are valid for all supported realtime/model backends, including OpenAI Realtime.

For OpenAI Realtime, generated `call_id`s should only contain characters matching:

```text
^[a-zA-Z0-9_-]+$
```

The final return value from an async tool should be injected back into the conversation and trigger the agent’s reply without causing a Realtime API error.

### Reproduction Steps

1. Install LiveKit Agents and the OpenAI plugin:

pip install "livekit-agents==1.5.7" "livekit-plugins-openai==1.5.7"

2. Configure an agent session using gpt-realtime in azure.

3. Register an `AsyncToolset` with an async tool that calls `ctx.update(...)`, then returns a final result later.

4. Ask the agent to call `long_running_tool`.

5. The initial `ctx.update(...)` may work, but when the async tool later returns, OpenAI Realtime rejects the generated `item.call_id`.

Observed error:

RealtimeError(message="Invalid 'item.call_id': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.", type='invalid_request_error', code='invalid_value', param='item.call_id')

### Operating System

macOS Tahoe

### Models Used

gpt-realtime in azure

### Package Versions

```bash
livekit-agents==1.5.7
livekit-plugins-openai==1.5.7
python==3.14
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

Avoid `/` in synthetic async tool call IDs.

For example in `livekit/agents/llm/async_toolset.py`, change:

```python
call_id=f"{self.function_call.call_id}/update_{self._step_idx}"
call_id=f"{call_id}/finished"
```

to:

```python
call_id=f"{self.function_call.call_id}_update_{self._step_idx}"
call_id=f"{call_id}_finished"
```

Or any alternative that matches the regex pattern:

```
^[a-zA-Z0-9_-]+$
```

### Additional Context

_No response_

_No response_

longcw:
## Summary
- Replace `/` with `_` in the synthetic `call_id`s generated for `ctx.update(...)` and the final return in `AsyncToolset`.
- The previous format violated the `^[a-zA-Z0-9_-]+$` pattern enforced by some realtime backends, causing tool outputs to be rejected and never delivered back to the agent.

Closes #5652

### DIFF
--- a/livekit-agents/livekit/agents/llm/async_toolset.py
+++ b/livekit-agents/livekit/agents/llm/async_toolset.py
@@ -107,7 +107,7 @@ class AsyncRunContext(RunContext[Userdata_T]):
         self._step_idx += 1
 
         tool_output = self._make_tool_output(
-            message, call_id=f"{self.function_call.call_id}/update_{self._step_idx}"
+            message, call_id=f"{self.function_call.call_id}_update_{self._step_idx}"
         )
         if tool_output.fnc_call_out is None:
             return
@@ -281,7 +281,7 @@ class AsyncToolset(Toolset):
                 if output is None:
                     return
 
-                tool_output = async_ctx._make_tool_output(output, call_id=f"{call_id}/finished")
+                tool_output = async_ctx._make_tool_output(output, call_id=f"{call_id}_finished")
                 if tool_output.fnc_call_out is None:
                     return
 


chenghao-mou (approved):

toubatbrian:

---

**Port status notice**

This PR is a core runtime fix (synthetic `call_id` format compliance with the `^[a-zA-Z0-9_-]+$` pattern enforced by realtime backends) and would normally be ported to `agents-js`. However, the `AsyncToolset` feature itself has not yet landed in `agents-js` `main` — it is currently being introduced via the open PR livekit/agents-js#1397 ("Add non-blocking tool calls").

Because there is no `async_toolset.ts` on `agents-js` `main` to patch, a standalone port PR cannot be opened today. Instead, the routine will leave a note on livekit/agents-js#1397 so the equivalent `_` (underscore) call_id format is incorporated into that branch before it merges, avoiding the same realtime-backend rejection on the JS side.

cc @toubatbrian @livekit/agent-devs

---

Issue #2359, opened 2025-05-22
Feature Request: Add Explicit Context Caching Support for Gemini Models


hsjun99:
## Problem
Current [implicit caching](https://ai.google.dev/gemini-api/docs/caching?lang=python) for Gemini models doesn't guarantee caching, even when exceeding minimum token thresholds (1,024 tokens for 2.5 Flash, 2,048 tokens for 2.5 Pro). This is the observed behavior after thorough testing.

## Solution
Add explicit context caching support using Gemini API's caching feature:

- **Cache Creation**: Create caches with configurable TTL
- **Guaranteed Savings**: Predictable caching mechanism

Mohd162004:
Hey, just wanted to ask, how did you do implicit caching for the gemini model? and was it the live api model? or was it the STT - LLM - TSS pipeline that you used?

exponentiallyio:
Any update on this please? Gemini Implicit Caching is still broken - I tested through the gemini API

csanz91:
Could you share a bit more about the reasoning or technical trade-offs behind marking this as not planned?

In conversational agent architectures, being able to reuse the system prompt across multiple independent sessions is highly valuable, for example, lower costs, better TTFT.

PR #5661, opened 2026-05-06
feat(plugins-google): add cached_content option for explicit context caching

kamil-bidus:
## Motivation

The Gemini plugin's `LLM` class supports many `GenerateContentConfig` options (thinking_config, retrieval_config, safety_settings, etc.) but not `cached_content`. The plugin already reads `cached_content_token_count` from response usage in `LLMStream._parse_part`, so cache hits surface in metrics — there's just no way to attach a `CachedContent` resource to outgoing requests.

## Change

Add `cached_content: NotGivenOr[str] = NOT_GIVEN` to `LLM.__init__`, propagated through `_LLMOptions` → `chat()` → `extra["cached_content"]` → `GenerateContentConfig` via `**self._extra_kwargs`.

### Request-side suppression

Gemini's API rejects `generateContent` requests that pass `cached_content` together with `system_instruction`, `tools`, or `tool_config` — those fields belong inside the `CachedContent` resource itself, and the API returns a 400 instructing callers to move them.

Without handling that, the parameter would 400 on any realistic agent. So `LLMStream._run` strips `system_instruction`, `tools`, and `tool_config` from the outgoing request whenever `cached_content` is attached. Behaviour is unchanged when `cached_content` is unset.

Cache lifecycle (creation, TTL refresh, deletion) and the choice of what to bake into the cache stay the application's responsibility.

## Compatibility

Default `NOT_GIVEN` keeps existing behaviour unchanged — verified by tests covering both the omission case (no key in `_extra_kwargs`) and the no-cache request path (`system_instruction` and `tools` propagate as before).

Works with both Gemini Developer API (`cachedContents/{id}`) and Vertex AI (`projects/{p}/locations/{l}/cachedContents/{id}`); the plugin passes the string through unmodified.

## Tests

`tests/test_plugin_google_llm.py` — 6 cases:

- **Propagation (3)** — `cached_content` round-trips through `_LLMOptions` and reaches `_extra_kwargs`; default `NOT_GIVEN` produces no key.
- **Suppression (3)** — patching `client.aio.models.generate_content_stream` to capture the `GenerateContentConfig`, the request omits `system_instruction` / `tools` / `tool_config` when `cached_content` is set, and includes them when it isn't (backward compat).

Existing google-plugin tests still pass. `ruff check` / `ruff format` clean.

## Refs

- #2359 (implicit caching reliability)
- https://ai.google.dev/gemini-api/docs/caching

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
@@ -81,6 +81,7 @@ class _LLMOptions:
     http_options: NotGivenOr[types.HttpOptions]
     seed: NotGivenOr[int]
     safety_settings: NotGivenOr[list[types.SafetySettingOrDict]]
+    cached_content: NotGivenOr[str]
 
 
 BLOCKED_REASONS = [
@@ -117,6 +118,7 @@ class LLM(llm.LLM):
         http_options: NotGivenOr[types.HttpOptions] = NOT_GIVEN,
         seed: NotGivenOr[int] = NOT_GIVEN,
         safety_settings: NotGivenOr[list[types.SafetySettingOrDict]] = NOT_GIVEN,
+        cached_content: NotGivenOr[str] = NOT_GIVEN,
         credentials: google.auth.credentials.Credentials | None = None,
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


kamil-bidus:
Superseded by #5675.

PR #5677, opened 2026-05-07
fix(soniox): prevent stream hang on server-side errors

morix1500:
When Soniox sends a WebSocket error frame such as `503 - Cannot continue request (code 4)`, the recv loop only logged the error and never reconnected, leaving the agent process hung indefinitely.

Root cause: the plugin defined `_reconnect_event` but never called `_reconnect_event.set()` anywhere in the codebase (dead code since the plugin's first commit), so the reconnection mechanism was effectively disabled.

This PR removes `_reconnect_event` and aligns the plugin with the Deepgram pattern — surface failures as `APIError` so the base class `SpeechStream._main_task` retry/backoff policy handles recovery.

Changes:
- in-band `error_code`/`error_message` frames → `APIStatusError`
- unexpected WS `CLOSED`/`CLOSE`/`CLOSING` → `APIStatusError`
- mid-stream `aiohttp.ClientError` from gather → `APIConnectionError`
- normal end of session (`finished` frame followed by clean close) → raise an internal `_SessionFinished` so the surrounding `finally` cancels sibling tasks and `_run` returns cleanly without triggering a retry (prevents the gather hang where `_send_audio_task` blocks on an empty queue)
- tolerate Soniox control frames that omit the `tokens` field
- decorate audio/keepalive tasks with `@utils.log_exceptions` for uniform task-level logging

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
@@ -51,6 +51,13 @@ END_TOKEN = "<end>"
 FINALIZED_TOKEN = "<fin>"
 
 
+class _SessionFinished(Exception):
+    """Internal sentinel raised by the recv task when Soniox signals a normal
+    end-of-session (`finished` frame followed by a clean WS close). `_run`
+    catches it so the gather unwinds and sibling tasks (e.g. send_audio
+    blocked on an empty queue) are cancelled by the surrounding finally."""
+
+
 def is_end_token(token: dict) -> bool:
     """Return True if the given token marks an end or finalized event."""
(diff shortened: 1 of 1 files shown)


morix1500:
@davidzhao @chenghao-mou @tinalenguyen
Hello. I am mentioning this to those who frequently review the Soniox plugin.
I would like to request a review for this pull request. Thank you for your time despite your busy schedule.

morix1500:
Closing this PR — the same WebSocket hang on Soniox server errors was independently fixed by #5770 (merged 2026-05-19), which takes a different approach: it wires up the existing `_reconnect_event` instead of removing it and propagating `APIError` through the base class retry path.

Since #5770 is already in upstream and resolves the original symptom, this PR is no longer needed.
Thanks for the reviews!

PR #5690, opened 2026-05-08
Add support for GPT-Realtime-2.0

ShayneP:
## Summary

Update realtime agent output handling for Realtime 2.0 responses and fix transcript synchronization races around overlapping segment lifecycle events.

Realtime 2.0 can produce multiple message items for a single response. The Agents output stack exposes playback through a shared `AudioOutput` segment contract, so this PR forwards realtime message outputs sequentially through the sink. That keeps playback-start/playback-finished events attributable to the correct message and avoids adding or truncating the wrong assistant item during interruption.

This PR also hardens `TranscriptSynchronizer` segment handling when audio/text flush timing is not perfectly aligned.

  ## Changes

  - Support multiple realtime message items in one generation.
  - Forward each realtime message’s audio/text output to completion before starting the next message’s sink forwarding.
  - Wait for per-message audio playout before registering the next message’s playback listener.
  - Preserve existing single-message realtime behavior for current providers.
  - Fix transcript synchronizer stale-duration handling by resetting per-segment pushed audio duration during flush.
  - Keep pending/finalizing transcript segment impls alive until their text/audio inputs are complete.
  - Emit interrupted or already-ready playback-finished events synchronously to preserve existing state transition ordering.
  - Delay non-interrupted synced playback-finished events only when needed to include complete synchronized transcript text.
  - Apply pause/resume to active, pending, and finalizing transcript segments.
  - Add focused regression tests for transcript segment advancement, delayed text completion, and pause/resume behavior.

  ## Compatibility

This is intended to be public-API compatible. It does not change the `AudioOutput` or `TextOutput` interfaces or event payload shapes.

Existing realtime models emitted a single message item per response, so the sequential forwarding path preserves prior behavior while correctly handling the new Realtime 2.0 multi-message response shape.

The transcript synchronizer changes are internal lifecycle fixes. The main observable timing change is that a non-interrupted synced `playback_finished` event may wait for text/audio inputs to complete when playback finishes before transcript input drains. Interrupted and already-ready playback events remain synchronous.

  ## Testing

  - `make check`
  - `make unit-tests`
  - `uv run pytest tests/test_agent_session.py tests/test_transcript_synchronizer.py -q`
  - `git diff --check`

  `make unit-tests` result: 635 passed, 2 skipped

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -75,6 +75,7 @@ from .generation import (
     remove_instructions,
     update_instructions,
 )
+from .io import PlaybackFinishedEvent
 from .speech_handle import DEFAULT_INPUT_DETAILS, InputDetails, SpeechHandle
 from .turn import EndpointingOptions, TurnDetectionMode
 
@@ -3012,6 +3013,8 @@ class AgentActivity(RecognitionHooks):
             2. _TextOutput.first_text_fut (None)
             """
             nonlocal started_speaking_at, started_forwarding_at
+            if started_speaking_at is not None:
+                return
             try:
                 started_speaking_at = fut.result() or time.time()
                 started_forwarding_at = (
@@ -3048,12 +3051,6 @@ class AgentActivity(RecognitionHooks):
             forward_tasks: list[asyncio.Task[Any]] = []
             try:
                 async for msg in generation_ev.message_stream:
-                    if len(forward_tasks) > 0:
-                        logger.warning(
-                            "expected to receive only one message generation from the realtime API"
-                        )
-                        break
-
                     msg_modalities = await msg.modalities
                     tts_text_input: AsyncIterable[str] | None = None
                     if "audio" not in msg_modalities and self.tts:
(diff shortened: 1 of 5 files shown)


tinalenguyen (commented):
tested it and it works well, though i think the start/stop speaking_at times recorded in the chat messages aren't accurate 🤔 all messages in the same generation would have the same started_speaking_at timestamps for the chatmessages

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -3139,16 +3132,22 @@ async def _read_messages(
...
                         text_out.first_text_fut.add_done_callback(_on_first_frame)
 
                     outputs.append((msg, text_out, audio_out))
 
-                await asyncio.gather(*forward_tasks)
+                    await asyncio.gather(*msg_tasks)
+                    msg_tasks.clear()
+
+                    if audio_output is not None and audio_out is not None:
+                        await audio_output.wait_for_playout()
```
longcw, line 3146: should we call `perform_audio_forwarding` once for multiple segment instead of calling it multiple times in one response, like merge the streams from `generation_ev.message_stream`? in that way we don't need to change the output and synchronizer?

longcw: ~btw, I didn't see multiple messages in a single response during my testing, is there a way to trigger that or it's just random?~

update: I saw multiple segment with asking it to do so, like tell me a story with two parts

PR #5692, opened 2026-05-08
(deepgram stt): add redact param

### LINKED ISSUE #5683: livekit-plugins-deepgram: expose redact parameter on STT for PII redaction
### Feature Type

## Summary

The `deepgram.STT` plugin (`livekit-plugins-deepgram`) does not expose Deepgram's `redact` query parameter, even though:

1. **Deepgram's API supports it** for both pre-recorded and streaming requests — see [Deepgram redaction docs](https://developers.deepgram.com/docs/redaction).
2. **The LiveKit Inference path already passes it through** via `extra_kwargs`, as documented in the [Deepgram STT plugin guide → Model parameters](https://docs.livekit.io/agents/models/stt/deepgram/#model-parameters):

   ```python
   inference.STT(
       model="deepgram/nova-3",
       extra_kwargs={"redact": ["pii", "numbers"]},
   )
   ```

So users on **LiveKit Inference get redaction**, but users on the **plugin path (own Deepgram API key) do not** — there is no clean way to enable it short of subclassing `SpeechStream` and overriding `_connect_ws` to inject `redact` into the WebSocket URL.

## Why it matters

Redaction is a hard requirement for voice agents handling regulated data (HIPAA, PCI, GDPR, etc.). Common scenarios:

- Job interview / HR voice agents (candidates may speak phone numbers, addresses, ID numbers).
- Customer support agents that must avoid storing card numbers, SSNs, etc.
- Any product that persists transcripts to long-term storage (S3, databases) and needs PII suppression at the source.

Today, anyone using `deepgram.STT` who needs redaction has to either (a) switch to LiveKit Inference (different billing / key management), (b) subclass the plugin, or (c) add post-hoc redaction in their own code.

## Current source

[`livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py`](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py) — `STTOptions` and the `STT.__init__` signature do not include `redact`. The `live_config` dict built inside `SpeechStream._connect_ws` does not pass it either.

The plugin already supports several Deepgram parameters (`punctuate`, `smart_format`, `numerals`, `profanity_filter`, `keyterm`, `tags`, etc.) — `redact` fits naturally alongside them.

## Proposed change

Add `redact` to the constructor and to `STTOptions`, then forward it into `live_config` in `_connect_ws` (and optionally `recognize_config` in `_recognize_impl` for batch use).

```python
class STT(stt.STT):
    def __init__(
        self,
        *,
        # ... existing args ...
        redact: NotGivenOr[str | list[str]] = NOT_GIVEN,
        # ...
    ):
        # ... existing init ...
        self._opts.redact = redact if is_given(redact) else None

# In SpeechStream._connect_ws:
if self._opts.redact:
    live_config["redact"] = self._opts.redact
```

`redact` accepts the standard Deepgram values (`pci`, `pii`, `phi`, `numbers`, or specific entity classes — see [Supported entity types](https://developers.deepgram.com/docs/supported-entity-types)). Passing a list maps to `redact=...&redact=...` repeated query params, which `urlencode(..., doseq=True)` already handles in [`_to_deepgram_url`](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/_utils.py).

Adding the matching field to `update_options` would also be useful so it can be toggled per-session.

## Environment

- `livekit-agents` 1.5.x
- `livekit-plugins-deepgram` 1.5.5
- Python 3.11

_No response_

### Additional Context

_No response_

tinalenguyen:
closes #5683

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -67,6 +67,7 @@ class STTOptions:
     keywords: list[tuple[str, float]]
     keyterm: str | Sequence[str]
     profanity_filter: bool
+    redact: str | list[str]
     endpoint_url: str
     vad_events: bool = True
     numerals: bool = False
@@ -94,6 +95,7 @@ class STT(stt.STT):
         keyterm: NotGivenOr[str | list[str]] = NOT_GIVEN,
         tags: NotGivenOr[list[str]] = NOT_GIVEN,
         profanity_filter: bool = False,
+        redact: NotGivenOr[str | list[str]] = NOT_GIVEN,
         api_key: NotGivenOr[str] = NOT_GIVEN,
         http_session: aiohttp.ClientSession | None = None,
         base_url: str = "https://api.deepgram.com/v1/listen",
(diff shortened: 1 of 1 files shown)


theomonnom (approved):

toubatbrian:

---

PR #5694, opened 2026-05-09
feat(mcp): per-tool options and progress update for MCPToolset

longcw:
## Summary

`MCPToolset` now extends `AsyncToolset`. Per-tool behavior is configured with `MCPToolOptions`, keyed by tool name via `tool_options`; any tool not listed there is a plain blocking call.

`MCPToolOptions` keeps the three concerns independent:

- `flags: ToolFlag` — e.g. `CANCELLABLE`, `IGNORE_ON_ENTER` (mirrors `@function_tool`).
- `on_duplicate: DuplicateMode` — duplicate-call policy (mirrors `@function_tool`).
- `forward_progress: bool` — forward the MCP server's `report_progress` notifications to `ctx.update()`, so the tool runs in the background and the agent can narrate progress.

When any tool is `ToolFlag.CANCELLABLE`, `lk_agents_get_running_tasks` / `lk_agents_cancel_task` are auto-exposed to the LLM by the activity.

```python
AgentSession(tools=[
    MCPToolset(
        id="mcp",
        mcp_server=MCPServerHTTP(url="...", client_session_timeout_seconds=120),
        tool_options={
            "book_flight": MCPToolOptions(
                flags=ToolFlag.CANCELLABLE,
                on_duplicate="confirm",
                forward_progress=True,
            ),
        },
    ),
])
```

## Example

`examples/voice_agents/mcp/server.py` ships a fast `get_weather` (sync, no progress) and a long-running `book_flight` that emits `ctx.report_progress` updates over ~70s. `mcp-agent.py` configures only `book_flight` (cancellable + forward progress) and bumps `client_session_timeout_seconds=120` since the per-request timeout otherwise fires before it returns.


theomonnom (commented):
lgtm, let's have async_mode enabled by default?

maybe the arguments becomes something like `nonblocking_tools: list[str]`?

theomonnom:
Let's wait for https://github.com/livekit/agents/pull/5711

livekit-agents/livekit/agents/llm/mcp.py
```diff
@@ -47,6 +50,24 @@
...
+    """Per-tool behavior for MCP tools exposed by an :class:`MCPToolset`.
+
+    ``flags`` and ``on_duplicate`` mirror ``@function_tool``. ``forward_progress``
+    forwards the server's ``report_progress`` to ``ctx.update()``, so the tool runs in
+    the background and the agent can narrate.
+    """
+
+    flags: ToolFlag = ToolFlag.NONE
+    on_duplicate: DuplicateMode = "allow"
+    forward_progress: bool = False
```
theomonnom: Should we use a TypedDict?

theomonnom: nit: mb we can use the same name. No strong opinion tho
```suggestion
    report_progress: bool = False
```

theomonnom (approved):

PR #5698, opened 2026-05-10
fix: do not republish tracks on reconnect

davidzhao:
republishing is handled automatically by the RTC SDK. it should be transparent to the agent.
added e2e tests for both resume and reconnect.

depends on: https://github.com/livekit/python-sdks/pull/654


theomonnom (approved):

tests/test_room.py
```diff
@@ -208,3 +219,203 @@ async def test_agent_joins(self):
...
+                publications_after = _agent_audio_publications(user_room)
+                assert len(publications_after) == 1, publications_after
+                assert publications_after[0].sid == sid_before, (
+                    "publication SID changed on resume — engine should have preserved it"
+                )
+
+                # Audio continues to flow uninterrupted.
+                await mon.assert_audio_continuous(
+                    min_rms=_AUDIO_RMS_THRESHOLD, duration=1.5
+                )
```
theomonnom: this makes sense, but make sure it's based on wall-clock, and not the frames being received, otherwise even if the reconnection failed, it may be continuous?

davidzhao: yeah this is based on wall clock

davidzhao: I forgot to check in the test file last time, it's here now

PR #5699, opened 2026-05-10
fix(core): make default user span time explicit

chenghao-mou:
tracer.start_span has a default value when start_time is None, so we need a default value for _user_turn_start as well to make sure they are consistent.

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -1267,7 +1267,8 @@ class AudioRecognition:
         if self._user_turn_span and self._user_turn_span.is_recording():
             return self._user_turn_span
 
-        start_time: float = start_time or time.time()
+        if start_time is None:
+            start_time = time.time()
         start_time_ns = int(start_time * 1_000_000_000)
         self._user_turn_span = tracer.start_span("user_turn", start_time=start_time_ns)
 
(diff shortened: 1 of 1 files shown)


longcw (approved):

toubatbrian:
This is an automated Claude Code Routine created by @toubatbrian. Right now it is in experimentation stage. The automation reviewed this PR for an auto-port into `agents-js`.

**Decision: skipping the port.**

The Python fix targets a consistency issue specific to the Python implementation: previously, calling `_ensure_user_turn_span(start_time=None)` would let `tracer.start_span(..., start_time=None)` fall back to the OTel SDK default while `self._user_turn_start` stayed `None`, leaving the span's actual start time and the separately-tracked `_user_turn_start` field out of sync (the latter is what `_bounce_eou_task` consumes for EOU metrics). The fix materializes `start_time = time.time()` once and assigns both from it.

The equivalent code in `agents-js` (`agents/src/voice/audio_recognition.ts`, `ensureUserTurnSpan`) does not have a sibling `userTurnStart` field that needs to stay in sync with the span's start time:

- The `bounceEOUTask` EOU metric path uses `this.speechStartTime` — which is populated exclusively by the VAD `INFERENCE_DONE` handler, backdated by `ev.rawAccumulatedSpeech`, and reset only on commit / clear. Its lifecycle is independent of `ensureUserTurnSpan`.
- `ensureUserTurnSpan(startTime?)` forwards `startTime` directly to `tracer.startSpan({ ..., startTime })`. When `undefined`, the OTel JS SDK already substitutes `hrtime()` internally, matching the Python post-fix behavior at the span level.

So in agents-js the two-field inconsistency the Python PR fixes does not exist; mirroring the diff verbatim would be a no-op (an explicit `Date.now()` fallback that the SDK already does). No port action needed.

— Automated routine

---

PR #5703, opened 2026-05-11
fix(voice): cancel realtime generation when speech is interrupted

### LINKED ISSUE #5642: await session.interrupt() with gpt-realtime model not reliable
### Bug Description

We have code in our agent which triggers speech with gpt-realtime via session.generate_reply() when certain events happen, e.g. user speaking too long etc.

When using livekit testing framework (text only mode) we noticed that we could not test the functionality well because it seems `await interrupt()` is not cancelling in flight requests before subequent requests, meaning OpenAI rejects subsequent requests (` RealtimeError(message='Conversation already has an active response in progress: resp_Dc7WDsAfSkZTNf34lhYGX. Wait until the response is finished before creating a new one.'`)

I have provided a minimal reproducible example of this behaviour on this issue.

I think that this issue _might_  not be present in production (outside of text only tests) because in real usage, the interrupt awaits the speech handle stopping, so in practice would correctly wait until the interrupt has finished in most cases (as interrupt event should trigger speech stopping). Nonetheless I think there is value in fixing this becuase:
a) any divergence between the test framework and production in behaviour is problematic for testing. 
b) I am not sure that speech handle ending would be 100% reliable indicator that the interrupt has been successfully completed in all possible scenarios (e.g. what about tool call cancellation?) 

### Expected Behavior

- await interrupt() whould await OpenAI ackowledgement that interrupt has been received
- The attached test would pass 

### Reproduction Steps

```bash
import asyncio
import logging

import pytest
from openai.types.realtime import AudioTranscription
from openai.types.realtime.realtime_audio_input_turn_detection import SemanticVad

load_dotenv(".env.local")

_REALTIME_ACTIVE_RESPONSE_SNIPPET = "conversation_already_has_active_response"


class Assistant(Agent):
    """Minimal agent for interrupt / follow-up turn tests."""

    def __init__(self) -> None:
        super().__init__(
            instructions="You are a helpful voice AI assistant. Speak English.",
        )


def _semantic_vad() -> SemanticVad:
    return SemanticVad(
        type="semantic_vad",
        create_response=True,
        interrupt_response=True,
        eagerness="low",
    )


def _realtime_llm() -> openai.realtime.RealtimeModel:
    return openai.realtime.RealtimeModel(
        modalities=["text"],
        input_audio_transcription=AudioTranscription(language="en", model="whisper-1"),
        max_session_duration=20 * 60,
        turn_detection=_semantic_vad(),
    )


@pytest.mark.asyncio
async def test_interrupt_then_follow_up_user_turn(caplog: pytest.LogCaptureFixture) -> None:
    """User-driven turns; interrupt mid first reply; second reply reflects the new ask."""
    with caplog.at_level(logging.ERROR, logger="livekit.plugins.openai"):
        async with (
            _realtime_llm() as realtime_llm,
            AgentSession(llm=realtime_llm, turn_detection="realtime_llm") as session,
        ):
            await session.start(Assistant())

            session.generate_reply(user_input="Reply with 5 sentence essay about cats")
            await asyncio.sleep(0.1)
            await session.interrupt()

            second = session.run(
                user_input=(
                    "Ignore any partial earlier reply. Reply with only the exact phrase "
                    "I HATE DOGS in all caps. Do not add any other words."
                )
            )
            await second

    error_messages = [r.getMessage() for r in caplog.records if r.levelno >= logging.ERROR]
    assert not any(_REALTIME_ACTIVE_RESPONSE_SNIPPET in m for m in error_messages), (
        "OpenAI Realtime logged active-response conflict "
        f"({_REALTIME_ACTIVE_RESPONSE_SNIPPET}): {error_messages!r}"
    )


````
Fails with 

E   AssertionError: OpenAI Realtime logged active-response conflict (conversation_already_has_active_response): ["OpenAI Realtime API returned an error: RealtimeError(message='Conversation already has an active response in progress: resp_Dc7WDsAfSkZTNf34lhYGX. Wait until the response is finished before creating a new one.', type='invalid_request_error', code='conversation_already_has_active_response', event_id='response_create_67056e52992a', param=None)", 'failed to generate a reply: generate_reply timed out.']

### Operating System

Linux

### Models Used

gpt-realtime

### Package Versions

```bash
1.5.7
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
For gpt-realtime, await interrupt() should wait for the OpenAI `cancelled 'acknowledgement' event to be received (with a timeout in case it never is received).
```

### Additional Context

_No response_

_No response_

longcw:
## Summary

`session.interrupt()` was a silent no-op when fired before the realtime server had emitted `response.created`. The local `SpeechHandle` was marked interrupted, but `_realtime_reply_task` had already passed its auth-wait block and continued to call `rt_session.generate_reply()`, after which the server produced a response the agent then discarded. Reported in #5642 ([comment](https://github.com/livekit/agents/issues/5642#issuecomment-4417273612)).

Two changes fix it:

- **Activity-side race.** `_realtime_reply_task` now races the `generate_reply` future against `speech_handle._interrupt_fut` via `wait_if_not_interrupted`. On interrupt, the future is cancelled and the task returns.
- **Plugin-side cancel propagation.** Each realtime plugin's `generate_reply` future now wires an `add_done_callback` that, on cancellation, cleans up tracking state and signals the server (or drops the queued send) where the API supports it:
  - OpenAI (and xAI by inheritance): emits `response.cancel`. 
  - Google: calls `interrupt()` to send `activity_start` to Gemini.
  - Phonic: cancels the queued send task (Phonic has no programmatic server-side cancel).
  - Ultravox: sends a deferred barge-in (`urgency=immediate, defer_response=True`).
  - AWS Nova (experimental): cancels the queued send task (Nova handles in-progress interruption automatically).

Added `test_generate_reply_cancellation` in `tests/test_realtime/test_realtime.py` (parameterized over OpenAI/Azure) which cancels a `generate_reply` future immediately and verifies a subsequent `generate_reply` does not hit `conversation_already_has_active_response`.

Fixes #5642

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2888,18 +2888,24 @@ class AgentActivity(RecognitionHooks):
                     ori_tools = self._rt_session.tools.flatten()
                     await self._rt_session.update_tools(llm.ToolContext(tools).flatten())
 
+            generate_reply_fut = self._rt_session.generate_reply(
+                instructions=instructions or NOT_GIVEN,
+                tool_choice=(model_settings.tool_choice if per_response_tool_choice else NOT_GIVEN),
+                tools=(
+                    llm.ToolContext(tools).flatten()
+                    if per_response_tool_choice and tools is not None
+                    else NOT_GIVEN
+                ),
+            )
+            await speech_handle.wait_if_not_interrupted([generate_reply_fut])
+            if speech_handle.interrupted:
+                # cancel the pending generation; the plugin emits response.cancel
+                if not generate_reply_fut.done():
+                    generate_reply_fut.cancel()
+                return
+
             try:
-                generation_ev = await self._rt_session.generate_reply(
-                    instructions=instructions or NOT_GIVEN,
-                    tool_choice=(
-                        model_settings.tool_choice if per_response_tool_choice else NOT_GIVEN
-                    ),
-                    tools=(
-                        llm.ToolContext(tools).flatten()
-                        if per_response_tool_choice and tools is not None
-                        else NOT_GIVEN
-                    ),
-                )
+                generation_ev = generate_reply_fut.result()
             except llm.RealtimeError as e:
                 logger.error(
                     "failed to generate a reply%s: %s",
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2905,7 +2905,7 @@ class AgentActivity(RecognitionHooks):
                 return
 
             try:
-                generation_ev = generate_reply_fut.result()
+                generation_ev = await generate_reply_fut
             except llm.RealtimeError as e:
                 logger.error(
                     "failed to generate a reply%s: %s",
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -717,7 +717,11 @@ class RealtimeSession(llm.RealtimeSession):
             logger.warning(
                 "generate_reply called while another generation is pending, cancelling previous."
             )
-            self._pending_generation_fut.cancel("Superseded by new generate_reply call")
+            # clear the slot before cancelling so the done callback doesn't treat it
+            # as an external cancellation and signal the server.
+            old_fut = self._pending_generation_fut
+            self._pending_generation_fut = None
+            old_fut.cancel("Superseded by new generate_reply call")
 
         fut = asyncio.Future[llm.GenerationCreatedEvent]()
         self._pending_generation_fut = fut
@@ -749,7 +753,17 @@ class RealtimeSession(llm.RealtimeSession):
                     self._pending_generation_fut = None
 
         timeout_handle = asyncio.get_event_loop().call_later(5.0, _on_timeout)
-        fut.add_done_callback(lambda _: timeout_handle.cancel())
+
+        def _on_fut_done(f: asyncio.Future[llm.GenerationCreatedEvent]) -> None:
+            timeout_handle.cancel()
+            is_current = self._pending_generation_fut is fut
+            if is_current:
+                self._pending_generation_fut = None
+            if f.cancelled() and is_current:
+                # external cancel: signal interrupt to Gemini via activity_start
+                self.interrupt()
+
+        fut.add_done_callback(_on_fut_done)
 
         return fut
 
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
@@ -2116,7 +2116,7 @@ class RealtimeSession(  # noqa: F811
                     if self._pending_generation_fut is fut:
                         self._pending_generation_fut = None
 
-            asyncio.create_task(_send_text())
+            send_task = asyncio.create_task(_send_text())
 
             # Set timeout from model configuration
             def _on_timeout() -> None:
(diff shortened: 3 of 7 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2888,18 +2888,24 @@ async def _realtime_reply_task(
...
+                    llm.ToolContext(tools).flatten()
+                    if per_response_tool_choice and tools is not None
+                    else NOT_GIVEN
+                ),
+            )
+            await speech_handle.wait_if_not_interrupted([generate_reply_fut])
+            if speech_handle.interrupted:
+                # cancel the pending generation; the plugin emits response.cancel
+                if not generate_reply_fut.done():
+                    generate_reply_fut.cancel()
```
chenghao-mou, line 2904: if the future is done, should we call `self._rt_session.interrupt()` instead?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -749,7 +753,17 @@ def _on_timeout() -> None:
...
-        fut.add_done_callback(lambda _: timeout_handle.cancel())
+
+        def _on_fut_done(f: asyncio.Future[llm.GenerationCreatedEvent]) -> None:
+            timeout_handle.cancel()
+            is_current = self._pending_generation_fut is fut
+            if is_current:
+                self._pending_generation_fut = None
+            if f.cancelled() and is_current:
+                # external cancel: signal interrupt to Gemini via activity_start
+                self.interrupt()
```
chenghao-mou, line 764: This is guarded by `_manual_activity_detection` in `start_user_activity`, should we send it here directly?

longcw: I think we still want the guard for `self._opts.realtime_input_config.activity_handling == types.ActivityHandling.NO_INTERRUPTION`?

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: maybe not, ideally this shouldn't happen

theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2888,18 +2888,24 @@ async def _realtime_reply_task(
...
-                    tool_choice=(
-                        model_settings.tool_choice if per_response_tool_choice else NOT_GIVEN
-                    ),
-                    tools=(
-                        llm.ToolContext(tools).flatten()
-                        if per_response_tool_choice and tools is not None
-                        else NOT_GIVEN
-                    ),
-                )
+                generation_ev = generate_reply_fut.result()
```
theomonnom: Isn't this raising CancelledError as well? 

This will propagate to the current task and have unwanted effects?

theomonnom:
can you make sure to test this PR with xAI, I know they have some discrepancies with the OpenAI API around generation cancellation

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: it's returned after the fut is cancelled.

longcw:
> can you make sure to test this PR with xAI, I know they have some discrepancies with the OpenAI API around generation cancellation

xAI realtime doesn't support `response.cancel` when server VAD is enabled. other than that, this pr shouldn't affect it.

bml1g12:
@longcw  I realised today that this PR improves things but we still do not wait for a cancelled() event emitted from OpenAI, so there is still the possibility of 

session.generate_reply()
session.interrupt() # cancel event sent to gpt-realtime, but is delayed in transit
session.generate_reply() # rejected by OpenAI due to "response in flight" as the cancelled() ack event from OpenAI has not yet been emitted/ackowledged 

I am not yet sure if we see this is reality though

longcw:
@bml1g12 

the cancel and new response create requests are sent to the server in order, the only open question is whether OpenAI's cancel is synchronous, and we have no evidence it isn't, unless someone reproduces the rejection against the API.

bml1g12:
> @bml1g12
> 
> the cancel and new response create requests are sent to the server in order, the only open question is whether OpenAI's cancel is synchronous, and we have no evidence it isn't, unless someone reproduces the rejection against the API.

I think its fair to assume it is not sychronous for two reasons:

1) The docs say the event detection should be part of the operation 

> The client monitors for new input_audio_buffer.speech_started events from the server, which indicate the user has started speaking. The server will automatically cancel any in-progress model response and a `response.cancelled` event will be emitted. **When the client detects this event, it should immediately stop playback** of any audio currently being played from the model. It should note how much of the last audio response was played before the interruption.

2) If it was synchronous, then the cancelled event would be fairly redundant 

That said, it might be that the time between sending a cancel and receiving this event is so small that for practical purposes it makes no difference...

We sometimes see generate reply rejected as OpenAI claims responses are are still in flight - which should not be possible as we now strictly queue all requests using an internal queue. I do wonder if this is related to sometimes us sometimes interrupting the in flight response without await of cancelled event and then calling generate reply whilst the interupt is still pending, but its tough to debug if that's the root cause :thinking:   if I find proof that's the case I'll share it but for now it's just a hypothesis.

PR #5708, opened 2026-05-11
Add Hume TTS input streaming

tayek333:
## Summary

This adds native websocket input streaming support to the Hume TTS plugin using Hume's `/v0/tts/stream/input` endpoint: https://dev.hume.ai/reference/text-to-speech-tts/stream-input

- Adds `SynthesizeStream` for Hume so LiveKit can forward text chunks directly to Hume instead of going through the sentence-based `StreamAdapter`
- Advertises `TTSCapabilities(streaming=True)` by default, with `streaming=False` available to preserve the previous sentence-adapter behavior in LiveKit's default TTS node
- Keeps the existing HTTP `synthesize()` path unchanged
- Sends Hume stream-input query params for `format_type`, `instant_mode`, `version`, `context_generation_id`, `no_binary=true`, and `strip_headers=true`
- Sends `close=true` on stream end and handles Hume audio/timestamp/error websocket messages
- Adds focused unit coverage for the capability toggle, context validation, query params, direct chunk forwarding, and PCM audio emission

## Tests

- `uv run pytest tests/test_plugin_hume_tts.py tests/test_plugin_elevenlabs_tts.py -q`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run mypy --untyped-calls-exclude=smithy_aws_core -p livekit.plugins.hume`

### DIFF
--- a/livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
+++ b/livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
@@ -18,9 +18,11 @@ import asyncio
 import base64
 import json
 import os
+import weakref
 from dataclasses import dataclass, replace
 from enum import Enum
 from typing import Any, Literal, TypedDict
+from urllib.parse import urlencode
 
 import aiohttp
 
@@ -28,6 +30,7 @@ from livekit.agents import (
     APIConnectionError,
     APIConnectOptions,
     APIError,
+    APIStatusError,
     APITimeoutError,
     tts,
     utils,
@@ -82,6 +85,7 @@ DEFAULT_HEADERS = {
 }
 API_AUTH_HEADER = "X-Hume-Api-Key"
 STREAM_PATH = "/v0/tts/stream/json"
+INPUT_STREAM_PATH = "/v0/tts/stream/input"
 DEFAULT_BASE_URL = "https://api.hume.ai"
 SUPPORTED_SAMPLE_RATE = 48000
 DEFAULT_VOICE = VoiceByName(name="Male English Actor", provider=VoiceProvider.hume)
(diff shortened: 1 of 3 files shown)


livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
```diff
@@ -302,3 +340,169 @@ async def _run(self, output_emitter: tts.AudioEmitter) -> None:
...
+            num_channels=self._tts.num_channels,
+            mime_type=f"audio/{self._opts.audio_format.value}",
+            stream=True,
+        )
+
+        async def send_task(ws: aiohttp.ClientWebSocketResponse) -> None:
+            async for data in self._input_ch:
+                if isinstance(data, self._FlushSentinel):
+                    if not self._input_ended:
+                        await ws.send_str(json.dumps({"flush": True}))
```
longcw, line 369: during my testing, the TTS generate audio based on the chunk sent to the API, i.e. there is no sentence buffer in the server side. we may need to add a sentence tokenizer in the plugin, just like other TTS plugins, for example https://github.com/livekit/agents/blob/09ca0cd3cb042b24995fedf6b7d226710195303b/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py#L418

```
Sending data: Hi
Sending data:  ther
Sending data: e! I'm
Sending data:  
Sending data: Kelly
Sending data: , your fr
Sending data: iendly
Sending data:  voice ass
Sending data: istant
```

tests/test_plugin_hume_tts.py
```diff
@@ -0,0 +1,153 @@
+from __future__ import annotations
```
longcw, line 1: let's delete this test file, it's not very helpful IMO.

livekit-plugins/livekit-plugins-hume/livekit/plugins/hume/tts.py
```diff
@@ -302,3 +340,169 @@ async def _run(self, output_emitter: tts.AudioEmitter) -> None:
...
+    message: dict[str, Any] = {"text": text}
+
+    if opts.voice:
+        message["voice"] = opts.voice
+    if opts.description:
+        message["description"] = opts.description
+    if opts.speed:
+        message["speed"] = opts.speed
+    if opts.trailing_silence:
+        message["trailing_silence"] = opts.trailing_silence
```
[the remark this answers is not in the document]
longcw: I think this comment makes sense.

longcw (commented):
thanks for the pr! I left some comments, otherwise looks good to me

PR #5712, opened 2026-05-12
feat(gemini llm): add media resolution option to LLM and RealtimeModel

### LINKED ISSUE #5710: gemini: add media_resolution parameter to Gemini LLM class
### Feature Type

Nice to have

### Feature Description

By default the new family of Gemini 3 models use the MEDIA_RESOLUTION_HIGH option, that results in high token usage when passing multiple images.
By allowing to set the media_resolution parameter to a lower setting, the token usage for the images can be reduced by a factor of around 4x while keeping the same level of understanding of the image contents.

_No response_

### Additional Context

_No response_

csanz91:
By default the new family of Gemini 3 models use the MEDIA_RESOLUTION_HIGH option, that results in high token usage when passing multiple images.
By allowing to set the media_resolution parameter to a lower setting, the token usage for the images can be reduced by a factor of around 4x while keeping the same level of understanding of the image contents.

Fixes #5710


tinalenguyen (approved):
lgtm, thank you!

PR #5713, opened 2026-05-12
chore(amd): update default models and default behavior

chenghao-mou:


livekit-agents/livekit/agents/utils/misc.py
```diff
@@ -56,3 +56,18 @@ def is_dev_mode() -> bool:
 def is_hosted() -> bool:
     """Return whether the agent is hosted on LiveKit Cloud."""
     return os.getenv("LIVEKIT_REMOTE_EOT_URL") is not None
+
+
+def is_using_cloud() -> bool:
```
theomonnom: `is_cloud` vs `is_using_cloud` is kind of confusing

theomonnom (approved):

livekit-agents/livekit/agents/utils/misc.py
chenghao-mou: I agree. I will inline this call instead.

PR #5737, opened 2026-05-14
fix: suppress expected STT failover input-ended errors

### LINKED ISSUE #5736: FallbackAdapter logs RuntimeError("input ended") per audio frame during STT failover
### Environment
- `livekit-agents` 1.5.8
- Python 3.12, prod
- Streaming STT primary + streaming STT fallback via `stt.FallbackAdapter`

### Summary
When the primary STT in `stt.FallbackAdapter` dies mid-stream (e.g. `APITimeoutError` during WebSocket init), the adapter logs an ERROR for **every audio frame** during the failover gap until the next provider's stream is constructed. In one prod call we observed **175 ERROR records in ~2.8 seconds** (≈ one per 20 ms audio frame), each with body `"error happened in forwarding input"` and a `RuntimeError: <primary>.SpeechStream input ended` exception attached.

[LiveKit Cloud session](https://cloud.livekit.io/projects/p_2tsga5unx60/sessions/RM_SsoPPLuZ5T9S/observability?mode=logs) showing the noise

### Root cause
In [`fallback_adapter.py:300-320`](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/stt/fallback_adapter.py#L300-L320), `_forward_input_task` runs concurrently with the for-loop that iterates `_stt_instances`. When the primary's `_run` raises, its `RecognizeStream._input_ch` is closed. The for-loop catches the `APIError` and proceeds to construct the fallback stream, but in the meantime the still-running forwarder is calling `main_stream.push_frame(data)` on the dead primary. Each call raises `RuntimeError("...SpeechStream input ended")` from `_check_input_not_ended` ([`stt.py:522-525`](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/stt/stt.py#L522-L525)), which is caught by the bare `except Exception` and logged via `logger.exception("error happened in forwarding input")`.

### Expected behavior
A successful failover should not emit any ERROR-level logs from the forwarder. Suggestions (any one would do):
- Skip the push when `main_stream._input_ch.closed`.
- Catch `RuntimeError("input ended")` separately and log at `debug` — it's an expected race during failover.
- Cancel/replace the forwarder task atomically with the `main_stream` swap (or have it re-read the active stream after each iteration).

### Impact
Failover still completes correctly — pure log noise. But it drowns genuine errors in observability and inflates alert volume.

nightcityblade:
## Summary

Suppresses expected `RuntimeError("... input ended")` races from the STT `FallbackAdapter` forwarder during failover, while preserving error logging for unexpected forwarding failures.

## Testing

- `uv run ruff check livekit-agents/livekit/agents/stt/fallback_adapter.py`
- `uv run pytest tests/test_stt_fallback.py -q`

Fixes #5736

### DIFF
--- a/livekit-agents/livekit/agents/stt/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/stt/fallback_adapter.py
@@ -314,6 +314,11 @@ class FallbackRecognizeStream(RecognizeStream):
                             main_stream.push_frame(data)
                         elif isinstance(data, self._FlushSentinel):
                             main_stream.flush()
+                    except RuntimeError as e:
+                        if "input ended" not in str(e):
+                            logger.exception(
+                                "error happened in forwarding input", extra={"streamed": True}
+                            )
                     except Exception:
                         logger.exception(
                             "error happened in forwarding input", extra={"streamed": True}
--- a/livekit-agents/livekit/agents/stt/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/stt/fallback_adapter.py
@@ -308,17 +308,16 @@ class FallbackRecognizeStream(RecognizeStream):
                     except Exception:
                         pass
 
-                if main_stream is not None:
+                if (
+                    main_stream is not None
+                    and not main_stream._input_ch.closed
+                    and not main_stream._event_ch.closed
+                ):
                     try:
                         if isinstance(data, rtc.AudioFrame):
                             main_stream.push_frame(data)
                         elif isinstance(data, self._FlushSentinel):
                             main_stream.flush()
-                    except RuntimeError as e:
-                        if "input ended" not in str(e):
-                            logger.exception(
-                                "error happened in forwarding input", extra={"streamed": True}
-                            )
                     except Exception:
                         logger.exception(
                             "error happened in forwarding input", extra={"streamed": True}


livekit-agents/livekit/agents/stt/fallback_adapter.py
```diff
@@ -318,6 +314,11 @@ async def _forward_input_task() -> None:
                             main_stream.push_frame(data)
                         elif isinstance(data, self._FlushSentinel):
                             main_stream.flush()
+                    except RuntimeError as e:
+                        if "input ended" not in str(e):
```
longcw: instead of suppressing the specific exception, I think it's better to add a check in L311 to ignore frames when the `main_stream` is closed, e.g.
```python
                # skip pushing during the gap between primary's _run raising and the
                # outer loop swapping main_stream to the next stt
                if (
                    main_stream is not None
                    and not main_stream._input_ch.closed
                    and not main_stream._event_ch.closed
                ):
```

longcw (changes requested):

livekit-agents/livekit/agents/stt/fallback_adapter.py
nightcityblade: Good call — I switched this to guard against closed input/event channels before pushing into the active stream, instead of suppressing the RuntimeError after the fact. Updated in 1b17793, and `uv run pytest tests/test_stt_fallback.py` still passes.

longcw (approved):

PR #5738, opened 2026-05-14
feat: support python -m livekit.agents.download for asset fetching

davidzhao:
enables more efficient docker images by installing plugin layer before application code gets copied over

### DIFF
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -1959,12 +1959,12 @@ def _build_cli(server: AgentServer) -> typer.Typer:
 
         c.print(
             "[yellow]`download-files` via the agent CLI is deprecated. "
-            "Use `python -m livekit.agents.download` instead — it discovers installed "
+            "Use `python -m livekit.agents download-files` instead — it discovers installed "
             "plugins without loading your agent code.[/yellow]"
         )
         warnings.warn(
             "`download-files` via the agent CLI is deprecated. "
-            "Use `python -m livekit.agents.download` instead.",
+            "Use `python -m livekit.agents download-files` instead.",
             DeprecationWarning,
             stacklevel=2,
         )
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -1950,16 +1950,26 @@ def _build_cli(server: AgentServer) -> typer.Typer:
 
     @app.command()
     def download_files() -> None:
+        import warnings
+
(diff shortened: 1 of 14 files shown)


real-danm (approved):
lgtm, but a bit of a rubber stamp, i'm not aware of any downstream effect this will have.

livekit-agents/livekit/agents/cli/cli.py
```diff
@@ -1950,16 +1950,26 @@ async def simulate_job() -> None:
...
         c.enabled = True
 
         _configure_logger(c, logging.DEBUG)
 
-        try:
-            # import_data = get_import_data(path=path)
-            # c.print(f"Importing from {import_data.module_data.extra_sys_path}")
-            # c.print(" ")
+        c.print(
+            "[yellow]`download-files` via the agent CLI is deprecated. "
```
longcw, line 1961: update the message in https://github.com/livekit/agents/blob/b4c1195ce2c9826d7dad242d650e13dd19c8ba2a/livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/base.py#L144 as well?

longcw (approved):
lgtm! somehting nit:

livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/__init__.py
```diff
@@ -17,10 +17,18 @@
...
+
+from .base import EOUPlugin
+from .english import _EUORunnerEn
+from .multilingual import _EUORunnerMultilingual
 from .version import __version__
 
 __all__ = ["english", "multilingual", "__version__"]
 
+Plugin.register_plugin(EOUPlugin(_EUORunnerEn))
+Plugin.register_plugin(EOUPlugin(_EUORunnerMultilingual))
```
longcw, line 30: this will download both en and multilingual models even only one of them is used?

davidzhao: yes, because there is no way to differentiate them as long as the plugin is installed. this is temporary until the next model is released.

chenghao-mou: about that, are we thinking about removing them entirely from the plugin for the new release?

theomonnom (commented):
Can we use a generic CLI script instead?

 You can add a `__main__.py` inside livekit/agents

Then maybe it can becomes` python3 -m livekit.agents download-files`

So it can be reused for more stuff in the future

PR #5744, opened 2026-05-15
fix(voice): preserve LLM tool reply on per-step interrupt

u9g:
## Summary
- When a tool called `ctx.speech_handle.interrupt()` (or `session.interrupt()` fired on a background speech), the owning `SpeechHandle`'s interrupt flag leaked into the subsequent tool-reply step. The scheduling task skipped the re-scheduled handle and the new reply task short-circuited after invoking the LLM — so the conversation paid for the LLM call but the user heard nothing.
- Scope the interrupt flag to one playout step: `SpeechHandle._reset_interrupt_for_next_step()` re-arms the interrupt future (and cancels the per-interrupt 5s safety timer), and `_pipeline_reply_task_impl` calls it unconditionally before scheduling the tool-reply step on the same handle.
- Tools that want to suppress the LLM reply intentionally should raise `StopResponse`.

## Test plan
- [x] `uv run pytest tests/test_say_interrupt_race.py` — control + repro both pass
- [x] `uv run pytest tests/test_agent_session.py tests/test_nested_agent_task.py tests/test_drain_timeout.py tests/test_ivr_activity.py` — 36 passed, no regressions
- [x] `make type-check` — clean
- [x] `uv run ruff check` / `ruff format --check` — clean

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2809,6 +2809,20 @@ class AgentActivity(RecognitionHooks):
                     add_if_missing=False,
                 )
 
+                # The interrupt flag is scoped to a single playout step. If the
+                # prior step (initial reply or earlier tool reply) was
+                # interrupted — either by user code calling
+                # `ctx.speech_handle.interrupt()` inside the tool or by
+                # `session.interrupt()` hitting this handle while it was in
+                # `_background_speeches` — clear it before scheduling the tool
+                # reply. Otherwise the scheduling task skips the handle when
+                # popping it from the queue (interrupted check in
+                # `_scheduling_task`) and the new reply task aborts on its own
+                # interrupted check after invoking the LLM, silently dropping
+                # the tool reply. To intentionally suppress the tool reply,
+                # tools should raise `StopResponse`.
+                speech_handle._reset_interrupt_for_next_step()
+
                 tool_response_task = self._create_speech_task(
                     self._pipeline_reply_task(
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2809,6 +2809,20 @@ def _tool_execution_completed_cb(out: ToolExecutionOutput) -> None:
...
+                # `ctx.speech_handle.interrupt()` inside the tool or by
+                # `session.interrupt()` hitting this handle while it was in
+                # `_background_speeches` — clear it before scheduling the tool
+                # reply. Otherwise the scheduling task skips the handle when
+                # popping it from the queue (interrupted check in
+                # `_scheduling_task`) and the new reply task aborts on its own
+                # interrupted check after invoking the LLM, silently dropping
+                # the tool reply. To intentionally suppress the tool reply,
+                # tools should raise `StopResponse`.
+                speech_handle._reset_interrupt_for_next_step()
```
longcw, line 2824: thanks for digging into this, the symptom is real but i don't think resetting the interrupt flag is the right fix.

`SpeechHandle` is turn-level by design and interrupt() means "abandon this turn." 

the actual cost you're hitting (LLM called but nothing audible) is real, but it comes from the late `if speech_handle.interrupted: return` check — the LLM call has already started by then. i'd rather fix it at the scheduling site: if the handle is already interrupted, don't schedule the tool-reply task at all.

could you share an example of the real-world case you're hitting? I want to understand what the tool is actually trying to do so we can find a fix.

PR #5748, opened 2026-05-16
feat(rime): add coda model support

MaCaki:
## Summary

Adds support for Rime's `coda` TTS model to `livekit-plugins-rime`, alongside the existing `mistv2`, `mistv3`, and `arcana` models.

Coda mirrors Arcana's chunked HTTP and WebSocket streaming paths but intentionally does not expose `temperature`, `top_p`, or `repetition_penalty`. Its server-side payload accepts only `lang` and `max_tokens` (plus `samplingRate` on the chunked endpoint).

Notable details:

- New `_CodaOptions` dataclass (`max_tokens`, `lang`, `sample_rate`) wired through `_TTSOptions`, `_model_params`, `TTS.__init__`, `update_options`, and `ChunkedStream._run`.
- `_timeout_for_model` returns the 4-minute Arcana timeout for `coda` as well.
- Default speaker is `lyra` when `speaker` is unset and `model="coda"`.
- Public constructor and `update_options` signatures are unchanged; `temperature` / `top_p` / `repetition_penalty` are silently ignored for coda (same pattern Mist already uses for the Arcana-only kwargs).

## Test plan

- [x] `make check` (format, lint, mypy) passes
- [x] Smoke-tested chunked HTTP path against the live Rime API — `rime.TTS(model="coda").synthesize(...)` returns audio with expected sample rate and final-frame flag
- [x] Smoke-tested WebSocket streaming path against the live Rime API — `rime.TTS(model="coda", use_websocket=True).stream()` streams audio and ends cleanly
- [x] Added integration entry `id="rime-coda"` in `tests/test_tts.py::SYNTHESIZE_TTS` (requires `RIME_API_KEY` + `DEEPGRAM_API_KEY` and toxiproxy)

### DIFF
--- a/livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
+++ b/livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
@@ -44,7 +44,7 @@ from livekit.agents.voice.io import TimedString
 
 from .langs import TTSLangs
 from .log import logger
-from .models import ArcanaVoices, DefaultMistVoice, TTSModels
+from .models import ArcanaVoices, DefaultCodaVoice, DefaultMistVoice, TTSModels
 
 # arcana can take as long as 80% of the total duration of the audio it's synthesizing.
 ARCANA_MODEL_TIMEOUT = 60 * 4
@@ -59,6 +59,7 @@ class _TTSOptions:
     model: TTSModels | str
     speaker: str
     arcana_options: _ArcanaOptions | None = None
+    coda_options: _CodaOptions | None = None
     mist_options: _MistOptions | None = None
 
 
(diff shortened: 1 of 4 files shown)


livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
```diff
@@ -87,7 +95,7 @@ def _is_mist_model(model: TTSModels | str) -> bool:
 
 
 def _timeout_for_model(model: TTSModels | str) -> int:
-    if model == "arcana":
+    if model == "arcana" or model == "coda":
```
MaCaki, line 98: While coda is faster than `arcana`, with an rtf of ~0.5, it probably still makes sense to reuse the `ARCANA_MODEL_TIMEOUT`.  

That said, we can rename it to acknowledge the dual use.

davidzhao (approved):

PR #5750, opened 2026-05-16
(speechmatics + inference): add VAD

tinalenguyen:
When set to `EXTERNAL` mode, the Speechmatics STT needs `finalize()` to be called to flush the partial transcripts and mark as end of speech. We pass a VAD, similar to Mistral's plugin, or initialize one so it works right out the box.

We mirror the plugin in the inference code; we also accept/initialize a VAD. Inference is already set to handle the VAD event.

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -207,7 +207,7 @@ def _parse_model_string(model: str) -> tuple[str, NotGivenOr[LanguageCode]]:
 def _resolve_vad_for_model(
     model: NotGivenOr[STTModels | str],
     vad_instance: vad.VAD | None,
-) -> tuple[bool, vad.VAD | None]:
+) -> vad.VAD | None:
     is_speechmatics = (
         is_given(model) and isinstance(model, str) and model.startswith("speechmatics/")
     )
@@ -216,7 +216,7 @@ def _resolve_vad_for_model(
             "`vad` will be ignored: model %r handles endpointing server-side.",
             model,
         )
-        vad_instance = None
+        return None
     if is_speechmatics and vad_instance is None:
         try:
             from livekit.plugins.silero import VAD as SileroVAD
@@ -226,7 +226,7 @@ def _resolve_vad_for_model(
                 f"{model!r} does not handle endpointing server-side."
             ) from e
         vad_instance = SileroVAD.load()
-    return is_speechmatics, vad_instance
+    return vad_instance
 
 
 def _normalize_fallback(
@@ -471,7 +471,7 @@ class STT(stt.STT):
             if is_given(parsed_language) and not is_given(language):
                 language = parsed_language
 
-        is_speechmatics, vad = _resolve_vad_for_model(model, vad if is_given(vad) else None)
+        vad = _resolve_vad_for_model(model, vad if is_given(vad) else None)
 
         super().__init__(
             capabilities=stt.STTCapabilities(
@@ -480,7 +480,6 @@ class STT(stt.STT):
                 diarization=diarization_enabled,
                 aligned_transcript="word",
                 offline_recognize=False,
-                server_endpointing=not is_speechmatics,
             ),
         )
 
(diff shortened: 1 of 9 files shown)


livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -434,23 +464,26 @@ def __init__(
...
         )
 
+        # Parse language from model string if provided: "provider/model:language"
+        if is_given(model) and isinstance(model, str):
+            parsed_model, parsed_language = _parse_model_string(model)
+            model = parsed_model
+            if is_given(parsed_language) and not is_given(language):
+                language = parsed_language
+
+        is_speechmatics, vad = _resolve_vad_for_model(model, vad if is_given(vad) else None)
```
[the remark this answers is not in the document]
tinalenguyen: as of right now, speechmatics inference STT models need a VAD to be run here. it is possible they will support server-side endpointing in the future. this differs from the speechmatics stt plugin approach, which already exposes `finalize()` which allows users to flush end-of-speech on their own

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -395,6 +396,21 @@ def __init__(
...
+        # If the STT does its own endpointing, the FINAL_TRANSCRIPT is the turn
+        # boundary — no additional `min_delay` buffer is needed. Only apply when
+        # the user didn't explicitly pin `min_endpointing_delay` or `turn_handling`.
+        if (
+            self._stt is not None
+            and self._stt.capabilities.server_endpointing
+            and not is_given(min_endpointing_delay)
+            and not _user_provided_turn_handling
+        ):
+            endpointing["min_delay"] = 0.0
```
longcw: `min_endpointing_delay` is deprecated and we should check if user specified `min_delay` in `turn_handling`? also, should we move this to a separate pr since it's not related to speechmatics plugin.

tinalenguyen: makes sense, i also thought that the stt capability would play well with indicating STTs that need VAD. i removed those changes here for now though

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -204,6 +204,31 @@ def _parse_model_string(model: str) -> tuple[str, NotGivenOr[LanguageCode]]:
...
+            from livekit.plugins.silero import VAD as SileroVAD
+        except ImportError as e:
+            raise ImportError(
+                "livekit-plugins-silero is required: model "
+                f"{model!r} does not handle endpointing server-side."
+            ) from e
+        vad_instance = SileroVAD.load()
+    return vad_instance
+
+
```
russellmartin-livekit, line 231: In the case where AgentSession has VAD wouldn't this mean we have 2 VAD instances?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -666,6 +717,18 @@ def add_message(message: dict[str, Any]) -> None:
...
+            if self._vad_stream is not None:
+                await self._vad_stream.aclose()
+                self._vad_stream = None
+
+            if vad_task is not None:
+                vad_task.cancel()
+                try:
+                    await vad_task
+                except asyncio.CancelledError:
+                    pass
```
longcw: nit: use `utils.cancel_and_wait`

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -667,12 +718,28 @@ async def send_task(ws: aiohttp.ClientWebSocketResponse) -> None:
...
 
+        @utils.log_exceptions(logger=logger)
+        async def vad_task(ws: aiohttp.ClientWebSocketResponse, stream: vad.VADStream) -> None:
+            async for ev in stream:
+                if ev.type != vad.VADEventType.END_OF_SPEECH:
+                    continue
+                if ws.closed:
+                    return
+                try:
+                    await ws.send_str(json.dumps({"type": "session.finalize"}))
```
longcw, line 738: one questions, what will happen if VAD fires EOS on noise, will the STT return an empty or a random transcript?

longcw (approved):
lgtm! something nit:

livekit-agents/livekit/agents/inference/stt.py
tinalenguyen: yes, to use just 1 would require the user to store it and pass the same instance

maybe it would be helpful for the user to have separate settings for stt and session level vad, but as of right now the session vad can't be connected to stt

livekit-agents/livekit/agents/inference/stt.py
tinalenguyen: i believe the STT will return an empty one ""

PR #5751, opened 2026-05-16
feat(examples): publish examples manifest to agents-jukebox

theomonnom:


davidzhao (approved):

examples/playground.yaml
```diff
@@ -0,0 +1,256 @@
...
+        000000000000
+        011111111110
+        011111111110
+        000000000000
+        000000000000
+    # The playground reads these `controls` and renders a labeled
+    # pill dropdown per control. Picking an option fires an RPC on
+    # the agent participant — `rpc` is the method name, payload is
+    # `{"value": "…"}`. Defaults match the agent's DEFAULT_* constants.
+    controls:
```
bcherry, line 172: we need to figure out a less brittle way to sync the model list here. these change in inference all the time - both our recommended defaults and the models available. It's already _almost_ available via `lk docs pricing-info --json`, but a couple details aren't quite there. but we should be able to make this not use a static list at all but instead query an API on the docs site live from the page

PR #5765, opened 2026-05-18
(speechmatics stt): update default mode to external

tinalenguyen:


livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -125,7 +125,7 @@ def __init__(
         *,
         api_key: NotGivenOr[str] = NOT_GIVEN,
         base_url: NotGivenOr[str] = NOT_GIVEN,
-        turn_detection_mode: TurnDetectionMode = TurnDetectionMode.ADAPTIVE,
+        turn_detection_mode: TurnDetectionMode = TurnDetectionMode.EXTERNAL,
```
toubatbrian, line 128: Since speechmatics doesn't come with an VAD, what would happen if user specify `TurnDetectionMode.ADAPTIVE` here, just curious

toubatbrian (approved):

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
tinalenguyen: their SDK has a VAD built in so it would work out of the box

PR #5770, opened 2026-05-19
fix(soniox): plugin reliability fixes

mihafabcic-soniox:
## Summary                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                   
Fixes 5 bugs in the Soniox plugin.                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                   
### TTS                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                   
- **Hallucinated speech on empty text** (`afa65855`, `7165e9c9`). Soniox TTS hallucinates 1-3s of random audio on empty text. Plugin now omits the empty `text` field from the payload and short-circuits `session.say("")` before it reaches the server.                                          
- **408 timeouts on slow LLM responses** (`c22910f4`). Server starts a ~5s idle timer on stream open. If the LLM is slow to produce its first token (long context, tool calls), the timer fires and the turn dies with a 408. Stream open is now deferred until the first non-empty `push_text`.   
- **Server abort treated as success** (`a49aa9b9`). `terminated` from an abort looked identical to normal completion, so the turn ended with no audio. Now distinguished via `audio_end` + `cancel_sent`; aborts raise a retryable error.                                                          
                                                                                                                                                                                                                                                                                                   
### STT                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                   
- **WS drops not recovered** (`cd4108d5`). A dropped WebSocket left STT silent for the rest of the session. Plugin now reconnects automatically.                                                                                                                                                   
- **Wrong language label on mixed input** (`60d49fde`). The transcript's language was set from the first token, so an English transcript starting with `"哎"` was tagged `zh`. Now uses the language with the most characters across the transcript; ties go to whichever reached the count first.


tinalenguyen (approved):
tested it and it lgtm, thank you!

PR #5771, opened 2026-05-19
chore(worker): update worker warnings AGT-2909

chenghao-mou:


livekit-agents/livekit/agents/worker.py
```diff
@@ -1136,10 +1137,22 @@ async def _recv_task() -> None:
...
+                    )
 
                 if msg.type != aiohttp.WSMsgType.BINARY:
-                    logger.warning("unexpected message type: %s", msg.type)
+                    logger.warning(
+                        "unexpected message type: %s",
+                        msg.type,
+                        extra={
+                            "type": msg.type.name,
+                            "ws_extra": msg.extra,
```
theomonnom: I think printing the data may even be more spammy?

I would only print it for error messages

theomonnom (approved):

livekit-agents/livekit/agents/worker.py
chenghao-mou: Do we send any non-error text messages?

chenghao-mou: Truncated the message. Since we only expect binary messages, a text message is most likely an error message.

PR #5776, opened 2026-05-19
examples(inference): editable system prompt + "Open in Builder" + Sonic-3 / 4.1-mini defaults

JackNDwyer:
## Summary

Turns the **Inference** example into a lightweight agent-builder playground that pairs with the corresponding Jukebox PR (livekit/agents-jukebox#TODO).

- **Editable system prompt.** Adds a `set_system_prompt` RPC that rewrites `agent.instructions` live. Deduped against the current value so the per-keystroke fires from the frontend don't spam `update_instructions`.
- **"Open in Builder" CTA.** New `type: link` control in the YAML with an `href_template` that opens Cloud Builder pre-populated with the current `instructions` (URL-encoded), `llm`, `stt`, `tts`, and `modelMode=pipeline`. `project_id` is the `p_` placeholder so Cloud routes through login when needed and preserves the params on redirect.
- **Live transcript view.** Adds a `views[].source: transcript` entry; the Jukebox frontend renders LiveKit `TranscriptionReceived` segments into the existing view-card slot — no agent-side push needed.
- **Defaults bumped.** `cartesia/sonic-2` → `cartesia/sonic-3`, `openai/gpt-4o-mini` → `openai/gpt-4.1-mini`, both in `playground.yaml` and the agent's `DEFAULT_*` constants. Seed `INSTRUCTIONS` aligned with the textarea's `default` so the first session matches what the user sees before any edit.

YAML schema also picks up doc comments for the new `textarea` / `link` control types and the `transcript` view source — the Jukebox PR has the corresponding parser + renderer changes.

## Test plan

- [ ] `cd examples/inference && python agent.py dev` — connect via Jukebox playground; verify prompt edits land (`system prompt updated (N chars)` in agent logs) and the agent's behaviour changes within a turn.
- [ ] Pick a new STT / LLM / TTS — confirm the model swap still works after the prompt-editing changes.
- [ ] Click "Open in Builder" — confirm a Cloud Builder URL opens in a new tab with `instructions`, `llm`, `stt`, `tts`, and `modelMode=pipeline` matching the current Playground state.
- [ ] Default-only session (no edits) — verify the agent's initial prompt matches the textarea's seed.

### DIFF
--- a/examples/playground.yaml
+++ b/examples/playground.yaml
@@ -165,11 +165,29 @@ examples:
         011111111110
         000000000000
         000000000000
-    # The playground reads these `controls` and renders a labeled
-    # pill dropdown per control. Picking an option fires an RPC on
-    # the agent participant — `rpc` is the method name, payload is
-    # `{"value": "…"}`. Defaults match the agent's DEFAULT_* constants.
+    # The playground reads these `controls` and renders one widget per
+    # entry. `type` selects the widget — defaults to `dropdown` for
+    # backwards-compat with the other examples. Picking an option fires
+    # an RPC on the agent participant: `rpc` is the method name,
+    # payload is `{"value": "…"}`. Defaults match the agent's
+    # DEFAULT_* constants in agent.py.
+    #
+    # Supported types:
+    #   - dropdown (default): pill with options list
+    #   - textarea: multiline string editor. Fires `rpc` with the
+    #     current text on every edit-commit (debounced by the FE) so
+    #     the agent can apply it live. Value is also exposed for URL
+    #     template substitution.
+    #   - link: a button-styled link. On click opens `href_template`
+    #     in a new tab with `{slug/rpc}` tokens substituted from the
+    #     current control values (URL-encoded). No RPC fired.
     controls:
+      - rpc: set_system_prompt
+        type: textarea
+        label: System prompt
+        # Keep in sync with INSTRUCTIONS in examples/inference/agent.py.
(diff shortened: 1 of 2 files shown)


examples/playground.yaml
```diff
@@ -254,3 +272,30 @@ examples:
...
+      # project_id; if the user isn't logged in on cloud.livekit.io
+      # they'll be routed through login first and params will be
+      # preserved on the redirect. Param names match the verified
+      # Builder schema: `instructions` (not `prompt`), `llm`, `stt`,
+      # `tts`, `modelMode`.
+      - rpc: open_in_builder
+        type: link
+        label: "Open in Builder \u2192"
+        primary: true
+        href_template: "https://cloud.livekit.io/projects/p_/agents/builder/new?modelMode=pipeline&instructions={inference/set_system_prompt}&llm={inference/set_llm_model}&stt={inference/set_stt_model}&tts={inference/set_tts_model}"
```
theomonnom: Should this be an RPC? And it open the URL that the agent is returning? 

So you don't need any template

theomonnom (approved):

JackNDwyer:
Done in ac546378a (agents) + livekit/agents-jukebox#10 343c8a6 (jukebox).

`open_in_builder` is now an RPC on the agent — it reads its own `instructions` + the live `session.{stt,llm,tts}.model` and returns the Cloud Builder URL. The YAML loses `href_template` and the jukebox loses the whole template-substitution + URL-encoder block; `type: link` controls now just fire the RPC and open whatever URL the agent returns.

One bit of glue: `window.open()` only escapes the popup blocker when called synchronously inside a click handler, so the jukebox side cracks `about:blank` open on click and assigns `.location` once `performRpc` resolves (closing the blank tab on failure so the user isn't stranded). Net it's still a single click, just routed through the agent.

PR #5801, opened 2026-05-21
feat(inworld tts): add delivery_mode parameter for inworld-tts-2

chrisackermann:
## Summary

- Adds `delivery_mode` to the Inworld TTS plugin, exposing the TTS-2-only
  `deliveryMode` output-variation control. Accepts `"DELIVERY_MODE_UNSPECIFIED"`,
  `"STABLE"`, `"BALANCED"`, or `"CREATIVE"` per the [Inworld REST API reference](https://docs.inworld.ai/api-reference/ttsAPI/texttospeech/synthesize-speech).
- Wired through both transport paths: the WebSocket streaming `create` packet
  (`SynthesizeStream` via `_InworldConnection._send_loop`) and the HTTP
  streaming request body (`ChunkedStream._run`). Mirrors the existing
  `text_normalization` / `timestamp_type` plumbing.
- Surfaced on `TTS.__init__`, `TTS.update_options`, and `_TTSOptions`, with
  the same `_validate_str_param` validation used by the sibling enum params.

## Why

Inworld TTS-2 introduced `deliveryMode` and silently ignores `temperature`
on that model. Previously the plugin gave TTS-2 users no way to steer output
variation. Defaults are unchanged — `delivery_mode` is opt-in and absent from
the wire when unset, letting the server default (`BALANCED`) apply.

No client-side model-gating: Inworld silently ignores the wrong knob on each
model, and gating client-side would require tracking Inworld's per-model field
support over time. The docstring calls out the TTS-2-only semantics.

## Test plan

- [x] `uv run pytest tests/test_plugin_inworld_tts.py` — 9 new unit tests pass
  (option round-trip + validation for all four documented values + rejection
  of unknown values + wire-payload assertions for both WS and HTTP paths,
  including the negative case that `deliveryMode` is omitted when unset)
- [x] `make check` — ruff lint, ruff format, mypy clean
- [x] Live-tested against the real Inworld API end-to-end: rendered the same
  line with `delivery_mode="STABLE"`, `"BALANCED"`, and `"CREATIVE"` against
  `inworld-tts-2`. All three were accepted (no fallback to `EXPRESSIVE` needed —
  the REST docs are authoritative, the Inworld Python SDK docstring is stale)
  and produced audibly different cadence/expressiveness.


tinalenguyen (approved):
lgtm and tested it, thank you @chrisackermann!

PR #5812, opened 2026-05-22
feat(vad): set default vad in agent session

chenghao-mou:
`AgentSession` now provisions a bundled silero VAD automatically. Existing call-site behaviour is preserved via an `is_default` marker on the VAD instance — the framework-supplied default is transparent to every code path that previously cared about "user configured a VAD."

## Before

```python
from livekit.agents import AgentSession, inference

session = AgentSession(
    stt=inference.STT("deepgram/nova-3"),
    llm=inference.LLM("openai/gpt-4.1-mini"),
    tts=inference.TTS("cartesia/sonic-3"),
    vad=inference.VAD(model="silero"),  # required boilerplate
)
``` 

…or, in examples using the prewarm pattern:

```python
def prewarm(proc: JobProcess) -> None:
    proc.userdata["vad"] = inference.VAD(model="silero")

server.setup_fnc = prewarm

@server.rtc_session()
    session = AgentSession(
        # ...
        vad=ctx.proc.userdata["vad"],
    )
```

## After

```python
from livekit.agents import AgentSession, inference

session = AgentSession(
    stt=inference.STT("deepgram/nova-3"),
    llm=inference.LLM("openai/gpt-4.1-mini"),
    tts=inference.TTS("cartesia/sonic-3"),
    # vad= omitted — bundled silero is provided automatically
)

# Opt out explicitly if needed:
# AgentSession(..., vad=None)

# Customise by passing your own:
# AgentSession(..., vad=inference.VAD(model="silero", min_silence_duration=0.3))
```

## What changes (and what doesn't)

| Concern | Behaviour |
|---|---|
| `AgentSession()` with no `vad=` | Now gets `inference.VAD(model="silero")` with `is_default=True` |
| `AgentSession(vad=None)` | Unchanged — no VAD |
| `AgentSession(vad=my_vad)` | Unchanged — user-supplied VAD, `is_default=False` |
| `RealtimeModel` with server-side turn detection | Default VAD is constructed but **not** wired into the audio pipeline (saves CPU); server-side VAD remains canonical |
| `turn_detection="vad"` / `AudioTurnDetector` opt-ins | Now satisfied by the default — no more "VAD model not provided" warnings |
| Realtime fallback `mode = "vad"` when server-side turn detection is off | **No change** — only flips when the user explicitly supplied a VAD |
| Adaptive interruption detection eligibility | **No change** — opt-in stays opt-in; default VAD doesn't auto-enable it |
| STT hook `speaking=` payload type and `_last_speaking_time` source | **No change** for sessions that didn't configure a VAD — default VAD is treated as absent at these sites |

The marker is exposed as a read-only `vad.VAD.is_default` property on the base class, so plugin-supplied VADs inherit `False` automatically.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -198,6 +198,12 @@ class VoiceActivityVideoSampler:
 DEFAULT_TTS_TEXT_TRANSFORMS: list[TextTransforms] = ["filter_markdown", "filter_emoji"]
 
 
+def _build_default_vad() -> vad.VAD:
+    instance = inference.VAD(model="silero")
+    instance._is_default = True
+    return instance
+
+
 class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
     @deprecate_params(
         {
@@ -266,7 +272,10 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
 
         Args:
             stt (stt.STT | str, optional): Speech-to-text backend.
-            vad (vad.VAD, optional): Voice-activity detector
+            vad (vad.VAD, optional): Voice-activity detector. Defaults to the
+                bundled silero VAD (``inference.VAD(model="silero")``) when
+                omitted. Pass ``vad=None`` to opt out, or pass an explicit
+                instance to customise options.
             llm (llm.LLM | llm.RealtimeModel | str, optional): LLM or RealtimeModel
             tts (tts.TTS | str, optional): Text-to-speech engine.
             tools (list[llm.FunctionTool | llm.RawFunctionTool], optional): List of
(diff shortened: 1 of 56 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -198,6 +198,12 @@ def __call__(self, frame: rtc.VideoFrame, session: AgentSession) -> bool:
 DEFAULT_TTS_TEXT_TRANSFORMS: list[TextTransforms] = ["filter_markdown", "filter_emoji"]
 
 
+def _build_default_vad() -> vad.VAD:
+    instance = inference.VAD(model="silero")
+    instance._is_default = True
```
theomonnom, line 203: Should we create a separate field? Like one inside the session itself? self._using_default_vad = True

theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_session.py
chenghao-mou: I thought about that, but then we need to pass extra flag to activity for resolution or audio recognition like in `update_vad`.

PR #5818, opened 2026-05-22
fix(voice): skip commit_user_turn on close when no turn is pending

chenghao-mou:
- Add `AudioRecognition.has_pending_user_turn` and gate the session-close commit on it, avoiding spurious empty-turn commits and the resulting timeout warnings during teardown.
- Make the close-path commit best-effort: catch timeout / API / runtime errors so STT or EOU failures don't block a clean shutdown.
- Wrap `activity.drain()` in a non-cancelling watchdog that logs a snapshot of running tasks if drain stalls past 300s, for shutdown observability.
- Mark inference STT error 2007 (idle session closed by server) as non-retryable; reconnecting just produces the same error.
- Honor `APIError.retryable` in STT (`stt/stt.py`) and TTS (`tts/tts.py` chunked stream) retry loops, matching the existing pattern in `llm/llm.py` and `inference/interruption.py`.

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -260,6 +260,11 @@ STTEncoding = Literal["pcm_s16le"]
 DEFAULT_ENCODING: STTEncoding = "pcm_s16le"
 DEFAULT_SAMPLE_RATE: int = 16000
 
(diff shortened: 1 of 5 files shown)


chenghao-mou:
Going to create separate PRs for each of the issues.

PR #5827, opened 2026-05-24
feat(cartesia): add ink-2 stt

charlotte-zhuang:
Adds the Cartesia Ink 2 STT model.

This model has a different API from Ink Whisper and supports different features, most importantly turn detection.

A decision was made to use the same `cartesia.STT()` class for both the new and old API to avoid confusion as to which class to use.

## Changes

These changes could be considered breaking in a way, but they should not affect expected usage patterns:

1. Default model for English is now `ink-2`
  1. `START_OF_SPEECH`, `INTERIM_TRANSCRIPT`, `PREFLIGHT_TRANSCRIPT`, and `END_OF_SPEECH` events will be emitted
  2. Aligned transcripts are no longer supported
2. `update_options(model="...")` is now a no-op. There is no reason to update your model mid-session so I'm not sure why this was added in the first place. This should not negatively effect any users since before Ink 2, there was only a single Ink Whisper snapshot, making the method even more bewildering.
3. `SpeechStream` is now an `ABC` to allow for different behavior based on the model. It is highly unexpected for users to be constructing their own `cartesia.stt.SpeechStream` instances.
4. Ink Whisper checks `self._FlushSentinel` and sends `"finalize"` commands to the server

These changes are purely additive:

1. `AutoFinalizeRecognizeStream`, which is the default for `ink-2`
2. `cartesia.py` example agent
3. `AUDIO_ENCODING` constant
4. doc strings

## Testing

Tested using `cartesia.py` using both `ink-2` and `ink-whisper`.

Also verified that `make check` passes and existing tests pass.


charlotte-zhuang:
It looks like `Test STT / test-stt` is failing because:

1. Some errors with `livekit.plugins.aws` and `livekit.plugins.fireworksai`
2. `OPENAI_API_KEY` is not set
3. The GitHub action was unable to post a comment to this PR

```
FAILED tests/test_stt.py::test_recognize[livekit.plugins.openai] - openai.OpenAIError: Missing credentials. Please pass an `api_key`, `workload_identity`, `admin_api_key`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
FAILED tests/test_stt.py::test_stream[livekit.plugins.fireworksai] - RuntimeError: livekit.plugins.fireworksai.stt.SpeechStream is closed
FAILED tests/test_stt.py::test_stream[livekit.plugins.openai] - openai.OpenAIError: Missing credentials. Please pass an `api_key`, `workload_identity`, `admin_api_key`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
FAILED tests/test_stt.py::test_stream[livekit.plugins.aws] - smithy_core.exceptions.CallError: Unknown error for operation com.amazonaws.transcribestreaming#StartStreamTranscription - status: 403 - id: com.amazonaws.transcribestreaming#UnrecognizedClientException
```

```
POST /repos/livekit/agents/issues/5827/comments - 403 with id 8009:24BDD6:12551F54:4647D1C7:6A1310E4 in 157ms
RequestError [HttpError]: Resource not accessible by integration - https://docs.github.com/rest/issues/comments#create-an-issue-comment
```

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/stt.py
```diff
@@ -14,123 +14,198 @@
...
+        self._audio_chunk_duration_ms = audio_chunk_duration_ms
+        self._model = resolved_model
+        self._api_key = resolved_api_key
+        self._ws_base_url = _base_url_to_ws_base_url(base_url=base_url)
         self._session = http_session
-        self._streams = weakref.WeakSet[SpeechStream]()
+        self._streams = weakref.WeakSet[TurnsRecognizeStream | LegacyRecognizeStream]()
 
     @property
+    @override
```
tinalenguyen: hmm i don't think we need `@override`s for these

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/stt.py
```diff
@@ -164,292 +273,59 @@ def update_options(
...
-
-            # Forward audio to Cartesia in chunks
-            samples_50ms = self._opts.sample_rate // 20
-            audio_bstream = utils.audio.AudioByteStream(
-                sample_rate=self._opts.sample_rate,
-                num_channels=1,
-                samples_per_channel=samples_50ms,
-            )
+@dataclass
+class STTOptions:
```
tinalenguyen, line 335: do you foresee more parameters introduced in the future that would apply to both models? i'm a bit hesitant on deprecating this at the moment

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/models.py
```diff
@@ -76,56 +83,31 @@
...
+"""
+STT models that support turn detection.
+
+See [the docs](https://docs.cartesia.ai/build-with-cartesia/stt-models/latest) for all options.
+"""
+
+STTModels = Literal["ink-whisper"] | TurnDetectingSTTModel
+"""See [the docs](https://docs.cartesia.ai/build-with-cartesia/stt-models/latest) for all options."""
+
+STTLanguages = Literal["en"]
```
tinalenguyen: could we list all the languages supported (maybe separated by model) and other applicable options? users often reference the literal here

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/_legacy_recognize_stream.py
```diff
@@ -0,0 +1,445 @@
...
+    ) -> None:
+        """Change Cartesia STT options. Use sparingly; this will interrupt transcription.
+
+        Args:
+            language: Used to change the language to match what the user is speaking.
+            model: Deprecated. This is a no-op. Construct a new STT instance to change the model.
+        """
+        # not changing the model and reconnecting since that is likely unexpected behavior
+        if is_given(model) and model != self._model:
+            warnings.warn(
```
tinalenguyen: should we use `logger.warning` instead of importing warnings for consistent logs?

.github/next-release/changeset-cartesia-ink-2-stt.md
```diff
@@ -0,0 +1,15 @@
+---
+"livekit-plugins-cartesia": minor
```
tinalenguyen: changes in the python repo don't require a changeset, we can omit this

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/_recognize_streams/cartesia_recognize_stream.py
```diff
@@ -0,0 +1,20 @@
...
+
+from livekit.agents import (
+    stt,
+)
+from livekit.agents.types import NOT_GIVEN, NotGivenOr
+
+from .models import STTLanguages
+
+
+class CartesiaRecognizeStream(stt.RecognizeStream):
```
tinalenguyen, line 17: i wonder if it would be possible to condense this to something like:
```
CartesiaRecognizeStream: TypeAlias = TurnsRecognizeStream | LegacyRecognizeStream
```
what do you think of this?

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/stt.py
charlotte-zhuang: I actually deprecated `STTOptions` since this class appears to be for internal use only.

The one public use was in `SpeechStream.__init__()`, but realistically users get a `SpeechStream` instance by calling `STT.stream()`, not by constructing their own instances. That was why I made `SpeechStream` an ABC, which removed the need for `STTOptions`.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/stt.py
charlotte-zhuang: Oh, this is true.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/_legacy_recognize_stream.py
charlotte-zhuang: Sure! I just wasn't sure what the norm was in this repo.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/_recognize_streams/cartesia_recognize_stream.py
charlotte-zhuang: Do you know if users call methods on the return of `STT.stream()`? i.e. is it important that all my `RecognizeStream` implementations have the same interface?

I just wanted to avoid a situation where you would need to do something like this

```py
stt = cartesia.STT(model="ink-whisper")
stream = cartesia.stream()

cast(stream, Any).update_options(language="es")
```

So the `CartesiaRecognizeStream` ABC helps to make sure all our `RecognizeStream` implementations follow the same interface.

If I were to implement this for the first time, I would probably make our `SpeechStream.update_options` method private.

charlotte-zhuang:
@tinalenguyen I addressed all your comments!

I also ended up adding a third implementation of `CartesiaRecognizeStream`, which is why I ended up opting to keep it.

Copied from my PR description:

> ## final_transcript_mode
> 
> I also added a way to use both `ink-whisper` and `ink-2` in conjunction with `stream.flush()` and called it `final_transcript_mode="emit_on_flush"`. This will send a final transcript speech event only after `stream.flush()` is called by the user.
> 
> I'm not entirely sure how this works in LiveKit, but we have some users who want to tell the model when turns end rather than our model telling them when the turn ended. Those users will see increased latency and unexpected transcripts without the new `final_transcript_mode="emit_on_flush"` behavior I added.
> 
> The reason why I couldn't use the `LegacyRecognizeStream` is because that implementation streams out `FINAL_TRANSCRIPT` in chunks as soon as they become available. Then, LiveKit concatenates all `FINAL_TRANSCRIPT` speech events with spaces:
> 
> ```py
> self._audio_transcript += f" {transcript}"
> ```
> 
> This space concatenation can cause issues with the transcript by inserting spaces in unexpected places. My solution was to create a new opt-in behavior where `FINAL_TRANSCRIPT` is only emitted after `stream.flush()` is called and all audio up until that point is flushed. This way, the entire human turn is put in a single transcript and the space concatenation issue is avoided.
> 
> An alternative solution would be to add a flag to `SpeechEvent` like `transcript_event_sep: str = " "` so that plugins can specify how their transcripts should be concatenated. Then, `LegacyRecognizeStream` can just emit transcripts with `transcript_event_sep=""`.

tinalenguyen:
hi @charlotte-zhuang, thanks for iterating! i think it's rare for users to choose manually flushing for final transcripts vs opting for a server-side VAD. ideally we set up the plugin so that everything works right out the box (i do see that the default is server-side vad). to mirror inference, what do you think of only exposing stt/turns/websocket for now? if we receive requests for the manual flush mechanism in the future, i think it would be worth a discussion of restructuring our framework to better accommodate that

also for the concatenation of final transcripts, do you think stripping the received chunks of whitespaces would resolve the issue?

charlotte-zhuang:
@tinalenguyen I would be ok with leaving out the manual flush behavior for now!

> also for the concatenation of final transcripts, do you think stripping the received chunks of whitespaces would resolve the issue?

Unfortunately not since our model will emit final text chunks with words broken up between chunks. So the more grievous error is inserting spaces between words.

A hack could be to buffer the last word in text chunks, so the last word from chunk 0 gets emitted from chunk 1, etc.

This feels like over-engineering though and ideally LiveKit could have a way to avoid inserting spaces natively.

charlotte-zhuang:
For example (I made this one up):

`Hello, to`
`morrow is Monday.`

tinalenguyen (approved):
super small suggestion, do you think we should move the types to their corresponding stream module?

charlotte-zhuang:
@tinalenguyen sure, I moved the types into their corresponding stream modules!

PR #5830, opened 2026-05-24
fix(plugins/sarvam): thread language_probability into SpeechData.confidence

### LINKED ISSUE #5829: plugins/sarvam: STT hardcodes confidence=1.0; should thread Sarvam's language_probability into SpeechData
## Bug

`livekit-plugins-sarvam` returns `SpeechData.confidence=1.0` for every transcript, but Sarvam's Saaras API returns a real `language_probability` field per response. The field is dropped at two sites in `stt.py`:

- **REST batch path** — `livekit/plugins/sarvam/stt.py:701` (v1.5.12)
  ```python
  alternatives = [
      stt.SpeechData(
          language=detected_language,
          text=transcript_text,
          start_time=start_time,
          end_time=end_time,
          confidence=1.0,  # Sarvam doesn't provide confidence score in this response
      )
  ]
  ```
  The comment is inaccurate — `language_probability` is in the response body. See: https://docs.sarvam.ai/api-reference-docs/speech-to-text/transcribe (response shows `language_probability: 0.95`).

- **Streaming WS path** — `livekit/plugins/sarvam/stt.py:1475-1480` (v1.5.12) in `_handle_transcript_data`
  ```python
  speech_data = stt.SpeechData(
      language=language,
      text=transcript_text,
      start_time=transcript_data.get("speech_start", 0.0),
      end_time=transcript_data.get("speech_end", 0.0),
  )
  ```
  `confidence` is omitted entirely, so `SpeechData`'s default (1.0) takes over. The WS payload likewise carries `language_probability` per chunk.

## Why it matters

Real per-chunk confidence values enable downstream language stickiness / code-mix handling. Without them, weighted-voting buffers (we built one for a multilingual voice assistant) can't tell a confident Hindi turn from a noisy guess.

## Question before PR

@dhruvladia-sarvam — can you confirm `language_probability` is:

1. A stable field in Saaras v3 streaming responses (not just REST)?
2. Present on every transcript chunk, or omitted on some (e.g. final-vs-interim, noisy chunks)?
3. Bounded to `[0.0, 1.0]`, or could it ever fall outside?

This determines whether the PR should treat `language_probability` as required (assert) or optional with a fallback (defensive `isinstance` + 1.0 default).

## Proposed fix

Once 1–3 are answered, I'd like to open a PR that:

- Parses `language_probability` from the response/payload at both sites.
- Threads it into `SpeechData(confidence=...)`.
- Defaults to `1.0` if absent or wrong type (defensive, in case the field is best-effort).
- Adds a unit test using a mocked WS payload `{"language_probability": 0.87, ...}` asserting `SpeechData.confidence == 0.87`.

Patch is ~20 lines per site, mostly the isinstance-guard helper. Happy to draft once field stability is confirmed.

## Reproducer

```python
from livekit.plugins.sarvam import STT
stt = STT(model="saaras:v3", api_key="...")
event = await stt.recognize(buffer=...)
print(event.alternatives[0].confidence)  # always 1.0, regardless of input quality
```

hashirventhodi:
## Summary

Fixes #5829. The `livekit-plugins-sarvam` STT silently dropped Sarvam's `language_probability` and reported `confidence=1.0` everywhere. This wires the real value through both the REST and WS paths.

## Changes

- `livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py`:
  - REST `recognize()`: parse `language_probability` from response JSON; thread into `SpeechData(confidence=...)`.
  - WS `_handle_transcript_data()`: parse `language_probability` from `transcript_data`; thread into `SpeechData(confidence=...)`.
  - Both sites: `isinstance(_, (int, float))` guard with 1.0 fallback so callers don't crash on null/missing/string values. `logger.debug` on unexpected types so contract drift is visible.

- `livekit-plugins/livekit-plugins-sarvam/tests/test_language_probability.py` (new):
  - Asserts `SpeechData.confidence == 0.87` for a mocked WS payload containing `language_probability: 0.87`.
  - Asserts fallback to 1.0 when the field is absent, null, or has a wrong type.
  - REST test stubbed (mocks aiohttp session — same isinstance-guard logic as WS).

## Validation

Tested in a downstream voice assistant via a vendored fork carrying these exact patches: worker logs show `confidence` varying across utterances rather than the previous flat 1.0. The defensive fallback handles missing/null cases without crashing.

## Notes / Open question for review

- `language_probability` is documented for Sarvam's REST batch endpoint. **Open question (see #5829)**: is it guaranteed present on Saaras v3 WS chunks, or best-effort? The defensive `isinstance + 1.0 fallback` makes the PR safe under either answer, but if it's guaranteed, the fallback path becomes dead code we could tighten later.
- `@dhruvladia-sarvam` — would appreciate confirmation on field stability so we know whether to keep or remove the fallback in a follow-up.
- Defensive fallback to 1.0 preserves backward behaviour for callers that read `confidence` defensively.
- No public API changes.
- CLA: will be signed before merge (first-time contributor).

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
@@ -692,13 +692,32 @@ class STT(stt.STT):
                 if start_time == 0.0 and end_time == 0.0:
                     end_time = _calculate_audio_duration(buffer)
 
+                # Sarvam Saaras returns ``language_probability``
+                # on every response — thread it into SpeechData.confidence
+                # so downstream consumers
+                # see real confidence values instead of a hardcoded 1.0.
+                # Defensive: defaults to 1.0 if the field is absent or has
+                # an unexpected type (API contract drift detection).
+                _lang_prob = response_json.get("language_probability")
+                if isinstance(_lang_prob, (int, float)):
+                    _confidence = float(_lang_prob)
+                else:
+                    if _lang_prob is not None:
+                        logger.debug(
+                            "Unexpected language_probability type: %s (value=%r); "
+                            "falling back to confidence=1.0",
+                            type(_lang_prob).__name__,
+                            _lang_prob,
+                        )
+                    _confidence = 1.0
+
                 alternatives = [
                     stt.SpeechData(
                         language=detected_language,
                         text=transcript_text,
                         start_time=start_time,
                         end_time=end_time,
-                        confidence=1.0,  # Sarvam doesn't provide confidence score in this response
+                        confidence=_confidence,
                     )
                 ]
 
@@ -1471,12 +1490,33 @@ class SpeechStream(stt.SpeechStream):
             )
             self._event_ch.send_nowait(usage_event)
 
+            # Sarvam Saaras streaming WS sends ``language_probability``
+            # in the transcript_data payload — thread it into SpeechData.confidence
+            # so downstream consumers see
+            # real confidence values instead of the SpeechData default (1.0).
+            # Defensive: defaults to 1.0 if the field is absent or has an
+            # unexpected type (the field is documented for REST but not
+            # explicitly for streaming — API contract drift detection).
+            _lang_prob = transcript_data.get("language_probability")
+            if isinstance(_lang_prob, (int, float)):
+                _confidence = float(_lang_prob)
+            else:
+                if _lang_prob is not None:
+                    self._logger.debug(
+                        "Unexpected language_probability type: %s (value=%r); "
+                        "falling back to confidence=1.0",
+                        type(_lang_prob).__name__,
+                        _lang_prob,
+                    )
+                _confidence = 1.0
+
             # Create speech data
             speech_data = stt.SpeechData(
                 language=language,
                 text=transcript_text,
                 start_time=transcript_data.get("speech_start", 0.0),
                 end_time=transcript_data.get("speech_end", 0.0),
+                confidence=_confidence,
             )
 
             # Create final transcript event with request_id
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
@@ -703,7 +703,7 @@ class STT(stt.STT):
                     _confidence = float(_lang_prob)
                 else:
                     if _lang_prob is not None:
-                        logger.debug(
+                        self._logger.debug(
                             "Unexpected language_probability type: %s (value=%r); "
                             "falling back to confidence=1.0",
                             type(_lang_prob).__name__,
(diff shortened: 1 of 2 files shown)


hashirventhodi:
Thanks @devin-ai-integration — addressed all four findings in `eb7a39af`:

**🔴 P1 (tests):** Rewrote `test_language_probability.py` to:
- Target `SpeechStream._handle_transcript_data` (not `STT` — that was just wrong).
- Use `async def` so pytest-asyncio (mode=auto, already in `pyproject.toml`) awaits the call.
- Drop the bogus `request_id="test"` kwarg — method signature is `(self, data: dict)`.
- Use the real WS payload shape: outer `{"type": "data", "data": {...}}` with nested `transcript` / `language_code` / `language_probability` / `speech_start` / `speech_end` / `metrics` / `request_id`.

14 tests pass locally:
- 5 happy-path values (0.0, 0.123, 0.5, 0.87, 1.0) thread through to `SpeechData.confidence`.
- missing / `None` / wrong-type (str, list, dict, object) fall back to `1.0`.
- 3 out-of-range values pass through verbatim (not clamping at this layer).

**🟡 P2 (REST logger):** Switched `logger.debug` → `self._logger.debug` to match every other log in `_recognize_impl` and the equivalent WS site at `stt.py:1505`.

Apologies for the busted first cut — the original tests were written before I had access to read the real method signature and silently never executed. Fresh CI run should reflect the green local result.

livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
```diff
@@ -1471,12 +1490,33 @@ async def _handle_transcript_data(self, data: dict) -> None:
...
             self._event_ch.send_nowait(usage_event)
 
+            # Sarvam Saaras streaming WS sends ``language_probability``
+            # in the transcript_data payload — thread it into SpeechData.confidence
+            # so downstream consumers see
+            # real confidence values instead of the SpeechData default (1.0).
+            # Defensive: defaults to 1.0 if the field is absent or has an
+            # unexpected type (the field is documented for REST but not
+            # explicitly for streaming — API contract drift detection).
+            _lang_prob = transcript_data.get("language_probability")
```
davidzhao: let's refactor this code instead of duplicating it in both places.

hashirventhodi:
@davidzhao thanks — refactored in `3493a50b` (pushed). Both call sites now read `confidence=_extract_confidence(payload, self._logger)`; the parse + isinstance-guard + 1.0 fallback + drift-log live in one module-level helper. 14 tests still pass.

hashirventhodi:
@devin-ai-integration good catch — fixed in `0b1fe91b`. Guard now reads `isinstance(value, (int, float)) and not isinstance(value, bool)`, matching the `livekit-plugins-slng` pattern you cited. Added `True`/`False` to the parametrize'd bad-value test (16 cases pass).

davidzhao (approved):

dhruvladia-sarvam:
Thank you @davidzhao @hashirventhodi 
For your context, this was upcoming this week wrt to recent updates to our STT model, as language confidence is only populated with the relevant custom value when language code is `unknown`.

PR #5841, opened 2026-05-25
first-class async tools via ctx.update()

longcw:
## Overview

Async tools are now first-class. Any `@function_tool` can call `ctx.update()` to stream progress and run in the background — no special type hint needed. `AsyncRunContext` is now an alias of `RunContext`, and per-tool duplicate policy / cancellation are explicit flags on `@function_tool`.

## Summary

- `AsyncRunContext` is now a module-level alias of `RunContext`. Tools typed `ctx: AsyncRunContext` keep working unchanged.
- Per-tool flags replace type-hint-driven dispatch: `@function_tool(on_duplicate="allow|reject|replace|confirm", flag=llm.ToolFlag.CANCELLABLE)`.
- Every tool routes through `_ToolExecutor` (`voice/tool_executor.py`). The default executor is owned by `AgentActivity`. Each `AsyncToolset` owns its own executor scoped to that toolset — when placed under `AgentSession(tools=[...])` its in-flight tools and deferred replies survive agent handoff; under `Agent(tools=[...])` they're bound to that agent.
- `lk_agents_cancel_task` / `lk_agents_get_running_tasks` companion tools are auto-exposed by `AgentActivity.tools` whenever any registered tool declares `allow_cancellation=True` (always-on, so the LLM-visible schema stays stable across turns and the prompt cache stays warm).
- Async-tool prompt templates are configured via `tool_handling=` (a `ToolHandlingOptions`) on `AgentSession`, `Agent`, or `AsyncToolset` (most specific wins). Its `async_options` field (`AsyncToolOptions`) holds the per-stage templates (`update_template`, `duplicate_reject_template`, `duplicate_confirm_template`, `reply_at_tail_template`, `reply_maybe_covered_template`), each either a `str.format()` template or a callable receiving typed `UpdatePromptArgs` / `DuplicatePromptArgs` / `ReplyPromptArgs`.

## Deprecations

- `AsyncToolset(on_duplicate_call=...)` logs a warning at construction and is applied as a fallback override onto contained tools' `on_duplicate`. Prefer the per-tool `@function_tool(on_duplicate=...)` flag.


livekit-agents/livekit/agents/llm/async_toolset.py
```diff
@@ -1,425 +1,101 @@
...
-    exe_task: asyncio.Task[Any]
-
-
-@dataclass
-class _PendingUpdate:
-    ctx: AsyncRunContext
-    items: list[ChatItem]
+# AsyncRunContext is a backwards-compat alias — the unified RunContext now carries
+# the same surface, so user tools typing `ctx: AsyncRunContext` keep working
+__all__ = ["AsyncRunContext", "AsyncToolset"]
```
theomonnom, line 23: Since AsyncToolset was in beta, I think we can safely remove AsyncRunContext?

livekit-agents/livekit/agents/voice/tool_executor.py
```diff
@@ -0,0 +1,564 @@
...
+
+    async def cancel(self, call_id: str) -> bool:
+        task = self._running_tasks.get(call_id)
+        if task is None:
+            return False
+        if not task.ctx.speech_handle.allow_interruptions:
+            raise ToolError(
+                f"Tool call {call_id} is not cancellable because interruptions are disallowed"
+            )
+        await utils.aio.cancel_and_wait(task.exe_task)
```
theomonnom, line 384: Should we cancel the asyncio.Task? Or just mark the SpeechHandle as cancelled/interrupted? 

My main concern is that most users don’t really understand asyncio cancellation logic. If we can hide that complexity from users, that would be great. (e.g not having to worry about all your method to be cancellation-safe)

livekit-agents/livekit/agents/llm/async_toolset.py
```diff
@@ -1,425 +1,101 @@
...
-    """A toolset for running long-running functions in the background.
+    """Session-scoped toolset whose tools survive agent handoff.
 
-    Tools with an :class:`AsyncRunContext` parameter are wrapped to run in the background.
-    Each ``ctx.update()`` and the final ``return`` inject a tool output into the conversation;
-    the agent then generates a natural-language reply to the user based on that output.
+    Background updates from tools in this toolset are delivered to whichever agent
+    is current at delivery time, so a ``ctx.update()`` started under agent A still
+    completes after a handoff to agent B. Tools placed on ``Agent(tools=...)``
+    instead use the activity-scoped executor and are cancelled/awaited on handoff.
```
theomonnom, line 43: 🙌

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -143,11 +154,16 @@ class ToolFlag(Flag):
...
+DuplicateMode = Literal["allow", "reject", "replace", "confirm"]
+
+
 @dataclass
 class FunctionToolInfo:
     name: str
     description: str | None
     flags: ToolFlag
+    on_duplicate: DuplicateMode = "allow"
+    allow_cancellation: bool = False
```
theomonnom: Maybe we add a ToolFlag.CANCELLABLE ?

livekit-agents/livekit/agents/voice/tool_executor.py
longcw: we have `allow_cancellation` default with False, I think user should be aware of it if they enable it specifically. it could be useful to stop or revert some write operations like updating the db or ordering something, etc...

livekit-agents/livekit/agents/llm/tool_context.py
longcw: sounds good!

livekit-agents/livekit/agents/llm/async_toolset.py
longcw: maybe log a deprecation warning and notify them to use RunContext directly when it's used

livekit-agents/livekit/agents/voice/generation.py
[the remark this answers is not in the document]
longcw: not related to this pr

livekit-agents/livekit/agents/voice/tool_executor.py
theomonnom: But should the user catch `asyncio.CancelledError`vs having something like `if speech_handle.cancelled: ...`?

longcw: perhaps something like `ctx.wait_if_not_cancelled` and `ctx.cancelled` since the speech handle is done after the first update. we have something similar using `speech_handle.wait_if_not_interrupted` in https://github.com/livekit/agents/blob/e3e992f9591445ab92740cb8631adccbdf08889c/examples/voice_agents/long_running_function.py#L44-L51

livekit-agents/livekit/agents/voice/tool_executor.py
```diff
@@ -0,0 +1,571 @@
...
+    """``fnc_calls_json`` joined by newlines — what the default string templates use."""
+
+
+class ReplyPromptArgs(TypedDict):
+    """Args for the ``reply_at_tail`` / ``reply_maybe_covered`` templates."""
+
+    call_ids: list[str]
+
+
+class AsyncToolPrompts(TypedDict, total=False):
```
theomonnom: Can we follow this similar pattern I did for expressiveness mode?  Like the "template" approach<br><br>[https://github.com/livekit/agents/blob/4962032dd6cfecd652e143f90f0b8a4440a21502/livekit-agents/livekit/agents/voice/agent_session.py#L135](<https://github.com/livekit/agents/blob/4962032dd6cfecd652e143f90f0b8a4440a21502/livekit-agents/livekit/agents/voice/agent_session.py#L135>)

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -225,6 +226,7 @@ def __init__(
         turn_handling: NotGivenOr[TurnHandlingOptions] = NOT_GIVEN,
         # Tool settings
         tools: NotGivenOr[list[llm.Tool | llm.Toolset]] = NOT_GIVEN,
+        async_tool_prompts: NotGivenOr[AsyncToolPrompts] = NOT_GIVEN,
```
theomonnom: Should we find something more future-proof? <br><br>like

```python
tool_handling={
    async_options={
        update_template=..
    } 
}
```

livekit-agents/livekit/agents/voice/tool_executor.py
```diff
@@ -0,0 +1,571 @@
...
+async def cancel_task(call_id: str) -> str:
+    """Cancel a running tool call by call_id."""
+    job_ctx = get_job_context(required=False)
+    task = _RunningTasks.get((job_ctx, call_id))
+    if task is None:
+        raise ToolError(f"Task {call_id} not found")
+
+    if not await task.executor.cancel(call_id):
+        raise ToolError(f"Task {call_id} not found or already completed")
+    return f"Task {call_id} cancelled successfully."
```
theomonnom: Should we scope all running tasks to an `AgentSession`?

With text mode, we'll allow creating a new `AgentSession` from within a `function_tool`, enabling reasoning agentic loops with slower LLMs. Ideally they don't capture unrelated tasks

theomonnom (commented):
My last main concern is that it is kinda hard to understand the difference between interrupted and cancelled for function tools.<br><br>`ctx.wait_if_not_cancelled`<br>`ctx.wait_if_not_interrupted`

Any idea on how we can improve that?

livekit-agents/livekit/agents/voice/tool_executor.py
longcw: done

livekit-agents/livekit/agents/voice/agent_session.py
longcw: done

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -141,13 +152,22 @@ def __init__(self) -> None:
...
 @dataclass
 class FunctionToolInfo:
     name: str
     description: str | None
     flags: ToolFlag
+    on_duplicate: DuplicateMode = "allow"
+
+    @property
+    def allow_cancellation(self) -> bool:
+        return ToolFlag.CANCELLABLE in self.flags
```
theomonnom: Do we need this utility? could be removed IMO

livekit-agents/livekit/agents/voice/tool_executor.py
```diff
@@ -0,0 +1,600 @@
...
+# used when the pending update is the most recent item in chat_ctx — the agent
+# can't have already talked about it.
+REPLY_INSTRUCTIONS_AT_TAIL = """New results arrived from background tool calls (call_ids: {call_ids}).
+Summarize the results naturally. Do NOT repeat information you have already told the user."""
+
+# used when newer items have been appended after the pending update — the agent
+# may have already verbalized the result in its most recent turn.
+REPLY_INSTRUCTIONS_MAYBE_COVERED = """New results arrived from background tool calls (call_ids: {call_ids}).
+Review your most recent assistant messages and check what you have already conveyed:
+- If they already cover all of these results, return an empty response (no text at all).
```
theomonnom: I'm not sure if lot of models will correctly handle this instruction
```suggestion
- If they already cover all of these results, return an empty response (no text at all).
```

livekit-agents/livekit/agents/voice/tool_executor.py
```diff
@@ -0,0 +1,600 @@
...
+async def cancel_task(call_id: str) -> str:
+    """Cancel a running tool call by call_id."""
+    job_ctx = get_job_context(required=False)
+    task = _RunningTasks.get((job_ctx, call_id))
+    if task is None:
+        raise ToolError(f"Task {call_id} not found")
+
+    if not await task.executor.cancel(call_id):
+        raise ToolError(f"Task {call_id} not found or already completed")
+    return f"Task {call_id} cancelled successfully."
```
theomonnom:
```python
async def get_running_tasks() -> list[dict]:
async def cancel_task(call_id: str) -> str:
```

Should we make those tools more "hidden".

Like we prefix them with lk_?

theomonnom (approved):
lgtm! nice work!

PR #5845, opened 2026-05-26
fix(elevenlabs): always close TTS stream context on cancellation

### LINKED ISSUE #5844: ElevenLabs - websocket connection closed unexpectedly (1008) on frequent interruptions
### Bug Description

I wrote a load testing script that had frequent interruptions, but after a few interruptions the ElevenLabs websocket closed with `livekit.agents._exceptions.APIStatusError: message='ElevenLabs websocket connection closed unexpectedly', status_code=1008, retryable=True`.

After digging into this some more, it looks like it's caused when a single ElevenLabs websocket is used for more than 5 contexts (couldn't find any official docs, but there are a few related pipecat issues: https://github.com/pipecat-ai/pipecat/issues/4027, https://github.com/pipecat-ai/pipecat/issues/4132, https://github.com/pipecat-ai/pipecat/issues/4041). Looks like the root cause is that open contexts aren't closed after an interruption.

### Expected Behavior

Shouldn't crash

### Reproduction Steps

https://gist.github.com/superMDguy/609e34a246d7a2aecc5ef0102468ebea

Disclaimer: written by an LLM, but I've verified it follows the same pattern as the official `tts_node`

### Operating System

macOS Sequoia 15.7.4

### Models Used

ElevenLabs Flash v2.5

### Package Versions

```bash
livekit                                  1.1.8
livekit-agents                           1.5.10
livekit-api                              1.0.7
livekit-blingfire                        1.1.0
livekit-plugins-elevenlabs               1.5.10
livekit-protocol                         1.1.8
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

Again, this comes from an LLM, but it makes sense to me:

In `SynthesizeStream._run()`, call `connection.close_context()` in the `finally` block, **after** `gracefully_cancel(input_t, stream_t)`:

```python
# tts.py (AFTER FIX)
finally:
    output_emitter.end_segment()
    await utils.aio.gracefully_cancel(input_t, stream_t)
    await sent_tokenizer_stream.aclose()
    # Queue close_context AFTER cancelling tasks: all pending
    # _SynthesizeContent messages are already in the queue, so
    # _send_loop will drain them then send this close packet while
    # the context is still in _active_contexts.
    if self._context_id and connection is not None:
        try:
            connection.close_context(self._context_id)
        except Exception:
            pass
```

### Additional Context

_No response_

_No response_

longcw:
close https://github.com/livekit/agents/issues/5844

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import asyncio
 import base64
+import contextlib
 import dataclasses
 import json
 import os
@@ -89,7 +90,7 @@ class PronunciationDictionaryLocator:
     version_id: str
 
 
-DEFAULT_VOICE_ID = "l7kNoIfnJKPg7779LI2t"
+DEFAULT_VOICE_ID = "hpp4J3VqNfWAUOO0d1Us"
 API_BASE_URL_V1 = "https://api.elevenlabs.io/v1"
 AUTHORIZATION_HEADER = "xi-api-key"
 WS_INACTIVITY_TIMEOUT = 180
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
lgtm

PR #5848, opened 2026-05-26
fix(amd): defer no-speech timer until SIP call is answered

chenghao-mou:
1. **Timeouts**

Previously AMD started all its timers as soon as the SIP audio track was subscribed. Because audio tracks can be published during ringing or carrier early media (before SIP ANSWER), this poisoned the classifier with pre-answer audio and burned the no-speech budget. 

This PR correctly waits for SIP active state to start the no-speech timer. Detection timeout is still armed when track is subscribed.

Also adds a reusable wait_for_participant_attribute helper in utils.participant with a dedicated `ParticipantAttributeWaitAborted` exception, and tracks/cleans up the deferred setup task properly.

2. **No speech timeout**

Now it maps to `uncertain`;

3. **EoT**

Previously, verdict is emitted when the prediction arrives and if the silence threshold is satisfied, but it doesn't mean the user turn is ended. A new option `wait_until_finished` (default False) now make sure we wait for both EOT and silence threshold for machines. 

This helps when previously, new normal generation can be triggered in parallel with a generate_reply call after AMD has interrupted any preemptive generations, emitted the verdict and released the hold for playout (the new generation is too late for AMD to interrupt, due to late STT for example).

If EOT is properly being waited, in addition to the silence threshold, then the parallel generation will be interrupted when `interrupt_on_machine=True`. 

This flag still respects the overall `timeout` value if there is no speech or transcript arrives.


livekit-agents/livekit/agents/utils/participant.py
```diff
@@ -68,6 +68,76 @@ def on_connection_state_changed(state: int) -> None:
         room.off("connection_state_changed", on_connection_state_changed)
 
 
+class ParticipantAttributeWaitAborted(RuntimeError):
+    """Raised by :func:`wait_for_participant_attribute` when the wait cannot
+    complete (room/participant disconnected or never present)."""
```
theomonnom: Should we just use RuntimeError?

theomonnom (approved):

PR #5863, opened 2026-05-27
fix(avatar): preserve audio wrappers across avatar hot-swaps

### LINKED ISSUE #4198: VideoAvatar handling does not allow per-agent session customization
### Bug Description

The VideoAvatar logic assumes that a single AgentSession is tied to one video avatar. This design prevents assigning separate avatars to different agents within the same session, even though logically it should be possible.

As a result:

- We cannot have multiple agents within the same session each using a distinct video avatar.
- The current implementation restricts flexibility, forcing one avatar per session instead of allowing per-agent avatar configuration.

### Expected Behavior

- Each agent within a session should be able to have its own video avatar.
- The system should still support shared session data while allowing per-agent avatar customization.

### Reproduction Steps

```bash
async def start(
        self,
        agent_session: AgentSession,
```

### Operating System

macOS 

### Models Used

_No response_

### Package Versions

```bash
livekit==1.0.20
livekit-agents==1.3.6
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

longcw:
## Summary

Avatar plugins set `session.output.audio = DataStreamAudioOutput(...)` on every start. On the first start this works because `AgentSession.start()` wraps the sink with the `TranscriptSynchronizer` / `RecorderAudioOutput` chain afterwards; on a mid-session rebind (avatar switch) the raw assignment blows the chain away, silently breaking transcription sync and recording.

Fix it by auto-inserting an `_AudioSinkProxy` at the bottom of any wrapper chain. Wrappers cache the proxy, the proxy holds the swappable leaf — so hot-swaps preserve the wrappers above. New `AgentOutput.swap_audio_endpoint(sink)` walks the chain to the proxy and swaps its downstream in place, leaving the wrappers attached; full replacement stays as `output.audio = sink`.

## Plugin migration

All 13 avatar plugins migrated to `swap_audio_endpoint(...)`: anam, avatario, avatartalk, bey, bithuman, did, keyframe, lemonslice, liveavatar, runway, simli, tavus, trugen.

## Example

`examples/avatar_agents/audio_wave` now demonstrates hot-swapping through a `swap_avatar` RPC method: it tears down the current avatar (removing it from the room) and launches a fresh one under the same identity, while the audio wrappers and the listeners attached to `session.output.audio` survive the swap.

## Related

Clean avatar-worker shutdown on swap depends on livekit/python-sdks#699.

close https://github.com/livekit/agents/issues/4198


livekit-agents/livekit/agents/voice/io.py
```diff
@@ -275,6 +287,87 @@ def __repr__(self) -> str:
...
+        super().flush()
+        if self.next_in_chain:
+            self.next_in_chain.flush()
+        else:
+            # no real sink; synthesize a playback_finished
+            self.on_playback_finished(playback_position=0.0, interrupted=True)
+
+    def clear_buffer(self) -> None:
+        if self.next_in_chain:
+            self.next_in_chain.clear_buffer()
```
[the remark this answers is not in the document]
longcw: report on flush is enough

livekit-agents/livekit/agents/voice/io.py
```diff
@@ -568,6 +661,27 @@ def audio(self, sink: AudioOutput | None) -> None:
             else:
                 self._audio_sink.on_detached()
 
+    def set_audio_sink(self, sink: AudioOutput | None, *, preserve_wrappers: bool = False) -> None:
```
theomonnom: Any way to avoid this method?

Ideally the audio.setter alone is sufficient 

```
    @audio.setter
    def audio(self, sink: AudioOutput | None) -> None:
```

longcw: we can make `preserve_wrappers` default for setter but then there is no way to fully replace it.

livekit-agents/livekit/agents/voice/io.py
```diff
@@ -568,6 +674,24 @@ def audio(self, sink: AudioOutput | None) -> None:
             else:
                 self._audio_sink.on_detached()
 
+    def swap_audio_endpoint(self, sink: AudioOutput) -> None:
```
theomonnom:
```suggestion
    def replace_audio_leadl(self, sink: AudioOutput) -> None:
```
not a fan of the name

theomonnom (approved):

livekit-agents/livekit/agents/voice/io.py
davidzhao: should it be something like `replace_audio_sink`?

longcw: "lead" reads as the head of the chain, but the method replaces the tail, how about
+ replace_audio_endpoint
+ replace_audio_destination
+ redirect_audio
+ replace_audio_tail

PR #5864, opened 2026-05-27
fix(soniox): surface STT server errors

### LINKED ISSUE #5854: soniox: SpeechStream silently swallows server-side errors (error_code/error_message), breaking FallbackAdapter recovery
## Bug

`livekit-plugins-soniox`'s `SpeechStream._recv_messages_task` swallows server-side errors that arrive as JSON payloads with `error_code` / `error_message`.

When Soniox returns a server-side error (e.g. 401 *"Incorrect API key provided"*, region mismatch, plan-limit rejection), the receive loop:

1. Calls `logger.error(f"WebSocket error: ...")`
2. Continues iterating

No exception is raised, so:

- `stt.FallbackAdapter` never sees a failure signal → can't flip to the next provider.
- The stream sits silently with no events.
- livekit-agents' 60 s "process is unresponsive" watchdog eventually SIGKILLs the worker.
- The in-flight call dies — the patient hears nothing after the welcome line.

Source (1.5.11): https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py#L549-L553

## Reproducer

Wire Soniox + a fallback STT, then point Soniox at credentials it will reject (e.g. US-region key against the EU `base_url`, or simply a bad key). The fallback should kick in; instead, the worker dies.

\`\`\`python
from livekit.agents import AgentSession, stt
from livekit.plugins import azure, soniox

session = AgentSession(
    stt=stt.FallbackAdapter(
        stt=[
            soniox.STT(api_key=\"invalid-or-wrong-region-key\", base_url=..., params=...),
            azure.STT(...),  # perfectly valid fallback
        ],
        attempt_timeout=3.0,
        max_retry_per_stt=1,
    ),
    # ...
)
\`\`\`

**Expected:** Soniox's failure surfaces as a Python exception from the stream → FallbackAdapter marks Soniox unhealthy → traffic routes to Azure → call completes.

**Actual:** Soniox logs the error and otherwise sits silent. The agent worker is SIGKILL'd by the 60 s unresponsive watchdog. The patient hears `Welcome ...` and then silence.

Confirmed in two real environments (Azure Container Apps + local dev) on `livekit-plugins-soniox 1.5.11` / `livekit-agents 1.5.11`.

## Suggested fix

In `SpeechStream._recv_messages_task`, raise `APIStatusError` after the `logger.error(...)` for the error_code/error_message branch:

\`\`\`python
if content.get(\"error_code\") or content.get(\"error_message\"):
    err_code = content.get(\"error_code\")
    err_msg = content.get(\"error_message\")
    logger.error(f\"WebSocket error: {err_code} - {err_msg}\")
    raise APIStatusError(
        message=f\"Soniox STT server error: {err_code} - {err_msg}\",
        status_code=int(err_code) if isinstance(err_code, int) else 500,
        request_id=None,
        body=msg.data,
    )
\`\`\`

…and either remove the bare \`except Exception\` at the outer level of the method, or make it re-raise \`APIError\` subclasses, so the exception escapes the method into \`_run\`'s \`asyncio.gather\` → \`task.result()\` chain (where the FallbackAdapter is awaiting).

The TTS plugin already follows this shape (raises on server errors); STT is the asymmetric one.

## Workaround

For anyone hitting this before the fix lands, we monkey-patch \`_recv_messages_task\` at plugin import time to raise on \`error_code\`. The patch is gated to known-good plugin versions so a future refactor skips it with a warning rather than break silently. Happy to share the snippet.

## Versions

- \`livekit-plugins-soniox==1.5.11\`
- \`livekit-agents==1.5.11\`
- Python 3.13.13

## PR

Happy to send a PR with the two-line raise + the broad-except removal if useful.

he-yufeng:
﻿Fixes #5854

## Summary
- turn Soniox STT `error_code` / `error_message` frames into `APIStatusError`
- preserve the existing final transcript / usage flush before surfacing the error
- let `APIError` propagate through the receive loop and `_run` instead of being swallowed by broad exception handlers
- add a unit test for an error frame without `tokens`

## Tests
- `python -m pytest tests\test_plugin_soniox_stt.py -q -p no:cacheprovider`
- `python -m ruff check livekit-plugins\livekit-plugins-soniox\livekit\plugins\soniox\stt.py tests\test_plugin_soniox_stt.py`
- `python -m ruff format --check livekit-plugins\livekit-plugins-soniox\livekit\plugins\soniox\stt.py tests\test_plugin_soniox_stt.py`
- `python -m py_compile livekit-plugins\livekit-plugins-soniox\livekit\plugins\soniox\stt.py tests\test_plugin_soniox_stt.py`
- `git diff --check`

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
@@ -27,6 +27,7 @@ from livekit import rtc
 from livekit.agents import (
     APIConnectionError,
     APIConnectOptions,
+    APIError,
     APIStatusError,
     APITimeoutError,
     LanguageCode,
@@ -335,6 +336,8 @@ class SpeechStream(stt.SpeechStream):
                     tasks_group.cancel()
                     tasks_group.exception()
 
+            except APIError:
+                raise
             except asyncio.TimeoutError as e:
                 logger.error(
                     f"Timeout during Soniox Speech-to-Text API connection/initialization: {e}"
@@ -486,7 +489,8 @@ class SpeechStream(stt.SpeechStream):
 
                 try:
                     content = json.loads(msg.data)
-                    tokens = content["tokens"]
+                    has_error = bool(content.get("error_code") or content.get("error_message"))
+                    tokens = content.get("tokens", []) if has_error else content["tokens"]
 
                     non_final = _TokenAccumulator()
                     non_final_original = _TokenAccumulator()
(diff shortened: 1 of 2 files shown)


he-yufeng:
I checked the failing `test-stt` job. The Soniox cases from this PR are skipped in that run because `SONIOX_API_KEY` is empty; the actual failures are in other hosted providers (`openai`, `fireworksai`, `aws`) plus the fork-comment `403` while posting the generated test summary.

The PR-specific local checks from the branch are still the focused Soniox regression tests plus repo lint/type checks. I don't see evidence in the CI log that the Soniox error-frame change caused the current hosted-provider failures.

tinalenguyen (approved):
thanks you for the PR!

PR #5865, opened 2026-05-27
fix(llm): serialize all provider tools for the Responses API + log server-side tool execution

u9g:
## Summary

While wiring the basic voice agent up to xAI (Grok) with its `WebSearch` / `XSearch` provider tools, the tools never reached the API — Grok answered from memory instead of searching. This PR fixes the root cause and adds visibility into server-side tool execution.

### `fix(llm): let the caller specify the provider tool type for responses`
`to_responses_fnc_ctx` hardcoded `openai.tools.OpenAITool`, so the xAI LLM (which reuses openai's Responses serializer) had its `WebSearch` / `XSearch` tools silently dropped — they subclass `XAITool`, not `OpenAITool`, and never reached the API.

- The caller now passes its plugin's provider-tool type via `provider_tool_type`, removing the core→plugin import of `livekit.plugins.openai`. The openai Responses LLM defaults it to `OpenAITool`; the xAI LLM overrides the `_provider_tool_type` class attribute to `XAITool`.
- Adds a shared `DictProviderTool` base (a `ProviderTool` declaring `to_dict()`). The dict-based plugin tool bases (`OpenAITool`, `XAITool`, `AnthropicTool`, `MistralTool`) now extend it, so the `to_dict` contract is declared once instead of in every plugin, and the serializer is typed without a Protocol or `type: ignore`. (Gemini keeps `ProviderTool` + `to_tool_config`, since its schema is a typed `types.Tool`.)

### `feat(openai): log server-side provider tool execution in responses LLM`
Emits an info log when the Responses API runs a tool server-side, e.g. xAI `web_search` and `x_search` (which decomposes into `custom_tool_call` subcalls like `x_keyword_search`), including the issued query and result. Detection is grounded in openai's `ResponseOutputItem` discriminated union (the authoritative parser): only `message`, `reasoning`, `function_call`, and `function_call_output` are produced/consumed by the agent itself — every other union member is a server-side tool, so the denylist is provably complete.

## Test plan
- Verified with the basic voice agent on `xai.responses.LLM`: provider tools now reach the API, searches execute, and `provider tool executed` logs appear on completion.
- `make lint` passes on touched files; runtime check confirms all four dict-based plugin tools subclass `DictProviderTool` and serialize correctly.

### DIFF
--- a/livekit-agents/livekit/agents/llm/_provider_format/openai.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/openai.py
@@ -222,6 +222,7 @@ def to_responses_fnc_ctx(
     tool_ctx: llm.ToolContext,
     *,
     strict: bool = True,
+    provider_tool_type: type[llm.ProviderTool],
 ) -> list[dict[str, Any]]:
     schemas: list[dict[str, Any]] = []
     for tool in tool_ctx.flatten():
@@ -232,7 +233,7 @@ def to_responses_fnc_ctx(
         elif isinstance(tool, llm.FunctionTool):
             schema = llm.utils.build_legacy_openai_schema(tool, internally_tagged=True)
             schemas.append(schema)
-        elif isinstance(tool, llm.ProviderTool) and hasattr(tool, "to_dict"):
+        elif isinstance(tool, provider_tool_type) and hasattr(tool, "to_dict"):
             schemas.append(tool.to_dict())
 
     return schemas
--- a/livekit-agents/livekit/agents/llm/_provider_format/openai.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/openai.py
@@ -218,9 +218,14 @@ def to_fnc_ctx(tool_ctx: llm.ToolContext, *, strict: bool = True) -> list[dict[s
     return schemas
 
 
-def to_responses_fnc_ctx(tool_ctx: llm.ToolContext, *, strict: bool = True) -> list[dict[str, Any]]:
-    from livekit.plugins import openai
-
+def to_responses_fnc_ctx(
+    tool_ctx: llm.ToolContext,
+    *,
+    strict: bool = True,
+    provider_tool_type: type[llm.DictProviderTool],
+) -> list[dict[str, Any]]:
+    # the caller passes its plugin's ProviderTool subclass so this core helper doesn't import any
+    # plugin to recognize server-side provider tools (web_search, x_search, ...).
     schemas: list[dict[str, Any]] = []
     for tool in tool_ctx.flatten():
         if isinstance(tool, llm.RawFunctionTool):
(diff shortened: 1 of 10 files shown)


livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -48,6 +48,13 @@ def id(self) -> str:
         return self._id
 
 
+class DictProviderTool(ProviderTool):
+    """A provider tool whose schema serializes to a plain dict (e.g. openai, xAI)."""
+
+    @abstractmethod
+    def to_dict(self) -> dict[str, Any]: ...
```
theomonnom: It's tricky to have an abstract class for that, because it really depends on the provider. 

Is there a way to avoid it?

livekit-agents/livekit/agents/llm/_provider_format/openai.py
```diff
@@ -230,7 +235,7 @@ def to_responses_fnc_ctx(tool_ctx: llm.ToolContext, *, strict: bool = True) -> l
         elif isinstance(tool, llm.FunctionTool):
             schema = llm.utils.build_legacy_openai_schema(tool, internally_tagged=True)
             schemas.append(schema)
-        elif isinstance(tool, openai.tools.OpenAITool):
+        elif isinstance(tool, provider_tool_type):
```
theomonnom: Maybe we should use hasattr? 

I'm not sure, but the reasoning is that for some providers, they only require the API name to mention the tool. It’s not even a dict; it’s literally just "web_search"."

u9g: What if we called it JSONProviderTool, then it encompasses just strings too.

Alternatively, we can just use `has_attr`, it's not a big deal, but it means we can't specify the type in the function argument and we can't have tools that extend the YProviderTool.

u9g: I removed DictProviderTool and replaced it with `hasattr(tool, "to_dict")`

livekit-agents/livekit/agents/llm/tool_context.py
u9g: done.

theomonnom (approved):

livekit-agents/livekit/agents/llm/tool_context.py
```diff
@@ -503,7 +503,16 @@ def copy(self) -> ToolContext:
...
+        self, format: Literal["openai"], *, strict: bool = True
+    ) -> list[dict[str, Any]]: ...
+
+    @overload
+    def parse_function_tools(
+        self,
+        format: Literal["openai.responses"],
+        *,
+        strict: bool = True,
+        provider_tool_type: type[ProviderTool],
```
theomonnom, line 515: This is unused?

u9g: No, this adds the `provider_tool_type` type argument so we can check is_instance against either `OpenAIProviderTool` or `XAIProviderTool` so we can avoid adding provider tools from other providers, before the is_instance hardcoded the `OpenAIProviderTool` type which meant `XAIProviderTools` were never sent.

PR #5869, opened 2026-05-27
(core): add function_tools_called event

tinalenguyen:


jeanprbt:
From what I understand, the new `FunctionToolsCalledEvent` contains both provider tools & locally-executed tools. Is there a specific reason why you want to group both, and not only include provider tools?

Also, is the `FunctionToolsCalledEvent` containing both the arguments & results of the called provider tool?

In my sense, it would be even more handy to have a `ProviderToolsExecutedEvent` instead of a `FunctionToolsCalledEvent`, which only contains provider tools, and contain both arguments and results for each provider tool. What do you think about this?

In any case, thanks a lot for addressing the provider tools events problem! 🙌

Issue #5868, opened 2026-05-27
Bug: jobs stall indefinitely when process initialization times out (ProcPool._warmed_proc_queue.get() has no timeout)


onur-yildirim-infinitusai:
## Summary

When worker processes fail to initialize (e.g. `initialize_process_timeout` is exceeded), jobs dispatched to the pool stall **indefinitely** and never emit the expected \"failed to launch job after N attempts\" error. The 3-retry logic in `launch_job` is unreachable in this failure mode.

## Affected versions

Confirmed on `livekit-agents==1.5.11` (likely affects all versions).

## How to reproduce

1. Start a worker on a resource-constrained host where cold module imports take longer than `initialize_process_timeout` (default: 10 s).
2. On Linux, the SDK uses `forkserver` by default, so child processes must re-import application code from scratch — heavy import trees can exceed 10 s under CPU contention.
3. Dispatch a job while all pre-warm slots are failing.
4. Observe: the job logs `no warmed process available for job, waiting for one to be created` and then **hangs forever**. No further logs appear for that job.

## Root cause

`ProcPool.launch_job` waits for an initialized process via:

```python
proc = await self._warmed_proc_queue.get()   # line ~133 in proc_pool.py
```

There is **no timeout** on this `get()`. When `_proc_spawn_task` fails (process init timeout), it cleans up and returns without ever adding anything to `_warmed_proc_queue`:

```python
# _proc_spawn_task failure path
if not initialized:
    self._executors.remove(proc)
    await proc.aclose()
    self.emit("process_closed", proc)
    return   # ← silently returns, _warmed_proc_queue never receives an entry
```

The `launch_job` coroutine is now stuck in `_warmed_proc_queue.get()` indefinitely. The `finally` block that decrements `_jobs_waiting_for_process` cannot run until `get()` returns, so the counter stays elevated and no new spawn is triggered by subsequent calls. The 3-retry loop that emits `"failed to launch job on process after N attempts"` is **after** the `get()` call and is therefore never reached.

## Observed log sequence (truncated)

```
ERROR  livekit.agents  error initializing process   # TimeoutError in supervised_proc.initialize()
ERROR  livekit.agents  error initializing process   # same for every pre-warm slot
WARNING livekit.agents  no warmed process available for job, waiting for one to be created
# ... silence forever, job is never served and never errors out
```

## Expected behavior

[...]

PR #5871, opened 2026-05-28
fix: unblock jobs after process init failure

he-yufeng:
﻿## Summary
- wake jobs waiting on `ProcPool` when a process fails initialization
- retry the existing 3-attempt launch path instead of waiting forever on an empty warmed-process queue
- cover the failure path with a fake executor test that does not depend on OS process spawning

## To verify
- `python -m py_compile livekit-agents\livekit\agents\ipc\proc_pool.py tests\test_ipc.py`
- `uv run ruff check livekit-agents\livekit\agents\ipc\proc_pool.py tests\test_ipc.py`
- `uv run pytest tests/test_ipc.py::test_proc_pool_launch_job_returns_when_initialization_fails -q --basetemp .tmp\pytest-5868-fast -p no:cacheprovider`

Related: #5868


longcw:
thanks for the pr! closing in favor of https://github.com/livekit/agents/pull/5874 which takes a slightly different approach

Issue #5849, opened 2026-05-26
ElevenLabs server_vad with turn_detection="stt" does not end user turns


IngLP:
### Bug Description

When using `livekit-plugins-elevenlabs` with ElevenLabs Scribe v2 Realtime and `server_vad`, `AgentSession(turn_detection="stt")` receives partial and final transcripts, but the user turn is not committed in time because no useful `END_OF_SPEECH` event is emitted.

In a real call, the ElevenLabs STT stream produced multiple partial/final transcripts for short user utterances, for example:

```text
Can you hear me?
Hello?
Are you receiving audio?
```

However, the agent did not respond. The transcript only appeared as a `user_turn` when the participant disconnected, and LiveKit then logged:

```text
skipping user input, speech scheduling is paused
```

This looks like a plugin/turn-detection integration issue rather than an STT recognition issue: Scribe did hear the user, but the turn stayed open until shutdown.

Related but not exactly the same as #4087. That issue describes Scribe v2 turns committing very late with local/Silero VAD. This issue is specifically about the new ElevenLabs `server_vad` path with `AgentSession(turn_detection="stt")`.

### Expected Behavior

With ElevenLabs `server_vad` enabled and `AgentSession(turn_detection="stt")`, the server-side VAD endpoint should cause LiveKit Agents to commit the user turn promptly, so the LLM/agent can answer as soon as ElevenLabs decides the utterance ended.

### Actual Behavior

The ElevenLabs plugin emits partial/interim and committed/final transcript events, but the user turn is not committed promptly. The agent keeps waiting and only sees the accumulated user transcript during shutdown/disconnect.

### Likely Cause

From `livekit.plugins.elevenlabs.stt` in `livekit-plugins-elevenlabs==1.5.12`:

- `_connect_ws()` uses `commit_strategy = "vad"` when `server_vad` is configured.
- `_process_stream_event()` maps non-empty `committed_transcript` / `committed_transcript_with_timestamps` to `SpeechEventType.FINAL_TRANSCRIPT` and keeps `_speaking = True`.
- `SpeechEventType.END_OF_SPEECH` is only emitted for an empty committed transcript.

With `AgentSession(turn_detection="stt")` and no local VAD, `audio_recognition.py` appears to rely on an STT `END_OF_SPEECH` event to mark `_user_turn_committed=True`. If ElevenLabs does not send an empty committed transcript after a server-VAD commit, LiveKit never gets the end-of-speech signal and the turn remains open.

### Reproduction Steps

[...]

PR #5872, opened 2026-05-28
fix(elevenlabs): end server vad turns

he-yufeng:
﻿## Summary
- map ElevenLabs server-VAD committed transcripts to LiveKit `END_OF_SPEECH`
- leave manual commit behavior unchanged: manual streams still wait for an empty commit before EOS
- avoid treating an omitted `server_vad` option as server VAD when building the realtime websocket URL

## To verify
- `python -m py_compile livekit-plugins\livekit-plugins-elevenlabs\livekit\plugins\elevenlabs\stt.py tests\test_plugin_elevenlabs_stt.py`
- `uv run ruff check livekit-plugins\livekit-plugins-elevenlabs\livekit\plugins\elevenlabs\stt.py tests\test_plugin_elevenlabs_stt.py`
- `uv run pytest tests/test_plugin_elevenlabs_stt.py -q --basetemp .tmp\pytest-5849 -p no:cacheprovider`

Related: #5849

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
@@ -350,6 +350,9 @@ class SpeechStream(stt.SpeechStream):
         )
         self._event_ch.send_nowait(usage_event)
 
+    def _uses_server_vad(self) -> bool:
+        return is_given(self._opts.server_vad) and self._opts.server_vad is not None
+
     async def _run(self) -> None:
         """Run the streaming transcription session"""
         closing_ws = False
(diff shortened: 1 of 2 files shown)


toubatbrian (approved):

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -350,6 +350,9 @@ def _on_audio_duration_report(self, duration: float) -> None:
         )
         self._event_ch.send_nowait(usage_event)
 
+    def _uses_server_vad(self) -> bool:
```
chenghao-mou: we can use a property here:

```python
@property
def _server_vad(self) -> VADOptions | None:
    return self._opts.server_vad if is_given(self._opts.server_vad) else None

```

so the later checks become  either `if (server_vad := self._server_vad) is not None` or `if self._server_vad is not None` without any assertions. (is not None is needed here to allow empty dict)

chenghao-mou: updated it in commit: 399e8666113512da45be8ba5727955d8adcdbc52

chenghao-mou (approved):

PR #5874, opened 2026-05-28
fix(ipc): retry proc acquisition when all in-flight spawns fail

### LINKED ISSUE #5868: Bug: jobs stall indefinitely when process initialization times out (ProcPool._warmed_proc_queue.get() has no timeout)

longcw:
## Summary

Fixes #5868. When every worker process fails to initialize, `launch_job` previously hung forever on `_warmed_proc_queue.get()` because nothing was ever put on the queue, and the 3-attempt retry loop was unreachable.

Split the responsibility: `_acquire_proc` now owns the wait-for-a-warmed-process loop. It races `queue.get()` against every in-flight spawn task and only retries once all in-flight spawns settle without producing a proc — so peer spawns still in flight don't burn a retry attempt. After `MAX_ACQUIRE_ATTEMPTS` such cycles, it raises a `RuntimeError`. `launch_job` keeps its own 3-attempt budget for post-acquire launch failures (the original retry semantics, untouched).

## Alternative considered

#5871 uses `None` sentinels on `_warmed_proc_queue` to unblock waiters when a spawn fails. Two issues with that approach:
1. It wakes too eagerly — a sentinel is pushed on each individual spawn failure, even though peer spawns may still succeed, burning a retry attempt.
2. The retry doesn't trigger a fresh spawn — on loop-back the queue isn't empty (it has more `None` sentinels), so `launch_job` consumes another sentinel without ever requesting a new process. `MAX_ATTEMPTS=3` becomes "fail 3 sentinels and give up".

### DIFF
--- a/tests/test_ipc.py
+++ b/tests/test_ipc.py
@@ -14,6 +14,7 @@ from multiprocessing.context import BaseContext
 from typing import ClassVar
 
 import psutil
+import pytest
 
 from livekit.agents import JobContext, JobProcess, ipc, job, utils
 from livekit.agents.ipc.log_queue import LogQueueHandler, LogQueueListener
@@ -368,6 +369,56 @@ async def test_slow_initialization():
 
     for exitcode in exitcodes:
         assert exitcode != 0, "process should have been killed"
+
+
+async def test_proc_pool_launch_job_raises_when_all_spawns_fail(monkeypatch):
+    """When every spawn task fails to initialize, launch_job should raise
+    instead of hanging on an empty warmed-process queue. Reproduces #5868."""
+
+    class FailingProc:
+        running_job = None
+
+        def __init__(self, **kwargs):
+            pass
+
+        async def start(self) -> None:
+            pass
+
+        async def initialize(self) -> None:
+            raise TimeoutError("init timed out")
+
+        async def aclose(self) -> None:
+            pass
+
+        def logging_extra(self) -> dict:
+            return {}
+
+    monkeypatch.setattr(ipc.proc_pool.job_proc_executor, "ProcJobExecutor", FailingProc)
+
+    mp_ctx = mp.get_context("spawn")
+    loop = asyncio.get_running_loop()
+    pool = ipc.proc_pool.ProcPool(
+        job_executor_type=job.JobExecutorType.PROCESS,
+        initialize_process_fnc=_initialize_proc,
+        job_entrypoint_fnc=_job_entrypoint,
+        session_end_fnc=None,
+        num_idle_processes=0,
+        initialize_timeout=0.1,
+        close_timeout=20.0,
+        session_end_timeout=300.0,
+        inference_executor=None,
+        memory_warn_mb=0,
+        memory_limit_mb=0,
+        http_proxy=None,
+        mp_ctx=mp_ctx,
+        loop=loop,
+    )
+
+    with pytest.raises(RuntimeError, match="no process became available"):
+        await pool.launch_job(_generate_fake_job())
+
+    assert pool._jobs_waiting_for_process == 0
+    assert len(pool.processes) == 0
 
 
 def _create_proc(
--- a/tests/test_ipc.py
+++ b/tests/test_ipc.py
@@ -413,12 +413,16 @@ async def test_proc_pool_launch_job_raises_when_all_spawns_fail(monkeypatch):
         mp_ctx=mp_ctx,
         loop=loop,
     )
+    await pool.start()
 
-    with pytest.raises(RuntimeError, match="no process became available"):
-        await pool.launch_job(_generate_fake_job())
+    try:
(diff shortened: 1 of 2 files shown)


tests/test_ipc.py
```diff
@@ -370,6 +371,56 @@ def _process_closed(proc: ipc.job_proc_executor.ProcJobExecutor):
...
+        session_end_timeout=300.0,
+        inference_executor=None,
+        memory_warn_mb=0,
+        memory_limit_mb=0,
+        http_proxy=None,
+        mp_ctx=mp_ctx,
+        loop=loop,
+    )
+
+    with pytest.raises(RuntimeError, match="no process became available"):
```
chenghao-mou: pool.start() seems missing

chenghao-mou (approved):
LGTM. One small nit.

PR #5881, opened 2026-05-28
fix(speechmatics): avoid fake confidence for segments

### LINKED ISSUE #5880: Speechmatics STT plugin always reports confidence=1.0 instead of provider confidence
### Bug Description

The Speechmatics STT plugin currently appears to hardcode transcript confidence to 1.0, so LiveKit Sessions always shows Speechmatics transcripts as 100% confident.

This makes confidence-based logic unusable, for example asking the user to repeat when ASR confidence is low.

It looks like this behavior may have been introduced in PR [#4703](https://github.com/livekit/agents/pull/4703):

Before that PR, the plugin seems to have used the reported confidence value instead of hardcoding 1.0. I do not see any explanation in the PR description/comments about why confidence was changed.

Questions:

Is hardcoding confidence=1.0 intentional?
Does the newer Speechmatics integration expose real provider confidence somewhere?
If real confidence is not available, would it be better to leave confidence unset/default instead of reporting 1.0?
Would you accept a PR that propagates Speechmatics confidence when available?

### Expected Behavior

Speechmatics confidence should be propagated when available, or the plugin should avoid reporting 1.0 as if it were real confidence.

### Reproduction Steps

```bash
1. Create a LiveKit agent using the Speechmatics STT plugin.
2. Log the transcription confidence from user transcript events.
3. Observe the logged confidence value or check LiveKit Sessions.
```

### Operating System

Cloud

### Models Used

_No response_

### Package Versions

```bash
livekit-agents=1.5
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

he-yufeng:
﻿## Summary
- stop reporting Speechmatics segment confidence as `1.0` when the segment does not include a confidence value
- propagate numeric segment confidence when present, with a fragment-confidence fallback for richer segment payloads
- add a small regression test around the emitted `SpeechData.confidence`

Fixes #5880

## To verify
- `python -m py_compile livekit-plugins\livekit-plugins-speechmatics\livekit\plugins\speechmatics\stt.py livekit-plugins\livekit-plugins-speechmatics\tests\test_stt_confidence.py`
- `python -m ruff check livekit-plugins\livekit-plugins-speechmatics\livekit\plugins\speechmatics\stt.py livekit-plugins\livekit-plugins-speechmatics\tests\test_stt_confidence.py`
- `python -m ruff format --check livekit-plugins\livekit-plugins-speechmatics\livekit\plugins\speechmatics\stt.py livekit-plugins\livekit-plugins-speechmatics\tests\test_stt_confidence.py`
- `python -m pytest livekit-plugins\livekit-plugins-speechmatics\tests -q`

### DIFF
--- a/livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
+++ b/livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
@@ -904,6 +904,7 @@ class SpeechStream(stt.RecognizeStream):
                 start_time=segment.get("metadata", {}).get("start_time", 0)
                 + self.start_time_offset,
                 end_time=segment.get("metadata", {}).get("end_time", 0) + self.start_time_offset,
+                confidence=_extract_confidence(segment),
             )
 
             # Create speech event
@@ -941,6 +942,27 @@ class SpeechStream(stt.RecognizeStream):
         # Remove from active streams
         if self in self._stt._streams:
             self._stt._streams.remove(self)
+
+
+def _extract_confidence(segment: dict[str, Any]) -> float:
+    value = segment.get("confidence")
+    if isinstance(value, (int, float)) and not isinstance(value, bool):
+        return float(value)
+
+    confidences: list[float] = []
+    for fragment in segment.get("fragments") or []:
+        if not isinstance(fragment, dict):
+            continue
+        fragment_confidence = fragment.get("confidence")
+        if isinstance(fragment_confidence, (int, float)) and not isinstance(
+            fragment_confidence, bool
+        ):
+            confidences.append(float(fragment_confidence))
+
+    if confidences:
+        return sum(confidences) / len(confidences)
+
+    return 0.0
 
 
 def _check_deprecated_args(kwargs: dict[str, Any], opts: STTOptions) -> None:
--- a/livekit-plugins/livekit-plugins-speechmatics/tests/test_stt_confidence.py
+++ b/livekit-plugins/livekit-plugins-speechmatics/tests/test_stt_confidence.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+
+from types import SimpleNamespace
+from typing import Any
+
+import pytest
+
+from livekit.agents import LanguageCode, stt
+from livekit.plugins.speechmatics.stt import SpeechStream, STTOptions, _extract_confidence
+
+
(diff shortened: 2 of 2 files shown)


he-yufeng:
Rebased this on the current `main` branch. The only conflict was the Speechmatics segment conversion line, and the branch still uses the reported segment/fragment confidence instead of defaulting to `1.0`.

Validated locally:

- `uv run --project . --frozen pytest tests/test_stt_confidence.py` from `livekit-plugins/livekit-plugins-speechmatics` — 3 passed
- `uv run --project . --frozen ruff check livekit/plugins/speechmatics/stt.py tests/test_stt_confidence.py`
- `uv run --project . --frozen ruff format --check livekit/plugins/speechmatics/stt.py tests/test_stt_confidence.py`
- `python -m py_compile livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py livekit-plugins/livekit-plugins-speechmatics/tests/test_stt_confidence.py`
- `git diff --cached --check`

I could not run the root `uv sync --all-extras --dev` / `scripts/check_types.py` path on Windows because the current lock resolves `bithuman==2.3.3`, which has no win_amd64 wheel. The previous Linux CI type-check path should not hit that Windows-only blocker.

he-yufeng:
Second and last nudge on this. The behavior is unchanged on current main: Speechmatics segments that carry no confidence value are reported downstream as `1.0`, which reads as "maximally certain" to anything filtering on confidence — the opposite of what "unknown" should mean. The patch stops fabricating the value (falls back to fragment confidence when the richer payload has it) and adds a regression test around `SpeechData.confidence`.

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -943,6 +944,27 @@ async def aclose(self) -> None:
             self._stt._streams.remove(self)
 
 
+def _extract_confidence(segment: dict[str, Any]) -> float:
+    value = segment.get("confidence")
```
longcw, line 948: the `segment` is from `SegmentMessage.to_dict` produced by speechmatics SDK, it's not raw server json and has no confidence field. so this is identical to current main (which already defaults to 0.0 after #5942)

longcw:
closing this now,  we can reopen it after speechmatics SDK exposes the confidence in the segment.

PR #5883, opened 2026-05-28
fix(inference): enable STT websocket heartbeat

### LINKED ISSUE #5717: fix(inference): add per-message recv timeout to inference STT WebSocket
## Problem

`livekit/agents/inference/stt.py`'s `recv_task` calls `await ws.receive()` with no timeout (line 381). If the gateway WebSocket connects successfully but the upstream STT provider (Deepgram, etc.) silently stops returning transcripts — e.g., the gateway-to-provider tunnel goes flaky, or TCP goes half-open — the recv loop will block indefinitely. No exception is raised, `stt.FallbackAdapter` never sees a failure signal, and the agent can sit unresponsive until OS-level TCP keepalive eventually surfaces a `ConnectionError` (which can take minutes).

## Inconsistency with TTS

`livekit/agents/inference/tts.py` already does this correctly — its `_recv_task` uses `await ws.receive(timeout=self._conn_options.timeout)` (line 430). On a stalled WS, this raises `asyncio.TimeoutError` which is mapped to `APITimeoutError` by the outer try/except, allowing `FallbackAdapter` to switch providers.

The STT recv path looks like the same pattern was intended but was never applied. From a quick read I don't see a protocol reason it would differ.

## Proposed fix

Mirror the TTS implementation:

```python
# inference/stt.py recv_task
msg = await ws.receive(timeout=self._conn_options.timeout)
```

Plus matching outer-handler additions if needed to map `asyncio.TimeoutError` → `APITimeoutError` consistent with how connection-time timeouts are handled at line 475.

## Why this matters

For users running with `stt.FallbackAdapter`, this is the single most likely silent-failure mode — the adapter only switches on raised exceptions, and right now a degraded streaming session simply doesn't raise. We hit it while evaluating fallback strategies for a production voice agent and it took a while to figure out why fallback wasn't engaging.

Happy to put up a PR if the approach sounds right.

he-yufeng:
﻿## Summary
- enable aiohttp WebSocket heartbeat for LiveKit Inference STT streams
- keep normal quiet-user sessions untouched because heartbeat uses WS PING/PONG, not application message timeouts
- cover the connection path with a fake-session unit test

Fixes #5717.

## To verify
- `python -m py_compile livekit-agents\livekit\agents\inference\stt.py tests\test_inference_stt_fallback.py`
- `uv run ruff check livekit-agents\livekit\agents\inference\stt.py tests\test_inference_stt_fallback.py`
- `uv run ruff format --check livekit-agents\livekit\agents\inference\stt.py tests\test_inference_stt_fallback.py`
- `uv run pytest tests\test_inference_stt_fallback.py -q`

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -259,6 +259,7 @@ STTEncoding = Literal["pcm_s16le"]
 
 DEFAULT_ENCODING: STTEncoding = "pcm_s16le"
 DEFAULT_SAMPLE_RATE: int = 16000
+WS_HEARTBEAT: float = 30.0
 
 
 @dataclass
@@ -843,7 +844,9 @@ class SpeechStream(stt.SpeechStream):
         try:
             ws = await asyncio.wait_for(
                 http_session.ws_connect(
-                    f"{base_url}/stt?model={self._opts.model}", headers=headers
+                    f"{base_url}/stt?model={self._opts.model}",
+                    headers=headers,
+                    heartbeat=WS_HEARTBEAT,
                 ),
                 self._conn_options.timeout,
             )
--- a/tests/test_inference_stt_fallback.py
+++ b/tests/test_inference_stt_fallback.py
@@ -1,8 +1,12 @@
+from unittest.mock import AsyncMock, MagicMock, patch
+
 import pytest
 
 from livekit.agents.inference.stt import (
     STT,
+    WS_HEARTBEAT,
     FallbackModel,
+    SpeechStream,
     _normalize_fallback,
     _parse_model_string,
 )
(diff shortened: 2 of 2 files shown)


he-yufeng:
Gentle bump. Long quiet STT sessions still drop on idle websocket timeouts. This enables aiohttp PING/PONG heartbeat only for the inference STT streams, so normal quiet-user sessions are completely untouched (no application-level pings). Happy to make it configurable if you'd rather gate it.

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -843,7 +844,9 @@ async def _connect_ws(
         try:
             ws = await asyncio.wait_for(
                 http_session.ws_connect(
-                    f"{base_url}/stt?model={self._opts.model}", headers=headers
+                    f"{base_url}/stt?model={self._opts.model}",
+                    headers=headers,
+                    heartbeat=WS_HEARTBEAT,
```
longcw, line 849: this only check the connection between client to the gateway, but cannot fix the issue described in the original issue 
> If the gateway WebSocket connects successfully but the upstream STT provider (Deepgram, etc.) silently stops returning transcripts — e.g., the gateway-to-provider tunnel goes flaky, or TCP goes half-open — the recv loop will block indefinitely.

he-yufeng:
longcw's review pointed out the heartbeat covers client↔gateway, not the gateway→provider silence from the original issue; I have not reworked it in 59 days. Closing until a real fix exists.

PR #5887, opened 2026-05-29
fix(openai): delete trimmed empty audio items

### LINKED ISSUE #5886: RealtimeSession.update_chat_ctx silently skips delete for audio items when input_audio_transcription=None → unbounded server context growth
### Bug Description

When `openai.realtime.RealtimeModel` is constructed with `input_audio_transcription=None`, server-side conversation context grows unbounded across the session, even when the application calls `Agent.update_chat_ctx(new_ctx)` every turn with a trimmed context.

`RealtimeSession._create_update_chat_ctx_events` (`livekit/plugins/openai/realtime/realtime_model.py:1337-1391`) computes a diff and for each `to_remove` id runs:

```python
def _is_content_empty(msg_id: str) -> bool:
    remote_item = remote_ctx.get_by_id(msg_id)
    if remote_item and remote_item.type == "message" and not remote_item.content:
        return True
    return False

for msg_id in diff_ops.to_remove:
    if _is_content_empty(msg_id):
        continue          # delete skipped
    _delete_item(msg_id)
```

With `input_audio_transcription=None`, the server emits `conversation.item.added` with no transcript, and `openai_item_to_livekit_item` (`livekit/plugins/openai/realtime/utils.py:272-284`) only appends `uc.transcript` when it is not `None`:

```python
elif uc.type == "input_audio" and uc.transcript is not None:
    content.append(uc.transcript)
```

Every user audio item therefore has empty `content`, every delete is skipped, and the server-side conversation grows by one item per user turn for the entire session.

Per-`response.create` input processing scales with retained items, so reply latency drifts linearly with turn count.

### Expected Behavior

`update_chat_ctx(new_ctx)` is an explicit instruction. Items the caller excluded from `new_ctx` should be deleted server-side regardless of whether the SDK has client-side text content for them. The "user can't have meant to delete it" heuristic should not silently override an explicit `to_remove` set.

Either:
- Always emit `conversation.item.delete` for `diff_ops.to_remove`, OR
- Expose `force_delete: bool = False` on `update_chat_ctx` to bypass the empty-content filter, OR
- Only apply the empty-content filter when the SDK is uncertain about the source of the diff (e.g. external mutations), not when the caller passed an explicit ctx.

### Reproduction Steps

```python
import asyncio
from livekit.agents import Agent, AgentSession, JobContext, llm


class Bug(Agent):
    def __init__(self):
        super().__init__(instructions="Speak briefly.")


async def my_agent(ctx: JobContext) -> None:
    session = AgentSession(
        llm=openai.realtime.RealtimeModel(
            model="gpt-realtime",
            voice="shimmer",
            turn_detection=None,
            input_audio_transcription=None,   # repro
        ),
    )
    await session.start(agent=Bug(), room=ctx.room)
    await ctx.connect()

    # Simulate 30 user audio turns. After each, trim to last 24 items and
    # assert server-side state shrinks. It does not.
    for turn in range(30):
        # ... pretend the user spoke and we committed audio ...
        await asyncio.sleep(1)
        items = list(session.current_agent.chat_ctx.items)
        if len(items) > 24:
            kept = items[-24:]
            new_ctx = llm.ChatContext()
            new_ctx.items.extend(kept)
            await session.current_agent.update_chat_ctx(new_ctx)

        activity = session.current_agent._get_activity_or_raise()
        remote_size = len(activity._rt_session._remote_chat_ctx.to_chat_ctx().items)
        print(f"turn={turn} local={len(items)} remote={remote_size}")
```

Expected: `remote` stays bounded at ~24.
Observed: `remote` grows by ~2 per turn (user audio item + assistant message); user audio items never delete.

Measurable side-effect (production): time from `generate_reply()` to first audio frame from server drifts linearly with turn count. Observed 650 ms on turn 1 → ~5000 ms by turn 50 in a real deployment with 50+ turn sessions.

### Operating System

Linux (GKE us-central1, Python 3.11)

### Models Used

OpenAI Realtime `gpt-realtime`, external STT (Soniox) — reason `input_audio_transcription` is set to `None`, because we already have an authoritative transcript from another pipeline.

### Package Versions

```bash
livekit-agents==1.5.4
livekit-plugins-openai==<bundled>
openai==<bundled>
```

### Session/Room/Call IDs

Internal — happy to share via private channel if useful.

### Proposed Solution

```python
# realtime_model.py:1370-1376
# Option A (simplest): always delete what the caller put in to_remove.
for msg_id in diff_ops.to_remove:
    _delete_item(msg_id)

# Option B (preserve heuristic, add escape hatch):
async def update_chat_ctx(self, ctx, *, force_delete: bool = False) -> None:
    ...
for msg_id in diff_ops.to_remove:
    if not force_delete and _is_content_empty(msg_id):
        continue
    _delete_item(msg_id)
```

If the empty-content guard exists for a real scenario (server-only items the SDK didn't insert), please document it — then at least applications can call `force_delete=True` when they're confident the diff is authoritative.

### Additional Context

Workaround for affected applications: set `input_audio_transcription` to a cheap model (e.g. `gpt-4o-mini-transcribe`). The transcription is wasted work for applications with their own STT, but it makes `_is_content_empty` return `False` so the existing delete path fires.

The silent-skip combined with no warning log makes this very hard to diagnose without latency profiling across a long-running session. Adding even a single `logger.debug("update_chat_ctx: skipped delete for empty-content item %s", msg_id)` would have made root-cause obvious.

he-yufeng:
## Summary
- delete items explicitly trimmed out by `update_chat_ctx()`, even if the local copy has empty content
- keep the empty-content guard for updates so unsynced audio/image items are not recreated
- add a unit test for empty audio-like remote items being deleted from the OpenAI realtime context diff

Fixes #5886.

## To verify
- `uv run pytest tests/test_realtime/test_openai_realtime_model.py -q`
- `uv run ruff check livekit-plugins\livekit-plugins-openai\livekit\plugins\openai\realtime\realtime_model.py tests\test_realtime\test_openai_realtime_model.py`
- `uv run ruff format --check livekit-plugins\livekit-plugins-openai\livekit\plugins\openai\realtime\realtime_model.py tests\test_realtime\test_openai_realtime_model.py`
- `uv run python -m py_compile livekit-plugins\livekit-plugins-openai\livekit\plugins\openai\realtime\realtime_model.py tests\test_realtime\test_openai_realtime_model.py`
- `git diff --check`

## Notes
- `uv run python scripts\check_types.py` currently fails on Windows in `livekit-agents\livekit\agents\cli\readchar.py` because mypy sees POSIX `termios`/`tty` attributes as missing. The errors are outside this diff.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1364,11 +1364,6 @@ class RealtimeSession(
             return False
 
         for msg_id in diff_ops.to_remove:
-            # we don't have content synced down for some types of content (audio/images)
-            # these won't be present in the Agent's view of the context
-            # so in those cases, we do not want to remove them from the server context
-            if _is_content_empty(msg_id):
-                continue
             _delete_item(msg_id)
 
         for previous_msg_id, msg_id in diff_ops.to_create:
@@ -1376,7 +1371,7 @@ class RealtimeSession(
 
         # update the items with the same id but different content
         for previous_msg_id, msg_id in diff_ops.to_update:
-            # likewise, empty content almost always means the content is not synced down
+            # empty content almost always means the content is not synced down
             # we don't want to recreate these items there
             if _is_content_empty(msg_id):
                 continue
--- a/tests/test_realtime/test_openai_realtime_model.py
+++ b/tests/test_realtime/test_openai_realtime_model.py
@@ -1,0 +1,29 @@
+from __future__ import annotations
+
+from types import SimpleNamespace
+from typing import cast
+
+from livekit.agents import llm
+from livekit.agents.llm.remote_chat_context import RemoteChatContext
+from livekit.plugins.openai.realtime.realtime_model import RealtimeSession
+
+
+def test_update_chat_ctx_deletes_empty_remote_items() -> None:
+    remote_ctx = RemoteChatContext()
+    audio_item = llm.ChatMessage(id="audio_item", role="user", content=[])
+    kept_item = llm.ChatMessage(id="assistant_item", role="assistant", content=["kept"])
+    remote_ctx.insert(None, audio_item)
+    remote_ctx.insert(audio_item.id, kept_item)
+
+    session = cast(RealtimeSession, SimpleNamespace(_remote_chat_ctx=remote_ctx))
+    events = RealtimeSession._create_update_chat_ctx_events(
+        session,
+        llm.ChatContext(items=[kept_item]),
+    )
+
+    delete_ids = [
+        getattr(event, "item_id", None)
+        for event in events
+        if getattr(event, "type", None) == "conversation.item.delete"
+    ]
+    assert delete_ids == ["audio_item"]


theomonnom (approved):

theomonnom:
Seems like https://github.com/he-yufeng/agents/blob/29ad338bcf2ce014f1f6c62f9cadc166a1f7e422/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py#L1360 is now unused ?

he-yufeng:
Good catch. I kept that helper for the update path below: when an existing item id is recreated with different content, it still avoids recreating server-only audio/image items whose content was never synced down. This PR only changes the remove path so those trimmed empty items can be deleted from the server context.

he-yufeng:
Approved a couple of weeks back and CI is green. Anything else needed before this can go in?

PR #5902, opened 2026-05-30
fix(openai realtime): pass through already-formed wss:// base URLs

u9g:
## Summary
- The `test_process_base_url` case for already-wss passthrough was silently a no-op: a trailing comma made the asserted RHS a 1-tuple (always truthy), and `# noqa: F631` suppressed ruff's warning about it.
- Removing the comma surfaced a real bug — `process_base_url` was appending `?model=<model>` to fully-formed `wss://` URLs with custom paths (e.g. LiveKit's voice gateway `wss://livekit.ai/voice/v1/chat/voice?client=oai&...`), corrupting them.
- Fix: only inject the `model` query param when the input had its scheme rewritten from `http(s)` or matched a known openai path. Already-`wss://` URLs with a custom path now pass through unchanged.
- Added a second passthrough regression test for a bare `wss://example.com/foo` input.

## Test plan
- [x] `uv run pytest tests/test_config.py -k test_process_base_url`
- [x] `uv run ruff check` on touched files

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -734,7 +734,8 @@ def process_base_url(
     azure_deployment: str | None = None,
     api_version: str | None = None,
 ) -> str:
-    if url.startswith("http"):
+    scheme_rewritten = url.startswith("http")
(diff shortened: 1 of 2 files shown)


u9g:
Closing — the test case being 'fixed' has unclear intent (passthrough behavior isn't backed by OpenAI/Azure docs and the LiveKit voice gateway contract isn't verified).

PR #5914, opened 2026-06-01
Add LiveKit cloud auth to krisp viva plugin

lukasIO:
simplified default usage for cloud users now looks like

```py
  await session.start(
        agent=MyAgent(),
        room=ctx.room,
                noise_cancellation=krisp.KrispVivaFilterFrameProcessor(noise_suppression_level=90)
            ),
        ),
    )
```


livekit-plugins/livekit-plugins-krisp/README.md
```diff
@@ -9,67 +9,46 @@ Real-time noise reduction for LiveKit voice agents using [Krisp's VIVA SDK](http
...
+> That path has additional prerequisites — see below.
 
 ## Quick Start
 
-### Human-to-Bot Noise Cancellation / Voice Isolation (Recommended)
-
-For cleaning up user audio before STT/VAD processing using the FrameProcessor approach:
+By default, `KrispVivaFilterFrameProcessor` uses **LiveKit Cloud** authentication: the
+bundled backend ships the noise-reduction model and authenticates against LiveKit Cloud
+using the room JWT the agent framework hands to the FrameProcessor automatically.
```
1egoman, line 26: thought: Should the default be livekit cloud, since that would be backwards incompatible? Or maybe something like checking to see if those special krisp environment variables are set, and if they are but the "livekit cloud" auth provider is enabled, then log some sort of warning, and (a little less sure if this part is a good idea) maybe also change the initial default mode to the krisp license key option?

pyproject.toml
```diff
@@ -31,9 +31,10 @@ livekit-plugins-hamming = { workspace = true }
 livekit-plugins-hume = { workspace = true }
 livekit-plugins-inworld = { workspace = true }
 livekit-plugins-krisp = { workspace = true }
+livekit-plugins-krisp-internal = { path = "../plugin-krisp-viva-internal/target/packages/python", editable = true } # TODO remove this after pypi publish
```
1egoman: thought: Is it worth calling this `-internal` package something that makes it more clear that this is where the livekit cloud auth provider lives?

1egoman (commented):
Generally looks good to me, once this is no longer draft I can take another look and probably approve.

livekit-plugins/livekit-plugins-krisp/README.md
lukasIO: That's a good point. Non breaking transition with a legacy warning log was already the case (gated on the env var presence), but your comment made me refine that behaviour and ensure that both env vars have to be present

livekit-plugins/livekit-plugins-krisp/examples/krisp_agent_example.py
```diff
@@ -77,13 +84,31 @@ async def entrypoint(ctx: JobContext):
...
+    #             # Both default to env vars KRISP_VIVA_SDK_LICENSE_KEY and
+    #             # KRISP_VIVA_FILTER_MODEL_PATH if omitted.
+    #             license_key="...",
+    #             model_path="/path/to/model.kef",
+    #         ),
+    #         noise_suppression_level=100,
+    #         frame_duration_ms=10,
+    #         sample_rate=16000,
+    #     )
+
```
1egoman, line 111: nitpick: Maybe add a note here that this path is required for use with a non hosted SFU?

livekit-plugins/livekit-plugins-krisp/livekit/plugins/krisp/_krisp.py
```diff
@@ -0,0 +1,365 @@
+# Copyright 2023 LiveKit, Inc.
```
1egoman: nitpick: Update the copyright year to 2026, this is a new file:


```suggestion
# Copyright 2026 LiveKit, Inc.
```

livekit-plugins/livekit-plugins-krisp/livekit/plugins/krisp/_krisp.py
```diff
@@ -0,0 +1,365 @@
...
+        assert cls._sample_rates is not None, "SDK not acquired"
+        if sample_rate not in cls._sample_rates:
+            supported = ", ".join(str(r) for r in sorted(cls._sample_rates))
+            raise ValueError(
+                f"Unsupported sample rate: {sample_rate} Hz. Supported rates: {supported} Hz"
+            )
+        return cls._sample_rates[sample_rate]
+
+    @classmethod
+    def frame_duration_enum(cls, frame_duration_ms: int) -> Any:
```
1egoman, line 128: nitpick: If you want to get rid of these `Any`s, you could do `if TYPE_CHECKING: import krisp_audio`. I'm guessing that's why you didn't make these `krisp_audio.SamplingRate` / `krisp_audio.FrameDuration` respectively?

UPDATE: Actually reading a bit further, I think I forgot that `krisp_audio` can't be imported since it's a krisp vendored dependency. So nevermind on this, unless you could think of another creative solution...

livekit-plugins/livekit-plugins-krisp/livekit/plugins/krisp/_krisp.py
```diff
@@ -0,0 +1,365 @@
...
+        # first full chunk is ready the deficit is zero-padded; samples in and
+        # out stay balanced over time, so there is no drift.
+        n = frame.samples_per_channel
+        if len(self._out_buf) >= n:
+            out = self._out_buf[:n]
+            self._out_buf = self._out_buf[n:].copy()
+        else:
+            padding = np.zeros(n - len(self._out_buf), dtype=np.int16)
+            out = np.concatenate((padding, self._out_buf))
+            self._out_buf = np.empty(0, dtype=np.int16)
```
1egoman: question: Is this zero padding behavior what should be done here? I'd think that you may pad it with zeros going into the model if the chunk size is too small, but that going out you'd pull off the zeros before returning it back to the user.

Context behind my question: Depending on how chunks go through the pipeline, especially if two chunks got backed up where the first were zero padded and the second wasn't, I think the current state of this could result in brief cuts in and out if these audio chunks were concatenated together. I've encountered this cutting in and out phenomenon locally in the aic plugin when I was doing something somewhat similar (only in that case, the aic library itself was what was doing the zero padding) and it's a particularly bad failure mode. I wonder if it would be possible to try to make this case occur by varying the feed rate of audio into the plugin (and maybe adding arbitrary delay in the plugin to force it to take longer than normal)?

livekit-plugins/livekit-plugins-krisp/pyproject.toml
```diff
@@ -37,12 +37,9 @@ dependencies = [
     "livekit-agents>=1.6.4",
     "livekit>=1.0.23,<2",  # FrameProcessor support requires livekit>=1.0.23 (PR #4145)
     "numpy>=1.20.0",
-    # Note: krisp-audio is a required runtime dependency but not available on public PyPI.
-    # It must be obtained and installed separately from Krisp.
-    # Install with: pip install krisp-audio (from Krisp's private repository or distribution)
-    # The plugin handles missing krisp-audio gracefully with ImportError.
-    # This dependency is declared here for documentation purposes only.
-    # Since it's not on PyPI, it cannot be listed in dependencies or optional-dependencies.
```
1egoman, line 45: thought: Is it still worth keeping this caveat around, since it's still true for the local SFU krisp plugin path?

1egoman (approved):
Generally makes sense to me!

1egoman (approved):

livekit-plugins/livekit-plugins-krisp/livekit/plugins/krisp/_krisp.py
lukasIO: addressed in https://github.com/livekit/agents/pull/5914/commits/65c07209667dde4a7a65930d24747e21a5dc1d6b

livekit-plugins/livekit-plugins-krisp/pyproject.toml
lukasIO: this should become clear from the README and the krisp-audio package isn't required anymore as it will work ootb with cloud usage

theomonnom (approved):

PR #5926, opened 2026-06-01
fix(recorder): prevent close hang and fix corrupt frame splits

chenghao-mou:
## Summary

Two distinct bugs in `RecorderIO`, both surfaced by a real session where a SIP caller hung up mid-goodbye while the "caller silence check-in" pause/resume feature was active. Symptom: session shutdown stalled ~60s and the recording was incomplete/garbled.

### Bug 1 — `aclose()` hangs forever if the encode thread dies
`_encode_thread` only resolved `_close_fut` at the end of its normal loop. If the thread raised anywhere in the loop, the future stayed unresolved and `RecorderIO.aclose()` blocked on `await asyncio.shield(self._close_fut)`, stalling session shutdown until the caller's aclose watchdog fired (~60s). Because the encoder runs on a `daemon` thread, the exception went to `threading.excepthook` (stderr) instead of the structured logger, so it was invisible in observability exports.

- Wrap the encode loop in `try/except/finally` so `_close_fut` is **always** resolved (idempotent `_resolve_close_fut`), even when encoding raises.
- Log encoder exceptions via `logger.exception` instead of losing them.
- Cancel the previously-leaked `_forward_atask` in `aclose()`.

### Bug 2 — `_split_frame` corrupts audio on interrupted/paused playback
`rtc.AudioFrame.data` is a memoryview of int16 **samples** (`format="h"`) — indexed in samples, not bytes. `_split_frame` computed `bytes_per_sample = num_channels * 2` and over-sliced by 2×, producing frames whose `data` length disagreed with `samples_per_channel`. This path runs only on interrupted playback and pause/silence insertion, corrupting the recording around those points.

- Index the buffer in samples (`split = samples_needed * num_channels`), clamp `samples_needed`, and set each half's sample count correctly.

## Tests
- Adds `_split_frame` regression tests (consistency + lossless round-trip for mono/stereo, plus boundaries). The encode path previously had no coverage.
- All 23 tests in `tests/test_recording.py` pass; lint clean.

## Note
The structural fix (Bug 1) makes the hang impossible regardless of which exception killed the thread, and the new `logger.exception` will surface the precise traceback if it recurs. Bug 2 is a proven correctness bug on the interrupted/paused code path exercised by this session.

### DIFF
--- a/tests/test_recording.py
+++ b/tests/test_recording.py
@@ -8,8 +8,10 @@ from typing import Any
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import aiohttp
+import numpy as np
 import pytest
 
+from livekit import rtc
 from livekit.agents import Agent, AgentSession
 from livekit.agents.telemetry.traces import _upload_session_report
 from livekit.agents.voice.agent_session import (
@@ -17,6 +19,7 @@ from livekit.agents.voice.agent_session import (
     _RECORDING_ALL_ON,
     RecordingOptions,
 )
+from livekit.agents.voice.recorder_io.recorder_io import _split_frame
 
 from .fake_io import FakeAudioInput, FakeAudioOutput, FakeTextOutput
 from .fake_llm import FakeLLM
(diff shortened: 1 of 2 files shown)


tests/test_recording.py
```diff
@@ -415,3 +418,60 @@ async def test_recorder_io_not_created_when_audio_false() -> None:
...
+    """Splitting at or beyond the edges returns an empty half and the original."""
+    frame = _ramp_frame(100, 1)
+
+    empty, whole = _split_frame(frame, 0.0)
+    assert empty.samples_per_channel == 0 and len(empty.data) == 0
+    assert whole.samples_per_channel == 100
+
+    whole2, empty2 = _split_frame(frame, frame.duration * 2)
+    assert whole2.samples_per_channel == 100
+    assert empty2.samples_per_channel == 0 and len(empty2.data) == 0
```
theomonnom, line 477: Are those tests really useful? They're self-contained, and don't even import framework code

theomonnom (approved):

tests/test_recording.py
theomonnom: nvm i see it imports _split_frame

longcw (approved):

PR #5935, opened 2026-06-02
feat(anam): forward output video dimensions via SessionOptions

### LINKED ISSUE #5954: Anam plugin: support explicit output video dimensions via SessionOptions
## Summary

The `livekit-plugins-anam` avatar plugin mints an Anam session with no way to request the avatar's output video dimensions. Anam's public session-token API now accepts explicit pixel output dimensions (`sessionOptions.videoWidth` / `videoHeight`), so the plugin should expose and forward them.

## Motivation

Anam resolves the avatar's output frame size at session-token creation. Today the plugin always gets the avatar model's default size — an Agents developer can't request a specific output (e.g. Cara 4 landscape `1152x768`, or portrait once Anam enables it server-side). Other provider plugins already expose this; e.g. `livekit-plugins-avatario` has a `VideoInfo(video_width, video_height)` dataclass.

## Proposal

Add an optional `anam.SessionOptions(video_width, video_height)` (pixels only, both-or-neither) that the plugin forwards into the Anam session-token request as `sessionOptions.videoWidth`/`videoHeight`:

```python
session = anam.AvatarSession(
    persona_config=anam.PersonaConfig(name="avatar", avatarId=ANAM_AVATAR_ID),
    session_options=anam.SessionOptions(video_width=1152, video_height=768),
)
```

- Pixels only — no portrait/landscape strings.
- Backward compatible: optional; omitted -> identical request to today -> Anam applies the model default.
- Validation stays server-side in Anam (single source of truth): an unsupported pair returns HTTP 400, already surfaced by the plugin as `APIStatusError`. New output sizes (e.g. portrait) can therefore be enabled by Anam server-side with no plugin release.

## Scope

Small, additive change to the anam plugin only (`types.py`, `avatar.py`, `api.py`, `__init__.py`) plus the example. Mirrors the existing Avatario precedent. No new dependencies; no change to the LiveKit token / engine-session / remote-track flow.

## Open question for maintainers

Naming: top-level `SessionOptions` (matches Anam's own `sessionOptions` API field and leaves room for future session options) vs a nested `AvatarSession.VideoInfo` (matches Avatario). Happy to match house preference.

Smidge:
Closes #5954

## What

Adds an optional `anam.SessionOptions(video_width, video_height)` that the Anam avatar plugin forwards into Anam's session-token request as `sessionOptions.videoWidth` / `videoHeight`. The avatar video then publishes at the Anam-resolved output size.

```python
session = anam.AvatarSession(
    persona_config=anam.PersonaConfig(name="avatar", avatarId=ANAM_AVATAR_ID),
    session_options=anam.SessionOptions(video_width=1152, video_height=768),
)
```

## Why

Anam's public session-token API now accepts explicit pixel output dimensions. This lets Agents users choose the avatar's output size without any provider-specific dimensions API.

## Design

- **Pixels only**, both-or-neither — no portrait/landscape strings.
- **Validation lives in Anam, not the plugin.** The plugin is a thin pass-through and doesn't know the per-model allow-list; Anam validates the requested pair against the avatar model and returns HTTP 400 for an unsupported (or half) pair, which the plugin already surfaces as `APIStatusError` with Anam's descriptive message. No silent downgrade.
- **Mirrors `livekit-plugins-avatario`** (`VideoInfo(video_width, video_height)`). I named it `SessionOptions` to match Anam's own `sessionOptions` API field and leave room for future session-level options — happy to switch to a nested `AvatarSession.VideoInfo` if you'd prefer consistency with Avatario.
- **Backward compatible:** `session_options` is optional (`NOT_GIVEN`); when omitted the request body is identical to today and Anam applies the avatar model's default. No behaviour change for existing users.

## Scope

5 files — `types.py` (new `SessionOptions` dataclass), `avatar.py` + `api.py` (thread through + emit camelCase), `__init__.py` (export), and the anam example. No new dependencies; no change to the LiveKit token / `start_engine_session` / `wait_remote_track` flow. `ruff format` + `ruff check` clean. Left `version.py`/changelog to the release bot per CONTRIBUTING.

Happy to add unit tests for the payload construction in whatever harness you prefer.

### DIFF
--- a/livekit-plugins/livekit-plugins-anam/livekit/plugins/anam/api.py
+++ b/livekit-plugins/livekit-plugins-anam/livekit/plugins/anam/api.py
@@ -97,17 +97,22 @@ class AnamAPI:
             "livekitToken": livekit_token,
         }
 
-        if session_options is not None:
-            # Anam's public API speaks camelCase pixel dimensions. Send both or
-            # neither; Anam rejects a half pair and any unsupported pair with an
-            # HTTP 400 (surfaced below as APIStatusError) rather than downgrading.
-            video_options: dict[str, int] = {}
-            if session_options.video_width is not None:
-                video_options["videoWidth"] = session_options.video_width
-            if session_options.video_height is not None:
-                video_options["videoHeight"] = session_options.video_height
-            if video_options:
-                payload["sessionOptions"] = video_options
+        if session_options is not None and (
+            session_options.video_width is not None or session_options.video_height is not None
+        ):
+            # Anam's public API speaks camelCase pixel dimensions and wants them
+            # as a matched pair: it rejects a lone width/height (and any
+            # unsupported pair) with an HTTP 400, surfaced below as
+            # APIStatusError, rather than downgrading. Fail fast on a half pair
+            # rather than round-tripping a 400.
+            if session_options.video_width is None or session_options.video_height is None:
+                raise ValueError(
+                    "video_width and video_height must be set together (both or neither)"
+                )
+            payload["sessionOptions"] = {
+                "videoWidth": session_options.video_width,
+                "videoHeight": session_options.video_height,
+            }
 
         headers = {
             "Authorization": f"Bearer {self._api_key}",  # Use API Key here
--- a/livekit-plugins/livekit-plugins-anam/livekit/plugins/anam/api.py
+++ b/livekit-plugins/livekit-plugins-anam/livekit/plugins/anam/api.py
@@ -14,7 +14,7 @@ from livekit.agents import (
 
 from .errors import AnamException
 from .log import logger
-from .types import PersonaConfig
+from .types import PersonaConfig, SessionOptions
 
 DEFAULT_API_URL = "https://api.anam.ai"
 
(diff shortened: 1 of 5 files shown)


livekit-plugins/livekit-plugins-anam/livekit/plugins/anam/api.py
```diff
@@ -80,13 +89,26 @@ async def create_session_token(
...
+            # Anam's public API speaks camelCase pixel dimensions. Send both or
+            # neither; Anam rejects a half pair and any unsupported pair with an
+            # HTTP 400 (surfaced below as APIStatusError) rather than downgrading.
+            video_options: dict[str, int] = {}
+            if session_options.video_width is not None:
+                video_options["videoWidth"] = session_options.video_width
+            if session_options.video_height is not None:
+                video_options["videoHeight"] = session_options.video_height
+            if video_options:
+                payload["sessionOptions"] = video_options
```
longcw: this doesn't match the comment above "Send both or neither"?

Smidge: Good catch, you're right. It builds `video_options` from whichever field is non-`None`, so a lone `video_width`/`video_height` still goes out as exactly the half pair the comment says Anam rejects.

Fixing it to send both or neither, and to fail fast locally on a half pair rather than round-tripping a 400:

```python
if session_options is not None and (
    session_options.video_width is not None
    or session_options.video_height is not None
):
    if session_options.video_width is None or session_options.video_height is None:
        raise ValueError(
            "video_width and video_height must be set together (both or neither)"
        )
    payload["sessionOptions"] = {
        "videoWidth": session_options.video_width,
        "videoHeight": session_options.video_height,
    }
```

Thanks for the review!

Smidge: Fixed in 73adbbc31 — `sessionOptions` now only goes out when both dimensions are present, and raises a `ValueError` on a lone width/height so it fails fast locally instead of round-tripping a 400.

tinalenguyen (approved):
lgtm, thank you for the PR!

PR #5942, opened 2026-06-02
(speechmatics): remove hard coded confidence 

### LINKED ISSUE #5880: Speechmatics STT plugin always reports confidence=1.0 instead of provider confidence

tinalenguyen:


livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -904,7 +904,6 @@ def _send_frames(self, segments: list[dict[str, Any]], is_final: bool) -> None:
                 start_time=segment.get("metadata", {}).get("start_time", 0)
                 + self.start_time_offset,
                 end_time=segment.get("metadata", {}).get("end_time", 0) + self.start_time_offset,
-                confidence=1.0,
```
davidzhao, line 907: lg.. do we have actual confidence values somewhere?

davidzhao (approved):

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
tinalenguyen: i don't think their voice SDK exposes it (their raw websocket api does), i will ask about it

PR #5952, opened 2026-06-03
fix(openai): skip empty messages in realtime chat updates

### LINKED ISSUE #5951: Bug: `update_chat_ctx` sends empty `item.content` array when `llm=None
### Bug Description

When `update_chat_ctx` is called while `llm=None` (e.g., during connection recovery or mid agent-handoff), it appears to push an empty chat history to the OpenAI Realtime API instead of failing gracefully. This results in the following error:

```
Session encountered a recoverable error: type='realtime_model_error' timestamp=1780062146.7014737 label='livekit.plugins.openai.realtime.realtime_model.RealtimeModel' error=APIError('OpenAI Realtime API returned an error', body=RealtimeError(message="Invalid 'item.content': empty array. Expected an array with minimum length 1, but got an empty array instead.", type='invalid_request_error', code='empty_array', event_id='chat_ctx_create_eafc907953bc', param='item.content'), retryable=True) recoverable=True
```

### Expected Behavior

`update_chat_ctx` should guard against `llm=None` and either skip the update, raise a local error, or wait until the LLM connection is restored — rather than forwarding an empty context to the API.

### Reproduction Steps

```bash
1. Trigger a session where `llm` is temporarily set to `None` (e.g., during an `AgentHandoff` or connection drop/recovery)
2. Have `update_chat_ctx` called during that window
3. Observe that an empty array is submitted to OpenAI rather than the call being skipped or raising a local error
```

### Operating System

MacOS

### Models Used

GPT-realtime

### Package Versions

```bash
"livekit-agents[azure,openai,turn-detector,silero,elevenlabs]==1.5.12"
"livekit-api~=1.1"
```

### Session/Room/Call IDs

"room_id": "RM_sd6JZrs4uxom",
"sid": "PA_mqSbRFjQXECC",

### Proposed Solution

```python
We described the problem to your deepwiki, and it proposes the following root causes:
[https://deepwiki.com/search/livekit-provides-an-agenthando_25fa9924-c873-4c3f-bb38-4694b13dfbbd?mode=deep](https://deepwiki.com/search/livekit-provides-an-agenthando_25fa9924-c873-4c3f-bb38-4694b13dfbbd?mode=deep)

- update_chat_ctx does not pass exclude_empty_message=True
    - realtime_model.py line 1298
- _create_update_chat_ctx_events has no empty-content guard on to_create items
    - realtime_model.py line 1374

We are not completely sure about the inner workings of livekit-agents, but those comments could be a good lead.
```

### Additional Context

_No response_

_No response_

nightcityblade:
Fixes #5951

## Summary
- filter empty chat messages before generating OpenAI Realtime chat context update events
- add a regression test covering update_chat_ctx with an empty message

## Tests
- uv run --group dev pytest tests/test_openai_realtime_chat_ctx.py -q
- uv run --group dev ruff check livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py tests/test_openai_realtime_chat_ctx.py

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1249,7 +1249,6 @@ class RealtimeSession(
     async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
         async with self._update_chat_ctx_lock:
             chat_ctx = chat_ctx.copy(
-                exclude_empty_message=True,
                 exclude_handoff=True,
                 exclude_config_update=True,
             )
@@ -1280,6 +1279,20 @@ class RealtimeSession(
     ) -> list[ConversationItemCreateEvent | ConversationItemDeleteEvent]:
         events: list[ConversationItemCreateEvent | ConversationItemDeleteEvent] = []
         remote_ctx = self._remote_chat_ctx.to_chat_ctx()
+
+        # Empty message content can mean either:
+        # - a local placeholder that should not be created remotely, or
+        # - an existing remote item with non-text content (audio/images) that is not
+        #   synced into the agent-side ChatContext.
+        # Keep empty messages that already exist remotely so we do not delete them.
+        remote_ids = {item.id for item in remote_ctx.items}
+        chat_ctx = llm.ChatContext(
+            [
+                item
+                for item in chat_ctx.items
+                if item.type != "message" or item.content or item.id in remote_ids
+            ]
+        )
         diff_ops = llm.utils.compute_chat_ctx_diff(remote_ctx, chat_ctx)
 
         def _delete_item(msg_id: str) -> None:
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1249,6 +1249,7 @@ class RealtimeSession(
     async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
         async with self._update_chat_ctx_lock:
             chat_ctx = chat_ctx.copy(
+                exclude_empty_message=True,
                 exclude_handoff=True,
                 exclude_config_update=True,
             )
--- a/tests/test_openai_realtime_chat_ctx.py
+++ b/tests/test_openai_realtime_chat_ctx.py
@@ -1,9 +1,13 @@
 import asyncio
 import types
 
+import pytest
+
 from livekit.agents import llm
 from livekit.agents.llm import remote_chat_context
 from livekit.plugins.openai.realtime.realtime_model import RealtimeSession
+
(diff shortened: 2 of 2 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1323,6 +1249,7 @@ def update_options(
     async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
         async with self._update_chat_ctx_lock:
             chat_ctx = chat_ctx.copy(
+                exclude_empty_message=True,
```
longcw: we cannot exclude empty message here, we don't have content synced down for some types of content (audio/images), these won't be present in the Agent's view of the context.

that means when the content is empty in agent's view, we should still keep it if it's also existed in the remote session.

maybe do the filtering in `_create_update_chat_ctx_events` that keep the item exists in remote chat context
```python
    def _create_update_chat_ctx_events(
        self, chat_ctx: llm.ChatContext
    ) -> list[ConversationItemCreateEvent | ConversationItemDeleteEvent]:
        events: list[ConversationItemCreateEvent | ConversationItemDeleteEvent] = []
        remote_ctx = self._remote_chat_ctx.to_chat_ctx()

        # prune empty items not present in remote_ctx
        remote_ids = {item.id for item in remote_ctx.items}
        filtered_chat_ctx = llm.ChatContext(
            [
                item
                for item in chat_ctx.items
                if item.type != "message" or item.content or item.id in remote_ids
            ]
        )
        ...
```

longcw (changes requested):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
nightcityblade: Thanks, updated in 7c6d1d2b496c4144e19c8901bfc98332e6c5b8b0. I moved the empty-message filtering into `_create_update_chat_ctx_events` so new local empty messages are pruned, but existing remote empty messages are retained to avoid deleting non-text content that is not synced into the agent view. Also added coverage for both cases.

longcw (approved):
thanks! could you add `pytestmark = pytest.mark.unit` to the test file so the unit tests can run through it, for [example](https://github.com/livekit/agents/blob/854832583b8c2ffa976b73496e3c63717e76e146/tests/test_tools.py)?

nightcityblade:
Added that in f7816bf — `tests/test_openai_realtime_chat_ctx.py` now includes `pytestmark = pytest.mark.unit` as requested.

Validation rerun:
- `uv run pytest tests/test_openai_realtime_chat_ctx.py` ✅
- `uv run ruff check tests/test_openai_realtime_chat_ctx.py` ✅

Local pytest emits a `PytestUnknownMarkWarning` for `unit` on this macOS setup because the mark does not appear to be registered in the current checkout, but I kept the exact marker you requested so the test can participate in the repo's unit-test selection path.

bml1g12:
thanks for the quick fix here!

PR #5960, opened 2026-06-03
fix(job): log unhandled exceptions in fire-and-forget tasks

theomonnom:


theomonnom:
Superseded by #5961. This branch was based on unmerged local work (the hotel_receptionist commit), so the diff swept in unrelated files. #5961 contains only the job.py change against main.

PR #5962, opened 2026-06-04
fix(ipc): rate-limit and contextualize high-memory warnings

Bobronium:
## Changes
- **Rate-limit** the warning: at most once per `_MEMORY_WARN_COOLDOWN` (120s) per process, but re-emits early if usage grew by ≥`_MEMORY_WARN_RESET_DELTA_MB` (50MB) since the last warning, so real leaks still surface promptly.
- **Reword** the message and mark it advisory when no hard limit is configured.
- Add extra: `uptime`, `has_running_job`, `baseline_memory_mb`, `growth_memory_mb` (current − baseline).
- **`process_kind` property**: base = `"process"`, overridden to `"job process"` / `"inference process"`.

## Behavior
- No change to *when* a process is killed — the `memory_limit_mb` path is untouched except for richer log context. Warning thresholds and defaults unchanged.
- Log message string and `extra` keys change

## Testing
- New unit tests (`tests/test_supervised_proc_memory.py`): cooldown suppression, growth-based re-emit, baseline/growth reporting, default `process_kind`.

### DIFF
--- a/livekit-agents/livekit/agents/ipc/job_proc_executor.py
+++ b/livekit-agents/livekit/agents/ipc/job_proc_executor.py
@@ -69,7 +69,7 @@ class ProcJobExecutor(SupervisedProc):
 
     @property
     def process_kind(self) -> SupervisedProcKind:
-        return SupervisedProcKind.job
+        return SupervisedProcKind.JOB
 
     @property
     def status(self) -> JobStatus:
--- a/livekit-agents/livekit/agents/ipc/job_proc_executor.py
+++ b/livekit-agents/livekit/agents/ipc/job_proc_executor.py
@@ -16,7 +16,7 @@ from . import channel, proto
 from .inference_executor import InferenceExecutor
 from .job_executor import JobStatus
 from .job_proc_lazy_main import ProcStartArgs, proc_main
-from .supervised_proc import SupervisedProc
+from .supervised_proc import SupervisedProc, SupervisedProcKind
 
 
 class ProcJobExecutor(SupervisedProc):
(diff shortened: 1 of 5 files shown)


longcw (approved):
lgtm!

livekit-agents/livekit/agents/ipc/job_proc_executor.py
```diff
@@ -67,6 +67,10 @@ def __init__(
     def id(self) -> str:
         return self._id
 
+    @property
+    def process_kind(self) -> str:
+        return "job process"
```
theomonnom: Let's use a string enum somewhere

theomonnom (approved):
lgtm

livekit-agents/livekit/agents/ipc/supervised_proc.py
```diff
@@ -29,6 +30,23 @@
...
+# process that sits above the threshold doesn't emit a warning on every sample
+_MEMORY_WARN_COOLDOWN = 120.0
+# re-emit the warning before the cooldown elapses if usage grew by at least this much
+# since the last warning (so a genuine leak still surfaces promptly)
+_MEMORY_WARN_RESET_DELTA_MB = 50.0
+
+
+class SupervisedProcKind(str, Enum):
+    job = "job"
+    inference = "inference"
```
theomonnom:
```suggestion
    JOB = "job"
    INFERENCE = "inference"
```
To be consistent with other stuff in the codebase, let's make them UPPERCASE

PR #5968, opened 2026-06-04
Deploy examples against the deployed git ref instead of PyPI

theomonnom:


theomonnom:
Superseded by #5969, which contains only the deploy-examples change (this branch also bundled unrelated hotel_receptionist work).

PR #5980, opened 2026-06-05
tests: speed up unit tests

Bobronium:
Running `pytest --unit` with these changes goes from 3 minutes 44 seconds to just **30 seconds**.  

#### Virtual time
Takes advantage of the fact that many of the tests rely on `asyncio.sleep()` to synchronize events in the loop. It unties it from the real time so that `await asyncio.sleep(60)` returns instantly with `loop.time()` reporting 60 seconds passed. `time.time()` is then patched to report loop time. This brings down `test_agent_session.py` test time form 90 seconds to just 1 second without any modifications to the test itself.

- To disable globally `--real-time` option can be used.
- To enable/disable per-module or per-test `pytest.mark.virtual_time`/`.real_time` can be used.

#### Concurrent tests
Tests with real I/O are groupped by their module and executed in a shared event loop. This shrinks wall time of the whole module to wall time of the slowest test + raw compute. `contextvars` are used to preserve input capture and track leaked tasks.

- To disable concurrent execution, `--no-concurrent` option can be used.
- To enable/disable per-module or per-test, `pytest.mark.no_concurrent`/`.no_concurrent` can be used.

### Changes
- Add `pytest-asyncio-concurrent` — gives the speedup for I/O bound tests
- Add `tests/concurrent.py` that extends on `pytest-asyncio-concurrent` and adds crucial features:
    - Captures stdout/stderr and reports on failure (same as `pytest`)
    - Allows leaked tasks detection in concurrent mode (so that `fail_on_leaked_tasks` continues to work)
    - adds `--concurrent`/`--no-concurrent` options
    - Adds live progress
- Add `tests/virtual_time.py` that provides instant and deterministic `asyncio` clock for time-bound execution

#### Changes to the tests sources
- https://github.com/livekit/agents/pull/5980/commits/04fe6cbb55c32d1058a01a4c0ddda790b5834770: Add `concurrent`/`virtual_time` markers to tests modules
- https://github.com/livekit/agents/pull/5980/changes/a8f5be0be8b76efefaa55743d8ad0a5d77b15b1b: Tighten timings in `test_audio_decoder.py`, `test_ipc.pt` and `test_room.py` 
- https://github.com/livekit/agents/pull/5980/changes/396bce8cbf4618a292ab28a257ea13f58bd3c5a6: Make `test_async_channel` and `test_proc_pool_launch_job_raises_when_all_spawns_fail` in `test_ipc.py` safe to run concurrently

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,9 +86,10 @@ dev = [
     "python-dotenv>=1.0.1",
     "mypy",
     "scikit-learn",
-    "pytest",
+    "pytest>=9.0.3,<9.1",
     "ruff",
     "pytest-asyncio>=0.25.3",
+    "pytest-asyncio-concurrent==0.5.2",
     "jiwer>=3.1.0",
     "scipy>=1.13.1",
     "tiktoken>=0.9.0",
@@ -125,7 +126,7 @@ convention = "google"
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "function"
-addopts = ["--import-mode=importlib", "--ignore=examples"]
+addopts = ["--import-mode=importlib", "--ignore=examples", "-p", "tests.concurrency"]
 pythonpath = ["."]
 testpaths = ["tests"]
 markers = [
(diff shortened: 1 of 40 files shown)


theomonnom (approved):
lgtm!

PR #5983, opened 2026-06-05
Hotel Receptionist example

theomonnom:


theomonnom:
Merging, I need the example to be on main for simulations. Lmk if you guys have any thoughts

PR #5995, opened 2026-06-07
fix(sarvam): emit speech timing for STT metrics

dhruvladia-sarvam:
## Summary
- Tracks Sarvam `STT` stream audio position and speech boundaries so `START_OF_SPEECH`, `FINAL_TRANSCRIPT`, and `END_OF_SPEECH` carry timing data that LiveKit can use for EOU metrics.
- Defers final transcript emission until a speech-end timestamp is known, while preserving `FINAL_TRANSCRIPT` before `END_OF_SPEECH` ordering for both transcript-before-EOS and EOS-before-transcript provider ordering.
- Adds focused Sarvam plugin tests for timestamp fallback behavior and event ordering without modifying `audio_recognition.py` or other STT providers.

## Test plan
- `uv run pytest livekit-plugins/livekit-plugins-sarvam/tests/test_speech_timing.py livekit-plugins/livekit-plugins-sarvam/tests/test_language_probability.py -v`
- `uv run ruff check livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py livekit-plugins/livekit-plugins-sarvam/tests/test_speech_timing.py`
- `python -m py_compile livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py livekit-plugins/livekit-plugins-sarvam/tests/test_speech_timing.py`
- Local console smoke test with `sarvam.STT(model="saaras:v3")`, `vad=None`, and `turn_handling={"turn_detection": "stt"}` produced non-zero EOU metrics, including `transcription_delay=0.146s` and `end_of_utterance_delay=0.501s`.

## Notes
- This is scoped to `livekit-plugins-sarvam`'s `sarvam.STT` path in `stt.py`; it does not change `STTStreaming` or framework turn detection behavior.
- The console smoke test emitted a non-interactive terminal key-listener traceback from console mode, but the agent continued and emitted the expected EOU metrics.

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
@@ -992,32 +992,7 @@ class SpeechStream(stt.SpeechStream):
             return None
         return float(value)
 
-    def _ensure_utterance_timing_state(self) -> None:
-        if not hasattr(self, "_audio_position"):
-            self._audio_position = 0.0
-        if not hasattr(self, "_utterance_start_audio_pos"):
-            self._utterance_start_audio_pos = 0.0
-        if not hasattr(self, "_utterance_speech_end_audio_pos"):
-            self._utterance_speech_end_audio_pos = None
-        if not hasattr(self, "_utterance_speech_start_wall"):
-            self._utterance_speech_start_wall = None
-        if not hasattr(self, "_utterance_speech_end_wall"):
-            self._utterance_speech_end_wall = None
-        if not hasattr(self, "_pending_final_data"):
-            self._pending_final_data = None
-        if not hasattr(self, "_pending_eos"):
-            self._pending_eos = False
-        if not hasattr(self, "_eos_fallback_task"):
-            self._eos_fallback_task = None
-        if not hasattr(self, "_eos_fallback_timeout"):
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
```diff
@@ -970,6 +984,165 @@ def _maybe_set_server_request_id(self, data: dict) -> None:
...
 
+    @staticmethod
+    def _positive_time(value: object) -> float | None:
+        if isinstance(value, bool) or not isinstance(value, (int, float)):
+            return None
+        if value <= 0:
+            return None
+        return float(value)
+
+    def _ensure_utterance_timing_state(self) -> None:
```
tinalenguyen: all these fields are already set in __init__, ithink we can remove this

tinalenguyen (approved):
thank you for the PR! i left a small comment but otherwise i tested it and it lgtm

livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
dhruvladia-sarvam: Ah I see, had left it there as a guard for tests that use SpeechStream.__new__() without calling __init__. Now removed as suggested, thank you

PR #6012, opened 2026-06-08
feat(voice): surface inference quota errors instead of going silent

### LINKED ISSUE #6009: Voice agent goes silently unresponsive on 429 inference_quota_exceeded — no audible/visible signal to the end user
**Repo:** `livekit/agents` (Python) · observed on `livekit-agents==1.5.2`
**Type:** bug / UX gap · **Area:** `voice` (AgentSession pipeline), `llm`, error surfacing
**Suggested labels:** `bug`, `enhancement`, `voice`, `dx`

---

## Summary

When the LLM endpoint returns **HTTP 429 with body `{"type": "inference_quota_exceeded", ...}`** (the project is out of LiveKit Inference credits / hit a quota), a voice agent **joins the room, publishes its audio track, and then never speaks**. There is no spoken message, nothing rendered in the Agent Builder preview, and — for the first few turns — not even a session close. From the caller's perspective the agent is simply, silently dead.

The gateway's behavior here is **correct**: it returns a well-formed, fully structured 429 (status code, `type`, human-readable `hint`, `quota_type`, `category`, `remaining_limit`, `documentation_url`). The gap is entirely on the **SDK surfacing side** — the SDK emits an `error` event but, by default, turns an unrecoverable LLM failure into **dead air** rather than anything a human (or a frontend) can perceive.

Reported by a user in the LiveKit community forum (LLM out of credits → "agent joins but never speaks", confirmed by staff as depleted inference credits, *not* a connection/config problem).

---

## Reproduction

1. Build a standard pipeline agent: `AgentSession(stt=deepgram, llm=<OpenAI-compatible / LiveKit Inference, e.g. gpt-5.2-chat>, tts=elevenlabs)`.
2. Use a project whose LLM token credit quota is exhausted (or otherwise force the LLM endpoint to return `429 inference_quota_exceeded`).
3. Join and speak to the agent.

**Expected:** the caller gets *some* perceptible signal — a spoken "this assistant is temporarily unavailable" message, and/or a frontend-renderable error/close state ("out of credits").

**Actual:** STT transcribes fine, then **silence**. The agent track is published but emits no audio. No reply, no spoken error. The Agent Builder preview shows nothing. The session does not even close until several consecutive failures (see root cause).

### The upstream response the SDK receives

HTTP `429`, JSON body (field names are stable; this is the LiveKit Inference gateway contract):

```json
{
  "type": "inference_quota_exceeded",
  "error": "LLM token credit quota exceeded, category: MaxGatewayCredits, remaining_limit: 0",
  "hint": "LLM token credit quota exhausted. Wait for the next billing cycle or upgrade your plan.",
  "quota_type": "llm",
  "category": "MaxGatewayCredits",
  "current_usage": "...",
  "remaining_limit": "0",
  "status": "...",
  "free_tier": "false",
  "documentation_url": "https://livekit.com/pricing"
}
```

Everything a good UX needs (`type`, `hint`, `quota_type`) is right there in the body. The SDK just never looks at it.

---

## Root cause (traced in `livekit-agents==1.5.2`)

**1. A 429 is classified non-retryable, which is correct for quota — but means it raises straight through.**
`APIStatusError` marks every 4xx (incl. 429) as `retryable=False`:
```python
# livekit/agents/_exceptions.py:66-70
if retryable is None:
    retryable = True
    if status_code >= 400 and status_code < 500:   # 429 falls here
        retryable = False
```
For `inference_quota_exceeded` that classification is *desirable* — retrying won't help until the billing window resets. The decoded JSON body is preserved on `APIError.body` (`_exceptions.py:23-29`), so the rich quota detail is technically available to callers.

**2. `LLMStream` emits an `error` event and re-raises — no audible side effect.**
```python
# livekit/agents/llm/llm.py:227-229
    self._emit_error(e, recoverable=False)   # -> self._llm.emit("error", LLMError(...))
    raise
```

**3. The error propagates to `AgentSession` as an `ErrorEvent`…**
```python
# livekit/agents/voice/agent_activity.py:1322-1324, 1345
if isinstance(error, llm.LLMError):
    error_event = ErrorEvent(error=error, source=self.llm)
    self._session.emit("error", error_event)
...
```
`ErrorEvent` / `CloseEvent` / `CloseReason` already exist (`voice/events.py:222-245`). So a developer who *explicitly* registered `@session.on("error")` can see it — but nothing happens by default.

**4. …and `AgentSession._on_error` deliberately does nothing for the first 3 unrecoverable LLM errors.** This is the silence:
```python
# livekit/agents/voice/agent_session.py:1363-1401  (max_unrecoverable_errors default = 3, line 129)
def _on_error(self, error: ...) -> None:
        return
    if error.type == "llm_error":
        self._llm_error_counts += 1
            return                      # <-- swallows the 1st, 2nd, 3rd failure entirely
    ...
        self._aclose_impl(error=error, reason=CloseReason.ERROR)   # only on the 4th
    )
```
The 3-strike tolerance is sensible for *transient* blips, but a `429 inference_quota_exceeded` is **terminal** — it will fail identically every turn. So the agent absorbs three silent dead turns and *then* closes with `CloseReason.ERROR` and still no spoken explanation.

**5. The reply turn produces no speech.** The pipeline reply path awaits the inference tasks with `wait_if_not_interrupted([*tasks])` (gathers with `return_exceptions=True`, `agent_activity.py:2410`). The raised 429 yields no reply text, so TTS is never invoked → the published audio track stays silent. (Unlike `_tts_task_impl`, the LLM reply path doesn't convert that failure into any fallback output.)

**Net:** `error` event fires (invisible unless subscribed) → no reply text → no TTS → silent track → repeats → session closes after 3 with no spoken reason. Hence "agent joins but never speaks."

> Realtime path (`RealtimeModel`, e.g. Gemini Live / OpenAI Realtime) routes through the same `_on_error` → `ErrorEvent` mechanism (`agent_activity.py:1325-1327`), so an error-event-driven fix covers both pipeline and realtime.

---

## Why this matters

- **Silent failure is the worst failure mode for voice.** A muted agent reads as "broken product," not "out of credits." The user has no way to self-diagnose.
- **The diagnosis exists and is being thrown away.** The gateway hands the SDK `type: inference_quota_exceeded` + a ready-to-speak `hint`. We surface none of it.
- **First-party surfaces are affected too.** The forum report notes the Agent Builder *preview* also showed nothing — so even LiveKit's own UI doesn't render this state.
- **It's not just credits.** The same path swallows any non-retryable LLM error (`quota_type` also covers `stt`/`tts`/`bargein`, and `category` includes rate-limit variants like `MaxConcurrentGatewayLLMRpm/Tpm`).

---

## Proposed fix (layered — ship the cheap parts first)

**A. Typed detection of quota/limit errors (core).**
Give developers (and the SDK) a non-opaque way to recognize this. Either:
- introduce `APIQuotaExceededError(APIStatusError)`, constructed when the response body has `type == "inference_quota_exceeded"` (carrying `quota_type`, `category`, `hint`, `remaining_limit`); or
- at minimum, guarantee the OpenAI/inference plugin populates `APIError.body` with the decoded JSON and document the field, so `ev.error.body["type"]` / `["hint"]` are reliable.

**B. Default user-perceptible surfacing (opt-in, sensible default).**
On an **unrecoverable** LLM error with audio enabled, optionally speak a fallback line before closing — e.g. an `AgentSession` option `on_unrecoverable_error` (callback) or `error_message: str | None`. For the quota case, the gateway's `hint` is already caller-appropriate. Keep backwards compatible (default could be a generic "this assistant is temporarily unavailable").

**C. Forward a structured signal to the room/frontend.**
So frontends (incl. Agent Builder preview) can render an "out of credits / quota exceeded" state instead of dead air — e.g. surface the `error`/`close` events over the wire or publish a data/attribute event carrying `type`, `quota_type`, `hint`. This is what the forum reporter actually needed.

**D. Don't silently absorb a terminal error 3×.**
`max_unrecoverable_errors=3` is right for transient errors but wrong for known-terminal ones. For `inference_quota_exceeded` (and other non-retryable quota categories), surface/close on the **first** occurrence rather than after three silent dead turns.

**E. Docs + example (zero-risk, do immediately).**
Document the `inference_quota_exceeded` body shape and the `@session.on("error")` + `session.say(...)` recipe; add an example. Extend the existing 429 test (`tests/test_interruption/test_interruption_failover.py::test_retries_then_emits_unrecoverable`) to assert the new surfaced/spoken path.

---

## Acceptance criteria

- [ ] `429 inference_quota_exceeded` is detectable via a typed error or documented `body` field — not an opaque blob.
- [ ] By default, or via a one-line documented option, the end user gets a perceptible signal (spoken message and/or frontend-renderable event) — never pure silence.
- [ ] A terminal quota error surfaces on the **first** occurrence, not after `max_unrecoverable_errors`.
- [ ] Agent Builder preview renders an "out of credits / quota exceeded" state for this case.
- [ ] Docs include the recipe + body contract; an example demonstrates it; a test covers 429-quota → surfaced.

---

## Workaround for users today

```python
from livekit.agents import ErrorEvent

@session.on("error")
def _on_error(ev: ErrorEvent):
    err = ev.error
    if getattr(err, "recoverable", True):
        return
    body = getattr(err, "body", None) or {}
    is_quota = (
        getattr(err, "status_code", None) == 429
        or (isinstance(body, dict) and body.get("type") == "inference_quota_exceeded")
    )
    if is_quota:
        msg = (body.get("hint") if isinstance(body, dict) else None) \
            or "Sorry — this assistant is temporarily unavailable due to usage limits. Please try again later."
        session.say(msg)
        # optionally: publish a data/attribute event so the frontend can render an "out of credits" state
```

---

## References

- LiveKit community forum: LLM out-of-credits → "agent joins but never speaks" (staff-confirmed depleted inference credits): https://community.livekit.io/t/1282
- Gateway 429 contract (for maintainer reference): structured body built in `agent-gateway` `pkg/quota/response.go` (`QuotaExceededResponse`) / `pkg/http/error.go` (`JSONErrorResponse`); returned from `pkg/handler/completions.go` on `QuotaStatusExceeded`.
- SDK code paths (v1.5.2): `_exceptions.py:48-96`, `llm/llm.py:211-259`, `voice/events.py:222-245`, `voice/agent_activity.py:1314-1345` & `:2410`, `voice/agent_session.py:129` & `:1363-1401`.

shawnfeldman:
## Summary

Fixes #6009 — a voice agent goes **silently unresponsive** when the LLM endpoint returns `HTTP 429 {"type": "inference_quota_exceeded", ...}` (e.g. the project is out of LiveKit Inference credits). Today the agent joins the room, publishes its track, and then never speaks: STT transcribes fine, the reply produces no text, TTS is never invoked, and the session absorbs three silent dead turns before closing with no audible or visible signal. The gateway hands us everything we need (`type`, `hint`, `quota_type`) in the response body — we just never looked at it.

This PR makes that failure perceptible by default.

## What changed

**A. Typed detection** — new `APIQuotaExceededError(APIStatusError)` exposing `quota_type`, `category`, `hint`, and `remaining_limit` decoded from the gateway body. It defaults to `retryable=False` since quota exhaustion won't recover on an immediate retry. `create_api_error_from_http` and a `APIQuotaExceededError.from_response(...)` factory construct it whenever the body is an `inference_quota_exceeded` payload. Exported from `livekit.agents`.

**B. Plugin wiring** — the inference LLM plugin (also the base class for the OpenAI-compatible `livekit-plugins-openai` `LLMStream`) raises the typed error on a 429 quota body, so `ev.error` / `APIError.body` are reliable.

**C. Surface on the first occurrence** — `AgentSession._on_error` no longer absorbs a terminal quota error up to `max_unrecoverable_errors`. The 3-strike tolerance still applies to transient errors; a known-terminal quota error closes immediately.

**D. Perceptible signal, by default** — a new `AgentSession(error_message=...)` option speaks a fallback line just before the session closes on an unrecoverable error:
- **omit it (default):** quota errors speak the gateway `hint` (generic fallback if absent); other errors stay silent — no behavior change for the generic path.
- **`error_message="..."`:** speak that message on any unrecoverable error.
- **`error_message=None`:** disable spoken errors entirely.

Spoken delivery is best-effort (bounded by a timeout, wrapped so a failing TTS can't stall teardown). The `error` and `close` events still fire regardless, carrying the typed error for frontend handlers.

**E. Docs, example, tests** — docstrings on the new error + option, a [`quota_exceeded.py`](examples/voice_agents/quota_exceeded.py) example, and unit tests.

## Acceptance criteria

- [x] `429 inference_quota_exceeded` is detectable via a typed error (`APIQuotaExceededError`) and a documented `body` field.
- [x] By default (or via the one-line `error_message` option) the user gets a perceptible signal — never pure silence.
- [x] A terminal quota error surfaces on the **first** occurrence, not after `max_unrecoverable_errors`.
- [x] Docs include the recipe + body contract; an example demonstrates it; tests cover 429-quota → surfaced + spoken.
- [ ] Agent Builder preview rendering — out of scope here (first-party frontend); the typed `error`/`close` events now carry the structured info needed to render it.

## Test plan

- `uv run pytest tests/test_exceptions.py tests/test_unrecoverable_error.py --unit` — new tests for the typed error, the inference-plugin conversion, surface-on-first-occurrence, and the spoken fallback (default hint / generic / disabled / custom).
- `make check` (ruff format + lint + strict mypy) passes.
- Full `uv run pytest --unit` suite: **832 passed, 4 skipped** (the 9 `test_room.py` errors are pre-existing `FileNotFoundError` from a missing local server binary, unrelated to this change).

## Notes / follow-ups

- The typed error covers `quota_type` `llm`/`stt`/`tts`/`bargein`, but only the LLM HTTP path is wired today — the inference STT/TTS WebSocket paths don't surface the JSON body on connect, so they'd need separate handling.
- Forwarding the structured error over the wire for the Agent Builder preview is a separate, frontend-facing change.


examples/voice_agents/quota_exceeded.py
```diff
@@ -0,0 +1,80 @@
+import logging
```
chenghao-mou, line 1: I am not sure if we need an example for gateway errors. Maybe we should put this in docs instead?

livekit-agents/livekit/agents/_exceptions.py
```diff
@@ -100,6 +103,105 @@ def __repr__(self) -> str:
...
+        request_id: str | None = None,
+        body: object | None = None,
+        retryable: bool | None = None,
+        quota_type: str | None = None,
+        category: str | None = None,
+        hint: str | None = None,
+        remaining_limit: str | None = None,
+    ) -> None:
+        # quota exhaustion won't recover on an immediate retry
+        if retryable is None:
```
chenghao-mou, line 228: should we differentiate retry behaviour based on concurrency/RPM quotas vs credits difference?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -243,6 +253,7 @@ def __init__(
         ivr_detection: bool = False,
         user_away_timeout: float | None = 15.0,
         session_close_transcript_timeout: float = 2.0,
+        error_message: NotGivenOr[str | None] = NOT_GIVEN,
```
chenghao-mou, line 256: In the docs, we recommend using pre-recorded audio for errors like this: https://docs.livekit.io/reference/agents/events/#pre-recorded-audio 
so that out of quota from TTS won't block this.

Should we just introduce APIError.terminal in addition to retryable so users can use the same error handling easier (so that this behaviour is not limited to quota errors)?

shawnfeldman (commented):
## PR Review Summary

Reviewed the diff against the **actual gateway contract** in `agent-gateway` (`pkg/quota/`), the base LLM retry loop, and the existing reviewer comments. The shape is solid and the spoken-fallback machinery is carefully guarded; one classification issue is worth fixing before merge.

### Important

**1. Transient rate-limit 429s get misclassified as terminal + non-retryable (regression)**

`APIQuotaExceededError` defaults `retryable=False` for *every* `inference_quota_exceeded` body (`_exceptions.py:157-158`), and `AgentSession._on_error` treats *every* instance as `terminal`, closing on the first occurrence (`agent_session.py:1517`). But the gateway emits `type: inference_quota_exceeded` for two very different classes of 429:

| category | meaning | recovers? |
|---|---|---|
| `MaxGatewayCredits` (+ `MaxBargeInRequests`) | credit / quota exhausted | only at next billing cycle → **terminal** |
| `MaxConcurrentGatewayLLMRpm`, `MaxConcurrentGatewayLLMTpm` (+ STT/TTS/bargein concurrency) | per-minute rate / concurrency limit | within ~a minute via backoff → **transient** |

Authoritative source: `agent-gateway/pkg/quota/check.go:68-79` and `pkg/quota/response.go:55-77` — the hint for the rate-limit categories literally says *"Reduce request rate … or upgrade"*, vs *"Wait for the next billing cycle"* for credits.

The live impact today is on the **LLM path** (the only one wired): before this PR, a pre-stream 429 raised `APIStatusError(retryable=True)` (`inference/llm.py:395`; 429 is exempt from the 4xx→non-retryable rule in `APIStatusError.__init__`), so the base stream retried it with backoff (`llm/llm.py:215-262`, gated on `e.retryable`). After this PR a single `MaxConcurrentGatewayLLMRpm`/`…Tpm` spike is now (a) non-retryable and (b) kills the session permanently on the first turn — and speaks an "out of credits / temporarily unavailable" line for what is actually a recoverable rate-limit blip. That's exactly what @chenghao-mou flagged on `_exceptions.py`.

The class already knows about the split — the `category` docstring (`_exceptions.py:134-135`) calls out `MaxConcurrentGatewayLLMRpm` as a "rate-limit variant" — it just doesn't act on it. Suggested fix: derive `retryable`/terminal from `category` rather than from the body `type` alone. Only the credit/quota-exhaustion categories (`MaxGatewayCredits`, `MaxBargeInRequests`) should be `retryable=False` + terminal; the rate/concurrency categories should stay retryable and fall through the existing `max_unrecoverable_errors` tolerance. Then have `_on_error` key off that derived flag instead of `isinstance(..., APIQuotaExceededError)`.

### Suggestions

**2. The spoken fallback can't be heard when TTS itself is the exhausted resource.** When `quota_type == "tts"` (`MaxConcurrentGatewayTTS`), `_try_speak_error_message` tries to `say()` through the very resource that's rate-limited. Latent today (only the LLM path is wired, so `quota_type == "llm"` and the separate TTS still works), but it becomes real once the STT/TTS WS paths surface the body — your own follow-up note. As @chenghao-mou said, the [pre-recorded audio path](https://docs.livekit.io/reference/agents/events/#pre-recorded-audio) is the robust signal there; worth a code comment acknowledging the limitation.

**3. Echoing @chenghao-mou: consider a generic `APIError.terminal`.** Hardcoding `isinstance(error.error, APIQuotaExceededError)` in the session couples teardown policy to one exception subclass. A `terminal` attribute on `APIError` (default `False`) would be the natural home for the credits-vs-rate-limit bit from #1, and would also let other genuinely terminal errors (401 bad key, 400 bad request) short-circuit the tolerance loop without further session changes.

**4. Test the transient case.** Both new test files only exercise `MaxGatewayCredits` bodies — i.e. the case the PR handles correctly. Adding a `MaxConcurrentGatewayLLMRpm` body that asserts `retryable is True` / non-terminal is what would have caught #1.

**5. Example vs docs placement** (@chenghao-mou) — defer to maintainer convention; no strong opinion.

### Strengths

- Typed error is clean: body backfill, `from_response` factory, exported from `livekit.agents`, and explicit fields correctly take precedence over the body (nicely tested).
- Best-effort spoken fallback is well-guarded — 16s playout timeout + `try/except` + `_activity`/`output.audio` None checks mean a failing TTS can't stall teardown; `error`/`close` still fire regardless.
- Default path is backwards-compatible (silent unless a quota error or an explicit `error_message`).
- Surfacing true credit exhaustion on the *first* occurrence instead of after 3 dead turns is the right call; docstrings, example, and tests are thorough and `make check` is green.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code) — gateway contract cross-checked against `agent-gateway/pkg/quota`.

livekit-agents/livekit/agents/_exceptions.py
shawnfeldman: Done in e7357c6 — `APIQuotaExceededError` now derives `retryable`/`terminal` from `category` (verified against `agent-gateway/pkg/quota/response.go::quotaHint`):

- **Terminal + non-retryable:** credit exhaustion — `MaxGatewayCredits`, `MaxBargeInRequests` ("Wait for the next billing cycle…").
- **Retryable + non-terminal:** rate/concurrency limits — `MaxConcurrentGatewayLLMRpm/Tpm`, `MaxConcurrentGatewaySTT/TTS`, `MaxBargeInRPM`, etc. (and any unknown/missing category, to avoid regressions).

So a transient rate-limit 429 is now retried with backoff by the stream and falls through `max_unrecoverable_errors` exactly as before this PR; only true credit exhaustion closes on the first turn. Added tests for both classes (retried vs terminal).

livekit-agents/livekit/agents/voice/agent_session.py
shawnfeldman: Both addressed in e7357c6:

**`APIError.terminal`** — added a generic `terminal: bool = False` on `APIError` (independent of `retryable`: `retryable` governs in-request retries, `terminal` governs whether higher-level loops should give up at once). `AgentSession._on_error` now keys off `error.error.terminal` instead of `isinstance(..., APIQuotaExceededError)`, so any terminal error (incl. future 401 bad-key / 400 bad-request cases) can short-circuit the `max_unrecoverable_errors` tolerance without further session changes. `APIQuotaExceededError` sets it from `category`.

**Pre-recorded audio** — documented the limitation: the `error_message` fallback synthesizes through the configured TTS, so it can't be heard if TTS itself is the exhausted/failed resource. The `error_message` docstring and a code comment in `_try_speak_error_message` now point to `@session.on("error")` + `session.say(..., audio=...)` with pre-recorded audio (linking the events docs) for a TTS-resilient signal.

examples/voice_agents/quota_exceeded.py
shawnfeldman: Fair point. For now I've kept the example but fixed a real bug in it (it used `ev.error` where it needs `ev.error.error` — `ErrorEvent.error` is the `LLMError`/`STTError` wrapper, so the `isinstance` guard was always False). The minimal `@session.on("error")` recipe also lives in the `APIQuotaExceededError` docstring.

Happy to delete the example and move it to the docs site instead if you'd prefer that — just say the word and I'll drop `examples/voice_agents/quota_exceeded.py` + its README entry.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1490,14 +1511,20 @@ def _on_error(
...
+
+        if not terminal:
+            if error.type == "llm_error":
+                self._llm_error_counts += 1
+                if self._llm_error_counts <= self.conn_options.max_unrecoverable_errors:
+                    return
+            elif error.type == "tts_error":
+                self._tts_error_counts += 1
+                if self._tts_error_counts <= self.conn_options.max_unrecoverable_errors:
+                    return
```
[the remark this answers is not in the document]
shawnfeldman: Correct that this is pre-existing — `stt_error` and `realtime_model_error` fall through to close on the first unrecoverable occurrence, and this PR preserves that (the new `terminal` short-circuit sits in front of the unchanged llm/tts counters). It's arguably intentional: an STT stream persists for the whole agent lifetime (unlike LLM/TTS, which are recreated per response), so it isn't recreated to "retry" the way the tolerance assumes.

This PR does make a clean future fix possible: `_on_error` now keys off the generic `APIError.terminal` flag, so extending `max_unrecoverable_errors` tolerance to STT/realtime (or marking specific errors terminal) is a localized change. Leaving it out of scope here to keep the diff focused on the quota issue.

shawnfeldman:
Thanks for the thorough review (and the gateway-contract cross-check) — pushed e7357c6 addressing it.

### Fixed

**1. Credit-vs-rate-limit regression (the important one).** The gateway returns `inference_quota_exceeded` for two different conditions, and the first cut treated them identically (non-retryable + terminal), which would kill the session on turn 1 and speak "out of credits" for a recoverable rate-limit blip. `APIQuotaExceededError` now derives `retryable`/`terminal` from `category` (verified against `agent-gateway/pkg/quota/response.go::quotaHint`):

| category | class | retryable | terminal |
|---|---|---|---|
| `MaxGatewayCredits`, `MaxBargeInRequests` | credit exhaustion | ❌ | ✅ |
| `MaxConcurrentGatewayLLMRpm/Tpm`, `…STT/TTS`, `MaxBargeInRPM`, … | rate / concurrency | ✅ | ❌ |
| unknown / missing | (safe default) | ✅ | ❌ |

So transient limits are retried with backoff and fall through `max_unrecoverable_errors` exactly as before; only true credit exhaustion surfaces on the first turn.

**2. Generic `APIError.terminal`.** Added `terminal: bool = False` on `APIError` (independent of `retryable`). `AgentSession._on_error` keys off `error.error.terminal` instead of `isinstance(..., APIQuotaExceededError)`, so any terminal error can short-circuit the tolerance loop in future without further session changes.

**3. `ev.error` → `ev.error.error`.** `ErrorEvent.error` is the `LLMError`/`STTError`/… wrapper; the underlying exception is at `.error`. The docstring example and `quota_exceeded.py` had the guard one level too shallow (always False) — fixed both.

**4. TTS-quota / pre-recorded audio.** Documented that the spoken fallback synthesizes through TTS (so it can't be heard if TTS itself is the failed resource); the docstring + a code comment point to the `@session.on("error")` + `say(..., audio=...)` pre-recorded-audio recipe.

### Tests added
Rate-limit body is retryable + non-terminal and is retried by the stream / tolerated by the session; unknown category defaults to transient; custom `error_message` overrides the hint; realtime quota path; spoken fallback survives a failing TTS; no-audio-output path. `make check` green; full `--unit` suite passes (the `test_room.py` errors are a pre-existing missing-binary issue).

### Open question
- **Example vs docs placement** (@chenghao-mou): kept + fixed for now; glad to move it to the docs site and drop the example file if you prefer.
- **STT/realtime first-occurrence close** (Devin): pre-existing and arguably intentional (STT streams persist for the agent lifetime); the new `terminal` flag makes a future tolerance change localized. Left out of scope here.

livekit-agents/livekit/agents/_exceptions.py
```diff
@@ -142,6 +297,12 @@ def create_api_error_from_http(
     else:
         display = f"{reason} ({status})"
 
+    quota_error = APIQuotaExceededError.from_response(
+        display, status_code=status, request_id=request_id, body=body
+    )
+    if quota_error is not None:
+        return quota_error
```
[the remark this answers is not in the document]
shawnfeldman: Confirmed and documented in 8411107. Traced the full path:

- The gateway's `limitsMiddleware` rejects a quota-exhausted STT/TTS connection **pre-upgrade** (`agent-gateway/pkg/middleware/limits.go` → `HandleJSONError`) with a JSON 429 body, setting only `Content-Type` (no useful headers).
- On a failed handshake aiohttp raises `WSServerHandshakeError`, which on 3.14 exposes only `status`/`message`/`headers` — the response **body is discarded**. So there's no `body=` to pass at the connect site.

So it's a genuine limitation, but there's a sharper reason to leave STT/TTS as a plain (retryable) `APIStatusError` rather than guess: without the body's `category` the SDK **can't tell terminal credit exhaustion from a transient rate limit**. Forcing terminal/non-retryable on every STT/TTS 429 would reintroduce exactly the rate-limit regression fixed earlier in this PR; leaving it untyped keeps the safe (retryable → existing tolerance) behavior.

Added a code comment at both connect sites (`inference/tts.py`, `inference/stt.py`) explaining this so it's discoverable. A real fix would need the gateway to surface the category another way (e.g. a response header aiohttp keeps, or a post-upgrade close frame) — tracked as future work.

shawnfeldman (commented):
## PR Review Summary

Reviewed with 6 specialized agents (code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer, comment-analyzer, code-simplifier); Critical/Important findings were independently verified against the gateway source and the pinned openai SDK before posting.

### Critical Issues (must fix)

1. **[code-reviewer, silent-failure-hunter, pr-test-analyzer — independently converged; verified empirically] The production quota path never fires: `e.body` is a `str`, not the quota dict** — `livekit-agents/livekit/agents/inference/llm.py:461-471`
   The openai SDK's `_make_status_error` (verified in 2.30.0, `openai/_client.py:809`) unwraps the body before constructing the exception: `data = body.get("error", body) if is_mapping(body) else body`. The gateway's 429 payload always carries a top-level `"error"` key whose value is a **string** (`agent-gateway/pkg/http/error.go` — `Error string`, no omitempty; set from `q.Error.Error()` in `pkg/quota/response.go::QuotaExceededResponse`). Reproduced with a gateway-shaped 429 through the real SDK:
   ```
   exception class: RateLimitError
   e.body type:     str   ('LLM token credit quota exceeded, category: MaxGatewayCredits, ...')
   isinstance dict: False
   re-parse works:  e.response.json()["type"] == "inference_quota_exceeded"
   ```
   So `from_response(body=e.body)` hits `isinstance(body, dict) → False` and returns `None`: every real quota 429 degrades to a plain retryable `APIStatusError` → retried → absorbed under `max_unrecoverable_errors` → the exact silent dead-turns of #6009 persist, while all new tests pass. The tests mask this by hand-constructing `openai.APIStatusError(body["error"], response=response, body=body)` with the full dict (`tests/test_exceptions.py:162-163`), bypassing the SDK's unwrapping that real traffic goes through.
   **Fix:** recover the full body before detection:
   ```python
   except openai.APIStatusError as e:
       body: object = e.body
       if not isinstance(body, dict):
           try:
               body = e.response.json()
           except Exception:
               body = e.body
       quota_error = APIQuotaExceededError.from_response(
           e.message, status_code=e.status_code, request_id=e.request_id, body=body
       )
   ```
   And make the test fixture go through the SDK path so this regression class is caught, e.g. `openai.AsyncOpenAI(api_key="x")._make_status_error_from_response(httpx.Response(429, json=quota_body, request=...))`, or swap in a real client over `httpx.MockTransport` (httpx already imported, no new dep).

### Important Issues (should fix)

2. **[type-design-analyzer — verified] Unvalidated wire-data extraction can crash inside the error handler** — `livekit-agents/livekit/agents/_exceptions.py:205-217`
   The four `body.get(...)` reads are unvalidated `Any`. An unhashable `category` (e.g. a list) raises `TypeError` at the frozenset membership test — inside the `except openai.APIStatusError` block, degrading the typed path to a generic error. Non-str values (int `remaining_limit`, list `hint`) silently violate the `str | None` annotations (strict mypy can't see it) and a non-str `hint` can flow into `session.say()`. The gateway sends strings today, but `inference.LLM(base_url=...)` is user-pointable. Fix (~4 lines): coerce each field, e.g. `quota_type = v if isinstance(v := body.get("quota_type"), str) else None`.

3. **[comment-analyzer, code-simplifier — both found independently] The example's copy-paste snippet is a SyntaxError, and its handler mislabels transient errors** — `examples/voice_agents/quota_exceeded.py`
   - Lines 69-72: the commented `await ctx.room.local_participant.set_attributes(...)` sits inside the **sync** `def on_error` handler — uncommenting it is `SyntaxError: 'await' outside async function`. Show `asyncio.create_task(...)` (the pattern `error_callback.py` already uses).
   - Line 57: `# quota errors are non-retryable; they will fail identically every turn` — wrong for half the class. Rate-limit-category `APIQuotaExceededError`s are retryable/non-terminal, and this handler **does** receive them (the LLM base emits an `error` event per retry attempt; `agent_activity._on_error` forwards every `LLMError` unconditionally). As written, the example logs/forwards "out of credits" for transient rate limits. Gate on `err.terminal` and fix the comment.
   - Lines 25-26: "By default such an error makes the agent join the room ... and then never speak" — stale; this PR changed that default, and it contradicts bullet (1) three lines later. Rephrase as before/after.

4. **[comment-analyzer — verified] Comment/docstring claims that contradict the code** —
   - `_exceptions.py:155-156`: "By default ``AgentSession`` already speaks the ``hint`` and closes on the first occurrence" — only true for **terminal** instances with `error_message` unset; a transient rate-limit instance of this same class does neither. Add the qualifier.
   - `inference/llm.py:462-463`: "surface it as a typed, **non-retryable** error" — wrong for rate-limit bodies; the PR's own `test_inference_llm_retries_rate_limit_429` asserts the retry. Reword.
   - `voice/agent_session.py:345`: `https://docs.livekit.io/reference/agents/events/#pre-recorded-audio` appears to be a guessed deep link — no `docs.livekit.io/reference/...` URL exists anywhere in the repo (the established pattern is `docs.livekit.io/agents/...`). Replace or drop before it 404-rots in a public docstring.

### Suggestions

5. **[silent-failure-hunter] Non-terminal closes still end in dead air by default** — `agent_session.py:1561-1567`. Sustained rate-limiting (or any generic LLM/TTS error) past the tolerance closes the session with nothing spoken — same end-user symptom as #6009. That matches the PR's stated scope ("other errors stay silent"), but the inline comment's justification ("speaking ... for a recoverable blip would be misleading") is wrong in context: `_resolve_error_message` is only reached when the session is irrevocably closing — the blip case never gets there. Either speak `DEFAULT_ERROR_MESSAGE` on any error-close by default, or fix the comment and note the limitation in the `error_message` docstring.
6. **[silent-failure-hunter] The speak-path warning can't fire in its dominant failure mode** — `agent_session.py:1590-1596`. `SpeechHandle._done_fut` is only ever resolved with `set_result(None)`, so a TTS that fails synthesis (e.g. the same depleted project) returns from `wait_for_playout()` normally — the `except Exception` log never fires, and the subsequent `TTSError` is dropped by the `_closing_task` guard. Log at info before speaking; log at warning when `_on_error` drops an error while `_closing_task` is set.
7. **[silent-failure-hunter] Document the STT-first ordering caveat** — for a fully-depleted all-inference project (the configuration the example itself promotes), the **STT** WS connect 429s first, STT errors have no tolerance counter, and the session closes untyped/silent within seconds — before the LLM path and the spoken hint ever engage. The stt.py/tts.py NOTEs are accurate per-path, but this end-to-end consequence is documented nowhere; worth a sentence in the `error_message` docstring and the example.
8. **[pr-test-analyzer] Test gaps worth closing:** the 16s `_ERROR_MESSAGE_PLAYOUT_TIMEOUT` is never exercised (a hanging-TTS test with the constant monkeypatched would pin it, rated 7/10); tolerance accumulation is unpinned (two errors with `max_unrecoverable_errors=1`, 6/10); `error_message=""` silently behaves like `None`, contradicting the docstring (6/10 — document or normalize); `_spy_say` discards kwargs so `add_to_chat_ctx=False` is unpinned (a regression would write error messages into chat history).
9. **[type-design-analyzer] Polish:** add `terminal` to `APIError.__repr__`/`APIStatusError.__repr__` (debugging "why did my session close on turn 1" wants it); one docstring line that `from_response` detection is body-keyed (status intentionally ignored); decide `terminal` vs `permanent` naming now — free before merge, impossible after in a public SDK (keeping `terminal` is defensible).
10. **[code-simplifier] Tests:** build `_rate_limit_error`'s body as `{**INFERENCE_QUOTA_BODY, "category": ..., "hint": ...}` so the category delta is visible by construction; optionally extract the repeated emit-error/await-close plumbing (~6 lines × 9 tests) into two helpers, matching `test_agent_session.py` style.

### Strengths

- The category→terminal/retryable taxonomy is exactly right per the gateway source (`MaxGatewayCredits`/`MaxBargeInRequests` terminal; everything else transient), and unknown categories failing open to transient is the correct version-skew posture — pinned by a dedicated test.
- The session-level tests are genuinely behavioral (close events, reason, error identity, exact spoken text) and are the first coverage ever for the `_on_error` close path, including the realtime path and TTS-failure-during-farewell resilience.
- `_close_on_error`'s try/finally + bounded playout + `_aclose_impl`'s force-interrupt means teardown is provably bounded even with a wedged TTS; concurrent `aclose()` is serialized safely.
- The stt.py/tts.py NOTE comments are fully accurate (verified against aiohttp's handshake behavior and the gateway middleware), and the `APIQuotaExceededError` docstring with a working copy-pasteable example is unusually good.
- `APIError.terminal` as a class attribute keeps reads safe on third-party subclasses that predate the PR — nice backward-compat detail.

shawnfeldman:
Pushed 4b8245102 addressing the Critical + Important findings. `make check` green; full `--unit` suite 842 passed / 4 skipped (same pre-existing `test_room.py` missing-binary errors).

### Fixed

**1. `e.body` is a string in production (the Critical).** Confirmed: the openai SDK's `_make_status_error` does `body.get("error", body)` before constructing the exception, and the gateway's `error` field is a bare string — so the typed quota path never fired on real traffic. The plugin now recovers the full payload via `e.response.json()` when `e.body` isn't a dict. The fixture in `tests/test_exceptions.py` now builds the 429 through the SDK's own `_make_status_error_from_response` (with a pinned `assert not isinstance(status_error.body, dict)`), so it exercises the narrowing real traffic goes through — verified it fails against the unfixed plugin exactly the way production did (string body → plain retryable 429 → retried → absorbed silently).

**2. Wire-data coercion.** `APIQuotaExceededError` now drops non-str body fields (`_str_or_none`) — an unhashable `category` raised `TypeError` at the frozenset check, inside the except handler, and non-str values violated the `str | None` annotations. The endpoint is user-pointable, so the body is untrusted. New test pins it (unhashable category → fields `None`, stays transient).

**3. Example.** `on_error` now gates on `err.terminal` (the handler also sees transient rate-limit/retry events — plain `isinstance` would report "out of credits" for a recoverable blip); rewrote the stale "by default … never speak" intro as before/after; the commented `set_attributes` snippet now spawns a task instead of `await`ing inside the sync handler.

**4. Doc accuracy.** Scoped the `APIQuotaExceededError` docstring claim to terminal errors; reworded the "non-retryable" comment in `inference/llm.py`; the pre-recorded-audio docs link was actually the right page (the same one linked above) with a wrong fragment — fixed to `#pre-recorded-audio-fallback`.

### Pushing back on the suggestions

- **Speak on every unrecoverable close by default** — out of scope on purpose. That flips the default for every error class and every existing app (including ones already rendering their own error UX off the `error` event), not just the silent-quota case this PR targets. And on a transient-quota close, the only message available is the gateway `hint` ("Reduce request rate…"), which is developer-facing — the wrong thing to speak at an end user. `error_message="..."` is the supported one-liner for exactly that behavior.
- **Speak-path logging** — the TTS base already warns per failed synthesis attempt, and `_try_speak_error_message` warns with traceback on timeout/sync failure; a session-level "speaking error message" info line is fine polish for a follow-up, not load-bearing here.
- **STT-first ordering caveat in the docstring** — the connect-site NOTEs in `inference/stt.py`/`tts.py` are the canonical home for that mechanism and already describe it; copying gateway middleware-ordering details into the `error_message` docstring bakes server behavior into API docs that will drift. Typed STT/TTS quota errors stay the "future work" called out in the description.
- **Hanging-TTS timeout / tolerance-accumulation tests** — agreed they'd be nice; the can't-stall invariant is already enforced by `asyncio.wait_for` + `_aclose_impl`'s force-interrupt and covered by `test_speaking_error_message_survives_tts_failure`. Follow-up material rather than gating.
- **`terminal` vs `permanent` naming** — keeping `terminal`: it's the name requested in review above, and it describes the consumer contract (the session gives up at once) rather than the error's persistence.
- Smaller polish (`terminal` in `__repr__`, documenting `error_message=""`, test-helper extraction) — happy to fold into a follow-up; skipped here to keep the diff reviewable.

examples/voice_agents/quota_exceeded.py
chenghao-mou: On a second thought, we can leave it here so we have some example for other error handling as well. This is hard to document for all other vendors.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1510,11 +1543,58 @@ def _on_error(
...
+
+        # default: only speak for a terminal quota error, preferring the gateway hint.
+        # Transient errors (incl. rate limits) stay silent — speaking "out of credits"
+        # for a recoverable blip would be misleading.
+        if isinstance(error, APIQuotaExceededError) and error.terminal:
+            return error.hint or DEFAULT_ERROR_MESSAGE
+
+        return None
+
+    async def _close_on_error(
```
chenghao-mou, line 1569: maybe instead of adding all these internal functions, we can add a default error hook (like what customers would do) if an error message is given?

PR #6013, opened 2026-06-08
(core): add function_tools_called event

tinalenguyen:


longcw:
I created a separate event for tool started, updated, and reply events https://github.com/livekit/agents/pull/6100

PR #6035, opened 2026-06-09
feat(openai): add LiveKit inference realtime model

russellmartin-livekit:
## Summary
- add an OpenAI-compatible realtime model authenticated through LiveKit Inference
- refresh LiveKit access tokens and inference headers for each websocket connection
- expose the minimal realtime session seams needed for gateway metadata and terminal error handling
- preserve realtime calls and session state across failover and reconnects
- add xAI-compatible defaults for `xai/*`: `eve`, `grok-transcribe`, and server VAD, while preserving explicit overrides
- preserve the framework ability to disable default xAI server VAD when client-side VAD is used
- retry after per-message `invalid_audio_payload` errors while keeping unsupported session configuration terminal

## Scope
This PR contains the SDK-side OpenAI-compatible realtime client and the xAI inference defaults formerly tracked in [#7093](https://github.com/livekit/agents/pull/7093). The xAI provider implementation and protocol normalization remain in [agent-gateway #1262](https://github.com/livekit/agent-gateway/pull/1262).

## Test plan
- [x] `uv run pytest tests/test_realtime/test_openai_inference_realtime_model.py tests/test_realtime/test_openai_realtime_model.py --unit`
- [x] OpenAI plugin type checking
- [x] `uv run ruff format --check` on changed Python files
- [x] `uv run ruff check` on changed Python files


shawnfeldman (approved):
Reviewed the seams in `realtime_model.py` and the new subclass. No blockers.

**Verified**
- The two `realtime_model.py` changes are behavior-preserving for existing OpenAI/Azure users: `_create_ws_url_and_headers()` is a byte-identical extraction from `_create_ws_conn()`, and the three `_is_fatal_error` call sites just route through `self.`.
- The `_wrap_session_update` override is in-contract — the base already declares `SessionUpdateEvent | dict[str, Any]` for the legacy-Azure path, uses the *same* `model_dump(by_alias=True, exclude_unset=True, exclude_defaults=False)` flags, and both dict-sensitive call sites already `isinstance(event, dict)`-check (`_create_tools_update_event`) or accept dicts (`send_event`). Popping `session.model` hits the right nesting level.
- Header replacement is not an observability regression: `get_inference_headers()` sets `User-Agent: LiveKit Agents/{ver} (python {ver})`, which is strictly richer than the base's bare `"LiveKit Agents"`, plus the room/job/agent/session IDs.
- `Authorization` is assigned *after* `get_inference_headers()`, so a job's `ctx.inference_headers` can't overwrite the credential. Good ordering.
- Tests cover what matters: provider-prefix validation, the env-var fallback chain, per-connection token refresh (two distinct `Authorization` values across two connects), URL shape, and `"model" not in dumped["session"]`. Adding `_is_fatal_error=_is_fatal_error` to the existing fake is the right minimal fix for the refactor.

**Suggestions**
1. `_is_fatal_error` drops the base's `isinstance(code, str)` guard — module-level does `isinstance(code, str) and code in _FATAL_ERROR_CODES`, the override does bare `code in {...}`. Only bites if a vendor ever ships an unhashable `code` (set membership would raise rather than return False), but matching the base costs nothing.
2. `provider` and `inference_class` end up with *opposite* precedence against `ctx.inference_headers`: `inference_class` is passed into `get_inference_headers()` and so is overridable by the job (as `_utils` documents), while `HEADER_INFERENCE_PROVIDER` is set after the `headers.update(ctx.inference_headers)` and so the model config wins. Worth picking one.
3. `test_gateway_configuration_errors_are_fatal` only exercises the four new codes; a parametrize entry for one inherited code would also pin the `or super()._is_fatal_error(error)` branch.

Approving.

PR #6039, opened 2026-06-10
conversation-aware STT recognition (keyterms + chat context)

longcw:
Makes STT recognition conversation-aware via a new `stt_context_options` on `AgentSession`, grouping static keyterms with two independent, composable mechanisms for biasing recognition during a call.

## Overview

- `keyterms`: user-defined terms applied wherever the STT accepts a term list; never modified by detection.
- `keyterm_detection`: an LLM-based detector (`enabled`, `llm`, `turn_interval`, `max_keyterms`, `instructions`) that runs a background pass per user turn over the recent transcript and maintains the keyterm set with a confirmation gate — a new term starts `pending` and only biases the STT once later transcript evidence confirms it; `remove` applies only to spellings the user explicitly corrected.
- `chat_context`: native conversation-context carryover for STTs that consume context directly (no LLM), forwarding each conversation turn to the provider's own field (e.g. AssemblyAI u3-rt-pro `agent_context`).

## Details

- The detection prompt treats USER lines as untrusted STT output and ASSISTANT lines as authoritative spelling, and rejects misrecognitions: sound-alike variants of tracked terms, garbled phrases the assistant never adopts, and fragments from interrupted lines.
- Detection state is owned by the session so keyterms survive agent handoffs; user-defined terms are shown to the detection LLM as applied but never modified by it.
- New `STT._push_conversation_item()` hook and `chat_context` capability flag, alongside the existing `_update_keyterms()` / `keyterms` flag; both are forwarded by the fallback and stream adapters.
- `keyterms` capability implemented for deepgram (v1/v2), assemblyai, google, and livekit inference STT; `chat_context` implemented for assemblyai u3-rt-pro.


livekit-agents/livekit/agents/voice/keyterms.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def set_user_keyterms(self, terms: list[str]) -> None:
+        self._user_terms = list(dict.fromkeys(terms))
+        if self._stt is not None:
+            self._stt.update_keyterms(self.keyterms)
+
+    def start(self, session: AgentSession, *, stt: STT | None, llm: LLM | None) -> None:
+        """Bind this activity's STT/LLM and start detection (if enabled)."""
+        if self._session is not None or not self._options["enabled"]:
+            return
```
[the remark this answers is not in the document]
longcw: Fixed in 65cd307f1 — `start()` now binds the STT unconditionally and only gates the detection setup on `enabled`, so user-defined terms are pushed (and `set_user_keyterms()` works) without `detection.enabled`. The push is skipped when the term list is empty so sessions that don't use keyterms see no capability warning or reconnect. Added `test_user_terms_pushed_without_detection` and `test_start_without_terms_does_not_push`.

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -375,6 +376,10 @@ def update_options(
                 endpoint_url=endpoint_url,
             )
 
+    def update_keyterms(self, keyterms: list[str]) -> None:
+        # keyterm is supported by Nova-3 models; older models fall back server-side.
+        self.update_options(keyterm=keyterms)
```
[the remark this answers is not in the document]
longcw: This one looks like a false positive: Deepgram v1's `SpeechStream.update_options` does end with `self._reconnect_event.set()` (stt.py:505, also present on main at :500), and `STT.update_options` forwards `keyterm` to every active stream — so `update_keyterms` reconnects the live websocket on v1 the same way v2 does. No change needed.

examples/voice_agents/basic_agent.py
```diff
@@ -116,6 +116,11 @@ async def entrypoint(ctx: JobContext) -> None:
             "filter_markdown",
             text_transforms.replace({"LiveKit": "<<ˈ|l|aɪ|v>> <<ˈ|k|ɪ|t>>"}),
         ],
+        # automatically detect keyterms and apply them to the STT per user turn
+        keyterm_options={
+            "terms": ["LiveKit"],
+            "detection": {"enabled": True, "turn_interval": 1},
+        },
```
chenghao-mou, line 112: For conversations where some context/user information is available before the call (e.g from a patient/customer profile loaded when starting), should we allow extracting keyterms from such context first?

livekit-agents/livekit/agents/voice/keyterm_detection.py
```diff
@@ -0,0 +1,396 @@
...
+    """User-defined keyterms. Never modified by auto-detection."""
+    detection: KeytermDetectionOptions
+    """Automatic keyterm detection configuration."""
+
+
+_KEYTERM_DETECTION_DEFAULTS: KeytermDetectionOptions = {
+    "enabled": False,
+    "llm": None,
+    "turn_interval": 1,
+    "max_keyterms": None,
```
chenghao-mou, line 75: some of vendors impose a max limit, maybe we should check this in the extract-for-model function.

examples/voice_agents/basic_agent.py
longcw: perhaps keep that on the developer side, they can pass the context like address, user name via `keyterm_options={"terms": [...]}` once the profile loads.

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -375,6 +376,10 @@ def update_options(
                 endpoint_url=endpoint_url,
             )
 
+    def _update_keyterms(self, keyterms: list[str]) -> None:
```
chenghao-mou: Only the flux models can support mid-stream keyterm update: [https://developers.deepgram.com/docs/keyterm#dynamic-keyterm-updates-flux-only](<https://developers.deepgram.com/docs/keyterm#dynamic-keyterm-updates-flux-only>) should we disable this for nova models?

examples/voice_agents/basic_agent.py
```diff
@@ -102,10 +102,17 @@ async def entrypoint(ctx: JobContext) -> None:
             "filter_markdown",
             text_transforms.replace({"LiveKit": "<<ˈ|l|aɪ|v|k|ɪ|t>>"}),
         ],
+        # automatically detect keyterms and apply them to the STT per user turn
+        stt_context_options={
+            "keyterms": ["LiveKit"],
+            "keyterm_detection": {"enabled": True, "turn_interval": 1},
```
u9g: `"turn_interval": 1` is the default, probably not necessary to put it here.

u9g: also, I had to search the userbase to understand what turn_interval represents, maybe a better name is `detect_every_n_turns`.

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -192,6 +192,24 @@ def _diarization_enabled(extra_kwargs: dict[str, Any] | None) -> bool:
...
+    list, it doubles as a capability check (non-None ⇒ supported). Keep every provider's
+    keyterm key here so capability inference and _update_keyterms can't diverge.
+    """
+    if not (is_given(model) and isinstance(model, str)):
+        return None
+    if model.startswith("deepgram/"):
+        return {"keyterm": list(keyterms)}
+    if model.startswith("assemblyai/"):
+        return {"keyterms_prompt": list(keyterms)}
+    if model.startswith("speechmatics/"):
```
u9g, line 217: would it make sense to add STTCapabilities.keyterms to the speechmatics plugin?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1073,6 +1093,7 @@ async def _close_session(self) -> None:
         if isinstance(self.stt, stt.STT):
             self.stt.off("metrics_collected", self._on_metrics_collected)
             self.stt.off("error", self._on_error)
+            self._session.off("conversation_item_added", self.stt._push_conversation_item)
```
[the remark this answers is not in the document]
longcw: `off` uses `discard` won't raise ValueError

livekit-agents/livekit/agents/voice/keyterm_detection.py
```diff
@@ -0,0 +1,466 @@
...
+
+    Returns ``(pending, confirm, remove)``.
+    """
+    current = current_keyterms or []
+    user_msg = _format_input(chat_ctx, current)
+    if user_msg is None:  # no transcript yet — nothing to detect
+        return [], [], []
+    req_ctx = ChatContext.empty()
+    req_ctx.add_message(role="system", content=instructions or _DEFAULT_KEYTERM_INSTRUCTIONS)
+    req_ctx.add_message(role="user", content=user_msg)
```
toubatbrian, line 380: small nit: should we format `user_msg` in a prefix caching friendly way so this detector LLM can run faster?

livekit-agents/livekit/agents/voice/keyterm_detection.py
```diff
@@ -0,0 +1,466 @@
...
+    turns: list[str] = []
+    for item in reversed(chat_ctx.items):
+        if not isinstance(item, llm_module.ChatMessage) or item.role not in ("user", "assistant"):
+            continue
+        if text := item.text_content:
+            # keep the message's line structure but drop blank lines, so the blank line
+            # between turns is the only blank line and reliably marks a turn boundary
+            body = "\n".join(line for line in text.splitlines() if line.strip())
+            turns.append(f"{item.role.upper()}: {body}")
+            if len(turns) >= _MAX_TRANSCRIPT_MESSAGES:
```
toubatbrian, line 411: should this be configurable?

livekit-agents/livekit/agents/voice/keyterm_detection.py
```diff
@@ -0,0 +1,466 @@
...
+        ):
+            return
+
+        self._turn_count += 1
+        if self._turn_count % self._turn_interval != 0:
+            return
+
+        # single-flight: skip while a pass is still running
+        if self._detect_task is not None and not self._detect_task.done():
+            return
```
toubatbrian, line 286: nit: maybe we can chain this task off `self._detect_task.add_done_callback` here?

Main concern is that we probably don’t want to miss a critical detection turn. For example, if turn_interval is 3 and the previous LLM call somehow doesn’t finish or gets stuck due to a failure, then skipping here would leave us with no detection for a total of 6 turns.

Maybe we could consider adding a timeout inside `_detect_keyterms`?

livekit-agents/livekit/agents/voice/keyterm_detection.py
longcw: the instructions is already fixed, but the turns added to the user message is a sliding window that will change in every request. and I think the total token per request won't reach the caching threshold for many models.

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
longcw: now `update_options` is called after `END_OF_SPEECH` event, to avoid reconnection in the middle of the turn. meanwhile, the audio during reconnection is buffered in `input_ch` so ideally it won't missing user transcript.

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -637,10 +683,37 @@ def update_options(
...
+            return
+
+        self._session_keyterms = list(keyterms)
+        # inference applies extra live via session.update; defer to END_OF_SPEECH since the
+        # gateway may reconnect upstream when the keyterms change
+        for stream in self._streams:
+            if stream._speaking:
+                stream._pending_extra = keyterm_extra
+            else:
+                stream.update_options(extra=keyterm_extra)
```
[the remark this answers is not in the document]
longcw: original `update_options` is sending delta already

livekit-agents/livekit/agents/stt/stream_adapter.py
```diff
@@ -42,6 +47,12 @@ def model(self) -> str:
     def provider(self) -> str:
         return self._stt.provider
 
+    def _update_session_keyterms(self, keyterms: list[str]) -> None:
+        self._stt._update_session_keyterms(keyterms)
+
+    def _push_conversation_item(self, item: ConversationItemAddedEvent) -> None:
+        self._stt._push_conversation_item(item)
```
theomonnom, line 54: Should it be a `update_chat_ctx` instead? Seems more consistent?

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -231,6 +233,7 @@ def __init__(
         llm: NotGivenOr[llm.LLM | llm.RealtimeModel | LLMModels | str] = NOT_GIVEN,
         tts: NotGivenOr[tts.TTS | TTSModels | str] = NOT_GIVEN,
         turn_handling: NotGivenOr[TurnHandlingOptions] = NOT_GIVEN,
+        stt_context_options: NotGivenOr[STTContextOptions] = NOT_GIVEN,
```
theomonnom: I think we should name this feature `keyterms_options` directly inside the `AgentSession`

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -565,6 +578,11 @@ def conn_options(self) -> SessionConnectOptions:
     def history(self) -> llm.ChatContext:
         return self._chat_ctx
 
+    @property
+    def keyterms(self) -> list[str]:
+        """The effective keyterms (user-defined + auto-detected) currently applied to the STT."""
+        return self._keyterm_detector.keyterms
```
theomonnom, line 583: I think we don't need a new property if it's inside AgentSessionOptions

theomonnom (approved):
lgtm!

livekit-agents/livekit/agents/voice/keyterm_detection.py
longcw: added a timeout to _detect_keyterms

livekit-agents/livekit/agents/stt/stream_adapter.py
longcw: using push because each call carries a single chat item as it's added and feeds it incrementally, `update_chat_ctx` feels like it's a replacing.

livekit-agents/livekit/agents/voice/agent_session.py
longcw: this is the keyterms applied to the STT (including the detected terms), expose it here in case user want to save the keyterms.

livekit-agents/livekit/agents/voice/agent_session.py
longcw: what about the chat_ctx options?

longcw: or we enable it if the STT.capability support chat context, and user can enable/disable it via STT args.

PR #6041, opened 2026-06-10
fix(deepgram): use stored language when validating model in update_options 

### LINKED ISSUE #6047: fix(deepgram): update_options ignores stored language when validating model compatibility
## Bug Description

`STT.update_options` and `SpeechStream.update_options` both call `_validate_model(model, language)` using the **caller's** `language` parameter. When only the model is updated, `language` is `NOT_GIVEN` — so `_validate_model` skips its English-only model guard entirely, allowing an incompatible model/language combination to be stored silently with no warning.

## Reproduction

```python
from livekit.plugins.deepgram import STT

stt = STT(api_key="...", language="fr")

# Should warn and fall back to nova-2-general (French + English-only model)
# Instead: silently accepts nova-2-meeting
stt.update_options(model="nova-2-meeting")

print(stt._opts.model)  # nova-2-meeting — wrong, should be nova-2-general
```

## Root Cause

In `stt.py` (lines 316 and 459):

```python
if is_given(model):
    self._opts.model = _validate_model(model, language)  # language is NOT_GIVEN here
```

`_validate_model` only warns when `is_given(language)` — so passing `NOT_GIVEN` silently bypasses the check, even if `self._opts.language` is already set to a non-English language.

## Fix

Use the stored language as fallback:

```python
self._opts.model = _validate_model(
    model, language if is_given(language) else (self._opts.language or NOT_GIVEN)
)
```

Same fix applies to both `STT.update_options` (line 316) and `SpeechStream.update_options` (line 459).

adityajha2005:
\`STT.update_options\` and \`SpeechStream.update_options\` both call \`_validate_model(model, language)\` using the caller's \`language\` parameter. When only the model is changed, \`language\` is \`NOT_GIVEN\`, so \`_validate_model\` skips its English-only model check entirely — allowing an incompatible model/language pair to be stored silently.

For example, after \`STT(language="fr")\`, calling \`update_options(model="nova-2-meeting")\` would accept \`nova-2-meeting\` (English-only) with no warning, instead of falling back to \`nova-2-general\`.

Fix: pass \`self._opts.language or NOT_GIVEN\` when the caller omits \`language\`, so the stored language is used for validation.

Fixes #6047

## Test plan

\`\`\`
uv run pytest tests/test_plugin_deepgram_stt.py --plugin deepgram -v
\`\`\`

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -313,7 +313,9 @@ class STT(stt.STT):
         if is_given(language):
             self._opts.language = LanguageCode(language)
         if is_given(model):
-            self._opts.model = _validate_model(model, language)
+            self._opts.model = _validate_model(
+                model, language if is_given(language) else (self._opts.language or NOT_GIVEN)
+            )
         if is_given(interim_results):
             self._opts.interim_results = interim_results
         if is_given(punctuate):
@@ -456,7 +458,9 @@ class SpeechStream(stt.SpeechStream):
         if is_given(language):
             self._opts.language = LanguageCode(language)
         if is_given(model):
-            self._opts.model = _validate_model(model, language)
+            self._opts.model = _validate_model(
+                model, language if is_given(language) else (self._opts.language or NOT_GIVEN)
+            )
         if is_given(interim_results):
             self._opts.interim_results = interim_results
         if is_given(punctuate):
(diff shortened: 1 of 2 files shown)


adityajha2005:
Hey @tinalenguyen @theomonnom  tagging you as you've both touched the Deepgram STT plugin recently. Would love a review when you get a chance!

The fix is in two symmetric spots (`STT.update_options` and `SpeechStream.update_options`) and covers the edge case where a user calls `update_options(model=...)` without re-specifying the language that was set at construction time.

tinalenguyen (approved):
lgtm, thank you for the PR!

PR #6046, opened 2026-06-10
fix(cli): restore terminal settings on console exit

### LINKED ISSUE #6045: Console mode leaves the tty in no-echo mode after ^C
### Bug Description

Stopping agents (e.g. the basic example agents) in "console" mode with Ctrl-C typically leaves the terminal in no-echo mode after exit.

### Expected Behavior

The terminal should be left in the same state at exit as it was at startup.

### Reproduction Steps

```bash
1. uv run examples/voice_agents/basic_agent.py console
2. Wait for it to start up, then press Ctrl-C to exit to your shell prompt
3. Note that your shell isn't echoing your keystrokes (run `reset` to fix it)
```

### Operating System

Ubuntu 24.04

### Models Used

_No response_

### Package Versions

```bash
livekit-agents==1.5.19.rc1
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
Snapshot tty settings, and restore at exit.
```

### Additional Context

_No response_

_No response_

kamalmostafa:
In console mode the key-listener runs in a daemon thread that puts the tty into no-echo mode. Ctrl-C tears down the process -- commonly while the key-listener is in mid-read so its per-read restore never gets to run.

The terminal is then left with echo disabled upon process exit.

Fix: Snapshot the tty settings at the start of _run_console and restore them in its finally, which always runs in the main thread on exit, regardless of what state the listener thread might leave the terminal in.

Fixes #6045

### DIFF
--- a/livekit-agents/livekit/agents/cli/_legacy.py
+++ b/livekit-agents/livekit/agents/cli/_legacy.py
@@ -1543,6 +1543,17 @@ def _run_console(
 
         c._validate_device_or_raise(input_device=input_device, output_device=output_device)
 
+        # Snapshot and restore tty settings, so we don't leave the terminal with echo disabled.
+        tty_fd: int | None = None
+        tty_settings: Any = None
+        try:
+            import termios
+
+            tty_fd = sys.stdin.fileno()
+            tty_settings = termios.tcgetattr(tty_fd)
+        except Exception:
+            tty_fd = None
+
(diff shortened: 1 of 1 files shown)


drasch:
love this, how can I help to get this merged? +1

longcw (approved):
thanks for the pr. looks good!

PR #6052, opened 2026-06-11
(sarvam stt): fix end of speech timing

tinalenguyen:
###   **Problem**

  The Sarvam streaming STT plugin tried to manufacture an audio-relative speech-end time from two sources that don't actually provide one:
  - The VAD END_SPEECH event's occured_at field — which logging proved is a wall-clock Unix epoch, not an audio-stream offset, so it was rejected 100% of the time by a range-check heuristic.
  - A local send-clock counter (_audio_position) that counts audio uploaded, not processed — biased and fabricated.

  Sarvam's streaming socket genuinely sends no usable word timing (no timestamps array; speech_start/speech_end come back null), so all this machinery produced misleading timestamps.

###   **Change**

  Aligned Sarvam with how every other STT plugin (Deepgram, AssemblyAI, Google, Azure…) handles this:

  1. END_OF_SPEECH is now emitted bare — no alternatives/end_time. Previously it carried an empty-text SpeechData with a fabricated end_time + unused speech_end_wall_time metadata.
  2. FINAL_TRANSCRIPT timing comes only from the provider — start_time/end_time read from speech_start/speech_end, falling back to 0.0. The voice pipeline then uses wall-clock for EOU timing (standard behavior for providers without streaming word timing).
  3. Deleted the dead machinery: _interpret_signal_time, _resolved_speech_end, the _audio_position send-clock (+ its hot-loop increment), _utterance_server_speech_end, _utterance_speech_end_audio_pos, _utterance_speech_end_wall, and the require_end_time param.

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
@@ -921,6 +921,12 @@ class SpeechStream(stt.SpeechStream):
 
         self._audio_position = 0.0
         self._utterance_start_audio_pos = 0.0
+        # Authoritative speech-end (audio-stream seconds) as measured by Sarvam:
+        # the END_SPEECH `occured_at` or the transcript `speech_end`. Preferred
+        # for all timing.
+        self._utterance_server_speech_end: float | None = None
+        # Fallback only: the send clock at END_SPEECH. Biased high (it counts audio
+        # uploaded, not processed), so used only when Sarvam gives no usable time.
         self._utterance_speech_end_audio_pos: float | None = None
         self._utterance_speech_start_wall: float | None = None
         self._utterance_speech_end_wall: float | None = None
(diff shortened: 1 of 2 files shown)


theomonnom (approved):

livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
```diff
@@ -1012,33 +1005,21 @@ def _cancel_eos_fallback(self) -> asyncio.Task[None] | None:
...
-
+        # Streaming reports timing via speech_start/speech_end (the batch
+        # `timestamps` array is not sent over the socket). When absent, end_time
+        # is 0.0 and the pipeline falls back to wall-clock for EOU timing.
         speech_data = stt.SpeechData(
             language=language,
             text=transcript_text,
-            start_time=start_time,
-            end_time=end_time,
+            start_time=self._positive_time(transcript_data.get("speech_start")) or 0.0,
```
chenghao-mou, line 1022: we can add the start_time_offset in _positive_time so that all values through that function includes session start -> audio start delay:

```python
    def _positive_time(self, value: object) -> float | None:
        if isinstance(value, bool) or not isinstance(value, (int, float)):
            return None
        if value <= 0:
            return None
        return float(value) + self.start_time_offset
```

chenghao-mou (approved):
lgtm, one small nitpick.

PR #6058, opened 2026-06-11
fix(google): correct pitch parameter type from int to float in TTS

### LINKED ISSUE #6057: fix(google): TTS pitch parameter typed as int but should be float
## Bug Description

`TTS.__init__` in `livekit-plugins-google` declares `pitch` as `int`:

```python
pitch: int = 0,
```

But the docstring, the Google Cloud TTS API, and the `_TTSOptions` dataclass all treat it as `float` (range -20.0 to 20.0 semitones). Passing `pitch=1.5` works at runtime since Python doesn't enforce annotations, but mypy flags it and IDEs will reject it.

## Root Cause

The `_TTSOptions` dataclass already has `pitch: float` (correct), but the `__init__` signature has `pitch: int = 0` (wrong). One-character mismatch between the dataclass field and the constructor annotation.

## Fix

Change `pitch: int = 0` to `pitch: float = 0` in `TTS.__init__`.

adityajha2005:
The `pitch` parameter in `TTS.__init__` was typed as `int` while the `_TTSOptions` dataclass, the docstring, and the Google Cloud TTS API all define it as a float in the range -20.0 to 20.0. This caused mypy errors and IDE type rejections when passing a float like `pitch=1.5`.

Fixes #6057

## Test plan

```
uv run pytest tests/test_plugin_google_tts.py --plugin google -v
```
image


tests/test_plugin_google_tts.py
```diff
@@ -0,0 +1,24 @@
+import pytest
```
longcw: maybe remove this file, it doesn't look very helpful.

longcw (approved):

tests/test_plugin_google_tts.py
adityajha2005: > maybe remove this file, it doesn't look very helpful.

done!

tinalenguyen (approved):

PR #6064, opened 2026-06-11
fix(neuphonic): use configured encoding and lang_code in TTS requests

### LINKED ISSUE #6063: fix(neuphonic): TTS ignores configured encoding and hardcodes language in WebSocket path
## Bug Description

Two bugs in `livekit-plugins-neuphonic/tts.py`:

**1. Hardcoded encoding in SSE request (line 255)**

`ChunkedStream` always sends `"encoding": "pcm_linear"` in the request body, ignoring the `encoding` parameter passed to the constructor. Any non-default encoding is silently dropped.

```python
# Before - ignores self._opts.encoding
"encoding": "pcm_linear",

# After
"encoding": self._opts.encoding,
```

**2. Hardcoded language in WebSocket URL path (line 141)**

`_connect_ws` builds the URL as `/speak/en?...` regardless of the configured `lang_code`. The SSE path correctly uses `/sse/speak/{lang_code}`, but the WebSocket path always connects to the English endpoint.

```python
# Before - always connects to /speak/en
f"/speak/en?...&lang_code={self._opts.lang_code.language}&..."

# After
f"/speak/{self._opts.lang_code.language}?...&lang_code={self._opts.lang_code.language}&..."
```

adityajha2005:
Two bugs in the Neuphonic TTS plugin where configured options were silently ignored.

**1. SSE encoding hardcoded** - \`ChunkedStream\` always sent \`"encoding": "pcm_linear"\` regardless of the \`encoding\` constructor argument. Fixed to use \`self._opts.encoding\`.

**2. WebSocket lang_code hardcoded** - \`_connect_ws\` built the URL as \`/speak/en\` regardless of \`lang_code\`. The SSE path already does this correctly (\`/sse/speak/{lang_code}\`). Fixed to use \`self._opts.lang_code.language\` in the WebSocket path too.

Fixes #6063

### DIFF
--- a/livekit-plugins/livekit-plugins-neuphonic/livekit/plugins/neuphonic/tts.py
+++ b/livekit-plugins/livekit-plugins-neuphonic/livekit/plugins/neuphonic/tts.py
@@ -43,6 +43,14 @@ from .models import TTSLangCodes  # noqa: I001
 API_AUTH_HEADER = "x-api-key"
 
 
+def _encoding_to_mime_type(encoding: str) -> str:
+    if encoding.startswith("pcm"):
+        return "audio/pcm"
+    elif encoding.startswith("mp3"):
+        return "audio/mp3"
+    return "audio/pcm"
+
+
 @dataclass
 class _TTSOptions:
     lang_code: LanguageCode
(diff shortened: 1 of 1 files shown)


adityajha2005:
@devin-ai-integration - addressed both points in the latest push:

1. Added `_encoding_to_mime_type()` helper (matching the ElevenLabs pattern) and replaced both hardcoded `mime_type="audio/pcm"` calls in `ChunkedStream` and `SynthesizeStream`.

2. On the WebSocket encoding param - the WS URL already passes `sampling_rate`, `lang_code`, `voice_id`, and `speed` as query params but not `encoding`. Left as-is for now since it matches the existing WS path behaviour and would need Neuphonic API docs to confirm whether the WS endpoint accepts an encoding param.

theomonnom (approved):

PR #6071, opened 2026-06-12
fix(voice): deadlock when session closes while awaiting an AgentTask

longcw:
Fixes a deadlock where the session never closes when a participant disconnects while `on_enter` (or a function tool) is about to `await AgentTask()` — `AgentSession.aclose()` then times out after 60s. Reported in https://community.livekit.io/t/solved-session-never-closes-when-participant-disconnects-before-agenttask/1411.

**Root cause**

`aclose()` drains the current activity while holding the activity lock and waits for all speech tasks, including the `on_enter` task. Meanwhile the `AgentTask` handoff calls `pause()` on the same activity, which blocks on that lock — a circular wait: drain → scheduling task → on_enter → drain's lock. The existing `blocked_tasks` mechanism was designed to break exactly this edge, but it was only threaded as an argument through `pause()`, so a drain initiated by the session close never knew about the blocked tasks.

**Fix**

- `AgentTask.__await_impl` now registers the handoff-blocked tasks on the old activity (`_add_drain_blocked_tasks`, before any await), so any drain — including one initiated by the session close — excludes them from its wait. `_drain_blocked_tasks` is now a set; `_pause_scheduling_task` merges into it instead of overwriting, and it's cleared on resume.
- `_update_activity` bails out early when the session is closing instead of pausing the previous activity for a handoff that will be skipped anyway.
- `AgentTask.__await_impl` skips the parent handoff in its `finally` when the activity never started due to the close, instead of resuming the parent activity mid-close.
- `pause()` returns early on an already-closed activity.

The `AgentTask` now completes with the existing `ToolError("activity doesn't start ...")` and the session closes promptly. Added a regression test that deterministically reproduces the race (it deadlocks without the fix).

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent.py
+++ b/livekit-agents/livekit/agents/voice/agent.py
@@ -869,6 +869,10 @@ class AgentTask(Agent, Generic[TaskResult_T]):
         ):
             blocked_tasks.append(old_activity._on_enter_task)
 
+        # register before any await so a concurrent drain (e.g. session close)
+        # won't wait for tasks blocked on this handoff
+        old_activity._add_drain_blocked_tasks(blocked_tasks)
+
         if (
             task_info.function_call
             and isinstance(old_activity.llm, RealtimeModel)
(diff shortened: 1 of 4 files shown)


theomonnom (approved):
lgtm, do we make sure the Activity is close tho?

longcw:
the Activity should be closed in agent session aclose

PR #6081, opened 2026-06-12
Refactor LiveKit API references to use lkapi

sbriceland:
Am I the only one with VS Code intellisense issues because of the property `api` being confused with the `api` module?

### DIFF
--- a/livekit-agents/livekit/agents/job.py
+++ b/livekit-agents/livekit/agents/job.py
@@ -402,7 +402,7 @@ class JobContext:
         LIVEKIT_URL, LIVEKIT_API_KEY, and LIVEKIT_API_SECRET so this API is always
         usable inside job entrypoints.
         """
-        return api.LiveKitAPI(session=http_context.http_session())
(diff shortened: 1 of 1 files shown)


chenghao-mou:
I don't think there is an `api` module in this repo. Can you provide more details?

sbriceland:
> I don't think there is an `api` module in this repo. Can you provide more details?

Please see screenshots. In the `JobContext` class, the `api` property on the class reports as `Unkown` when you hover over the type. I looked into the `job.JobContext` and it appears the issue could be VS Code is confused when trying to infer `api.LiveKitAPI` because the class has a property named `api`.

Screenshot 2026-06-15 at 10 02 24 AM
Screenshot 2026-06-15 at 10 02 39 AM

chenghao-mou:
With a fresh install of vs code and the recommended python plugins, I don't see any alerts/errors: 
CleanShot 2026-06-19 at 11 27 26@2x

PR #6084, opened 2026-06-12
fix(voice): include conversation items added during tool execution in reply context

ShayneP:
When a function tool awaits an inline `AgentTask`, the sub-conversation it runs is merged into the agent's `chat_ctx` at handoff-return. But the reply generated after tool execution uses a `chat_ctx` snapshot taken before the tool ran, so it's blind to everything said inside the task — in practice the agent re-asks for fields the task already collected.

This refreshes the snapshot with any chat items added during tool execution before generating the tool reply, mirroring the existing `update_instructions()` refresh just below it.

Found while benchmarking the hotel receptionist example, where the card-collection `AgentTask` would complete and the agent would immediately ask for the card number again.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -3045,7 +3045,15 @@ class AgentActivity(RecognitionHooks):
 
             tool_messages = new_calls + new_fnc_outputs
             if fnc_executed_ev._reply_required:
+                # refresh conversation items added during tool execution: a tool that
+                # awaits an inline AgentTask runs a whole sub-conversation, merged into
+                # the agent's chat_ctx at handoff-return - this turn's snapshot predates
+                # it. Without the refresh, the tool response is generated blind to what
+                # was actually said inside the tool call (and re-asks captured fields).
+                # Conversational analog of the update_instructions() refresh below.
+                # tool_messages must be added first so merge() dedups them by id.
                 chat_ctx.items.extend(tool_messages)
+                chat_ctx.merge(self._agent._chat_ctx, exclude_instructions=True)
 
                 # refresh instructions in chat_ctx so that any update_instructions()
                 # calls made inside tool functions are reflected in the tool response


theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -3045,6 +3045,17 @@ async def _next_segment() -> _SpeechSegment | None:
...
             if fnc_executed_ev._reply_required:
+                # refresh conversation items added during tool execution: a tool that
+                # awaits an inline AgentTask runs a whole sub-conversation, merged into
+                # the agent's chat_ctx at handoff-return - this turn's snapshot predates
+                # it. Without the refresh, the tool response is generated blind to what
+                # was actually said inside the tool call (and re-asks captured fields).
+                # Conversational analog of the update_instructions() refresh below.
+                known_ids = {item.id for item in chat_ctx.items}
+                known_ids.update(item.id for item in tool_messages)
+                chat_ctx.items.extend(
```
theomonnom: Maybe https://github.com/livekit/agents/blob/e87d3eec4cd6fa46870027464dafc8bc339043ed/livekit-agents/livekit/agents/voice/agent.py#L966-L972 isn't working properly?

theomonnom (changes requested):

theomonnom (approved):
lgtm! discussed in a meeting

livekit-agents/livekit/agents/voice/agent_activity.py
ShayneP: (Just adding what we talked about in meeting for posterity)

That part works well, I think it's that the tool response is generated from the turn's local snapshot, not from the agent's chat_ctx when [_pipeline_reply_task_impl copies the ctx at the start of the turn](https://github.com/livekit/agents/blob/04ac3bed5/livekit-agents/livekit/agents/voice/agent_activity.py#L2614) and the same local copy is [passed back in for the tool reply](https://github.com/livekit/agents/blob/04ac3bed5/livekit-agents/livekit/agents/voice/agent_activity.py#L3068) 

The copy is from before the tool execution, so the items the merge added to agent._chat_ctx  aren't in there so the **next** turn sees them, but the **immediate tool reply** doesn't, so sometimes the tool reply asks for things that have actually already been given _between the context that it copied and the current state_.

theomonnom (approved):

PR #6087, opened 2026-06-12
Improve hotel receptionist example: policies, card collection, richer DB

ShayneP:
Builds on the hotel receptionist example (#5983) based on what we learned running it against a larger benchmark suite:

- **Policies knowledge base**: a `policies/` directory of markdown files (rooms & amenities, dining, payments & currency, guest privacy, group bookings, transport, tours, accessibility, …) loaded via `policies/__init__.py`, so the agent answers from grounded hotel policy instead of improvising.
- **Card collection**: replaces the `get_consent` flow with a `GetCardTask`-based payment card collection flow (`get_card.py`).
- **Richer hotel DB**: walk handling for sold-out nights, tour bookings, wake-up calls, guest messages, and no-show guarantee data, plus expanded seed data.
- **Tool tightening**: `book_room`, `modify_booking`, `verify_booking`, and `book_restaurant` get better validation and clearer outputs; persona instructions updated to match.
- **Benchmark support**: dumps final DB state per room when `LIVEKIT_SESSION_REPORT_DIR` is set, so runs can be graded post-hoc by diffing expected vs actual state.

The accompanying benchmark scenarios are in a follow-up PR stacked on this one.


examples/hotel_receptionist/get_card.py
theomonnom, line 1: We should reuse or edit our built-in GetCreditCard

tinalenguyen (approved):

PR #6095, opened 2026-06-14
fix(xai): expose TTS websocket options

### LINKED ISSUE #6077: xai TTS: documented API parameters (optimize_streaming_latency, speed, text_normalization) cannot be passed
### Plugin / version

`livekit-plugins-xai` 1.6.0, `livekit-agents` 1.6.0, Python 3.13

**Verified on the latest release:** inspected the `livekit-plugins-xai` 1.6.0 wheel from PyPI (uploaded 2026-06-11) and the current `main` branch ([`livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/tts.py`](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/tts.py)) — `TTS.__init__` and `_connect_ws` are unchanged; none of the parameters below are exposed. Behavior was originally observed in production on 1.5.17, which is code-identical.

### Problem

xAI's TTS WebSocket endpoint (`wss://api.x.ai/v1/tts`) accepts several connection query parameters documented in the [xAI TTS docs](https://docs.x.ai/developers/model-capabilities/audio/text-to-speech):

- `optimize_streaming_latency` (0/1/2) — first-chunk latency mode
- `speed` (0.7–1.5) — speaking-rate multiplier
- `text_normalization` (bool) — spoken-form normalization of numbers/abbreviations

The plugin's `TTS.__init__` only accepts `api_key`, `voice`, `language`, `tokenizer`, `http_session` (keyword-only, no `**kwargs`), and `_connect_ws` sends only `voice`, `language`, `codec`, `sample_rate`. There is no way to set any of the parameters above — passing them raises `TypeError: __init__() got an unexpected keyword argument`.

### Why it matters

`optimize_streaming_latency` in particular has a large effect on real-time voice agents. Measured against the live endpoint (same text, voice `ara`, 4 runs each):

| level | generation TTFB |
|---|---|
| 0 (plugin's effective behavior) | mean 741 ms |
| 2 | **mean 379 ms** |

For a voice-agent use case this nearly halves TTS time-to-first-byte, but plugin users are stuck on the slowest setting.

### Expected behavior

`TTS.__init__` exposes these documented parameters (and forwards them as WS query params), similar to how other TTS plugins expose their providers' generation options.

nightcityblade:
Fixes #6077

## Summary
- expose xAI TTS websocket options for optimize_streaming_latency, speed, and text_normalization
- forward provided values as websocket query parameters
- allow updating these options via update_options

## Testing
- uv run ruff format livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/tts.py
- uv run ruff check livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/tts.py
- uv run python - <<'PY' ... xAI TTS option smoke test

### DIFF
--- a/livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/tts.py
+++ b/livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/tts.py
@@ -20,6 +20,7 @@ import json
 import os
 import weakref
 from dataclasses import dataclass, replace
+from typing import Literal
 from urllib.parse import urlencode
 
 import aiohttp
@@ -51,6 +52,9 @@ class _TTSOptions:
     voice: GrokVoices | str
     language: TTSLanguages | str
     tokenizer: tokenize.WordTokenizer
+    optimize_streaming_latency: NotGivenOr[Literal[0, 1, 2]]
+    speed: NotGivenOr[float]
+    text_normalization: NotGivenOr[bool]
 
 
 class TTS(tts.TTS):
@@ -60,6 +64,9 @@ class TTS(tts.TTS):
         api_key: NotGivenOr[str] = NOT_GIVEN,
         voice: GrokVoices | str = DEFAULT_VOICE,
         language: TTSLanguages | str = "auto",
+        optimize_streaming_latency: NotGivenOr[Literal[0, 1, 2]] = NOT_GIVEN,
+        speed: NotGivenOr[float] = NOT_GIVEN,
+        text_normalization: NotGivenOr[bool] = NOT_GIVEN,
         tokenizer: tokenize.WordTokenizer | None = None,
         http_session: aiohttp.ClientSession | None = None,
     ) -> None:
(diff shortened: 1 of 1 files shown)


nightcityblade:
Thanks for the review. I pushed f38e7d5 to forward the new xAI TTS WebSocket options (optimize_streaming_latency, speed, text_normalization) as query params, and changed the WebSocket connection helper to use the stream's captured options snapshot so changes made after stream creation don't leak into that stream's connection.

livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/tts.py
```diff
@@ -51,6 +52,9 @@ class _TTSOptions:
     voice: GrokVoices | str
     language: TTSLanguages | str
     tokenizer: tokenize.WordTokenizer
+    optimize_streaming_latency: NotGivenOr[Literal[0, 1, 2]]
```
tinalenguyen: i think we should accept `int` for this, in case more levels are added in the future so this won't break

tinalenguyen (commented):
thanks for the PR! i left a comment

nightcityblade:
Thanks for the review — addressed in `7f605e3`.

I changed `optimize_streaming_latency` to accept `int` instead of `Literal[0, 1, 2]`, since the xAI API allows a wider numeric range and validates it server-side.

Validation run:
- `uv run ruff check livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/tts.py`
- `uv run ruff format --check livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/tts.py`

tinalenguyen (approved):
thank you for the PR!

PR #6097, opened 2026-06-14
fix(clova): stop _recognize_impl from mutating instance language as a side effect

### LINKED ISSUE #6096: fix(clova): _recognize_impl permanently mutates instance language as a side effect
## Bug Description

`STT._recognize_impl` accepts a per-call `language` parameter but mutates `self._language` when it is provided, permanently changing the instance's language for all future calls:

```python
async def _recognize_impl(self, buffer, *, language=NOT_GIVEN, ...):
    if is_given(language):
        normalized = LanguageCode(language).iso
        self._language = clova_languages_mapping.get(normalized, normalized)  # ← persists!
    payload = json.dumps({"language": self._language, ...})
```

## Reproduction

```python
stt = STT(language="en-US", ...)
await stt.recognize(buffer1)              # uses "en-US" ✓
await stt.recognize(buffer2, language="fr")  # uses "fr" ✓ but also silently sets self._language = "fr"
await stt.recognize(buffer3)              # now uses "fr" ✗ — expected "en-US"
```

## Fix

Use a local `lang` variable scoped to the request instead of mutating `self._language`. Thread it through to `_transcription_to_speech_event` so the response also reflects the correct per-call language.

adityajha2005:
`_recognize_impl` accepted a per-call `language` override but wrote it back to `self._language`, permanently changing the instance's language for all subsequent calls that don't specify one.

Fix: compute a local `lang` variable from the per-call override (falling back to `self._language`), use it in the request payload, and pass it through to `_transcription_to_speech_event` so the response language also reflects the correct per-call value. `self._language` is now only ever mutated by `__init__` and `update_options`.

Fixes #6096

### DIFF
--- a/livekit-plugins/livekit-plugins-clova/livekit/plugins/clova/stt.py
+++ b/livekit-plugins/livekit-plugins-clova/livekit/plugins/clova/stt.py
@@ -111,10 +111,11 @@ class STT(stt.STT):
     ) -> stt.SpeechEvent:
         try:
             url = self.url_builder()
+            lang = self._language
             if is_given(language):
                 normalized = LanguageCode(language).iso
-                self._language = clova_languages_mapping.get(normalized, normalized)
-            payload = json.dumps({"language": self._language, "completion": "sync"})
+                lang = clova_languages_mapping.get(normalized, normalized)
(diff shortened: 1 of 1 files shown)


chenghao-mou (approved):
Thanks for the PR!

Issue #5669, opened 2026-05-07
Bug: end_of_turn_delay collapses to min_endpointing_delay when running without VAD in turn_detection="stt" mode


miguelmoralai:
## Summary

When an `AgentSession` is configured with `vad=NOT_GIVEN` and `turn_detection="stt"`, the `end_of_turn_delay` (and `transcription_delay`) values reported on user `ChatMessage` metrics no longer measure what their names suggest. They collapse to roughly `endpointing.min_delay` regardless of how long the STT model actually took to detect end of turn.

This is a problem for users who rely on these metrics to monitor real end of utterance latency or to compare STT providers.

## Repro

1. Build an `AgentSession` with a streaming STT that emits `END_OF_SPEECH` (Soniox or Deepgram Flux for example).
2. Pass `turn_detection="stt"` and `vad=NOT_GIVEN`.
3. Run a normal turn and inspect the `metrics` dict on the resulting user `ChatMessage`.

Observed: `end_of_turn_delay` is approximately equal to `endpointing.min_delay` on every turn, even on ambiguous turns where the STT model spends close to its `MAX_ENDPOINT_DELAY` ceiling before firing.

Expected: `end_of_turn_delay` should reflect the time from the user actually stopping to the EOU pipeline committing, including the STT model decision time.

## Root cause

In `livekit/agents/voice/audio_recognition.py` the `_last_speaking_time` anchor is set in three places when there is no VAD attached:

```python
# line 846 (FINAL_TRANSCRIPT)
if not self._vad or self._last_speaking_time is None:
    self._last_speaking_time = time.time()

# line 903 (PREFLIGHT_TRANSCRIPT)
if not self._vad or self._last_speaking_time is None:
    self._last_speaking_time = time.time()

# line 944 (END_OF_SPEECH, stt mode)
if not self._vad or self._last_speaking_time is None:
    self._last_speaking_time = time.time()
```

Without VAD, the first event handler that runs stamps `_last_speaking_time = time.time()`. For Soniox and Flux this is essentially the moment `END_OF_SPEECH` arrives, that is, **after** the model has already decided end of turn. The bounce task then runs and sleeps for `endpointing_delay`, then computes:

```python
end_of_turn_delay = time.time() - last_speaking_time  # line 1121
```

Since `last_speaking_time` was just set to "now" before the sleep, the resulting value is approximately `endpointing_delay` and hides the real STT detection time.

The TODO already in the file at line 848 acknowledges this:

```python
[...]

PR #6098, opened 2026-06-14
fix(voice): skip end-of-turn metrics on stale/out-of-order speaking anchor

### LINKED ISSUE #6093: transcription_delay / end_of_turn_delay incorrectly large (>200s) or missing — stopped_speaking_at predates started_speaking_at
## Summary
For some user turns, the reported `transcription_delay` and `end_of_turn_delay` metrics on `ChatMessage` are extremely large (often >200s) despite no actual user-perceived delay in the corresponding session recordings. In other turns within the same session, these fields are missing entirely.

## Environment
- python 3.14
- livekit-agents==1.6.0
- livekit-plugins-deepgram==1.6.0
- livekit-plugins-turn-detector==1.6.0
- STT: custom Deepgram provider (own API key), model `nova-3-medical`
- Endpoint config:
  `deepgram/v1/listen?model=nova-3-medical&punctuate=true&smart_format=true&no_delay=true&interim_results=true&encoding=linear16&vad_events=true&sample_rate=16000&channels=1&endpointing=100&filler_words=true&profanity_filter=false&numerals=false&mip_opt_out=false&language=en-GB`
- WE use multilangual turn detector with `endpointing={"min_delay": 0.1, "max_delay": 1.5}` and min_silence_duration in vad is 0.3

## Observed behavior

Example metric payload from an affected turn:

```json
{
  "started_speaking_at": "2026-06-12T18:52:03.881Z",
  "stopped_speaking_at": "2026-06-12T18:48:40.647Z",
  "transcription_delay": 203.2745988368988,
  "end_of_turn_delay": 203.40400671958923,
  "on_user_turn_completed_delay": 0.00006804399890825152
}
```

Note that `stopped_speaking_at` is ~203.2s *before* `started_speaking_at` — i.e. it precedes the start of the turn it's supposedly the end of. The reported `transcription_delay` (203.27s) and `end_of_turn_delay` (203.40s) are almost exactly equal to this gap, which suggests both are being computed against a stale `_last_speaking_time` anchor left over from an earlier point (e.g. previous turn or session start) rather than the actual end of the current utterance.

Session recordings for these turns show no perceptible delay between the user finishing speaking and the agent responding.

## Affected sessions (high/incorrect E2E latency)
- `RM_8QjN2S9pRA4m` — timestamp 8:10.88
- `RM_5CcWSoM77Npx` — timestamp 9:13.26
- `RM_vrwpheZtBoDx` — timestamp 8:07.65
- `RM_8GAcHsSvxCkq` — timestamp 8:53.57

## Affected sessions (E2E latency data missing entirely)
- `RM_8GAcHsSvxCkq`
  - 7:25.12
  - 6:24.22
  - 5:04.62
  - 4:03.84
  - and several other timestamps within the same session

## Expected behavior
`transcription_delay` and `end_of_turn_delay` should reflect real sub-second-to-low-single-digit-second delays consistent with the configured `endpointing`/`max_endpointing_delay`, and should be present for every committed user turn.

## Possibly related issues
- #2361 — `transcription_delay` occasionally receives extremely large values that look like raw timestamps; suspected cause is `_last_speaking_time` being 0/stale.
- #5669 — `end_of_turn_delay` misbehaves when the `_last_speaking_time` anchor is set via STT events rather than VAD.

## Additional context
For analytics we currently rely primarily on TTSFB and LLM TTFT rather than E2E latency, given the reliability concerns above. It would help to know whether this is a known/tracked issue and whether there's a recommended workaround or upcoming fix.

## Additional findings — concrete chat-history evidence

Pulling raw `ChatMessage` metrics from one of the affected sessions (items 4–5 in the conversation, `role: "user"` for both, no assistant turn in between — i.e. one continuous user utterance split into two committed turns by the turn detector):

```json
// Turn N (first split of the utterance) — looks correct
{
  "started_speaking_at": 1781342795.3221886,
  "stopped_speaking_at": 1781342804.3863783,
  "created_at": 1781342803.7775867,
  "transcription_delay": 0.0,
  "end_of_turn_delay": 0.3565683364868164
}

// Turn N+1 (second split, same utterance, no intervening assistant turn)
{
  "started_speaking_at": 1781342804.815377,
  "stopped_speaking_at": 1781342584.6181495,
  "created_at": 1781342804.9027314,
  "transcription_delay": 220.2844967842102,
  "end_of_turn_delay": 220.3811547756195
}
```

Empirically, across multiple turns in this session:

- `transcription_delay ≈ max(created_at − stopped_speaking_at, 0)`
- `end_of_turn_delay ≈ time_of_commit − stopped_speaking_at` (time_of_commit ≈ created_at / "now")

This confirms `stopped_speaking_at` in the reported metrics corresponds to the internal `_last_speaking_time` anchor.

For turn N, `stopped_speaking_at` (804.386) is current relative to `created_at` (803.778) — it's just *slightly ahead* of `created_at`, so the `max(x, 0)` clamp produces `transcription_delay = 0.0` (the "always-zero" symptom from #4388).

For turn N+1, `stopped_speaking_at` (584.618) is ~220s in the past relative to `created_at`/`started_speaking_at` (~804.8–804.9, i.e. "now") — even though turn N+1 is the *very next* committed message, with the user speaking continuously and no assistant turn between N and N+1. This produces the huge bogus delay (#2361 symptom).

**Hypothesis:** when the turn detector commits multiple consecutive same-role turns from a single continuous user utterance (no VAD speech-stop/start cycle between them), `_last_speaking_time` is not refreshed for the second+ split turn and falls back to a stale value from an earlier point in the session (here, ~220s prior — possibly the last genuine VAD speech-start/stop event before this monologue began). This would explain both the "delay = 0" and "delay = huge" symptoms as two manifestations of the same anchor-refresh bug, occurring on adjacent turns within the same long user utterance.

One more detail: in this session, `stt_metadata`/`llm_metadata`/`tts_metadata` all report `model_name: "FallbackAdapter"`, `model_provider: "livekit"` for these turns — i.e. the configured Deepgram STT (and primary LLM/TTS) appear to have fallen back to LiveKit's FallbackAdapter for this segment. Not sure if relevant, but flagging in case the fallback path has separate/duplicated `_last_speaking_time` tracking that isn't kept in sync with the primary path.

Happy to share the full session JSON if useful for repro.

### Expected Behavior

`stopped_speaking_at` (the internal `_last_speaking_time` anchor) should always reflect a timestamp close to the actual end of the audio relevant to the committed turn — never a value from many seconds/turns earlier in the session.

Concretely:

- `transcription_delay` and `end_of_turn_delay` should be small, bounded values (roughly 0 to `max_endpointing_delay` + STT processing latency), never on the order of hundreds of seconds when there's no real silence gap.
- `transcription_delay` should reflect the true (small, positive) STT finalization delay, not be clamped to exactly `0.0` just because `stopped_speaking_at` happens to be slightly ahead of `created_at`/`last_final_transcript_time`.
- When the turn detector commits multiple consecutive same-role turns from one continuous utterance (no VAD stop/start cycle between them), each committed turn's `stopped_speaking_at` should be derived from that segment's own audio timing, not inherited stale from an earlier point in the session.
- The corresponding assistant turn's `e2e_latency` should reflect the actual perceived response latency (consistent with `tts_node_ttfb`/`llm_node_ttft` + small overhead), not be inflated by a stale anchor on the preceding user turn.

### Reproduction Steps

```bash
This issue does not appear to be reliably reproducible on demand (similar to what was noted in #2361 — "cannot reproduce intentionally, but happens quite often during normal usage"), but it occurs consistently in many sessions for some of the turns mostly max 1-2 time in any session.
```

### Operating System

Livekit cloud

### Models Used

Deepgram nova-3, Openai 5.4 mini, Cartesia sonic 3.5

### Package Versions

```bash
- python 3.14
- livekit-agents==1.6.0
- livekit-plugins-deepgram==1.6.0
- livekit-plugins-turn-detector==1.6.0
- livekit-api>=0.8
```

### Affected session (E2E latency data mismatch)
- `RM_8QjN2S9pRA4m` — timestamp 8:10.88
- `RM_5CcWSoM77Npx` — timestamp 9:13.26
- `RM_vrwpheZtBoDx` — timestamp 8:07.65
- `RM_8GAcHsSvxCkq` — timestamp 8:53.57

### Affected sessions (E2E latency data missing entirely)
- `RM_8GAcHsSvxCkq`
  - 7:25.12
  - 6:24.22
  - 5:04.62
  - 4:03.84
  - and several other timestamps within the same session
### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

anshulkulhari7:
## Summary

Fixes #6093.

For some user turns, the reported `transcription_delay` / `end_of_turn_delay` metrics on `ChatMessage` are extremely large (often >200s) even though the session recordings show no real delay, and `stopped_speaking_at` can precede `started_speaking_at`. In other turns the fields are missing entirely.

### Root cause

The metrics are computed in `_bounce_eou_task` (`audio_recognition.py`) from three captured anchors:

```python
started_speaking_at  = speech_start_time
stopped_speaking_at  = last_speaking_time           # the internal _last_speaking_time anchor
transcription_delay  = max(last_final_transcript_time - last_speaking_time, 0)
end_of_turn_delay    = time.time() - last_speaking_time
```

The guard around this block only checked that the three values are **not `None`**. When the turn detector commits a user turn whose `_last_speaking_time` was never refreshed for that segment — e.g. consecutive same-role turns split from one continuous utterance, with no VAD speech-stop/start cycle between them — the anchor is left over from an earlier point in the session and can predate the start of the current turn.

In that case the not-`None` guard still passes, so `end_of_turn_delay = now - last_speaking_time` becomes ~200s and `stopped_speaking_at` ends up *before* `started_speaking_at`, exactly the payload reported in the issue.

This is the same class of bug noted in #2361 / #5669 / #4388 (stale/`0` anchor), now manifesting as an *out-of-order* anchor on adjacent turns within one long utterance.

### Fix

An anchor that predates the start of the turn (`last_speaking_time < speech_start_time`) is logically impossible — you cannot stop speaking before the turn started. The existing code already has a policy for unreliable timing (see the in-code comment): *skip the calculation and report the metrics as `None`, because that is better than emitting a likely-wrong value.* This change extends that same policy to the out-of-order case.

The computation is extracted into a small pure helper, `_compute_end_of_turn_metrics`, which:

- returns `None` for all four metrics when any anchor is missing **or** when `last_speaking_time < speech_start_time` (stale/out-of-order), and
- otherwise returns the same values as before (with `end_of_turn_delay` now clamped to `>= 0`, consistent with the existing `transcription_delay` clamp).

This makes the behaviour directly unit-testable without audio/STT/VAD.

## Testing

New unit test module `tests/test_end_of_turn_metrics.py` exercises the pure helper with crafted timestamps (no audio):

- `test_normal_turn_produces_small_bounded_delays` — well-ordered turn yields the expected sub-second delays.
- `test_stale_anchor_predating_turn_start_is_skipped` — regression for this issue, using the exact `~220s` numbers from the reported payload; all four metrics must be `None`.
- `test_anchor_equal_to_start_is_accepted` — boundary (`last_speaking_time == speech_start_time`) stays valid.
- `test_missing_anchor_is_skipped` — any missing anchor skips the calculation.

```
$ uv run pytest tests/test_end_of_turn_metrics.py --unit -q
......                                                                   [100%]
6 passed in 0.02s
```

Confirmed RED before the fix (reverting the ordering guard): `test_stale_anchor_predating_turn_start_is_skipped` failed with `started_speaking_at=1781342804.815377`, `end_of_turn_delay=220.28458189964294` — i.e. the bogus >200s value. The existing `tests/test_speech_start_time_persistence.py` still passes.

`ruff check`, `ruff format --check`, and `mypy` are clean on the changed files.

## AI disclosure

This change was AI-assisted; all logic, tests, and verification were reviewed by the author.

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -49,12 +49,7 @@ class _EndOfTurnInfo:
     """If True, a reply was already triggered and should be skipped after end of turn detection."""
     new_transcript: str
     transcript_confidence: float
-
-    # metrics report
-    started_speaking_at: float | None
-    stopped_speaking_at: float | None
-    transcription_delay: float | None
-    end_of_turn_delay: float | None
+    metrics: _EndOfTurnMetrics
 
 
 @dataclass
@@ -1255,20 +1250,12 @@ class AudioRecognition:
                 last_final_transcript_time=last_final_transcript_time,
                 now=time.time(),
             )
-            started_speaking_at = metrics.started_speaking_at
-            stopped_speaking_at = metrics.stopped_speaking_at
-            transcription_delay = metrics.transcription_delay
-            end_of_turn_delay = metrics.end_of_turn_delay
-
             committed = self._hooks.on_end_of_turn(
                 _EndOfTurnInfo(
                     skip_reply=skip_reply,
                     new_transcript=self._audio_transcript,
                     transcript_confidence=confidence_avg,
-                    transcription_delay=transcription_delay or 0,
-                    end_of_turn_delay=end_of_turn_delay,
-                    started_speaking_at=started_speaking_at,
-                    stopped_speaking_at=stopped_speaking_at,
+                    metrics=metrics,
                 )
             )
             if committed:
@@ -1276,8 +1263,8 @@ class AudioRecognition:
                     {
                         trace_types.ATTR_USER_TRANSCRIPT: self._audio_transcript,
                         trace_types.ATTR_TRANSCRIPT_CONFIDENCE: confidence_avg,
-                        trace_types.ATTR_TRANSCRIPTION_DELAY: transcription_delay or 0,
-                        trace_types.ATTR_END_OF_TURN_DELAY: end_of_turn_delay or 0,
+                        trace_types.ATTR_TRANSCRIPTION_DELAY: metrics.transcription_delay or 0,
+                        trace_types.ATTR_END_OF_TURN_DELAY: metrics.end_of_turn_delay or 0,
                     }
                 )
                 if self._stt_request_ids:
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -58,6 +58,55 @@ class _EndOfTurnInfo:
 
 
 @dataclass
+class _EndOfTurnMetrics:
+    started_speaking_at: float | None
+    stopped_speaking_at: float | None
+    transcription_delay: float | None
+    end_of_turn_delay: float | None
+
+
+def _compute_end_of_turn_metrics(
+    *,
+    speech_start_time: float | None,
+    last_speaking_time: float | None,
+    last_final_transcript_time: float | None,
+    now: float,
+) -> _EndOfTurnMetrics:
+    """Compute the end-of-turn timing metrics from the captured turn anchors.
+
+    ``last_speaking_time`` is the internal ``_last_speaking_time`` anchor (reported
+    as ``stopped_speaking_at``). When the turn detector commits a turn whose anchor
+    was never refreshed for this segment, that value can be stale and predate the
+    start of the current turn, producing wildly inflated delays (see issue #6093).
+
+    We treat such an inconsistent anchor the same way we treat unreliable VAD: skip
+    the calculation and return ``None`` rather than emit a likely wrong value. A
+    valid anchor must satisfy ``last_speaking_time >= speech_start_time`` (you cannot
+    stop speaking before the turn started).
+    """
+    if (
+        speech_start_time is None
+        or last_speaking_time is None
+        or last_final_transcript_time is None
+        # stale/out-of-order anchor: stopping to speak cannot predate the turn start
+        or last_speaking_time < speech_start_time
+    ):
+        return _EndOfTurnMetrics(
+            started_speaking_at=None,
+            stopped_speaking_at=None,
+            transcription_delay=None,
+            end_of_turn_delay=None,
+        )
+
+    return _EndOfTurnMetrics(
+        started_speaking_at=speech_start_time,
+        stopped_speaking_at=last_speaking_time,
+        transcription_delay=max(last_final_transcript_time - last_speaking_time, 0),
+        end_of_turn_delay=max(now - last_speaking_time, 0),
+    )
+
+
+@dataclass
 class _PreemptiveGenerationInfo:
     new_transcript: str
     transcript_confidence: float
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -57,6 +57,55 @@ class _EndOfTurnInfo:
     end_of_turn_delay: float | None
 
 
+@dataclass
+class _EndOfTurnMetrics:
```
chenghao-mou: since _EndOfTurnInfo is internal, I would go one step further to replace the four variables in _EndOfTurnInfo with this directly so we don't have to unpack or pass around those values individually.

chenghao-mou (commented):
Thanks for the PR! I have one small comment.

livekit-agents/livekit/agents/voice/audio_recognition.py
anshulkulhari7: Done — folded the four metric fields on `_EndOfTurnInfo` into a single `metrics: _EndOfTurnMetrics`. The computed object is now passed straight through; `_user_turn_completed_task`, `_init_metrics_from_end_of_turn`, and the turn span read `info.metrics.*`. mypy strict (593 files) and the unit tests pass.

chenghao-mou: Thanks! nitpicking here: can you move this above _EndOfTurnInfo so we have a correct reading order?

chenghao-mou (approved):

livekit-agents/livekit/agents/voice/audio_recognition.py
anshulkulhari7: Good call — since _EndOfTurnInfo now holds a _EndOfTurnMetrics, defining it first reads better. Moved it up in 002b0cb.

PR #6099, opened 2026-06-14
feat(eot): emit agent backchannel opportunity events

chenghao-mou:
## What

The multimodal EOT model now returns a **backchannel probability** alongside the end-of-turn probability. The turn detector computes a backchannel *opportunity* (a window where the agent could say a short "mm-hmm" while the user still holds the floor) and passes it to `AgentActivity`.

**Internal only:** this is not a public `AgentSession` event. `AgentActivity.on_agent_backchannel_opportunity` is currently a no-op with a TODO — the detection plumbing is in place so an internal consumer can be wired up later. Treated the same way as the internal `EotPredictionEvent`.

Stacked on top of `feat/AGT-2520-multimodal-EOU`.

## Protocol

Relies on new fields (already landing in `livekit/protocol`):
- `EotPrediction.backchannel_probability`
- `SessionCreated.default_backchannel_thresholds` / `default_backchannel_threshold`

## How it works

- `backchannel_probability` is threaded through `TurnDetectionEvent` (transport → `_resolve_prediction`).
- `ThresholdOptions` gains server-provided backchannel thresholds + `lookup_backchannel()`; exposed via `detector.backchannel_threshold()`.
- In `audio_recognition`, at the same point the EOT prediction is consumed, `on_agent_backchannel_opportunity` is invoked **only when** `backchannel_probability >= threshold` **and** the turn is still continuing (`end_of_turn_probability < unlikely_threshold`) — so it never collides with a committed end-of-turn reply.
- `AgentBackchannelOpportunityEvent` is kept as the internal hook payload; it is **not** in `EventTypes`, the `AgentEvent` union, or the package exports.

## Disabled cases (hook never fires)

- Gateway sends no backchannel thresholds (older server) → `lookup_backchannel` returns `None`.
- After cloud→local fallback — the mini model produces no backchannel probability.
- Backchannel has **no** user overrides (server-driven only).

## Tests

- `tests/test_audio_turn_detector_fallback.py::TestBackchannelThresholds` — lookup per-language/default, disabled when absent / non-positive / mini model / after fallback.
- `tests/test_audio_recognition_turn_detection.py::TestBackchannelOpportunityEmit` — gating (above threshold + turn continuing), no call on turn-end / below-threshold / disabled / text detector.
- `make check` green; `--audio_eot` category green.


examples/voice_agents/basic_agent.py
```diff
@@ -104,6 +110,10 @@ async def entrypoint(ctx: JobContext) -> None:
         ],
     )
 
+    @session.on("agent_backchannel_opportunity")
+    def _on_agent_backchannel_opportunity(ev: AgentBackchannelOpportunityEvent) -> None:
+        session.say("""<emotion value="curious"/>yeah...""")
```
theomonnom: let's make sure we use prerecorded audio. And let's still use session.say (with the AsyncIterable of frames)

If you check our utils, we already have an utility for file->audio stream


session.say is nice because we can still reuse all our SpeechHandle scheduling code

theomonnom (changes requested):

theomonnom:
Let's make sure nothing is added to the ChatContext when using session.say (should already be the case if text is empty)

examples/voice_agents/basic_agent.py
chenghao-mou: I am going to drop the event and the example change. 

I think we should merge this and eot changes into main first and create a new PR from the synced expressive mode branch so options can live there?

theomonnom: Yeah lgtm to me

theomonnom: Let's expose everything inside the expressiveness mode PR

theomonnom (approved):

PR #6106, opened 2026-06-15
feat(cerebras): expose reasoning_format for parity with reasoning_effort

### LINKED ISSUE #5989: Expose `reasoning_format` for Cerebras and xAI (Grok) — parity with `reasoning_effort`
### Feature Type

### Feature Description

When using Cerebras or Grok models that support extended thinking (`gpt-oss-120`) with `reasoning_effort` set, the model returns thinking tokens as part of the streamed response. **the TTS pipeline picks these up and reads them aloud** — the user hears the model's raw internal monologue instead of just the final answer.

_No response_

### Additional Context

_No response_

anshulkulhari7:
## Summary

Closes #5989.

Reasoning models such as `gpt-oss-120b` served by **Cerebras** stream their *thinking* tokens as part of the chat-completions response. Today there is no way to tell the provider to keep that internal monologue out of the message `content`, so the TTS pipeline ends up reading the model's raw reasoning aloud.

Cerebras supports a `reasoning_format` request field (`"parsed"`, `"raw"`, `"hidden"`, `"none"`) that controls where reasoning ends up. This PR exposes it on the relevant LLM constructors, mirroring exactly how `reasoning_effort` is already plumbed.

> **Scope note:** the first revision also exposed `reasoning_format` on `LLM.with_x_ai(...)`. That was removed after review feedback ([comment](https://github.com/livekit/agents/pull/6106#issuecomment-4867523296)) — xAI's API does not document a `reasoning_format` field (its reasoning control is `reasoning_effort`), so shipping it there would have been a silent no-op. The feature is now scoped to Cerebras only.

## Changes

- **`livekit-plugins-openai`** (`llm.py`)
  - Add a `ReasoningFormat = Literal["parsed", "raw", "hidden", "none"]` type (values per the Cerebras reasoning docs).
  - Add `reasoning_format` to `_LLMOptions`, the `LLM.__init__` signature, and the `LLM.with_cerebras(...)` factory.
  - In `chat()`, forward it to the request. Because `reasoning_format` is a provider-specific body field and **not** an OpenAI SDK keyword argument (passing it top-level raises `TypeError: unexpected keyword argument`), it is merged into `extra_body` — the same mechanism already used for OpenRouter-specific body fields. The user-supplied `extra_body` dict is never mutated in place.
- **`livekit-plugins-cerebras`** (`llm.py`)
  - Add `reasoning_format` to the Cerebras `LLM.__init__` and forward it to the base class.

The change is additive and backward-compatible: when `reasoning_format` is not set, nothing is added to the request.

## How `reasoning_format` reaches the API

`reasoning_effort` is a named param in the OpenAI Python SDK (`AsyncCompletions.create`), so it is passed top-level. `reasoning_format` is not, so it must travel through `extra_body`. Verified empirically against `openai==2.40.0`:

```
top-level reasoning_format: TypeError -> AsyncCompletions.create() got an unexpected keyword argument 'reasoning_format'
extra_body call body has reasoning_format: True
```

## Testing

Added `tests/test_plugin_reasoning_format.py` (unit, no live keys — constructs the LLM and inspects the request kwargs the stream would send):

- `reasoning_format` set on Cerebras `LLM` lands in `extra_body` of the outgoing request.
- `reasoning_format` set via `LLM.with_cerebras(...)` lands in `extra_body`.
- `"none"` is accepted and forwarded verbatim.
- When unset, no `reasoning_format` is added.

```
$ uv run pytest tests/test_plugin_reasoning_format.py --unit
4 passed
```

Quality gates on the changed files:

- `uv run ruff check` — All checks passed
- `uv run ruff format --check` — already formatted
- `uv run mypy -p livekit.plugins.openai -p livekit.plugins.cerebras` (strict) — Success: no issues found

---

*AI-assisted: this change was prepared with AI assistance and reviewed by the author.*

### DIFF
--- a/livekit-plugins/livekit-plugins-cerebras/livekit/plugins/cerebras/llm.py
+++ b/livekit-plugins/livekit-plugins-cerebras/livekit/plugins/cerebras/llm.py
@@ -33,6 +33,7 @@ from livekit.agents.types import (
 )
 from livekit.agents.utils import is_given
 from livekit.plugins.openai import LLM as OpenAILLM
+from livekit.plugins.openai.llm import ReasoningFormat
 
 from .models import CerebrasChatModels
 
@@ -112,6 +113,7 @@ class LLM(OpenAILLM):
         parallel_tool_calls: NotGivenOr[bool] = NOT_GIVEN,
         tool_choice: NotGivenOr[ToolChoice] = NOT_GIVEN,
         reasoning_effort: NotGivenOr[ReasoningEffort] = NOT_GIVEN,
+        reasoning_format: NotGivenOr[ReasoningFormat] = NOT_GIVEN,
         safety_identifier: NotGivenOr[str] = NOT_GIVEN,
         prompt_cache_key: NotGivenOr[str] = NOT_GIVEN,
         top_p: NotGivenOr[float] = NOT_GIVEN,
@@ -125,6 +127,10 @@ class LLM(OpenAILLM):
 
         ``api_key`` must be set to your Cerebras API key, either using the argument or by setting
         the ``CEREBRAS_API_KEY`` environmental variable.
+
+        ``reasoning_format`` controls how reasoning models (e.g. ``gpt-oss-120b``) return their
+        thinking tokens. Set it to ``"hidden"`` or ``"parsed"`` to keep the model's internal
+        monologue out of the spoken message content.
 
         When ``gzip_compression`` is True (default), request payloads are gzip-compressed,
         which can reduce TTFT for requests with large prompts.
@@ -167,6 +173,7 @@ class LLM(OpenAILLM):
             parallel_tool_calls=parallel_tool_calls,
             tool_choice=tool_choice,
             reasoning_effort=reasoning_effort,
+            reasoning_format=reasoning_format,
             safety_identifier=safety_identifier,
             prompt_cache_key=prompt_cache_key,
             top_p=top_p,
--- a/livekit-plugins/livekit-plugins-cerebras/livekit/plugins/cerebras/llm.py
+++ b/livekit-plugins/livekit-plugins-cerebras/livekit/plugins/cerebras/llm.py
@@ -130,7 +130,8 @@ class LLM(OpenAILLM):
 
         ``reasoning_format`` controls how reasoning models (e.g. ``gpt-oss-120b``) return their
         thinking tokens. Set it to ``"hidden"`` or ``"parsed"`` to keep the model's internal
-        monologue out of the spoken message content.
+        monologue out of the spoken message content; ``"none"`` keeps the model's default
+        behavior.
 
         When ``gzip_compression`` is True (default), request payloads are gzip-compressed,
         which can reduce TTFT for requests with large prompts.
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -61,6 +61,7 @@ lk_oai_debug = int(os.getenv("LK_OPENAI_DEBUG", 0))
 
 Verbosity = Literal["low", "medium", "high"]
 PromptCacheRetention = Literal["in_memory", "24h"]
+ReasoningFormat = Literal["parsed", "raw", "hidden"]
 
 
 @dataclass
@@ -78,6 +79,7 @@ class _LLMOptions:
     max_completion_tokens: NotGivenOr[int]
     service_tier: NotGivenOr[str]
     reasoning_effort: NotGivenOr[ReasoningEffort]
+    reasoning_format: NotGivenOr[ReasoningFormat]
     verbosity: NotGivenOr[Verbosity]
     prompt_cache_retention: NotGivenOr[PromptCacheRetention]
     extra_body: NotGivenOr[dict[str, Any]]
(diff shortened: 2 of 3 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -120,6 +123,12 @@ def __init__(
 
         ``api_key`` must be set to your OpenAI API key, either using the argument or by setting the
         ``OPENAI_API_KEY`` environmental variable.
+
+        ``reasoning_format`` controls how reasoning models (e.g. ``gpt-oss-120b`` served by
```
davidzhao: nit: reword the description. it makes it sound like `gpt-oss-120b` can be served by xAI

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
```diff
@@ -120,6 +123,12 @@ def __init__(
...
+        Cerebras or xAI) return their thinking tokens. Set it to ``"hidden"`` or ``"parsed"`` to
```
davidzhao: how do we handle parsed? if we are parsing it, there should be a way to expose it to the end user.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
anshulkulhari7: Reworded in 258fde2 — now reads "`gpt-oss-120b` on Cerebras, or Grok on xAI", so it no longer implies xAI serves gpt-oss-120b.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
anshulkulhari7: Good catch — right now the plugin doesn't surface the parsed reasoning. `ChoiceDelta` only has `role`/`content`/`tool_calls`/`extra` (no dedicated reasoning field), and the OpenAI stream handler doesn't read a `reasoning`/`reasoning_content` field, so with `reasoning_format="parsed"` the separated reasoning would currently be dropped.

My instinct is to surface it through `delta.extra` (the same provider-extra channel already used for xAI encrypted reasoning) rather than adding a new field — but how would you prefer reasoning to be exposed to the end user? Happy to wire it up that way in this PR, or drop `"parsed"` here and do the exposure as a focused follow-up if you'd rather keep this one to the request-param plumbing.

immuhammadfurqan:
Thanks for tackling this @anshulkulhari7 — the `extra_body` routing and covering all four entry points (`__init__`, `with_cerebras`, `with_x_ai`, cerebras plugin) looks clean. Two things I hit while cross-checking the provider docs, in case they're useful:

**1. Does xAI/Grok actually support `reasoning_format`?** xAI's documented reasoning control is `reasoning_effort` (`none`/`low`/`medium`/`high`) — I couldn't find `reasoning_format` anywhere in their API reference or reasoning guide. If Grok silently ignores the field, `with_x_ai(reasoning_format=...)` would be a no-op and the original goal (keeping Grok's thinking tokens out of TTS) wouldn't actually be met. Did you get a chance to verify it against a live xAI endpoint? ([xAI reasoning guide](https://docs.x.ai/docs/guides/reasoning))

**2. Cerebras also documents a `none` value.** The `ReasoningFormat` literal is `"parsed" | "raw" | "hidden"`, but Cerebras lists `parsed | raw | hidden | none` (`none` = model default) — might be worth including for completeness. ([Cerebras reasoning docs](https://inference-docs.cerebras.ai/capabilities/reasoning))

FWIW `parsed` and `hidden` are the two that keep reasoning out of the `content` field on Cerebras, so the docstring guidance there is spot on. :+1:

anshulkulhari7:
Thanks for cross-checking the provider docs — both points were right, and both are addressed in 180af172c:

1. **xAI:** confirmed — `reasoning_format` doesn't appear anywhere in xAI's API reference; their documented reasoning controls are `reasoning_effort` and the Responses API `include: ["reasoning.encrypted_content"]`. I don't have a live xAI endpoint to probe for undocumented support, so rather than ship a field Grok would likely ignore silently, I've removed `reasoning_format` from `with_x_ai()` and scoped the feature to Cerebras only (docstrings + tests updated, PR title/body updated to match). The pre-existing `# TODO(long): add provider fmt for grok` comment in `with_x_ai` suggests Grok format handling was already known to be an open item — if xAI documents a control later it can ride the same `extra_body` plumbing.

2. **`none` added:** `ReasoningFormat` is now `"parsed" | "raw" | "hidden" | "none"` per the Cerebras docs, with a test asserting it's forwarded verbatim.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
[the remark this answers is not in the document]
anshulkulhari7: This is intentional, not an omission — `reasoning_format` was deliberately **removed** from `with_x_ai` in 180af172c after review feedback: xAI's API has no `reasoning_format` field (its documented reasoning controls are `reasoning_effort` and the Responses API `include: ["reasoning.encrypted_content"]`), so exposing it there would have been a silently-ignored option. The PR description's scope note covers this; the PR is now scoped to Cerebras only.

livekit-plugins/livekit-plugins-cerebras/livekit/plugins/cerebras/llm.py
```diff
@@ -126,6 +128,11 @@ def __init__(
         ``api_key`` must be set to your Cerebras API key, either using the argument or by setting
         the ``CEREBRAS_API_KEY`` environmental variable.
 
+        ``reasoning_format`` controls how reasoning models (e.g. ``gpt-oss-120b``) return their
+        thinking tokens. Set it to ``"hidden"`` or ``"parsed"`` to keep the model's internal
+        monologue out of the spoken message content; ``"none"`` keeps the model's default
+        behavior.
```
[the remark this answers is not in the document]
anshulkulhari7: `"none"` is a documented Cerebras value, not an invalid one — their reasoning docs list `parsed | raw | hidden | none`, with `none` defined as "uses model's default behavior" (https://inference-docs.cerebras.ai/capabilities/reasoning). So forwarding it verbatim is the correct implementation of that semantic, and the docstring matches the provider's definition. The "only parsed/raw/hidden" list you're quoting is from the pre-review PR description; the current description and `ReasoningFormat` literal include all four.

PR #6107, opened 2026-06-15
fix(voice): scope playout state per generation step in SpeechHandle

### LINKED ISSUE #5545: Track voice playout and pause state by generation step instead of reused SpeechHandle
### Feature Type

Would make my life easier

### Feature Description

Today `SpeechHandle` represents a full assistant turn, but some state on it is generation-step local. In multi-step turns with tool replies, the same handle is reused while playout/pause state belongs to the current reply step only.

Proposed design change: add a private generation-step ref inside `SpeechHandle` and use it for playout and false-interruption bookkeeping. Playout callbacks, pause state, and cleanup timers should only mutate state when their captured generation ref still matches the handle’s current generation.

This keeps `SpeechHandle` as the public whole-turn handle while making per-step runtime state explicit internally.

Another alternative would be to create a new public SpeechHandle for each tool/reply step, but that seems like a much larger API and behavior change.

### Additional Context

This is a follow-up design issue after the stale playout-state fix(https://github.com/livekit/agents/pull/5533).

anshulkulhari7:
## Summary

Follow-up to #5594 (the cross-step stale-paused-speech fix).

`SpeechHandle` is the public whole-turn handle, but a multi-step turn reuses the
same handle across generation steps while some state (playout / pause) is
generation-step local. #5594 resets `_paused_speech` in `_scheduling_task` *after*
the current generation's `_wait_for_generation` completes, which covers the
cross-step leak. It does **not** cover the case @jayeshp19 flagged in #5533:

> the false-interruption timer fires **before** the silent tool-call generation
> finishes. In that case the runtime can still do a `thinking → listening →
> thinking` false-resume **during the silent step itself** … the pause-on-thinking
> path can still record `_paused_speech`.

This is the "pausing while no audio is actually playing" edge case, and it is what
issue #5545 proposes to fix by scoping playout/pause state per generation step.

## The bug (reproduced)

1. User asks "what's the weather in Tokyo?"
2. The model emits only a function call — a **silent** tool-call step, no audio.
3. VAD-only speech starts and ends during that silent step, so the pause-on-thinking
   path records `_paused_speech` and the false-interruption timer is scheduled.
4. The timer fires **while the silent step is still active** (before it advances to
   the tool reply).

On `main`, `_on_false_interruption` resumes and emits
`AgentFalseInterruptionEvent(resumed=True)` even though no audio ever played for
this step — leaking stale per-step playout state. The added regression test
asserts no such resume is emitted; it fails on `main`:

```
>       assert not any(ev.resumed for ev in false_interruption_events)
E       assert not True
tests/test_agent_session.py: AssertionError
```

## Approach

Mirroring the issue's proposal ("a private generation-step ref … playout callbacks,
pause state, and cleanup timers should only mutate state when their captured
generation ref still matches the handle's current generation"), this scopes playout
state to the current step:

- `SpeechHandle._playout_started` — set when the current step starts audio playout
  (in the existing `_on_first_frame` callbacks), reset on every step advance
  (`_authorize_generation`). It records whether the *current* step is actually
  playing audio.
- `_on_false_interruption` now only treats a pause as a resumable false interruption
  when the current step actually started playout. For a silent step it undoes the
  preemptive pause silently and emits no resume event.

The change is additive and backward-compatible. The existing #5594 regression test
(`test_silent_tool_call_pause_state_does_not_leak_into_tool_reply`) and the rest of
the interruption suite still pass, since a real tool reply has started playout by the
time its timer fires.

## Question for maintainers

@longcw @davidzhao — does this per-step `_playout_started` gating match the internal
direction you intended for #5545? I kept it minimal (a per-step playout flag + a
single early-return in the false-interruption timer) rather than introducing a
broader public per-step handle. Happy to fold playout-state tracking more fully into
a generation-step ref if you'd prefer that shape.

## Verification

- `uv run pytest tests/test_agent_session.py --unit` — 34 passed (incl. new test and
  the #5594 regression test)
- `uv run ruff check` / `ruff format --check` — clean on changed files
- `uv run mypy -p livekit.agents` (strict) — Success, no issues

Closes #5545

---

*AI-assisted: implemented with AI assistance; all changes were reviewed, tested, and verified by the author.*

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2713,6 +2713,7 @@ class AgentActivity(RecognitionHooks):
             except BaseException:
                 return
 
+            speech_handle._mark_generation_playout_started()
             self._session._update_agent_state(
                 "speaking",
                 start_time=started_speaking_at,
@@ -3174,6 +3175,7 @@ class AgentActivity(RecognitionHooks):
                 )
             self._session._early_assistant_metrics = early_metrics
 
+            speech_handle._mark_generation_playout_started()
             self._session._update_agent_state(
                 "speaking",
                 start_time=started_speaking_at,
@@ -3708,6 +3710,7 @@ class AgentActivity(RecognitionHooks):
             except BaseException:
                 return
 
+            speech_handle._mark_generation_playout_started()
             self._session._update_agent_state(
                 "speaking",
                 start_time=started_speaking_at,
@@ -4166,6 +4169,22 @@ class AgentActivity(RecognitionHooks):
             ):
                 # already new speech is scheduled, do nothing
                 self._paused_speech = None
+                return
+
+            if not self._paused_speech.handle._generation_playout_started:
+                # The current generation step never started audio playout (e.g. a
+                # silent tool-call step). The pause was recorded preemptively, so
+                # there is nothing to "resume" — undo the preemptive pause without
+                # emitting a false-interruption resume that would otherwise leak
+                # stale playout state into a later step.
+                if (
+                    self._session.options.interruption["resume_false_interruption"]
+                    and (audio_output := self._session.output.audio)
+                    and audio_output.can_pause
+                ):
+                    audio_output.resume()
+                self._paused_speech = None
+                self._false_interruption_timer = None
                 return
 
             resumed = False
--- a/livekit-agents/livekit/agents/voice/speech_handle.py
+++ b/livekit-agents/livekit/agents/voice/speech_handle.py
@@ -50,6 +50,18 @@ class SpeechHandle:
         self._num_steps = 1
         self._agent_turn_context: otel_context.Context | None = None
 
+        # per generation-step playout state.
+        #
+        # ``SpeechHandle`` is the public whole-turn handle, but a multi-step turn
+        # (e.g. a silent tool-call step followed by a tool-reply step) reuses the
+        # same handle across steps. ``_playout_started`` is scoped to the current
+        # step: it becomes ``True`` once the step produces audio playout and is
+        # reset on every step advance (see ``_authorize_generation``). It lets the
+        # runtime tell whether the *current* step is actually playing audio, so
+        # pause/false-interruption bookkeeping captured during a silent step does
+        # not leak into a later step.
+        self._playout_started = False
+
         self._interrupt_timeout_handle: asyncio.TimerHandle | None = None
 
         self._item_added_callbacks: set[Callable[[llm.ChatItem], None]] = set()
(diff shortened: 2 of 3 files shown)


tsushanth (approved):
Solid fix. The reproduction in the PR body is precise, the regression test exhibits the bug on `main`, and the chosen scope (per-step `_playout_started` flag) is the right minimal shape for closing #5545 without dragging in the broader per-step generation-ref refactor that the issue gestures at. A few items worth checking before merge but nothing structural.

### Worth verifying

**1. Coverage of all audio-playout entry points.**

This PR patches three `_on_first_frame` callbacks in `livekit-agents/livekit/agents/voice/agent_activity.py` (around lines 2411, 2806, and 3299), each adding `speech_handle._mark_generation_playout_started()` immediately before the `"speaking"` state transition. If there's a fourth audio-start callsite anywhere (a different playout path that emits the "speaking" state without going through one of these three callbacks), it would skip the marker, leave `_playout_started=False` while audio actually plays, and the new guard in `_on_false_interruption` would then *suppress a legitimate resume event* on that path.

A quick scope check like `rg '_update_agent_state\(\s*"speaking"' livekit-agents` to confirm three matches and no others — or `rg 'on_first_frame|on_playout_started' livekit-agents` to confirm three callbacks total — would close that risk. The bug being prevented (false suppress) is the inverse of the bug being fixed, so it's worth a moment of due diligence.

**2. Timer cleanup symmetry in `_on_false_interruption`.**

The new early-return branch in `agent_activity.py` nullifies the timer explicitly:

```python
self._paused_speech = None
self._false_interruption_timer = None
return
```

Whereas the existing audio-bearing path further down doesn't appear to do that in the diff context. If the existing path also resets the timer somewhere downstream, this is consistent. If not, the new branch is doing extra cleanup the old one doesn't — defensible, but worth a one-line comment explaining why the silent-step path specifically needs explicit nullification (e.g. "no resume event is emitted, so downstream resume-completion teardown won't fire and clear the timer for us").

### Worth discussing

**A.** `_generation_playout_started` (property in `speech_handle.py`) duplicates `_playout_started` (the field at the same line):

```python
# speech_handle.py
self._playout_started = False
...
@property
def _generation_playout_started(self) -> bool:
    return self._playout_started

def _mark_generation_playout_started(self) -> None:
    self._playout_started = True
```

The property is a pure pass-through. Two clean options:

- Drop the property entirely; have the caller in `_on_false_interruption` access `self._paused_speech.handle._playout_started` directly. Less indirection, same readability since the field name is already descriptive.
- Or keep the property and add an `@property.setter` for the write side, so there's exactly one symbol pair (`_generation_playout_started` getter+setter) instead of three names referring to the same state.

The current shape forces a reader to verify three names point at one piece of state. Not a blocker, but the indirection costs more than it earns when the property does nothing beyond returning the field.

**B.** On the design question raised in the PR body (per-step generation refs vs the minimal `_playout_started` flag): I'd argue the minimal approach is right here. The bug is narrowly scoped (playout state leaking across steps), and a broader per-step ref refactor would expand the public/internal surface (the issue gestures at "a private generation-step ref … passed through playout callbacks") without a current second use case. If a future bug surfaces another piece of state that needs per-step scoping, the broader refactor can be done then with two concrete examples to design against. YAGNI applies — landing the narrow fix now keeps the change reviewable and reversible.

### Decision

Approving on the strength of the reproduction + regression test + correct minimal scope. The two verification items above are quick to confirm and would tighten merge confidence; the property/field consolidation is a follow-up nit, not a blocker.

anshulkulhari7:
Rebased onto current main (`105d952ea`) to clear the merge conflict — the diff is unchanged from the approved revision (same 3 files, +116). Re-verified before pushing: the regression test still fails on current main with only the source files reverted, and the full unit gate shows exactly +1 pass vs baseline.

PR #6111, opened 2026-06-15
fix(worker): log the immediate first reconnect below WARNING (closes #6108)

### LINKED ISSUE #6108: Worker logs first control-plane reconnect at WARNING ("retrying in 0s") for expected, self-healing churn
### Summary

`Worker._run` logs the **first** control-plane websocket reconnect at `WARNING`
with `retrying in 0s`, even though that first immediate retry is expected churn
that almost always succeeds. In a long-running deployment this produces a steady
trickle of self-healing `WARNING`s that trip warning-level alerting, with nothing
actionable behind them. It would be better to log the immediate first reconnect
below `WARNING` and only escalate to `WARNING` once a retry actually has to back
off (i.e. from the second attempt / non-zero delay onward).

### Where

`livekit/agents/worker.py`, `Worker._run` (livekit-agents **1.5.17**):

```python
retry_delay = min(retry_count * 2, 10)
retry_count += 1

logger.warning(
    f"failed to connect to livekit, retrying in {retry_delay}s",
    extra={"retry_count": retry_count, "max_retry": self._max_retry, "error": e},
)
await asyncio.sleep(retry_delay)
```

and just after a successful connect:

```python
ws = await self._http_session.ws_connect(...)
retry_count = 0
```

### Why it's always `0s`

Because `retry_count` is reset to `0` the moment a connection succeeds, the first
reconnect after any websocket drop computes `retry_delay = min(0 * 2, 10) = 0s`.
The long-lived control-plane websocket drops fairly often in cloud environments
(LB/proxy idle timeouts, edge nodes recycling, transient network blips). Each drop
emits:

```
failed to connect to livekit, retrying in 0s
```

at `WARNING`, then reconnects immediately and succeeds — so `retry_count` resets
to `0` and it never escalates. The `WARNING` is, in practice, "a websocket
blipped and we already recovered."

We see these a few times a day across worker replicas, every one recovering, no
real outage behind any of them.

### Suggested change

Treat the immediate first reconnect as expected and reserve `WARNING` for retries
that actually have to wait. The simplest fix needs no new config: log below
`WARNING` when `retry_delay == 0` (the first reconnect after `retry_count` resets)
and at `WARNING` from the first backed-off attempt onward:

```python
level = logging.WARNING if retry_delay > 0 else logging.DEBUG
logger.log(
    level,
    f"failed to connect to livekit, retrying in {retry_delay}s",
    extra={"retry_count": retry_count, "max_retry": self._max_retry, "error": e},
)
```

This keeps real, sustained connection problems loud (escalating `2s/4s/.../10s`
retries and the terminal `failed to connect to livekit after N attempts`) while
removing the self-healing first-hop noise.

If you'd rather make it configurable, the natural home is `WorkerOptions` /
`AgentServer` — the worker owns this control-plane connection — e.g. a
`log_connection_retries` toggle, or a minimum-attempt threshold below which
reconnects log at `DEBUG`. It would not belong at the session level: this
websocket is sessionless and lives in the worker's main process, so a per-session
option could never reach it. Either way the level fix above already covers the
common case without new API surface, so a flag is optional.

### Workaround we're using

For now we demote only the literal `failed to connect to livekit, retrying in 0s`
record from `WARNING` to `INFO` via a `logging.Filter` on the `livekit.agents`
logger, leaving every escalating retry and the terminal failure at `WARNING` so a
genuine outage still alerts.

tsushanth:
Closes #6108.

`Worker._run` logs every control-plane reconnect attempt at `WARNING`, including the very first one whose computed `retry_delay` is `0`. That first immediate retry is expected churn — control-plane websocket lifecycle, NLB / proxy idle timeouts, brief network hiccups — and it almost always succeeds. In long-running deployments this produces a steady trickle of self-healing `WARNING`s that trip warning-level alerting with nothing actionable behind them; the operator reads "WARNING: failed to connect ... retrying in 0s" once per blip and learns to filter all of them out, which then masks the real backoff events that actually warrant attention.

The fix is a one-line conditional on the log level: `INFO` while `retry_delay == 0` (the immediate first attempt), `WARNING` once a retry has to back off (second attempt onward, where `retry_count * 2` becomes non-zero). The message and `extra={retry_count, max_retry, error}` payload are unchanged, so existing log scrapers / dashboards keyed on the message text keep working — only the level shifts for the first attempt.

No tests touch this path; the change is a log-level adjustment with no behavior change otherwise. Verified locally that the second iteration still emits `WARNING ... retrying in 2s` and the immediate first emits `INFO ... retrying in 0s`.

### DIFF
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -1134,7 +1134,14 @@ class AgentServer(utils.EventEmitter[EventTypes]):
                 retry_delay = min(retry_count * 2, 10)
                 retry_count += 1
 
-                logger.warning(
+                # The first reconnect runs at delay=0 and almost always succeeds —
+                # it's expected churn for control-plane websocket lifecycle and
+                # produces a steady trickle of self-healing WARNINGs that trip
+                # warning-level alerting with nothing actionable behind them.
+                # Only escalate to WARNING once a retry actually has to back
+                # off (second attempt onward, where retry_delay > 0).
+                log_fn = logger.info if retry_delay == 0 else logger.warning
+                log_fn(
                     f"failed to connect to livekit, retrying in {retry_delay}s",
                     extra={"retry_count": retry_count, "max_retry": self._max_retry, "error": e},
                 )
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -1052,6 +1052,7 @@ class AgentServer(utils.EventEmitter[EventTypes]):
         assert self._http_session is not None
 
         retry_count = 0
+        had_successful_connection = False
         ws: aiohttp.ClientWebSocketResponse | None = None
         while not self._closed:
             try:
@@ -1116,6 +1117,7 @@ class AgentServer(utils.EventEmitter[EventTypes]):
 
                 self._handle_register(msg.register)
                 self._connecting = False
+                had_successful_connection = True
 
                 # report all active jobs to the server after registration
                 await self._report_active_jobs()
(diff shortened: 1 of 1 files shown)


tsushanth:
Good catch — that's a real gap. The previous patch collapsed two distinct cases under `retry_delay == 0`:

1. **Boot-time first attempt** — wrong URL / bad API key / cert error. Operators want to see this at WARNING.
2. **Reconnect immediately after a successful connection drops** — control-plane churn, almost always self-heals.

Pushed [d9dda52](https://github.com/livekit/agents/pull/6111/commits/d9dda52) adding a `had_successful_connection` flag set right after `_handle_register` returns. The INFO downgrade now only applies when `had_successful_connection and retry_delay == 0` — the post-success churn path. The very first attempt (and every backoff retry of it) stays at WARNING, so misconfig at boot is still visible.

tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #6113, opened 2026-06-15
fix(voice): honor dict form of preemptive_generation kwarg (closes #6112)

### LINKED ISSUE #6112: Bug: preemptive_generation={"enabled": False} and endpointing={ "min_delay": 1 } is ignored — session report always shows defalt values
### Bug Description

Passing `preemptive_generation={"enabled": False}` to `Agent(...)` does not disable preemptive generation. The end-of-session report (`ctx.make_session_report().to_dict()`) shows `options.preemptive_generation.enabled` as `True` regardless of the value passed at session creation. Same for endpointing={ "min_delay": 1 }, its always using 0.5 default

### Expected Behavior

`report_dict["options"]["preemptive_generation"]["enabled"]` should be `False`, matching the value explicitly passed when constructing the session. Preemptive generation should not run during the session (no speculative LLM/TTS calls before end-of-turn is detected).

### Actual behavior
`report_dict["options"]` shows `preemptive_generation.enabled` as `True`, along with default values for the other sub-fields:
```python
{
    'allow_interruptions': True,
    'discard_audio_if_uninterruptible': True,
    'min_interruption_duration': 0.5,
    'min_interruption_words': 0,
    'min_endpointing_delay': 0.5,
    'max_endpointing_delay': 3.0,
    'max_tool_steps': 3,
    'user_away_timeout': 30,
    'min_consecutive_speech_delay': 0.0,
    'preemptive_generation': {
        'enabled': True,
        'preemptive_tts': False,
        'max_speech_duration': 10.0,
        'max_retries': 3,
    },
}
```

The `enabled: False` passed by the caller has no effect — the reported (and apparently effective) value is always `True`.

### Reproduction Steps

```bash
1. Create an `AgentSession` and explicitly disable preemptive generation:




class TempAgent(Agent)
    def __init__(self):
        super().__init__(
                turn_handling=TurnHandlingOptions(
                turn_detection=build_turn_detection(),
                endpointing={
                    "mode": "fixed",
                    "min_delay": 0.2,
                    "max_delay": 1.2,
                },
                interruption={
                    "enabled": True,
                    "resume_false_interruption": True,
                    "mode": "adaptive",
                    "min_words": 2,
                },
                preemptive_generation={"enabled": False},
            ),
        )


session = AgentSession(
    userdata=userdata
)

await session.start(
            agent=TempAgent,
            room=ctx.room,
  )


2. Run the session through to completion.
3. At session end, generate the session report:


report_dict = userdata.ctx.make_session_report().to_dict()
print(report_dict["options"]["preemptive_generation"])
```

### Operating System

livekit cloud

### Models Used

deepgram nova3, openai gpt 5.4 mini, cartesia tts

### Package Versions

```bash
- `livekit-agents` version: 1.6.0
- `livekit-api` version: 1.1.0
- Python version:  3.14
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional context

If i pass these settings directly in AgentSession than it works, but we need to have different agent having different turn handling and other settings, not at root agent session level

### Impact
Anyone trying to opt out of preemptive generation (e.g., to avoid extra LLM token usage from speculative responses, or to avoid duplicate `on_user_turn_completed`/`tts_node` invocations as described in #3414) cannot reliably do so via `preemptive_generation={"enabled": False}`, and cannot trust the session report to confirm whether it was actually disabled.

_No response_

tsushanth:
Closes #6112.

The deprecated `preemptive_generation` keyword argument on `AgentSession` (and the same migration path used by `Agent`) is documented as accepting either a bare `bool` (legacy shorthand) or a full `PreemptiveGenerationOptions` dict — `agent_session.py:322` says exactly `NotGivenOr[bool | PreemptiveGenerationOptions]`. But the implementation at `turn.py:298-299` was:

```python
if is_given(preemptive_generation):
    result["preemptive_generation"] = {"enabled": preemptive_generation}
```

That blindly wraps the input in `{"enabled": <input>}`. When the input is already a dict, you get `{"enabled": {"enabled": False, ...}}`. The resolver merges that with defaults, the session-report serializer reads `enabled` as a truthy dict, and the report shows `preemptive_generation.enabled: True` regardless of what the caller passed. The bug is in the dict form only — the bool form was correct.

## Fix

Two parts:

- **`turn.py`**: in `_migrate_turn_handling`, branch on `isinstance(preemptive_generation, dict)`. Dict form passes through untouched; bool form keeps the legacy `{"enabled": <bool>}` wrap. Type annotation widened from `NotGivenOr[bool]` to `NotGivenOr[bool | PreemptiveGenerationOptions]` to match the docstring.
- **`agent_session.py`**: matching widening of the `preemptive_generation` parameter type on `AgentSession.__init__` (still marked deprecated; the recommended path remains `turn_handling={"preemptive_generation": {...}}`).

The bool form for `Agent(turn_detection=..., allow_interruptions=False, ...)` and `AgentSession(preemptive_generation=False)` paths are unchanged. The recommended `turn_handling=` path (used by every test in `tests/test_agent_session.py::test_preemptive_generation`) doesn't go through `_migrate_turn_handling` at all and is untouched.

## Tests

New `tests/test_migrate_turn_handling.py` with 5 cases on the migration shim directly:

- `preemptive_generation=False` → `{"preemptive_generation": {"enabled": False}}` (legacy bool, unchanged)
- `preemptive_generation=True`  → `{"preemptive_generation": {"enabled": True}}` (legacy bool, unchanged)
- `preemptive_generation={"enabled": False}` → `{"preemptive_generation": {"enabled": False}}` (the bug case — fails on main with `{"enabled": {"enabled": False}}`)
- `preemptive_generation={"enabled": True, "preemptive_tts": True, "max_speech_duration": 5.0, "max_retries": 1}` → passes through untouched (the second regression, where dict form is the only way to set non-`enabled` fields via the deprecated kwarg)
- Omitting the kwarg yields no `preemptive_generation` key in the result

```
$ pytest tests/test_migrate_turn_handling.py --unit -q
5 passed in 0.03s
```

Without the fix, cases 3 and 4 fail with the exact double-wrap shape from #6112.

## The reporter's second complaint (`endpointing={"min_delay": 1}`)

The reporter mentions `endpointing={ "min_delay": 1 }` is also ignored. That kwarg doesn't exist on `AgentSession`/`Agent` directly — only the underscore form `min_endpointing_delay=1` (deprecated) or `turn_handling={"endpointing": {"min_delay": 1}}` (current). I don't think there's a real bug there to fix — likely the report conflated two different calls. Happy to add a follow-up if I'm reading that wrong.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -248,7 +248,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         conn_options: NotGivenOr[SessionConnectOptions] = NOT_GIVEN,
         loop: asyncio.AbstractEventLoop | None = None,
         # deprecated
-        preemptive_generation: NotGivenOr[bool] = NOT_GIVEN,
+        preemptive_generation: NotGivenOr[bool | PreemptiveGenerationOptions] = NOT_GIVEN,
         min_endpointing_delay: NotGivenOr[float] = NOT_GIVEN,
         max_endpointing_delay: NotGivenOr[float] = NOT_GIVEN,
         false_interruption_timeout: NotGivenOr[float | None] = NOT_GIVEN,
--- a/livekit-agents/livekit/agents/voice/turn.py
+++ b/livekit-agents/livekit/agents/voice/turn.py
@@ -258,7 +258,7 @@ def _migrate_turn_handling(
     allow_interruptions: NotGivenOr[bool] = NOT_GIVEN,
     resume_false_interruption: NotGivenOr[bool] = NOT_GIVEN,
     agent_false_interruption_timeout: NotGivenOr[float | None] = NOT_GIVEN,
-    preemptive_generation: NotGivenOr[bool] = NOT_GIVEN,
+    preemptive_generation: NotGivenOr[bool | PreemptiveGenerationOptions] = NOT_GIVEN,
 ) -> TurnHandlingOptions:
     """Build a TurnHandlingOptions from deprecated keyword arguments."""
     if is_given(agent_false_interruption_timeout):
@@ -296,6 +296,18 @@ def _migrate_turn_handling(
         result["turn_detection"] = turn_detection
 
     if is_given(preemptive_generation):
-        result["preemptive_generation"] = {"enabled": preemptive_generation}
+        # `preemptive_generation` is documented (and now typed) as
+        # `bool | PreemptiveGenerationOptions`. The bool form is the legacy
+        # shorthand for `{"enabled": <bool>}`; the dict form passes through
+        # so callers can also set `preemptive_tts` / `max_speech_duration`
+        # / `max_retries`. Without this branch the dict was re-wrapped as
+        # `{"enabled": {"enabled": False}}`, the `_resolve_preemptive_generation`
+        # merge produced a truthy `enabled` value, and the session report
+        # showed `enabled: True` regardless of what the caller passed
+        # (https://github.com/livekit/agents/issues/6112).
+        if isinstance(preemptive_generation, dict):
+            result["preemptive_generation"] = preemptive_generation
+        else:
+            result["preemptive_generation"] = {"enabled": preemptive_generation}
 
     return result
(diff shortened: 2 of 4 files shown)


longcw:
closing in favor of https://github.com/livekit/agents/pull/6114

PR #6116, opened 2026-06-15
feat: expressive mode

tinalenguyen:
LLM-driven TTS prosody via provider-specific markup tags, end to end:

- Provider-keyed markup pipeline in tts/_provider_format.py (LLM tag instructions, transcript stripping, normalize + convert to native syntax); base TTS.Markup delegates on _provider_key() so plugins override one method.
- In-house per-provider input chunking: when expressiveness is on, sentences batch up to the provider send limit (Inworld 900, Cartesia 400) for continuous prosody; per-sentence otherwise. Defaults live in _provider_format.
- ExpressiveContent captured on assistant messages and re-injected on later turns so the model sees its own expressive style.
- Conversational / customer-service / healthcare presets, wired into the drive-thru, frontdesk, healthcare and survey examples.
- Inference gateway + cartesia/elevenlabs/inworld plugin markup; assorted STT/telemetry/audio-recognition updates carried on the branch.


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -35,148 +35,117 @@
     from ..llm import LLM, Tool, Toolset
 
 
-class Instructions(str):
-    """Instructions that adapt based on the user's input modality (audio vs. text).
+class Instructions:
```
Bobronium, line 38: I suggest keeping `Instructions` a  `str` subclass:


```suggestion
class Instructions(str):
```

Reasons articulated well in [Devin's comment](https://github.com/livekit/agents/pull/6116#discussion_r3416010800):


> The `Instructions` class was previously a `str` subclass (inheriting all string behavior). It's now a regular class with `__str__` returning `self.common`. This is a significant API break for any external code that relied on `isinstance(instr, str)` being True, or passing `Instructions` directly where a `str` was expected. [...], any downstream plugin or user code that stored `Instructions` objects directly in `ChatMessage.content` (relying on `Instructions` being a `str`) would silently break.

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -35,148 +35,117 @@
...
 
-    def __new__(
-        cls, audio: str, *, text: str | None = None, _represent: str | None = None
-    ) -> Instructions:
-        """Create an Instructions object.
+        instr.render()                              # → common text
+        instr.render(modality="audio")               # → common + audio addition
+        instr.render(modality="text", name="Alex")   # → common + text, with {name} filled
+    """
+
```
Bobronium, line 59: Goes together with [previous suggestion](https://github.com/livekit/agents/pull/6116#discussion_r3420152959): 

```suggestion

    def __new__(cls, common: str = "", **kwargs: Any) -> Instructions:
        return super().__new__(cls, common)

```

`__init__` is then supplementary for setting instance attributes.

Bobronium (dismissed):

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -68,6 +68,86 @@ class TTS(
...
+        ) -> AsyncGenerator[str, None]:
+            """Strip TTS markup from a stream of text chunks.
+
+            Buffers partial XML tags across chunks so that ``to_text`` always
+            receives complete tags to strip.
+            """
+            buf = ""
+            async for chunk in text_stream:
+                buf += chunk
+                if buf.rfind("<") > buf.rfind(">"):
```
chenghao-mou: Not a huge issue right now, but not all vendors use XML syntax. e.g. [https://docs.fish.audio/developer-guide/core-features/emotions](<https://docs.fish.audio/developer-guide/core-features/emotions>) use `[]`

livekit-agents/livekit/agents/tts/_provider_format.py
```diff
@@ -0,0 +1,450 @@
...
+
+
+# --- Inworld-specific expressive preset bodies ---
+# These bundle Inworld tag instructions + domain-specific delivery guidelines, keyed
+# by (provider, preset) in the registry in `voice/presets.py`. The public, provider-
+# agnostic markers (`presets.CUSTOMER_SERVICE`, ...) resolve to one of these based on
+# the active TTS. They do NOT use the {tts.markup.llm_instructions} placeholder — the
+# Inworld tag reference is inlined directly, so the prompt is self-contained.
+
+_INWORLD_CUSTOMER_SERVICE: ExpressiveOptions = {
```
chenghao-mou, line 290: Q: Do we expect users to use these instructions for other languages too?

livekit-agents/livekit/agents/tts/markup_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+def convert_expression_tags(text: str) -> str:
+    """Convert ``<expression>`` and ``<sound>`` XML tags to ``[...]`` bracket format."""
+    text = _EXPRESSION_RE.sub(lambda m: f"[{m.group(1)}]", text)
+    text = _SOUND_RE.sub(lambda m: f"[{m.group(1)}]", text)
+    return text
+
+
+def convert_break_to_ellipsis(text: str) -> str:
+    """Replace ``<break time="..."/>`` tags with an ellipsis (``...``).
```
chenghao-mou: This might break a niche use case:

> User: Can you just stop for 10 seconds before responding?
>
> LLM: \<break time="10s"\>
>
> TTS: sees only "…"

definitely nitpicking here. maybe we should instruct the LLM know to use ellipsis for natural short pauses and only use break when needed.

livekit-agents/livekit/agents/tts/_provider_format.py
tinalenguyen: i believe we discussed that the LLM should be smart enough to translate, we can also test it via the demo

livekit-agents/livekit/agents/llm/chat_context.py
tinalenguyen: Hmm I think this change was introduced in the original branch theo/expressiveness-mode, the commit message associated with it was:

> Rework Instructions from str subclass to stateless class with common/audio/text fields. No Pydantic dependency, no runtime state.

I will double check on the intention there

livekit-agents/livekit/agents/tts/tts.py
tinalenguyen: do you think this further justifies LK having our own syntax that can be converted to XML and other syntax? thinking out loud

chenghao-mou: It is tricky. Even if we have our own syntax, different plugins support different values. Then we still need to translate something. If we only allow the same set of values, LLM will be generating unused/wasteful tokens. But it should be fine here as plugin can override this function.

livekit-agents/livekit/agents/llm/chat_context.py
tinalenguyen: i think the sentiment is that this change is fine, it was in beta and doesn't seem to be widely used

cc @theomonnom

examples/healthcare/agent.py
```diff
@@ -751,11 +757,41 @@ async def entrypoint(ctx: JobContext):
     session = AgentSession(
         userdata=userdata,
         stt=inference.STT("deepgram/nova-3", language="multi"),
-        llm=inference.LLM("openai/gpt-4.1-mini"),
-        tts=inference.TTS("inworld/inworld-tts-1"),
+        llm=inference.LLM("google/gemma-4-31b-it"),
+        tts=inference.TTS(
+            "inworld/inworld-tts-2",
+            voice="Luna",
+            extra_kwargs={"delivery_mode": "CREATIVE", "speaking_rate": 1.1},
```
bcherry, line 764: is this required to make expressive work?

tinalenguyen: speaking rate was preference from feedback but creative delivery_mode is optimal for expressive mode, not required

bcherry: Can't we just set it automatically?

livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -36,6 +39,29 @@ class ModelSettings:
 
 
 class Agent:
+    llm_output_format: type[BaseModel] | None = None
```
bcherry, line 42: this seems like a meaningful additional change right? Are we rewriting our examples and docs for structured output to use i?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -272,13 +273,14 @@ def __init__(
...
         self._backchannel_boundary_timer: asyncio.TimerHandle | None = None
-        self.backchannel_boundary_callback: Callable[[], None] | None = None
+        self._backchannel_boundary_callback: Callable[[], None] | None = None
         # endregion
 
         self._user_turn_span: trace.Span | None = None
         self._user_turn_start: float | None = None
         self._stt_request_ids: list[str] = []
         self._closing = asyncio.Event()
+        self.__stt_context: BaseModel | None = None
```
bcherry, line 285: Are we sure we're ready to release voice profile? Shouldn't this be an independent change?

theomonnom:
can you rebase on main? I see a lot of unrelated changes

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -239,6 +281,9 @@ def __init__(
         use_tts_aligned_transcript: NotGivenOr[bool] = NOT_GIVEN,
         tts_text_transforms: NotGivenOr[Sequence[TextTransforms] | None] = NOT_GIVEN,
         min_consecutive_speech_delay: float = 0.0,
+        # Expressive
+        expressive: bool | ExpressiveOptions = False,
+        strip_expressive_markup: bool = False,
```
```suggestion
```
What is this doing?

Also why it isn't inside ExpressiveOptions?

also nit: expressive vs expressiveness?

tinalenguyen: removed, it was leftover logic

we also decided on expressive mode

livekit-agents/livekit/agents/voice/audio_recognition.py
tinalenguyen: this has been removed

examples/healthcare/agent.py
tinalenguyen: it seems that inworld recommends stable mode for customer service use cases, maybe we should give the user freedom to set it? it may be personal preference

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2874,6 +2998,66 @@ def _end_segment() -> None:
...
+        async def _read_segment_text(
+            segment_text: AsyncIterable[str | BaseModel],
+        ) -> AsyncIterable[str]:
+            # structured-output BaseModel chunks carry no spoken text for the transcript
+            async def _plain_chunks() -> AsyncGenerator[str, None]:
+                async for chunk in segment_text:
+                    if isinstance(chunk, BaseModel):
+                        continue
+                    if chunk:
+                        yield chunk
```
longcw: `BaseModel` carries response content under `response_field_name`, "structured-output BaseModel chunks carry no spoken text for the transcript" is not true?

livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -509,12 +564,29 @@ async def tts_node(
...
             conn_options = activity.session.conn_options.tts_conn_options
             async with wrapped_tts.stream(conn_options=conn_options) as stream:
 
                 async def _forward_input() -> None:
+                    prev_response = ""
                     async for chunk in text:
-                        stream.push_text(chunk)
+                        if isinstance(chunk, str):
+                            stream.push_text(chunk)
+                        elif isinstance(chunk, BaseModel) and response_field:
```
longcw, line 594: I think we should move this to `_tts_inference_task` in generation.py before calling the `tts_node`, otherwise a `BaseModel` input breaks user's custom `tts_node`

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
```diff
@@ -973,6 +973,14 @@ def __init__(
             )
         )
 
+    class Markup(tts.TTS.Markup):
+        # markup delegation lives in the base class, keyed on _provider_key()
+        def _provider_key(self) -> str:
+            # only inworld-tts-2 understands the markup tags; older models get no
+            # markup so the tags aren't injected, converted, or stripped (matches
+            # the inference gateway's behavior)
+            return "inworld" if "tts-2" in self._tts.model else ""
```
longcw, line 982: in tts plugins, the `Markup` is defined but not used?

livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -35,148 +35,117 @@
...
-            ),
-            _represent=str(self).format(*args, **kwargs),
-        )
+            rendered = utils.misc.safe_render(template, kwargs)
+            return Instructions(common=rendered)
 
-    def as_modality(self, modality: Literal["audio", "text"]) -> Instructions:
-        """Return a copy whose ``str`` value is the correct variant for *modality*.
+    def __str__(self) -> str:
+        return self.common
```
longcw, line 131: `str(Instructions)` now returns only `common`:
- `agent_activity.py:858` — `_rt_session._update_session(instructions=str(self._agent.instructions))`: a realtime model running one of these tasks gets **empty instructions**
- `agent_activity.py:885` — the initial `AgentConfigUpdate` records `""` 
- `remote_session.py:647` — `GetAgentInfo` reports `""`.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2406,11 +2412,92 @@ def retrieve_chat_ctx(self) -> llm.ChatContext:
...
+        messages are replaced with copies rather than mutated.
+
+        Markup is only restored for messages captured under the same TTS (matched by
+        label): tag vocabularies differ between providers, so after a TTS handoff
+        another provider's tags would teach the LLM the wrong vocabulary.
+        """
+        if self.tts is None:
+            return
+
+        current_label = self.tts.label
```
longcw: for `inference.TTS` the label is always `livekit.agents.inference.tts.TTS`, the handoff from different providers (e.g. inworld to carteisa) won't restore the content.

livekit-agents/livekit/agents/inference/tts.py
```diff
@@ -635,7 +662,11 @@ async def _sentence_stream_task(ws: aiohttp.ClientWebSocketResponse) -> None:
             base_pkt["type"] = "input_transcript"
             async for ev in sent_tokenizer_stream:
                 token_pkt = base_pkt.copy()
-                token_pkt["transcript"] = ev.token + " "
+                # re-normalize at sentence level: tags split across input chunks
+                # aren't caught by the per-chunk normalize in _input_task
+                converted = self._tts.markup.convert(self._tts.markup.normalize(ev.token))
+                logger.debug("[TTS→API] %s", converted)
```
longcw:
```suggestion
```

livekit-agents/livekit/agents/inference/tts.py
```diff
@@ -646,13 +677,16 @@ async def _sentence_stream_task(ws: aiohttp.ClientWebSocketResponse) -> None:
                 token_pkt["generation_config"] = generation_config
                 token_pkt["extra"] = self._opts.extra_kwargs if self._opts.extra_kwargs else {}
                 self._mark_started()
-                await ws.send_str(json.dumps(token_pkt))
+                payload = json.dumps(token_pkt)
+                logger.debug("[TTS→gateway] %s", payload)
```
longcw:
```suggestion
```

livekit-agents/livekit/agents/tokenize/token_stream.py
```diff
@@ -18,13 +145,17 @@ def __init__(
         tokenize_fnc: TokenizeCallable,
         min_token_len: int,
         min_ctx_len: int,
+        max_token_len: int | None = None,
         retain_format: bool = False,
+        xml_aware: bool = True,
```
longcw: `_has_unclosed_xml_tags` treats any `<` after the last `>` as unclosed, so plain text like "3 < 5" holds all subsequent tokens until flush? If so, for a non-expressive agent this turns streaming TTS into end-of-turn batching for the rest of that turn. Consider enabling it only when expressive is active or optimize the hold?

livekit-agents/livekit/agents/tts/_provider_format.py
```diff
@@ -0,0 +1,561 @@
...
+    is set, it also raises the *minimum* so consecutive sentences are batched up to
+    that size, keeping prosody continuous across the turn; otherwise tokens emit per
+    sentence (the unchanged default). Providers with no configured limit are uncapped
+    and always per-sentence.
+    """
+    from .. import tokenize
+
+    max_len = _MAX_INPUT_LEN.get(provider)
+    return tokenize.blingfire.SentenceTokenizer(
+        max_token_len=max_len,
```
longcw, line 636: maybe worth to state in docs, `min_token_len` is raised to the provider max (900 chars for Inworld), so first audio per turn typically waits for the *entire* LLM turn.

I am wondering if this is really needed?

theomonnom: agree with that, let's verify if it's really needed (if expressiveness is clearly better). 

otherwise we need to update docs and make it clear that if you don't provide a tokenizer ourself, we pick the default depending on the used provider

theomonnom (approved):
lgtm!

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2874,6 +3063,68 @@ def _end_segment() -> None:
...
+                # expression lands on the segment's opening header (the frontend gets it
+                # at segment start). by here to_text_stream has collected the leading tag.
+                if not published:
+                    _publish_expressive_tags()
+                    published = True
+                yield chunk
+
+            # refresh on the flush trailer in case the leading expression was only
+            # collected after the header was sent
+            _publish_expressive_tags()
```
theomonnom: Move that to generation.py? 

Ideally the text source, receives the raw content from the LLM (including the tags). And the transcript forwarder implementation, strip the tags on its own.  (It may have to be done in both: TranscriptSynchronizer, and RoomTextOutput)

theomonnom (approved):
I would remove expressive_content

livekit-agents/livekit/agents/voice/io.py
```diff
@@ -402,6 +404,17 @@ async def capture_text(self, text: str) -> None:
...
+        """Attach extra attributes to the current (in-progress) output segment.
+
+        Used to surface side-channel metadata — e.g. expressive tags stripped
+        from the transcript — that should ride along with the segment on its next
+        flush. The base implementation forwards to the chained output; outputs that
+        publish to a transport (e.g. RoomIO transcription) override this to apply them.
+        """
+        if self.next_in_chain:
+            self.next_in_chain.set_segment_attributes(attributes)
+
```
theomonnom: If you extract the tags downstream, `set_segment_attributes` can be removed. (No new method needed inside io.py)

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -477,21 +501,33 @@ async def _text_forwarding_task(
...
+        async for chunk in display:
+            if text_output is not None and chunk:
+                await text_output.capture_text(chunk)
     finally:
         if isinstance(source, _ACloseable):
             await source.aclose()
 
         if text_output is not None:
-            text_output.flush()
+            text_output.flush(_expression_attributes(tags))
```
theomonnom: Do we need to get _expression_attributes there? Can't we do it directly inside the TranscriptSynchronizer? So nothing needs to be done inside generation.py?

theomonnom: Let's do that if it's feasible

tinalenguyen: the tags are stripped upstream of `TranscriptSynchronizer` so we don't duplicate the logic in `RoomTextOutput` as well

PR #6118, opened 2026-06-15
fix(azure): flush buffered audio on end_input to avoid truncated STT

jose-speak:
## Problem

When the input channel drained (`end_input`/`aclose`), `SpeechStream._run` broke out of the loop and tore the recognizer down via `stop_continuous_recognition` **without first closing the SDK push stream**. Azure was stopped mid-stream, so trailing phrase finals for already-buffered audio were never emitted and the transcript was truncated to the first phrase.

## Fix

Route the input-ended case through the same flush already used for reconnect:
- Close the push stream (signals end-of-stream so Azure emits the remaining finals)
- Await session-stopped
- Then exit the loop

## Verification

Verified against the real Azure service: a multi-phrase utterance pushed in full + `end_input` now returns the complete transcript instead of just the first phrase.


davidzhao (approved):
lg!

PR #6119, opened 2026-06-16
feat(assemblyai): add universal-3-5-pro (now default) and Voice Focus streaming params

dlange-aai:
## Summary
- **universal-3-5-pro**: new streaming model, now the **plugin default** (was `universal-streaming-english`). It joins the u3-rt-pro parameter family, so it accepts the full u3-pro parameter set (`prompt`, `agent_context`, `previous_context_n_turns`, `continuous_partials`, `interruption_delay`) and inherits the family's connect-time defaults (`continuous_partials=True`, `language_detection=True`, punctuation-based turn detection).
- **voice_focus**: new connect-time parameter (u3-rt-pro family only) — `"near-field"` for headsets/handsets/close-talking mics, `"far-field"` for conference rooms/laptop mics. Isolates the primary voice and suppresses background noise before audio reaches the model. See https://www.assemblyai.com/docs/streaming/voice-focus.
- **voice_focus_threshold**: new connect-time parameter (u3-rt-pro family only) — float in `[0.0, 1.0]` controlling how aggressively background audio is suppressed.
- Consolidated the per-parameter u3-rt-pro-family validation into a single loop so the `ValueError` lists the actual supported models (it previously named only `u3-rt-pro`, now misleading since `universal-3-5-pro` is the default).

> **Behavior change:** because the default model is now a u3-rt-pro-family model, a bare `STT()` now defaults to `continuous_partials=True`, `language_detection=True`, and punctuation-based turn detection (`min_turn_silence=100`, `max_turn_silence=min`). Callers who relied on the previous `universal-streaming-english` default should set `model="universal-streaming-english"` explicitly.

## Test Plan
- `uv run pytest --plugin assemblyai` — 51 tests (default-model assertions, u3-rt-pro-family gating + acceptance across all family models, `voice_focus`/`voice_focus_threshold` plumbing, connect-config forwarding incl. `voice_focus_threshold=0.0`, `update_options` exclusion). 5 existing gating tests were updated to pin `model="universal-streaming-english"` since the default now joins the family.
- ruff format + lint + mypy clean on touched files.

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -56,7 +56,8 @@ class STTOptions:
         "u3-rt-pro",
         "u3-rt-pro-beta-1",
         "u3-pro",
-    ] = "universal-streaming-english"
+        "universal-3-5-pro",
+    ] = "universal-3-5-pro"
     language_detection: NotGivenOr[bool] = NOT_GIVEN
     end_of_turn_confidence_threshold: NotGivenOr[float] = NOT_GIVEN
     min_turn_silence: NotGivenOr[int] = NOT_GIVEN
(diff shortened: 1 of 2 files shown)


tinalenguyen (approved):
lgtm, thank you for the PR!

PR #6120, opened 2026-06-16
fix: preserve transcription item id

### LINKED ISSUE #6109: UserInputTranscribedEvent drops item_id, making per-utterance dedup of interim transcripts impossible without provider-specific events
### Bug Description

`AgentSession`'s `user_input_transcribed` event fires once per interim transcript. For realtime models (OpenAI/xAI), the input-transcription delta handler emits a new `input_audio_transcription_completed(is_final=False)` for *every* streamed delta chunk, so a single user utterance produces many `user_input_transcribed` events with `is_final=False`.
A common need is to react exactly once per utterance — e.g. notify the frontend "user speech received" so it can render a placeholder before the agent responds. To do that you need a stable key to correlate all the interim events of the same utterance.
That key already exists internally: `livekit.agents.llm.InputTranscriptionCompleted` carries `item_id` (and `confidence`). But when `AgentActivity` re-emits it upward, the id is discarded:
```python
    )  # ev.item_id is dropped
UserInputTranscribedEvent itself only has transcript, is_final, speaker_id, language, created_at (livekit/agents/voice/events.py) — no item_id.

Consequence: to dedup per utterance, a consumer must either (a) keep manual last_item_id state that the event can't actually provide, or (b) bypass the provider-agnostic event entirely and subscribe to the raw openai_server_event_received to read item_id from the raw dict. Option (b) is not portable — Gemini's realtime plugin doesn't emit openai_server_event_received at all, so the same code can't work across providers

### Expected Behavior

UserInputTranscribedEvent should expose the item_id that is already present on llm.InputTranscriptionCompleted, so consumers can correlate all interim/final transcripts of a single utterance and act exactly once per utterance — using the provider-agnostic user_input_transcribed event, without dropping down to provider-specific raw events.

Concretely:

Add item_id: str | None to UserInputTranscribedEvent and pass ev.item_id through in _on_input_audio_transcription_completed.
(Optional) include a stable timestamp marking when this item/utterance first started, so consumers can dedup/order without tracking state themselves.
This keeps a single, provider-agnostic subscription point that works uniformly across OpenAI, xAI, and Gemini.

### Reproduction Steps

```bash
1.Start an AgentSession with any realtime model (e.g. OpenAI/xAI gpt-realtime).
2.session.on("user_input_transcribed", handler).
3.Speak one sentence.
4.Observe handler fires many times with is_final=False during the single utterance, and the event provides no item_id to group them.
```

### Operating System

macOS

### Models Used

Realtime xai realtime , openai gpt-realtime , gemini 

### Package Versions

```bash
livekit-agents 1.5.17
livekit (rtc) 1.1.8
livekit-plugins-openai 1.5.17
Python 3.13, macOS
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

nightcityblade:
Fixes #6109

## Summary
- expose `item_id` on `UserInputTranscribedEvent`
- pass through realtime input transcription item IDs when re-emitting user transcription events
- add a regression test for item ID propagation

## Testing
- `uv run pytest tests/test_agent_session.py -q`
- `uv run pytest tests/test_agent_session.py::test_realtime_user_input_transcription_preserves_item_id -q`
- `uv run ruff check livekit-agents/livekit/agents/voice/events.py livekit-agents/livekit/agents/voice/agent_activity.py tests/test_agent_session.py`

### DIFF
--- a/livekit-agents/livekit/agents/voice/events.py
+++ b/livekit-agents/livekit/agents/voice/events.py
@@ -315,6 +315,8 @@ class UserInputTranscribedEvent(BaseModel):
     type: Literal["user_input_transcribed"] = "user_input_transcribed"
     transcript: str
     is_final: bool
+    item_id: str | None = None
+    """Provider-specific ID for the transcribed input item, when available."""
     speaker_id: str | None = None
     language: LanguageCode | None = None
     created_at: float = Field(default_factory=time.time)
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1728,7 +1728,11 @@ class AgentActivity(RecognitionHooks):
 
     def _on_input_audio_transcription_completed(self, ev: llm.InputTranscriptionCompleted) -> None:
         self._session._user_input_transcribed(
-            UserInputTranscribedEvent(transcript=ev.transcript, is_final=ev.is_final)
+            UserInputTranscribedEvent(
+                transcript=ev.transcript,
+                is_final=ev.is_final,
+                item_id=ev.item_id,
+            )
         )
 
         if ev.is_final:
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -26,9 +26,7 @@ from livekit.agents import (
     inference,
     vad,
 )
-from livekit.agents.llm import (
-    FunctionToolCall,
-)
+from livekit.agents.llm import FunctionToolCall, InputTranscriptionCompleted
 from livekit.agents.llm.chat_context import ChatContext, ChatMessage
 from livekit.agents.stt import SpeechData, SpeechEvent, SpeechEventType
 from livekit.agents.utils import aio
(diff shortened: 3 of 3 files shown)


areebkhan-tech:
it seems to have merge conflicts

nightcityblade:
Thanks for flagging it — I'm working through a rebase now. The branch drift turned out to be larger than expected, so I want to untangle it cleanly before I push an update.

nightcityblade:
Updated the branch to resolve the merge conflicts by merging the latest `livekit/agents:main` into `fix/issue-6109` (no force-push).

Validation run:
- `uv run pytest tests/test_agent_session.py::test_realtime_user_input_transcription_preserves_item_id` ✅

The PR now reports as mergeable on GitHub. I also noticed the CLA check is still pending/not signed, which appears to be the remaining non-code item.

areebkhan-tech:
Hi, can u pls request re-review this has not been merged yet, or could it be you have not signed CLA yet .

Monkiia:
I checked the branch again. The code change itself is already rebased and the targeted test passed in the last update; the remaining blocker appears to be the CLA status rather than an outstanding code review issue.

Once the CLA is recognized on GitHub, I'm happy to request re-review immediately.

davidzhao:
@nightcityblade could you please sign the CLA?

nightcityblade:
I’m checking the CLA flow from this account now. The branch itself is already up to date and the targeted regression test passed in the last code update; once GitHub recognizes the CLA, I’ll request re-review immediately.

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -315,6 +315,8 @@ class UserInputTranscribedEvent(BaseModel):
     type: Literal["user_input_transcribed"] = "user_input_transcribed"
     transcript: str
     is_final: bool
+    item_id: str | None = None
```
theomonnom, line 318: Pretty sure we will need to expose even more fields in the future, extra seems more appropriate
```suggestion
    extra: dict = {}
```

We can then do something like extra["openai"]["item_id"] ?

davidzhao (approved):

PR #6121, opened 2026-06-16
fix(worker): fix timeout option overrides and update initialize_process_timeout

parkky21:
This PR fixes `update_options` to use `NotGivenOr[float]` defaults for timeouts, ensuring they are not reset to default values if not explicitly specified. It also adds the missing assignment for `initialize_process_timeout` in the `update_options` function body.


longcw (approved):

longcw:
thanks for the pr!

PR #6124, opened 2026-06-16
fix(workflows): normalize two-digit years in GetDOBTask

### LINKED ISSUE #6067: GetDOBTask silently accepts two-digit years as ancient dates (e.g. year 90 AD)
### Bug Description

The prompt instructions in `beta/workflows/dob.py` tell the model to normalize two-digit years ("Handle two-digit years appropriately: '90' likely means 1990"), but the `update_dob` tool accepts `year` as a raw int with no lower bound. If the model passes the spoken value through literally — which smaller/faster models commonly do — `date(90, 5, 15)` is a valid Python date (year 90 AD), the future-date check passes, no `ToolError` is raised, and the task completes with a corrupted birthdate.

The existing validation catches impossible dates (e.g. Feb 30) and future dates, but not implausible ones. The two-digit-year normalization currently exists only in the prompt; the deterministic tool layer does not enforce it, so the one failure mode the instructions themselves anticipate is exactly the one that slips through. Since this workflow is likely to feed identity, healthcare, or fintech intake systems, the failure is silent data corruption rather than a visible error.

### Expected Behavior

Either the two-digit year is deterministically normalized to the intended century (matching what the prompt already promises), or `update_dob` raises a `ToolError` for implausible years so the agent re-elicits the date. In no case should `GetDOBResult` complete successfully with a year like 90 AD.

### Reproduction Steps

```bash
The tool-layer gap is reproducible without a live session:


from datetime import date

d = date(90, 5, 15)        # valid Python date: year 90 AD
print(d <= date.today())   # True — the future-date check in _update_dob_impl passes


End-to-end: run `GetDOBTask` and have the model call `update_dob(year=90, month=5, day=15)` (e.g. user says "May fifteenth, ninety" and the model passes the literal value). `_update_dob_impl` constructs the date, the future check passes, and the task completes with `GetDOBResult(date_of_birth=datetime.date(90, 5, 15))`.
```

### Operating System

Linux (not OS-dependent — pure validation logic)

### Models Used

Model-independent: any LLM that passes a spoken two-digit year through literally triggers it. Smaller/faster models do this most often.

### Package Versions

```bash
livekit-agents @ current main (beta/workflows/dob.py)
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
Two options, happy to submit a PR for whichever direction maintainers prefer:

1. **Pivot-window normalization** in `_update_dob_impl`: deterministically map `year < 100` with a sliding window (e.g. `year + 2000 if year <= current_year % 100 else year + 1900`), matching the behavior the prompt already promises.

2. **Hard floor**: raise `ToolError` for `year < 1900` (or a configurable minimum), forcing the model to re-elicit instead of guessing.

Option 1 silently matches user intent; option 2 is stricter and keeps all normalization observable. Either is better than accepting year 90. A test for the two-digit-year case would accompany the fix.
```

### Additional Context

_No response_

_No response_

he-yufeng:
Closes #6067

`GetDOBTask`'s prompt tells the model to normalize two-digit years ("90" likely means 1990), but `_update_dob_impl` takes `year` as a raw int with no lower bound. Smaller/faster models often pass the spoken value through literally, and `date(90, 5, 15)` is a valid Python date (year 90 AD): the future-date check passes, no `ToolError` is raised, and the task completes with a corrupted birthdate. Because this workflow tends to feed identity/healthcare/fintech intake, it's silent data corruption rather than a visible error.

Fix: normalize `year < 100` at the top of `_update_dob_impl` with a pivot window keyed on the current year, which is exactly what the prompt already promises. 90 -> 1990, 05 -> 2005, 26 -> 2026, 27 -> 1927. Four-digit years are left untouched, so there's no regression. The issue floated a hard floor (raise on `year < 1900`) as an alternative; I went with normalization since it matches the documented prompt contract and the smaller behavioral change.

Verified:
- Added `tests/test_dob.py` (unit, no LLM). It fails on `main` (`date(90, 5, 15) != date(1990, 5, 15)`) and passes with the fix.
- `pytest tests/test_dob.py --unit` -> 2 passed.
- `ruff check` and `ruff format --check` clean on both files.

### DIFF
--- a/livekit-agents/livekit/agents/beta/workflows/dob.py
+++ b/livekit-agents/livekit/agents/beta/workflows/dob.py
@@ -151,6 +151,14 @@ class GetDOBTask(AgentTask[GetDOBResult]):
     async def _update_dob_impl(
         self, year: int, month: int, day: int, ctx: RunContext
     ) -> str | None:
+        # Normalize two-digit years to the intended century, matching what the
+        # prompt already asks the model to do ("90" -> 1990, "05" -> 2005). A
+        # literal two-digit year is otherwise a valid date (e.g. 90 -> year 90 AD)
+        # that passes the future-date check and silently corrupts the result.
+        if 0 <= year < 100:
+            current_yy = date.today().year % 100
+            year += 2000 if year <= current_yy else 1900
+
         try:
             dob = date(year, month, day)
         except ValueError as e:
--- a/tests/test_dob.py
+++ b/tests/test_dob.py
@@ -1,0 +1,32 @@
+from datetime import date
+
+import pytest
+
+from livekit.agents import beta
+
+pytestmark = pytest.mark.unit
+
+
+@pytest.mark.asyncio
+async def test_dob_two_digit_year_normalized() -> None:
+    # The prompt asks the model to normalize two-digit years, but smaller/faster
+    # models often pass the spoken value through literally. The tool layer must not
+    # accept "90" as year 90 AD. https://github.com/livekit/agents/issues/6067
+    task = beta.workflows.GetDOBTask(require_confirmation=True)
+
+    await task._update_dob_impl(90, 5, 15, ctx=None)  # type: ignore[arg-type]
+    assert task._current_dob == date(1990, 5, 15)
+
+    await task._update_dob_impl(5, 3, 1, ctx=None)  # type: ignore[arg-type]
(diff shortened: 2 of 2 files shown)


tinalenguyen (approved):
thank you for the PR!

PR #6126, opened 2026-06-16
feat(soniox): support stt-rt-v5 with endpoint_sensitivity option

mihafabcic-soniox:
Updates the LiveKit Soniox plugin for the v5 model.

- Add `endpoint_sensitivity` to `STTOptions` (`float | None`, range `-1.0` to `1.0`). Controls how quickly the model commits endpoints. Higher values finalize sooner. Only supported by v5; earlier models reject the field. Skipped on the wire when `None` so the server uses its default.
- Default STT model is now `stt-rt-v5`.
- Default `max_endpoint_delay_ms` raised from `500` (the API minimum) to `2000`. The old default was too aggressive on phone-call audio: short pauses between word or digit groups would cause Soniox to finalize a segment too early, before the model had enough context. `2000` matches the Soniox API's own default.

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
@@ -124,12 +124,20 @@ class STTOptions:
     Range: 500–3000.
     See: https://soniox.com/docs/stt/rt/endpoint-detection"""
 
+    endpoint_sensitivity: float | None = None
+    """How readily the model emits speech endpoints. Range: -1.0 to 1.0.
+    Higher values make endpoints more likely (finalize sooner); lower values make them
+    less likely. Leave as None to use the server-side default.
+    Introduced in the Soniox v5 model; earlier models reject it."""
+
     client_reference_id: str | None = None
     translation: TranslationConfig | None = None
 
     def __post_init__(self) -> None:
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
```diff
@@ -121,17 +119,25 @@ class STTOptions:
     enable_speaker_diarization: bool = False
     enable_language_identification: bool = True
 
-    max_endpoint_delay_ms: int = 500
+    max_endpoint_delay_ms: int = 2000
```
tinalenguyen, line 122: @mihafabcic-soniox could you provide more context on why you changed the default here?

tinalenguyen (commented):
lgtm, left a small comment

livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/stt.py
mihafabcic-soniox: Some context is in the PR description. The change came from a real customer who hit transcription issues from endpoints firing too aggressively at 500ms (e.g. extra digits added to phone numbers). Bumping to 2000ms (the Soniox API's own default) resolved them.

tinalenguyen (approved):
thank you for the PR!

PR #6132, opened 2026-06-17
Add Hopper LLM provider

pavanyellow:
## Summary

Adds a `LLM.with_hopper()` factory to the OpenAI plugin, alongside the other
OpenAI-compatible providers (Cerebras, Together, Nebius, Telnyx, …).

[Hopper](https://withhopper.com) serves open-source models optimized for low
time-to-first-token, aimed at voice agents. The API is OpenAI-compatible, so
this follows the existing `with_*` provider pattern — no new plugin, just a
factory method plus a `HopperChatModels` type.

## Usage

```python

llm = openai.LLM.with_hopper(
    # api_key defaults to HOPPER_API_KEY env var
    model="Qwen/Qwen3.6-35B-A3B",
)
```

Get an API key at https://withhopper.com.

## Changes

- `with_hopper()` classmethod on `openai.LLM` (defaults: `base_url=https://api.withhopper.com/v1`, `HOPPER_API_KEY`).
- `HopperChatModels` literal in `models.py`.
- `_strict_tool_schema=False`, matching the other open-model providers served on vLLM-style backends (same as Cerebras, #3134).

## Latency

TTFT (time to first token) over a warm connection, voice-agent-shaped context
(~2k-token system prompt + short user turn), 10-run p50:

- From **us-west-2** (same region as the model server): **p50 62ms** (min 56, max 77)
- From a residential laptop in SF: ~170ms — the difference is network round-trip

First-token latency is dominated by where your agent runs relative to the model,
not the serving itself; colocated, it's ~60ms.

## Testing

Verified against the live endpoint through the plugin:

- factory builds with the correct base URL and model
- streaming `chat()` returns a valid completion
- function calling emits a correct tool call (validates `_strict_tool_schema=False`)
- `ruff check` passes on the changed files

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -44,6 +44,7 @@ from .models import (
     ChatModels,
     CometAPIChatModels,
     DeepSeekChatModels,
+    HopperChatModels,
     NebiusChatModels,
     OctoChatModels,
     OpenRouterProviderPreferences,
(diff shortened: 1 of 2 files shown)


tinalenguyen (commented):
hi, thank you for the PR! i tried it and got this error:
```shell
livekit.agents._exceptions.APIStatusError: message="Error code: 400 - {'error': {'message': 'No user query found in messages.', 'type': 'BadRequestError', 'param': None, 'code': 400}}", status_code=400, retryable=False, body={'message': 'No user query found in messages.', 'type': 'BadRequestError', 'param': None, 'code': 400}
```
i tried this with our survey example and basic_agent.py example

PR #6131, opened 2026-06-16
feat(openai): add LLM.with_cloudflare() for Cloudflare AI Gateway

mcdgavin:
## Summary

Closes #6101. Adds `openai.LLM.with_cloudflare(...)`, a convenience constructor that routes
LLM traffic through the [Cloudflare AI Gateway](https://developers.cloudflare.com/ai-gateway/)
OpenAI-compatible [REST API](https://developers.cloudflare.com/ai-gateway/usage/rest-api/)
endpoint — mirroring the existing `with_openrouter` / `with_azure` helpers.

```python
from livekit.plugins import openai

# Minimal: account_id from CLOUDFLARE_ACCOUNT_ID, API token from CLOUDFLARE_API_TOKEN
llm = openai.LLM.with_cloudflare(model="openai/gpt-4o")

# Explicit, with a specific gateway and per-request gateway options
llm = openai.LLM.with_cloudflare(
    model="@cf/meta/llama-3.3-70b-instruct-fp8-fast",
    account_id="<account_id>",
    gateway_id="my-gateway",
    api_key="<cloudflare_api_token>",
    gateway_options={"cache_ttl": 3600, "max_attempts": 3, "backoff": "exponential"},
)
```

## Design notes

- **REST API endpoint.** Uses
  `https://api.cloudflare.com/client/v4/accounts/<account_id>/ai/v1` per review feedback — the
  legacy `gateway.ai.cloudflare.com/.../compat` unified endpoint is deprecated for single-model
  calls. The URL is built from `account_id` (env `CLOUDFLARE_ACCOUNT_ID`), with a `base_url`
  override.
- **Auth.** A Cloudflare API token (**Account > Workers AI > Read** permission) is sent as the
  standard `Authorization: Bearer` header via the OpenAI client's `api_key`; env fallback
  `CLOUDFLARE_API_TOKEN`. The legacy endpoint's `cf-aig-authorization` dual-auth scheme is not
  used.
- **Gateway selection.** Via the `cf-aig-gateway-id` header (`gateway_id` param). When omitted,
  third-party requests route through the account's default gateway; Workers AI (`@cf/...`)
  models require it.
- **Models.** Third-party providers use `provider/model` form (e.g. `openai/gpt-4o`); Workers AI
  models use `@cf/author/model` form.
- **Per-request options are headers, not body.** Unlike OpenRouter (JSON body fields),
  Cloudflare's controls are `cf-aig-*` HTTP headers. `gateway_options` is a
[...]

Issue #3163, opened 2025-08-15
Request: Cloudflare TTS, LLM, STT plugins


leemuljadi:
Hi would like to request plugins for Cloudflare TTS, LLM, STT

davidzhao:
please consider contributing one! it's pretty straight forward to build plugins

beingPro007:
@davidzhao, currently, Cloudflare won't provide STT over streaming; instead, they are doing batch processing. Is there any way to do it, then suggest to me that I am going to build it?.

davidzhao:
sure thing, please go for it. [OpenAI STT](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py) is also non-streaming, you can use it as a reference implementation

leemuljadi:
@davidzhao cool, how about for the Cloudflare meloTTS as it's using batch process as well? any example implementation of batch processing mp3?
Will the Cloudflare stack be a concern for latency since they are not providing streaming for STT and TTS?
And will it be possible to deploy Livekit agents in Clouldflare?

leemuljadi:
@beingPro007 thanks for collaborating!

mcdgavin:
Opened #6133 covering the **STT and TTS** parts of this request, and the **LLM** part shipped separately in #6131. Could a maintainer reopen this to track them? 🙏

**What's changed since this issue was filed:**

- **Streaming is now available.** Earlier discussion here assumed Cloudflare was batch-only request/response. That's no longer true: Cloudflare now hosts Deepgram partner models — **Nova-3** / **Flux** (STT) and **Aura** (TTS) — exposed over a real-time **WebSocket through the AI Gateway**, allowing for true streaming plugins rather than batch-wrapped.
- **The wire protocol is Deepgram's.** The gateway proxies Deepgram's streaming protocol verbatim, so #6133 adds `deepgram.STT.with_cloudflare(...)` / `deepgram.TTS.with_cloudflare(...)` that reuse the existing Deepgram streaming implementation — auth via the `cf-aig-authorization` header, no Deepgram key required.

```python
from livekit.plugins import deepgram, openai

stt = deepgram.STT.with_cloudflare(model="@cf/deepgram/nova-3")
tts = deepgram.TTS.with_cloudflare(model="@cf/deepgram/aura-1")
llm = openai.LLM.with_cloudflare(model="openai/gpt-4o")  # #6131
```

PR #6133, opened 2026-06-17
feat(deepgram): add STT/TTS with_cloudflare() for Cloudflare AI Gateway

mcdgavin:
## Summary

Adds Cloudflare AI Gateway support to the Deepgram plugin — streaming **STT** (Nova-3 via `STT`,
Flux via `STTv2`) and **TTS** (Aura) — via `with_cloudflare(...)` constructors.

Addresses the STT/TTS portion of #3163. (The LLM portion shipped separately in #6131 as
`openai.LLM.with_cloudflare`.)

```python

# account_id from CLOUDFLARE_ACCOUNT_ID, token from CLOUDFLARE_AI_GATEWAY_TOKEN
stt = deepgram.STT.with_cloudflare(model="@cf/deepgram/nova-3")
flux = deepgram.STTv2.with_cloudflare()  # @cf/deepgram/flux
tts = deepgram.TTS.with_cloudflare(model="@cf/deepgram/aura-1")
```

## Design notes

- **Rides on the Deepgram plugin.** Cloudflare's AI Gateway proxies Deepgram's streaming
  WebSocket protocols verbatim (`channel.alternatives[0].transcript` + `is_final` for v1 STT;
  Flux `TurnInfo` events for v2; `Speak`/`Flush`/`Flushed` for TTS), so these reuse the existing
  Deepgram `STT`/`STTv2`/`TTS` streaming implementations rather than duplicating them. This
  mirrors the LLM approach, where Cloudflare-LLM rode on the OpenAI plugin because it's
  OpenAI-compatible — each Cloudflare modality lives on the plugin whose protocol it matches.
- **Backward-compatible auth seam.** The connection auth headers (previously hardcoded
  `Authorization: Token <key>`) become an overridable `_connect_headers`, defaulting to the
  exact prior value. Existing Deepgram usage is unchanged; `with_cloudflare` injects
  `cf-aig-authorization` (the raw Cloudflare token — no Deepgram key required).
- **Flux via `STTv2`.** Cloudflare's catalog has a single Flux model (`@cf/deepgram/flux`, no
  `flux-general-en`/`-multi` variants), so `STTv2.with_cloudflare()` defaults to `model="flux"`
  and forwards the Flux tuning options (`eot_threshold`, `eager_eot_threshold`,
  `eot_timeout_ms`, `keyterm`). The gateway URL/token resolution shared by the three
  constructors lives in `_utils._resolve_cloudflare_gateway`.

## Testing

- `tests/test_deepgram_with_cloudflare.py` — 35 hermetic `--unit` tests covering URL building,
  `CLOUDFLARE_ACCOUNT_ID` env fallback, `base_url` override, the `cf-aig-authorization` header,
  default models, `streaming` capability, the `ValueError` paths, and prefix-aware model
  validation (`@cf/deepgram/...` seen as the underlying Deepgram model). Non-streaming calls
  (`STT.recognize()`, `TTS.synthesize()`) are not proxied by the gateway and now raise
  `NotImplementedError` up front instead of failing with a connection error after retries.
- A regression check confirms default (non-Cloudflare) Deepgram STT/STTv2/TTS auth is
  byte-identical.
- `make check` passes (ruff format + lint, strict mypy).


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -203,6 +209,64 @@ def model(self) -> str:
     def provider(self) -> str:
         return "Deepgram"
 
+    @staticmethod
+    def with_cloudflare(
+        *,
+        model: str = "@cf/deepgram/nova-3",
```
chenghao-mou: is it possible to construct "@cf/deepgram" part internally? Then we can reuse the same `DeepgramModels | str` type

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -102,6 +102,7 @@ def __init__(
         numerals: bool = False,
         mip_opt_out: bool = False,
         vad_events: bool = True,
+        _connect_headers: NotGivenOr[dict[str, str]] = NOT_GIVEN,
```
chenghao-mou: We can just call it `extra_headers: NotGivenOr[dict[str, str]] = NOT_GIVEN`. We use similar parameters in other plugins as well, it could also be useful for self-hosted DG instances too.

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -203,6 +209,64 @@ def model(self) -> str:
...
+            base_url = f"https://gateway.ai.cloudflare.com/v1/{account_id}/{gateway_id}/workers-ai"
+
+        return STT(
+            model=model,
+            language=language,
+            interim_results=interim_results,
+            sample_rate=sample_rate,
+            base_url=base_url,
+            http_session=http_session,
+            _connect_headers={"cf-aig-authorization": cf_aig_token},
```
chenghao-mou: double checking here: the code in their doc says `"cf-aig-authorization": "Bearer {CF_AIG_TOKEN}"`: https://developers.cloudflare.com/ai-gateway/usage/providers/deepgram/

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
mcdgavin: Thank you for the feedback @chenghao-mou, this is done.

 `with_cloudflare` now takes a bare Deepgram model name typed `DeepgramModels | str` (default `nova-3` for STT, `aura-1` for TTS) and prepends `@cf/deepgram/` internally. 
A value already prefixed with `@cf/` is passed through unchanged, so the full form still works.

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
mcdgavin: Switched `_connect_headers` to a public `extra_headers`, additive, matching the other plugins and supporting self hosting.

One note, a purely additive `extra_headers` merges on top of the default `Authorization: Token <key>`, so the Cloudflare path would send *both* `Authorization` and `cf-aig-authorization`  which requires a Deepgram key it doesn't need and risks the gateway rejecting the stray header. To keep it additive without that, the default `Authorization: Token` is now built only when a Deepgram key is present, and at least one auth source (key or `extra_headers`) is required:

```python
headers = ({"Authorization": f"Token {key}"} if key else {}) | extra_headers
```

Behavior:
- Normal Deepgram (key set): `Authorization: Token` + extras — unchanged.
- Self-hosted DG with custom auth (no key): just `extra_headers`.
- Cloudflare via `with_cloudflare` (no key): just `cf-aig-authorization`.

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
mcdgavin: I double-checked against the [realtime WebSockets page](https://developers.cloudflare.com/ai-gateway/usage/websockets-api/realtime-api/). The `Bearer` form is documented on the `/deepgram` **provider-proxy** endpoint, but this plugin targets the `/workers-ai` endpoint, and its **Deepgram (Workers AI)** examples (both `@cf/deepgram/nova-3` and `@cf/deepgram/aura-1`) use the **raw** `cf-aig-authorization` token:

```js
"wss://gateway.ai.cloudflare.com/v1/<account_id>/<gateway>/workers-ai?model=@cf/deepgram/nova-3..."
headers: { "cf-aig-authorization": process.env.CLOUDFLARE_API_KEY }
```

 Since it flows through `extra_headers`, it's still overridable either way.

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -203,6 +213,67 @@ def model(self) -> str:
...
+            account_id = account_id or os.environ.get("CLOUDFLARE_ACCOUNT_ID")
+            if not account_id:
+                raise ValueError(
+                    "Cloudflare account_id is required, either as argument or set"
+                    " CLOUDFLARE_ACCOUNT_ID environment variable (or pass base_url directly)"
+                )
+            base_url = f"https://gateway.ai.cloudflare.com/v1/{account_id}/{gateway_id}/workers-ai"
+
+        if not model.startswith("@cf/"):
+            model = f"@cf/deepgram/{model}"
```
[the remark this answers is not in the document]
mcdgavin: Fixed in acf23e3db. `_validate_model` and `_validate_keyterm` now strip the `@cf/deepgram/` routing prefix before the model-name checks (via a small `_bare_model` helper), so `@cf/deepgram/nova-3` is correctly treated as `nova-3` — keyterm is allowed and keywords are rejected as expected. Added a test (`test_keyterm_allowed_for_cf_prefixed_nova3`).

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -203,6 +213,67 @@ def model(self) -> str:
...
+
+        return STT(
+            model=model,
+            language=language,
+            interim_results=interim_results,
+            sample_rate=sample_rate,
+            base_url=base_url,
+            http_session=http_session,
+            extra_headers={"cf-aig-authorization": cf_aig_token},
+        )
```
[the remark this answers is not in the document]
mcdgavin: Good catch — confirmed and fixed in acf23e3db. `with_cloudflare` now passes `api_key=""`, which opts out of the `DEEPGRAM_API_KEY` env fallback, so the gateway only ever receives `cf-aig-authorization` (no stray `Authorization: Token`). TTS only reads the env var when `api_key is None`, so an explicit `""` means "no key". Added a regression test that sets `DEEPGRAM_API_KEY` and asserts the header is absent (`test_with_cloudflare_ignores_deepgram_api_key_env`).

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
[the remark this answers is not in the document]
mcdgavin: Fixed in 7014597db. `_validate_model` now preserves the routing prefix on the fallback:

```python
bare = _bare_model(model)
...
prefix = model[: len(model) - len(bare)]
return f"{prefix}nova-2-general"
```

so `@cf/deepgram/nova-2-meeting` + a non-English language → `@cf/deepgram/nova-2-general` (bare models unchanged). Added `test_en_only_fallback_preserves_cf_prefix`.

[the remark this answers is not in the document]
mcdgavin: Intentional. `STTv2` is the Deepgram v2 / Flux class (different protocol, `/v2/listen`). This PR scopes Cloudflare support to the stable `STT` (nova-3) and `TTS` (Aura).

A `STTv2.with_cloudflare` for Flux is a possible follow-up rather than part of this change.

PR #6137, opened 2026-06-17
feat(core): allow non-primary sessions without job_ctx

Bobronium:
Add private optional argument `_register_as_room_host` for `AgentSession.start()`, behavior stays the same when it's not used.

This allows using `AgentSession` as a non-primary session, which is required in simulations.

We can promote it to public interface when (if) it'll find any other use-cases.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -613,6 +613,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         # deprecated
         room_input_options: NotGivenOr[room_io.RoomInputOptions] = NOT_GIVEN,
         room_output_options: NotGivenOr[room_io.RoomOutputOptions] = NOT_GIVEN,
+        _register_as_room_host: NotGivenOr[bool] = NOT_GIVEN,
     ) -> RunResult: ...
 
     @overload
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -781,7 +777,7 @@ async def start(
                 self._room_io = room_io.RoomIO(room=room, agent_session=self, options=room_options)
                 await self._room_io.start()
 
-                if is_primary:
+                if _register_as_room_host or not is_given(_register_as_room_host):
```
chenghao-mou: nitpicking: `_register_as_room_host is not False`

chenghao-mou (approved):

livekit-agents/livekit/agents/voice/agent_session.py
Bobronium:
```suggestion
                if _register_as_room_host is not False:
```

tests/test_room_host.py
```diff
@@ -0,0 +1,135 @@
...
+# fmt: off
+# @formatter:off
+_CASES = [
+    # "_register_as_room_host", "job_ctx primary marker" and  "expected outcome")
+    pytest.param(True,      None,     Outcome.REGISTERS,     id="no_ctx/opt-in"),
+    pytest.param(True,      _OTHER,   Outcome.REGISTERS,     id="non_primary/force"),
+    pytest.param(NOT_GIVEN, None,     Outcome.REGISTERS,     id="no_ctx/default"),
+    pytest.param(NOT_GIVEN, _PRIMARY, Outcome.REGISTERS,     id="primary/default"),
+    pytest.param(False,     _PRIMARY, Outcome.NOT_REGISTERS, id="primary/opt-out"),
+    pytest.param(NOT_GIVEN, _OTHER,   Outcome.NOT_REGISTERS, id="non_primary/default"),
```
Bobronium:
```suggestion
    pytest.param(NOT_GIVEN, _OTHER,   Outcome.NOT_REGISTERS, id="non_primary/default"),
    pytest.param(False,     _PRIMARY, Outcome.NOT_REGISTERS, id="primary/opt-out"),
```

PR #6158, opened 2026-06-18
fix(openai): skip realtime truncate when no audio was played fixes : #6157

### LINKED ISSUE #6157: RealtimeModel: "Only model output audio messages can be truncated" when interrupted before first audio frame
### Bug Description

When using the OpenAI Realtime model, if the user interrupts (barge-in) in the
narrow window *after* the model has declared an audio response but *before* the
first audio frame is actually played, the plugin emits a
`conversation.item.truncate` with `audio_end_ms=0`. The Realtime API rejects it:
APIError('OpenAI Realtime API returned an error',
body=RealtimeError(
message='Only model output audio messages can be truncated',
type='invalid_request_error',
code='unsupported_content_type'),
retryable=True) recoverable=True

Root cause
`AgentActivity` calls `truncate()` on interruption with
`audio_end_ms = int(entry.out.playback_position * 1000)`, which is `0` when no
frame has played yet
(`livekit-agents/livekit/agents/voice/agent_activity.py:3609`).
`RealtimeSession.truncate()` then unconditionally sends a
`ConversationItemTruncateEvent` whenever `"audio"` is in `modalities`
(`livekit-plugins/livekit-plugins-openai/.../realtime/realtime_model.py:1609`).
Because the item has no committed model-output audio, the server rejects it.

Event sequence
1. `response.created`
2. `response.output_item.added` — message_id assigned
3. `response.content_part.added` — modalities resolve to `["audio", "text"]`
4. *(user interrupts here — VAD fires)*  ← `response.audio.delta` has NOT happened yet
5. interruption path calls `truncate(..., audio_end_ms=0)` → server error

### Expected Behavior

Interrupting before any audio has played should be a no-op for truncation —
there is no committed audio to truncate, so the plugin should not send a
`conversation.item.truncate` with `audio_end_ms=0`. No error should be raised
and the session should continue cleanly.

### Reproduction Steps

```bash
1. Start an AgentSession with `openai.realtime.RealtimeModel` (audio modality).
2. Trigger an initial agent reply (e.g. a welcome message).
3. Speak / send input that interrupts within the first few hundred ms,
   before the first audio frame plays.
4. Observe the error in logs.
```

### Operating System

macOS

### Models Used

_No response_

### Package Versions

```bash
livekit-agents == 1.5.4
livekit-plugins-openai == 1.5.4
OpenAI Realtime model (e.g. gpt-4o-realtime / gpt-realtime), server VAD turn detection
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

ByteMaster-1:
Fixes #6157

Skip the realtime `conversation.item.truncate` when the generation is
interrupted before any audio frame has played (`audio_end_ms == 0`), which
the Realtime API rejects with `unsupported_content_type`
("Only model output audio messages can be truncated").

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1625,15 +1625,13 @@ class RealtimeSession(
                     )
                 )
             else:
-                remote_ids = {item.id for item in self._remote_chat_ctx.to_chat_ctx().items}
-                if message_id in remote_ids:
-                    self.send_event(
-                        ConversationItemDeleteEvent(
-                            type="conversation.item.delete",
-                            item_id=message_id,
-                            event_id=utils.shortuuid("chat_ctx_delete_"),
-                        )
+                self.send_event(
+                    ConversationItemDeleteEvent(
+                        type="conversation.item.delete",
+                        item_id=message_id,
+                        event_id=utils.shortuuid("chat_ctx_delete_"),
                     )
+                )
         elif utils.is_given(audio_transcript):
             # sync the forwarded text to the remote chat ctx
             chat_ctx = self.chat_ctx.copy(
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1624,6 +1624,16 @@ class RealtimeSession(
                         audio_end_ms=audio_end_ms,
                     )
                 )
+            else:
+                remote_ids = {item.id for item in self._remote_chat_ctx.to_chat_ctx().items}
+                if message_id in remote_ids:
+                    self.send_event(
+                        ConversationItemDeleteEvent(
+                            type="conversation.item.delete",
+                            item_id=message_id,
+                            event_id=utils.shortuuid("chat_ctx_delete_"),
+                        )
+                    )
         elif utils.is_given(audio_transcript):
             # sync the forwarded text to the remote chat ctx
             chat_ctx = self.chat_ctx.copy(
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1615,14 +1615,15 @@ class RealtimeSession(
         audio_transcript: NotGivenOr[str] = NOT_GIVEN,
     ) -> None:
         if "audio" in modalities:
-            self.send_event(
-                ConversationItemTruncateEvent(
-                    type="conversation.item.truncate",
-                    content_index=0,
-                    item_id=message_id,
-                    audio_end_ms=audio_end_ms,
+            if audio_end_ms > 0:
+                self.send_event(
+                    ConversationItemTruncateEvent(
+                        type="conversation.item.truncate",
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1615,14 +1615,15 @@ def truncate(
         audio_transcript: NotGivenOr[str] = NOT_GIVEN,
     ) -> None:
         if "audio" in modalities:
-            self.send_event(
-                ConversationItemTruncateEvent(
-                    type="conversation.item.truncate",
-                    content_index=0,
-                    item_id=message_id,
-                    audio_end_ms=audio_end_ms,
+            if audio_end_ms > 0:
```
longcw, line 1618: could there be two situations with `audio_end_ms == 0`:
1. the generation was interrupted/cancelled before any audio generated on the server side
2. there are audio generated but agent didn't play any of frames 

this fix works for 1 but leaks the not played message for 2? maybe we should delete this item if the audio end ms is 0

areebkhan-tech: Good catch — you're right that audio_end_ms == 0 conflates two cases. Updated to delete the item (case 2) so the unplayed message isn't left dangling in the remote chat ctx, instead of just skipping. One thing I found while doing this: _handle_error only suppresses "Cancellation failed", so an unconditional delete would itself surface an error in case 1 (item never committed server-side). So I guarded the delete on the item being present in _remote_chat_ctx. Added tests for all three branches.

One open question: there's a small race where the item exists server-side but the conversation.item.added hasn't reflected into _remote_chat_ctx yet — in that window we'd skip the delete and the leak could persist. Happy to handle that differently (e.g. always delete and suppress the not-found error) if you'd prefer.

areebkhan-tech (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -1615,14 +1615,25 @@ def truncate(
...
+                        type="conversation.item.truncate",
+                        content_index=0,
+                        item_id=message_id,
+                        audio_end_ms=audio_end_ms,
+                    )
                 )
-            )
+            else:
+                remote_ids = {item.id for item in self._remote_chat_ctx.to_chat_ctx().items}
+                if message_id in remote_ids:
```
longcw: nit: the `RealtimeModelError` won't cause a reconnect or break so I think it's fine to remove the guard unless we see too many such error logs.

longcw (approved):

ByteMaster-1:
@longcw ,please can we now merge this or u want me to remove the guard first, mentioned in your comment please let me know.

davidzhao:
> @longcw ,please can we now merge this or u want me to remove the guard first, mentioned in your comment please let me know.

please address the comment first, it'll reduce the back & forth

ByteMaster-1:
> > @longcw ,please can we now merge this or u want me to remove the guard first, mentioned in your comment please let me know.
> 
> please address the comment first, it'll reduce the back & forth

Addressed — dropped the guard, deleting unconditionally now. Pushed. @longcw @davidzhao ready for another look.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
ByteMaster-1: Addressed , dropped the guard.

PR #6172, opened 2026-06-20
fix(voice): expose item_id on UserInputTranscribedEvent

### LINKED ISSUE #6110: Expose item_id (and speech-start timestamp) on UserInputTranscribedEvent for per-utterance dedup
### Feature Type

### Feature Description

`AgentSession`'s `user_input_transcribed` event fires once per interim transcript. With realtime models (OpenAI/xAI), the input-transcription delta handler emits a new `input_audio_transcription_completed(is_final=False)` for every streamed delta chunk, so a single user utterance produces many `user_input_transcribed` events with `is_final=False`.

A common need is to react exactly once per utterance — e.g. send an RPC to the frontend so it can show a "user is being transcribed" placeholder before the agent responds. To do that, a consumer needs a stable key to correlate all interim events belonging to the same utterance.

That key already exists internally: `livekit.agents.llm.InputTranscriptionCompleted` carries `item_id`. But it is discarded when the framework re-emits the event upward:

        )  # ev.item_id is dropped

`UserInputTranscribedEvent` (livekit/agents/voice/events.py) only exposes `transcript`, `is_final`, `speaker_id`, `language`, `created_at` — no `item_id`.

Request:
- Add `item_id: str | None` to `UserInputTranscribedEvent` and forward `ev.item_id` in `_on_input_audio_transcription_completed`.
- Optionally, include a stable timestamp marking when the utterance/item first started, so consumers can dedup/order without holding their own state.

Why: this lets a single, provider-agnostic `user_input_transcribed` subscription dedup to one action per utterance, instead of forcing consumers to either track `last_item_id` manually or drop down to a provider-specific raw event.

Two workarounds, both unsatisfying:

1. Manual state: keep a `last_item_id` in the handler closure and skip duplicates. But `UserInputTranscribedEvent` has no `item_id`, so this isn't actually possible from the high-level event — the only stable correlation key is unavailable.

2. Bypass the abstraction: subscribe to the OpenAI plugin's raw `openai_server_event_received`, parse the raw dict, and read `item_id` to dedup:

       session.current_agent.realtime_llm_session.on("openai_server_event_received")(handler)
       # handler filters event["type"] == "conversation.item.input_audio_transcription.delta"
       # and dedups on event["item_id"]

   This works for OpenAI/xAI but is NOT portable — the Gemini realtime plugin (livekit-plugins-google) never emits `openai_server_event_received`, so the same code silently does nothing on Gemini.

Subscribing to the provider-agnostic `input_audio_transcription_completed` on the realtime session works across providers, but still can't dedup because the re-emitted `UserInputTranscribedEvent` lacks `item_id` (the underlying `InputTranscriptionCompleted` has it). Hence this request.

### Additional Context

Versions:
- livekit-agents 1.5.17
- livekit (rtc) 1.1.8
- livekit-plugins-openai 1.5.17
- livekit-plugins-google 1.5.17
- Python 3.13, macOS

Relevant code references:
- UserInputTranscribedEvent fields: livekit/agents/voice/events.py (class UserInputTranscribedEvent)
- item_id dropped on re-emit: livekit/agents/voice/agent_activity.py (_on_input_audio_transcription_completed)
- item_id present at source: livekit/agents/llm/realtime.py (class InputTranscriptionCompleted)
- OpenAI/xAI emit one interim per delta chunk: livekit/plugins/openai/realtime/realtime_model.py (_handle_conversion_item_input_audio_transcription_delta)
- Gemini emits the same normalized event from a different base class: livekit/plugins/google/realtime/realtime_api.py (_handle_server_content)

Confirmed behavior: with the same handler bound to `input_audio_transcription_completed`, interim events fire repeatedly per utterance on both OpenAI/xAI (token-by-token) and Gemini (whole-sentence in one interim). A passed-through `item_id` would let consumers collapse these to one action per utterance uniformly.

DIYA73:
Closes livekit/agents#6110

## What

Adds `item_id: str | None = None` to `UserInputTranscribedEvent` and forwards<br>`ev.item_id` from `llm.InputTranscriptionCompleted` in<br>`AgentActivity._on_input_audio_transcription_completed`.

## Why

Realtime models (OpenAI/xAI/Gemini) emit multiple interim<br>`user_input_transcribed` events per utterance. `InputTranscriptionCompleted`<br>already carries a stable `item_id` to correlate these, but it was being<br>dropped when re-emitted as `UserInputTranscribedEvent`, forcing consumers to<br>either track state manually (impossible — the field wasn't exposed) or<br>subscribe to provider-specific raw events.

## Testing

* Added `test_user_input_transcribed_event_item_id` covering both the<br>populated and default-None cases
* `ruff check` / `ruff format`: clean
* `mypy` (full project, 601 source files): 0 issues
* Existing test suite: no regressions

### DIFF
--- a/livekit-agents/livekit/agents/voice/events.py
+++ b/livekit-agents/livekit/agents/voice/events.py
@@ -331,7 +331,11 @@ class UserInputTranscribedEvent(BaseModel):
     transcript: str
     is_final: bool
     item_id: str | None = None
-    """Provider-specific ID for the transcribed input item, when available."""
+    """Stable id correlating all interim transcripts belonging to the same utterance.
+
+    Mirrors ``llm.InputTranscriptionCompleted.item_id``. Useful for deduplicating
+    repeated interim events from realtime models (e.g. only reacting once per
+    utterance instead of once per streamed delta chunk)."""
     speaker_id: str | None = None
     language: LanguageCode | None = None
     created_at: float = Field(default_factory=time.time)
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -114,6 +114,7 @@ def test_user_input_transcribed_event_item_id():
     ev_no_id = UserInputTranscribedEvent(transcript="hello", is_final=True)
     assert ev_no_id.item_id is None
 
+
 async def test_events_and_metrics() -> None:
     speed = 1
     actions = FakeActions()
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -31,7 +31,6 @@ from livekit.agents import (
 from livekit.agents.llm import (
     FunctionTool,
     FunctionToolCall,
-    InputTranscriptionCompleted,
     RawFunctionTool,
     ToolContext,
     ToolFlag,
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -100,30 +100,19 @@ class MyAgent(Agent):
 SESSION_TIMEOUT = 60.0
 
 
-def test_realtime_user_input_transcription_preserves_item_id() -> None:
-    captured_events: list[UserInputTranscribedEvent] = []
+def test_user_input_transcribed_event_item_id():
+    """UserInputTranscribedEvent should accept and expose item_id (#6110).
 
-    class DummySession:
-        def _user_input_transcribed(self, ev: UserInputTranscribedEvent) -> None:
-            captured_events.append(ev)
+    item_id is the stable correlation key from llm.InputTranscriptionCompleted,
+    forwarded by AgentActivity._on_input_audio_transcription_completed so realtime
+    consumers can dedup repeated interim transcripts belonging to the same utterance.
+    """
+    ev = UserInputTranscribedEvent(transcript="hello", is_final=False, item_id="item_abc123")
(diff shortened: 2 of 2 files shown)


livekit-agents/livekit/agents/voice/events.py
```diff
@@ -331,7 +331,11 @@ class UserInputTranscribedEvent(BaseModel):
     transcript: str
     is_final: bool
     item_id: str | None = None
-    """Provider-specific ID for the transcribed input item, when available."""
+    """Stable id correlating all interim transcripts belonging to the same utterance.
```
longcw, line 334: this pr is only for doc change, have you confirmed the `item_id` here is stable id, and it's different across turns?

DIYA73: Confirmed both.

Forwarding is intact — `_on_input_audio_transcription_completed` still sets `item_id=ev.item_id` (agent_activity.py:1842), so this diff really is just the docstring catching up to what the code already does.

On stability: every realtime plugin sources `item_id` from the provider's own per-turn id — OpenAI and xAI both pass through `event.item_id` from the provider event, Google uses `gen.input_id`, AWS uses `turn.input_id`. Same id across a turn's interim → final events, new id for the next turn. Checked the call sites rather than assuming.

PR #6173, opened 2026-06-20
fix: bound fallback LLM time to first chunk

### LINKED ISSUE #6171: FallbackAdapter attempt_timeout is a read (inter-byte) timeout, not time-to-first-token — a provider that stalls before its first token causes long dead air with no failover
### Bug Description

`llm.FallbackAdapter(attempt_timeout=...)` does not bound **time-to-first-token** — it ends up as an HTTP **read (inter-byte) timeout**, not a total or first-token budget.

If a provider accepts the request and keeps the SSE stream alive (periodic keep-alive frames, or a role-only delta with empty content) but never emits a real content/tool token, the read timeout is continually reset, no `ChatChunk` is ever produced, and the `FallbackAdapter` **never fails over to the next provider**. The turn hangs far beyond `attempt_timeout` (we saw ~60s, bounded only by the upstream's own timeout / the user hanging up), even though healthy fallback providers are configured — which defeats the purpose of the adapter.

Root cause chain:
1. `FallbackAdapter(attempt_timeout=5.0)` → `FallbackAdapterStream._try_generate` passes it as `conn_options.timeout`.
2. The OpenAI plugin applies it as `timeout=httpx.Timeout(self._conn_options.timeout)` on `client.chat.completions.create(..., stream=True)` (`agents/inference/llm.py`). A scalar `httpx.Timeout` is a per-read (inter-byte) timeout.
3. The stream loop only forwards chunks with real content; empty/role-only/keep-alive frames yield no `ChatChunk`.

So a provider that trickles bytes without ever producing a token keeps the read timeout from firing, the attempt never raises, and `_try_generate`'s `except APIError`/`asyncio.TimeoutError` failover path is never reached. (Proof it's a trickle and not a dead socket: a silent socket *would* trip the read timeout at `attempt_timeout`; in our case the request stayed open ~60s.)

### Expected Behavior

If a provider produces no first token within the configured budget, the attempt should be aborted and `FallbackAdapter` should fail over to the next provider — instead of hanging indefinitely with no output. The adapter's core promise (route around a non-responsive provider) should hold even when the provider keeps the connection alive without emitting tokens.

### Reproduction Steps

```bash
1. Start an OpenAI-compatible mock that emits a role chunk then trickles empty keep-alives forever (never a real token):


# mock_stall.py  — run on :8999
import asyncio, json, time
from aiohttp import web

async def chat(req):
    r = web.StreamResponse(headers={"Content-Type": "text/event-stream"})
    await r.prepare(req)
    def sse(delta):
        return f'data: {json.dumps({"id":"x","object":"chat.completion.chunk","created":int(time.time()),"model":"m","choices":[{"index":0,"delta":delta,"finish_reason":None}]})}\n\n'.encode()
    await r.write(sse({"role": "assistant", "content": ""}))   # not a real token
    while True:                                                # keep-alive trickle, no content
        await asyncio.sleep(1.5); await r.write(sse({"content": ""}))
    return r

app = web.Application(); app.router.add_post("/v1/chat/completions", chat)
web.run_app(app, port=8999)

2. Put the stalling endpoint as the primary in a FallbackAdapter with a healthy fallback and a short attempt_timeout:

# repro.py
import asyncio

async def main():
    stalling = openai.LLM(model="m", base_url="http://localhost:8999/v1", api_key="x")
    healthy  = openai.LLM(model="gpt-4o-mini")  # real OPENAI_API_KEY
    fb = llm.FallbackAdapter([stalling, healthy], attempt_timeout=5.0)
    async with fb.chat(chat_ctx=llm.ChatContext.empty()) as s:
        async for chunk in s:
            print(chunk)

asyncio.run(main())

3. Observe: nothing is printed for far longer than attempt_timeout=5s; the adapter never fails over to healthy. Expected: failover within ~5s and a response from healthy.
```

### Operating System

Windows 11 (platform-independent — the behavior is in the async/HTTP layer, not OS-specific)

### Models Used

LLM: OpenAI-compatible endpoint (originally observed with Azure OpenAI gpt-4.1 and an OpenAI-compatible proxy). STT/TTS not relevant — bug is in llm.FallbackAdapter.

### Package Versions

```bash
livekit==1.1.8
livekit-agents==1.5.12
livekit-api==1.1.0
livekit-plugins-openai==1.5.12
livekit-protocol==1.1.15
```

### Session/Room/Call IDs

N/A — fully reproducible locally with the snippet above; not tied to a LiveKit Cloud session.

### Proposed Solution

```python
### Proposed Solution

Bound time-to-first-token per attempt inside `FallbackAdapterStream._try_generate`: wrap the wait for the first real chunk in `asyncio.timeout(...)` and raise a retryable `APITimeoutError` on expiry, so the existing `except APIError` path fails over. Two options:

- (a) Reuse `attempt_timeout` as a first-token deadline (pure behavior fix, but changes timing for existing users), or
- (b) Add an opt-in `first_token_timeout: float | None = None` param so default behavior is unchanged.

I'd lean toward (b) for backward-compat. The guard should disarm after the first real token (content **or** tool-call delta) so mid-stream behavior is untouched, and close the inner stream on timeout to release the hung connection. Happy to open a PR once a maintainer confirms the preferred option (a vs b).
```

### Additional Context

_No response_

_No response_

nightcityblade:
Fixes #6171

## Summary
- enforce `attempt_timeout` while waiting for the first yielded chat chunk in `FallbackAdapter`
- fall through to the next LLM when the current provider stalls before producing a chunk
- add a regression test covering slow time-to-first-chunk fallback

## Tests
- `uv run ruff format --check livekit-agents/livekit/agents/llm/fallback_adapter.py tests/test_llm_fallback.py`
- `uv run ruff check livekit-agents/livekit/agents/llm/fallback_adapter.py tests/test_llm_fallback.py`
- `uv run pytest tests/test_llm_fallback.py -q`

### DIFF
--- a/livekit-agents/livekit/agents/llm/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/fallback_adapter.py
@@ -177,7 +177,21 @@ class FallbackLLMStream(LLMStream):
                 ),
             ) as stream:
                 should_set_current = not check_recovery
-                async for chunk in stream:
+                stream_iter = stream.__aiter__()
+                try:
+                    first_chunk = await asyncio.wait_for(
+                        stream_iter.__anext__(),
+                        timeout=self._fallback_adapter._attempt_timeout,
+                    )
+                except StopAsyncIteration:
+                    return
+
+                if should_set_current:
+                    should_set_current = False
+                    self._current_stream = stream
+                yield first_chunk
+
+                async for chunk in stream_iter:
                     if should_set_current:
                         should_set_current = False
                         self._current_stream = stream
--- a/tests/test_llm_fallback.py
+++ b/tests/test_llm_fallback.py
@@ -1,0 +1,33 @@
+import asyncio
+import contextlib
+
+import pytest
+
+from livekit.agents.llm import ChatContext
+from livekit.agents.llm.fallback_adapter import FallbackAdapter
+
+from .fake_llm import FakeLLM, FakeLLMResponse
+
+pytestmark = pytest.mark.unit
+
+
+@pytest.mark.asyncio
+async def test_fallback_adapter_attempt_timeout_bounds_time_to_first_chunk() -> None:
+    slow = FakeLLM(
+        fake_responses=[FakeLLMResponse(input="hello", content="slow", ttft=1, duration=1)]
+    )
+    fast = FakeLLM(
+        fake_responses=[FakeLLMResponse(input="hello", content="fast", ttft=0, duration=0)]
+    )
+    fallback = FallbackAdapter([slow, fast], attempt_timeout=0.01)
+    chat_ctx = ChatContext()
+    chat_ctx.add_message(role="user", content="hello")
+
+    response = await fallback.chat(chat_ctx=chat_ctx).collect()
+
+    assert response.text == "fast"
+    if recovery_task := fallback._status[0].recovering_task:
+        recovery_task.cancel()
+        with contextlib.suppress(asyncio.CancelledError):
+            await recovery_task
+    await fallback.aclose()


ZaafirRizwan (commented):
Thanks for tackling this @nightcityblade — we hit this exact issue in production (a streaming provider that trickles keep-alive frames without ever emitting a token, so the read timeout never fires and the turn hangs with no failover). The adapter-level asyncio.wait_for on the first chunk is a clean fix, and nice that it disarms after the first chunk so mid-stream gaps aren't affected.

A few things worth considering:

1. Backward-compat of reusing attempt_timeout. Since this now also bounds time-to-first-chunk (default 5s), providers with a legitimately slow first token — reasoning models that think for several seconds, or a healthy-but-loaded endpoint — would start failing over where they didn't before. Is the intent to change attempt_timeout's meaning, or would a separate opt-in (e.g. first_token_timeout, off or larger by default) be safer for existing users? Happy to go whichever way maintainers prefer.

2. Test coverage for the original trigger. The current test exercises a ttft delay, which validates the adapter-level bound. It might be worth also simulating the real-world cause — a stream that trickles empty/keep-alive chunks without a real token — so the bug that motivated this is documented and guarded against regressions. I have a small OpenAI-compatible mock that reproduces it and am happy to share/contribute it.

3. Tool-call turns. The OpenAI plugin buffers a tool call and only emits the ChatChunk at completion, so for a tool-only turn the "first chunk" effectively arrives when the tool call finishes streaming. With a tight timeout, a turn streaming a large tool-argument payload could trip it mid-call — is that a concern for the targeted use cases?

PR #6175, opened 2026-06-20
fix(slng): expose speed in update_options

### LINKED ISSUE #6174: fix(slng): update_options missing speed parameter
## Bug Description

`TTS.update_options` exposes `voice` and `language` but omits `speed`, even though `speed` is a first-class field in `_TTSOptions` and is sent to the SLNG gateway in the WebSocket init payload via `build_tts_init_payload`.

Users can set speed at construction time but cannot dynamically change it afterward via `update_options`.

## Reproduction

```python
tts = TTS(speed=1.0, ...)
tts.update_options(speed=1.5)  # TypeError: unexpected keyword argument 'speed'
```

## Fix

Add `speed` to `update_options` and invalidate the connection pool when the value changes, since speed is part of the WebSocket init payload rather than a per-request parameter.

adityajha2005:
`TTS.update_options` accepted `voice` and `language` but silently dropped `speed`, even though speed is a first-class `_TTSOptions` field sent in the WebSocket init payload via `build_tts_init_payload`. Users who set speed at construction time had no way to change it afterward.

Fix: add `speed` to `update_options` and invalidate the connection pool when the value changes, consistent with how `voice` and `language` are handled.

Fixes #6174

### DIFF
--- a/livekit-plugins/livekit-plugins-slng/livekit/plugins/slng/tts.py
+++ b/livekit-plugins/livekit-plugins-slng/livekit/plugins/slng/tts.py
@@ -301,11 +301,13 @@ class TTS(tts.TTS):
         *,
         voice: NotGivenOr[str] = NOT_GIVEN,
         language: NotGivenOr[str] = NOT_GIVEN,
+        speed: NotGivenOr[float] = NOT_GIVEN,
     ) -> None:
         """
         Args:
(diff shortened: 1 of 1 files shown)


tinalenguyen (approved):
lgtm, thanks!

PR #6178, opened 2026-06-21
chore(example): update otel trace example

chenghao-mou:
make the otel trace example more generic (tested with langfuse)


longcw (approved):
lgtm! maybe fix the comment from devin before merge?

PR #6193, opened 2026-06-23
fix(google): clean up STT input frame task

### LINKED ISSUE #6181: Unhandled ChanClosed task warning with Google STT
### Bug Description

PR #5591 introduced a noisy asyncio warning when using Google STT.  After the change, I constantly see logs like:
```
Task exception was never retrieved
future: <Task finished ... coro=<Chan.recv()> exception=ChanClosed()>
livekit.agents.utils.aio.channel.ChanClosed
```

### Expected Behavior

No asyncio exceptions should appear

### Reproduction Steps

```bash
Use Google STT with livekit-agents 1.6.2 or newer
```

### Operating System

Ubuntu 24.04

### Models Used

chirp_3/google/gemini-3.5-flash/eleven_flash_v2_5

### Package Versions

```bash
livekit-agents==1.6.1
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

gyx09212214-prog:
## Summary

Fixes #6181.

This cleans up the Google STT streaming request generator when the surrounding gRPC stream is cancelled while it is waiting for the next audio frame. The user-visible streaming behavior is unchanged, but the pending `self._input_ch.recv()` task is now cancelled and awaited before the generator exits.

## Background

PR #5591 added a race between `self._input_ch.recv()` and a stop event so Google STT can reconnect without leaving an idle request generator parked on the input channel. That fixed the idle-stream leak, but left a smaller cleanup gap: if the request generator itself is cancelled while `frame_task` is still waiting on `Chan.recv()`, the generator's `finally` block only cancelled `stop_task`.

When the input channel later closes, that orphaned task can finish with `ChanClosed`, causing:

`Task exception was never retrieved`
`future: <Task finished ... coro=<Chan.recv()> exception=ChanClosed()>`

## Changes

- Track the active Google STT input `frame_task`.
- Cancel and await both `stop_task` and any pending `frame_task` via `utils.aio.gracefully_cancel`.
- Add a regression test covering stream cancellation while waiting for audio.
- Mock Google ADC only for config-only tests so `test_plugin_google_stt.py` can run locally without real credentials.

## Validation

- `python -m pytest tests/test_plugin_google_stt.py -q`
  - `16 passed`
- `python -m ruff format --check livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py tests/test_plugin_google_stt.py`
- `python -m ruff check livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py tests/test_plugin_google_stt.py`
- `python -m py_compile livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py tests/test_plugin_google_stt.py`
- `git diff --check`

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
@@ -786,7 +786,12 @@ class SpeechStream(stt.SpeechStream):
                 tasks: list[asyncio.Task[object]] = [stop_task]
                 if frame_task is not None:
                     tasks.append(frame_task)
-                await utils.aio.gracefully_cancel(*tasks)
+                cleanup_task = asyncio.create_task(utils.aio.gracefully_cancel(*tasks))
+                try:
+                    await asyncio.shield(cleanup_task)
+                except asyncio.CancelledError:
+                    await cleanup_task
+                    raise
 
         async def process_stream(
             client: SpeechAsyncClientV2 | SpeechAsyncClientV1,
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
@@ -753,6 +753,7 @@ class SpeechStream(stt.SpeechStream):
         ]:
             nonlocal audio_pushed
             stop_task = asyncio.create_task(should_stop.wait())
+            frame_task: asyncio.Task[object] | None = None
             try:
                 yield self._build_init_request(client)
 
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -781,7 +783,15 @@ async def input_generator(
...
-                stop_task.cancel()
+                tasks: list[asyncio.Task[object]] = [stop_task]
+                if frame_task is not None:
+                    tasks.append(frame_task)
+                cleanup_task = asyncio.create_task(utils.aio.gracefully_cancel(*tasks))
+                try:
+                    await asyncio.shield(cleanup_task)
+                except asyncio.CancelledError:
+                    await cleanup_task
+                    raise
```
longcw: no need to shield the cleanup task, otherwise looks good to me
```suggestion
                await utils.aio.gracefully_cancel(
                    stop_task, *([frame_task] if frame_task is not None else [])
                )
```

longcw (approved):

PR #6197, opened 2026-06-23
chore(elevenlabs): update `language_code` doc to remove error case

inickt:
Tiny doc update after experimenting with fixed language codes and talking with ElevenLabs support:

> `language_code` isn’t supported by `eleven_multilingual_v2`, since it determines the language from the surrounding context. As of the last update, it’s simply ignored, and we don’t throw an explicit error.
>
> `language_code` does work with `eleven_flash_v2_5`.
>
> The LiveKit docs you linked also seem very old.


tinalenguyen (approved):
thank you for the correction!

PR #6199, opened 2026-06-24
(fish audio tts): add model str

tinalenguyen:


longcw (approved):

livekit-plugins/livekit-plugins-fishaudio/livekit/plugins/fishaudio/models.py
```diff
@@ -1,6 +1,6 @@
 from typing import Literal
 
-TTSModels = Literal["s1", "s2-pro", "s2.1-pro"]
+TTSModels = Literal["s1", "s2-pro", "s2.1-pro", "s2.1-pro-free"]
```
davidzhao, line 3: pretty sure this is temporary and will be gone after promo. should we add this?

PR #6202, opened 2026-06-24
feat: add multi-turn answering-machine detection

chenghao-mou:
AMD handles screening, voicemail, and IVR across turns using rolling history and successful DTMF calls. `WAIT` suppresses replies during announcements; stale predictions are discarded.

A category FSM constrains predictions. AMD owns deadlines and reply authorization, supports realtime replies, and restores interruption settings after detection. `execute()` waits for completion. See [design and usage](https://github.com/livekit/agents/blob/chenghao/feat/AGT-3038-call-screening-support-in-amd/examples/telephony/amd.md).

Addresses [AGT-3038](https://linear.app/livekit/issue/AGT-3038/handle-call-screening). Adopts @karan-dhir's [screening proposal](https://github.com/livekit/agents-js/pull/1386) ([source issue](https://github.com/livekit/agents-js/issues/1374)) and @eashwar-mp's [rolling-history guidance](https://github.com/livekit/agents/pull/6202#discussion_r3837424261).

### DIFF
--- a/livekit-agents/livekit/agents/voice/amd/classifier.py
+++ b/livekit-agents/livekit/agents/voice/amd/classifier.py
@@ -139,7 +139,7 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
         self._classify_task: asyncio.Task[None] | None = None
         self._no_speech_timer: asyncio.TimerHandle | None = None
         self._silence_timer: asyncio.TimerHandle | None = None
-        self._silence_timer_trigger: Literal["short_speech", "long_speech"] | None = None
+        self._silence_timer_kind: Literal["human_verdict", "classification"] | None = None
         self._detection_timeout_timer: asyncio.TimerHandle | None = None
         self._eot_timer: asyncio.TimerHandle | None = None
 
@@ -160,7 +160,7 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
         self._extension_count = 0
         self._amd_stt_seen = False
 
-    def start_detection_timer(self) -> None:
+    def arm_detection_timer(self) -> None:
         """Arm or reset the overall detection-timeout budget."""
         if self._closed or self._emitted:
             return
@@ -175,8 +175,8 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
             ),
         )
 
-    def start_listening(self, *, arm_no_speech_timer: bool = True) -> None:
-        """Open the input gate and optionally arm the no-speech timer.
+    def start_listening(self) -> None:
+        """Open the input gate.
 
         Call once we expect audible speech to begin (e.g. after sip answer
         for outbound calls). Until this fires, all input methods are no-ops.
@@ -184,18 +184,16 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
         if self._closed or self._emitted or self._listening:
             return
         self._listening = True
-        if arm_no_speech_timer and self._no_speech_timer is None:
-            self._arm_no_speech_timer()
 
-    def arm_turn_timers(self) -> None:
-        """Start a fresh turn budget after message playout."""
-        if self._closed or self._emitted or not self._listening:
+    def arm_no_speech_timer(self) -> None:
+        if (
+            self._closed
+            or self._emitted
+            or not self._listening
+            or self._speech_started_at is not None
+            or self._transcript
+        ):
             return
-        self.start_detection_timer()
-        if self._speech_started_at is None and not self._transcript:
-            self._arm_no_speech_timer()
-
-    def _arm_no_speech_timer(self) -> None:
         if self._no_speech_timer is not None:
             self._no_speech_timer.cancel()
         self._no_speech_timer = asyncio.get_running_loop().call_later(
@@ -212,7 +210,7 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
         if self._silence_timer is not None:
             self._silence_timer.cancel()
             self._silence_timer = None
-            self._silence_timer_trigger = None
+            self._silence_timer_kind = None
         if self._no_speech_timer is not None:
             self._no_speech_timer.cancel()
             self._no_speech_timer = None
@@ -240,7 +238,7 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
             if self._silence_timer is not None:
                 self._silence_timer.cancel()
                 self._silence_timer = None
-                self._silence_timer_trigger = None
+                self._silence_timer_kind = None
             if not self._transcript:
                 self._silence_timer = asyncio.get_running_loop().call_later(
                     max(0, self._human_silence_threshold - silence_duration),
@@ -251,13 +249,13 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
                         speech_duration=speech_duration,
                     ),
                 )
-                self._silence_timer_trigger = "short_speech"
+                self._silence_timer_kind = "human_verdict"
             else:
                 self._silence_timer = asyncio.get_running_loop().call_later(
                     max(0, self._machine_silence_threshold - silence_duration),
                     self._on_silence_reached,
                 )
-                self._silence_timer_trigger = "long_speech"
+                self._silence_timer_kind = "classification"
             return
 
         if self._classify_task is None:
@@ -266,12 +264,12 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
         if self._silence_timer is not None:
             self._silence_timer.cancel()
             self._silence_timer = None
-            self._silence_timer_trigger = None
+            self._silence_timer_kind = None
         self._silence_timer = asyncio.get_running_loop().call_later(
             max(0, self._machine_silence_threshold - silence_duration),
             self._on_silence_reached,
         )
-        self._silence_timer_trigger = "long_speech"
+        self._silence_timer_kind = "classification"
 
     def _set_verdict(self, result: AMDPredictionEvent) -> None:
         self._verdict_result = result
@@ -360,7 +358,7 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
         if self._closed:
             return
         self._silence_timer = None
-        self._silence_timer_trigger = None
+        self._silence_timer_kind = None
         self._silence_reached = True
         self._try_emit_result()
 
@@ -390,7 +388,7 @@ class _AMDClassifier(EventEmitter[Literal["amd_prediction"]]):
         if self._silence_timer:
             self._silence_timer.cancel()
             self._silence_timer = None
-            self._silence_timer_trigger = None
+            self._silence_timer_kind = None
 
         self._silence_reached = True
         has_speech = self._speech_started_at is not None or bool(self._transcript)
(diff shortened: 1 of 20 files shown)


livekit-agents/livekit/agents/voice/amd/classifier.py
```diff
@@ -75,8 +85,7 @@ def is_machine(self) -> bool:
 Output: machine-ivr
 
 Input: "Please state your name and why you're calling, and I will check if the person is available"
-Output: machine-ivr
-Note: this should apply for any call screening prompts.
+Output: machine-screening
```
eashwar-mp, line 88: @chenghao-mou from running outbound screening in production, this category is genuinely non-deterministic on the LLM alone. We saw the same Google Call Screen greeting classified as machine-vm once and machine-ivr twice across 11 calls, and a single example won't cover the real spread. Two things that helped us: a lot more examples, and a deterministic transcript substring pre-gate before trusting the LLM (matched at the start of the greeting so STT truncation doesn't break it). The failure if it's wrong: a screener tagged machine-vm plays the voicemail message and ends, so you leave a voicemail on a call screen instead of reaching the human.

Screener markers we substring-match on:
`call assist`, `by google`, `automatically screening`, `record your name`, `reason for calling`, `asking for more information`, `if this person is available`, `see if this person is available`

And the phrases we use to detect a screener rolling over to message-taking:
`leave a message`, `leave me a message`, `leave an additional message`, `leave your message`, `record your message`, `after the tone`, `after the beep`, `at the tone`, `at the beep`, `forwarded to voicemail`, `reached the voicemail` (plus combos like `take the call` + `add anything else`). Happy to share the full lists as examples to fold into the prompt.

livekit-agents/livekit/agents/voice/amd/detector.py
```diff
@@ -121,26 +143,34 @@ class AMD(EventEmitter[Literal["amd_prediction"]]):
...
         session: The :class:`AgentSession` to wire AMD to.
         llm: LLM used for greeting classification. Accepts an :class:`LLM`
-            instance or an inference model string (e.g.
-            ``"openai/gpt-4.1-mini"``). When omitted, AMD auto-selects:
-            if LiveKit inference credentials are available in the environment
-            it uses ``"google/gemini-3.1-flash-lite"`` via the
-            inference gateway; otherwise it falls back to the session's own
-            LLM.
+            instance or an inference model string (e.g. ``"openai/gpt-4.1-mini"``).
+            When omitted, AMD reuses the session's own LLM (``session.llm``).
```
eashwar-mp, line 147: Is dropping the auto-selected gemini-flash-lite / ink-whisper defaults intended as part of this PR? It's a behavior change for anyone relying on auto-select, and reusing the session's main LLM for classification can be slower and, in our experience, less reliable on the vm-vs-screening call specifically. Might be worth calling out in the description.

livekit-agents/livekit/agents/voice/amd/detector.py
```diff
@@ -513,10 +533,141 @@ async def _receive() -> None:
...
+                    playback = await self._play(self._screening_message, verdict)
+                    if playback == "not_played":
+                        self._finish(verdict, playback)
+                        return
+                    logger.info("screening message: %s", playback)
+                    classifier.arm_detection_timer()
+                    classifier.arm_no_speech_timer()
+                    # remember the screening playback so a later human verdict reports it
+                    self._last_playback = playback
+                    continue
```
eashwar-mp, line 593: Is there a cap on screening turns here? Each machine-screening verdict replays screening_message and re-arms the detection timer, so a chatty screener or a flaky re-classification could loop the message with no upper bound. A max-iterations guard might be worth adding.

livekit-agents/livekit/agents/voice/amd/classifier.py
```diff
@@ -382,24 +403,37 @@ def push_text(self, text: str, source: str = "stt") -> None:
         if self._input_ch.closed:
             logger.debug("push_text called after close")
             return
-        # ignore text from other sources (e.g. when both session and AMD have STT specified)
         if source != self._source:
-            return
-        self._arm_eot_timer()
+            # pick session stt when it arrives before AMD stt
+            if self._source == "amd_stt" and source == "stt" and not self._amd_stt_seen:
+                logger.warning("amd: session STT won the transcript race, using session STT")
```
eashwar-mp, line 409: nit: this fires on the normal dual-STT path whenever session STT wins the race, which is expected. Might read better as debug so it's not warning-level noise.

livekit-agents/livekit/agents/voice/amd/detector.py
```diff
@@ -513,10 +533,141 @@ async def _receive() -> None:
...
+                        await self._session.interrupt(force=True)
+                    if not is_given(self._screening_message):
+                        # surface the screening verdict instead of timing out
+                        self._finish(verdict, "not_played")
+                        return
+
+                    logger.info("playing screening message")
+                    await classifier.reset()
+                    classifier.start_listening()
+                    self._session._on_aec_warmup_expired()
```
eashwar-mp, line 582: Since the classifier keeps listening while screening_message plays, is there a risk AMD transcribes the agent's own message as the callee's next greeting? Looks like _on_aec_warmup_expired is meant to cover that, just want to confirm the playout can't poison the next turn's classification.

livekit-agents/livekit/agents/voice/remote_session.py
```diff
@@ -276,6 +276,8 @@ async def __anext__(self) -> agent_pb.AgentSessionMessage:
 _AMD_CATEGORY_MAP: dict[AMDCategory, agent_pb.AmdCategory] = {
     AMDCategory.HUMAN: agent_pb.AmdCategory.AMD_HUMAN,
     AMDCategory.MACHINE_IVR: agent_pb.AmdCategory.AMD_MACHINE_IVR,
+    # TODO: @chenghao-mou The remote-session protocol does not yet represent screening.
+    AMDCategory.MACHINE_SCREENING: agent_pb.AmdCategory.AMD_UNKNOWN,
```
eashwar-mp, line 280: Worth flagging for remote-session users: screening collapses to AMD_UNKNOWN here, and screening_detected/message_playback aren't on the proto, so the whole signal is invisible over the wire. Fine as a follow-up given the TODO, just noting the gap.

livekit-agents/livekit/agents/voice/amd/detector.py
```diff
@@ -513,10 +533,141 @@ async def _receive() -> None:
...
+                        await self._session.interrupt(force=True)
+                    if self._ivr_detection:
+                        await self._session._start_ivr_detection(transcript=verdict.transcript)
+                    self._finish(verdict, self._last_playback)
+                    return
+
+                if verdict.is_machine and self._interrupt_on_machine:
+                    await self._session.interrupt(force=True)
+
+                if verdict.category == AMDCategory.MACHINE_VM:
```
eashwar-mp, line 606: One case from our screener handling worth considering: Google Call Screen often rolls over to message-taking with the decline split across several short STT turns ("okay", "they can't take the call", "feel free to leave a message"). Since each turn is classified independently after reset(), a rollover spread across turns may never produce a single transcript the LLM tags machine-vm, so the rollover-to-voicemail gets missed and the call just times out. We catch it by matching a voicemail-takeover over a rolling multi-turn window.

livekit-agents/livekit/agents/voice/amd/classifier.py
```diff
@@ -497,6 +531,79 @@ async def _run(transcript: str) -> None:
...
 
+    def _synthesize_short_utterance(self) -> None:
+        """Create a short speech bracket for a transcript missed by VAD."""
+        now = time.time()
+        self._speech_started_at = now
+        self._speech_ended_at = now
+        if self._silence_timer is not None:
+            self._silence_timer.cancel()
+        self._silence_timer = asyncio.get_running_loop().call_later(
+            self._human_silence_threshold,
```
eashwar-mp, line 542: Heads-up from a prod bug (we ended up widening human_silence from 0.5 to 1.0): a voicemail greeting whose start VAD misses runs through this synthesized bracket on the human_silence window. With the shorter default there's a risk the silence timer fires before a slow classification lands. Probably fine since this waits for the verdict rather than pre-baking human, but flagging since the short window bit us.

eashwar-mp (commented):
Nice work, this lands right where we need it. Sharing a few things from running outbound AMD (screeners + voicemail) in production that might save some pain, especially around trusting the LLM for the screening category. Happy to dig into any of these.

livekit-agents/livekit/agents/voice/amd/classifier.py
chenghao-mou: Thanks for the feedback. Just to make sure I understand this scope when you say "the same Google Call Screen greeting classified as machine-vm once and machine-ivr twice across 11 calls, and a single example won't cover the real spread." do you mean running the existing no-screening version with the default AMD LLM?

eashwar-mp: Yes, exactly, this was the pre-#6202 build with no `machine-screening` category, running the default auto-selected AMD LLM (`gemini-3.1-flash-lite` + `ink-whisper`, we pass no explicit `llm`/`stt`). So a screener could only surface as `machine-ivr` (which we then sub-classify screener-vs-keypad by substring) or misfire to `machine-vm`.

What we saw: across a small sample, the identical "Colossus by Google" greeting came back `machine-ivr` twice and `machine-vm` once, with `reason=llm` on the verdict, and roughly 1 in 11 screener calls misfired to a non-screener category. Confirmed against the CallState telemetry and the STT transcripts, same greeting text each time.

Fair caveat: this was without your new dedicated category and its prompt example, so that should help. My concern is just that the same greeting doesn't always get the same verdict, so one example won't fully fix it on its own.

chenghao-mou: Thanks for confirming. We now include Google call screening support in scope and will evaluate its performance before release.

livekit-agents/livekit/agents/voice/amd/detector.py
chenghao-mou: That was a detour during the initial design. We now have the following rank:
- explicit values
- omitted and credentials present: our chosen models
- None or otherwise: fall back to session or agent settings

livekit-agents/livekit/agents/voice/amd/detector.py
chenghao-mou: Current design: AMD has a hard timeout from the start of listening, defaulting to 120 seconds, which new turns cannot extend. There is no screening-specific turn cap, so extended exchanges remain possible within that time limit (there is also no fixed screening message now that Google call screening support is added)

livekit-agents/livekit/agents/voice/amd/classifier.py
chenghao-mou: Agreed. That warning is gone. When session STT wins, AMD selects it and closes the dedicated stream without logging a warning. An actual AMD STT failure still logs a warning before falling back.

livekit-agents/livekit/agents/voice/amd/detector.py
chenghao-mou: AMD receives participant input audio, and its classifier history excludes the Agent’s replies. That prevents direct inclusion of our output, but it does not rule out echo returning through the participant’s audio. The same can be said for normal agent sessions as well. This also depends on device, environment, and many other factors where typically you already have echo cancellation embedded in the browser or microphones.

livekit-agents/livekit/agents/voice/remote_session.py
chenghao-mou: Yeah, the gap won't be closed until we ship core protobuf changes. It will be ready when releasing.

livekit-agents/livekit/agents/voice/amd/detector.py
chenghao-mou: The new design does include recent messages so the classification has more context about the current stage.

livekit-agents/livekit/agents/voice/amd/classifier.py
chenghao-mou: The synthesized human-silence is dropped. AMD now classifies at committed end of turn and waits for the prediction or its inference deadline.

PR #6206, opened 2026-06-24
fix(agents): keep STT input anchor on the pipeline across handoff

chenghao-mou:
## Problem

After an agent/AgentTask handoff that **reuses the STT pipeline**, a user turn could stall for a long time between the end-of-turn prediction and the actual commit. Observed in the async-tool example: a ~68s gap between `eot prediction` and `user turn committed`, with the (preemptively generated) agent turn span staying open the whole time while its LLM finished in 0.5s.

```
13:35:07  eot prediction  prob=0.82  endpointing_delay=0.3
13:36:15  user turn committed  last_speaking_time=<68s in the future>
```

## Root cause

The reused `_STTPipeline` keeps its STT stream open, so `end_time` keeps measuring from the **original** stream start. But `_input_started_at` (the wall-clock anchor for `end_time = 0`) was reset to `None` on the new `AudioRecognition` and re-anchored to the **handoff** time on the next frame.

`stt_last_speaking_time = end_time + _input_started_at` then double-counts the pre-handoff interval, yielding a `_last_speaking_time` minutes in the future. `_bounce_eou_task` waits `endpointing_delay + (last_speaking_time - now)`, so it sleeps for that entire drift before committing.

This surfaces whenever the STT timestamp is used for the speaking anchor — including the common default-VAD case (`using_default_vad=True`), where the VAD-derived time is intentionally discarded. That's why VAD never reported the user as speaking yet the turn still hung.

## Fix

Move the `input_started_at` anchor onto `_STTPipeline`, so it travels with the stream it describes, and stop resetting it in `update_stt` / `update_interruption_detection` when a pipeline is reused. `AudioRecognition._input_started_at` becomes a thin property delegating to the current pipeline, leaving all existing call sites unchanged.

## Tests

- `test_input_anchor_preserved_when_pipeline_reused` — `update_stt` with a reused pipeline keeps its anchor.
- `test_input_anchor_delegates_to_pipeline` — anchor lives on the pipeline.
- Updated `test_audio_recognition_push_audio.py` stub for the pipeline-owned anchor.

`make check` (format/lint/type) and the recognition/turn-detection unit suites pass.

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -154,6 +154,11 @@ class _STTPipeline:
         self._event_ch = aio.Chan[stt.SpeechEvent]()
         self._pump_task = asyncio.create_task(self._stt_pump())
         self._pump_task.add_done_callback(lambda _: self._event_ch.close())
+        # wall-clock anchor for this stream's `end_time=0`; lives on the pipeline so it
+        # survives agent handoff together with the stream it describes. Resetting it on a
+        # reused pipeline would desync STT `end_time` (relative to the original stream
+        # start) from wall clock, pushing the derived speaking time far into the future.
+        self.input_started_at: float | None = None
 
     @property
     def audio_ch(self) -> aio.Chan[rtc.AudioFrame]:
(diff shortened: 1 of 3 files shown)


chenghao-mou:
Superseded by #6207 (renamed branch to the correct `chenghao/` prefix).

PR #6208, opened 2026-06-24
feat(agents): allow updating recording options mid-session via update_options

chenghao-mou:
Adds a public way to change Agent Insights recording options during an active session, replacing the private `session._recording_options.update(...)`.

- `AgentSession.update_options(record=...)` — follows the canonical session `update_options` pattern. Accepts `True`/`False` to toggle every feature, or a partial `RecordingOptions` mapping that updates only the given keys (audio, traces, logs, transcript) and leaves the rest unchanged.
- `AgentSession.recording_options` — read-only property returning a copy of the current options.

This makes it easy to toggle recording, traces, logs, or transcript collection mid-session, e.g. after recording consent is obtained.

Note: session audio is only captured when audio recording was enabled at `start()`, so enabling `audio` afterwards does not retroactively start the recorder; disabling always takes effect.


chenghao-mou:
closing this for now

PR #6219, opened 2026-06-24
feat(assemblyai): add language_code streaming param for language steering

gsharp-aai:
## Summary

Adds a `language_code` connect-time parameter to the AssemblyAI STT plugin so users can **steer transcription toward a specific language** (e.g. `"en"`, `"es"`, `"fr"`) instead of relying on automatic detection / code-switching.

Today the plugin only exposes `language_detection`, which is an output toggle (whether `language_code`/`language_confidence` are returned on turn messages) — there is no way to *steer* the model toward a language. The AssemblyAI streaming API already accepts `language_code` as a connect-time parameter, so this just plumbs it through.

This is useful for known-monolingual sessions, where steering improves accuracy on short/ambiguous utterances (e.g. disambiguating "see" vs. "si").

## Details

- Added `language_code: NotGivenOr[str]` to `STTOptions` and the `STT.__init__` signature, forwarded into the connect-time `live_config` querystring (omitted when unset).
- **Gated to the u3-rt-pro family** (`u3-rt-pro`, `u3-rt-pro-beta-1`, `universal-3-5-pro`) via the existing `_U3_PRO_MODELS` validation — passing it with another model raises `ValueError`. Language *steering* is applied by the u3-pro ASR; on the universal-streaming models `language_code` does not steer, so this matches the parameter's documented behavior and how `mode`/`voice_focus` are handled.
- **Connect-time only**, matching the AssemblyAI streaming API — `language_code` is not part of `UpdateConfiguration`, so it is not added to `update_options`.
- Follows the structure of the recent `mode` param PR (#6156).

## Test plan

Added unit tests in `tests/test_plugin_assemblyai_stt.py`:

- default is `NOT_GIVEN`
- value is stored from the constructor
- raises `ValueError` on a non-u3-rt-pro-family model
- accepted across every u3-rt-pro-family model
- present in the connect config querystring when set
- absent from the querystring when unset
- not exposed via `update_options` (connect-time only)

```
65 passed   # full tests/test_plugin_assemblyai_stt.py
```

`ruff check` and `ruff format` pass on both changed files.

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -59,6 +59,7 @@ class STTOptions:
         "universal-3-5-pro",
     ] = "universal-3-5-pro"
     language_detection: NotGivenOr[bool] = NOT_GIVEN
+    language_code: NotGivenOr[str] = NOT_GIVEN
     end_of_turn_confidence_threshold: NotGivenOr[float] = NOT_GIVEN
     min_turn_silence: NotGivenOr[int] = NOT_GIVEN
     max_turn_silence: NotGivenOr[int] = NOT_GIVEN
@@ -101,6 +102,7 @@ class STT(stt.STT):
             "universal-3-5-pro",
         ] = "universal-3-5-pro",
         language_detection: NotGivenOr[bool] = NOT_GIVEN,
+        language_code: NotGivenOr[str] = NOT_GIVEN,
         end_of_turn_confidence_threshold: NotGivenOr[float] = NOT_GIVEN,
         min_turn_silence: NotGivenOr[int] = NOT_GIVEN,
         max_turn_silence: NotGivenOr[int] = NOT_GIVEN,
@@ -134,6 +136,12 @@ class STT(stt.STT):
                 0 and 1 that determines how sensitive the VAD is. Lower values make the VAD
                 more sensitive (detects quieter speech). Higher values make it less sensitive.
                 Defaults to 0.4.
+            language_code: Steer transcription toward a specific language (e.g. 'en', 'es',
+                'fr'). When set, the model is biased toward this language instead of
+                automatically detecting/code-switching across the supported languages.
+                Leave unset to use the model's default multilingual behavior. Only
+                supported with the 'u3-rt-pro' / 'u3-rt-pro-beta-1' / 'universal-3-5-pro'
+                models. Set at construction (connect) time only.
             min_turn_silence: Minimum silence in ms before a confident end-of-turn is finalized.
             min_end_of_turn_silence_when_confident: Deprecated. Use min_turn_silence instead.
             continuous_partials: Whether to emit additional partial transcripts during long
@@ -204,6 +212,7 @@ class STT(stt.STT):
                 "voice_focus": voice_focus,
                 "voice_focus_threshold": voice_focus_threshold,
                 "mode": mode,
+                "language_code": language_code,
             }
             for _param_name, _param_value in _u3_pro_only_params.items():
                 if is_given(_param_value):
@@ -250,6 +259,7 @@ class STT(stt.STT):
             encoding=encoding,
             speech_model=model,
             language_detection=language_detection,
+            language_code=language_code,
             end_of_turn_confidence_threshold=end_of_turn_confidence_threshold,
             min_turn_silence=min_turn_silence,
             max_turn_silence=max_turn_silence,
@@ -669,6 +679,9 @@ class SpeechStream(stt.SpeechStream):
             if "multilingual" in self._opts.speech_model
             or self._opts.speech_model in _U3_PRO_MODELS
             else False,
+            "language_code": self._opts.language_code
+            if is_given(self._opts.language_code)
+            else None,
             "prompt": self._opts.prompt if is_given(self._opts.prompt) else None,
             "agent_context": self._opts.agent_context
             if is_given(self._opts.agent_context)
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -662,6 +670,9 @@ async def _connect_ws(self) -> aiohttp.ClientWebSocketResponse:
             if "multilingual" in self._opts.speech_model
             or self._opts.speech_model in _U3_PRO_MODELS
             else False,
+            "language_code": self._opts.language_code
+            if is_given(self._opts.language_code)
+            else None,
```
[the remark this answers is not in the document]
gsharp-aai: 1. The flagged scenario can't occur. `language_code` is gated to the u3-rt-pro family via the `_U3_PRO_MODELS` validation — passing it with `universal-streaming-english` (or any non-family model) raises ValueError before a connect URL is ever built. So the `language_detection=false` + `language_code=es` combination on `universal-streaming-english` isn't reachable.

2. The dependency doesn't exist. `language_code` and `language_detection` are independent on the u3-pro family:

- `language_code` drives language steering (a system-prompt prepend in the u3-pro ASR).
- `language_detection` only controls whether language_code/language_confidence are reported on turn messages.

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -662,6 +672,9 @@ async def _connect_ws(self) -> aiohttp.ClientWebSocketResponse:
...
```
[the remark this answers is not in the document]
gsharp-aai: 1. The flagged scenario can't occur. `language_code` is gated to the `u3-rt-pro` family via the `_U3_PRO_MODELS` validation — passing it with `universal-streaming-english` (or any non-family model) raises `ValueError` before a connect URL is ever built. So the `language_detection=false` + `language_code=es` combination on `universal-streaming-english` isn't reachable.

2. The dependency doesn't exist. `language_code` and `language_detection` are independent on the u3-pro family:

- `language_code` drives language steering (a system-prompt prepend in the u3-pro ASR).
- `language_detection` only controls whether `language_code/language_confidence` are reported on turn messages.

gsharp-aai:
https://www.assemblyai.com/docs/api-reference/streaming-api/universal-3-pro-streaming
Screenshot 2026-06-24 at 4 35 38 PM

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -134,6 +136,12 @@ def __init__(
                 0 and 1 that determines how sensitive the VAD is. Lower values make the VAD
                 more sensitive (detects quieter speech). Higher values make it less sensitive.
                 Defaults to 0.4.
+            language_code: Steer transcription toward a specific language (e.g. 'en', 'es',
+                'fr'). When set, the model is biased toward this language instead of
+                automatically detecting/code-switching across the supported languages.
+                Leave unset to use the model's default multilingual behavior. Only
+                supported with the 'u3-rt-pro' / 'u3-rt-pro-beta-1' / 'universal-3-5-pro'
```
chenghao-mou: nit: we should use an umbrella term similar to this PR: https://github.com/livekit/agents/pull/6220/changes

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -59,6 +59,7 @@ class STTOptions:
         "universal-3-5-pro",
     ] = "universal-3-5-pro"
     language_detection: NotGivenOr[bool] = NOT_GIVEN
+    language_code: NotGivenOr[str] = NOT_GIVEN
```
chenghao-mou: we do have a pattern of using `LanguageCode` instead of str in options, and by default it normalizes the value to ISO 639-1 format.

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
gsharp-aai: Done — updated the docstring to use the "Universal-3 Pro family models" umbrella term, matching the wording from #6220. Also rebased onto main to pick up that change.

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
gsharp-aai: Switched to LanguageCode.

One intentional deviation: I normalize down to the bare base code via LanguageCode(...).language rather than relying on the default LanguageCode() normalization. The reason is that AssemblyAI's language_code steering only accepts bare ISO 639-1 codes (es, en, pt), so an unrecognized key like es-ES silently falls back to no steering rather than erroring.

The default LanguageCode() preserves region subtags (es-ES stays es-ES), which would silently break steering for any caller passing a region-tagged code. So 'es-ES' / 'Spanish' / 'en-US' all resolve to 'es' / 'es' / 'en'. Happy to drop the .language step and use plain LanguageCode if you'd prefer strict consistency with the pattern — just flagging the functional tradeoff.

Added test_language_code_normalized_to_iso_639_1 covering these cases.

livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -101,6 +102,7 @@ def __init__(
             "universal-3-5-pro",
         ] = "universal-3-5-pro",
         language_detection: NotGivenOr[bool] = NOT_GIVEN,
+        language_code: NotGivenOr[LanguageCode | str] = NOT_GIVEN,
```
tinalenguyen: i think we can just accept a str here so we don't need to normalize it like this later: `LanguageCode(LanguageCode(language_code).language)`
```suggestion
        language_code: NotGivenOr[str] = NOT_GIVEN,
```

gsharp-aai: Got it — reverted the signature to str and simplified to a single LanguageCode(language_code).language

tinalenguyen (approved):
thanks for the PR :D

PR #6221, opened 2026-06-25
fix(openai): expose azure_deployment on LLM and fix model property fallback

### LINKED ISSUE #6209: Remove default for "model" parameter of Azure OpenAI LLM
### Bug Description

[This is not strictly a bug but IMHO confusing and potentially risky]

When using OpenAI LLM with Azure, there is a default value for the `model` parameter. The LLM doesn't seem to use this parameter at all (instead uses `azure_deployment`), but given the relation to "real" OpenAI LLM this may change in the future. Then this default might become active unnoticed.

Also, the `model` property of the LLM object will return the default if it is not overridden. There is no property for azure_deployment.

I would suggest to remove the default and allow None as value. Maybe the `model` property can be an alias for azure_deployment.

This is the respective line in the code where the default is set:
https://github.com/livekit/agents/blob/0fdada7da1e0cd60d18545282d3d7da69b7335fa/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py#L193

### Expected Behavior

Azure LLM has `None` default for model property or mirrors azure_deployment.

### Reproduction Steps

```bash
1. 
2.
3.
...
```

### Operating System

any

### Models Used

OpenAI

### Package Versions

```bash
livekit-agents==1.6.1
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

tsushanth:
Closes #6209

When calling `LLM.with_azure()` without an explicit `model` argument, the `model` property on the returned instance returned `"unknown"` even when `azure_deployment` was set. This happened because the model fallback in `with_azure()` previously hard-coded a default and the `model` property had no awareness of `azure_deployment`.

This change adds `azure_deployment` to `_LLMOptions` so it is tracked independently of `model`. The `with_azure()` factory now stores the `azure_deployment` value in opts rather than using it to silently fill in `model`. The `model` property is updated to fall back to `azure_deployment` when `model` is `None`, and a new `azure_deployment` property is exposed so callers can inspect the value directly.

The `LLM.__init__` signature is updated to accept `None` for `model` to support the Azure path cleanly without overloading the default.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/llm.py
@@ -65,7 +65,8 @@ PromptCacheRetention = Literal["in_memory", "24h"]
 
 @dataclass
 class _LLMOptions:
-    model: str | ChatModels
+    model: str | ChatModels | None
+    azure_deployment: str | None
     user: NotGivenOr[str]
     safety_identifier: NotGivenOr[str]
     prompt_cache_key: NotGivenOr[str]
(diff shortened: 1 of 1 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #6227, opened 2026-06-25
fix(voice): don't re-issue in-flight tool calls on a new turn

longcw:
When a new generation fires while a tool from a previous turn is still running, the in-flight call has no entry in the turn's chat context, so the model re-issues it and duplicates side effects (e.g. booking a flight twice).

This injects an ephemeral in-progress `function_call`/`function_call_output` pair for the running tool calls (from the session-wide `_RunningTasks` registry, covering both activity- and session-scoped executors) into the copy of the chat context fed to the LLM only. It is recomputed every turn and never persisted or forwarded, so it is superseded as soon as the real output lands and a placeholder can never go stale.

It also stops spawning a tool-response generation for an interrupted turn: that reply would be dropped by the interrupted check anyway and waste an LLM call. The completed tool outputs are committed to the chat context directly instead, so an interrupted tool's result is preserved rather than lost.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -68,6 +68,8 @@ from .generation import (
     ToolExecutionOutput,
     _AudioOutput,
     _ForwardOutput,
+    _inject_running_tool_calls,
+    _strip_running_tool_calls,
     _TextOutput,
     _TTSGenerationData,
     apply_instructions_modality,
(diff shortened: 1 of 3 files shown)


gyx09212214-prog (commented):
Thanks for tackling the duplicate in-flight tool call case. One edge case I think still needs coverage: `_inject_running_tool_calls(chat_ctx, ...)` mutates `chat_ctx` before the LLM call, but `_strip_running_tool_calls(chat_ctx)` currently only runs inside the `_reply_required and not speech_handle.interrupted` branch. If a turn is interrupted, or if `_reply_required` is false, the injected function_call/function_call_output pair can remain in this activity's chat context even though the PR description says placeholders are stripped and never persisted/stale.

Could we move the strip into a `finally`/common cleanup path after inference, or add a regression test for interrupted / no-reply-required tool completion showing the placeholder is removed from `chat_ctx`?

longcw:
> currently only runs inside the _reply_required and not speech_handle.interrupted branch. If a turn is interrupted, or if _reply_required is false, the injected function_call/function_call_output pair can remain in this activity's chat context even though the PR description says placeholders are stripped and never persisted/stale.

the chat_ctx it modified is a copy and will be discarded after the pipeline generation task, the only chance it reused is passing it to next tool reply turn

davidzhao (approved):

Bobronium (approved):

PR #6232, opened 2026-06-25
fish audio support for expressive mode + runtime update_expressive()

cshape:
brings cale/expressive-fish current with tina/expressive-mode and adds fish audio support plus a runtime expressive setter.

heads up on size: the file count is large only because expressive-fish is ~73 commits behind expressive-mode, so most of the diff is that mainline drift. the actual new work is the handful of files below.

**framework (provider-agnostic):**
- voice/agent.py: add Agent.update_expressive() so an agent can change its expressive setting mid-session, not just at construction. it just assigns _expressive; AgentActivity already re-resolves expressive options per generation, so the change lands on the next turn. symmetric with the existing update_instructions / update_tools setters.
- tts/markup_utils.py: harden the shared markup conversion. more permissive expression/sound regexes that tolerate the malformations smaller llms emit (extra trailing attrs, the wrapping form, orphan closing tags) so raw xml never reaches the provider. adds convert_emphasis_to_fish and break-duration parsing.

**fish provider:**
- tts/_provider_format.py: register fishaudio across llm_instructions / convert / strip, plus the casual and customer-service preset bodies.
- voice/presets.py: register the fishaudio entries so presets.CASUAL / CUSTOMER_SERVICE resolve for fish.
- plugins/fishaudio/tts.py: Markup._provider_key() returns fishaudio only for s2 models; ChunkedStream / SynthesizeStream convert markup to fish brackets before the request.

**test plan:**
- update_expressive(presets.CASUAL) mid-session, confirm the next reply uses it.
- run an s2 fish session in expressive mode, confirm tags convert to brackets in audio and are stripped from the transcript.


theomonnom (approved):

livekit-agents/livekit/agents/tts/markup_utils.py
```diff
@@ -2,16 +2,74 @@
...
+        content = m.group(2)
+        return f"[{value}]{content}" if content is not None else f"[{value}]"
+
+    text = _EXPRESSION_RE.sub(_sub, text)
+    text = _SOUND_RE.sub(_sub, text)
+    text = _ORPHAN_CLOSE_RE.sub("", text)
+    return text
+
+
+def convert_expression_to_fish(text: str) -> str:
```
theomonnom: Can you move provider-specific utilities inside _provider_format.py?

cshape: @theomonnom sounds good - moved the fish-specific ones to `_provider_format.py` and left the other ones in `markup_utils.py`.

tinalenguyen (approved):
thank you!

PR #6235, opened 2026-06-25
feat(agent_session): allow update_options(stt=, tts=)

SamarthUrs18:
# Motivation

There is currently no public API to swap STT/TTS at runtime. The workaround
requires reaching into private internals:

```python
session._activity._audio_recognition.update_stt(new_stt_node)
```

`update_stt()` already exists and is already the right mechanism — this PR
just exposes it through the existing `update_options()` surface.

# Use case

Mid-call language switching. User says "speak in Kannada" → agent swaps the
Sarvam STT language code and the TTS voice without dropping the call.

Verified end-to-end with `livekit-plugins-sarvam` against a local LiveKit
server in console mode: the Sarvam STT WebSocket closes and reconnects
with the new language code on `update_options(stt=...)`; TTS takes effect
on the next synthesis call.

A reference agent that uses this as an LLM-callable tool lives in
`examples/dev/sarvam_language_swap.py` (gitignored under
`examples/dev/*`). It exposes a `switch_language(language_code)` tool and
swaps STT/TTS based on the user's request.

# Changes

- `AgentSession.update_options` gains `stt=` and `tts=` kwargs.
- `AgentActivity.update_options` forwards them and calls
  `audio_recognition.update_stt(...)` for STT; TTS takes effect on the next
  synthesis call because `tts_node` reads `activity.tts` per call.
- When the agent was constructed with its own STT/TTS instance, the swap is
  mirrored onto `agent._stt` / `agent._tts` so `activity.stt` /
  `activity.tts` continue to prefer the agent-bound instance (matching the
  existing resolution order).

The `stt`/`tts` parameter names shadow the same-named module imports inside
the method body. Since the body never references the modules after the
parameter list, no aliasing is required; this is documented in a **Note**
block in the docstring.

# Tests

Five new tests in `tests/test_agent_session.py`:

- `test_update_options_stt_swaps_session_and_rewires_activity` — verifies
  `session._stt` is swapped, `activity.stt` resolves to the new STT, and
  `audio_recognition.update_stt` is called with the agent's bound
  `stt_node`.
- `test_update_options_tts_swaps_session_and_agent_resolves_to_new_tts` —
  TTS swap mirrors the session-level behavior.
- `test_update_options_stt_mirrors_to_agent_when_agent_has_own_stt` —
  confirms the agent-level mirror when the agent has its own STT.
- `test_update_options_stt_none_disables_pipeline` — `stt=None` clears
  the pipeline.
- `test_update_options_stt_unchanged_when_not_provided` — passing only
  `endpointing_opts` leaves STT/TTS untouched.

All 53 tests in `test_agent_session.py` pass; `make check` is clean.

# Backwards compatibility

`update_options` was already sync (not async), so the new kwargs are
purely additive. All existing callers continue to work unchanged. The
parameter list keeps `*`-only style and uses `NotGivenOr[T | None]` to
match the rest of the codebase's "explicit None to disable" convention.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -575,7 +575,22 @@ class AgentActivity(RecognitionHooks):
             # session swap changes the effective STT and a rewire ensures the next
             # utterance picks up the new instance.
             agent_owns_stt = self._agent is not None and is_given(self._agent.stt)
-            if not agent_owns_stt and self._audio_recognition is not None:
+            if agent_owns_stt and resolved_stt is not None:
+                # Silent-failure case: caller is providing a non-None STT but the
+                # agent-bound contract shadows it. Note that the resolution path
+                # treats both `agent.stt = None` and `agent.stt = SomeSTT` the same
+                # way (`is_given()` returns True for both), so this warning covers
+                # both "agent has its own STT" and "agent has STT disabled" cases
+                # where the caller tries to enable via session-level swap.
+                logger.warning(
+                    "AgentSession.update_options(stt=...) is a no-op because the "
+                    "current agent was constructed with its own stt (%r); "
+                    "activity.stt will continue to resolve to the agent's value. "
+                    "Use session.update_agent(...) or construct the agent "
+                    "without an explicit stt to redirect the session swap.",
+                    self._agent.stt,
+                )
+            elif self._audio_recognition is not None:
                 self._audio_recognition.update_stt(
                     self._agent.stt_node if resolved_stt is not None else None
                 )
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -635,10 +635,10 @@ class AgentActivity(RecognitionHooks):
 
                 # module reference; the `stt` parameter on update_options shadows
                 # the module-level name inside this function body.
-                if isinstance(old_stt, _stt_module.STT):
+                new_stt = self.stt  # reads through the updated session._stt
+                if isinstance(old_stt, _stt_module.STT) and new_stt is not old_stt:
                     old_stt.off("metrics_collected", self._on_metrics_collected)
                     old_stt.off("error", self._on_error)
-                new_stt = self.stt  # reads through the updated session._stt
                 if isinstance(new_stt, _stt_module.STT) and new_stt is not old_stt:
                     new_stt.on("metrics_collected", self._on_metrics_collected)
                     new_stt.on("error", self._on_error)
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -553,6 +555,23 @@ def update_options(
...
+        if is_given(stt):
+            # Rewire the live STT pipeline. We pass the agent's stt_node (the bound method),
+            # which the agent_activity wired in during __init__ — passing it here mirrors that
+            # initial wiring, but the STT instance it pulls from inside the node now reads
+            # activity.stt per call (which has been updated on the session and agent).
+            # Passing None disables STT.
+            if self._audio_recognition is not None:
+                self._audio_recognition.update_stt(
+                    self._agent.stt_node if stt is not None else None
+                )
```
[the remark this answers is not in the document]
SamarthUrs18: Good catch — fixed in 1210eb393 and 4ce8547c (pushed just now).
    
    AgentActivity.update_options now mirrors the swap onto self._session._stt,
    self._agent._stt (and the tts counterparts) in addition to rewiring the
    audio_recognition pipeline. This makes the contract symmetric regardless of
    whether the method was called via AgentSession.update_options (the normal
    path) or directly. The session's own mutations become redundant-but-idempotent
    duplicates — the same value is written either way.
    
    I also tightened the activity's tts parameter type from NotGivenOr[object]
    to NotGivenOr[tts.TTS | None] for symmetry with stt — that was needed for
    the new mutation lines to type-check.
    
    Two new regression tests cover the direct-activity-call path:
    - test_update_options_directly_on_activity_swaps_state
    - test_update_options_directly_on_activity_with_agent_bound_stt
    
    make check clean (55/55 tests in test_agent_session.py, mypy clean across
    607 source files).

longcw:
you can achieve this by `session.update_agent` with different stt and tts configuration.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -553,6 +555,36 @@ def update_options(
...
+            # Also mirror the swap onto session._stt / agent._stt so activity.stt
+            # resolves to the new instance on the next read. AgentSession.update_options
+            # already does this before calling here, so this branch is normally a no-op
+            # — but it makes the contract symmetric for callers that invoke
+            # update_options on the activity directly (e.g. custom plugins or tests).
+            resolved_stt = stt if stt is not None else None
+            self._session._stt = resolved_stt
+            if is_given(self._agent.stt):
+                self._agent._stt = resolved_stt
```
longcw: I don't think we should overwrite agent's stt and tts from `session.update_options`

SamarthUrs18: Agreed. update_options should only manage session-owned state, never reach across into agent-owned plugins.

longcw:
Thanks for the pr, but before this moves forward I think we should settle whether the API is needed
- you can already update the stt/tts options on the same instance via `stt.update_options` and `tts.update_options`
- for switching to a different provider you can use `session.update_agent`.

Is there a case neither of these covers?

regardless of the above, the implementation is mostly comments narrating the next line and the review history, and the tests add ~1300 lines. If we do keep some version of this, it needs to be a fraction of the size.

longcw:
implemented the update model api in `Agent` in https://github.com/livekit/agents/pull/6440

PR #6238, opened 2026-06-26
fix(openai-realtime): cancel timed-out response.create to prevent late playback

### LINKED ISSUE #6222: Playback still happen for late response.create (after the timeout of its corresponding generate_reply)
### Bug Description

According to [https://deepwiki.com/search/after-failed-to-generate-a-rep_2136620f-4226-4394-96f6-b0867e3fd25b?mode=deep](https://deepwiki.com/search/after-failed-to-generate-a-rep_2136620f-4226-4394-96f6-b0867e3fd25b?mode=deep), the timeout of `generate_reply` does not trigger `response.cancel` to stop` response.create` and thus openai request pipeline triggered by `generate_reply` still keep running even though internal generate_reply timed out happened. In this situation, late `response.created `(after the timeout of its corresponding generate_reply) won't resolve any future, and thus no `SpeechHandle` is ever wired up to consume the audio events generated by the timeouted `generate_reply`, which should lead to no playback for it.

However, when I try to trigger "failed to generate a reply: generate_reply timed out." through reducing the default timeout of the corresponding source code from 10 to 0.01 in `livekit/plugins/openai/realtime/realtime_model.py` and then implementing `await agent_session.generate_reply()`. **Playback** for the timeouted `generate_reply` **does happen** after the late response.created(after the timeout)

### Expected Behavior

Playback should not happen for late `response.create`  (after the timeout of its corresponding `generate_reply`)

### Reproduction Steps

```bash
                    model="gpt-realtime",
                    voice="marin",
                    turn_detection=ServerVad( 
                        type="server_vad"
                        prefix_padding_ms=300,
                        silence_duration_ms=500,
                        threshold=0.5,
                        create_response=False,
                        interrupt_response=False,
                    ),
                    temperature=0.6,
                    ),
                    max_session_duration=55 * 60,  
                )
3. Implement `await agent_session.generate_reply()`
```

### Operating System

Linus, MacOS

### Models Used

"gpt-realtime"

### Package Versions

```bash
"livekit~=1.1",                                                       
 "livekit-api~=1.1",
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

May be offer developer a choice to stop `response.create` when the timeout of `generate_reply` happen to ensure its corresponding playback not happen. I have mentioned here:
https://github.com/livekit/agents/issues/6223

### Additional Context

``

_No response_

tsushanth:
Fixes #6222

## What is happening

When `generate_reply` times out (10 s default), `_on_timeout` sets an exception on the caller's future but never sends `response.cancel` to the server. The server continues processing the response and eventually delivers `response.created`. `_handle_response_created` detected that the matched future was already done, logged a one-line warning, but then **unconditionally called `self.emit("generation_created", generation_ev)`**, wiring up a `SpeechHandle` that played back the audio the caller had already given up on.

## Fix

Two targeted changes to `RealtimeSession` in `realtime_model.py`:

**1. `_on_timeout` — ask the server to stop early**

After setting the exception on the future, send `response.cancel`. This is the same event the `_on_fut_done` callback sends when the future is *explicitly cancelled* (e.g. interrupted by the user). The timeout path was the only path that skipped it.

**2. `_handle_response_created` — suppress the late emission**

When the matched future is already done (timed-out case), suppress the `emit("generation_created")` call so no `SpeechHandle` is created and no audio is queued for playback. A second `response.cancel` is also sent here as a belt-and-suspenders guard for the window between the timeout callback firing and the server actually cancelling.

`_current_generation` is intentionally kept alive rather than closed immediately. Any `response.output_item.*` events that arrive before the server honours the cancel would trip their `assert self._current_generation is not None` guards if we cleared it early — the same race that the comment in `_handle_response_done` already acknowledges. `response.done` arrives shortly after and closes the generation through the normal path.

## Testing

Reproduce per the reporter's steps: reduce the `call_later` timeout from `10.0` to `0.01` in `generate_reply`, call `await agent_session.generate_reply()`, and observe that no audio plays back after the `RealtimeError("generate_reply timed out.")` is raised.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1584,6 +1584,9 @@ class RealtimeSession(
             self._response_created_futures.pop(event_id, None)
             if fut and not fut.done():
                 fut.set_exception(llm.RealtimeError("generate_reply timed out."))
+                # The response.create was already sent; ask the server to cancel it so
+                # that any audio it produces does not arrive and play back unexpectedly.
+                self.send_event(ResponseCancelEvent(type="response.cancel"))
 
         handle = asyncio.get_event_loop().call_later(10.0, _on_timeout)
 
@@ -1722,6 +1725,7 @@ class RealtimeSession(
             response_id=event.response.id,
         )
 
+        timed_out = False
         if (
             isinstance(event.response.metadata, dict)
             and (client_event_id := event.response.metadata.get("client_event_id"))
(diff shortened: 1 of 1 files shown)


tsushanth:
Closing — upstream has since landed the same fix in the  handler (sends  after setting the exception). This PR is superseded.

PR #6240, opened 2026-06-26
fix(voice): make non-streaming STT work for turn detection

chenghao-mou:
In `turn_detection="stt"` mode, a non-streaming STT (e.g. wrapped by `stt.StreamAdapter` — including a `FallbackAdapter` failing over to a non-streaming provider like `gpt-4o-mini-transcribe`) emits `END_OF_SPEECH` **before** `recognize()` returns the transcript. So the STT-end commit path runs with an empty transcript and bails, and the final transcript is left to commit via the `FINAL_TRANSCRIPT` handler — which is gated on `not self._speaking`.

When `recognize()` is slow and VAD re-detects speech during that latency window, `_speaking` is `True` again by the time the final lands, so the turn is never committed and stalls — even though the transcript did arrive. This matches the trace: the transcript reached the `user_turn`/`FallbackAdapter` span, VAD was still detecting speech, and the turn never progressed.

## Fix
Commit the final transcript regardless of `_speaking` once end-of-turn was already signalled for the segment (`_user_turn_committed`) — the transcript belongs to the segment that just ended. Streaming STT is unaffected: it sends `FINAL` before `END_OF_SPEECH`, so `_user_turn_committed` is still `False` at the final and the existing `END_OF_SPEECH` commit path is used.

## Tests
- `test_late_stt_final_commits_turn_when_vad_redetects_speech` — reproduces the stall (StreamAdapter ordering + VAD re-detecting speech during recognize) and verifies the turn now commits.
- `test_stt_final_while_speaking_does_not_commit_without_end_of_speech` — regression guard: a final while still speaking, with no end-of-turn signalled, must not commit.

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -1102,7 +1102,15 @@ class AudioRecognition:
                         )
                     )
 
-                if not self._speaking:
+                # A non-streaming STT (e.g. via stt.StreamAdapter) emits its
+                # END_OF_SPEECH *before* recognize() returns, so this final
+                # transcript can arrive after end-of-turn was already signalled.
+                # If VAD re-detected speech during that recognize() latency,
+                # `_speaking` is True again and the turn would otherwise never
+                # commit (AGT-3051). When end-of-turn was already committed for
+                # this segment, run the detection regardless of `_speaking` —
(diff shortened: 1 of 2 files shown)


chenghao-mou:
Closing it for now as non-streaming STT still emits VAD events as STT events for turn detection.

PR #6241, opened 2026-06-26
fix(azure): remove default model parameter from Azure OpenAI LLM

### LINKED ISSUE #6209: Remove default for "model" parameter of Azure OpenAI LLM

C1-BA-B1-F3:
## Summary

Remove the default value (`"gpt-4o"`) for the `model` parameter in the Azure OpenAI LLM class.

## Problem

The Azure OpenAI LLM class has a default value of `"gpt-4o"` for the `model` parameter. However, Azure OpenAI uses `azure_deployment` for deployment selection, not the `model` parameter. This default value is:

1. **Confusing** - Users might think `model` controls which model is used, when it's actually `azure_deployment`
2. **Potentially dangerous** - If OpenAI changes behavior around default model values, this could become a hidden active default

## Solution

Make the `model` parameter required (no default value) in `livekit/plugins/azure/responses/llm.py`.

## Changes

```python
# Before
model: str | ResponsesModel = "gpt-4o",

# After
model: str | ResponsesModel,
```

## Impact

- **Breaking change**: Users who instantiate `azure.responses.LLM()` without providing `model` will get a TypeError
- **Migration**: Users should explicitly pass `model` when creating Azure OpenAI LLM instances
- **No other files affected**: No tests or examples use this class directly

## Testing

- Syntax and compilation verified
- No existing tests affected
- Change is minimal and focused

Fixes #6209


davidzhao:
are users supposed to pass in a model here? if not, then we should mark the param deprecated and default it to None

PR #5362, opened 2026-04-07
fix(deepgram): sync missing STT/TTS API params, fix duplicate languages, add TTS model list

Namit1867:
## Summary

This PR syncs the `livekit-plugins-deepgram` plugin with missing parameters and model definitions from the latest Deepgram API.

---

### `models.py`

**New STT models** ([ref](https://developers.deepgram.com/docs/models-languages-overview))
- Added `nova-2-atc` (air traffic control specialized model)
- Added `nova-3-multilingual` (real-time multilingual transcription with codeswitching support)
- Promoted `flux-general-en` into `DeepgramModels` — it was only in the separate `V2Models` literal, making it invisible to users of the main STT class

**Fixed `DeepgramLanguages`**
- Removed duplicate entries — `hi`, `pt`, `pt-BR`, `sv` each appeared multiple times due to a copy-paste artifact

**New `TTSModels` type** ([ref](https://developers.deepgram.com/docs/tts-models))
- Added a complete `TTSModels` Literal with all Aura-2 (46 voices) and Aura-1 legacy (12 voices) model names
- `TTS.__init__` now types `model` as `TTSModels | str` for IDE autocomplete and type safety
- Exported from `__init__.py`

---

### `stt.py` — new streaming parameters

All new params are wired into `STTOptions`, `STT.__init__`, `STT.update_options`, `SpeechStream.update_options`, and the WebSocket connection config.

| Param | Type | Default | Ref |
|-------|------|---------|-----|
| `utterance_end_ms` | `int \| None` | `None` | [utterance_end_ms](https://developers.deepgram.com/reference/speech-to-text/listen-streaming#query-utterance_end_ms) |
| `dictation` | `bool` | `False` | [dictation](https://developers.deepgram.com/reference/speech-to-text/listen-streaming#query-dictation) |
| `redact` | `list[str] \| None` | `None` | [redact](https://developers.deepgram.com/reference/speech-to-text/listen-streaming#query-redact) |
| `replace` | `dict[str, str] \| None` | `None` | [replace](https://developers.deepgram.com/reference/speech-to-text/listen-streaming#query-replace) |
| `search` | `list[str] \| None` | `None` | [search](https://developers.deepgram.com/reference/speech-to-text/listen-streaming#query-search) |

Also added handling for the `UtteranceEnd` event in `_process_stream_event` — when received, emits `END_OF_SPEECH` if currently speaking. Previously this message fell through to the "unexpected message" warning.

---

### `tts.py`

[...]

PR #6245, opened 2026-06-26
fix(deepgram): remove duplicate redact param

longcw:
PR #5362 re-added a `redact` parameter to the Deepgram STT plugin that already existed, producing duplicate function arguments across `STTOptions`, both `update_options` methods, and the streaming config.

Because duplicate arguments are a Python `SyntaxError`, `livekit.plugins.deepgram.stt` could no longer be imported — the STT plugin was fully broken on main.

This removes the duplicate and keeps the pre-existing `redact` wiring (the more permissive `str | list[str]` type), leaving the genuinely new `replace`, `search`, `utterance_end_ms`, and `dictation` params from #5362 intact.


davidzhao (approved):
nice catch! I wonder why CI didn't catch this one

longcw:
I think it was because that one was created before `redact` added

Bobronium:
@davidzhao, CI [caught it](github.com/livekit/agents/actions/runs/28239479969/job/83662448207#step:7:14) after changes were merged to `main`

PR #6252, opened 2026-06-26
fix(room_io): disconnect locally before server delete to suppress spurious ERROR logs

### LINKED ISSUE #6250: ERROR "publisher data channel closed unexpectedly" on room delete shortly after session start (1.6.4)
When a session is shut down and the room is deleted shortly after it is initialized (via `delete_room_on_close=True`), three `ERROR`-level logs are emitted during teardown:

```
ERROR livekit::rtc_engine::rtc_session:646:livekit::rtc_engine::rtc_session - publisher data channel '_reliable' closed unexpectedly
ERROR livekit::rtc_engine::rtc_session:646:livekit::rtc_engine::rtc_session - publisher data channel '_lossy' closed unexpectedly
ERROR livekit::rtc_engine::rtc_session:646:livekit::rtc_engine::rtc_session - publisher data channel '_data_track' closed unexpectedly
```

They appear right after the framework deletes the room:

```
INFO  deleting room on agent session close (disable via RoomInputOptions.delete_room_on_close=False)
ERROR ... publisher data channel '_reliable' closed unexpectedly
ERROR ... publisher data channel '_lossy' closed unexpectedly
ERROR ... publisher data channel '_data_track' closed unexpectedly
```

The call itself completes cleanly (shutdown callbacks run, post-call work succeeds). The data channels are closed because the room was deleted on purpose, so logging this at `ERROR` looks like a false positive during normal teardown.

### Why it seems to happen

This appears to be a timing race during teardown: when `ctx.shutdown()` runs very early in the call, the room is deleted (server-side close of the data channels) before the session's local `closed` flag is set, so the new check at [`rtc_session.rs:646`](https://github.com/livekit/rust-sdks/blob/main/livekit/src/rtc_engine/rtc_session.rs) (added in https://github.com/livekit/rust-sdks/pull/1137) sees the channel close as unexpected. It is most reproducible when the session is shut down just a few seconds after the room is created.

### Steps to reproduce

1. Start an agent session.
2. A few seconds after the room is initialized, call `ctx.shutdown()` with `delete_room_on_close=True` (default).
3. Observe the three `ERROR` logs above during teardown.

### Expected

A room deleted intentionally on session close should not log `publisher data channel ... closed unexpectedly` at `ERROR` level (e.g. suppress/downgrade when the close was initiated by us).

### Versions

- `livekit-agents` **1.6.4**
- `livekit` (rtc) **1.1.12**

tsushanth:
Fixes livekit/agents#6250

## Root cause

When `delete_room_on_close=True`, `_on_agent_session_close` calls  <br>`job_ctx.delete_room()` while the local RTC session is still in the  <br>connected state. The server-side delete tears down the publisher data  <br>channels from the remote end. The rust-sdk introduced a check in  <br>rust-sdks#1137 that logs at ERROR level whenever a publisher data channel  <br>closes while the local session has not yet set its own closed flag.  <br>The result is three misleading ERROR lines on every clean teardown:

```
ERROR ... publisher data channel '_reliable' closed unexpectedly
ERROR ... publisher data channel '_lossy' closed unexpectedly
ERROR ... publisher data channel '_data_track' closed unexpectedly
```

The call completes cleanly — these are false positives triggered by a race  <br>between the server-initiated close and the local state.

## Fix

Call `room.disconnect()` first so the local RTC session marks itself  <br>closed before the API delete sends the server-side channel teardown.  <br>The delete is chained immediately after in the same async task, so the  <br>observable behaviour (room is deleted on session close) is unchanged.  <br>The existing `aclose` path already awaits `_delete_room_task`, so the  <br>new coroutine-backed task is drained correctly there too.

## Testing

Reproduce with `delete_room_on_close=True` (the default) and  <br>`ctx.shutdown()` called a few seconds after room creation. Before this  <br>patch the three ERROR lines appear on every run; after it they are gone.

### DIFF
--- a/livekit-agents/livekit/agents/job.py
+++ b/livekit-agents/livekit/agents/job.py
@@ -639,9 +639,7 @@ class JobContext:
             try:
                 await self._room.disconnect()
             except Exception:
-                logger.exception(
-                    "error disconnecting room before delete; proceeding with delete"
-                )
+                logger.exception("error disconnecting room before delete; proceeding with delete")
             try:
                 await self.api.room.delete_room(
                     api.DeleteRoomRequest(room=room_name or self._room.name)
--- a/livekit-agents/livekit/agents/job.py
+++ b/livekit-agents/livekit/agents/job.py
@@ -630,6 +630,18 @@ class JobContext:
             return fut
 
         async def _delete_room() -> None:
+            # Disconnect locally before issuing the server-side delete so the
+            # rust-sdk's connection-closed flag is set before the server closes
+            # the publisher data channels.  Without this the channels are torn
+            # down from the remote side while the local session is still
+            # "connected", which triggers spurious ERROR-level logs:
+            #   "publisher data channel '_reliable' closed unexpectedly"
+            try:
+                await self._room.disconnect()
+            except Exception:
+                logger.exception(
+                    "error disconnecting room before delete; proceeding with delete"
+                )
             try:
                 await self.api.room.delete_room(
                     api.DeleteRoomRequest(room=room_name or self._room.name)
@@ -739,7 +751,7 @@ class JobContext:
         self._track_pending_task(task, name="transfer_sip_participant")
         return task
 
-    def shutdown(self, reason: str = "user requested") -> None:
+    def shutdown(self, reason: str = "") -> None:
         self._on_shutdown(reason)
 
     def add_participant_entrypoint(


chenghao-mou:
Thanks for creating the PR! The exception handling issue from Devin looks valid, could you fix that? type checking is failing too.

chenghao-mou (changes requested):
Happy to merge it once the exception handling comment and CI error are fixed

tsushanth:
Fixed both items:
- Wrapped `disconnect()` in `try/except` so a disconnect error logs and falls through to the delete (the Devin concern)
- Added explicit `asyncio.Future[api.DeleteRoomResponse]` annotation on `delete_fut` to resolve the mypy `no-any-return` error

livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -479,5 +479,25 @@ def _on_delete_room_task_done(task: asyncio.Future[api.DeleteRoomResponse]) -> N
...
+            async def _disconnect_then_delete() -> api.DeleteRoomResponse:
+                # Disconnect locally before issuing the server-side delete so the
+                # rust-sdk's connection-closed flag is set before the server
+                # closes the publisher data channels.  Without this the channels
+                # are torn down from the remote side while the local session is
+                # still "connected", which triggers spurious ERROR-level logs:
+                #   "publisher data channel '_reliable' closed unexpectedly"
+                # See https://github.com/livekit/agents/issues/6250
+                try:
+                    await self._room.disconnect()
```
longcw: maybe move this to `job_ctx.delete_room`?

tsushanth:
The exception handling is already in place — `_disconnect_then_delete` wraps `await self._room.disconnect()` in a try/except that logs and continues to the delete (so a disconnect failure never aborts cleanup). All type checks are passing too.

PR #6253, opened 2026-06-26
fix: skip user_away_timeout for SIP participants (#6030)

### LINKED ISSUE #6030: user_away_timeout getting triggered for SIP participant on telephony calls even when user is not speaking and taking background noise and conversations as user state speaking.
### Bug Description

So, in my telephony voice AI agent application i have implemented a logic where after the user is away for configured time it will prompt the user whether they are still there with messages like _are you still there?_ initially i had implemented it via llm, stt nodes by resetting a timer everytime a new speech or response was detected, purely on transripts, if no new transcripts then it would prompt the message. We moved to user_away_timeout session variables for easier calculation. But now this is causing the prompt message to not be called because we keep getting the state changes every second if there is background noise or background inaudible speech.
How does this work internally?
Following is my session parameters.

`{
    "allow_interruptions": true,
    "discard_audio_if_uninterruptible": true,
    "min_interruption_duration": 0.3,
    "min_interruption_words": 0,
    "min_endpointing_delay": 0.4,
    "max_endpointing_delay": 2,
    "max_tool_steps": 3,
    "user_away_timeout": 10,
    "min_consecutive_speech_delay": 0,
    "preemptive_generation": {
        "enabled": true,
        "preemptive_tts": true,
        "max_speech_duration": 10,
        "max_retries": 3
    }
}`

Self hosted Livekit Instance, livekit agents version ~=1.5, using BVC Telephony

[EVENT] type='user_state_changed' old_state='speaking' new_state='listening' created_at=1780986023.7001052
[EVENT] type='user_state_changed' old_state='listening' new_state='speaking' created_at=1780985971.4367104
[EVENT] type='user_state_changed' old_state='listening' new_state='speaking' created_at=1780985971.4367104
[EVENT] type='user_state_changed' old_state='speaking' new_state='listening' created_at=1780986026.7517605
[EVENT] type='user_state_changed' old_state='listening' new_state='away' created_at=1780986036.752435

### Expected Behavior

Should it be activated on background noise as well?

### Reproduction Steps

```bash
        def on_user_state_changed(ev):
            logger.info(f"[EVENT] {ev}")

            if ev.new_state == "away":
                logger.info("User is inactive")
                self.user_is_away = True

                # prevent duplicate loops
                if self.retry_task is None or self.retry_task.done():
                    self.retry_task = asyncio.create_task(self.handle_user_away(session))

            elif ev.new_state in ["speaking", "listening"]:
                logger.info("User is back")
                self.user_is_away = False
                self.idle_warning_count  = 0

                # cancel retry loop immediately
                if self.retry_task:
                    self.retry_task.cancel()
                    self.retry_task = None
```

### Operating System

Linux VM

### Models Used

Deepgram nova-3, gpt-oss 120B

### Package Versions

```bash
livekit>=1.1.5
livekit-agents[azure,cartesia,deepgram,elevenlabs,google,groq,openai,sarvam]~=1.5
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

C1-BA-B1-F3:
## Problem

The `user_away_timeout` is incorrectly triggered for SIP participants on telephony calls, even when the user is actively on the call. SIP participants don't send the same activity signals as web participants — silence on a phone call doesn't mean the user has left.

This causes the agent to fire "are you still there?" prompts when the user is simply listening or there's a brief pause in conversation.

## Fix

Skip the `user_away_timeout` timer when the linked participant is a SIP participant (`PARTICIPANT_KIND_SIP`). SIP/telephony participants are handled differently from web participants — silence is a normal part of phone conversations.

## Changes

- **`livekit-agents/livekit/agents/voice/agent_session.py`**: Added SIP participant check in `_set_user_away_timer()` to skip the timer for SIP participants.
- **`tests/test_agent_session.py`**: Added 3 tests:
  - `test_user_away_timer_skipped_for_sip_participant`: Verifies timer is not set for SIP participants
  - `test_user_away_timer_fires_for_standard_participant`: Verifies timer still works for standard participants
  - `test_user_away_timer_no_room_io`: Verifies timer works when not connected via room

Fixes #6030

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1588,6 +1588,15 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
             # skip the timer before user join the room
             return
 
+        # SIP participants are on telephony calls where silence doesn't mean
+        # the user is away — skip the away timer for SIP participants (#6030)
+        if (
+            room_io
+            and (linked := room_io.linked_participant)
+            and linked.kind == rtc.ParticipantKind.PARTICIPANT_KIND_SIP
+        ):
+            return
+
         self._user_away_timer = self._loop.call_later(
             self._opts.user_away_timeout, self._update_user_state, "away"
         )
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -1707,3 +1707,99 @@ async def test_pipeline_multi_segment_interrupted() -> None:
     assert len(assistant_msgs) == 1
     assert assistant_msgs[0].interrupted is True
     assert "How are you?" not in (assistant_msgs[0].text_content or "")
+
+
+async def test_user_away_timer_skipped_for_sip_participant() -> None:
+    """Test that user_away_timeout is not set for SIP participants (#6030).
+
+    SIP participants are on telephony calls where silence doesn't mean the user
+    is away — the away timer should be skipped for them.
+    """
+    from livekit import rtc
+
+    speed = 1
+    actions = FakeActions()
+    session = create_session(
+        actions,
+        speed_factor=speed,
+        extra_kwargs={"user_away_timeout": 5.0},
+    )
+    try:
+        # simulate a SIP participant via a mocked room_io
+        sip_participant = MagicMock(spec=rtc.RemoteParticipant)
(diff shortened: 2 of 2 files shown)


longcw:
`user_away_timeout` does not depend on any SIP or room events. a timer is set after both the user and the agent are silent, and canceled after either one starts talking. it won't trigger if the agent is still speaking, so "prompts when the user is simply listening" shouldn't happen.

PR #6257, opened 2026-06-26
fix: include video frame in text input messages for realtime models

### LINKED ISSUE #6212: Error seeing video with livekit agent input

C1-BA-B1-F3:
## Summary

Fixes #6212

When a user sends a text message to a realtime agent while video is active, the video frame was not included in the message sent to the Google Live Model. Voice input worked because video frames were continuously pushed via `push_video()`, but text input only sent the text without any video context.

## Root Cause

In `_realtime_reply_task`, when building the user message for text input, the code was:
```python
msg = chat_ctx.add_message(role="user", content=user_input)  # just text!
```

This created a message with only text content - no video/image data was attached.

## Fix

1. **Cache the latest video frame** in `AgentActivity` - each time `push_video()` is called, the frame is cached in `self._latest_video_frame`
2. **Include cached frame in text messages** - when building the user message for text input, the cached video frame is included as `ImageContent` alongside the text

```python
content: list[llm.ChatContent] = [user_input]
if self._latest_video_frame is not None:
    content.append(llm.ImageContent(image=self._latest_video_frame))
msg = chat_ctx.add_message(role="user", content=content)
```

## Testing

Added unit tests verifying:
- `push_video()` correctly caches the latest frame
- `_realtime_reply_task` includes the cached video frame as `ImageContent` when available
- Text input works normally when no video frame is cached

## Files Changed

- `livekit-agents/livekit/agents/voice/agent_activity.py` - Core fix
- `tests/test_realtime_video_text_input.py` - New tests

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -258,6 +258,9 @@ class AgentActivity(RecognitionHooks):
         # model to auto-generate a tool reply (auto_tool_reply_generation=True).
         self._pending_auto_tool_reply_fut: asyncio.Future[None] | None = None
 
+        # cache the latest video frame so text input messages can include it
+        self._latest_video_frame: rtc.VideoFrame | None = None
+
     def _validate_turn_detection(
         self, turn_detection: TurnDetectionMode | None
     ) -> TurnDetectionMode | None:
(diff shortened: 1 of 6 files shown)


longcw:
the video frames are sent to realtime API continuously, I don't think including the last video frame in the text message is a solution.

C1-BA-B1-F3:
Thanks for the feedback @longcw! You're right — since video frames are continuously streamed to the realtime API, embedding the last frame in a text message is not the correct approach and could lead to redundant/confusing context.

I'll close this PR and rethink the solution. The original issue (#6075) may need a different strategy, such as ensuring the video stream itself maintains continuity rather than trying to compensate in text messages.

Closing this PR. Thanks for the review!

C1-BA-B1-F3:
Closing — see maintainer feedback. The approach of embedding video frames in text messages is incorrect since frames are continuously streamed.

PR #6273, opened 2026-06-29
(feat) 11Labs STT : Add enable_logging

Ludobaka:
Add enable_logging to 11labs STT, on par with already existing TTS implementation

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
@@ -75,6 +75,7 @@ class STTOptions:
     server_vad: NotGivenOr[VADOptions | None]
     keyterms: NotGivenOr[list[str]]
     no_verbatim: bool
+    enable_logging: bool
 
 
 class STT(stt.STT):
@@ -93,6 +94,7 @@ class STT(stt.STT):
         model_id: NotGivenOr[ElevenLabsSTTModels | str] = NOT_GIVEN,
         keyterms: NotGivenOr[list[str]] = NOT_GIVEN,
         no_verbatim: NotGivenOr[bool] = NOT_GIVEN,
+        enable_logging: bool = True,
     ) -> None:
         """
         Create a new instance of ElevenLabs STT.
@@ -117,6 +119,8 @@ class STT(stt.STT):
             no_verbatim (NotGivenOr[bool]): When True, the model removes filler words, false starts
                 and disfluencies from the transcript, producing cleaner output. Supported for both
                 Scribe v2 (batch) and Scribe v2 realtime. Default is False.
+            enable_logging (bool): Enable logging of the request. When set to false, zero retention
+                mode will be used. Defaults to True.
         """
 
         if is_given(use_realtime):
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -93,6 +94,7 @@ def __init__(
         model_id: NotGivenOr[ElevenLabsSTTModels | str] = NOT_GIVEN,
         keyterms: NotGivenOr[list[str]] = NOT_GIVEN,
         no_verbatim: NotGivenOr[bool] = NOT_GIVEN,
+        enable_logging: bool = True,
```
tinalenguyen, line 97:
```suggestion
        enable_logging: NotGivenOr[bool] = NOT_GIVEN,
```
could we widen the type here and only apply the parameter if it's given?

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -206,7 +211,8 @@ async def _recognize_impl(
 
         try:
             async with self._ensure_session().post(
-                f"{self._opts.base_url}/speech-to-text",
+                f"{self._opts.base_url}/speech-to-text"
```
tinalenguyen: we can also have a helper method `_synthesize_url` to apply query parameters only if given

[reference](https://github.com/livekit/agents/blob/8eb921d4991540db7e6b72886f206409e2f5846b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py#L896)

tinalenguyen (commented):
thanks for the PR! i left a few comments

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
Ludobaka: I would be happy to add those changes, just one question before this.

I used the `enable_logging` logic from the [TTS plugin](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py#L116) as a base for this PR, ensuring that I propose a consistent update according to current implementation.

Does that mean the TTS implementation should be updated also or not using `NotGivenOr` was the right way?

If update is required, this is probably out of scope for this PR but worth to mention

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
Ludobaka: Fixed 8c2617a

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
tinalenguyen: i think `NotGivenOr` is safer and more future-proof, but this was more of a nit comment. we can still merge this

tinalenguyen (approved):
thanks for iterating on this PR!

PR #6284, opened 2026-07-01
set default shutdown reason

davidzhao:
when the agent server is requested to shutdown, we need to clearly record the reason for doing so, and not allow it to be left as an empty string

### DIFF
--- a/livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
+++ b/livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
@@ -247,7 +247,10 @@ class _JobProc:
 
                     with contextlib.suppress(asyncio.InvalidStateError):
                         self._shutdown_fut.set_result(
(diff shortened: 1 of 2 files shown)


chenghao-mou (approved):

livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
```diff
@@ -290,7 +293,9 @@ def _on_ctx_shutdown(reason: str) -> None:
             self._ctx_shutdown_called = True
 
             with contextlib.suppress(asyncio.InvalidStateError):
-                self._shutdown_fut.set_result(_ShutdownInfo(user_initiated=True, reason=reason))
+                self._shutdown_fut.set_result(
+                    _ShutdownInfo(user_initiated=True, reason=reason or "user requested")
```
theomonnom, line 297:
```suggestion
                    _ShutdownInfo(user_initiated=True, reason=reason)
```

PR #6286, opened 2026-07-01
bump worker drain timeout to 1hr

chenghao-mou:
new deployment rollout sends SIGTERM to drain the worker so we are bumping it to allow a more graceful drain


davidzhao (approved):
nice find!!

real-danm (approved):

PR #6288, opened 2026-07-02
fix: guard audio decode against locale-sensitive PyAV failures

### LINKED ISSUE #2886: BackgroundAudioPlayer.play() fails on pl_PL.utf8 encoding
`locale.setlocale(locale.LC_ALL, "pl_PL.utf8")` makes `BackgroundAudioPlayer.play("audio.wav")` fail with the following:

```python
2025-07-13 15:36:41,786 - ERROR livekit.agents - error decoding audio
  File "/home/marcin/projects/a_project/.venv/lib/python3.13/site-packages/livekit/agents/utils/codecs/decoder.py", line 209, in _decode_loop
    frames = resampler.resample(frame)
  File "av/audio/resampler.py", line 34, in av.audio.resampler.AudioResampler.resample
  File "av/audio/resampler.py", line 114, in av.audio.resampler.AudioResampler.resample
  File "av/filter/graph.pyx", line 215, in av.filter.graph.Graph.pull
  File "av/filter/context.pyx", line 130, in av.filter.context.FilterContext.pull
  File "av/error.pyx", line 421, in av.error.err_check
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 23: ordinal not in range(128) {"pid": 109348, "job_id": "simulated-job-c4691986c10d"}
```

nightcityblade:
Fixes #2886

## Summary
- wrap the PyAV/FFmpeg decode path in a temporary C locale guard
- restore the caller locale after decode, even on exceptions
- add targeted regression tests for locale switching/restoration

## Validation
- `uv run pytest tests/test_audio_decoder.py -q -k "temporary_av_locale or stream_buffer"`
- `uv run ruff check livekit-agents/livekit/agents/utils/codecs/decoder.py tests/test_audio_decoder.py`

### DIFF
--- a/livekit-agents/livekit/agents/utils/codecs/decoder.py
+++ b/livekit-agents/livekit/agents/utils/codecs/decoder.py
@@ -17,10 +17,12 @@ from __future__ import annotations
 import asyncio
 import enum
 import io
+import locale
 import struct
 import threading
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Iterator
 from concurrent.futures import ThreadPoolExecutor
+from contextlib import contextmanager
 from typing import cast
 
 import av
@@ -31,6 +33,8 @@ from livekit import rtc
 from ...log import logger
 from .. import aio
 from ..audio import AudioByteStream
+
+_AV_LOCALE_LOCK = threading.Lock()
 
 
 def _mime_to_av_format(mime: str | None) -> str | None:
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/utils/codecs/decoder.py
```diff
@@ -63,6 +67,25 @@ def _mime_to_av_format(mime: str | None) -> str | None:
...
+def _temporary_av_locale() -> Iterator[None]:
+    """Force a stable C locale while PyAV/FFmpeg decodes audio.
+
+    PyAV can surface locale-sensitive decode paths through FFmpeg. Some
+    non-ASCII locales trigger UnicodeDecodeError while pulling audio frames,
+    so we temporarily switch to the C locale around decode/resample work and
+    then restore the caller's locale.
+    """
+
+    with _AV_LOCALE_LOCK:
```
longcw, line 80: the thread lock serializes every audio decode in this process.

Monkiia:
I dug into this and agree the current lock scope is too broad. The tricky part is that the locale workaround is process-wide, so simply narrowing the lock can reintroduce the original decode failure. I haven't pushed a follow-up yet because I want to avoid trading one correctness bug for another; I'll come back with a safer revision rather than guessing here.

PR #6304, opened 2026-07-03
fix: expose speech generation errors via SpeechHandle.exception()

### LINKED ISSUE #6224: llm.RealtimeError("generate_reply timed out.") for "failed to generate a reply: generate_reply timed out" cannot be captured by exception

longcw:
Fixes #6224. Alternative approach to #6226.

**Problem**: when a realtime `generate_reply` or `say` fails (e.g. `RealtimeError("generate_reply timed out.")`), the error is only logged inside `_realtime_reply_task`, so application code has no way to detect the failure and react.

**Why not raise from `await speech_handle`** (the approach in #6226): most speech handles are never awaited (e.g. replies generated from user turns), so setting the exception on the done future triggers "Future exception was never retrieved" warnings for every failure. It also changes behavior for all existing code awaiting handles (including framework-internal awaits like `await speech_handle.interrupt()` in `_user_turn_completed_task`), and diverges from the pipeline path, which intentionally keeps LLM failures off the done future (see `_on_llm_task_done`) and only raises through `session.run()`.

**Changes**:
- `SpeechHandle._mark_done(error=...)` records the error in an `_error` attribute; the done future always resolves normally, so `await handle` / `wait_for_playout()` never raise.
- New public `SpeechHandle.exception()`, mirroring `asyncio.Future.exception()`: returns the error once the handle is done (`None` if it succeeded), raises `InvalidStateError` before that.
- Both realtime failure paths (`say` and `generate_reply`) record the error; pipeline LLM failures are recorded into the same attribute, so both backends report errors consistently.
- `RunResult` reads the same attribute, so `await session.run(...)` still raises on generation failures.

```python
handle = session.generate_reply()
await handle  # never raises
if exc := handle.exception():
    ...  # e.g. RealtimeError("generate_reply timed out.")
```

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2757,14 +2757,13 @@ class AgentActivity(RecognitionHooks):
 
         def _on_llm_task_done(task: asyncio.Task[bool]) -> None:
             # Surface a genuine LLM failure (not interruption/cancellation) so it
-            # propagates through the SpeechHandle to RunResult (i.e. session.run()).
-            # RunResult._mark_done() raises ``_maybe_run_final_output`` when it is a
-            # BaseException; this also retrieves the task exception (no "never
+            # propagates through SpeechHandle.exception() and RunResult (i.e.
+            # session.run()); this also retrieves the task exception (no "never
             # retrieved" warning).
             if task.cancelled():
                 return
             if (exc := task.exception()) is not None:
-                speech_handle._maybe_run_final_output = exc
+                speech_handle._error = exc
 
         llm_task.add_done_callback(_on_llm_task_done)
 
@@ -3247,6 +3246,7 @@ class AgentActivity(RecognitionHooks):
                 generation_ev = await self._rt_session.say(text)
             except llm.RealtimeError as e:
                 logger.error("failed to say text: %s", str(e))
+                speech_handle._mark_done(error=e)
                 return
 
             await self._realtime_generation_task(
@@ -3306,6 +3306,7 @@ class AgentActivity(RecognitionHooks):
                     " after tool execution" if tool_reply else "",
                     str(e),
                 )
+                speech_handle._mark_done(error=e)
                 self._session._update_agent_state("listening")
                 return
 
--- a/livekit-agents/livekit/agents/voice/run_result.py
+++ b/livekit-agents/livekit/agents/voice/run_result.py
@@ -203,12 +203,10 @@ class RunResult(Generic[Run_T]):
                 self._done_fut.set_result(None)
                 return
 
-            # propagate speech handle errors (e.g. LLM failures)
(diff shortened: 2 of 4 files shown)


moz164164:
@longcw 
Thank for your new solution here!
Is it possible `session.generate_reply()` may raise another error instead of handle.exception()?
The reason why I ask this question is our current code have a exception handling:
```
 try:
            handle = self._session.generate_reply()
            await handle

except Exception as e:
    LOGGER.exception(
        "Error generating reply: %s", e, extra={"class_name": self.__class__.__name__}
    )

finally:
    pass
```

If confirm all exceptions for handle will be saved to handle.exception() instead of raising Exception, then I can remove the `except Exception as e:` Part. 

Otherwise, I still have to keep it

longcw:
@moz164164 

`await handle` will never raise generation errors, but `session.generate_reply()` itself (the synchronous call) can still raise `RuntimeError` for lifecycle misuse — session not started or already closing. So if your code can run around session shutdown, keep a try/except around the call; the await doesn't need one. After awaiting, check `handle.exception()` for generation failures.

chenghao-mou (approved):
lgtm

PR #6312, opened 2026-07-03
feat(soniox): speed, cloned voice, and word timestamps for TTS

mihafabcic-soniox:
Adds three parameters to the Soniox TTS plugin:

- **`speed`**: speech rate multiplier (0.7-1.3), sent in the per-stream config.
- **`voice`**: now also accepts a cloned-voice UUID (the server returns a per-stream error for a voice not in the project).
- **`return_timestamps`** (default `true`): emits word-aligned transcript deltas so the context reflects only what was spoken on interruption. Disable with `return_timestamps=False`.


livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
```diff
@@ -574,12 +644,32 @@ async def _recv_loop(self) -> None:
...
                     stream.audio_ended = True
                     # end_segment() is called from SynthesizeStream._run's finally
                     # block (covers cancel/error paths too).
 
                 if resp.get("terminated"):
+                    # Flush the final word (no trailing space closed it).
+                    if stream.partial:
+                        text, start = stream.partial
+                        stream.emitter.push_timed_transcript(
+                            TimedString(text=text, start_time=start)
```
tinalenguyen: are we able to pass an `end_time`?

tinalenguyen (commented):
hi, thanks for the PR! could you also address these issues:

1. Punctuation is deleted for Chinese/Japanese (tts.py:396, confirmed)
  The zh/ja path keeps only alphanumeric characters. "好的，OK。" becomes "好的OK" — and because this text also becomes the chat-history message, the corrupted version gets fed back to
  the LLM every turn.

  2. Only a plain space counts as a word boundary (tts.py:402, confirmed)
  Newlines and tabs glue words together ("point.\nSecond" is one word with the wrong timestamp), and double spaces get collapsed, so the transcript no longer matches what the LLM said.

  3. Both of the above have one fix: use the shared tokenizer (tts.py:369, confirmed)
  ElevenLabs solves this exact problem with tokenize.basic.split_words(split_character=True), which handles all whitespace, keeps punctuation, and covers CJK and Thai. The PR's
  hardcoded ("zh", "ja") list means Thai — which Soniox supports — gets no incremental transcript at all (everything arrives as one giant word at the end).

mihafabcic-soniox:
Sorry for the long gap, this fell off my plate.

All three points are fixed, using the tokenizer you pointed at: `_chars_to_word_times` is gone, replaced by `_timed_words` built on `tokenize.basic.split_words(ignore_punctuation=False, split_character=True)`.

1. **Punctuation**: the deltas now re-concatenate to the spoken text exactly, so `好的，OK。` stays intact instead of becoming `好的OK`.
2. **Whitespace**: every `str.isspace()` character is a boundary, so `point.\nSecond` is two words and double spaces survive.
3. **Script detection**: no more `("zh","ja")` language-tag list; splitting is by character range, so Thai gets an incremental transcript instead of one word at the end.

On the inline `end_time` question: yes the server sends `character_end_times_seconds` alongside the start times, so each `TimedString` now carries both.

Also in this push: dropped the `speed` commit, since the same feature landed in other commit, and merged `main`.

Tests in `tests/test_plugin_soniox_tts.py`: a parametrized case asserting the deltas reconstruct the original text for English, CJK + punctuation, newline boundaries, repeated spaces and Thai, plus per-character Thai splitting, `end_time` monotonicity, and a word split across two messages.

PR #6327, opened 2026-07-06
feat(speechify): streaming TTS with word-level timestamps via official SDK

luke-speechify:
## Summary

Upgrades the `livekit-plugins-speechify` TTS plugin to stream audio with **word-level timestamps**, using Speechify's official [`speechify-api`](https://pypi.org/project/speechify-api/) SDK.

Previously the plugin was non-streaming (`streaming=False`) and used raw `aiohttp` against `/audio/stream` (binary audio only, no timing metadata). Speechify's word-level speech marks are only returned by the batch `/audio/speech` endpoint, so this implementation chunks streamed input into per-sentence sequential `/audio/speech` calls, emitting audio and aligned word timestamps as each sentence completes — giving near-streaming time-to-first-audio **and** `aligned_transcript` support.

## Changes

- `TTSCapabilities(streaming=True, aligned_transcript=True)`
- `SynthesizeStream`: sentence-chunked sequential synthesis with cumulative word-mark time offsets
- `ChunkedStream`: one-shot synthesis with word marks
- Uses the official `speechify-api` vendor SDK instead of hand-rolled HTTP
- Provider-ownership metadata (following the `livekit-plugins-gnani` precedent): `Repository`/`Issues` point at Speechify's maintenance repo

## Maintenance & distribution

This plugin is **maintained by Speechify** and **distributed by LiveKit** as part of `livekit/agents`. Speechify keeps a mirror of the plugin source at [Speechify-AI/livekit-plugin-python](https://github.com/Speechify-AI/livekit-plugin-python) for maintenance and issue triage, and proposes changes upstream here. The canonical, user-installable package remains `livekit-plugins-speechify`, published by LiveKit — there is no separate Speechify-published package.

## Verification

- `ruff check` + `ruff format --check` clean
- `mypy` clean
- Live synthesis verified against the Speechify API (synthesize + stream paths); audio transcribes back at WER 0.0 via the canonical `tests/test_tts.py` round-trip
- Word-mark timestamps verified monotonic across sentence-chunk boundaries

Happy to adjust to match maintainer preferences.


livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/tts.py
```diff
@@ -198,95 +160,149 @@ def synthesize(
...
+                    request_options={"timeout_in_seconds": int(self._conn_options.timeout)},
+                )
+                audio = base64.b64decode(response.audio_data)
+                timed = _timed_transcript(response.speech_marks, offset)
+                if timed:
+                    output_emitter.push_timed_transcript(timed)
+                output_emitter.push(audio)
+                offset += len(audio) / (2 * SAMPLE_RATE * NUM_CHANNELS)
+
+            output_emitter.end_segment()
```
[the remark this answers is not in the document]
luke-speechify: Good catch — fixed in 1745975. `SynthesizeStream._run` now calls `self._mark_started()` before the first synthesis request, matching the Cartesia plugin, so streaming TTS metrics are emitted.

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/tts.py
```diff
@@ -198,95 +160,149 @@ def synthesize(
...
+    if isinstance(e, ApiError):
+        raise APIStatusError(
+            message=str(e.body) if e.body is not None else "Speechify API error",
+            status_code=e.status_code or -1,
+            request_id=None,
+            body=None,
+        ) from None
+    if isinstance(e, asyncio.TimeoutError):
+        raise APITimeoutError() from None
+    raise APIConnectionError() from e
```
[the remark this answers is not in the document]
luke-speechify: Fixed. `_raise_from` now re-raises existing `APIError` subclasses unchanged (`if isinstance(e, APIError): raise e`) before the fallback, so decode/attribute errors are no longer misclassified as retryable `APIConnectionError`. Matches the targeted handling in the other plugins.

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/__init__.py
```diff
@@ -14,26 +14,26 @@
...
 __all__ = [
     "TTS",
-    "Voice",
-    "TTSEncoding",
+    "ChunkedStream",
+    "SynthesizeStream",
     "TTSModels",
     "DEFAULT_VOICE_ID",
     "__version__",
 ]
```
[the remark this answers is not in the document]
luke-speechify: Restored `list_voices()` on the TTS class (now backed by the official SDK's `voices.list()`), and re-exported `GetVoice` from the package, so voice enumeration has a migration path. The old `TTSEncoding` is intentionally dropped — synthesis is now fixed to 24 kHz PCM (the SDK's stream accept-set no longer exposes container/rate selection); noted for release notes.

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/__init__.py
```diff
@@ -14,26 +14,29 @@
...
+from .tts import DEFAULT_MODEL, DEFAULT_VOICE_ID, TTS, ChunkedStream, SynthesizeStream
 from .version import __version__
 
 __all__ = [
     "TTS",
-    "Voice",
-    "TTSEncoding",
+    "ChunkedStream",
+    "SynthesizeStream",
+    "GetVoice",
```
tinalenguyen: i'm not sure if we should re-export an external sdk's function here

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/models.py
```diff
@@ -11,18 +11,14 @@
...
 # limitations under the License.
+
 from typing import Literal
 
 TTSModels = Literal[
     "simba-english",
     "simba-multilingual",
-]
-
-TTSEncoding = Literal[
```
tinalenguyen, line 21: is there a reason for removing this? this could break existing code

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/tts.py
```diff
@@ -35,114 +35,99 @@
...
-        model: NotGivenOr[TTSModels] = NOT_GIVEN,
-        base_url: NotGivenOr[str] = NOT_GIVEN,
-        api_key: NotGivenOr[str] = NOT_GIVEN,
+        voice_id: str = DEFAULT_VOICE_ID,
+        model: TTSModels = DEFAULT_MODEL,
         language: NotGivenOr[str] = NOT_GIVEN,
         loudness_normalization: NotGivenOr[bool] = NOT_GIVEN,
         text_normalization: NotGivenOr[bool] = NOT_GIVEN,
-        http_session: aiohttp.ClientSession | None = None,
-        follow_redirects: bool = True,
```
tinalenguyen, line 98: i think we should throw an error to users when using the deprecated arguments instead of just removing. how about we accept `**kwargs` and parse through those to inform users?

[example](https://github.com/livekit/agents/blob/f1af0690aaf68fc675caaa722c22d818186beb19/livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py#L946-L983)

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/tts.py
```diff
@@ -35,114 +35,99 @@
...
 
 class TTS(tts.TTS):
     def __init__(
         self,
         *,
-        voice_id: NotGivenOr[str] = DEFAULT_VOICE_ID,
-        encoding: NotGivenOr[TTSEncoding] = NOT_GIVEN,
-        model: NotGivenOr[TTSModels] = NOT_GIVEN,
-        base_url: NotGivenOr[str] = NOT_GIVEN,
-        api_key: NotGivenOr[str] = NOT_GIVEN,
```
tinalenguyen, line 93: i don't think we should remove `api_key`

livekit-plugins/livekit-plugins-speechify/pyproject.toml
```diff
@@ -9,8 +9,8 @@ description = "Agent Framework plugin for voice synthesis with Speechify's API."
 readme = "README.md"
 license = "Apache-2.0"
 requires-python = ">=3.10.0"
-authors = [{ name = "LiveKit", email = "hello@livekit.io" }]
-keywords = ["webrtc", "realtime", "audio", "video", "livekit", "speechify"]
+authors = [{ name = "Speechify", email = "hello@speechify.ai" }]
```
tinalenguyen: i saw your comment about seeing this precedent before, i believe that was left over from transferring that pypi package. thanks for catching that, since we own this pypi package already we should remove this change

tinalenguyen (commented):
thanks for the PR! left a few comments and questions

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/__init__.py
luke-speechify: yeah, dropped the `GetVoice` re-export. `list_voices()` now returns a plugin-owned `Voice` dataclass that mirrors the pre-3.0 shape (`id`, `type`, `display_name`, `gender`, `avatar_image`, `models`, `locale`), adapting from the SDK internally. `730339a3d`.

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/models.py
luke-speechify: good call, restored `TTSEncoding` in `models.py`. output is fixed to 24 kHz PCM under the SDK now so the alias no longer selects an encoding, but keeping the name exported means the import doesn't break. paired with the `**kwargs` shim on `TTS.__init__` that warns if someone still passes `encoding=` at construction. `730339a3d`.

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/tts.py
luke-speechify: yes, added the `**kwargs` shim on `TTS.__init__`, modeled on `_check_deprecated_args` in `speechmatics/stt.py`. warns on `encoding`, `http_session`, and `follow_redirects` with a reason for each. `730339a3d`.

livekit-plugins/livekit-plugins-speechify/livekit/plugins/speechify/tts.py
luke-speechify: fair, restored `api_key` and dropped `token` entirely. `token` only appeared in this PR (never released), so no compat shim needed for it. init and env-var name (`SPEECHIFY_API_KEY`) now match `main`. `730339a3d`.

livekit-plugins/livekit-plugins-speechify/pyproject.toml
luke-speechify: agreed, reverted `authors`, `Repository` and `Issues` back to LiveKit. I noticed your `tina/update-gnani-pyproject` branch doing the same revert on gnani which confirmed the direction. kept only the functional changes: the `aiohttp` to `speechify-api` dep swap, the `tts` keyword, and the `[tool.uv]` block. `730339a3d`.

tinalenguyen (approved):

PR #6339, opened 2026-07-07
feat(soniox): support speed parameter in TTS

yui0303:
The Soniox TTS WebSocket API accepts a speed field in the stream config to control the speaking rate (supported range 0.7–1.3, default 1.0), but the plugin never exposed it, so there was no way to change the speaking rate.

Add speed to `TTS.__init__` and `update_options`, validate it against the supported range, and include it in the config message sent over the WebSocket, consistent with how `bitrate` is handled.


tinalenguyen (approved):
thank you for the pr!

PR #6346, opened 2026-07-07
fix(tts): restart ChunkedStream retries under a fresh request_id

u9g:
## Problem

`test_tts_synthesize[deepgram]` failed on main ([run](https://github.com/livekit/agents/actions/runs/28864391089/job/85610676702)): a transient connection error mid-response caused `ChunkedStream` to retry after partial audio had already been emitted, so the consumer received frames from two attempts with mixed `request_id`s — and frames from the failed attempt could still trickle out after the retry started, interleaving the two attempts.

## Changes

- **`ChunkedStream._main_task`**: settle the emitter (`aclose`) before retrying, so no stale frames from the failed attempt are delivered once the retry begins. The retry restarts synthesis under a fresh `request_id`, signaling downstream that any partial audio from the failed attempt is stale. Also skip retries for non-retryable errors (matching `SynthesizeStream`).
- **`tests/test_tts.py`**: `_do_synthesis` now accepts a request_id restart — it asserts request_ids don't interleave between attempts and validates completeness/audio on the final attempt's frames. New test `test_tts_synthesize_retry_after_partial_audio` covers fail-once-then-succeed: exactly one retry, recoverable error event, two non-interleaved request_ids, full audio from the retry, one metrics event.
- **`tests/fake_tts.py`**: `fake_exception_count` option to fail the first N attempts, and yield control between pushes so partial audio reaches the consumer like a real provider streaming over the network.

## Notes

`SynthesizeStream` still refuses to retry after partial audio (#5242); making it restart under a fresh request_id like this would be a follow-up.

### DIFF
--- a/livekit-agents/livekit/agents/tts/tts.py
+++ b/livekit-agents/livekit/agents/tts/tts.py
@@ -440,16 +440,29 @@ class ChunkedStream(ABC):
                 if isinstance(e, APIStatusError) and e.status_code == 499:
                     return
 
-                retry_interval = self._conn_options._interval_for_retry(i)
-                if self._conn_options.max_retry == 0 or self._conn_options.max_retry == i:
+                # settle the emitter so no frames from this attempt are delivered after
+                # the retry starts; the retry restarts the synthesis under a fresh
+                # request_id, signaling downstream that any partial audio is stale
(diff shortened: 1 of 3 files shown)


longcw (approved):
lgtm!

PR #6349, opened 2026-07-07
fix(voice): preserve completed tool results in chat context on interruption

### LINKED ISSUE #3702: Tool Call Results Lost During User Interruption Leading to Duplicate Executions
```markdown
## 🐛 Bug Report: Tool Call Results Lost During User Interruption

### Summary
When a user interrupts the agent during or immediately after tool execution, the completed tool calls and their results are not saved to the chat history. This causes the LLM to be unaware that tools have already been executed in the next inference turn, leading to duplicate tool executions.

### Environment
- **livekit-agents version**: 1.1.6 (verified the issue still exists in 1.2.15)
- **Python version**: 3.9+
- **Use case**: Voice agent handling restaurant reservations

### Impact
- **Critical**: In production environments, this causes duplicate operations (e.g., creating duplicate orders)
- **Data integrity**: Multiple identical records created in database
- **User experience**: Users receive incorrect confirmations and duplicated resources

---

### 🔄 Steps to Reproduce

1. **Setup**: Voice agent with a tool that creates resources (e.g., `create_reservation`)
2. **User request**: User initiates a modification flow (e.g., "Change to 6 people")
3. **LLM executes tool**: Agent calls `create_reservation` and successfully creates Order #1
4. **User interrupts**: User speaks (VAD triggered) immediately after tool completion but before agent finishes speaking
5. **LLM continues**: Agent processes user's next input
6. **Result**: Agent calls `create_reservation` again, creating duplicate Order #2

### Expected Behavior
The completed tool call and its result should be saved to chat history even when interrupted, so the LLM knows not to re-execute the same tool.

### Actual Behavior
The tool executes successfully but its result is not recorded in the chat history when interruption occurs, causing the LLM to repeat the tool call.

---

### 📊 Real-World Example

#### Timeline from Production Logs

| Time | Event | Chat History State |
|------|-------|-------------------|
| 14:29:38.705 | User: "我要改成6个人" (Change to 6 people) | - |
| 14:29:50.289 | **LLM outputs tool call**: `create_reservation` | - |
| 14:29:50.583 | **✅ Order created successfully**: NT02011172025102200047 | ❌ Not recorded |
| 14:29:52.063 | **⚠️ User interrupts** (VAD triggered) | Only partial text saved |
| 14:29:53.736 | **LLM inference #2** starts | ❌ Unaware of Order #047 |
| 14:29:54.599 | LLM outputs tool call again: `create_reservation` | - |
| 14:29:59.909 | **💥 Duplicate order created**: NT02011172025102200048 | - |

#### Key Log Excerpts

```
livekit.agents|14:29:50,584|tool耗时:create_reservation elapsed: 292.86 ms
livekit.agents|14:29:52,063|Speech handle 打断
livekit.agents|14:29:52,063|_pipeline_reply_task interrupted position2
```

**Result**: Two identical orders created with same parameters (6 people, 2025-10-22 15:30:00, same location and decorations)

---

### 🔍 Root Cause Analysis

#### Problem Location

**File**: `livekit-agents/livekit/agents/voice/agent_activity.py`
**Function**: `_pipeline_reply_task()`
**Lines**: ~1577-1593 (v1.1.6) and similar in v1.2.15

#### Current Code (Interrupted Path)

```python
if forwarded_text:
    msg = chat_ctx.add_message(
        role="assistant",
        content=forwarded_text,  # ❌ Only saves partial text
        id=llm_gen_data.id,
        interrupted=True,
        created_at=reply_started_at,
    )
    self._agent._chat_ctx.insert(msg)
    self._session._conversation_item_added(msg)

await utils.aio.cancel_and_wait(exe_task)
return  # ❌ Early return skips tool result saving
```

**Issues**:
1. ❌ No `tool_calls` parameter in the message
2. ❌ No tool result messages (role="tool") created
3. ❌ Early `return` bypasses the tool saving logic at line ~1607

#### Why Tools Complete But Results Are Lost

**File**: `livekit-agents/livekit/agents/voice/generation.py`
**Function**: `_execute_tools_task()`

```python
except asyncio.CancelledError:
    # Wait for pending tools to complete
    await asyncio.gather(*tasks)  # ✅ Tools do finish
    # ❌ But results in tool_output.output are never accessed
```

Tools are protected by `asyncio.shield()` and complete successfully, but when `exe_task` is cancelled via `cancel_and_wait()`, the caller never retrieves `tool_output.output`.

#### Flow Diagram

```
┌──────────────────────────────────────────────┐
│  LLM generates response + calls tools        │
└──────────────┬───────────────────────────────┘
               │
        User interrupts?
        ┌──────┴──────┐
     Yes│             │No
        ↓             ↓
  ┌───────────┐  ┌──────────────┐
  │ INT path  │  │ Normal path  │
  │ • Save    │  │ • Save text  │
  │   text    │  │ • ✅ Save    │
  │ • ❌ Skip │  │   tool calls │
  │   tools   │  │ • ✅ Save    │
  │ • return  │  │   results    │
  └───────────┘  └──────────────┘
```

**I'm willing to contribute this fix!**

hello-rory:
Fixes #3702
Extends #4877 (credit @CarlosGarciaPro) — covers the remaining interruption path, keeps the realtime server-side context in sync, and adds regression tests.

## Problem

Completed tool results only reach the agent's `chat_ctx` if the speech carrying them survives its interruption gates. If the user interrupts at the wrong moment, the completed `FunctionCall`/`FunctionCallOutput` are silently dropped: the next inference has no record the tool ran, so the LLM re-issues the same call — duplicating side effects for non-idempotent tools (bookings, payments, ...).

Two ways this happens:

1. The tool reply turn — which carries the previous turn's tool messages via `_previous_tools_messages` — is interrupted before reaching the commit site (not covered by #4877; the path we hit in production).
2. The turn is interrupted during playout after its own tools completed; the results in `tool_output.output` were discarded (the variant #4877 fixes).

## Changes

All in `agent_activity.py`, no API changes:

- Commit `_previous_tools_messages` at the start of the reply turn, before any interruption gate (old site removed; `insert` orders by `created_at`, so conversation ordering is unchanged).
- On the post-playout interrupted return, commit completed pairs from `tool_output.output` (pipeline and realtime paths, as in #4877). The realtime path also pushes the outputs to the realtime session, since its server-side state drives the next generation.
- Handoff tools are excluded: an interrupted handoff is not applied, and recording it as completed would stop the LLM from retrying it.

## Tests

Five hermetic tests in `tests/test_tool_results_preserved_on_interruption.py`: both pipeline variants (including a tool still in flight at interruption), the realtime path, and the handoff exclusion. The four preservation tests fail on `main` and pass with this change. Existing suites and `make check` pass.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -3851,8 +3851,8 @@ class AgentActivity(RecognitionHooks):
                         )
                 else:
                     logger.warning(
-                        "tool results preserved locally but not sent to the realtime session; "
-                        "the model may re-execute the interrupted function calls",
+                        "interrupted tool results added to the local chat context "
+                        "but not sent to the realtime session",
                         extra={"function_calls": [out.call_id for out in interrupted_fnc_outputs]},
                     )
             return
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2790,6 +2790,14 @@ class AgentActivity(RecognitionHooks):
     ) -> None:
         from .agent import ModelSettings
 
+        # commit the tool messages that triggered this reply before any interruption
+        # gate: the tools already executed, and discarding their results on
+        # interruption makes the next LLM inference re-issue the same calls (see #3702).
+        # ChatContext.insert orders by `created_at`, so ordering is unchanged.
+        if _previous_tools_messages:
+            self._agent._chat_ctx.insert(_previous_tools_messages)
+            self._session._tool_items_added(_previous_tools_messages)
+
         current_span = trace.get_current_span(context=speech_handle._agent_turn_context)
         current_span.set_attribute(trace_types.ATTR_SPEECH_ID, speech_handle.id)
         if instructions is not None:
@@ -3163,11 +3171,6 @@ class AgentActivity(RecognitionHooks):
 
         current_span.set_attribute(trace_types.ATTR_SPEECH_INTERRUPTED, speech_handle.interrupted)
 
-        # add the tools messages that triggers this reply to the chat context
-        if _previous_tools_messages:
-            self._agent._chat_ctx.insert(_previous_tools_messages)
-            self._session._tool_items_added(_previous_tools_messages)
-
         forwarded_text = "".join(out.forwarded_text for out in segment_outputs)
         if speech_handle.interrupted:
             # forward_generation already cleared the buffer and waited for playout
@@ -3220,6 +3223,24 @@ class AgentActivity(RecognitionHooks):
 
         if speech_handle.interrupted:
             await utils.aio.cancel_and_wait(exe_task)
+
+            # cancel_and_wait lets in-flight tool tasks run to completion, so
+            # tool_output.output may hold completed results. Commit them before
+            # returning: the tool reply turn is never scheduled here, and dropping
+            # them makes the next LLM inference re-execute the same tools (see #3702).
+            # Agent handoffs are excluded: the handoff is not applied on an
+            # interrupted speech, and recording its call as completed would prevent
+            # the LLM from retrying it.
+            interrupted_calls: list[llm.FunctionCall] = []
+            interrupted_fnc_outputs: list[llm.FunctionCallOutput] = []
+            for sanitized_out in tool_output.output:
+                if sanitized_out.fnc_call_out is not None and sanitized_out.agent_task is None:
+                    interrupted_calls.append(sanitized_out.fnc_call)
+                    interrupted_fnc_outputs.append(sanitized_out.fnc_call_out)
+
+            if interrupted_tool_messages := interrupted_calls + interrupted_fnc_outputs:
+                self._agent._chat_ctx.insert(interrupted_tool_messages)
+                self._session._tool_items_added(interrupted_tool_messages)
             return
 
         # wait for the tool execution to complete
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -3822,6 +3843,34 @@ def _create_assistant_message(
...
+                # call and re-issue it
+                chat_ctx = self._rt_session.chat_ctx.copy()
+                chat_ctx.items.extend(interrupted_fnc_outputs)
+                try:
+                    await self._rt_session.update_chat_ctx(chat_ctx)
+                except llm.RealtimeError as e:
+                    logger.warning(
+                        "failed to update chat context with the interrupted function calls results",
+                        extra={"error": str(e)},
+                    )
```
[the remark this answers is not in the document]
hello-rory: 1. The failure mode is the pre-fix status quo. If the push fails, the server side degrades to exactly what happens on main today (dangling call, possible re-issue) — but local durable history is now correct, which is strictly better. The alternative prioritization (don't commit locally unless the server accepts) would lose the result everywhere.

2. It mirrors the repo's own idiom exactly. The non-interrupted path 90 lines below (agent_activity.py:3959-3967) does the identical thing: same update_chat_ctx push, same except llm.RealtimeError → logger.warning, no retry, no fallback. The maintainers already accepted this tradeoff for the more common path; adding a retry mechanism only in the interrupted branch would be inconsistent and scope creep for this PR.

3. No concrete failure scenario beyond the generic. "Server might reject due to stale state" applies equally to every update_chat_ctx call in the file.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -3822,6 +3837,27 @@ def _create_assistant_message(
...
+                    interrupted_fnc_outputs.append(sanitized_out.fnc_call_out)
+                    self._agent._chat_ctx._upsert_item(sanitized_out.fnc_call_out)
+                    self._session._tool_items_added([sanitized_out.fnc_call_out])
+
+            if len(interrupted_fnc_outputs) > 0:
+                # without this the server-side state keeps a dangling call and can re-issue it
+                chat_ctx = self._rt_session.chat_ctx.copy()
+                chat_ctx.items.extend(interrupted_fnc_outputs)
+                try:
+                    await self._rt_session.update_chat_ctx(chat_ctx)
```
longcw: This push is unconditional, but its effect depends on `auto_tool_reply_generation`:
- OpenAI (`False`): just records the result — clean fix.
- Gemini (`True`): sending the function response *is* the trigger to generate, so we'd start a spoken tool reply right after the user interrupted. 

Suggest gating: keep the local `_upsert_item`, but only `update_chat_ctx` when `not capabilities.auto_tool_reply_generation`. Note the local upsert alone doesn't prevent re-execution (realtime generation is driven by server-side state), so maybe log a warning in that case?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2790,6 +2790,12 @@ async def _pipeline_reply_task_impl(
     ) -> None:
         from .agent import ModelSettings
 
+        # commit before any interruption gate: dropping already-executed tool
+        # results makes the next inference re-issue the calls (see #3702)
+        if _previous_tools_messages:
```
longcw: Could we instead commit `tool_messages` to the agent's chat_ctx right after the tools finish, before creating the reply speech task? Then the commit is guaranteed regardless of what happens to that speech, `_previous_tools_messages` and the no-reply commit both go away, and the messages keep their natural `created_at`.

Only the interrupted-during-playout return still needs its own commit.

hello-rory: Yeah, that makes sense, thanks for pointing it out. I dropped `_previous_tools_messages`

livekit-agents/livekit/agents/voice/agent_activity.py
hello-rory: Hmm... I was digging into this a bit. I am less confident in the realtime code path changes here, so I just dropped them from the diff. Our issue was specifically with pipeline so I don't want to get out of scope. I was drilling Fable about it, and it came back at me with this:


> The one finding that actually matters: the Gemini "local-only" outputs aren't local-only forever
> 
> The gate holds interrupted outputs back from the server, but they still go into self._agent._chat_ctx. Google's update_chat_ctx (realtime_api.py:614) works by diffing the incoming context against the plugin's own _chat_ctx — and converts every new function_call_output in the diff into a LiveClientToolResponse (utils.py:78-84), which is precisely the generation trigger we're gating against.
> 
> So the withheld outputs sit in local context as a landmine: the next time any code path pushes the full agent context to the session — the any_skipped sync at agent_activity.py:3810 after a later interruption, or a new session getting seeded from agent context after reconnect/handoff — they show up in the diff and get delivered as tool responses at that arbitrary moment, potentially triggering an unprompted spoken reply mid-conversation.


I don't think I'm familiar enough with this codepath to judge the pertinence of this claim though. To be safe, I just reverted realtime changes. PR now affects only pipline path

longcw (approved):

PR #6359, opened 2026-07-09
(expressive mode): pass expressions at segment start, fix timedstring sync

tinalenguyen:
- lk.expression passed at header start
- providers that support timed strings are now aligned 
- strip any trailing whitespaces around expression tags


livekit-agents/livekit/agents/tts/_provider_format.py
```diff
@@ -821,8 +824,21 @@ def split_all_markup(text: str) -> tuple[str, list[ExpressiveTag]]:
     return clean, expr_tags + [{"type": tag, "value": value} for tag, value in raw_tags]
 
 
-def strip_all_markup(text: str) -> str:
-    """:func:`split_all_markup` returning only the clean text (tags discarded)."""
+def strip_all_markup(text: str, *, drop_open_tail: bool = False) -> str:
```
theomonnom, line 955: Do we need this new parameter?

Seems like we always want it to be true?

livekit-agents/livekit/agents/tts/_provider_format.py
```diff
@@ -821,8 +824,21 @@ def split_all_markup(text: str) -> tuple[str, list[ExpressiveTag]]:
     return clean, expr_tags + [{"type": tag, "value": value} for tag, value in raw_tags]
```
theomonnom, line 926: Let's make it inside another function
```python
def _drop_open_tail(text: str) -> str:
    last_lt = text.rfind("<")
    if last_lt > text.rfind(">"):
        nxt = text[last_lt + 1 : last_lt + 2]
        if not nxt or nxt == "/" or nxt.isalpha():
            text = text[:last_lt]
    return text
```

Also I removed the [] brackets checks, I don't think we need it.

theomonnom: What's tricky with this method also is that it can be fragile.
e.g:

"I typed <expr this expression by mistake". This will result in "I typed"

livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -29,6 +30,9 @@
 from .. import io
 from ..transcription import find_micro_track_id
 
+# a complete self-closing expressive marker (<expr/>, <expression/>, or <emotion/>)
+_EXPR_MARKER_SPLIT_RE = re.compile(r"(<(?:expr|expression|emotion)\b[^>]*?/\s*>)")
```
theomonnom: I think it's redundant to have _output.py re-parses those tags again

livekit-agents/livekit/agents/tts/_provider_format.py
tinalenguyen: i think only synchronizer needs it to be true

theomonnom: When does someone wants to strip tags? and still have a partial tag in the returned value?

livekit-agents/livekit/agents/tts/_provider_format.py
```diff
@@ -821,9 +824,41 @@ def split_all_markup(text: str) -> tuple[str, list[ExpressiveTag]]:
...
+    name = _TAIL_TAG_NAME_RE.match(frag).group()  # type: ignore[union-attr]
+    return name in _TAIL_TAG_NAMES or (
+        name == frag and any(t.startswith(name) for t in _TAIL_TAG_NAMES)
+    )
+
+
+def _drop_open_tail(text: str) -> str:
+    """Drop a trailing unterminated markup tag (text sliced/cut mid-tag)."""
+    if _open_tag_fragment(text):
+        return text[: text.rfind("<")]
```
theomonnom, line 951: Isn't it quite fragile? 

Is it possible that it will cut very simple sentences like

`In math, 5 < 10`

PR #6362, opened 2026-07-09
fix(azure): recover from STT error cancellations instead of hanging

### LINKED ISSUE #5322: Azure STT plugin does not handle timeout error properly
### Bug Description

In case of a timeout error from azure cognitiveservices speech sdk specifically in case of Speech Recognition the _on_canceled method does not handle the timeout error properly. There is no retry initiated nor is there any error emitted so that it can be further picked by fallback adaptor and it stays in a suspended state.

Here i am attaching the error log

`{"level": "WARNING", "name": "livekit.plugins.azure", "message": "Speech recognition canceled: CancellationDetails(reason=CancellationReason.Error, error_details=\"Timeout while waiting for service to stop SessionId: <session_id>\")", "code": "CancellationErrorCode.ServiceTimeout", "reason": "CancellationReason.Error", "error_details": "Timeout while waiting for service to stop SessionId:<session_id>", "pid": 90811, "job_id": "<job_id>", "room_id": "<room_id>", "timestamp": "2026-04-03T05:28:45.466820+00:00"}`

### Expected Behavior

The expected behaviour for this issue could have been to add explicit retries as well as after retry exhaustion it should emits this error forward so that it could be picked up by fallback adapter.

### Reproduction Steps

```bash
It can be only reproduced only in case of timeout error from azure sdk. It can be mocked  in a local setup as well.

here is the code snippet which is throwing this warning :-

def _on_canceled(self, evt: speechsdk.SpeechRecognitionCanceledEventArgs) -> None:
        if evt.cancellation_details.reason == speechsdk.CancellationReason.Error:
            logger.warning(
                f"Speech recognition canceled: {evt.cancellation_details}",
                extra={
                    "code": evt.cancellation_details.code,
                    "reason": evt.cancellation_details.reason,
                    "error_details": evt.cancellation_details.error_details,
                },
            )
```

### Operating System

Amazon linux

### Models Used

Azure STT

### Package Versions

```bash
livekit-agents[azure]==1.5.1; python_version >= '3.9'
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

apoorva-01:
If Azure STT hits an error cancellation (like a service timeout), the stream just goes silent. `_on_canceled` only logged a warning, and Azure doesn't reliably send `session_stopped` after that, so `_run` blocks on its `asyncio.wait` forever. Audio keeps flowing into a dead recognizer and nothing raises, so retry and fallback never kick in.

Now an error cancel records the details and sets the stopped event, so `_run` wakes and raises `APIConnectionError`. The base retry loop and FallbackAdapter handle it from there.

I kept it on the base retry path instead of adding retry inside the plugin like the earlier attempt (#5333). Happy to flip it if you'd rather.

Tests in `tests/test_plugin_azure_stt.py` cover both cases.

Fixes #5322

### DIFF
--- a/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
+++ b/livekit-plugins/livekit-plugins-azure/livekit/plugins/azure/stt.py
@@ -233,6 +233,7 @@ class SpeechStream(stt.SpeechStream):
 
         self._loop = asyncio.get_running_loop()
         self._reconnect_event = asyncio.Event()
+        self._cancellation_error: speechsdk.CancellationDetails | None = None
         self._audio_duration = 0.0
         self._last_audio_duration_report_time = time.monotonic()
 
@@ -257,6 +258,7 @@ class SpeechStream(stt.SpeechStream):
     async def _run(self) -> None:
         while True:
             self._session_stopped_event.clear()
+            self._cancellation_error = None
 
             self._stream = speechsdk.audio.PushAudioInputStream(
                 stream_format=speechsdk.audio.AudioStreamFormat(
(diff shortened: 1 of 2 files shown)


longcw (approved):
thanks for the pr! looks good to me

PR #6363, opened 2026-07-09
feat(smallestai): use Waves continuous WebSocket streaming protocol for TTS

harshitajain165:
## Summary

The SmallestAI `SynthesizeStream` currently buffers all incoming tokens locally and sends **one isolated request per segment**, without using the Waves continuous-streaming protocol (`continue` / `flush` / `max_buffer_flush_ms`) that comparable LiveKit TTS plugins (ElevenLabs, Cartesia, Rime, Neuphonic, Inworld, …) use.

This PR switches the plugin to the continuous protocol so it speaks the Waves streaming API the way the official SDK and other providers' plugins do:

- Forwards tokens to `/waves/v1/tts/live` as they arrive with `"continue": true`, instead of buffering the whole segment locally.
- Closes each segment with an explicit `"flush": true` frame.
- Exposes the server-side `max_buffer_flush_ms` buffer bound as a `TTS(...)` constructor arg and `update_options(...)` option (default `0`).
- Runs send/receive as concurrent tasks over the pooled connection. The connection is only returned to the pool after a clean `complete`; the pool discards it on error/interruption, so reuse stays safe.
- Collapses to **one `SynthesizeStream` instance = one segment**, matching the current base-class contract (multi-segment instances are deprecated upstream).

Response-frame handling (`chunk` / `word_timestamp` / `complete` / `error`) is unchanged, and the HTTP `ChunkedStream` path is untouched.

## Scope / note

This change is about **protocol parity** — using the continuous-streaming fields the Waves WS API exposes, matching the official SDK and other plugins. It is **not** a latency change: with current server behavior the endpoint synthesizes on the explicit `flush`, so output timing is unchanged from the previous buffer-until-flush implementation. The client is now ready to benefit from any future server-side incremental emission with no further plugin changes.

## Testing

Verified against the **live** Waves `/tts/live` endpoint:

- **Continuous streaming:** tokens streamed with `continue`/`flush`/`max_buffer_flush_ms` are accepted and return audio (`lightning_v3.1` / `magnus`).
- **Word timestamps over streaming:** with the correct voice/model pairing (`lightning_v3.1_pro` / `meher`, `word_timestamps=True`), the endpoint returns `word_timestamp` frames — all words surfaced as `TimedString` transcript entries with sensible timings alongside audio. The `word_timestamps` flag on the streamed frames is accepted (no "Invalid input data").
- **Interruption / reuse:** cancelling a stream mid-synthesis leaves the pool clean — a subsequent stream on the same TTS instance completes normally.
- **HTTP path:** `synthesize()` (`ChunkedStream`) is unchanged and still returns audio.
- Hermetic tests (local mock WS server) cover the wire protocol and the `word_timestamps` flag — kept local for now, not included in this PR.
- `ruff format`, `ruff check`, `mypy` all clean.

### DIFF
--- a/livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/tts.py
+++ b/livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/tts.py
@@ -319,7 +319,14 @@ class SynthesizeStream(tts.SynthesizeStream):
                 send_task = asyncio.create_task(self._send_task(ws))
                 recv_task = asyncio.create_task(self._recv_task(ws, output_emitter))
                 try:
-                    await asyncio.gather(send_task, recv_task)
+                    # send_task reports whether any text was actually sent; if the
+                    # segment was empty (no non-whitespace tokens) no flush is sent and
+                    # the server produces no `complete`, so don't wait on recv_task.
+                    sent_any = await send_task
+                    if sent_any:
+                        await recv_task
+                    else:
+                        await utils.aio.gracefully_cancel(recv_task)
                 finally:
                     await utils.aio.gracefully_cancel(send_task, recv_task)
         except asyncio.TimeoutError:
@@ -350,10 +357,12 @@ class SynthesizeStream(tts.SynthesizeStream):
             payload["word_timestamps"] = True
         return payload
 
-    async def _send_task(self, ws: aiohttp.ClientWebSocketResponse) -> None:
+    async def _send_task(self, ws: aiohttp.ClientWebSocketResponse) -> bool:
         # Forward each token as it arrives (continuous streaming) instead of buffering
         # the whole segment, so the server can begin synthesis before the text is
-        # complete. The flush sentinel (or end of input) closes the segment.
+        # complete. The flush sentinel (or end of input) closes the segment. Returns
+        # whether any non-whitespace text was actually sent.
+        sent_any = False
         async for data in self._input_ch:
             if isinstance(data, self._FlushSentinel):
                 break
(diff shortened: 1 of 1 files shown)


harshitajain165:
### Problem

The Waves WS API supports incremental streaming — continue / flush / max_buffer_flush_ms — but livekit-plugins-smallestai (as of 1.6.4) doesn't use any of it. Its SynthesizeStream buffers every token locally until a flush, then sends one isolated request per segment with none of the streaming fields.

### Change

- Stream tokens to /waves/v1/tts/live as they arrive with continue: true instead of buffering the whole segment, so the server can start synthesizing before the text is complete.
- Close each segment with an explicit flush: true.
- Expose the server-side max_buffer_flush_ms buffer bound as a constructor / update_options option.
- Send and receive run as concurrent tasks over a pooled WS connection; the connection is only returned to the pool after a clean complete (discarded on error/interruption), keeping reuse safe.
- Empty segments send no flush and don't wait on a complete the server never emits — so they finish promptly instead of hanging until the connection timeout.


@tinalenguyen Requesting a review whenever you get a chance

livekit-plugins/livekit-plugins-smallestai/livekit/plugins/smallestai/tts.py
```diff
@@ -272,106 +286,142 @@ async def _run(self, output_emitter: tts.AudioEmitter) -> None:
 
 
 class SynthesizeStream(tts.SynthesizeStream):
-    """WebSocket-based streaming synthesis — primary path used by the agent pipeline."""
+    """WebSocket-based streaming synthesis — primary path used by the agent pipeline.
+
+    Uses the Waves continuous-streaming protocol: text tokens are forwarded to the
+    server as they arrive (``continue: true``) rather than buffered locally, and an
+    explicit ``flush: true`` message closes each segment. This lets synthesis begin
+    before the full segment text is known, lowering time-to-first-byte.
```
hvmzack, line 294: The docstring says this lowers time-to-first-byte, but the scope note under PR description says output timing is unchanged since the server synthesizes on the explicit flush — and the only flush is at end-of-segment, so TTFB should be the same as before.

Would sentence-level flushing be in scope? other providers  e.g elevenlabs,cartesia run a tokenize.blingfire.SentenceTokenizer inside SynthesizeStream and flush per sentence instead of per segment. Since Waves already synthesizes on flush, that would let audio start after sentence one rather than after the full LLM response — the TTFB win the docstring describes, with no server-side change needed.

tinalenguyen (approved):
thanks for the PR!

PR #6372, opened 2026-07-09
fix(workflows): return corrective instructions as tool output instead of generate_reply

u9g:
## What

The data-capture workflows' `confirm_*` tools (and `credit_card`'s `update_*` validation branches) called `session.generate_reply()` from inside tool execution to voice corrective messages — scheduling a separate speech that competes with the turn's own tool_response reply and leaves nothing in the chat context. This PR returns the instruction as the tool's output instead, so it flows through the normal tool-response reply and merges with sibling tool outputs in the same turn.

`on_enter` hooks and the DTMF event handlers keep `generate_reply` — they run outside tool execution where there is no tool-output channel.

## Testing

Adds a simulation harness (`examples/data_capture_sim`) — an intake agent exposing one `collect_*` tool per data-capture workflow (email, phone, address, DOB with time, name, credit card), all with `require_confirmation=True` so text-only simulations still exercise the `confirm_*` paths, plus an `lk agent simulate` scenarios file driving each corrective branch: value corrected at the confirmation step, time of birth before the date, and the card number length / repeat-mismatch / expired-date validations. Expectations grade the two regressions this refactor could introduce: going silent where `generate_reply` used to speak, and completing with a stale value.

The sim runs in CI (`.github/workflows/data-capture-sim.yml`) on any change to `livekit-agents/livekit/agents/beta/workflows/**` or the sim harness itself.

### DIFF
--- a/.github/workflows/data-capture-sim.yml
+++ b/.github/workflows/data-capture-sim.yml
@@ -1,0 +1,59 @@
+name: data-capture-sim
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "livekit-agents/livekit/agents/beta/workflows/**"
+      - "examples/data_capture_sim/**"
+      - ".github/workflows/data-capture-sim.yml"
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "livekit-agents/livekit/agents/beta/workflows/**"
+      - "examples/data_capture_sim/**"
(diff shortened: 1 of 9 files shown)


chenghao-mou (approved):
lgtm. CI is broken due to access/permission issues. Not sure who owns the CI project though.

theomonnom (approved):

PR #6373, opened 2026-07-09
feat(metrics): broadcast llm tps + ttfs

Bobronium:
- `llm_node_tps` — output tokens/sec, from the provider's usage chunk
- `llm_node_ttfs` — inference start until the first sentence is handed off for synthesis (reuses [`livekit.agents.types.USERDATA_TTS_STARTED_TIME`](https://github.com/livekit/agents/blob/a2264ff37ebcd7950508610740252b8917e5b9da/livekit-agents/livekit/agents/types.py#L76))
- ~~`agent_false_interruption` — the event, forwarded over the session host~~ 
  was implemented in #6691 

Needs: https://github.com/livekit/protocol/pull/1661 (released, already used)

### DIFF
--- a/livekit-agents/livekit/agents/voice/generation.py
+++ b/livekit-agents/livekit/agents/voice/generation.py
@@ -24,7 +24,7 @@ from ..llm import (
 from ..llm.chat_context import Instructions
 from ..log import logger
 from ..telemetry import trace_types, tracer
-from ..tokenize import blingfire
+from ..tokenize import SentenceTokenizer, blingfire
 from ..types import (
     USERDATA_TIMED_TRANSCRIPT,
     USERDATA_TTS_STARTED_TIME,
@@ -66,13 +66,6 @@ class _LLMGenerationData:
 # output for an injected in-progress tool call, phrased so the model waits instead of
 # re-issuing the call.
 _RUNNING_TOOL_PLACEHOLDER = "The tool call is still in progress."
-
-
-# Stateless and cheap to construct, min_sentence_len=1
-# keeps short first sentences (e.g. "Hi there.") counted when timing ttfs.
-_SENTENCE_TOKENIZER = blingfire.SentenceTokenizer(min_sentence_len=1)
-
-
 # extra flag marking an injected pair so it can be stripped before the ctx is forwarded.
 _RUNNING_PLACEHOLDER_KEY = "__lk_running_placeholder__"
 
@@ -160,12 +153,22 @@ def perform_llm_inference(
     model_settings: ModelSettings,
     model: str | None = None,
     provider: str | None = None,
+    sentence_tokenizer: SentenceTokenizer | None = None,
 ) -> tuple[asyncio.Task[bool], _LLMGenerationData]:
     text_ch = aio.Chan[str | FlushSentinel]()
     function_ch = aio.Chan[llm.FunctionCall]()
     data = _LLMGenerationData(text_ch=text_ch, function_ch=function_ch)
     llm_task = asyncio.create_task(
-        _llm_inference_task(node, chat_ctx, tool_ctx, model_settings, data, model, provider)
+        _llm_inference_task(
+            node,
+            chat_ctx,
+            tool_ctx,
+            model_settings,
+            data,
+            model,
+            provider,
+            sentence_tokenizer=sentence_tokenizer,
+        )
     )
     llm_task.add_done_callback(lambda _: text_ch.close())
     llm_task.add_done_callback(lambda _: function_ch.close())
@@ -189,6 +192,8 @@ async def _llm_inference_task(
     data: _LLMGenerationData,
     model: str | None = None,
     provider: str | None = None,
+    *,
+    sentence_tokenizer: SentenceTokenizer | None = None,
 ) -> bool:
     start_time = time.perf_counter()
     current_span = trace.get_current_span()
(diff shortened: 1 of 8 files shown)


livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -206,6 +216,20 @@ async def _llm_inference_task(
...
+
+    async def _set_time_first_sentence() -> None:
+        try:
+            async for _ in sentence_stream:
+                data.ttfs = time.perf_counter() - start_time
+                return
+        except Exception:
+            logger.exception("failed to time first sentence (llm_node_ttfs)")
+
+    ttfs_task = asyncio.create_task(_set_time_first_sentence())
```
theomonnom: The tokenizer configuration may actually be different than the one being really used. We should try to use the real one

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -179,6 +192,8 @@ async def _llm_inference_task(
     data: _LLMGenerationData,
     model: str | None = None,
     provider: str | None = None,
+    *,
+    sentence_tokenizer: SentenceTokenizer | None = None,
```
u9g: let's make this a non-keyword argument

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -228,6 +243,20 @@ async def _llm_inference_task(
         return False
 
     # forward llm stream to output channels
+    usage: Any = None
+    if sentence_tokenizer is None:
+        sentence_tokenizer = blingfire.SentenceTokenizer(retain_format=True)
```
u9g: can we use the sentence tokenizer off of the activity?

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -228,6 +243,20 @@ async def _llm_inference_task(
         return False
 
     # forward llm stream to output channels
+    usage: Any = None
```
u9g: is there any way we can make this !any

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -283,9 +314,22 @@ async def _llm_inference_task(
...
                 data.generated_text += content
                 text_ch.send_nowait(content)
+                if data.ttfs is None:
+                    sentence_stream.push_text(content)
     finally:
         if isinstance(llm_node, _ACloseable):
             await llm_node.aclose()
+        try:
+            sentence_stream.end_input()  # flush any trailing sentence, then close
+        except RuntimeError:
```
u9g: except Exception*?

tests/test_session_host.py
```diff
@@ -158,6 +158,18 @@ def test_partial_fields(self) -> None:
         pb = _metrics_to_proto(metrics)
         assert pb.transcription_delay == pytest.approx(0.42)
 
+    @pytest.mark.skipif(
```
u9g: let's remove this

u9g (approved):

livekit-agents/livekit/agents/voice/generation.py
Bobronium: This actually prompted me to think about a better design overall. Now PR just uses timings from the real STT node, avoiding a re-tokenization all-together.

u9g:
Tested end-to-end against a real voice agent on this branch (deepgram/nova-3 + gpt-4.1-mini + cartesia/sonic-3, `resume_false_interruption=True`, `false_interruption_timeout=1.0`), driving it over `lk.agent.session` with an external client. All three features arrive on the wire.

| turn | ttft | ttfs | tts_ttfb | tps |
|---|---|---|---|---|
| on_enter greeting | 1.432 | 1.648 | 0.204 | 7.9 |
| short factual answer | 0.518 | 0.710 | 0.245 | 40.7 |
| long reply, false-interrupted + resumed | 0.602 | 0.966 | 0.224 | 54.3 |

- `ttfs > ttft` by 0.19–0.36s every turn (the segmentation gap), consistent with `ttfb` anchoring on the same synthesis-start stamp. No `tts_node_ttfb` regression from the `.get(key, default)` → explicit-fallback change.
- `agent_false_interruption{resumed:true}` fired 3.3s after `overlapping_speech{is_interruption:true}` when a non-lexical barge-in produced no final transcript; the paused speech resumed and committed with full metrics.
- Text-mode session: `llm_node_tps` present, `llm_node_ttfs` absent — including when TTS 429'd mid-session (no frames → no stamp → no ttfs, tps still reported).

One semantic worth a second look: `tps = completion_tokens / (perf_counter() - start_time)` with `start_time` stamped before inference, so TTFT is in the denominator. The greeting shows it — 7.9 tok/s reported, but 1.43s of that 1.77s was time-to-first-token. It washes out on long replies (54.3 vs ~68 excluding TTFT) but understates short ones a lot. If the metric means "output throughput", anchoring on first-token time is truer; if it means "effective tokens per wall-second", the docstring could say so.

Note for external clients: this needs `livekit-protocol>=1.1.19`. On 1.1.11 the fields don't exist and the new event/metrics decode as silently absent.

Bobronium:
> One semantic worth a second look: tps = completion_tokens / (perf_counter() - start_time) with start_time stamped before inference, so TTFT is in the denominator. 

Constrained it to actual stream duration:

turn | ttft | ttfs | tts_ttfb | tps
-- | -- | -- | -- | -- 
greeting | 1.896 | 2.024 | 0.344 | 78.8
short factual answer | 1.480 | 1.612 | 0.364 | 101.5
long reply | 1.432 | 1.569 | 0.388 | 94.1

theomonnom (approved):

PR #6385, opened 2026-07-10
fix(openai): omit unset fields when serializing Responses WS requests

markine:
## Summary

The OpenAI **Responses WebSocket** transport (`use_websocket=True`, the default for `responses.LLM`) serializes request models by hand — `json.dumps(msg, default=lambda o: o.model_dump(mode="json"))` — instead of going through the openai SDK. The SDK omits unset fields; this hand-rolled path did not (`exclude_none` was missing), so every Optional field that defaults to `None` was emitted as an explicit `null` on the wire.

That is fine until the OpenAI schema grows an Optional enum field. `openai-python` **2.45.0** added `Reasoning.mode` (`Literal["standard","pro"] | None`, default `None`). The plugin auto-attaches `Reasoning(effort="none")` for gpt-5.x models, which now serializes as:

```json
"reasoning": {"context": null, "effort": "none", "generate_summary": null, "mode": null, "summary": null}
```

and the Responses API rejects it with **HTTP 400**:

```
Invalid type for 'reasoning.mode': expected one of 'standard' or 'pro', but got null instead.
```

Worse, that 400 was invisible: the error-frame parser crashed while trying to report it (see fix #2), so users saw a misleading pydantic `ValidationError` about `sequence_number` instead of the real message.

This breaks any Responses-API user on the WS transport running openai ≥ 2.45.0 (gpt-5.x, where `reasoning` is auto-set). The HTTP transport (`use_websocket=False`) is unaffected because the SDK omits unset fields.

## What changed

Two small fixes in `responses/llm.py`, plus a hermetic regression test.

1. **Serialization** — `_ResponsesWebsocket.generate_response._default` now uses `model_dump(mode="json", exclude_none=True)`. This mirrors the SDK's on-the-wire shape and is forward-compatible with any future Optional field OpenAI adds to these models. (Matches the file's own existing convention — `_handle_response_completed` already uses `exclude_none=True`.)

2. **Error reporting** — `_parse_ws_event` defaults `sequence_number` when validating top-level `error` frames. Those frames (e.g. a request-validation 400) don't carry `sequence_number`, which `ResponseErrorEvent` marks required, so validating them raised a `ValidationError` that masked the real API message. With the default, genuine errors surface cleanly as an `APIStatusError` via `_handle_error`.

3. **Test** — `tests/test_plugin_openai_responses.py`, hermetic (no network/keys):
   - drives the real `generate_response` through a fake WS pool and asserts the serialized `reasoning` has no null-valued keys;
   - asserts a `sequence_number`-less error frame parses cleanly and preserves the message.

## What this changes for users

Before: on openai ≥ 2.45.0, a gpt-5.x agent using the Responses API over WebSocket (the default) fails every turn with an opaque `ValidationError` about `sequence_number`; the underlying cause (a 400 on `reasoning.mode`) is hidden.

After: the request serializes the way the SDK would, the turn succeeds, and if any request *is* rejected the real API error message is surfaced instead of a parser crash.

## Testing

- `uv run pytest tests/test_plugin_openai_responses.py --plugin openai` → 2 passed.
- Verified end-to-end against the live OpenAI Responses WebSocket on **openai 2.45.0**: a gpt-5.4 turn with `Reasoning(effort="none")` now returns `response.completed` (400 before the fix); a deliberately invalid request surfaces the real error message (pydantic crash before the fix).
- Negative control: both the test and the live checks fail on the pre-patch code, confirming they guard the regression.
- `ruff format --check` and `ruff check` pass on the changed file.

## Notes

- The two fixes are independent; #1 stops the 400, #2 makes any future error debuggable. Happy to split into two commits if preferred.
- No changelog/changeset tooling found in the repo, so no fragment added.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
@@ -95,7 +95,19 @@ class _ResponsesWebsocket:
     async def generate_response(self, msg: dict) -> AsyncGenerator[dict, None]:
         def _default(o: object) -> object:
             if isinstance(o, openai.BaseModel):
-                return o.model_dump(mode="json")
+                # exclude_none is load-bearing, not cosmetic. This hand-rolled WS
+                # transport serializes request models itself instead of going
+                # through the openai SDK (which omits unset fields). Without
+                # exclude_none, every Optional field the model defaults to None
+                # is emitted as an explicit `null` on the wire. The Responses API
+                # rejects explicit nulls on fields that expect an enum: e.g. after
+                # openai-python added `Reasoning.mode` (default None), a plain
+                # `Reasoning(effort=...)` began serializing `"mode": null`, which
+                # the API 400s with "Invalid type for 'reasoning.mode': expected
+                # one of 'standard' or 'pro', but got null instead." Omitting None
+                # mirrors the SDK's on-the-wire shape and is forward-compatible
+                # with future Optional additions to these models.
+                return o.model_dump(mode="json", exclude_none=True)
             raise TypeError(f"unexpected type {type(o)}")
 
         try:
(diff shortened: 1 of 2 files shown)


markine:
Hi folks, this is a Claude-generated diff. Better to have it visible so others facing issues with openai-2.45.0 can find it. Please advise whether you want both changes and any cleanup/testing that's commonplace in your repo. Thank you.

davidzhao (approved):
lg

PR #6387, opened 2026-07-11
fix(inference): flush text preamble to TTS when a tool call starts

AKomplished-bug:


bnovik0v (commented):
Thanks for picking this up, and shape B looks like the right call — keeping `_event_ch` as `Chan[ChatChunk]` avoids the `_metrics_monitor_task` guards that shape A would have needed, and the marker-carries-no-`tool_calls` invariant is exactly the property that makes it safe.

One structural issue with the patch as written: the marker `return` sits **inside** the `for tool in delta.tool_calls` loop (`inference/llm.py:512-523`), so it exits before the rest of the loop body and before the `finish_reason` block. Any delta that carries more than a single bare tool name now loses information that `main` handles today:

1. **`delta.content` and `extra_content` are dropped.** If a provider puts the last text token in the same delta as the first named tool call, `main` falls through to the bottom of `_parse_choice` and emits it; the patch returns the marker with `content=None` and the text is gone. Same for `extra_content` (Google thought signatures) on that delta. Slightly ironic failure mode given the issue is about getting the preamble out.

2. **Parallel tool calls in one delta are dropped.** Two named tool calls in a single `delta.tool_calls` list → the second never gets accumulated, because we return during the first iteration. The pre-existing `return call_chunk` has the same shape, but it only fires on an index change mid-accumulation, which is precisely the streaming case where one-tool-per-delta holds. The new return fires on the first tool of *every* turn, so it meets batched deltas head-on.

3. **The tool call can be lost entirely.** If a provider emits the assembled tool call *and* `finish_reason="tool_calls"` in the same delta, `_parse_choice` returns the marker and never reaches the `finish_reason` block — that chunk is consumed, the call is never assembled, and the agent silently never executes the tool. OpenAI proper doesn't do this (finish_reason arrives in its own chunk with an empty delta), but I'd want to rule it out for the OpenAI-compatible gateways and vLLM tool parsers this change is aimed at before relying on it.

I haven't reproduced 1-3 against live providers, so how often they fire is empirical — but `main` handles all three and the patch doesn't, on a code path every `inference` user shares, so it seems worth closing off regardless.

The fix looks small: hoist the marker out of the loop (set a `pending` flag inside, emit after), fold `content`/`extra_content` into the marker chunk, and guard on `finish_reason is None` so it can never preempt the assembled call.

```python
# inside the loop, replacing the early return:
if tool.function.name:
    ...
    if not self._tool_start_signaled:
        pending_tool_start = True

# after the `for` loop, before the finish_reason block:
if pending_tool_start and call_chunk is None and choice.finish_reason is None:
    self._tool_start_signaled = True
    return llm.ChatChunk(
        id=id,
        delta=llm.ChoiceDelta(
            role="assistant",
            content=delta.content,
            extra=getattr(delta, "extra_content", None),
            tool_call_started=True,
        ),
    )
```

On the tests: driving `_parse_choice` directly is a good call, no reason to need a network. They currently only feed one-tool-per-delta with content and tool deltas always separated, which is why the three cases above pass. Three cases worth adding:

- content + first named tool in the same delta → asserts the content survives on the marker
- two named tool calls in one delta → asserts both are accumulated
- tool call + `finish_reason="tool_calls"` in one delta → asserts the assembled call is still emitted

Two smaller notes, neither blocking:

- Only the `inference` plugin emits the marker, so the standalone `openai` / `anthropic` / `google` plugins keep the old behavior. Fine as a first cut — probably worth stating in the PR description so it doesn't read as a global fix.
- The PR is currently conflicting with `main` and needs a rebase.

The example update reads well. Happy to help with any of this if it's useful — this is the fix I want on our side too.

AKomplished-bug:
Went with your suggestion: the marker is now emitted after the `delta.tool_calls` loop, carries `content`/`extra_content` from the same delta, and is guarded on `finish_reason is None` so it can never eat the assembled call. Also made `generation.py` send the flush after routing the chunk's content, otherwise a preamble token riding the marker delta would get stuck behind its own flush. Added the three tests you listed — the finish_reason-same-delta shape was actually already covered by a test from #6398 on main, which my old patch broke, so that confirmed it.

PR #6392, opened 2026-07-12
fix(xai): self-heal the realtime chat context when an item's anchor is missing

### LINKED ISSUE #6391: xAI realtime: an interruption desyncs the local chat context, then every later item is dropped (unrecoverable)
### Bug Description

On a `livekit-plugins-xai` realtime call, an interruption can desync the plugin's local `RemoteChatContext` from xAI's server-side item list. Once that happens, **every subsequent `conversation.item.added` is dropped**, because each new item anchors to an item the local copy no longer has — and the drop cascades: the next item anchors to the item that was just dropped, so it fails too, and so on for the rest of the call. The realtime model loses all conversation state and the agent stalls (multi-second silences, then hard `Item not found` API errors).

This reproduces on `livekit-plugins-xai==1.6.5` / `livekit-agents==1.6.5` (the current releases, which already include the item-duplication fix from #6271).

### What the logs look like

An interruption early in the call:

```
ERROR  speech not done in time after interruption, cancelling the speech arbitrarily.
```

then, later, the first desync and the cascade that never recovers:

```
WARN  failed to insert item `f6dce9e1…`: previous_item_id `55273a67…` not found
WARN  failed to insert item `f8cccbcc…`: previous_item_id `f6dce9e1…` not found
WARN  failed to insert item `c1581c9d…`: previous_item_id `f6dce9e1…` not found
WARN  failed to insert item `1e7b9805…`: previous_item_id `f6dce9e1…` not found
...
ERROR xAI Realtime API returned an error: RealtimeError(message='Item not found: …', type='invalid_request_error')
```

`f6dce9e1` fails once (its anchor `55273a67` is gone from the local context), and from then on every item that chains off `f6dce9e1` also fails. In a trace of one ~2:16 call, the context broke ~69s in and never recovered; the remaining ~65s had reply gaps of 12–16s.

### Root cause

Two xAI-specific workarounds in `livekit/plugins/xai/realtime/realtime_model.py` are fragile under the delete/truncate traffic that an interruption generates:

```python
def _handle_conversion_item_added(self, event: ConversationItemAdded) -> None:
    # xAI always sends previous_item_id=None; anchor to the local tail
    if event.previous_item_id is None:
        event.previous_item_id = (
            self._remote_chat_ctx._tail.item.id if self._remote_chat_ctx._tail else None
        )
    super()._handle_conversion_item_added(event)

def _handle_conversion_item_deleted(self, event: ConversationItemDeletedEvent) -> None:
    # xAI sends item_id empty; GUESS it's the first pending delete
    if event.item_id == "" and self._item_delete_future:
        event.item_id = list(self._item_delete_future.keys())[0]
    super()._handle_conversion_item_deleted(event)
```

On an interruption, `truncate()` deletes items. Because the `conversation.item.deleted` event arrives with an empty `item_id`, the handler **guesses** which item was removed (the first pending delete future). When several deletes are in flight the guess can pick the wrong item, so the local context deletes an item the server kept (and keeps one the server dropped). After that, a server `conversation.item.added` referencing the still-present-server item finds no local anchor, and `RemoteChatContext.insert()` raises `previous_item_id … not found`. The base `_handle_conversion_item_added` catches that and **drops the item**, which strands every later item that anchors to it — the unrecoverable cascade above.

### Proposed fix

Make item insertion self-healing so a single desync can't corrupt the whole context: when the anchor for a newly-added item isn't present locally, fall back to appending at the current tail instead of dropping the item. In `livekit/plugins/xai/realtime/realtime_model.py`:

```python
def _handle_conversion_item_added(self, event: ConversationItemAdded) -> None:
    # xAI's conversation.item.added always has previous_item_id=None; anchor to the tail.
    # It can also reference an item our local copy already dropped after a truncation/
    # delete desync. Anchoring to a missing item makes the base handler drop this item,
    # which strands every later item that chains off it and corrupts the whole context.
    # Fall back to the current tail so the context self-heals instead of cascading.
    prev = event.previous_item_id
    if prev is None or self._remote_chat_ctx.get(prev) is None:
        prev = self._remote_chat_ctx._tail.item.id if self._remote_chat_ctx._tail else None
    event.previous_item_id = prev
    super()._handle_conversion_item_added(event)
```

This keeps `RemoteChatContext.insert()` strict but guarantees the item-added path never anchors to a missing item, so at worst one item is slightly out of order rather than the entire remaining conversation being lost. (A more thorough fix would also make `_handle_conversion_item_deleted` identify the deleted item precisely instead of guessing, but that needs the deleted item's id, which xAI's event doesn't currently provide.)

Happy to open a PR with the self-heal above.

### Environment

- `livekit-agents==1.6.5`
- `livekit-plugins-xai==1.6.5`
- Model: `grok-voice-think-fast-1.0`
- Reproduces most readily on interruption-heavy calls (the caller talks over the agent).

thecarlmartin:
## Summary

Fixes an unrecoverable conversation-state cascade on `livekit-plugins-xai` realtime calls (filed as #6391).

On an interruption, `truncate()` deletes items. xAI's `conversation.item.deleted` event arrives with an empty `item_id`, so `_handle_conversion_item_deleted` **guesses** which item was removed (the first pending delete future). When several deletes are in flight the guess can pick the wrong item, desyncing the local `RemoteChatContext` from xAI's server-side list. After that, a server `conversation.item.added` referencing a still-present-server item finds no local anchor, `RemoteChatContext.insert()` raises `previous_item_id … not found`, and the base handler **drops the item** — which strands every later item that anchors to it. The whole context cascades into nothing and the model stalls (multi-second silences, then hard `Item not found` API errors).

In one ~2:16 trace the context broke ~69s in and never recovered:

```
WARN failed to insert item `f6dce9e1…`: previous_item_id `55273a67…` not found
WARN failed to insert item `f8cccbcc…`: previous_item_id `f6dce9e1…` not found   # chains off the dropped f6dce9e1
WARN failed to insert item `c1581c9d…`: previous_item_id `f6dce9e1…` not found
...
```

## Fix

Anchor a newly-added item to the current tail whenever its `previous_item_id` is missing from the local context, so a single desync **self-heals** (at worst one item is slightly out of order) instead of dropping the item and cascading:

```python
prev_id = event.previous_item_id
if prev_id is None or self._remote_chat_ctx.get(prev_id) is None:
    prev_id = self._remote_chat_ctx._tail.item.id if self._remote_chat_ctx._tail else None
event.previous_item_id = prev_id
```

This keeps `RemoteChatContext.insert()` strict and is scoped to the xAI handler (which already fills a `None` `previous_item_id` from the tail). A more thorough fix would also make `_handle_conversion_item_deleted` identify the deleted item precisely instead of guessing, but that needs the deleted item's id, which xAI's event doesn't currently provide — so this hardens the recovery path regardless.

## Tests

Adds two tests to `tests/test_realtime/test_xai_realtime_model.py`:
- `test_item_added_with_missing_anchor_self_heals_to_tail` — an item whose anchor is absent is kept and lands at the tail (previously dropped).
- `test_item_added_missing_anchor_does_not_cascade` — a following item that anchors to the self-healed item also inserts (the cascade is broken).

All tests in that file pass locally. I don't have access to run the live-xAI integration suite, so a maintainer verification against the real API would be appreciated.

Fixes #6391.

### DIFF
--- a/livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/realtime/realtime_model.py
@@ -180,12 +180,20 @@ class RealtimeSession(openai.realtime.RealtimeSession):
         super()._handle_function_call(item)
 
     def _handle_conversion_item_added(self, event: ConversationItemAdded) -> None:
-        # xAI's `conversation.item.added` event always has the previous_item_id as None
-        # replace it with the last item in the remote chat context
-        if event.previous_item_id is None:
-            event.previous_item_id = (
-                self._remote_chat_ctx._tail.item.id if self._remote_chat_ctx._tail else None
-            )
+        # xAI's `conversation.item.added` event always has the previous_item_id as None;
+        # anchor it to the last item in the remote chat context.
+        #
+        # The event can also reference an item our local copy no longer has, after an
+        # interruption's truncation/delete traffic desyncs us from the server (the
+        # `conversation.item.deleted` event carries no item id, so the delete is guessed
+        # and can drop the wrong item). Anchoring to a missing item makes the base handler
+        # raise `previous_item_id not found` and drop this item, which strands every later
+        # item that chains off it and corrupts the whole context. Fall back to the current
+        # tail so a single desync self-heals instead of cascading.
+        prev_id = event.previous_item_id
+        if prev_id is None or self._remote_chat_ctx.get(prev_id) is None:
+            prev_id = self._remote_chat_ctx._tail.item.id if self._remote_chat_ctx._tail else None
+        event.previous_item_id = prev_id
 
         super()._handle_conversion_item_added(event)
 
--- a/tests/test_realtime/test_xai_realtime_model.py
+++ b/tests/test_realtime/test_xai_realtime_model.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
 import pytest
-from openai.types.realtime import ConversationItemInputAudioTranscriptionCompletedEvent
+from openai.types.realtime import (
+    ConversationItemAdded,
+    ConversationItemInputAudioTranscriptionCompletedEvent,
+    RealtimeConversationItemUserMessage,
+)
 
 from livekit.agents import llm
 from livekit.agents.llm.remote_chat_context import RemoteChatContext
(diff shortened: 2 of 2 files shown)


longcw:
the issue should be fixed in https://github.com/livekit/agents/pull/6729

PR #6398, opened 2026-07-13
Fix stripping of Gemma reasoning markers in streamed output

### LINKED ISSUE #6375: Inference: strip Gemma 4 <|channel>thought…<channel|> markers (strip_thinking_tokens only knows <think>)
## Summary

LiveKit Inference's `google/gemma-4-31b-it` emits Gemma 4 channel-thought markup that leaks into customer-facing agent output (transcription / TTS / chat history). `strip_thinking_tokens` only strips `<think>`/`</think>`, so Gemma's markers pass through unchanged.

## Markers

Per [Gemma 4 model card](<https://ai.google.dev/gemma/docs/core/model_card_4>):

* Open: `<|channel>thought` (often followed by `\n`)
* Close: `<channel|>`
* Body: internal reasoning (or empty when thinking is "off" — larger models still emit the empty wrapper)

Example stream content that reaches the agent:

```text
<|channel>thought
Customer wants a $100 refund; check policy.
<channel|>I can help start a return instead.
```

Or empty:

```text
<|channel>thought
<channel|>Happy to help with that.
```

## Where it leaks

In the Agents SDK Inference LLM path, deltas go through `strip_thinking_tokens` before becoming `ChatChunk`s:

[https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/inference/llm.py](<https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/inference/llm.py>)

```python
THINK_TAG_START = "<think>"
THINK_TAG_END = "</think>"

def strip_thinking_tokens(content: str | None, thinking: asyncio.Event) -> str | None:
    ...
```

Those tags don't match Gemma's channel markers, so the markup reaches `llm_node` → TTS / `transcription_node` / session history.

## Expected

Inference (or `strip_thinking_tokens`) should treat Gemma channel-thought blocks like `<think>` blocks: drop the markers and any body between them so only the final answer is customer-visible. Prior-turn thoughts also must not re-enter chat history (Gemma model card: "No Thinking Content in History").

## Actual

Raw `<|channel>thought` / `<channel|>` (and sometimes the thought body) appear in the LiveKit transcription stream used by text chat UIs, and would be spoken on voice.

## Repro (Agents SDK + Inference)

1. `AgentSession` with `inference.LLM(model="google/gemma-4-31b-it")`
2. Multimodal/text chat so agent replies ride `lk.transcription`
3. Ask something that triggers a tool-using support turn (refund / damage claim)
4. Inspect the transcription / chat history for `<|channel>thought` / `<channel|>`

## Workaround

App-side `llm_node` filter that strips the Gemma markers before chunks reach TTS/transcription/history (same state-machine pattern as the chain-of-thought recipe, with Gemma's tags).

## Related

* Gemini thought-part handling: livekit/agents#4190
* `strip_thinking_tokens` list-delta crash: livekit/agents#6323

C0deRatoR:
Fixes livekit/agents#6375

## Problem

Gemma channel-thought markers and their hidden reasoning pass through the shared thinking-token filter into customer-visible output, including TTS, transcripts, and chat history.

## Root cause

`strip_thinking_tokens` only recognizes `<think>` blocks and uses an `asyncio.Event`, which records whether the stream is inside a reasoning block but cannot retain marker fragments split across chunks. It also does not handle complete or multiple reasoning blocks within one chunk.

## Change

* add a stateful streaming filter for existing `<think>` and Gemma `<|channel>thought` / `<channel|>` blocks
* buffer only possible marker prefixes across chunk boundaries
* preserve visible content before and after reasoning blocks
* flush incomplete visible marker-like text when the stream finishes while dropping unclosed reasoning

## Tests

* `uv run --frozen pytest tests/test_llm_utils.py --unit -q` — 11 passed
* `uv run --frozen pytest --unit --no-concurrent --ignore=tests/test_room.py -q` — 1116 passed, 2 skipped
* `make check` — formatting, Ruff, and mypy passed

`tests/test_room.py` could not run locally because this environment has no Docker executable; its nine failures were fixture setup errors before test execution.

## Scope

No new dependencies, provider calls, or unrelated inference changes.

### DIFF
--- a/livekit-agents/livekit/agents/inference/llm.py
+++ b/livekit-agents/livekit/agents/inference/llm.py
@@ -72,6 +72,10 @@ _UNSUPPORTED_PARAMS: dict[str, set[str]] = {
 # models that don't support reasoning_effort when function tools are present
 _REASONING_EFFORT_TOOL_INCOMPATIBLE_PREFIXES: set[str] = {"gpt-5.2", "gpt-5.4"}
 
+_MODEL_THINK_TAGS = {
+    "google/gemma-4-31b-it": ("<|channel>thought", "<channel|>"),
+}
+
 
 def drop_unsupported_params(
     model: str, params: dict[str, Any], tools: list[Any] | None = None
@@ -426,7 +430,11 @@ class LLMStream(llm.LLMStream):
                 **self._extra_kwargs,
             )
 
-            thinking_filter = llm_utils.ThinkingTokenFilter()
+            thinking_filter = llm_utils.ThinkingTokenFilter(
+                *_MODEL_THINK_TAGS.get(
+                    self._model, (llm_utils.THINK_TAG_START, llm_utils.THINK_TAG_END)
+                )
+            )
             async with stream:
                 async for chunk in stream:
                     for choice in chunk.choices:
@@ -473,6 +481,10 @@ class LLMStream(llm.LLMStream):
         # the delta can be None when using Azure OpenAI (content filtering)
         if delta is None:
             return None
+
+        delta.content = llm_utils.strip_thinking_tokens(
+            delta.content, thinking_filter, final=choice.finish_reason is not None
+        )
 
         if delta.tool_calls:
             for tool in delta.tool_calls:
@@ -533,10 +545,6 @@ class LLMStream(llm.LLMStream):
             self._tool_call_id = self._fnc_name = self._fnc_raw_arguments = None
             self._tool_extra = None
             return call_chunk
-
-        delta.content = llm_utils.strip_thinking_tokens(
-            delta.content, thinking_filter, final=choice.finish_reason is not None
-        )
 
         # Extract extra from delta (e.g., Google thought signatures on text parts)
         delta_extra = getattr(delta, "extra_content", None)
--- a/livekit-agents/livekit/agents/inference/llm.py
+++ b/livekit-agents/livekit/agents/inference/llm.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
 import os
 from dataclasses import dataclass
 from typing import Any, Literal, cast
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -42,6 +42,10 @@
 
 THINK_TAG_START = "<think>"
 THINK_TAG_END = "</think>"
+THINK_TAG_PAIRS = (
+    (THINK_TAG_START, THINK_TAG_END),
+    ("<|channel>thought", "<channel|>"),
```
longcw: since this is gemma only, can we make the start and end tags configurable per model, so it applies to only gamma. for future models we can set that tag directly, and if not set fallback to the default <think>

livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -535,7 +534,9 @@ def _parse_choice(
             self._tool_extra = None
             return call_chunk
 
-        delta.content = llm_utils.strip_thinking_tokens(delta.content, thinking)
+        delta.content = llm_utils.strip_thinking_tokens(
```
longcw: maybe move strip to the top of `_parse_choice` so it also applied for the text content alongside a tool call?

livekit-agents/livekit/agents/llm/utils.py
C0deRatoR: Done in 68b535da6. `ThinkingTokenFilter` now accepts configurable start/end tags; the Inference path maps only `google/gemma-4-31b-it` to the Gemma markers and falls back to `<think>`/`</think>` for every other model. I also added a regression test confirming Gemma markers remain untouched without explicit tag configuration.

livekit-agents/livekit/agents/inference/llm.py
C0deRatoR: Done in 68b535da6. Filtering now runs immediately after the `delta is None` guard, before tool-call handling, so text emitted alongside a tool call is filtered too. Added a regression covering visible text followed by an empty Gemma thought block on a tool-call chunk.

longcw (approved):

PR #6406, opened 2026-07-13
(expressive mode): refactor options, remove prompt presets

tinalenguyen:


tests/test_expr_markup.py
theomonnom, line 1: Not sure if those tests are really useful, they're mostly noise IMO

theomonnom (approved):

theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -141,23 +140,59 @@ class SessionConnectOptions:
...
+    breathing: bool
+    """audible breath, inhale, exhale"""
+    sighing: bool
+    crying: bool
+    vocalizing: bool
+    """non-lexical voiced sounds — humming a tune, sing-song or sung delivery"""
+    mouth_sounds: bool
+    """tsk, tongue-click, lip-smack"""
+    reflex_sounds: bool
+    """cough, clearing the throat, yawn"""
```
theomonnom, line 177: cc @ShayneP on that for a review as well

theomonnom: maybe we need to be more specific. 

idk if `clearing the throat, yawn` is useful in any usecase lol

cough could be a nice one tho

tinalenguyen: i think it would be useful for NPCs or character voice agents

livekit-agents/livekit/agents/tts/_provider_format.py
```diff
@@ -193,403 +192,254 @@ def _xai_break_to_bracket(match: re.Match[str]) -> str:
...
+
+
+# NonverbalOptions field -> the provider's expr sound labels it governs. A provider
+# absent here (cartesia) has no non-verbal sounds; an empty list means the provider
+# has no sound for that field (nothing to filter). _allowed_sounds uses this to
+# remove disabled labels from the advertised vocabulary, so a sound steering turns
+# off is never exposed to the LLM in the first place. Every label in
+# _PROVIDER_SOUNDS must be governed by exactly one field, so a preset controls
+# the full vocabulary.
+_NONVERBAL_SOUND_LABELS: dict[str, dict[str, list[str]]] = {
```
ShayneP, line 395: Can we expose the provider / capabilities matrix? Right now the only way to know which `NonverbalOptions` fields do anything for a given voice is to read this here. The info ideally needs to be queryable. Something like:

`tts.markup.supports_expressive -> bool`
`tts.markup.supported_nonverbals -> dict[str, list[str]] `

Basically we need this for builder: the UI needs to gray out or hide sound toggles per selected voice and without an accessor, the web repo needs to hardcode a copy of these labels which isn't great since it will probably drift.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -141,23 +140,59 @@ class SessionConnectOptions:
...
+
+
+class SpeechSteeringOptions(TypedDict, total=False):
+    """Steers verbal delivery and non-verbal sounds in generated speech.
+
+    ``nonverbal_sounds`` is atomic: passing it replaces the default value
```
ShayneP: Loosely held opinion, but this feels like a bit of a footgun to me. 

In `presets.py` we say:

```
A preset is a plain :class:`SpeechSteeringOptions` value — pass it as
``speech_steering`` directly, or spread it and override individual fields
(``{**presets.CASUAL, "pace": "slow"}``)
```

For top level fields like `pace` it works fine, but if you try to do the same with sounds (which is where most of the presets are different from each other anyways) we blow away all of the other sounds by design.

So maybe we should add docs to `presets.py` showing like:

```
{**presets.CASUAL, "nonverbal_sounds": {**presets.CASUAL["nonverbal_sounds"], "laughing": True}}
```

Which is somewhat ugly ... but I think it's kind of easy to miss that `nonverbal_sounds` is atomic.

livekit-agents/livekit/agents/voice/agent_session.py
cshape: anecdotally, i've heard customer support-oriented folks liking `clear throat` in their voice agents. from the ones that fish can do, i only included it and `laugh`.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -141,23 +140,70 @@ class SessionConnectOptions:
...
+    """laugh, chuckle, giggle — and laugh-speak delivery"""
+    breathing: bool
+    """audible breath, inhale, exhale"""
+    sighing: bool
+    crying: bool
+    vocalizing: bool
+    """non-lexical voiced sounds — humming a tune, sing-song or sung delivery"""
+    mouth_sounds: bool
+    """tsk, tongue-click, lip-smack"""
+    reflex_sounds: bool
```
theomonnom, line 176: I think we need more granularity if we enable clearing the throat and yawn. 

If we're not happy with the result of expressive mode, I don't think adding more sounds make sense. Let's focus on what works. 

I would just do coughing.


What's tricky with those options: is that you may be OK with cough, but not with yawn at all (can be seen as disrespectful for some agents).

Same for breathing, audible breath may be OK, but not exhale?

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -57,6 +58,16 @@ class TTSCapabilities:
     """Whether this TTS supports aligned transcripts with word timestamps"""
 
 
+@dataclass
+class MarkupCapabilities:
```
theomonnom: Maybe MarkupInfo?

Not sure if those are really "capabilities"

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -57,6 +58,16 @@ class TTSCapabilities:
...
+    """What the expressive markup pipeline can do with a given voice."""
+
+    expressive: bool = False
+    """Whether this TTS declares an expressive markup dialect"""
```
theomonnom: Can we directly check if tts.markup.llm_instructions is not None?
```suggestion
```

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -94,15 +105,28 @@ def _provider_key(self) -> str:
...
+            """The queryable markup capabilities matrix for this voice."""
+            from ._provider_format import llm_instructions, supported_nonverbals
+
+            key = self._provider_key()
+            return MarkupCapabilities(
+                expressive=llm_instructions(key) is not None,
+                nonverbals=supported_nonverbals(key),
+            )
+
+        def llm_instructions(self, steering: SpeechSteeringOptions | None = None) -> str | None:
```
theomonnom:
```suggestion
        def llm_instructions(self, *, speech_steering: SpeechSteeringOptions | None = None) -> str | None:
```

theomonnom (approved):

PR #6410, opened 2026-07-13
feat(voice): opt-in audio fade-out on interruption (#6165)

### LINKED ISSUE #6165: Feature request: smooth fade-out when the agent is interrupted
# Feature request: smooth fade-out when the agent is interrupted

## The problem

When a user starts talking over the agent and the SDK confirms an interruption, the agent's voice stops **instantly**. The decision to stop is correct — but the abrupt cut, often mid-word, is the single most "robotic" artifact we hear on real phone calls. 

Here's an example of what it sounds like today — the user interrupts and the agent stops abruptly:
https://drive.google.com/file/d/128KuaXsuUTkGBBxy9ycaxNI__Q4Pap-9/view?usp=sharing

We're on `livekit-agents 1.6.0`.

## What we'd like

A supported option to let the agent's voice **fade out over a short window (~100–250ms)** when an interruption is confirmed, instead of cutting instantly.

To be clear about scope: this is **only** about the *way* the agent stops, not *when*. Interruption detection would stay exactly as it is — same rules, same timing. The only thing that changes is that the last fraction of a second fades out smoothly instead of being chopped off.

Ideally this would be a simple, declarative setting — something like:

```
interruption_fade_out = 0.2   # seconds; 0 = instant cut (today's behavior)
```

dorukdumlu:
Closes #6165

## What

Adds an opt-in linear fade-out applied to the agent's audio when it gets interrupted, instead of cutting playback abruptly:

```python
await session.start(
    room=room,
    room_options=RoomOptions(
        audio_output=AudioOutputOptions(fade_out_on_interruption=0.08),  # seconds
    ),
)
```

Default is `0.0` (disabled) — existing behavior is unchanged.

## How (and why not the holdback buffer I proposed in the issue)

In my comment on #6165 I proposed holding back a small tail buffer before forwarding frames downstream. I dropped that approach while implementing: it adds first-frame playback latency equal to the tail size, which is the one metric a voice agent shouldn't pay for a cosmetic feature.

Instead, the fade is built inside `_ParticipantAudioOutput`, which already knows everything needed at interruption time:

- `_forward_audio` keeps a small rolling history (~400ms: queue size + fade + margin) of frames already pushed to the `rtc.AudioSource`.
- On interruption, `AudioSource.queued_duration` tells us how much of that history hadn't played yet, so the exact playback position is `end_of_history - queued_duration`.
- After `clear_queue()`, a single frame is re-pushed: the next `fade_out_on_interruption` seconds of the *actual* audio from the playback position, with a linear 1 → 0 gain ramp.

This gives zero added latency and no content discontinuity — the fade is the same audio the user was hearing, just ramped down. The history bookkeeping is only active when the option is enabled.

`playback_position` accounting is untouched: the faded tail replays audio that was already subtracted via `queued_duration`, so reported position stays correct for chat-history truncation.

## Notes

- The fade duration is clamped to what's actually unplayed (interruption near the end of an utterance fades over the remainder).
- Known pre-existing race, unchanged by this PR: if `_forward_audio` is blocked in `AudioSource.capture_frame` under backpressure when the interruption fires, that in-flight frame can land after `clear_queue()`. Today it plays at full gain after the cut; with the fade enabled it may play before the faded tail. Fixing that needs coordination between `_forward_audio` and `_wait_for_playout` and felt out of scope here.

## Testing

- `_build_fade_out_frame` is a pure function; added hermetic unit tests (`tests/test_interruption_fade_out.py`, marked `unit`): ramp shape/monotonicity, correct offset from playback position, clamping to unplayed duration and to available history, stereo, and the no-op cases.
- Full `--unit` suite passes (same result as `main` baseline on my machine).
- `ruff check` / `ruff format` clean; no new mypy errors in `room_io`.

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/_output.py
+++ b/livekit-agents/livekit/agents/voice/room_io/_output.py
@@ -3,7 +3,10 @@ from __future__ import annotations
 import asyncio
 import json
 import time
+from collections import deque
+from collections.abc import Sequence
 
+import numpy as np
 from google.protobuf.json_format import MessageToDict
 
 from livekit import rtc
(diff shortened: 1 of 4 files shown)


dorukdumlu:
Ended up implementing this in #6410 went with a zero-latency approach inside _ParticipantAudioOutput instead of the holdback buffer, details in the PR description

dorukdumlu:
Friendly ping — CI is green and CLA is signed. Happy to adjust the approach (fade location / API surface) if you'd rather this live somewhere else than `_ParticipantAudioOutput`.

dorukdumlu:
Rebased onto current main. Kept #6728 / #6636 playback accounting (_source_pushed_duration, _forwarding_idle, _interruption_generation). Fade-out still rebuilds a zero-latency tail from already-queued frames. Pause discard now clears fade history so an interrupt after pause cannot rebuild a tail from frames that are no longer in the audio source.

dorukdumlu:
Both Devin findings were real, addressed in fbce451.

The cancellation one: the fade `capture_frame` await did sit between `clear_queue()` and the state reset, so a cancellation landing there (overlapping `flush()` or `aclose()`) left `_interrupted_event` set forever — `_forward_audio` kept discarding frames and the speech never reported finished. The reset and `on_playback_finished` now run in a `finally`, which restores the pre-fade guarantee: once an interruption is observed, the finish notification fires. If the cancel lands mid-capture, the fade deadline is also cleared since it's unknown whether the tail made it into the source.

The accounting one: the tail bypasses `_forward_audio`, so it was never counted as pushed, and a next segment interrupted or paused while the tail was still draining got its playback position shaved by up to the fade duration. The tail now records when it will have drained (`_fade_tail_until`) and both the interrupt and pause-discard reads subtract the residual. This also corrects the `unplayed_duration` fed to `_build_fade_out_frame`, which could otherwise start the fade earlier than the real playback position when a leftover tail sat in the queue.

Three regression tests: the cancelled fade push still finishing playout, and both accounting paths.

PR #6417, opened 2026-07-14
fix(voice): enforce interruption min_duration on STT transcript triggers (#3515)

### LINKED ISSUE #3515: `min_interruption_duration` seems to be ignored over `min_interruption_words`
## Description
Using `min_interruption_duration` seems to be ignored over `min_interruption_words` , when these are set up in the settings:

```python
        session = AgentSession[InboundCallData](
            preemptive_generation=preemptive_generation, # False
            stt=stt_engine,  # Amazon Transcribe
            llm=llm_engine,  # OpenAI
            tts=tts_engine,  # Elevenlabs
            vad=session_vad, # Silero VAD
            turn_detection=turn_detection, # "vad"
            userdata=call_data,
            mcp_servers=mcp_servers,
            user_away_timeout=user_away_timeout,
            allow_interruptions=allow_interruptions, # True
            min_interruption_duration=min_interruption_duration, # Tested as described below
            min_interruption_words=min_interruption_words, # Tested as described below
            min_endpointing_delay=min_endpointing_delay,
            max_endpointing_delay=max_endpointing_delay,
        )
```

Here's the behavior I'm getting:
- If I keep the defaults (`min_interruption_duration = 0.5` and `min_interruption_words = 0`), it behaves as expected, with the user interrupting the agent with short utterances.
- If I increase the `min_interruption_duration` to `10.0`, but keep the `min_interruption_words` at default (`0`), the behavior doesn't change. The model is still interrupted after ~1 second.
  - One important note about this: I've read https://github.com/livekit/agents/issues/2197, and @longcw mentions that _"`min_interruption_duration` won't block STT event to interrupt the agent"_. However, don't think it's an STT event that is blocking the agent. See the section "Why I suspect the STT is not interrupting the agent" below.
- If I keep the default the `min_interruption_duration` to `0.5`, but increase `min_interruption_words` to 50 (`0`), the model will _only_ be interrupted after a very long phrase (I presume, after 50 words). Talking for more than 0.5 seconds will not interrupt it.
- If I increase both to high values, the model is triggered

That is, the `min_interruption_duration` seems to have no effect in the behavior of the model, and _only_ setting `min_interruption_words` higher or lower seems to make a difference on the tolerance of the model for interruption from the user.

### Why I suspect the STT is not interrupting the agent
If you say a small utterance, as expected, an STT event will be triggered, and the agent will be interrupted.

However, even if you keep talking with no breathing (giving no breaks for the STT to get the transcript), the agent will still be interrupted after roughly 1 second.

We have an `stt_node` set up, and can confirm though the logs that it's not triggered _until_ the utterance finishes. The agent however, will be interrupted _way_ before that. That is, if you talk non-stop for 5 seconds, the agent will still be interrupted after just the first second (even though the `stt_node` is only triggered at the end of the 5th second).

Therefore, I suspect it's not the case that the model is being interrupted by an STT event.

## Environment
- OS: Windows 11
- Livekit Version: 1.2.11

dorukdumlu:
Closes #3515

## Problem

`interruption.min_duration` (formerly `min_interruption_duration`) is only enforced on the VAD path: `on_vad_inference_done` checks `ev.speech_duration >= min_duration` before calling `_interrupt_by_audio_activity()`. The STT paths bypass it entirely — `on_interim_transcript` and `on_final_transcript` interrupt as soon as any non-empty transcript arrives, and `_interrupt_by_audio_activity()` itself only gates on `min_words`.

With providers that stream continuous interim results (Amazon Transcribe, etc.), the first interim lands well before `min_duration` is reached, so the option effectively does nothing — which is what was reported in #3515 (details in [my comment there](https://github.com/livekit/agents/issues/3515#issuecomment-4935433280)). This implements Option A from that comment: all entry paths share the same `min_duration` semantics.

## Fix

- `AudioRecognition` gains a `current_speech_duration` property derived from the speech timestamps it already tracks (`_speech_start_time` / `_last_speaking_time`): wall-clock elapsed while the user is speaking, actual segment length once they've stopped, `None` when no speech start was ever tracked.
- `_interrupt_by_audio_activity()` enforces `min_duration` against that value, next to the existing `min_words` gate, so the interim/final transcript paths can no longer bypass it.

Two deliberate exemptions:

- **Unknown duration (`None`) is allowed through.** This preserves the existing failsafe in `on_final_transcript` ("agent speech might not be interrupted if VAD failed and a final transcript is received"), and keeps VAD-less sessions behaving as before rather than becoming un-interruptible.
- **The interruption detection model path (`on_interruption`) skips the check** via `enforce_min_duration=False` — that model already made an explicit "this is a real interruption" decision.

The VAD path keeps its existing pre-check; the new gate can't disagree with it (wall-clock elapsed ≥ VAD-measured speech duration), so VAD-triggered interruptions are unaffected.

## Testing

- New regression test `test_min_interruption_duration_applies_to_stt_transcripts`: user speaks for 4s with interim transcripts streaming every 0.3s and `min_duration=2.0`. Before the fix the first interim interrupts at ~5.3s (playback position ~1.8s); after the fix the interruption fires at 7.0s (position 3.5s). Verified it fails without the fix and passes with it.
- `FakeSTT`/`FakeActions` gained an opt-in `interim_interval` to simulate continuous-interim providers; default behavior of existing fixtures is unchanged.
- Full `--unit` suite passes; `ruff` clean; no new mypy errors.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1808,7 +1808,10 @@ class AgentActivity(RecognitionHooks):
         self._schedule_speech(handle, SpeechHandle.SPEECH_PRIORITY_NORMAL)
 
     def _interrupt_by_audio_activity(
-        self, *, ignore_user_transcript_until: float | None = None
+        self,
+        *,
+        ignore_user_transcript_until: float | None = None,
+        enforce_min_duration: bool = True,
     ) -> None:
         """
         Interrupt the current speech or generation, and optionally ignore the user transcript until the given timestamp.
@@ -1816,6 +1819,9 @@ class AgentActivity(RecognitionHooks):
         Args:
             ignore_user_transcript_until: The timestamp until which the user transcript should be ignored.
                 If None, the user transcript will be ignored until the current time.
+            enforce_min_duration: Whether to enforce ``interruption.min_duration`` against the
+                tracked user speech duration. Set to False by callers that already made an
+                explicit interruption decision (e.g. the interruption detection model).
         """
         if not self._interruption_by_audio_activity_enabled:
             return
@@ -1839,6 +1845,20 @@ class AgentActivity(RecognitionHooks):
             # TODO(long): better word splitting for multi-language
             if len(split_words(text, split_character=True)) < interruption_options["min_words"]:
                 return
+
+        # enforce min_duration on every audio-activity trigger, not only the VAD path;
+        # STT interim/final transcripts would otherwise interrupt regardless of how
+        # short the user speech was (https://github.com/livekit/agents/issues/3515).
+        # an unknown duration (None) is allowed through to keep the VAD-failure
+        # failsafe in on_final_transcript working.
+        if (
+            enforce_min_duration
+            and interruption_options["min_duration"] > 0
+            and self._audio_recognition is not None
+            and (speech_duration := self._audio_recognition.current_speech_duration) is not None
+            and speech_duration < interruption_options["min_duration"]
+        ):
+            return
 
         if self._rt_session is not None:
             self._rt_session.start_user_activity()
@@ -1972,10 +1992,13 @@ class AgentActivity(RecognitionHooks):
             self._user_silence_event.set()
 
     def on_interruption(self, ev: inference.OverlappingSpeechEvent) -> None:
-        # restore interruption by audio activity and then immediately interrupt
+        # restore interruption by audio activity and then immediately interrupt.
+        # the interruption detection model already decided this is a real interruption,
+        # so min_duration is not re-checked here.
         self._restore_interruption_by_audio_activity()
         self._interrupt_by_audio_activity(
-            ignore_user_transcript_until=ev.overlap_started_at or ev.detected_at
+            ignore_user_transcript_until=ev.overlap_started_at or ev.detected_at,
+            enforce_min_duration=False,
         )
         # flush held transcripts again if possible
         if self._audio_recognition:
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2003,7 +2003,10 @@ class AgentActivity(RecognitionHooks):
         self._schedule_speech(handle, SpeechHandle.SPEECH_PRIORITY_NORMAL)
 
     def _interrupt_by_audio_activity(
-        self, *, ignore_user_transcript_until: float | None = None
+        self,
+        *,
+        ignore_user_transcript_until: float | None = None,
+        enforce_min_duration: bool = True,
     ) -> None:
         """
         Interrupt the current speech or generation, and optionally ignore the user transcript until the given timestamp.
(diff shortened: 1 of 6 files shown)


dorukdumlu:
Friendly ping — CI is green on this one too. The change is intentionally small: gate STT-triggered interruptions on the same `min_duration` the VAD path already uses, with an escape hatch for the adaptive interruption model. Happy to tweak if you'd prefer a different call site.

dorukdumlu:
﻿Addressed Devin's finding on `current_speech_duration`:

Prefer VAD-measured `speech_duration` (same metric `on_vad_inference_done` already uses for `min_duration`) instead of wall-clock elapsed while `_speaking`. Wall-clock remains the fallback for STT-only setups with no VAD events.

Keeps short coughs/noise from clearing `min_duration` just because the interim transcript arrived after trailing silence.

dorukdumlu:
Addressed the post-EOS Silero overwrite: after `END_OF_SPEECH`, Silero resets `pub_speech_duration` to 0, so later `INFERENCE_DONE` events were clobbering `_vad_speech_duration` and making late STT finals look too short for `interruption.min_duration`.

Now we only update from `INFERENCE_DONE` while still speaking or when `speech_duration > 0`, and keep the EOS final otherwise. Added a unit regression in `tests/test_speech_start_time_persistence.py`.

dorukdumlu:
Addressed Devin's stale-duration finding: _bounce_eou_task now clears _vad_speech_duration with the other speech anchors on turn commit, so the next STT-failsafe path sees unknown duration (None) instead of the previous segment. Regression in 	ests/test_speech_start_time_persistence.py.

dorukdumlu:
Addressed Devin's remaining yellow: _vad_speech_duration is now written before on_vad_inference_done, so _interrupt_by_audio_activity sees the same frame the VAD pre-check used. Post-EOS zero guard is unchanged. Regression in 	ests/test_speech_start_time_persistence.py.

dorukdumlu:
Addressed Devin's remaining yellow: `turn_detection=stt` can set `_speaking` before Silero has onset, so `if self._speaking or ev.speech_duration > 0` wrote `0.0` and `min_duration` blocked the STT-failsafe barge-in.

INFERENCE_DONE now updates `_vad_speech_duration` only when `speech_duration > 0`. Post-EOS zero guard is unchanged. Rebased onto current main. Regression in `tests/test_speech_start_time_persistence.py`.

dorukdumlu:
Devin's stale-duration finding was real. A VAD blip that never produces a transcript never commits a turn (`_run_eou_detection` bails without one), so the blip's voiced duration survived indefinitely — and if VAD then missed the real barge-in, the `on_final_transcript` failsafe gated against that stale value instead of unknown. The "None passes through" exemption never got a chance to trigger.

Fixed in aec38f5 by scoping the measurement in time: once the measured segment ended more than 2s ago, `current_speech_duration` reports None. That's long enough for a late STT final of the same segment (the reason the EOS value is kept at all) and short enough not to leak into an unrelated one. Added a regression test that walks the exact blip → VAD-miss → late-final sequence from the review.

PR #6418, opened 2026-07-14
fix(voice): recover STT stream after unrecoverable errors

### LINKED ISSUE #5665: bug: STT errors close AgentSession immediately, lacking the tolerance counter that LLM and TTS errors have

longcw:
Fixes #5665.

Unlike the LLM and TTS streams, which are recreated every turn, the STT stream is long-lived: it is created once per agent activity and iterated by a single pump task. When it exhausted its per-utterance retries it emitted an unrecoverable `stt_error` and the pump ended, so nothing recreated it — and because `stt_error` had none of the session-level tolerance that `llm_error`/`tts_error` have, the session closed immediately on the first unrecoverable STT error.

This makes the long-lived STT stream self-healing and gives it the same session-level tolerance:

- `_STTPipeline` recreates the STT stream (with a short backoff) after an unrecoverable error, so the agent keeps hearing the user if the provider recovers.
- `AgentSession` applies `max_unrecoverable_errors` to `stt_error` as the circuit breaker: each failed stream increments the count, and it resets on a user transcript (STT's success signal, mirroring how llm/tts reset on agent speaking).
- The pump does not recreate the stream while the session is closing, so a teardown in progress can't spawn provider connections that are torn down at once.

Supersedes #5666, #5997, and #6001

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -177,28 +177,29 @@ class _STTPipeline:
         """Iterate the STT node and forward events into *event_ch*.
 
         Owns the generator lifecycle — never cancelled during handoff, only the
-        consumer is swapped. On an unrecoverable error the long-lived stream is
+        consumer is swapped. On a connection failure the long-lived stream is
         recreated after a backoff; the session tolerance is what closes it.
         """
         from .agent import ModelSettings
 
         while True:
-            node = self._stt_node(self._audio_ch, ModelSettings())
-            if asyncio.iscoroutine(node):
-                node = await node
-
-            if not isinstance(node, AsyncIterable):
-                # None or a non-streaming node: nothing to iterate or recover
-                return
-
             try:
+                node = self._stt_node(self._audio_ch, ModelSettings())
+                if asyncio.iscoroutine(node):
+                    node = await node
+
+                if not isinstance(node, AsyncIterable):
+                    # None or a non-streaming node: nothing to iterate or recover
+                    return
+
                 async for ev in node:
                     assert isinstance(ev, stt.SpeechEvent), (
                         f"STT node must yield SpeechEvent, got: {type(ev)}"
                     )
                     self._event_ch.send_nowait(ev)
-            except Exception:
-                # the error is already emitted and counted by the session; recreate to resume
+            except APIError:
+                # only a connection failure is retried (it was emitted and counted by the
+                # session); any other error propagates and stops the pump
                 if self._is_closing():
                     return
                 logger.warning(
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -215,6 +215,12 @@ class _STTPipeline:
             # node ended without error (audio input closed): stop
             return
 
+    def _rebind_node(self, stt_node: io.STTNode) -> None:
+        # the pipeline outlives the agent that created it (reused across handoff);
+        # recreation must call the currently-active node, not the previous agent's
+        # bound node whose activity is torn down (would raise, stopping the pump)
+        self._stt_node = stt_node
+
     async def aclose(self) -> None:
         await aio.cancel_and_wait(self._pump_task)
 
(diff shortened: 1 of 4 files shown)


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -53,6 +53,8 @@
 MIN_LANGUAGE_DETECTION_LENGTH = 5
 # Mirrors turn_detector.base.MAX_HISTORY_TURNS for tracing
 _EOU_MAX_HISTORY_TURNS = 6
+# backoff before recreating the stt stream after an unrecoverable error
+_STT_RECONNECT_INTERVAL = 0.5
```
theomonnom, line 57: Should we use an exponential backoff?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -748,7 +774,11 @@ async def _aclose(self) -> None:
     def _update_stt(self, stt: io.STTNode | None, *, pipeline: _STTPipeline | None = None) -> None:
         self._stt = stt
         if pipeline is None and stt is not None:
-            pipeline = _STTPipeline(stt)
+            # _closing_task is set synchronously at close, earlier than _closing
+            session = self._session
+            pipeline = _STTPipeline(
+                stt, is_closing=lambda: session._closing_task is not None or session._closing
+            )
```
theomonnom: I'm wondering if it's fine to directly pass an instance of self._session? 

Also nit: 

I think we should add a private var directly to the session like:

```
def _is_closing(self) -> bool: 
      self._closing_task is not None or self._closing
```

theomonnom (approved):

PR #6428, opened 2026-07-15
(google llm): strip thought summaries

tinalenguyen:
todo: expose in separate field in chatitem ?

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
@@ -619,6 +619,12 @@ class LLMStream(llm.LLMStream):
         if not part.text:
             return None
 
+        # Drop thought summaries: when include_thoughts is enabled Gemini streams its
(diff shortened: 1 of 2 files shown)


davidzhao (approved):

davidzhao:
exposing as an attribute sounds good. let's do that as a separate one

livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -619,6 +623,12 @@ def _parse_part(self, id: str, part: types.Part) -> llm.ChatChunk | None:
         if not part.text:
             return None
 
+        # Drop thought summaries: when include_thoughts is enabled Gemini streams its
+        # reasoning back as parts flagged part.thought=True. These must not reach
+        # ChoiceDelta.content, otherwise TTS would speak the model's internal reasoning.
+        if getattr(part, "thought", None):
```
theomonnom: do we need to use getattr? maybe it's an exposed attribute if we bump their sdk?

theomonnom (approved):

PR #6429, opened 2026-07-15
(deepgram stt): fix reconnect on send-side socket drops instead of crashing the session

tinalenguyen:
context: https://community.livekit.io/t/think-i-found-a-bug-in-livekit-plugins-deepgrams-stt/1596

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt_v2.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt_v2.py
@@ -437,25 +437,34 @@ class SpeechStreamv2(stt.SpeechStream):
             )
 
             has_ended = False
-            async for data in self._input_ch:
-                frames: list[rtc.AudioFrame] = []
-                if isinstance(data, rtc.AudioFrame):
-                    frames.extend(audio_bstream.write(data.data.tobytes()))
-                elif isinstance(data, self._FlushSentinel):
-                    frames.extend(audio_bstream.flush())
-                    has_ended = True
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt_v2.py
```diff
@@ -437,25 +437,34 @@ async def send_task(ws: aiohttp.ClientWebSocketResponse) -> None:
...
+
+                # tell deepgram we are done sending audio/inputs
+                closing_ws = True
+                await ws.send_str(SpeechStreamv2._CLOSE_MSG)
+            except (aiohttp.ClientError, ConnectionError) as e:
+                # a mid-write socket drop surfaces here as a raw connection error.
+                # if the close is expected (aclose or the http session closing) just
+                # return; otherwise re-raise as a retryable APIError so _main_task
+                # reconnects, symmetric with recv_task.
+                if closing_ws or self._session.closed:
```
davidzhao, line 465: would we run into the same error in recv_task? should the retry be a bit higher level from here?

is this deepgram specific? or should we fix the other plugins/inference too?

theomonnom (approved):

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt_v2.py
tinalenguyen: i believe this is send side specific, i think only some plugins are vulnerable to this. we should unify the design

PR #6431, opened 2026-07-15
fix(google): close realtime output streams on generation_complete

### LINKED ISSUE #6421: google.realtime.RealtimeModel sends AudioSegmentEnd too late
### Bug Description

Hi!
I'm a developer at Beyond Presence where we provide a virtual avatar for voice agents, see [this setup](https://docs.livekit.io/agents/models/avatar/#avatar-workers). We noticed an issue with the Gemini Realtime model where the `AudioSegmentEnd` is sent too late from either LiveKit or the Google RealtimeModel (Gemini) while OpenAI Realtime works fine. This makes it difficult for us to react on time and sometimes causes lags for the user. On some occasions, there is also a residual audio chunk, too.

### Expected Behavior

OpenAI sends audio data like this:
```
rtc.AudioFrame -> ... -> rtc.AudioFrame -> AudioSegmentEnd
```

Gemini on the other hand looks like this:

Case A: Without residual audio
```
rtc.AudioFrame -> ... -> rtc.AudioFrame -> [nothing for several seconds] -> AudioSegmentEnd
```

Case B: With residual audio
```
rtc.AudioFrame -> ... -> rtc.AudioFrame -> [nothing for several seconds] -> rtc.AudioFrame -> AudioSegmentEnd
```

### Reproduction Steps

```bash
1. Talk to an agent using Gemini Realtime model
2. Notice that all `rtc.AudioFrame` come in really fast except the last one
3. Notice that `AudioSegmentEnd` comes in too late. It should come in right after the last audio chunk.
```

### Operating System

Linux, macOS

### Models Used

google.realtime.RealtimeModel

### Package Versions

```bash
"livekit-agents == 1.6.5",
    "livekit-plugins-bey == 1.6.5",
    "livekit-plugins-google==1.6.5",
    "livekit-plugins-openai>=1.6.5",
    "python-dotenv ~= 1.0",
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
Right after the last audio chunk, a AudioSegmentEnd should be sent to indicate that the audio has finished.
```

### Additional Context

_No response_

I confirmed the issue in our logs. Notice that there is a 4s delay between the last audio chunk and the segment end

Image

longcw:
Gemini's realtime API delays `turn_complete` until it estimates client-side playback has finished, so ending the response on `turn_complete` made `AudioSegmentEnd` (and the finalized transcript) arrive several seconds after the last audio chunk. This closes the audio and text streams on `generation_complete` instead, matching the OpenAI realtime behavior. Full finalization (input transcription, metrics) still waits for `turn_complete`, and any audio/text received after `generation_complete` is dropped with a warning.

Fixes #6421 and https://github.com/livekit/agents/pull/5821#issuecomment-4958860133

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -178,8 +178,18 @@ class _ResponseGeneration:
     """The timestamp when the generation is completed"""
     _done: bool = False
     """Whether the generation is done (set when the turn is complete)"""
+    _extra_content_warned: bool = False
+    """Whether we've warned about audio/text arriving after generation completed"""
 
     def push_text(self, text: str) -> None:
+        if self.text_ch.closed:
+            # generation_complete already finalized the output; a turn should not emit
+            # more text, so drop it (see _handle_server_content)
+            if not self._extra_content_warned:
+                self._extra_content_warned = True
+                logger.warning("Gemini sent text after generation completed; dropping it")
+            return
+
         if self.output_text:
             self.output_text += text
         else:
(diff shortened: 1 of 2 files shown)


3DRX:
I tested this, it's working great now.

theomonnom (approved):

PR #6433, opened 2026-07-15
fix(soniox): sentence-buffer tts input, rotate streams on idle and age

chenghao-mou:
Fixes AGT-3049 (408s breaking the stream mid-reply); also addresses the open-stream 408 exposure described in AGT-3132.

**Why:** Soniox caps each TTS stream at 2 minutes and kills streams left waiting on slow LLM output, cutting speech mid-reply.

**What:** LLM chunks are now buffered into complete sentences (blingfire, overridable) and fed into one shared stream so prosody stays continuous. If no sentence arrives for `stream_idle_timeout` (default 5s), the stream is finalized with `text_end` before the server can time it out; the next sentence opens a fresh stream. Streams also rotate at a sentence boundary near the 2-minute cap, and a stream that fails before emitting any audio replays its sentence batch on a new `stream_id` (the framework never retries once audio reached the user).

Verified with a fake-connection harness (steady flow, idle rotation, batch replay); live end-to-end needs `SONIOX_API_KEY`.

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
@@ -20,7 +20,7 @@ import json
 import os
 import time
 import weakref
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, field, replace
 from typing import Any
 
 import aiohttp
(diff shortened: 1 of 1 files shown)


chenghao-mou:
Superseded by the correctly named branch, see the new PR from chenghao/fix/AGT-3049-soniox-tts-stream-timeout.

PR #6438, opened 2026-07-15
fix(plugins-aws): depend on aiobotocore directly instead of aioboto3

### LINKED ISSUE #6437: livekit-plugins-aws base dependency on aioboto3 transitively hard-pins botocore to a narrow window
### Summary

`livekit-plugins-aws` declares `aioboto3>=14.1.0` as a **base dependency**. The latest `aioboto3` (15.x) hard-pins `aiobotocore[boto3]==2.25.1`, which in turn constrains `botocore>=1.40.46,<1.40.62` (and `boto3` to the same window) — an ~15-patch range.

The practical effect: any application whose dependency graph already resolves `botocore` outside that window (e.g. 1.43.x) cannot install `livekit-plugins-aws`. In a monorepo with a single unified lockfile (uv / Poetry workspace), this makes the plugin uninstallable unless **every** other package is pinned back to that old botocore window — holding back the AWS SDK across the whole project.

For reference, the transitive chain is:

```
livekit-plugins-aws  ->  aioboto3>=14.1.0        (base dependency)
                              └─> aiobotocore[boto3]==2.25.1   (exact pin)
                                     └─> botocore >=1.40.46,<1.40.62
```

A compatible modern `aiobotocore` exists (3.7.0 accepts `botocore` 1.42.90–1.43.0), but `aioboto3`'s `==` pin refuses it.

### Questions

1. Why is `aioboto3` a base dependency of `livekit-plugins-aws`, rather than being gated behind the specific extras/features that actually use it?
2. Could the `aioboto3` dependency be loosened or made optional so it doesn't transitively pin `botocore`/`boto3` to a narrow window?
3. Is there a recommended way to use `livekit-plugins-aws` alongside a modern `botocore` (1.43.x)?

### Environment

- `livekit-plugins-aws` 1.6.5
- Python 3.12
- Resolving in a workspace that uses `botocore` 1.43.x

xavieralmendros-aily:
## Summary

`livekit-plugins-aws` depends on `aioboto3`, but `aioboto3` hard-pins its `aiobotocore` dependency with `==` (e.g. `aioboto3` 15.5.0 → `aiobotocore[boto3]==2.25.1`). Because each `aiobotocore` release in turn narrow-pins `botocore`/`boto3` to a ~15-patch window (2.25.1 → `botocore>=1.40.46,<1.40.62`), this transitively makes it impossible to install `livekit-plugins-aws` alongside any app that is on a more recent `botocore`. The resolver has exactly one `aiobotocore` release to choose from, and it drags an old `botocore` with it.

The plugin's actual usage of the AWS SDK is minimal: only the async **client** API is used — `session.client("polly", ...).synthesize_speech(...)` in `tts.py` and `session.client("bedrock-runtime", ...).converse_stream(...)` in `llm.py`. It never uses `.resource()`, which is the only thing `aioboto3` adds on top of `aiobotocore`. STT and the experimental realtime (Nova Sonic) paths use the smithy async SDKs, not `aioboto3` at all.

This PR drops `aioboto3` and uses `aiobotocore` directly:

- `aioboto3.Session(...)` → `aiobotocore.session.get_session()` (credentials/region applied via `set_credentials` / `set_config_variable`, matching the previous behaviour of only setting them when explicitly provided).
- `session.client(name, config=...)` → `session.create_client(name, config=...)` (same async-context-manager API).
- Dependency `aioboto3>=14.1.0` → `aiobotocore>=3.0.0`.

Because the dependency is now a **range** rather than a wrapper that `==`-pins a single old `aiobotocore`, the resolver can pick the `aiobotocore` whose `botocore` window matches the surrounding application (e.g. `aiobotocore` 3.7.0 accepts `botocore` 1.42.90–1.43.0).

## Compatibility note

The optional `session=` parameter on `aws.TTS` / `aws.LLM` changes type from `aioboto3.Session` to `aiobotocore.session.AioSession`. Callers relying on the default (no `session` passed) are unaffected. Callers passing their own session must pass an `aiobotocore` session instead.

## Test plan

- [ ] `tts.py` (Polly `synthesize_speech`) and `llm.py` (Bedrock `converse_stream`) smoke-tested with default credential chain.
- [ ] Verify install resolves against a modern `botocore` without backtracking to the old `aioboto3`-pinned window.

Refs #6437

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/utils.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/utils.py
@@ -37,7 +37,7 @@ def _resolve_session(session: object | None) -> AioSession:
             DeprecationWarning,
             stacklevel=3,
         )
-        return session._session  # type: ignore[attr-defined]
+        return session._session
 
     raise TypeError(
         "session must be an aiobotocore.session.AioSession "
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/utils.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/utils.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-import warnings
-
 from aiobotocore.session import AioSession, get_session  # type: ignore
+
+from .log import logger
 
 DEFAULT_REGION = "us-east-1"
 
@@ -30,12 +30,10 @@ def _resolve_session(session: object | None) -> AioSession:
         aioboto3 = None
 
     if aioboto3 is not None and isinstance(session, aioboto3.Session):
-        warnings.warn(
+        logger.warning(
             "Passing an aioboto3.Session is deprecated; pass an "
             "aiobotocore.session.AioSession instead. The AWS plugin no longer depends "
-            "on aioboto3.",
-            DeprecationWarning,
-            stacklevel=3,
+            "on aioboto3."
         )
         return session._session
 
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/utils.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/utils.py
@@ -1,7 +1,45 @@
 from __future__ import annotations
 
+import warnings
+
+from aiobotocore.session import AioSession, get_session  # type: ignore
+
 DEFAULT_REGION = "us-east-1"
 
 
 def _strip_nones(d: dict) -> dict:
     return {k: v for k, v in d.items() if v is not None}
+
+
+def _resolve_session(session: object | None) -> AioSession:
+    """Return an ``AioSession`` for the given ``session`` argument.
+
+    Accepts ``None`` (creates a fresh session), an ``aiobotocore.session.AioSession``,
+    or — for backwards compatibility — a legacy ``aioboto3.Session``, which wraps an
+    ``AioSession`` and is unwrapped here with a ``DeprecationWarning``. ``aioboto3`` is
+    imported lazily so it does not become a dependency of this plugin.
+    """
+    if session is None:
+        return get_session()
+    if isinstance(session, AioSession):
+        return session
+
+    try:
+        import aioboto3  # type: ignore
+    except ImportError:
+        aioboto3 = None
+
+    if aioboto3 is not None and isinstance(session, aioboto3.Session):
+        warnings.warn(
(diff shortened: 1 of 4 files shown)


tinalenguyen (commented):
hi, thanks for the PR! should we widen the type to still accept the `aioboto3.session` and throw a warning to switch to `AioSession`? by default we can use the `AioSession`

xavieralmendros-aily:
Hey @tinalenguyen  updated with your comments! Thanks for the fast answer 🚀

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/utils.py
```diff
@@ -1,7 +1,45 @@
...
+    if isinstance(session, AioSession):
+        return session
+
+    try:
+        import aioboto3  # type: ignore
+    except ImportError:
+        aioboto3 = None
+
+    if aioboto3 is not None and isinstance(session, aioboto3.Session):
+        warnings.warn(
```
tinalenguyen: could we use `logger.warning` instead?

xavieralmendros-aily:
Hi @tinalenguyen done :) , switched the deprecation notice from `warnings.warn(...)` to `logger.warning(...) `using the plugin's `livekit.plugins.aws `logger (pushed in `8175e1441`).

Heads-up on the failing check-types CI: it's not coming from this PR. The check runs against the merge with main, and main currently has a type error unrelated to the AWS plugin:

`livekit-agents/livekit/agents/voice/agent_session.py:1183: error: "AgentSessionOptions" has no attribute "expressive" [attr-defined]`

At that line `self._opts.expressive = cast("bool | ExpressiveOptions", expressive)`, but the AgentSessionOptions dataclass has no expressive field,  the attribute set elsewhere is `self._expressive `(line 440). This PR only touches `livekit/plugins/aws/utils.py`, so once main is fixed a re-run should be green. 

Thanks!

tinalenguyen (approved):
thank you again for the context and iterating on this @xavieralmendros-aily! :D

xavieralmendros-aily:
@tinalenguyen  thanks to you! It's been a pleasure! btw do you know when this fix will be released ?

tinalenguyen:
@xavieralmendros-aily likewise!! we're doing a release today, sometime in the next couple hours

xavieralmendros-aily:
@tinalenguyen  amazing!!!!!! thanks 🚀

PR #6439, opened 2026-07-15
fix: avoid mid-word splits in text_transforms.replace

longcw:
`replace` held back a fixed `max(len(key)) - 1` trailing characters on every chunk to catch keys split across streaming boundaries, so ordinary words were emitted split mid-word (e.g. "connect." arrived as "conne" + "ct.").

It now buffers only a trailing run that could still start a replacement key and forwards the rest immediately. Substitution uses a single compiled alternation, which also avoids cascading replacements and prefers the longest match on overlapping keys.


theomonnom (approved):
wow nice catch

PR #6440, opened 2026-07-15
feat(agent): add update_options() to swap STT/VAD/LLM/TTS at runtime

longcw:
`Agent.update_models(*, stt, vad, llm, tts)` swaps a model in place. Only the models passed are changed; strings resolve to inference models like the constructor, and `None` disables a model (overriding the session), matching `Agent(stt=None)`.

When the agent isn't running it replaces the stored model, applied on the next start. When running, it applies to the live pipeline: the STT and VAD streams are rewired, the LLM and TTS take effect on the next generation and synthesis respectively, and each model's `metrics_collected`/`error` listeners follow the new instance. The call is synchronous and atomic.

The Agent is the source of truth — the activity resolves `agent.<model> if given else session.<model>` live — so `update_models` just writes to the agent. This is distinct from `AgentSession.update_agent` (full handoff, new instance) and from a model's own `update_options` (same instance, provider options).

Alternative to https://github.com/livekit/agents/pull/6235

## Realtime models

A `RealtimeModel` owns a live session created when the agent starts, so it can't be swapped in place. While running, passing a `RealtimeModel` as `llm` (or swapping away from one) raises `RuntimeError` pointing to `AgentSession.update_agent`. Validation happens before any mutation, so the call stays all-or-nothing.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent.py
+++ b/livekit-agents/livekit/agents/voice/agent.py
@@ -260,7 +260,7 @@ class Agent:
             chat_ctx, exclude_invalid_function_calls=exclude_invalid_function_calls
         )
 
-    def update_models(
+    def update_options(
         self,
         *,
         stt: NotGivenOr[stt.STT | STTModels | str | None] = NOT_GIVEN,
--- a/livekit-agents/livekit/agents/voice/agent.py
+++ b/livekit-agents/livekit/agents/voice/agent.py
@@ -260,6 +260,46 @@ class Agent:
             chat_ctx, exclude_invalid_function_calls=exclude_invalid_function_calls
         )
 
+    def update_models(
+        self,
+        *,
+        stt: NotGivenOr[stt.STT | STTModels | str | None] = NOT_GIVEN,
+        vad: NotGivenOr[vad.VAD | None] = NOT_GIVEN,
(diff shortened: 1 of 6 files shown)


livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -260,6 +260,46 @@ async def update_chat_ctx(
             chat_ctx, exclude_invalid_function_calls=exclude_invalid_function_calls
         )
 
+    def update_models(
```
theomonnom: What do you think about naming it `update_options` to keep it consistent with how we do it in other parts of the codebase?

theomonnom (approved):

theomonnom:
Just double checking, when switching to another TTS mid-conversation? Will an "active" generation still use the old TTS, or the new one?

longcw:
> Just double checking, when switching to another TTS mid-conversation? Will an "active" generation still use the old TTS, or the new one?

update the TTS won't cancel the active generation, the active one still uses the old TTS.

livekit-agents/livekit/agents/voice/agent.py
longcw: updated to `update_options` for Agent, agent activity keeps `_update_models`

PR #6443, opened 2026-07-15
feat(assemblyai): add language_codes steering param, enable agent_context_carryover by default

dlange-aai:
## What

Two related upgrades to the AssemblyAI streaming STT plugin:

### 1. `language_codes` — multi-language steering

- New `language_codes: list[str]` parameter: `language_codes=["en", "es"]` biases transcription toward that set of languages instead of open detection/code-switching (Universal-3 Pro family only). The existing singular `language_code` keeps working unchanged as documented shorthand for a one-element list; passing both raises `ValueError`.
- Each entry is normalized to a bare ISO 639-1 code (`"en-US"` / `"english"` → `"en"`) and deduped preserving order; server limits are validated client-side (max 10 codes, `"multi"` must be sent alone) so misconfiguration fails at construction instead of as a websocket error.
- Mid-session re-steering via `STT.update_options(language_codes=...)` / `SpeechStream.update_options(...)`, sent as an `UpdateConfiguration` message; an explicit `[]` clears steering back to the model default. Useful for agents that detect the caller's language and re-steer without reconnecting. Validation runs before any option mutation, so a rejected update never leaves options partially applied.
- Wire format: sent as a JSON-array query param (`language_codes=["en","es"]`), the same convention as `keyterms_prompt`. **Note for existing `language_code` users:** the singular is now also sent as `language_codes=["es"]` — behaviorally identical, since the server declares `language_code`/`language_codes`/`language` as aliases for the same field.

### 2. `agent_context_carryover` — on by default, with a server-cap guard

- **Fixes a live-session kill:** the server rejects `agent_context` longer than 1750 chars, and an invalid mid-stream `UpdateConfiguration` cancels the entire streaming session. The plugin now enforces the cap client-side: the auto-carryover handler truncates oversize assistant replies (keeping the tail — the end of the reply is what the user responds to), and explicit `agent_context` input fails fast with `ValueError`. Also fixes the stale "max 1500 characters" docstring (server cap is 1750).
- With the guard in place, `agent_context_carryover` defaults to **enabled on Universal-3 Pro family models** (tri-state `NotGivenOr[bool]`): the server already carries prior user turns automatically, so assistant-reply carryover completes the conversation context the models are designed to use. Opt out with `agent_context_carryover=False`; setting `previous_context_n_turns=0` (documented as "disable context carryover") also suppresses the default. On unsupported models the default is silently off; explicitly passing `True` there keeps today's warning.
- **Behavior change to be aware of:** on supported models, assistant reply text now flows to AssemblyAI as transcription context by default, and each reply replaces any manually set `agent_context` (documented in the docstring; disable carryover to manage context manually).

## Testing

- `tests/test_plugin_assemblyai_stt.py`: 43 new tests — normalization/dedup, mutual exclusion, U3-Pro gating on both construction and update paths, max-10/multi validation, JSON-array wire format for both spellings, `UpdateConfiguration` enqueue incl. empty-list clear, no-partial-update on validation failure, STT→stream propagation, truncation boundaries at exactly 1750/1751, tail-keeping, the four tri-state carryover cases, and handler safety for user/textless/handoff conversation items. Full file: 101 passed; ruff and strict mypy clean.
- Verified live against the AssemblyAI streaming API (console-mode voice agent, full multi-turn booking flow): steering param accepted at connect, 13 carryover context pushes and mid-stream config updates over one session with no errors or reconnects.

### DIFF
--- a/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
+++ b/livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
@@ -105,8 +105,8 @@ def _normalize_language_codes(language_codes: list[str]) -> list[str]:
         )
     if "multi" in normalized and len(normalized) > 1:
         raise ValueError(
-            "language_code 'multi' routes to the unsteered multilingual model "
-            "and cannot be combined with other codes"
+            "'multi' routes to the unsteered multilingual model "
+            "and cannot be combined with other language codes"
         )
     return normalized
 
@@ -117,14 +117,13 @@ def _normalize_language_codes(language_codes: list[str]) -> list[str]:
 _MAX_AGENT_CONTEXT_CHARS = 1750
 
 
-def _validate_agent_context(agent_context: str) -> str:
+def _validate_agent_context(agent_context: str) -> None:
     """Reject explicit agent_context longer than the server cap."""
     if len(agent_context) > _MAX_AGENT_CONTEXT_CHARS:
         raise ValueError(
             f"agent_context exceeds maximum length of {_MAX_AGENT_CONTEXT_CHARS} "
             f"characters (got {len(agent_context)})"
         )
-    return agent_context
 
 
 class STT(stt.STT):
@@ -192,7 +191,8 @@ class STT(stt.STT):
                 Leave unset to use the model's default multilingual behavior. Only
                 supported with the Universal-3 Pro family models. Can be updated
                 mid-session via ``update_options``; pass an empty list there to clear
-                steering back to the model default.
+                steering back to the model default. At construction an empty list is
+                equivalent to leaving it unset.
             min_turn_silence: Minimum silence in ms before a confident end-of-turn is finalized.
             min_end_of_turn_silence_when_confident: Deprecated. Use min_turn_silence instead.
             continuous_partials: Whether to emit additional partial transcripts during long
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-assemblyai/livekit/plugins/assemblyai/stt.py
```diff
@@ -104,6 +143,7 @@ def __init__(
         ] = "universal-3-5-pro",
         language_detection: NotGivenOr[bool] = NOT_GIVEN,
         language_code: NotGivenOr[str] = NOT_GIVEN,
+        language_codes: NotGivenOr[list[str]] = NOT_GIVEN,
```
tinalenguyen: what do you think of deprecating `language` and throwing a warning, and then widening the type like:

`language_codes: NotGivenOr[str | list[str]] = NOT_GIVEN,`

dlange-aai: @tinalenguyen do you mean deprecate the singular 'language_code' parameter?

dlange-aai: @tinalenguyen good call out, thank you! Latest commit should be good now!

tinalenguyen (approved):

PR #6444, opened 2026-07-15
fix(llm): apply tool defaults across all schema shapes; keep response formats sentinel-free

u9g:
## Summary

Strict JSON schemas require every property and don't support the `default` keyword. This PR settles how defaulted fields are handled on each of the two encode paths.

**Tool schemas keep the null sentinel, and it now works for every schema shape.** A defaulted field additionally accepts `null`, meaning "use the Python default" — the LLM never has to invent a value it can't know. This is safe because the framework controls the decode: `_prepare_function_arguments` resolves the sentinel before validation.

- The sentinel is advertised for all shapes, not just bare types: `Literal` (const/enum), unions (`anyOf`), and `$ref`-typed defaults (nested models, enums) previously either contradicted themselves or never became nullable.
- Resolution is now recursive: nested model arguments, `dict[str, Model]` values, and tuple items are decoded (previously a top-level-only signature walk).
- Union variants are rejected when the payload carries keys outside their declared `properties`, matching the closed-object semantics of the strict wire schema — prevents a payload meant for variant B from silently matching variant A and dropping data.
- Genuine nulls declared via enum membership (`Literal["x", None]`) are preserved instead of being replaced with the default.
- When wrapping a defaulted `$ref`-typed field (e.g. a documented nested-model param), sibling keys like `description` stay on the wrapper — strict mode forbids siblings on `$ref` but allows them next to `anyOf`.

**Response format schemas get no sentinel.** `to_openai_response_format` output is decoded by users with plain Pydantic validation, so the wire schema must be exactly what validation accepts. Defaulted fields stay required and non-nullable; the model produces their values and the Python default never fires. `to_strict_json_schema` takes a `null_sentinel_for_defaults` flag (tools pass it; response formats don't).

## Behavior changes

- `response_format`: the model can no longer send `null` for a defaulted non-nullable field. Previously bare-type defaulted fields were widened to `["type", "null"]`, producing payloads that plain `model_validate` rejects.
- Tool args: a `null` for a required parameter with no default now surfaces as a Pydantic `ValidationError` (still wrapped in `ToolError`) instead of a hand-written `ValueError`.

## Known limitation

`Field(default_factory=...)` emits no schema `default`, so factory-defaulted tool fields get no sentinel and remain required on the wire. Left for a follow-up.

## Testing

- 134 tests across the response-format, tools, gemini-schema, tool-proxy, and placeholder suites (new regression tests written to fail first)
- Ruff check/format clean; full repo type check (618 files) passes

### DIFF
--- a/livekit-agents/livekit/agents/llm/_strict.py
+++ b/livekit-agents/livekit/agents/llm/_strict.py
@@ -7,46 +7,13 @@ from pydantic import BaseModel, TypeAdapter
 _T = TypeVar("_T")
 
 
-def _model_json_schema(model: type[BaseModel] | TypeAdapter[Any]) -> dict[str, Any]:
-    if isinstance(model, TypeAdapter):
-        return model.json_schema()
-    return model.model_json_schema()
-
-
 def to_strict_json_schema(model: type[BaseModel] | TypeAdapter[Any]) -> dict[str, Any]:
-    schema = _model_json_schema(model)
+    if isinstance(model, TypeAdapter):
+        schema = model.json_schema()
+    else:
+        schema = model.model_json_schema()
 
     return _ensure_strict_json_schema(schema, path=(), root=schema)
-
-
-def to_strict_tool_json_schema(model: type[BaseModel] | TypeAdapter[Any]) -> dict[str, Any]:
-    schema = _model_json_schema(model)
-    _make_defaults_nullable(schema)
-
-    return _ensure_strict_json_schema(schema, path=(), root=schema)
-
-
-def _make_defaults_nullable(json_schema: object) -> None:
-    """Encode defaulted tool arguments as nullable fields for strict schemas.
-
-    OpenAI strict schemas require every property and do not support JSON Schema defaults, so
-    ``null`` represents "use the Python default". Tool argument preparation replaces that value
-    before validation. Response formats intentionally skip this conversion because their output
-    is validated directly against the declared Pydantic nullability.
-    """
-    if isinstance(json_schema, dict):
-        if "default" in json_schema:
-            typ = json_schema.get("type")
-            if isinstance(typ, str):
-                json_schema["type"] = [typ, "null"]
-            elif isinstance(typ, list) and "null" not in typ:
-                json_schema["type"] = [*typ, "null"]
-
-        for value in json_schema.values():
-            _make_defaults_nullable(value)
-    elif isinstance(json_schema, list):
-        for value in json_schema:
-            _make_defaults_nullable(value)
 
 
 # from https://platform.openai.com/docs/guides/function-calling?api-mode=responses&strict-mode=disabled#strict-mode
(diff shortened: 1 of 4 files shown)


livekit-agents/livekit/agents/llm/_strict.py
```diff
@@ -105,32 +158,27 @@ def _ensure_strict_json_schema(
...
             )
             json_schema.pop("allOf")
         else:
             json_schema["allOf"] = [
-                _ensure_strict_json_schema(entry, path=(*path, "allOf", str(i)), root=root)
+                _ensure_strict_json_schema(
+                    entry,
+                    path=(*path, "allOf", str(i)),
+                    root=root,
+                )
```
theomonnom, line 166: Let's keep one line, there is a formatting issue

livekit-agents/livekit/agents/llm/_strict.py
```diff
@@ -105,32 +158,27 @@ def _ensure_strict_json_schema(
...
-        t = json_schema.get("type")
-        if isinstance(t, str):
-            json_schema["type"] = [t, "null"]
-
-        elif isinstance(t, list):
-            types = t.copy()
-            if "null" not in types:
-                types.append("null")
-
-            json_schema["type"] = types
```
theomonnom, line 132: Can you verify it doesn't break existing behaviour? I think the reasoning behind that is that users using default values on parameters don't want to require the LLMs to generate a value for it

livekit-agents/livekit/agents/llm/_strict.py
u9g: done

livekit-agents/livekit/agents/llm/_strict.py
u9g: Updated in 80a65382a. I kept the existing default-as-null schema behavior, so the LLM is still not required to generate a value it cannot know. The new validate_response_format helper recursively replaces null with the Pydantic default before validation, while preserving null for fields whose declared type is actually nullable. The focused response-format and tool suites pass (101 tests).

Bobronium: It seems that `validate_response_format` is not used anywhere, except tests?

u9g: Sorry this was an LLM comment, after this comment was sent, I added to claude code a manual guard to never autosend comments without human approval. Me and theo chatted offline. The idea that we agreed on is that we will make the schema for defaults be type | null and then if the model gives null then we will put in the default value.

Bobronium: The intent is clear, thanks! My comment was about `validate_response_format` being defined, but not used — it's in the PR changes, not just comment above.

P.S. I should add such guard as well — Claude has also needlessly commented once on my behalf.

livekit-agents/livekit/agents/llm/_strict.py
```diff
@@ -184,6 +234,43 @@ def _ensure_strict_json_schema(
...
+        json_schema["type"] = [*typ, "null"]
+        return
+
+    # const / enum / $ref / allOf: wrap so null is a valid instance without violating
+    # the original constraints. An empty schema already accepts null — leave it.
+    inner = dict(json_schema)
+    if not inner:
+        return
+    json_schema.clear()
+    json_schema["anyOf"] = [inner, {"type": "null"}]
```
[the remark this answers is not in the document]
u9g: done.

livekit-agents/livekit/agents/llm/_strict.py
u9g: You are right. This PR was hanging on the assumption that people would call a different function to normalize the nulls out of their structured outputs into the defaults. We no longer do that.

theomonnom (approved):

PR #6445, opened 2026-07-15
fix(voice): don't drop the realtime turn when the chat-ctx push times out

ByteMaster-1:
Incident. Calling session.generate_reply(user_input=...) on a realtime session pushes the message via update_chat_ctx before requesting the reply. That await was unguarded: when the OpenAI plugin's 5s ack timeout fired (RealtimeError("update_chat_ctx timed out.")), the error propagated as an unhandled exception in the turn task — the turn died, the user got silence, and nothing was reported on the SpeechHandle.

Fix. The push is best-effort — the plugin sends all conversation.item.create events before waiting for acks, so a timeout means missing acknowledgements, not missing items. So:

RealtimeError → log a warning and still call generate_reply (the server usually has the context).
Unexpected exceptions → mark the SpeechHandle done with the error, observable via SpeechHandle.exception() #6304 , without crashing the turn task.
Deliberately out of scope: no retry of the push itself, and no change to the plugin's timeout; the two other update_chat_ctx call sites (interrupted-sync and tool-output paths) already have their own RealtimeError handling and are untouched.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -3387,7 +3387,19 @@ class AgentActivity(RecognitionHooks):
         if user_input is not None:
             chat_ctx = self._rt_session.chat_ctx.copy()
             msg = chat_ctx.add_message(role="user", content=user_input)
-            await self._rt_session.update_chat_ctx(chat_ctx)
+            try:
+                await self._rt_session.update_chat_ctx(chat_ctx)
+            except llm.RealtimeError as e:
+                # the push is best-effort (the items were sent; only the ack timed out),
+                # so still generate the reply rather than dropping the whole turn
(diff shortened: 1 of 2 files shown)


ByteMaster-1:
@longcw This is the second follow-up from the #6352  rescope: the update_chat_ctx push inside generate_reply no longer kills the turn on the plugin's ack timeout (best-effort — the items were already sent), and unexpected failures land on SpeechHandle.exception() instead of crashing the turn task. Small diff + unit tests. PTAL when you get a chance.

ByteMaster-1:
@davidzhao @longcw Gentle bump on this one too — the second #6352 follow-up (best-effort update_chat_ctx push inside generate_reply). Small diff, CI green; happy to rework if you'd prefer a different shape.

longcw (approved):

PR #6447, opened 2026-07-15
fix(workflows): unstall credit card repeat-back confirmation

chenghao-mou:
During repeat-back confirmation the LLM often routes the repeated value back to `update_*` instead of `confirm_*`; the task re-arms confirmation, the model then claims the value is confirmed without any tool call, and the task group stalls (dead air until the caller speaks again).

`update_*` now raises a ToolError redirecting the model to `confirm_*` when it receives the pending value again, all three instruction templates get the explicit tool-invocation guard from `name.py`, and the update return strings cover the late-update case where the caller already repeated the value. A different value still re-arms confirmation, so mid-confirmation corrections are unchanged.

**Known gap:** if the model never calls `update_*` until after the repeat (or calls no tool at all), the task can still stall; prompt guidance alone did not fix that for gpt-5.2 at low reasoning effort. Structural fix (forced `tool_choice` while confirmation is pending) tracked in AGT-3140.

Repro used for both the bug and verification: `lk agent daemon` text session against `GetCreditCardTask(require_confirmation=True)`.

Fixes AGT-3139

### DIFF
--- a/livekit-agents/livekit/agents/beta/workflows/credit_card.py
+++ b/livekit-agents/livekit/agents/beta/workflows/credit_card.py
@@ -25,6 +25,7 @@ You are solely responsible for collecting the credit card number.
 {modality_specific}
 If the user refuses to provide a credit card number, call decline_card_capture().
 If the user wishes to start over the credit card collection process, call restart_card_collection().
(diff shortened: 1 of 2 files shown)


tinalenguyen (approved):

chenghao-mou:
Superseded by #6487, which incorporates the confirmation-routing fix into the stateful credit-card capture change.

PR #6448, opened 2026-07-15
fix(workflows): support chunked card number reading

chenghao-mou:
Callers naturally read card numbers in groups and expect acknowledgment between them ("the first four are 4 5 3 9"); GetCardNumberTask previously pushed them to read the whole number in one utterance (`update_card_number` forbids incremental calls and the instructions had no grouped-capture guidance).

Prompt-level fix: the audio instructions now describe group-wise collection with brief, digit-free acknowledgments (the agent never utters digits, per the existing PCI constraint in the base instructions), and the tool docstring instructs concatenating the groups in order into a single call. Verified via `lk agent daemon` with four distinct groups: correct in-order concatenation, Luhn pass, caller repeat-back confirmation, and a mid-flow restart. No stateful chunk-buffer tool needed.

Stacked on #6447 (base branch `chenghao/fix/AGT-3139-credit-card-confirm-routing`); will retarget to main once that merges.

Fixes AGT-3141


theomonnom (approved):

chenghao-mou:
Superseded by #6487, which moves incremental card-number accumulation and confirmation into a server-owned state machine.

PR #6454, opened 2026-07-16
Retry OpenAI realtime responses that fail before output

### LINKED ISSUE #6205: Realtime models have no response-level retry for recoverable generate_reply failures (parity gap with the pipeline LLM)

RitwijParmar:
Fixes #6205.

OpenAI realtime can return response.done with status failed after response.created but before any output item starts. Before this change, that closes the generation and the agent turn can disappear without another response attempt.

This retries that narrow case with the existing connection retry policy:
- only failed responses
- only before output starts
- only retryable errors
- bounded by max_retry
- reuses the same generation streams so callers do not get a duplicate generation event

I did not retry after output starts. That needs a separate policy for partial audio/text cleanup and transcript history.

Checked locally:
- uv run pytest tests/test_realtime/test_openai_realtime_model.py tests/test_openai_realtime_chat_ctx.py -q
- uv run ruff check livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py tests/test_realtime/test_openai_realtime_model.py
- uv run ruff format --check livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py tests/test_realtime/test_openai_realtime_model.py
- uv run mypy livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -264,6 +264,9 @@ class _ResponseGeneration:
     """timestamp when the response was created"""
     _first_token_timestamp: float | None = None
     """timestamp when the first token was received"""
+    response_create_params: RealtimeResponseCreateParams | None = None
+    retry_count: int = 0
+    output_started: bool = False
 
     def _close(self) -> None:
         for msg in self.messages.values():
@@ -297,6 +300,26 @@ _FATAL_ERROR_CODES = frozenset(
 def _is_fatal_error(error: object | None) -> bool:
     code = getattr(error, "code", None) or getattr(error, "type", None)
     return isinstance(code, str) and code in _FATAL_ERROR_CODES
+
+
+def _response_done_error(*, provider_label: str, event: ResponseDoneEvent) -> APIError | None:
+    if event.response.status != "failed":
+        return None
+
+    if event.response.status_details and hasattr(event.response.status_details, "error"):
+        error_type = getattr(event.response.status_details.error, "type", "unknown")
+        error_body = event.response.status_details.error
+        message = f"{provider_label} response failed with error type: {error_type}"
+    else:
+        error_body = None
+        message = f"{provider_label} response failed with unknown error"
+
+    recoverable = not _is_fatal_error(error_body)
+    return APIError(
+        message=message,
+        body=error_body,
+        retryable=recoverable,
+    )
 
 
 class RealtimeModel(llm.RealtimeModel):
@@ -857,12 +880,14 @@ class RealtimeSession(
         self.send_event(self._create_session_update_event())
 
         self._response_created_futures: dict[str, asyncio.Future[llm.GenerationCreatedEvent]] = {}
+        self._response_create_params: dict[str, RealtimeResponseCreateParams] = {}
         self._item_delete_future: dict[str, asyncio.Future] = {}
         self._item_create_future: dict[str, asyncio.Future] = {}
 
         # generate_reply event_ids cancelled or timed out before response.created arrived; the
         # response is cancelled by id and discarded when it finally arrives
         self._discarded_event_ids: set[str] = set()
+        self._response_retry_event_ids: set[str] = set()
 
         # accumulates partial input-audio transcripts per (item_id, content_index)
         self._input_transcript_accumulators: dict[str, dict[int, str]] = {}
@@ -1626,12 +1651,12 @@ class RealtimeSession(
         if is_given(tools):
             params.tools = self._convert_tools_to_oai(tools)  # type: ignore
 
-        self.send_event(
-            ResponseCreateEvent(type="response.create", event_id=event_id, response=params)
-        )
+        self._response_create_params[event_id] = params
+        self._send_response_create(event_id=event_id, params=params)
 
         def _on_timeout() -> None:
             self._response_created_futures.pop(event_id, None)
+            self._response_create_params.pop(event_id, None)
             if fut and not fut.done():
                 # discard the response if the server still creates it after the timeout
                 self._discarded_event_ids.add(event_id)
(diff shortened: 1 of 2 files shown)


RitwijParmar:
One thing I would want to pin down before this grows is the generation identity across a retry. Reusing streams avoids a duplicate generation in the caller, but the trace and metrics still need to say whether the first attempt failed before output and the second attempt completed it. I would carry one generation id with an attempt number and emit the retry reason once. Then a failed first attempt cannot look like two user generations in dashboards. Is that useful for the direction here?

PR #6461, opened 2026-07-17
fix(elevenlabs): populate SpeechData.confidence from Scribe logprobs

captainbanan:
## What / Why

The ElevenLabs STT plugin never populates `SpeechData.confidence`, so it always stays at its `0.0` default — even though Scribe returns a per-word `logprob` in every response. Consumers that use `confidence` (e.g. to tell a clean transcript from a low-confidence / garbled one) get no signal from ElevenLabs.

Verified live: both `scribe_v1`/`scribe_v2` (REST) and `scribe_v2_realtime` return a natural-log `logprob` per `words[]` token, but `_process_stream_event` / `_transcription_to_speech_event` read only `text` + word `start`/`end`, so it is discarded and `confidence` stays `0.0`.

## Change

Add `_speech_confidence(words)`: average the spoken-word (`type == "word"`) logprobs and exponentiate → a `[0, 1]` confidence (geometric mean of the token probabilities). Returns `0.0` when per-word logprobs aren't available (e.g. a non-timestamped commit), so behaviour is unchanged there.

Wired into both the REST and WebSocket `SpeechData` construction.

Note: this deliberately does **not** use Scribe's `language_probability` — that is language-identification confidence, a different quantity from transcription confidence, so folding it into `SpeechData.confidence` would be misleading.

Same class of fix as #5829 / #5830 (a sibling STT plugin not threading its confidence into `SpeechData`).

## Tests

Unit tests for the helper (confident vs low-confidence logprobs, no-logprob → `0.0`) and an end-to-end check that a committed transcript sets `confidence`. `ruff check` / `ruff format` clean.

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import asyncio
 import base64
 import json
+import math
 import os
 import weakref
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


captainbanan:
@darryncampbell 
Hey! Could you check it please

davidzhao (approved):

PR #6463, opened 2026-07-17
fix(drive-thru): assume quantity of one when unstated to avoid count confirmation

u9g:
## Summary

`test_consecutive_order` has flaked twice on main in the last week (runs [29438348555](https://github.com/livekit/agents/actions/runs/29438348555) and [29261765536](https://github.com/livekit/agents/actions/runs/29261765536), both on commits unrelated to the drive-thru eval). In both cases the model responded to "Can I get a keychup sauce and a McFlurry Oreo ?" with a clarifying question ("Just to check, did you want just one of each?") instead of calling `order_regular_item`, so `contains_function_call` found no tool call.

#6383 fixed the stated-quantity variant of this ("two mayonnaise sauces" → add twice without confirming). This extends the same instruction bullet to cover the unstated-quantity case: if no quantity is stated, assume one and add the item without confirming the count.

## Testing

Ran `examples/drive-thru/test_agent.py::test_consecutive_order` 4 times locally with the change — 4/4 passed.


examples/drive-thru/database.py
```diff
@@ -43,7 +43,8 @@
     "duplicates. \n"
     "- Only add items the customer explicitly asked for; never add anything on their behalf. \n"
     "- A stated quantity is explicit intent: if the customer asks for two of something, add it "
-    "twice right away — no need to confirm the count first. \n"
+    "twice right away — no need to confirm the count first. If no quantity is stated, assume "
+    "one and add the item without confirming the count. \n"
     "- Don’t assume unstated details — especially the drink in a combo. If a required detail is "
```
chenghao-mou: nit: Should we have a carve-out here to exclude quantity?

chenghao-mou (approved):

examples/drive-thru/database.py
u9g: This is a great idea, honestly I didn't think of it. I just implemented it though, it's a better path forward imo.

u9g:
I a/b tested whether the "other than quantity" carve-out alone is enough without the explicit assume-one sentence (across 40 runs of `test_consecutive_order` on gpt-4.1 @ temp 0.45). Carve-out-only failed 5/40 times with the same reason as CI failed before this PR vs 0/40 with this PR and 1/20 for the baseline.

PR #6466, opened 2026-07-17
(assemblyai inference): support agent_context_carryover

tinalenguyen:
already introduced in the assembly ai plugin

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -546,15 +546,8 @@ class STT(stt.STT):
             api_secret (str, optional): LIVEKIT_API_SECRET, if not provided, read from environment variable.
             http_session (aiohttp.ClientSession, optional): HTTP session to use.
             extra_kwargs (dict, optional): Extra kwargs to pass to the STT model.
-            agent_context_carryover (bool, optional): When the model supports it, let an
-                ``AgentSession`` push each assistant reply into AssemblyAI's ``agent_context``
-                so it biases recognition of the user's next turn. Enabled by default on models
-                that support it (the AssemblyAI Universal-3 Pro family: ``assemblyai/u3-rt-pro``,
-                ``assemblyai/universal-3-5-pro``); pass False to opt out. On other models it is
-                off; explicitly passing True logs a warning and is ignored. Also disabled by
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -718,6 +755,21 @@ def _update_session_keyterms(self, keyterms: list[str]) -> None:
             else:
                 stream.update_options(extra=keyterm_extra)
 
+    def _push_conversation_item(self, ev: ConversationItemAddedEvent) -> None:
+        if (
```
chenghao-mou, line 759: should we check if chat context is enabled here? The current content is assembly specific and if we have other implementation needed, we should have vendor + feature specific conditions

tinalenguyen: i believe it's already gated in `agent_activity.py`: 
https://github.com/livekit/agents/blob/e7ac7f07d67ee78bcc9e9e1050e828b5a2b7b688/livekit-agents/livekit/agents/voice/agent_activity.py#L611-L614

chenghao-mou: should we check if assembly ai model is used too? I know this is the only supported vendor for now, so it's fine to skip it now.

tinalenguyen: do you mean that we should add another check besides this one: https://github.com/livekit/agents/pull/6466/changes#diff-6fb78e0bc4c1a972691de905aada78e293c2d797083d5a488c12ec3cd7843f64R251-R253

since other providers might have different names for the context field in the future?

chenghao-mou: I am referring to a case where, for example, Cartesia support this feature later, but it needs first sentence of each agent message. Then this callback would need to differentiate how to process the data based on vendor + model, since the caller only has the `chat_context` flag.

tinalenguyen: ahh okay, i see

chenghao-mou (approved):
LGTM. I don't have a strong opinion on the default enablement here. We can always move toward having it on by default or bundling with keyterm extraction once we have our eval results.

tinalenguyen:
thanks @chenghao-mou! 

> We can always move toward having it on by default


just wanted to note that it is on by default here with the compatible models, not sure if this is what you meant

PR #6468, opened 2026-07-17
feat(gnani): update TTS voices/models and STT sample rates + streaming toggle

tinalenguyen:


livekit-plugins/livekit-plugins-gnani/livekit/plugins/gnani/tts.py
```diff
@@ -101,15 +85,164 @@ def _check_deprecated_tts_args(kwargs: dict[str, Any], *, caller: str = "TTS.__i
         )
 
 
+def _strip_wav_header(data: bytes) -> bytes:
```
chenghao-mou: maybe we can use `AudioStreamDecoder`?

livekit-plugins/livekit-plugins-gnani/livekit/plugins/gnani/models.py
```diff
@@ -0,0 +1,56 @@
...
+    "mr-IN",
+    "bn-IN",
+    "gu-IN",
+    "pa-IN",
+]
+"""``language`` is honored by ``timbre-v2.5`` only. See
+https://docs.gnani.ai/api/TTS/tts-inference#supported-languages"""
+
+GnaniTTSEncodings = Literal["linear_pcm", "oggopus"]
+GnaniTTSContainers = Literal["raw", "mp3", "wav", "mulaw", "ogg"]
```
chenghao-mou: I don't think we support mulaw in the emitter. We can drop it.

livekit-plugins/livekit-plugins-gnani/livekit/plugins/gnani/utils.py
```diff
@@ -0,0 +1,39 @@
...
+
+def ws_header_kwargs(headers: dict[str, str]) -> dict[str, Any]:
+    """Return the correct ``connect()`` header kwarg for the installed websockets.
+
+    The new asyncio implementation uses ``additional_headers`` while the legacy
+    client uses ``extra_headers``. The top-level ``websockets.connect`` alias
+    (used by the STT/TTS clients) only switched to the asyncio implementation in
+    websockets 14.0 — in 12.x and 13.x it is still the legacy client. Select the
+    kwarg by that boundary so WebSocket STT/TTS works across ``websockets>=12``.
+    """
```
chenghao-mou: Q: if we enforce >=14.0 in this plugin, we then don't need this file, right?

chenghao-mou (approved):
Two comments. Otherwise LGTM. Tested locally with both STT and TTS.

livekit-plugins/livekit-plugins-gnani/livekit/plugins/gnani/utils.py
tinalenguyen: great callout, will bump and remove

livekit-plugins/livekit-plugins-gnani/livekit/plugins/gnani/tts.py
```diff
@@ -358,58 +347,51 @@ async def _run(self, output_emitter: tts.AudioEmitter) -> None:
...
-                        data_buf += line.strip()
+                    if payload.get("is_final", False):
+                        audio_b64 = payload.get("audio", "")
+                        if audio_b64:
+                            output_emitter.push(base64.b64decode(audio_b64))
+                        break
+
+                    audio_b64 = payload.get("audio", "")
+                    if audio_b64:
+                        output_emitter.push(base64.b64decode(audio_b64))
```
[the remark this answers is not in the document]
tinalenguyen: Verified — not an issue. For audio/wav, AudioStreamDecoder uses _WavInlineDecoder (not PyAV), a state machine that resets on each new RIFF header, so concatenated complete-WAV chunks decode correctly (decoder.py:180-187). I drove the real AudioEmitter as the plugin does — 8 complete WAV files, one per push — and reconstruction is byte-exact in both modes. So per-chunk stripping / audio/pcm isn't needed.

PR #6478, opened 2026-07-18
fix(core): reset user away timer on final STT transcript AGT-3149

chenghao-mou:
VAD can miss speech while STT still emits a final transcript. In that case the user stays `listening`, so the away timer kept counting down and could fire despite recent STT activity. Observed from a user session where short utterance was ignored by the timer.

On a final transcript when the user is not already speaking, refresh `user_away_timeout` without changing user state. The existing `away` → `listening` recovery path is unchanged.


davidzhao (approved):
lg

theomonnom (approved):

PR #6480, opened 2026-07-18
fix(voice/realtime): add tool_reply_after_audio to prevent duplicate speech

### LINKED ISSUE #6411: voice/realtime: post-tool reply fires even when the function-call response already produced audio — agent audibly answers twice (bundled message + function_call)
**Environment**: livekit-agents 1.5.1, `livekit-plugins-openai` realtime path against the xAI Grok realtime API (`grok-voice-think-fast-1.0`), server-side VAD. Verified the same decision logic in 1.6.5 `agent_activity.py`: `fnc_executed_ev._reply_required` → `interrupt()` + `_realtime_reply_task(...)` with no check on whether the triggering response already produced audio.

**Behavior**: reasoning-class realtime models frequently bundle an audible assistant message AND a `function_call` in ONE response — despite prompt instructions to call tools silently. The framework's post-tool reply decision (`reply_required`, i.e. the tool returned a value) is computed *after* that bundled audio has already played, and fires unconditionally: `interrupt()` (a blanket `response.cancel`) followed by a fresh `response.create`. The caller audibly hears the agent answer the same turn twice.

Wire capture from a real session (timestamps to the ms):

```
09:39:40.532  input_audio_buffer.committed
09:39:40.755  response.created            (VAD-created, id=df812bbb…)
09:39:41.048  response.output_item.added  (message, content=[] at add time)
09:39:41.050  response.content_part.added (audio)   ← ~6.4s of audible speech streams
09:39:47.451  response.output_item.added  (function_call)
09:39:47.453  response.done
09:39:47.479  FUNCTION_TOOLS_EXECUTED     reply_required=True
09:39:47.768  response.created            (framework continuation, id=19a93a3c…, client response.create)
              → a second audible message answering the SAME turn again
```

Gap between the bundled audio and the duplicate: ~0.3s. In a live test the caller reacted with *"Can you stop speaking?"*

**Why applications cannot fix this at their layer** (we tried, extensively): at `function_tools_executed` time an app-side guard knows the turn already delivered audio, but cannot know whether that audio was a throwaway filler ("one moment…" → continuation NEEDED) or the complete reply (continuation = audible duplicate). No wire signal distinguishes the two at decision time. We shipped both guard polarities: cancelling the continuation froze the agent mid-call whenever the bundle was a filler; allowing it produces the duplicate whenever the bundle was complete. Prompt instructions ("call the tool silently") are exactly what reasoning models fail to follow here — that is what creates the bundle.

**The framework CAN know**: `AgentActivity`/`RealtimeSession` sees the response that carried the `function_call`, including whether it contained audio output items/deltas and how much played out — information the app never gets at decision time.

**Ask** — gate the post-tool reply on whether the function-call response already produced audible output, or expose the decision. Any of these shapes would unblock:

1. A config/capability like `tool_reply_generation="skip_if_audio"`.
2. Expose `had_audio` (or an output-item summary of the triggering response) on `FunctionToolsExecutedEvent`, so `cancel_tool_reply()`-style decisions can be informed rather than blind.
3. A user callback consulted before `_realtime_reply_task` when the triggering response contained an audible message item.

**Related**: #4554 (Gemini speaks twice after function calls — closed with a prompt-instruction workaround; the trace above is a case where prompt instructions are structurally unreliable) and #2407 (`auto_tool_reply_generation` configurability). This report adds the wire-level trace of the bundling and evidence that both app-side workaround polarities fail.

Happy to provide full dual-direction JSONL wire traces for both the duplicate case and the freeze case.

Utkarshsteve:
## Summary

- Adds a configurable `tool_reply_after_audio` policy to `ToolHandlingOptions` that controls whether the framework generates a post-tool reply when the realtime response already produced audible audio
- Default is `"always"` (backwards-compatible); setting `"skip"` suppresses the duplicate reply
- Exposes `response_had_audio` on `FunctionToolsExecutedEvent` so user handlers can inspect the signal

## Problem

When a realtime model (e.g. xAI Grok) bundles audio output **and** a `function_call` in one response, the framework fires a post-tool reply after executing the tool — causing the agent to audibly answer twice (#6411).

## Solution

Detection: after tool execution, check whether the triggering response produced audio that was actually played (`started_speaking_at is not None` and at least one message output wasn't skipped).

Guard: when `tool_reply_after_audio == "skip"` and audio was produced, skip the post-tool reply generation.

```python
AgentSession(
    tool_handling={"tool_reply_after_audio": "skip"},
)
```

## Test plan

- [x] Unit tests for event property defaults and mutation
- [x] Unit tests for `ToolHandlingOptions` config dictionary
- [x] Unit tests for `AgentSession` config storage (default, skip, always, empty dict)
- [x] `make check` passes (format, lint, mypy strict)
- [x] `uv run pytest --unit` — all 1132 + 10 new tests pass
- [ ] Manual verification with xAI Grok realtime model (requires API key)

Closes #6411

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -3853,11 +3853,16 @@ class AgentActivity(RecognitionHooks):
         if len(tool_output.output) > 0:
             speech_handle._num_steps += 1
 
+            response_had_audio = started_speaking_at is not None and any(
+                entry.out.played != "skipped" for entry in message_outputs
+            )
+
             new_fnc_outputs: list[llm.FunctionCallOutput] = []
             generate_tool_reply: bool = False
             fnc_executed_ev = FunctionToolsExecutedEvent(
                 function_calls=[], function_call_outputs=[]
             )
+            fnc_executed_ev._response_had_audio = response_had_audio
             new_agent_task: Agent | None = None
             ignore_task_switch = False
 
@@ -3950,9 +3955,11 @@ class AgentActivity(RecognitionHooks):
                             self._pending_auto_tool_reply_fut = None
                         auto_reply_fut.set_result(None)
 
+            skip_for_audio = response_had_audio and self._session._tool_reply_after_audio == "skip"
             if (
                 fnc_executed_ev._reply_required
                 and not self._rt_session.capabilities.auto_tool_reply_generation
+                and not skip_for_audio
             ):
                 self._rt_session.interrupt()
 
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -474,6 +474,11 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         self._async_tool_options = _resolve_async_tool_options(
             tool_handling.get("async_options") if is_given(tool_handling) else None
         )
+        self._tool_reply_after_audio: Literal["always", "skip"] = (
+            tool_handling.get("tool_reply_after_audio", "always")
+            if is_given(tool_handling)
+            else "always"
+        )
 
         # unrecoverable error counts, reset after agent speaking
         self._llm_error_counts = 0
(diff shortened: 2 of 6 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -3853,11 +3853,16 @@ def _create_assistant_message(
         if len(tool_output.output) > 0:
             speech_handle._num_steps += 1
 
+            response_had_audio = started_speaking_at is not None and any(
```
Sanjays2402, line 3856: `started_speaking_at` is also set by `_TextOutput.first_text_fut` in `_on_first_frame`, so this becomes true for a text-only response with no audio at all. With transcription enabled and `tool_reply_after_audio="skip"`, that suppresses the required post-tool reply even though nothing was spoken.

PR #6484, opened 2026-07-18
prewarm LLM to skip connection establishment

davidzhao:
warms up DNS + TLS cache/handshakes before the first LLM request is sent

### DIFF
--- a/livekit-agents/livekit/agents/llm/llm.py
+++ b/livekit-agents/livekit/agents/llm/llm.py
@@ -144,7 +144,7 @@ class LLM(
         extra_kwargs: NotGivenOr[dict[str, Any]] = NOT_GIVEN,
     ) -> LLMStream: ...
 
-    def prewarm(self) -> None:
+    def prewarm(self, *, loop: asyncio.AbstractEventLoop | None = None) -> None:
         """Pre-warm connection to the LLM service.
 
         Establishes DNS resolution and the TLS connection to the provider before the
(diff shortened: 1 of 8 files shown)


theomonnom (approved):

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -477,6 +477,15 @@ def __init__(
         self._llm = llm or None
         self._tts = tts or None
 
+        # eagerly establish DNS/TLS to the LLM provider so the first inference
+        # request doesn't pay connection setup costs
+        if isinstance(self._llm, LLM):
+            try:
+                asyncio.get_running_loop()
```
theomonnom: There is always an event loop

livekit-agents/livekit/agents/voice/agent_session.py
theomonnom, line 292: here

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -144,10 +145,47 @@ def chat(
...
+
+        async def _prewarm() -> None:
+            try:
+                await self._prewarm_impl()
+            except Exception:
+                pass
+
+        self._prewarm_task = asyncio.create_task(_prewarm())
+
+    async def _prewarm_impl(self) -> None:
```
chenghao-mou, line 179: Q: should we add support for FallbackAdapter?

livekit-agents/livekit/agents/voice/agent_session.py
chenghao-mou: Adding to this, I think we can skip checking it here since _llm.prewarm already does that internally.

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -144,10 +145,47 @@ def chat(
...
+
+        try:
+            asyncio.get_running_loop()
+        except RuntimeError:
+            return  # no running event loop, prewarm will be retried when the agent starts
+
+        async def _prewarm() -> None:
+            try:
+                await self._prewarm_impl()
+            except Exception:
```
chenghao-mou, line 174: Should we clear the `_prewarm_task` here?

davidzhao: I don't think we would? if this has failed, we'd just wait for normal cleanup?

PR #6488, opened 2026-07-20
Support adaptive interruption for realtime models

longcw:
Enables client-side adaptive interruption (barge-in) for OpenAI-compatible realtime models when the model's own server-side turn detection is disabled — so a backchannel ("uh-huh") while the agent is speaking doesn't trigger a spurious reply, without requiring a separate STT.

**The gate.** With no STT there's no transcript to gatekeep, so the audio recognition layer tracks a turn-scoped signal — whether the turn's speech overlapped agent speech and was classified a backchannel — and surfaces it on the end-of-turn info. When set, the user turn is dropped and the buffered input audio is cleared, so no reply is generated.

**Difference from the STT gate.** The STT path is timestamp-windowed: it drops the backchannel portion of a turn and re-emits the rest. Realtime is all-or-nothing — `clear_audio()` wipes the whole input buffer and the model buffers the turn as one unit, so there's no per-timestamp split. It's correct for the dominant cases (pure backchannel, real interruption) and drops only on a confirmed backchannel; a slow verdict after the agent has stopped falls back to committing rather than silently dropping a possible real barge-in.

**STT + realtime.** If you want the finer-grained STT gate, pass an STT alongside the realtime model — the realtime drop path is skipped when an STT is present and transcript-based backchanneling takes over.

### DIFF
--- a/livekit-agents/livekit/agents/inference/interruption.py
+++ b/livekit-agents/livekit/agents/inference/interruption.py
@@ -136,6 +136,13 @@ class OverlappingSpeechEvent(BaseModel):
     is_interruption: bool = False
     """Whether interruption is detected."""
 
+    agent_ended: bool = False
+    """True when the overlap ended because the agent finished speaking rather than the user.
+
+    The user may still be talking, so ``is_interruption`` (always ``False`` here) is inconclusive
+    and must not be treated as a confirmed backchannel verdict.
+    """
+
     total_duration: float = 0.0
     """RTT (Round Trip Time) time taken to perform the inference, in seconds."""
 
@@ -180,6 +187,7 @@ class OverlappingSpeechEvent(BaseModel):
         is_interruption: bool,
         started_at: float | None = None,
         ended_at: float | None = None,
+        agent_ended: bool = False,
     ) -> OverlappingSpeechEvent:
         """Initialize the event from a cache entry.
 
@@ -188,6 +196,7 @@ class OverlappingSpeechEvent(BaseModel):
             is_interruption: Whether the interruption is detected.
             started_at: The timestamp when the overlap speech started.
             ended_at: The timestamp when the overlap speech ended.
+            agent_ended: Whether the overlap ended because the agent finished speaking.
 
         Returns:
             The initialized event.
@@ -196,6 +205,7 @@ class OverlappingSpeechEvent(BaseModel):
             type="overlapping_speech",
             detected_at=ended_at or time.time(),
             is_interruption=is_interruption,
+            agent_ended=agent_ended,
             overlap_started_at=started_at,
             speech_input=entry.speech_input,
             probabilities=entry.probabilities,
@@ -232,8 +242,9 @@ class _OverlapSpeechStartedSentinel:
 
 
 class _OverlapSpeechEndedSentinel:
-    def __init__(self, ended_at: float) -> None:
+    def __init__(self, ended_at: float, agent_ended: bool = False) -> None:
         self._ended_at = ended_at
+        self._agent_ended = agent_ended
 
 
 class _FlushSentinel:
@@ -623,6 +634,7 @@ class InterruptionStreamBase(ABC):
                             is_interruption=False,
                             started_at=self._overlap_started_at,
                             ended_at=input_frame._ended_at,
+                            agent_ended=input_frame._agent_ended,
                         )
                         ev.num_requests = await self._num_requests.get_and_reset()
                         self.send(ev)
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1381,6 +1393,11 @@ async def _on_overlap_speech_event(self, ev: inference.OverlappingSpeechEvent) -
             )
             return
 
+        # only honor the verdict while this turn's overlap is unresolved so a late verdict
+        # can't leak into the next turn; an interruption supersedes a prior backchannel
+        if self._overlap_in_current_turn:
+            self._turn_backchannel_over_agent = not ev.is_interruption
```
chenghao-mou, line 1406: The overlap speech also ends when agent finishes speaking first, so we trigger a synthesized `is_interruption=False`, but the user speech might not be done, and we might carry over the wrong flag (til the end of turn).

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2235,7 +2233,31 @@ def on_end_of_turn(self, info: _EndOfTurnInfo) -> bool:
...
+            and isinstance(self.llm, llm.RealtimeModel)
+            and not self.llm.capabilities.turn_detection
+            and self._interruption_detection_enabled
+            and (
+                # confirmed backchannel for this turn (survives the agent stopping)
+                info.backchannel_over_agent
+                # or agent still speaking with nothing flagged yet
+                or (
+                    not self._interruption_detected
+                    and self._current_speech is not None
```
chenghao-mou, line 2260: Q: should we check a narrower agent speaking state instead of speech handle?

livekit-agents/livekit/agents/voice/audio_recognition.py
longcw: added a flag `agent_ended` to distinguish two types of `is_interruption=False`

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: `_current_speech` is intentionally broader than `agent_state == "speaking"`, during response generation (before the first audio frame) `agent_state` is still listening while `_current_speech` is set, and we don't want the backchannel to interrupt that right?

chenghao-mou: right, I guess in this branch/under this condition, we are not running barge-in (agent not actually speaking) so this falls back to VAD interruption, and if it didn't interrupt on end of turn, it must be failing VAD interruption gate as well. So it is fine to keep it this way.

livekit-agents/livekit/agents/voice/audio_recognition.py
chenghao-mou: if there are one backchannel ev + one agent ends first ev during the agent speech, `self._turn_backchannel_over_agent` would be stale, right?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -509,8 +514,15 @@ def _on_start_of_speech(
...
         )
+        if not self._agent_speaking:
+            # non-overlapping speech drops a stale backchannel verdict
+            self._turn_backchannel_over_agent = False
+            self._overlap_in_current_turn = False
+
         if not self._adaptive_interruption_active or not self._agent_speaking:
             return
+        # overlap over agent speech started this turn; gates verdict acceptance below
+        self._overlap_in_current_turn = True
```
chenghao-mou, line 526: we probably should reset `self._turn_backchannel_over_agent` too?

livekit-agents/livekit/agents/voice/audio_recognition.py
longcw: fixed.

livekit-agents/livekit/agents/inference/interruption.py
```diff
@@ -136,6 +136,13 @@ class OverlappingSpeechEvent(BaseModel):
     is_interruption: bool = False
```
chenghao-mou, line 136: There seems to be a race condition with false interruption:

```
    11:57:42.097 TRACE    livekit.agents     overlap speech started, starting interruption inference {"overlap_count": 1, "room": "console"}
    11:57:43.893 DEBUG    livekit.agents     resumed false interrupted speech {"timeout": 1.0, "room": "console"}
```

namely false interruption should not be trigger-able when adaptive interruption is still inferencing? (if it detects interruption, then we can start a false interruption timer)

chenghao-mou: I think this is expected as during the boundary, both VAD and barge-in are active.

chenghao-mou (approved):

Issue #1034, opened 2024-11-04
Modify context before LLM


tpy37:
Thank you for the great package. I am looking for a method to change the context before sending it to the LLM in MultimodalAgent class. I think it exists in VoicePipelineAgent, and I am wondering how I could implement it with OpenAI realtime API.

tpy37:
I think

tpy37:
I would be happy if you could implement the RAG part of the MultimodalAgent, as in the documentation! :) 
before_llm_cb=_enrich_with_rag

![image](https://github.com/user-attachments/assets/a360741e-c548-43d9-97da-bcb8a5aea894)

davidzhao:
it's a bit difficult with multimodal agent, because it's from voice input directly to voice output.

the way to handle RAG is with function calling. If you are defining a function for the LLM to look up information with the user's query, it should be straight forward to pick up the function call and return the RAG results that way

tpy37:
Thank you very much David! 
I see... I was trying the function call, but changing to tool "required" seem to completely halt the process and break the conversation, so I stopped using it. 

Since we have the transcribed text from the user's audio in the 
openai.realtime.RealtimeResponse

I was thinking that we could analyze the transcribed texts, and then use that to do RAG, and send the results async to the openai api as texts? 
`const event = {
  type: 'conversation.item.create',
  item: {
    type: 'message',
    role: 'user',
    content: [
      {
        type: 'input_text',
        text: 'Hello!'
      }
    ]
  }
};
ws.send(JSON.stringify(event));
ws.send(JSON.stringify({type: 'response.create'}));`
https://platform.openai.com/docs/guides/realtime?text-generation-quickstart-example=text

Just some thoughts...

prashantmetadome:
@tpy37 can you guide me on how to change context in VoicePipelineAgent?

tpy37:
I am sorry, I haven't done it myself in VoicePipeline. But the example is available in https://docs.livekit.io/agents/voice-agent/voice-pipeline/#modify-context-before-llm

I think there was also example to send it to RAG using this before_llm_cb in one of the Github repository:
examples/voice-pipeline-agent/simple-rag/assistant.py

prashantmetadome:
@tpy37 thank you very much, it does help. But I am still unsure.. my use case is I want to manipulate the prompt based on a tool call. 

To go deeper in the specific requirement, the conversation has outgrown the current prompt and has gone into different territory that need to be handled by a different prompt.
[...]

PR #6492, opened 2026-07-20
feat(inference): add inference.AvatarSession (avatar provisioning via the gateway)

shawnfeldman:
## What
Adds `livekit.agents.inference.AvatarSession` — a thin subclass of `voice.avatar.AvatarSession` that provisions an avatar provider session through the **LiveKit Inference gateway** (`POST /v1/avatar/sessions`) instead of a BYOK provider key.

```python
from livekit.agents import AgentSession, inference

avatar = inference.AvatarSession("lemonslice", image_url="https://…", prompt="…")
await avatar.start(session, room=ctx.room)
await avatar.wait_for_join()
```

## Why
BYOK avatar plugins require a customer provider key and bill LiveKit nothing for the avatar. Routing provisioning through Inference gives a **single-key** experience (`LIVEKIT_API_KEY` / `LIVEKIT_API_SECRET` only) and lets LiveKit meter avatar usage — matching `inference.STT/LLM/TTS`. BYOK plugins stay as-is; this is an additive path.

## How
- The agent mints the avatar worker's room token **locally** (identical to the BYOK plugins — `kind=agent`, `room_join`, `lk.publish_on_behalf`, plus an `lk.avatar_provider` attribute for future metering) and sends a normalized payload to the gateway; the gateway calls the provider with LiveKit's wholesale key. Media/lip-sync still flow in-room over `DataStreamAudioOutput` — unchanged from BYOK.
- Model string `"lemonslice"` or `"lemonslice/<agent_id>"`; `image_url` / `prompt` / `idle_prompt` first-class; `extra_kwargs` for provider-specific fields.
- **Idempotency-Key** per `start()` (stable across retries) so a retried create replays on the gateway instead of creating a second paid session.
- **`aclose()`** terminates the provider session via the gateway (`/v1/avatar/sessions/terminate`) so it stops billing on shutdown rather than lingering to the idle timeout.
- Uses the gateway-reported **`sample_rate`** for `DataStreamAudioOutput` (no per-provider hardcoding). Works inside an agent job or standalone (derives identity/room-sid from the connected room when there's no job context).
- Lazy PEP 562 export in `inference/__init__.py` to avoid a circular import during voice-package init.

## Files
- `livekit-agents/livekit/agents/inference/avatar.py` (new)
- `livekit-agents/livekit/agents/inference/__init__.py` — lazy `AvatarSession` export
- `examples/avatar/inference_agent.py` (new) + a note in `examples/avatar/README.md`
- `tests/test_inference_avatar.py` (new)

## Testing
`pytest tests/test_inference_avatar.py` — 19 passing: model-string parsing, credential resolution/env fallbacks, create payload/headers (incl. stable Idempotency-Key), retry/error mapping (403 non-retryable, 502 retried), response `sample_rate` → `DataStreamAudioOutput`, and `aclose()` terminate (+ no-op when no session). `ruff` clean.

## Dependencies / rollout
- Requires the gateway endpoints from **livekit/agent-gateway#1034** and the `avatar_lemonslice` feature flag enabled for the project (staging first). Until then, `start()` returns a `403 not_enabled`.
- E2E coverage: **livekit/e2e#1255** (`agents/avatar/` suite), which activates once this ships and the flag is on.

Refs: LKINF-386

### DIFF
--- a/livekit-agents/livekit/agents/inference/avatar.py
+++ b/livekit-agents/livekit/agents/inference/avatar.py
@@ -37,9 +37,14 @@ if TYPE_CHECKING:
     from ..voice import AgentSession
 
 # Fallback audio sample rate when the gateway response omits one. The gateway
-# is authoritative (it returns the per-provider rate) so this is only a floor
-# for older gateways.
+# is authoritative (it returns the per-provider rate) so this is only a
+# fallback for older gateways.
 _DEFAULT_SAMPLE_RATE = 16000
+# Total request timeout for gateway calls. sock_connect (conn_options.timeout)
+# bounds the TCP connect; this bounds the whole request so a gateway that
+# accepts the connection but never responds can't hang start()/aclose()
+# indefinitely. Matches the BYOK lemonslice plugin's request timeout.
+_REQUEST_TIMEOUT = 60.0
 # lk.avatar_provider tags the avatar worker participant with its provider so
 # server-side (participant-lifetime) avatar-minutes metering can attribute the
 # worker's room time to the right SKU. See the inference-avatar plan.
@@ -174,6 +179,9 @@ class AvatarSession(BaseAvatarSession):
 
         self._session_id: str | None = None
         self._provider_session_id: str | None = None
+        # HMAC issued by the gateway alongside provider_session_id; required by
+        # /avatar/sessions/terminate to prove this project owns the session.
+        self._terminate_token: str | None = None
 
     @property
     def avatar_identity(self) -> str:
@@ -202,6 +210,12 @@ class AvatarSession(BaseAvatarSession):
         livekit_api_key: NotGivenOr[str] = NOT_GIVEN,
         livekit_api_secret: NotGivenOr[str] = NOT_GIVEN,
     ) -> None:
+        if self._session_id is not None or self._provider_session_id is not None:
+            raise RuntimeError(
+                "AvatarSession.start() may only be called once per instance; "
+                "create a new AvatarSession to start another avatar"
+            )
+
         await super().start(agent_session, room)
 
         lk_url = livekit_url or (os.getenv("LIVEKIT_URL") or NOT_GIVEN)
@@ -221,7 +235,7 @@ class AvatarSession(BaseAvatarSession):
         if job_ctx is not None:
             local_participant_identity = job_ctx.local_participant_identity
             room_sid = job_ctx.job.room.sid or await room.sid
-        elif room.local_participant is not None:
+        elif room.isconnected():
             local_participant_identity = room.local_participant.identity
             room_sid = await room.sid
         else:
@@ -257,7 +271,22 @@ class AvatarSession(BaseAvatarSession):
 
         self._session_id = create_resp.get("session_id")
         self._provider_session_id = create_resp.get("provider_session_id")
+        self._terminate_token = create_resp.get("terminate_token")
         sample_rate = create_resp.get("sample_rate") or _DEFAULT_SAMPLE_RATE
+
+        if self._provider_session_id and not self._terminate_token:
+            # aclose() cannot terminate without this; the provider session will
+            # run to its own idle timeout instead of being stopped explicitly.
+            logger.warning(
+                "avatar gateway create response had no terminate_token; this "
+                "session cannot be explicitly terminated and will bill until "
+                "its provider idle timeout",
+                extra={
+                    "provider": self._provider,
+                    "session_id": self._session_id,
+                    "provider_session_id": self._provider_session_id,
+                },
+            )
 
         # Rebind the audio tail to the avatar worker using the gateway-reported
         # sample rate. Done after the create response (unlike BYOK, which
@@ -289,22 +318,40 @@ class AvatarSession(BaseAvatarSession):
     async def aclose(self) -> None:
         # Terminate the provider session through the gateway so it stops billing
         # immediately instead of lingering until its idle timeout. Best-effort:
-        # a failure must not block participant cleanup in the base class.
+        # a failure must not block participant cleanup in the base class, so
+        # base cleanup runs in `finally` even if terminate raises or is
+        # cancelled (e.g. job-shutdown deadline). The ids are cleared only on
+        # a confirmed terminate so a second aclose() call can still retry.
         provider_session_id = self._provider_session_id
-        self._provider_session_id = None
-        if provider_session_id:
-            try:
-                await self._terminate_session(provider_session_id)
-            except Exception:
-                logger.warning(
-                    "failed to terminate inference avatar session",
+        terminate_token = self._terminate_token
+        try:
+            if provider_session_id and terminate_token:
+                try:
+                    await self._terminate_session(provider_session_id, terminate_token)
+                except Exception:
+                    logger.warning(
+                        "failed to terminate inference avatar session; it will "
+                        "keep billing until its provider idle timeout unless "
+                        "aclose() is called again",
+                        extra={
+                            "provider": self._provider,
+                            "provider_session_id": provider_session_id,
+                        },
+                        exc_info=True,
+                    )
+                else:
+                    self._provider_session_id = None
+                    self._terminate_token = None
+            elif provider_session_id:
+                logger.debug(
+                    "no terminate_token for this avatar session; skipping explicit terminate",
                     extra={
                         "provider": self._provider,
                         "provider_session_id": provider_session_id,
                     },
-                    exc_info=True,
                 )
-        await super().aclose()
+        finally:
+            await super().aclose()
 
     async def _create_session(
         self,
@@ -360,7 +407,7 @@ class AvatarSession(BaseAvatarSession):
                         headers=headers,
                         json=payload,
                         timeout=aiohttp.ClientTimeout(
-                            total=None,
+                            total=_REQUEST_TIMEOUT,
                             sock_connect=self._conn_options.timeout,
                         ),
                     ) as response:
@@ -372,13 +419,20 @@ class AvatarSession(BaseAvatarSession):
                                 body=text,
                             )
                         return await response.json()  # type: ignore[no-any-return]
-                except asyncio.TimeoutError as e:
-                    last_exc = APITimeoutError()
-                    logger.warning("avatar gateway request timed out", extra={"attempt": i})
-                    _ = e
+                except asyncio.TimeoutError:
+                    last_exc = APITimeoutError(
+                        f"avatar gateway create timed out after attempt {i + 1}"
+                    )
+                    logger.warning(
+                        "avatar gateway request timed out",
+                        extra={"provider": self._provider, "attempt": i},
+                    )
                 except aiohttp.ClientError as e:
                     last_exc = APIConnectionError(str(e))
-                    logger.warning("failed to call avatar gateway", extra={"error": str(e)})
+                    logger.warning(
+                        "failed to call avatar gateway",
+                        extra={"provider": self._provider, "error": str(e)},
+                    )
                 except APIError as e:
                     last_exc = e
                     if not e.retryable:
@@ -398,7 +452,7 @@ class AvatarSession(BaseAvatarSession):
             if self._http_session is None:
                 await session.close()
 
-    async def _terminate_session(self, provider_session_id: str) -> None:
+    async def _terminate_session(self, provider_session_id: str, terminate_token: str) -> None:
         url = f"{self._base_url.rstrip('/')}/avatar/sessions/terminate"
         headers = {
             **get_inference_headers(),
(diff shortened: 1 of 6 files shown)


shawnfeldman (commented):
## PR Review Summary

Reviewed at head `25e916b4f`. Verified locally: **19/19 unit tests pass**, `ruff check`/`format` clean, full `scripts/check_types.py` mypy pass, and all three import paths for the lazy `inference.AvatarSession` export work (PEP 562 hook, `from`-import, direct module import). Payloads/headers were also cross-checked field-by-field against the gateway handler in livekit/agent-gateway#1034 — which is how the critical issue below surfaced.

### Critical Issues (must fix)

1. **[code-reviewer] `aclose()` termination can never succeed against the gateway: `terminate_token` is neither captured nor sent** — `avatar.py:258-259`, `avatar.py:409`.
   The #1034 create response returns a `terminate_token` (an HMAC binding project+provider+provider_session_id, minted in `pkg/handler/avatar.go` specifically so terminate can't be invoked cross-tenant), and `handleTerminate` hard-requires it: missing → `400 "provider, provider_session_id, and terminate_token are required"`, wrong → `403`. `start()` reads only `session_id`/`provider_session_id`/`sample_rate` from the create response, and the terminate payload is `{provider, provider_session_id}` only. So every terminate 400s, `aclose()` swallows it into a warning, and the provider session keeps billing until idle timeout — the exact failure this feature exists to prevent. The unit test (`test_aclose_terminates_session`) asserts the token-less body, baking the wrong contract into the suite, and the fake gateway accepts any body, so tests stay green.
   **Fix:** capture `self._terminate_token = create_resp.get("terminate_token")` in `start()`, include it in the terminate payload (skip with a debug log if the gateway didn't return one, for older gateways), and update the test to assert it's sent. If the gateway side is meant to change instead, reconcile #1034 — as the two PRs stand today they are incompatible.

### Important Issues (should fix)

2. **[silent-failure-hunter] Unbounded HTTP awaits: `total=None` on both create and terminate** — `avatar.py:362-365`, `avatar.py:417`.
   Only the TCP connect is bounded; `total=None` explicitly disables even aiohttp's built-in 5-minute default. A gateway/LB that accepts the connection but stalls hangs `start()` forever — the retry loop and the `except asyncio.TimeoutError` arm are unreachable for post-connect stalls. Terminate is worse: `aclose()` runs as a job-shutdown callback in an unbounded `gather` after the room is already disconnected, so a hung terminate blocks base cleanup until the supervisor SIGKILLs the process (`shutdown_process_timeout` default 10s) — taking the "failed to terminate" warning down with it. The lemonslice BYOK plugin this mirrors uses `total=60.0` (`lemonslice/api.py:251-253`); bey/tavus omit it, but this path is billing-sensitive.
   **Fix:** bounded `total` (e.g. `60.0`) on both calls, and move `await super().aclose()` into a `finally:` so base cleanup survives terminate cancellation/timeout.

3. **[code-reviewer] Dead `else` branch in `start()`: the designed `RuntimeError` is unreachable** — `avatar.py:224-231`.
   `rtc.Room.local_participant` raises `Exception("cannot access local participant before connecting")` when the room isn't connected — it never returns `None`. In the standalone-no-job-context path with an unconnected room, evaluating the `elif` condition itself raises, so the friendly error (the only text that mentions the job-context alternative) is dead code. Reproduced with a real `rtc.Room()`.
   **Fix:** `elif room.isconnected():` — the codebase-standard readiness guard (`inference/_utils.py` uses it for exactly this reason).

4. **[type-design-analyzer] Double-`start()` is unguarded and costs real money** — `avatar.py:343`, `avatar.py:259`.
   A second `start()` mints a fresh Idempotency-Key → a second wholesale-billed provider session — and overwrites `_provider_session_id`, permanently losing the only terminate handle for the first (it bills until idle timeout, with zero signal). The BYOK plugins share this gap, but here the wholesale key is LiveKit's own.
   **Fix:** one line at the top of `start()`: raise if `_session_id`/`_provider_session_id` is already set. Guarding on the ids (not a new flag) keeps retry-after-a-*failed*-start working.

5. **[silent-failure-hunter] Terminate is strictly one-shot, and the create-response contract is trusted silently** — `avatar.py:293-294`, `avatar.py:259`.
   `_provider_session_id` is cleared *before* the terminate attempt, so one transient failure (502, connection reset) permanently forfeits the billing stop — create gets `max_retry` attempts, terminate gets one. Separately, a 2xx create response missing `provider_session_id` silently disables termination (`aclose()` skips with no log; the ids only appear at `debug` level). The gateway documents terminate as idempotent ("a not-found session is reported as success"), so clear-on-success is safe against double-terminate.
   **Fix:** clear the id only on success (or non-retryable 4xx), consider one retry, and `logger.warning` when a 2xx create response lacks `provider_session_id`.

6. **[pr-test-analyzer] Billing-lifecycle test gaps.** The 19 tests pin the stated guarantees well, but the costliest paths are uncovered: double-`start()` (see #4), `aclose()` when terminate fails (assert it doesn't raise and base cleanup still runs), and partial-start orphan safety — session ids are assigned (`avatar.py:258-259`) *before* `replace_audio_tail`, correctly keeping a failed rebind terminable, but nothing pins that ordering against an innocent reorder. Also unasserted: the minted worker JWT (`lk.avatar_provider` metering attribution and `lk.publish_on_behalf` could be dropped or typo'd with the suite green — decode the captured token in a `start()`-level test) and `_FakeSink` ignores `destination_identity`/`wait_remote_track`. All expressible in the existing fake-gateway style, ~60-80 lines total.

### Suggestions

- **[comment-analyzer]** Two comment inaccuracies: `avatar.py:40` — "only a floor" should be "fallback" (a gateway-returned 8000 would be used as-is, not clamped up); `avatar.py:262-264` — "unlike BYOK, which hardcodes the rate and rebinds first" is true only of lemonslice (bey and tavus rebind *after* their create calls). Also `avatar.py:46` — "See the inference-avatar plan" is a dangling internal reference in a public repo; drop it.
- **[comment-analyzer]** `avatar.py:115` — "Mutually exclusive with a catalog id in the model string" is enforced by prose only: both `avatar_id` and `image_url` are sent when both are given. Validate in `__init__` or reword to "the gateway rejects the combination."
- **[code-simplifier]** `avatar.py:375-378` — drop the `_ = e` lint shim: bare `except asyncio.TimeoutError:` (tts.py:767 precedent). While there, give the terminal `APITimeoutError` a message with attempt count and URL (it currently surfaces as a context-free "Request timed out.") and add `provider` to the retry warnings' `extra`.
- **[code-simplifier]** Extract the duplicated auth-header dict in `_create_session`/`_terminate_session` into a `_auth_headers()` helper (keep the call inside the retry loop so each attempt still mints a fresh JWT); parametrize the `_gateway` test fixture's route so the terminate test stops hand-rolling its own server.
- **[type-design-analyzer]** `self._extra_kwargs` stores the caller's dict by reference; `tts.py:402` defensively copies (`dict(...)`) — do the same.
- **[pr-test-analyzer]** Cheap tests worth adding: import via the public PEP 562 path (`inference.AvatarSession` — the suite imports the module directly, so a typo'd `__getattr__` stays green); `LIVEKIT_INFERENCE_API_KEY` precedence with both env pairs set; missing `sample_rate` → 16000 fallback; distinct Idempotency-Keys across two creates. Consider a `__dir__` in `inference/__init__.py` so `dir()` lists the lazy export.

### Strengths

- Convention fidelity is excellent: credential resolution is byte-for-byte the `inference/tts.py` idiom, worker-token minting byte-for-byte the BYOK pattern (with the `lk.avatar_provider` metering attribute the only — documented — addition), and `start()` is LSP-clean where lemonslice needs a `# type: ignore[override]`.
- Stable-Idempotency-Key-per-`start()` is exactly right for the gateway's singleflight+replay design, and the retry taxonomy (403 non-retryable, 5xx retried, exact attempt counts) is correct and tested against a real aiohttp `TestServer` — no mocked HTTP anywhere.
- The PEP 562 lazy export is the right cycle-break, and its justifying comment is verifiably accurate end-to-end (`voice/__init__` → `agent.py` → `inference`).
- Gateway-reported `sample_rate` removes the per-provider hardcoded-constant bug class; rebind-after-create is a deliberate, documented, correctly-reasoned deviation from the lemonslice ordering.
- Comments and docstrings are unusually substantive and — with the two exceptions above — all check out against the implementation, including the genuinely useful dual-project credential caveat.

---
*Verified: 19/19 unit tests, ruff clean, mypy (`scripts/check_types.py`) clean, CI fully green at `25e916b4f`. Contract cross-checked against livekit/agent-gateway#1034 `pkg/handler/avatar.go`.*

livekit-agents/livekit/agents/inference/avatar.py
```diff
@@ -0,0 +1,488 @@
...
+      gateway credentials when ``LIVEKIT_INFERENCE_*`` points at a different
+      project.
+    """
+
+    def __init__(
+        self,
+        model: str,
+        *,
+        image_url: NotGivenOr[str] = NOT_GIVEN,
+        prompt: NotGivenOr[str] = NOT_GIVEN,
```
tinalenguyen: should we separate these params into `LemonSliceOptions` similar to `inference.STT` and `inference.TTS`?

tinalenguyen: i think we're also missing `agent_id`

livekit-agents/livekit/agents/inference/avatar.py
```diff
@@ -0,0 +1,488 @@
...
+        )
+        if not self._api_secret:
+            raise ValueError(
+                "api_secret is required, either as argument or set LIVEKIT_API_SECRET environment variable"
+            )
+
+        self._avatar_participant_identity = (
+            avatar_participant_identity
+            if is_given(avatar_participant_identity)
+            else f"{self._provider}-avatar-agent"
```
tinalenguyen: should we denote that the provider is from inference in the participant identity as well? maybe `<provider>-inference-avatar` ?

livekit-agents/livekit/agents/inference/avatar.py
shawnfeldman: i think @adrian-cowham was intending to keep it flat for all the avatar sdks.  

@adrian-cowham thoughts?

adrian-cowham: Yeah, wanted to try to avoid using `extra_kwargs` if possible but I don't think it's going to work. We could keep it flat for now and add the provider options later. Also, I'm ok with using `LemonSliceOptions`.

shawnfeldman: Kept these flat for now: unlike `inference.STT`/`TTS` — where the per-provider `TypedDict` types the pass-through `extra_kwargs` — `image_url`/`prompt`/`idle_prompt`/`idle_timeout` are first-class fields on the gateway's avatar request (see `avatarRequest` in `pkg/handler/avatar.go`), and there's already a separate `extra_kwargs` for true pass-through extras. Can revisit a typed `LemonSliceOptions` if/when we add more avatar providers. Clarified the docstrings in 5a268c83e.

shawnfeldman: `agent_id` is already supported via the model string: `lemonslice/<agent_id>`. The gateway maps our `avatar_id` onto LemonSlice's `agent_id` (`pkg/provider/avatar/lemonslice/lemonslice.go`), so `inference.AvatarSession("lemonslice/<agent_id>")` selects a pre-built agent. Clarified this in the docstrings in 5a268c83e so it's discoverable.

livekit-agents/livekit/agents/inference/avatar.py
shawnfeldman: Good call — changed the default to `<provider>-inference-avatar` in 5a268c83e, so inference-provisioned avatars are distinguishable from the BYOK plugin's `<provider>-avatar-agent` worker (and it avoids an identity collision if both run in one room).

livekit-agents/livekit/agents/inference/avatar.py
shawnfeldman: Update: implemented this in 245154f00. Added `LemonSliceOptions` (`TypedDict, total=False`: `image_url` / `prompt` / `idle_prompt` / `idle_timeout`), passed via `extra_kwargs` like the `CartesiaOptions`/`DeepgramOptions` dicts in `inference.stt`/`tts`, and exported it from `livekit.agents.inference`. Known keys map onto the gateway's existing first-class request fields and unknown keys pass through under `extra_kwargs`, so no agent-gateway change was needed. e2e suite updated to match in livekit/e2e#1255.

tinalenguyen: agree that it would be ideal to avoid extra_kwargs

tinalenguyen (approved):

PR #6495, opened 2026-07-21
feat(voice): auto-disable realtime server-side turn detection

longcw:
Auto-disables a realtime model's server-side turn detection for the session when the user configured client-side turn-taking (a TurnDetector, `vad`, `manual`, or an explicit interruption mode) and the model supports it, so client-side EoT/barge-in works without also passing `turn_detection=None` to the RealtimeModel.

Adds `RealtimeCapabilities.can_disable_turn_detection` (OpenAI sets it from the ctor `turn_detection` given-ness) and a `session(turn_detection_disabled=...)` seam applied on the initial `session.update` via the per-session opts copy, so the model is never mutated and stays reusable. `AgentActivity` resolves the effective state once (`_rt_turn_detection_enabled`) and routes every server-side turn-detection check through it.

Other realtime plugins accept-and-ignore the flag (`can_disable_turn_detection` stays False): xAI (client turn-taking not stable in testing) and Gemini/others (no `commit_audio`/`clear_audio`; turns are activity-signal based).

Stacked on `longc/realtime-barge-in`; this PR targets that branch, not main.

### DIFF
--- a/livekit-agents/livekit/agents/llm/realtime.py
+++ b/livekit-agents/livekit/agents/llm/realtime.py
@@ -70,6 +70,9 @@ class RealtimeCapabilities:
     """Whether the model can produce audio output directly"""
     manual_function_calls: bool
     """Whether function call items already in the chat context can be resumed"""
+    can_disable_turn_detection: bool = False
+    """Whether server-side turn detection can be disabled for a session so the client drives
+    turn-taking. Set by plugins that implement ``session(turn_detection_disabled=True)``."""
     mutable_chat_context: bool = False
     """Whether the chat context can be updated mid-session"""
     mutable_instructions: bool = False
@@ -113,7 +116,12 @@ class RealtimeModel:
         return self._label
 
     @abstractmethod
-    def session(self) -> RealtimeSession: ...
+    def session(self, *, turn_detection_disabled: bool = False) -> RealtimeSession:
+        """Create a new session, optionally with server-side turn detection disabled.
+
+        ``turn_detection_disabled`` is honored only by plugins reporting
+        ``can_disable_turn_detection``; the model itself is left unchanged and reusable."""
+        ...
 
     @abstractmethod
     async def aclose(self) -> None: ...
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2294,7 +2294,7 @@ class AgentActivity(RecognitionHooks):
             self.stt is None
             and self._turn_detection != "manual"
             and isinstance(self.llm, llm.RealtimeModel)
-            and not self.llm.capabilities.turn_detection
+            and not self._rt_turn_detection_enabled
             and self._interruption_detection_enabled
             and (
                 # confirmed backchannel for this turn (survives the agent stopping)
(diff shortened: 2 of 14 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -457,6 +457,7 @@ def __init__(
             capabilities=llm.RealtimeCapabilities(
                 message_truncation=True,
                 turn_detection=turn_detection is not None,
+                can_disable_turn_detection=not is_given(turn_detection),
```
chenghao-mou, line 460: nit: we should update `with_azure` too

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -263,6 +262,55 @@ def __init__(self, agent: Agent, sess: AgentSession) -> None:
         # model to auto-generate a tool reply (auto_tool_reply_generation=True).
         self._pending_auto_tool_reply_fut: asyncio.Future[None] | None = None
 
+    def _resolve_rt_turn_detection_enabled(self) -> bool:
```
chenghao-mou, line 274: should we handle update_options with turn_detection for realtime too?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
chenghao-mou, line 1: we should handle with_azure too since it assigns the default turn detection before init.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
longcw: done

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: toggle turn detection at runtime will need a restart of rt_session if the server-side turn detection is disabled. I added a warning for now.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -592,6 +597,21 @@ def update_options(
             self._rt_session.update_options(tool_choice=self._tool_choice)
 
         if utils.is_given(turn_detection):
+            # a realtime model's server-side turn detection is resolved once at session start;
```
chenghao-mou, line 607: we should have a similar warning for `RealtimeModel.update_options(turn_detection=...)` too

chenghao-mou:
Another thing I noticed in testing: if the interruption is detected by the barge-in model, it leads to the false interruption path, and for realtime model, transcript isn't always available immediately for that interruption to take effect. Should we consider disable false interruption when no STT is configured with realtime model?

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -70,6 +70,9 @@ class RealtimeCapabilities:
     """Whether the model can produce audio output directly"""
     manual_function_calls: bool
     """Whether function call items already in the chat context can be resumed"""
+    can_disable_turn_detection: bool = False
```
chenghao-mou, line 73: nit: maybe `optional_turn_detection` so it matches other `adj_noun` patterns

livekit-agents/livekit/agents/llm/realtime.py
```diff
@@ -113,7 +116,12 @@ def label(self) -> str:
         return self._label
 
     @abstractmethod
-    def session(self) -> RealtimeSession: ...
+    def session(self, *, turn_detection_disabled: bool = False) -> RealtimeSession:
```
chenghao-mou, line 119: nit: should we rename it `enable_turn_detection: bool = False` instead?

chenghao-mou (approved):
LGTM overall. Tested locally with OAI Realtime and VAD or turn detector. Two nits and two issues from previous comments.

PR #6497, opened 2026-07-21
feat: tag simulation and redaction telemetry

chenghao-mou:
Simulated sessions and redaction previews need lightweight collector signals without scenario metadata or metrics attrs. 

**Passing simulation metadata to collector**
This adds only `lk.simulation.enabled` and opt-in `lk.redaction.enabled` to **trace/log metadata**, session-report logger metadata, and recording multipart part headers. 

**Force redaction with client side opt-in**
`record={"redaction": True}` to enable redaction for a session if the project setting is unset. `record=True` stays unchanged: without `record={"redaction": True}`, redaction is left to the project setting, and `False` is ignored when redaction is enabled globally for the project.

Safe to merge before redaction release as this is a client side addition, no-op for collector endpoints today.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -107,7 +107,7 @@ class RecordingOptions(TypedDict, total=False):
     * ``record=True``  → all on (backward compatible)
     * ``record=False`` → all off (backward compatible)
     * ``record={"audio": True, "traces": False}`` → granular
-    * ``record={"redaction": True}`` → preview collector-side redaction
+    * ``record={"redaction": True}`` → enable redaction for the session
     """
 
     audio: bool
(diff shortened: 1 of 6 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -115,19 +118,23 @@ class RecordingOptions(TypedDict, total=False):
     """Export OpenTelemetry logs. Defaults to ``True``."""
     transcript: bool
     """Upload the conversation transcript (chat history). Defaults to ``True``."""
+    redaction: bool
+    """Request collector-side redaction preview. ``False`` does not disable project redaction."""
```
```suggestion
    """Enable redaction. ``False`` does not disable project redaction."""
```

Our users don't have context about `collector`

theomonnom (approved):

PR #6498, opened 2026-07-21
Allow local VAD inference to use a dedicated executor

basnijholt:
## Summary

Add an optional caller-owned executor to the local inference VAD. When supplied, native prediction runs there instead of the event loop default executor. Existing behavior remains unchanged when omitted.

## Why

Applications embedding AgentSession alongside other blocking workloads may heavily use the default executor. VAD inference then waits behind unrelated work and can fall behind realtime audio. A caller-provided executor gives latency-sensitive inference an isolated queue without changing VAD internals or global event-loop configuration.

## Validation

- make check
- uv run pytest --plugin silero tests/test_vad.py

### DIFF
--- a/livekit-agents/livekit/agents/inference/vad.py
+++ b/livekit-agents/livekit/agents/inference/vad.py
@@ -54,7 +54,8 @@ class VAD(vad.VAD):
 
     The native model singleton is loaded once at module import (via the
     pybind11 ``.so`` constructor); each stream allocates its own per-instance
-    LSTM/context state.
+    LSTM/context state. Pass ``executor`` to isolate inference from the event
+    loop's default executor; the caller remains responsible for shutting it down.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


basnijholt:
Additional consumer context: [MindRoom](https://github.com/mindroom-ai/mindroom) lets users call its agents over MatrixRTC. Its cascaded voice path runs local VAD alongside the normal agent and tool runtime in one asyncio process.

The default executor is shared with unrelated blocking work, so a burst of agent/tool activity can delay the 32 ms VAD inference loop even though the native inference itself is fast. Allowing the application to provide a small dedicated executor isolates that latency-sensitive work while preserving today's behavior for every caller that omits the option. MindRoom would own and shut down the executor at call teardown.

longcw (approved):

longcw:
thanks for the pr! could you sign the CLA before we can merge?

basnijholt:
@longcw done! ✅

PR #6500, opened 2026-07-21
fix(voice): don't hold session.run() open on update_agent handoffs with long-lived on_enter

k-zaher:
## What

`session.run()` deadlocks when a function tool calls `session.update_agent(new_agent)` and the new agent's `on_enter` awaits an `AgentTask` (or `TaskGroup`) that needs a future user turn to complete.

The chain: `update_agent()` → `_update_activity_task` (watched by the active `RunResult` so the handoff result lands before run completion) → `_update_activity(agent)` with the default `wait_on_enter=True` → awaits the new agent's `on_enter` → which awaits an `AgentTask` that needs the *next* user turn. But the driver can't send the next turn until `run()` returns — a circular wait. The run hangs until an external timeout cancels everything (surfacing as `ToolError: AgentTask ... is cancelled`).

Both other activity-transition paths already guard against exactly this:

- `session.start()` passes `wait_on_enter=False` (`agent_session.py:916`), which is why the `examples/survey` pattern (agent `on_enter` awaiting a whole `TaskGroup`) works when the agent is the starting agent.
- `AgentTask.__await_impl` passes `wait_on_enter=False` with a comment describing this exact deadlock: *"on_enter may spawn nested AgentTasks that require user input, but session.run() can't return until all watched handles complete — creating a circular wait."*

`update_agent()` was the only path still waiting. Entering the *same* agent that works via `session.start()` through an `update_agent()` handoff deadlocks — reproduced on 1.6.4, 1.6.5, and 1.6.6.

## Fix

Pass `wait_on_enter=False` in `_update_activity_task`, matching the other two paths.

## Test

`tests/test_update_agent_long_on_enter.py` — FakeLLM-scripted: a greeter tool calls `update_agent()` into an agent whose `on_enter` awaits an `AgentTask` needing another user turn; asserts `run()` returns on the handoff turn and the task completes on the following turn. Deadlocks (times out) without the fix; passes with it.

Validation:
- `tests/test_agent_session.py`, `test_nested_agent_task.py`, `test_audio_recognition_handoff.py`, `test_agent_task_close_race.py`, `test_drain_timeout.py`, `test_tool_results_preserved_on_interruption.py`, `test_user_turn_exceeded.py`, `test_speech_handle_exception.py`, `test_agent_update_options.py` + the new test: 116 passed.
- `examples/survey/test_survey_agent.py`: 5 passed (live LLM).
- A production voice agent using this shape (greeter agent routing via `update_agent()` to an agent whose `on_enter` drives a multi-task loop): its full offline suite of ~1000 tests passes against this patch; on current releases every workflow test deadlocks.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1601,7 +1601,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         if old_task is not None:
             await old_task
 
-        await self._update_activity(agent)
+        await self._update_activity(agent, wait_on_enter=False)
 
     def _emit_debug_message(self, payload: dict[str, Any]) -> None:
         """:meta private: internal — emit a debug/trace payload to the debugger/recorder."""
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1603,6 +1603,11 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
 
         await self._update_activity(agent, wait_on_enter=False)
 
+        # watch on_enter so the run captures its output without awaiting it
+        if (activity := self._activity) is not None and activity._on_enter_task is not None:
+            if (run_state := self._global_run_state) is not None and not run_state.done():
+                run_state._watch_handle(activity._on_enter_task)
+
     def _emit_debug_message(self, payload: dict[str, Any]) -> None:
         """:meta private: internal — emit a debug/trace payload to the debugger/recorder."""
         st = Struct()
--- a/tests/test_update_agent_long_on_enter.py
+++ b/tests/test_update_agent_long_on_enter.py
@@ -79,6 +79,52 @@ def _build_fake_llm() -> FakeLLM:
     )
 
 
+class GreetingAgent(Agent):
+    """Agent whose on_enter does async work before speaking — the run must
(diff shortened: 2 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1601,7 +1601,7 @@ async def _update_activity_task(
         if old_task is not None:
             await old_task
 
-        await self._update_activity(agent)
+        await self._update_activity(agent, wait_on_enter=False)
```
longcw, line 1604: One thing though: `wait_on_enter=False` alone doesn't just stop *waiting* on `on_enter`, it stops the run from *tracking* it.  So the run can now complete before the new agent's `on_enter` emits its greeting.

On the common shape where `on_enter` does a bit of async work before speaking (`await something; self.session.generate_reply(...)`), `run()` returns before the greeting — and under `async with AgentSession(): await session.run(...)` the delayed `generate_reply` then raises "RuntimeError: AgentSession is closing"

```suggestion
        await self._update_activity(agent, wait_on_enter=False)

        # watch on_enter so the run captures its output without awaiting it
        if (activity := self._activity) is not None and activity._on_enter_task is not None:
            if (run_state := self._global_run_state) is not None and not run_state.done():
                run_state._watch_handle(activity._on_enter_task)
```

longcw (approved):

longcw:
thanks for the pr! could sign the CLA before we can merge

k-zaher:
I have signed the CLA but the bot is not picking up the updated status for some reason

PR #6515, opened 2026-07-22
Fix stale OpenAI Responses WebSocket reuse

### LINKED ISSUE #6513: OpenAI Responses WebSocket pool may reuse stale connections after concurrent requests
### Bug Description

When `openai.responses.LLM` uses its default WebSocket transport, concurrent
requests create multiple pooled WebSocket connections. After those connections
remain idle, a later LLM request can receive a connection that is no longer
writable.

The request then fails at `ws.send_str()` with:

```text
WARNING livekit.agents failed to generate LLM completion:
failed to send request over WebSocket, retrying in 0.1s
```

If multiple stale connections exist in the pool, the warning can occur multiple
times before a fresh/healthy connection is acquired:

```text
failed to send request over WebSocket, retrying in 0.1s
failed to send request over WebSocket, retrying in 2.0s
```

The request eventually succeeds, but the retries add noticeable latency to a
live voice turn.

This was observed after adding three concurrent application-level LLM warm-up
requests. The warm-ups grow the Responses WebSocket pool from one connection to
multiple connections. After an idle period, production requests started
encountering repeated send failures.

The current implementation appears susceptible to this behavior:

- `ConnectionPool.get()` checks `max_session_duration`, but does not validate
  whether an available connection is still open/usable before returning it.
- `_ResponsesWebsocket.generate_response()` acquires a pooled socket and writes
  to it directly with `ws.send_str(data)`.
- A send failure removes that connection through the pool context manager, but
  the outer LLM retry loop must fail once for each stale pooled connection.

Relevant source:

- https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/utils/connection_pool.py
- https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py

### Expected Behavior

Connections that are already known to be closed should be discarded before
they are returned to the OpenAI Responses transport.

A stale idle connection should not consume a full LLM retry attempt or add
retry backoff latency to a voice turn.

The existing retry behavior should remain in place for races where a connection
becomes unusable during the actual send operation.

### Reproduction Steps

```bash
The issue is timing/network dependent because the idle close can originate from
OpenAI, a NAT gateway, VPN, proxy, or another intermediary.

A representative reproduction is:


import asyncio

async def request(client: openai.responses.LLM) -> None:
    chat_ctx = llm.ChatContext()
    chat_ctx.add_message(role="user", content="Reply with OK.")
    await client.chat(chat_ctx=chat_ctx).collect()


async def main() -> None:
    client = openai.responses.LLM(
        model="gpt-4.1",
        use_websocket=True,
    )

    # A single Responses WebSocket supports one in-flight response, so this
    # causes the pool to create multiple connections.
    await asyncio.gather(*(request(client) for _ in range(3)))

    # Allow one or more pooled sockets to become stale in the environment.
    await asyncio.sleep(60)

    # This may first acquire stale pooled sockets and emit repeated warnings.
    await request(client)

    await client.aclose()


asyncio.run(main())


The observed application sequence was:

1. Three concurrent requests started.
2. All three completed.
3. The connections remained idle for approximately 30–90 seconds.
4. A subsequent request encountered two consecutive
   `failed to send request over WebSocket` warnings.
5. A later retry succeeded.

A deterministic unit test should be possible using fake WebSocket objects:

1. Populate `ConnectionPool` with connections whose `closed` state is true.
2. Add one healthy connection.
3. Acquire a connection for `_ResponsesWebsocket.generate_response()`.
4. Verify that closed connections are discarded without surfacing an outer LLM
   retry.
5. Verify that the healthy connection receives the payload.
```

### Operating System

macOS 26.5.1

### Models Used

_No response_

### Package Versions

```bash
livekit-agents==1.6.6
livekit-plugins-openai==1.6.6
openai==2.44.0
aiohttp==3.13.5
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
One possible minimal approach:

1. Add an optional connection validation callback to `ConnectionPool`.
2. Invoke it before returning an available pooled connection.
3. If validation fails, remove/close that connection and continue looking for a
   healthy connection.
4. Configure the OpenAI Responses pool to reject sockets where `ws.closed` is
   true.
5. Preserve the existing retry behavior for failures that occur after a
   connection passes validation.

For example, conceptually:


ConnectionPool(
    connect_cb=self._create_ws,
    close_cb=self._close_ws,
    validate_cb=lambda ws: not ws.closed,
    max_session_duration=3600,
)


An OpenAI-plugin-local stale-socket check could also avoid expanding the generic
`ConnectionPool` API.

WebSocket heartbeat/ping support may help detect remotely closed idle
connections, but that may be better handled separately because it affects
latency, traffic, and intermediary compatibility.
```

### Additional Context

OpenAI documents that one Responses WebSocket connection handles one in-flight
response at a time, so parallel work requires multiple connections:

https://developers.openai.com/api/docs/guides/deployment-checklist#use-websocket-mode

The main user impact is additional retry latency during a real-time voice turn.
The application-level workaround is to use a separate HTTP
`openai.responses.LLM(use_websocket=False)` instance for concurrent warm-up
requests so they do not expand the production WebSocket pool.

_No response_

MMYurt:
## Summary

- add an optional `validate_cb` to `ConnectionPool` so unavailable pooled connections can be discarded before reuse
- configure the OpenAI Responses WebSocket pool to reject connections whose `closed` state is already known
- add hermetic regression coverage for both the generic pool and the OpenAI Responses transport

Fixes #6513.

## Problem

Concurrent Responses API requests can grow the WebSocket pool to multiple connections. If one or more of those connections close while idle, a later request may reuse a known-closed socket and fail at `send_str()`. The outer LLM retry eventually recovers, but each stale connection can add a warning and retry backoff to a live voice turn.

This change removes connections that fail a cheap, synchronous validation before returning them from the pool. The OpenAI transport uses `not ws.closed` for that check. Existing retry behavior remains unchanged for races where a connection becomes unusable after validation or during the send itself.

## Testing

- `uv run pytest tests/test_connection_pool.py --unit -q` (5 passed)
- `uv run pytest tests/test_plugin_openai_responses.py --plugin openai -q` (3 passed)
- `make check` (format, lint, and strict mypy across 623 source files)

Additional local unit-suite validation completed 1,216 tests with 4 skipped. Nine `tests/test_room.py` cases could not set up because the local machine does not have the external `livekit-server` binary installed.

### DIFF
--- a/livekit-agents/livekit/agents/utils/connection_pool.py
+++ b/livekit-agents/livekit/agents/utils/connection_pool.py
@@ -25,6 +25,7 @@ class ConnectionPool(Generic[T]):
         mark_refreshed_on_get: bool = False,
         connect_cb: Callable[[float], Awaitable[T]] | None = None,
         close_cb: Callable[[T], Awaitable[None]] | None = None,
+        validate_cb: Callable[[T], bool] | None = None,
         connect_timeout: float = 10.0,
     ) -> None:
         """Initialize the connection wrapper.
@@ -34,11 +35,13 @@ class ConnectionPool(Generic[T]):
             mark_refreshed_on_get: If True, the session will be marked as fresh when get() is called. only used when max_session_duration is set.
             connect_cb: Optional async callback to create new connections
             close_cb: Optional async callback to close connections
+            validate_cb: Optional callback to check whether a connection can be reused
         """  # noqa: E501
         self._max_session_duration = max_session_duration
         self._mark_refreshed_on_get = mark_refreshed_on_get
         self._connect_cb = connect_cb
         self._close_cb = close_cb
+        self._validate_cb = validate_cb
         self._connections: dict[T, float] = {}  # conn -> connected_at timestamp
         self._available: set[T] = set()
         self._connect_timeout = connect_timeout
@@ -106,10 +109,12 @@ class ConnectionPool(Generic[T]):
             # try to reuse an available connection that hasn't expired
             while self._available:
                 conn = self._available.pop()
-                if (
+                is_valid = self._validate_cb is None or self._validate_cb(conn)
+                has_expired = not (
                     self._max_session_duration is None
                     or now - self._connections[conn] <= self._max_session_duration
-                ):
+                )
+                if is_valid and not has_expired:
                     if self._mark_refreshed_on_get:
                         self._connections[conn] = now
                     self.last_acquire_time = 0.0
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
@@ -73,6 +73,7 @@ class _ResponsesWebsocket:
         self._pool = utils.ConnectionPool[aiohttp.ClientWebSocketResponse](
             connect_cb=self._create_ws,
             close_cb=self._close_ws,
+            validate_cb=lambda ws: not ws.closed,
             max_session_duration=3600,
         )
 
--- a/tests/test_connection_pool.py
+++ b/tests/test_connection_pool.py
@@ -10,6 +10,7 @@ pytestmark = [pytest.mark.unit, pytest.mark.virtual_time, pytest.mark.no_concurr
 class DummyConnection:
     def __init__(self, id):
         self.id = id
+        self.is_usable = True
 
     def __repr__(self):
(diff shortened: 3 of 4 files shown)


longcw:
closing in favor of https://github.com/livekit/agents/pull/6523

PR #6516, opened 2026-07-22
fix(voice): skip committing late STT finals after a turn flush (#6504)

### LINKED ISSUE #6504: Turn commits with end_of_turn_probability: null when STT transcript arrives after turn flush, splitting one utterance into two turns
### Bug Description

# Turn commits with `end_of_turn_probability: null` when STT transcript arrives after turn flush, splitting one utterance into two turns

## Summary

When a turn is committed by VAD and the final STT transcript arrives shortly afterwards, the late transcript triggers a **second** turn commit with `end_of_turn_probability: null` and `unlikely_threshold: null`. The single spoken utterance is delivered to the LLM as two separate conversation items.

Two distinct problems:

1. **A turn commits with no end-of-turn decision.** `end_of_turn_probability` and `unlikely_threshold` are both `null`, and `speech_start_time` is `null`. The pipeline proceeds anyway.
2. **The log contradicts itself.** In one variant, `stt transcript arrived after a turn flush, skipping eot prediction` is logged, and an `eot prediction` event is then emitted with null fields on the next line.

The user-visible effect is that a continuous sentence reaches the LLM fragmented, with the second fragment having bypassed turn detection entirely.

## Environment

```
livekit                        1.1.13
livekit-agents                 1.6.5
livekit-api                    1.2.0
livekit-blingfire              1.1.0
livekit-local-inference        0.2.6
livekit-plugins-ai-coustics    0.3.0
livekit-protocol               1.1.19
Python 3.14.5, Windows 11
```

- `agent-starter-python` template, running in `console` mode
- STT: `deepgram/nova-3`, `language="multi"` (via `inference.STT`)
- Turn detection: `inference.TurnDetector()` (no arguments) — hosted inference path, **not** `livekit-plugins-turn-detector` (that package is not installed)
- `preemptive_generation=True`
- Turn detector reports `"model": "turn-detector-v1"`, `default_threshold: 0.56`

## Reproduction

Speak a sentence with a pause of roughly one second mid-phrase, so that VAD commits the turn before STT has finalised the trailing words.

Observed on multiple occasions across two sessions.

## Expected

Either the late transcript is merged into the already-committed turn, or it is genuinely skipped as the log message states. A turn should not commit with a null end-of-turn probability.

### Case 1 — enhancement disabled

Spoken: *"No. Date needed now."*

```
04:48:24.760 DEBUG  received user transcript
             {"user_transcript": "Date", "transcript_delay": 0.058}
04:48:25.794 DEBUG  eot prediction
             {"probability": 0.737298846244812, "unlikely_threshold": 0.56,
              "endpointing_delay": 0.3, "trigger": "vad", "from_cache": false}
04:48:25.805 DEBUG  user turn committed
             {"delay_completed": false, "source": "vad",
              "end_of_turn_probability": 0.737298846244812}
04:48:25.822 DEBUG  conversation_item_added
             {"role": "user", "text": "No. Date"}
04:48:25.857 DEBUG  received user transcript
             {"user_transcript": "needed now."}
04:48:25.864 WARNING  transcript arrives after turn has been committed. consider
             raising `min_delay` in the endpointing options to accommodate a slow stt.
04:48:25.876 DEBUG  eot prediction
             {"probability": null, "unlikely_threshold": null,
              "endpointing_delay": 0.3, "trigger": "stt", "from_cache": false}
04:48:26.158 DEBUG  user turn committed
             {"speech_start_time": null, "delay_completed": true, "source": "stt",
              "end_of_turn_probability": null, "unlikely_threshold": null}
04:48:26.191 DEBUG  conversation_item_added
             {"role": "user", "text": "needed now."}
```

One utterance, two conversation items, second turn committed on a null probability.

### Case 2 — the logged contradiction

Spoken: *"Yeah. Can you share the link of that property?"*

```
21:53:44.634 DEBUG  user turn committed
             {"delay_completed": false, "source": "vad",
              "end_of_turn_probability": 0.7216790318489075}
21:53:44.640 DEBUG  conversation_item_added
             {"role": "user", "text": "Yeah. Can you share the link of the link of that"}
21:53:44.756 DEBUG  received user transcript
             {"user_transcript": "property?"}
21:53:44.759 DEBUG  stt transcript arrived after a turn flush, skipping eot prediction
21:53:44.762 DEBUG  eot prediction
             {"probability": null, "unlikely_threshold": null,
              "endpointing_delay": 0.3, "trigger": "stt", "from_cache": false}
21:53:45.059 DEBUG  user turn committed
             {"speech_start_time": null, "end_of_turn_probability": null,
              "unlikely_threshold": null}
21:53:45.078 DEBUG  conversation_item_added
             {"role": "user", "text": "property?"}
```

`skipping eot prediction` is logged, then an `eot prediction` event is emitted 3 ms later with null fields.

## Ruled out

The starter template applies `ai_coustics` audio enhancement on the input path by default. I disabled it and re-ran; the behaviour reproduces identically. The enhancement layer is not the cause.

## Note on the `min_delay` suggestion

The warning suggests raising `min_delay` to accommodate slow STT. That may reduce how often the race occurs, but it does not address either problem above: a turn still commits with a null probability when it does occur, and the log still contradicts itself. Raising `min_delay` also directly increases response latency, which is costly in a telephony context.

## Not the same as #3070

#3070 (closed) was the end-of-turn model timing out under CPU pressure, resolved by moving to a larger instance. Here there is no timeout and no `CancelledError` — the prediction is emitted, after the turn has already flushed, with null fields. That issue was also on the local `turn_detector` plugin; this is the hosted `inference.TurnDetector()` path.

## Impact

In a voice pipeline where users produce short utterances with natural mid-sentence pauses, the LLM receives fragmented input and one of the fragments has bypassed turn detection. In my case the agent responded to `"No. Date"` before `"needed now."` had arrived.

### Expected Behavior

Either the late transcript is merged into the already-committed turn, or it is genuinely skipped as the log message states — a turn should not commit with end_of_turn_probability: null and unlikely_threshold: null.

A single continuous utterance should reach the LLM as one conversation item, not split across two turns where the second bypassed turn detection entirely.

### Reproduction Steps

```bash
Create a project from agent-starter-python (unmodified template).

Run uv run src/agent.py console.

Speak a sentence with a pause of roughly one second mid-phrase, so VAD commits the turn before STT has finalised the trailing words. Example: "No. Date … needed now."

Watch the debug log for transcript arrives after turn has been committed, immediately followed by an eot prediction with probability: null.

Observed on multiple occasions across two sessions. Reproduces with the template's default ai_coustics audio enhancement both enabled and disabled — the enhancement layer is not the cause.

The only deviation from the stock template in my runs was commenting out the room_options block to disable enhancement for the control test.
```

### Operating System

Windows 11

### Models Used

STT: deepgram/nova-3 (language="multi") via inference.STT | LLM: google/gemma-4-31b-it via inference.LLM | TTS: cartesia/sonic-3 via inference.TTS | Turn detection: inference.TurnDetector() — hosted, reports "model": "turn-detector-v1"

### Package Versions

```bash
livekit                        1.1.13
livekit-agents                 1.6.5
livekit-api                    1.2.0
livekit-blingfire              1.1.0
livekit-local-inference        0.2.6
livekit-plugins-ai-coustics    0.3.0
livekit-protocol               1.1.19
Python 3.14.5
```

### Session/Room/Call IDs

N/A - console mode, no Cloud session

### Proposed Solution

```python
Resolve the race in one direction rather than both: either merge the late transcript into the already-committed turn, or genuinely skip it as the "skipping eot prediction" log line claims. A turn should never commit with end_of_turn_probability and unlikely_threshold both null. Note that the warning's suggestion to raise min_delay reduces how often the race occurs but does not fix either problem — and it increases response latency, which is costly in telephony.
```

### Additional Context

Config runs entirely through LiveKit's hosted inference gateway (inference.LLM / STT / TTS / TurnDetector); no third-party model keys. Distinct from #3070, which was an EOT model timeout under CPU pressure on the local turn_detector plugin.

_No response_

SHAI-sambhav-suri:
A late STT final transcript could arrive after VAD had already committed and flushed the turn. With no prediction future left in flight, the eou bounce logged "skipping eot prediction" but then fell through and committed a second turn with a null end_of_turn_probability, splitting one utterance into two conversation items where the second bypassed turn detection.

Genuinely short-circuit that case: when a streaming detector has been flushed and there is no prediction in flight, skip the commit instead of shipping a null-probability turn. The transcript is left intact so the trailing words fold into the next turn rather than being lost.

Fixes livekit/agents#6504

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -1431,6 +1431,22 @@ class AudioRecognition:
             speech_start_time: float | None = None,
         ) -> None:
             endpointing_delay = self._endpointing.min_delay
+
+            # A late STT final transcript can arrive after the turn detector was
+            # already flushed for a committed turn. There is no prediction in flight,
+            # so continuing would commit a second turn with a null end_of_turn_probability
+            # and split one utterance into two turns (issue #6504). Genuinely skip here
+            # — as the log message states — leaving the transcript to fold into the next
+            # turn instead of shipping a fragment that bypassed turn detection.
+            if (
+                trigger == "stt"
+                and self._turn_detector_flushed
+                and self._turn_detector_prediction_fut is None
+                and isinstance(turn_detector, _StreamingTurnDetectorStream)
+            ):
+                self._on_missing_eot_prediction()
+                return
+
             user_turn_span = self._ensure_user_turn_span()
 
             end_of_turn_probability: float | None = None
--- a/tests/test_audio_recognition_turn_detection.py
+++ b/tests/test_audio_recognition_turn_detection.py
@@ -540,10 +540,14 @@ class TestPredictionFutureLifecycle:
         self, caplog: pytest.LogCaptureFixture
     ) -> None:
         """A late stt final after the turn was flushed must not start an
-        inference request; it warns once, then logs at debug level."""
+        inference request and must not commit a second turn (issue #6504):
+        committing here would ship a turn with a null ``end_of_turn_probability``
+        and split one utterance into two. It warns once, then logs at debug level,
+        and leaves the transcript untouched so it folds into the next turn."""
         caplog.set_level(logging.WARNING, logger="livekit.agents")
         ar = _make_full_recognition_for_eou()
         ar._turn_detector_flushed = True
+        ar._audio_transcript = "date needed now"
         chat_ctx = _make_chat_ctx_stub()
 
         for _ in range(2):
@@ -553,7 +557,13 @@ class TestPredictionFutureLifecycle:
 
         assert ar._turn_detector_stream.predict.call_count == 0
         ar._hooks.on_eot_prediction.assert_not_called()
-        flush_warnings = [r for r in caplog.records if "already flushed" in r.getMessage()]
+        # the late transcript is genuinely skipped, never committed with a null prediction
+        ar._hooks.on_end_of_turn.assert_not_called()
+        # transcript is preserved so it merges into the next turn instead of being lost
+        assert ar._audio_transcript == "date needed now"
+        flush_warnings = [
+            r for r in caplog.records if "after turn has been committed" in r.getMessage()
+        ]
         assert len(flush_warnings) == 1
 
     async def test_predict_timeout_signals_fallback_and_drops_future(self) -> None:


SHAI-sambhav-suri:
**What this fixes**

Fixes #6504. An utterance with a mid-phrase pause ("No." … ~1s … "Date needed
now.") was split into two conversation items, and the second committed with
`end_of_turn_probability: null` — reaching the LLM having bypassed turn detection.

**Why it happens**

VAD detects the silence and commits the turn, which flushes the streaming turn
detector (`_turn_detector_flushed = True`, prediction future cleared). STT then
finalizes the trailing words a beat later, re-entering
`_run_eou_detection(trigger="stt")`. With no prediction future left, the eou
bounce logs *"skipping eot prediction"* — but doesn't actually skip: it falls
through with `end_of_turn_probability = None` and commits a second, bypassed turn.

**The change**

Make the skip real. In `_bounce_eou_task`, short-circuit before committing when a
late STT final arrives after the detector was flushed and no prediction is in flight:

```python
if (
    trigger == "stt"
    and self._turn_detector_flushed
    and self._turn_detector_prediction_fut is None
    and isinstance(turn_detector, _StreamingTurnDetectorStream)
):
    self._on_missing_eot_prediction()
    return
  ```
    
No null-probability commit, and the log is now truthful. We return before
_audio_transcript is cleared, so the trailing words fold into the next turn
instead of being lost. The guard is scoped to this exact case, so the VAD path,
manual mode, and text detectors are unaffected.

**Tests**
Extended test_late_stt_final_after_flush_short_circuits to assert the contract
that was missing: on_end_of_turn is not called and the transcript is preserved
(plus no inference started / no on_eot_prediction). It fails on the old behavior
(commits twice) and passes with the fix. Full --audio_eot suite and
voice-related --unit tests pass; ruff and mypy clean.

Propconnect-ai:
Reporter of #6504 here — this matches what I observed.

The four-condition guard covers the exact path from my logs: STT-triggered, detector already flushed, no prediction in flight, streaming detector. Returning before the commit means no phantom `eot prediction` event and no null-probability turn, so both halves of the issue are addressed rather than just the commit.

Preserving the transcript so it folds into the next turn is the right call. In my repro the trailing fragment became its own conversation item ("No. Date" then "needed now."), and the agent responded to the first half before the second arrived. Merging forward fixes the behaviour, not just the log.

One question on coverage: the guard requires `_turn_detector_prediction_fut is None`. I saw two variants in my sessions —

1. `stt transcript arrived after a turn flush, skipping eot prediction` followed by an `eot prediction` with null fields
2. the `transcript arrives after turn has been committed. consider raising min_delay...` warning, also followed by a null prediction

Both ended with `end_of_turn_probability: null` and `speech_start_time: null`, but I don't know from the logs alone whether a prediction future was in flight in either case. If a late final can arrive while one is still pending, that path would fall through this guard. Worth confirming, or explicitly out of scope.

Happy to re-run my repro against this branch if that's useful.

Propconnect-ai:
Following up on the coverage question - happy to just write it. I can push a commit to this branch adding the pending-future case, or open a separate PR if you'd rather keep this one focused.

Propconnect-ai:
Re-ran the repro on LiveKit Cloud (1.6.5) — session with audio and support access enabled:

https://cloud.livekit.io/projects/p_19mwyin9l4l/sessions/RM_DBfSGxQqRj3t/observability?mode=transcript

Caught the null-probability turn again on an email input ("It's john at" commits via VAD, then "example dot com" STT final arrives after the flush): end_of_turn_probability: null, unlikely_threshold: null, with the "transcript arrives after turn has been committed" warning. This is variant 2 from my comment above — happy to dig into the in-flight _turn_detector_prediction_fut case if you want me to try to trigger it.

PR #6520, opened 2026-07-22
feat(fishaudio): add normalize_loudness and generation-tuning TTS options

adrian-cowham:
## Summary

Exposes five Fish Audio TTS params the plugin did not forward, per the [Fish Audio TTS API](https://docs.fish.audio/api-reference/endpoint/openapi-v1/text-to-speech):

**Top-level request fields**
- `max_new_tokens` — max audio tokens generated per text chunk (Fish default 1024)
- `min_chunk_length` — minimum characters before splitting a chunk, 0–100 (Fish default 50)
- `condition_on_previous_chunks` — use previous audio as context for voice consistency (Fish default true)
- `early_stop_threshold` — early stopping threshold, 0–1 (Fish default 1.0)

**Prosody sub-object** (alongside the existing `speed`/`volume`)
- `normalize_loudness` — consistent output loudness (Fish default true). **S2-Pro family only:** dropped with a warning at set time (constructor, `update_options`, and on a model switch to `s1`) rather than sent for a model that cannot honor it

All five default to `NOT_GIVEN` and are omitted from the request unless set, so Fish's server-side defaults stay in effect and the default wire request is byte-for-byte unchanged. Values are validated in both the constructor and `update_options`. Follows the same pattern as #6456.

A sibling change adds the same params to the LiveKit inference gateway (livekit/agent-gateway#1066).

## Testing

Explicit unit tests in `tests/test_plugin_fishaudio_tts.py` cover prosody placement of `normalize_loudness`, omission-when-unset, constructor and `update_options` forwarding, and range validation. `ruff check` / `ruff format` clean.

### DIFF
--- a/livekit-plugins/livekit-plugins-fishaudio/livekit/plugins/fishaudio/tts.py
+++ b/livekit-plugins/livekit-plugins-fishaudio/livekit/plugins/fishaudio/tts.py
@@ -58,6 +58,11 @@ class _TTSOptions:
     mp3_bitrate: MP3Bitrate
     opus_bitrate: OpusBitrate
     normalize: bool
+    normalize_loudness: NotGivenOr[bool]
+    max_new_tokens: NotGivenOr[int]
+    min_chunk_length: NotGivenOr[int]
+    condition_on_previous_chunks: NotGivenOr[bool]
+    early_stop_threshold: NotGivenOr[float]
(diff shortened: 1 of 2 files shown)


tinalenguyen (approved):
lgtm, thank you!

PR #6521, opened 2026-07-22
feat(ipc): configure entrypoint shutdown timeout

### LINKED ISSUE #6512: Allow configuring the job entrypoint shutdown grace period
### Feature Type

Would make my life easier

### Feature Description

Allow the job entrypoint shutdown grace period to be configured on `AgentServer` / `ServerOptions`.

When a worker shuts down, `_JobProc` currently waits for the job entrypoint for a hard-coded 15 seconds before cancelling it. Some applications need shutdown to terminate an active call immediately rather than allow voice playout or other entrypoint work to drain for that full grace period.

A setting such as `entrypoint_shutdown_timeout: float = 15.0`, passed through to both process and thread job executors, would preserve existing behavior by default while allowing `0.0` to cancel the entrypoint immediately.

The current options do not control this wait. `drain_timeout=0` skips worker draining, while `shutdown_process_timeout` controls the outer job-process shutdown deadline; neither replaces the hard-coded entrypoint grace period.

The remaining workarounds are to accept the additional 15 seconds during shutdown or maintain a downstream patch/fork.

### Additional Context

Observed on `livekit-agents==1.6.6`: after the worker logs that it is shutting down, it continues active session/voice work for 15 seconds, then logs `entrypoint did not exit in time, cancelling` and proceeds with application shutdown callbacks.

The proposed setting should only alter how long LiveKit waits before cancelling the entrypoint. Existing cleanup and shutdown callbacks should continue to run after cancellation.

tmtcam-042:
## Summary

The job process currently waits for the entrypoint for a hard-coded 15 seconds during shutdown before cancelling it. This makes that grace period configurable through `ServerOptions` and `AgentServer` while preserving the existing 15-second default.

Setting `entrypoint_shutdown_timeout=0.0` cancels the entrypoint immediately, allowing applications to terminate active calls without waiting for voice playout or other entrypoint work to drain. Shutdown callbacks continue to run after cancellation.

Closes #6512.

## Changes

- add `entrypoint_shutdown_timeout` to `ServerOptions`, `AgentServer`, and `update_options`
- pass the setting through `ProcPool` to process and thread job executors
- use the configured timeout when waiting for the job entrypoint during shutdown
- cover the default, option forwarding, updates, and immediate-cancellation behavior

## Testing

- `uv run pytest tests/test_ipc.py tests/test_worker_config.py --unit -q` — 26 passed
- `make check` — formatting, Ruff, and mypy passed across all 623 source files

### DIFF
--- a/livekit-agents/livekit/agents/ipc/job_proc_executor.py
+++ b/livekit-agents/livekit/agents/ipc/job_proc_executor.py
@@ -30,6 +30,7 @@ class ProcJobExecutor(SupervisedProc):
         inference_executor: InferenceExecutor | None,
         initialize_timeout: float,
         close_timeout: float,
+        entrypoint_shutdown_timeout: float,
         session_end_timeout: float,
         memory_warn_mb: float,
         memory_limit_mb: float,
@@ -60,6 +61,7 @@ class ProcJobExecutor(SupervisedProc):
         self._job_entrypoint_fnc = job_entrypoint_fnc
         self._session_end_fnc = session_end_fnc
         self._simulation_end_fnc = simulation_end_fnc
+        self._entrypoint_shutdown_timeout = entrypoint_shutdown_timeout
         self._session_end_timeout = session_end_timeout
         self._inference_executor = inference_executor
         self._inference_tasks: set[asyncio.Task[None]] = set()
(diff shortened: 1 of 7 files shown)


tmtcam-042:
Closed because reviewer was not auto-assigned, don't think CI picks up when a PR promotes from draft to ready!

PR #6524, opened 2026-07-23
Add 'eleven_v3_conversational' to model list

devb-enp:
eleven_v3_conversational is added to some of the early access enterprise customers


theomonnom (approved):

devb-enp:
I've tested eleven_v3_conversational with a custom script and it was working for my API key since my company has access to this model but when testing in livekit, it gave errors. I don't know why it's not working. I've attached a tracelog below 
[traceback.log](https://github.com/user-attachments/files/30331801/traceback.log)
@theomonnom Can you please check this?

Issue #5742, opened 2026-05-15
Gemini realtime: trailing audio truncated when a tool call ends the turn


joneriktalvio:
# Gemini realtime: trailing audio truncated when a tool call ends the turn

## Summary

`livekit/plugins/google/realtime/realtime_api.py:_handle_tool_calls` closes the
current generation's `audio_ch` synchronously on receipt of a `toolCall`
message. Because Gemini Live emits `toolCall` mid-turn (before flushing the
final `serverContent.modelTurn.parts` audio chunks), the close is premature:
the framework's speech handle resolves `_done_fut`, `RunContext.wait_for_playout()`
returns as a no-op, and any tool that does "wait for playout then tear down
the room" (e.g. an `end_call` tool) cuts the audio mid-syllable.

OpenAI's Realtime API doesn't have this shape — its plugin only closes audio
on `response.done`, which the API guarantees is sent after the final audio
delta.

PR #2439 fixed a *related* downstream symptom (transcript not finalised) on
the now-removed `beta/realtime/` path. Audio truncation is a separate
symptom of the same upstream cause and is still present on `main`.

## Reproduction

Minimal recipe (no SIP/Twilio needed — bug is at the LiveKit-room layer):

1. Run a local LiveKit dev server (`docker run livekit/livekit-server --dev`).
2. Connect two participants to the same room:
   - **Agent**: `AgentSession(llm=google.realtime.RealtimeModel(...))` with
     instructions *"Say 'Goodbye and have a great day.' then call the
     `end_call` tool."* and a single tool:

     ```python
     @function_tool()
     async def end_call(ctx: RunContext) -> str:
         await ctx.wait_for_playout()
         await lk_api.room.delete_room(api.DeleteRoomRequest(room=room_name))
         return "Call ended"
     ```

   - **Listener**: subscribes to the agent's audio track via
     `rtc.AudioStream`, writes received PCM frames to a WAV.
3. Run the same scenario against `gpt-realtime-2` and a Gemini Live model.
   Compare the WAVs.

The Gemini WAV ends at the last non-silent sample (zero trailing silence,
i.e. mid-syllable cut). The OpenAI WAV ends with the model's natural
silence pad. Same prompt, same tool, same room teardown logic; only the
realtime model differs.

A fuller automated harness — with quiescence-based deferred-close patching
for A/B, a macOS-`say` user-input fixture for audio-driven turns, and a
WAV trailing-silence analyzer for verdicts — is available on request.

[...]

PR #6526, opened 2026-07-23
fix(google): preserve non-blocking realtime generation

usluio:
## Summary

- keep Gemini realtime output streams open when a `NON_BLOCKING` function call arrives
- continue using the existing immediate finalization behavior for blocking function calls
- add regression coverage for both blocking and non-blocking tool behavior

## Why

Gemini 2.5 asynchronous function calling allows the model to continue the active conversation while a non-blocking tool executes. The Google realtime plugin currently calls `_mark_current_generation_done()` for every tool call, which closes the audio and text streams before trailing output in the same turn can be delivered.

This change only defers finalization for declarations configured with `Behavior.NON_BLOCKING`. The generation still closes on Gemini's existing `generation_complete` and `turn_complete` events. Blocking calls retain the current behavior.

This is intentionally narrower than #6285, which addresses blocking/tool-terminated turns such as Gemini 3.1 with an audio-drain timeout.

Related: #5742

## Tests

- `uv run --python 3.12 pytest tests/test_plugin_google_realtime.py --unit -q`
- `uv run --python 3.12 ruff check livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py tests/test_plugin_google_realtime.py`
- `uv run --python 3.12 ruff format --check livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py tests/test_plugin_google_realtime.py`
- `uv run --python 3.12 mypy --untyped-calls-exclude=smithy_aws_core -p livekit.plugins.google`

Full repository type checking is currently blocked locally by a missing `boto3` stub in the AWS experimental plugin.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -1443,6 +1443,12 @@ class RealtimeSession(llm.RealtimeSession):
                     arguments=arguments,
                 )
             )
+
+        # NON_BLOCKING calls may be followed by more output in the same turn.
+        # Keep the generation open until Gemini sends its completion events.
+        if self._opts.tool_behavior == types.Behavior.NON_BLOCKING:
+            return
+
         self._mark_current_generation_done()
 
     def _handle_tool_call_cancellation(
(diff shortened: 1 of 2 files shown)


usluio:
Closing this in favor of a fresh PR rebased on the current LiveKit Agents 1.6.8 codebase.

PR #6528, opened 2026-07-23
(expressive mode): add fish audio

cshape:
adds fish audio (s2 models) to expressive mode, same shape as inworld/xai/cartesia:

- emotion tags (discrete vocabulary: regretful, hopeful, happy, excited, curious, surprised, sad, empathetic, sarcastic) -> `[very EMOTION]`
- breaks -> `[break]` / `[long-break]` by duration
- emphasis, the only wrapping prosody -> prefix form: `are you [emphasis] sure`
- sounds: `[laughing]`, `[chuckling]`, `[clear throat]`, governed by the `laughing` / `reflex_sounds` steering fields; disabled sounds are dropped from the advertised vocabulary and examples like the other providers
- disfluent few-shot examples render only while `disfluencies` is enabled, so examples never contradict the delivery guidelines

no `_MAX_INPUT_LEN` entry on purpose: fish's markers are sentence-scoped (every sentence carries its own emotion tag), so we stick with normal per-sentence synthesis instead of batching longer expressive chunks — no steering is lost and time-to-first-audio stays low.

verified live through the inference gateway (`fishaudio/s2.1-pro` + gemma-4-31b-it) and with offline sweeps across several llms checking tag placement, conversion, transcript stripping, and steering filters.

### DIFF
--- a/livekit-agents/livekit/agents/tts/_provider_format.py
+++ b/livekit-agents/livekit/agents/tts/_provider_format.py
@@ -11,6 +11,7 @@ Provider docs:
 - Inworld: https://docs.inworld.ai/tts/best-practices/prompting-for-tts-2
 - xAI: https://docs.x.ai/developers/model-capabilities/audio/text-to-speech
 - xAI: https://docs.x.ai/developers/model-capabilities/audio/voice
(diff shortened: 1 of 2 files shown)


cshape:
rebased onto main as #6631

PR #6531, opened 2026-07-23
fix(voice): retain concrete language for turn detection

### LINKED ISSUE #6530: MutlilingualModel Turn Detector model does not support "multi" language code
### Bug Description

It says in the documentation that MutlilingualModel supports LanguageCode, but in LanguageCode there is an example that specifies LanguageCode("multi") -> "multi".  Which itself is not wrong, but confusing. Could we instead type LanguageCode as an Enum? so we'd know that "multi" is not part of that language list?

On another note, Deepgram nova3 multilingual can predict that the language detected for a SpeechEvent is "multi". So passing this value directly to `MutlilingualModel` simply fails. There is a detect_language parameter, but it only takes a boolean as a parameter, but looking at Deepgram, it is possible to narrow the detected language: https://developers.deepgram.com/docs/language-detection#restricting-the-detectable-languages

For example, detect_language=en&detect_language=es will choose either English or Spanish as the detected language.

Or there could be a way to ensure that the detected language in the MutlilingualModel Turn Detector model is always a valid code by adding a normalizing step before calling the model. Or add a better logic around using "multi" with the turn detector model by reusing last language value that was not "multi"

### Expected Behavior

The turn detection model still works even if "mutli" is passed to it, or have the STT models not predict "multi" in the SpeechEvent

### Reproduction Steps

```bash
1. In a session with a Voice Agent that uses Livekit Inference Deepgram nova3 multi and MutlilingualModel,
2. Speak a quick ambiguous word ("we" in English, or "oui" in French), Look at the SpeechEvent and see that the detected language is "multi", see that the turn handling model does not work because it does not allow "multi".
...
```

### Operating System

MacOS

### Models Used

Deepgram nova3

### Package Versions

```bash
[[package]]
name = "livekit"
version = "1.1.13"

[[package]]
name = "livekit-agent"
version = "0.3.46"

[[package]]
name = "livekit-agents"
version = "1.6.6"

[[package]]
name = "livekit-api"
version = "1.2.0"

[[package]]
name = "livekit-protocol"
version = "1.1.20"

[[package]]
name = "livekit-plugins-turn-detector"
version = "1.6.6"

[[package]]
name = "livekit-plugins-silero"
version = "1.6.6"

[[package]]
name = "livekit-api"
version = "1.2.0"
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

Oxygen56:
Fixes #6530

## Summary
- keep `multi` and `auto` language-detection modes from replacing the last concrete language
- share the same language-tracking logic across final and preflight transcripts
- add regression coverage for ambiguous multilingual results and later concrete-language updates

## Root Cause
`AudioRecognition` stored every initial STT language value in `_last_language`. When a multilingual STT provider returned the mode value `multi` for an ambiguous utterance, the turn detector treated it as a concrete language code, could not find a matching threshold, and skipped end-of-turn prediction.

The STT event still retains the provider's original language value. Only the language tracked for language-specific turn-detection behavior ignores non-specific detection modes.

## Verification
- `uv run --no-sync pytest tests/test_audio_recognition_turn_detection.py::TestLanguageTracking --audio_eot -q` -> 3 passed
- `uv run --no-sync ruff check livekit-agents/livekit/agents/voice/audio_recognition.py tests/test_audio_recognition_turn_detection.py` -> passed
- `uv run --no-sync ruff format --check livekit-agents/livekit/agents/voice/audio_recognition.py tests/test_audio_recognition_turn_detection.py` -> passed
- `uv run --group typing mypy -p livekit.agents` -> passed (205 source files)
- `uv run --no-sync pytest --unit -q` -> 1142 passed, 4 skipped before the repository's concurrent-test runner stopped with 9 event-loop errors on Python 3.13

## Notes for Reviewers
- If no concrete language has been detected yet, `multi` leaves the tracked language unset and the existing endpointing fallback remains unchanged.
- No public API or STT event payload is changed.

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -51,6 +51,7 @@ if TYPE_CHECKING:
     from .agent_session import AgentSession
 
 MIN_LANGUAGE_DETECTION_LENGTH = 5
+_NON_SPECIFIC_LANGUAGE_CODES = frozenset({"auto", "multi"})
 # Mirrors turn_detector.base.MAX_HISTORY_TURNS for tracing
 _EOU_MAX_HISTORY_TURNS = 6
 # backoff before recreating the stt stream after an unrecoverable error
@@ -367,6 +368,13 @@ class AudioRecognition:
                     if self._turn_detector_stream is not None:
                         self._turn_detector_stream.cancel_inference()
                     self._turn_detector_prediction_fut = None
+
+    def _update_last_language(self, language: LanguageCode, transcript: str) -> None:
+        if not language or language.language in _NON_SPECIFIC_LANGUAGE_CODES:
+            return
+
+        if not self._last_language or len(transcript) > MIN_LANGUAGE_DETECTION_LENGTH:
+            self._last_language = language
 
     @property
     def _input_started_at(self) -> float | None:
@@ -1163,10 +1171,7 @@ class AudioRecognition:
             language = ev.alternatives[0].language
             confidence = ev.alternatives[0].confidence
 
-            if not self._last_language or (
-                language and len(transcript) > MIN_LANGUAGE_DETECTION_LENGTH
-            ):
-                self._last_language = language
+            self._update_last_language(language, transcript)
 
             self._final_transcript_received.set()
             if not transcript:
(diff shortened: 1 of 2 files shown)


longcw (approved):

longcw:
could you sign the CLA before we merge?

PR #6534, opened 2026-07-24
feat(elevenlabs): support previous_text on realtime STT

### LINKED ISSUE #6074: feat(elevenlabs): support previous_text context on STT realtime input chunks
### Feature Type

Would make my life easier

### Feature Description

The ElevenLabs realtime STT WebSocket API accepts a `previous_text` field on the `input_audio_chunk` message that primes the model with text context to improve accuracy. Per the [docs](https://elevenlabs.io/docs/api-reference/speech-to-text/v-1-speech-to-text-realtime#send.inputAudioChunk), it "can only be sent alongside the first audio chunk."

The `livekit-plugins-elevenlabs` STT plugin doesn't expose this. In `stt.py`, every `input_audio_chunk` is built with only `message_type`, `audio_base_64`, `commit`, and `sample_rate`, so there's no way to pass context.

This helps agent scenarios where prior conversation context (names, domain/medical terms, homophones) would disambiguate the next utterance.

Proposed implementation, matching the existing `keyterms`/`no_verbatim` style: add `previous_text: NotGivenOr[str]` to `STTOptions` and `STT.__init__`, then send it as the first `input_audio_chunk` per connection (respecting the first-chunk-only constraint). Applies to the `scribe_v2_realtime` path.

I already have this implemented and would happily open the PR if the maintainers are fine with the approach. Filing this first per `CONTRIBUTING.md` to confirm scope.

`keyterms` (#5618) covers vocabulary biasing but not free-form preceding-text context, so there's no current way to feed conversational context into realtime STT.

### Additional Context

Complementary to #5618 (`keyterms`/`no_verbatim`/`enable_logging`); `previous_text` is a distinct field not covered there.

tulanisahil1-beep:
## Summary
- Add `previous_text` to ElevenLabs realtime STT (`scribe_v2_realtime`)
- After each WebSocket connect (including reconnect), send it on a priming empty `input_audio_chunk` so it is always the first chunk on the connection (ElevenLabs requires this)

Fixes #6074

## Test plan
- [x] `uv run pytest tests/test_plugin_elevenlabs_stt.py --plugin elevenlabs`

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
@@ -76,6 +76,7 @@ class STTOptions:
     keyterms: NotGivenOr[list[str]]
     no_verbatim: bool
     enable_logging: bool
+    previous_text: str | None
 
 
 class STT(stt.STT):
@@ -96,6 +97,7 @@ class STT(stt.STT):
         keyterms: NotGivenOr[list[str]] = NOT_GIVEN,
         no_verbatim: NotGivenOr[bool] = NOT_GIVEN,
         enable_logging: bool = True,
+        previous_text: NotGivenOr[str] = NOT_GIVEN,
     ) -> None:
         """
         Create a new instance of ElevenLabs STT.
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
```diff
@@ -179,6 +183,7 @@ def __init__(
             keyterms=keyterms,
             no_verbatim=no_verbatim if is_given(no_verbatim) else False,
             enable_logging=enable_logging,
+            previous_text=previous_text if is_given(previous_text) else None,
```
longcw: should we add a warning and ignore it when it's set for other models than the scribe v2 realtime?

tulanisahil1-beep: Good call — done. We now warn and ignore `previous_text` when it's set on models other than `scribe_v2_realtime`, matching the `server_vad` pattern.

longcw (approved):

PR #6535, opened 2026-07-24
Add first interim STT latency metric

### LINKED ISSUE #6533: STT time-to-first-partial metric (speech onset → first interim), the STT analog of ttft/ttfb
### Feature Type

Would make my life easier

### Feature Description

STT has no head-latency metric. `LLMMetrics` has `ttft` and `TTSMetrics` has `ttfb`, but `STTMetrics` exposes only `duration` (0.0 for streaming), `audio_duration`, and `streamed`. There is no equivalent for "how fast did the first word come back."

Two different STT latencies exist, and only the tail is measurable today:

| Interval | Meaning | Status |
|---|---|---|
| end-of-speech → FINAL transcript | `EOUMetrics.transcription_delay` | exists (includes endpointing) |
| audio-end → FINAL transcript | proposed in #5063 / PR #4966 | in review |
| **speech onset → first INTERIM** | **first-partial responsiveness** | **missing** |

The head metric is the one that predicts barge-in and perceived snappiness, and it is endpointing-free. `transcription_delay` cannot stand in for it: it is dominated by the endpointing/EOU wait, so it moves when you tune silence thresholds even if the STT's first-word speed never changed.

Both operands are already present in `AudioRecognition._on_stt_event`: the VAD-backdated onset `self._speech_start_time`, and the `INTERIM_TRANSCRIPT` branch. The metric is a wall-clock subtraction there, no provider timestamps involved:

```python
# AudioRecognition._on_stt_event, INTERIM_TRANSCRIPT branch
elif ev.type == stt.SpeechEventType.INTERIM_TRANSCRIPT:
    if self._first_interim_at is None and self._speech_start_time is not None:
        first_interim_delay = time.time() - self._speech_start_time
```

Proposed carrier (leaning `EOUMetrics`, since that is where it is computed and it stays symmetric with `transcription_delay`):

```python
first_interim_delay: float
"""Seconds from speech onset (VAD start) to the first interim transcript.
The STT-responsiveness analog of LLMMetrics.ttft / TTSMetrics.ttfb.
Unset / 0.0 when the STT emits no interims (non-streaming or interim_results=False)."""
```

Alternative placement is `STTMetrics.first_interim_latency`, but the plugin never sees VAD onset, so it would need framework enrichment anyway. Happy to take direction on which surface you'd prefer.

Why the #5063 `end_time`-reliability concern does not apply here: this metric never reads a provider `end_time`. Onset comes from VAD; the first interim's timestamp is just its wall-clock arrival. It is unaffected by the endpointing-accuracy objection that blocks the tail metric.

- `EOUMetrics.transcription_delay`: measurable today, but it is the tail (end-of-speech → FINAL) and folds in endpointing, so it conflates STT speed with EOU tuning.
- Timing `user_input_transcribed` at the `AgentSession` layer: works downstream but adds the session emit hop on top of the real STT latency, so it overstates it and drifts with unrelated scheduling.
- Both are proxies. The value already exists one layer down in `_on_stt_event`; surfacing it removes the guesswork.

### Additional Context

- Not a duplicate of #5063 / #4966 / #4795. Those are all the tail (audio/speech end → FINAL). This is the head (speech onset → first interim), and the two are independent numbers.
- The `_on_stt_event` mechanism above was confirmed by a community member on the LiveKit forum.
- I have this working as a downstream proxy in my own tooling and would be glad to bring a PR (latch + field + test) if the direction and carrier are agreeable.

This thread was first started at [Livekit Community](https://community.livekit.io/t/is-there-an-stt-time-to-first-partial-metric-the-ttft-ttfb-analog-or-am-i-missing-it/1701?u=mahimai_raja)

spyrux:
closes #6533 — STT time-to-first-partial metric (https://github.com/livekit/agents/issues/6533)
  ## Summary

  Adds `first_interim_delay`: the elapsed time from VAD detecting speech onset to the first interim (partial) STT
  transcript for a user turn.

  The metric is:

  - captured once, on the first interim STT event for each turn
  - reset when the user turn is cleared or completed
  - exposed in `EOUMetrics`
  - included in the per-turn metrics report
  - recorded as the `lk.agents.turn.first_interim_delay` OpenTelemetry histogram

  A missing interim transcript is represented as unavailable in the turn report and `0.0` in the existing `EOUMetrics`
  event convention.

  ## Tests

  ```text
  pytest tests/test_end_of_turn_metrics.py tests/test_stt_first_interim_metric.py

  - verifies the delay is calculated from speech onset
  - verifies the metric is unavailable when no interim transcript arrives

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -63,6 +63,7 @@ class _EndOfTurnMetrics:
     stopped_speaking_at: float | None
     transcription_delay: float | None
     end_of_turn_delay: float | None
+    first_interim_delay: float | None
 
 
 @dataclass
@@ -82,6 +83,7 @@ def _compute_end_of_turn_metrics(
     last_speaking_time: float | None,
     last_final_transcript_time: float | None,
     now: float,
+    first_interim_time: float | None = None,
 ) -> _EndOfTurnMetrics:
     """Compute the end-of-turn timing metrics from the captured turn anchors.
 
@@ -107,6 +109,7 @@ def _compute_end_of_turn_metrics(
             stopped_speaking_at=None,
             transcription_delay=None,
             end_of_turn_delay=None,
+            first_interim_delay=None,
         )
 
     return _EndOfTurnMetrics(
@@ -114,6 +117,11 @@ def _compute_end_of_turn_metrics(
         stopped_speaking_at=last_speaking_time,
         transcription_delay=max(last_final_transcript_time - last_speaking_time, 0),
         end_of_turn_delay=max(now - last_speaking_time, 0),
+        first_interim_delay=(
+            max(first_interim_time - speech_start_time, 0)
+            if first_interim_time is not None
+            else None
+        ),
     )
 
 
@@ -271,6 +279,7 @@ class AudioRecognition:
         self._last_final_transcript_time: float | None = None
         self._last_speaking_time: float | None = None
         self._speech_start_time: float | None = None
+        self._first_interim_time: float | None = None
 
         # used for manual commit_user_turn
         self._final_transcript_received = asyncio.Event()
@@ -979,6 +988,7 @@ class AudioRecognition:
         self._final_transcript_confidence = []
         self._last_final_transcript_time = None
         self._speech_start_time = None
+        self._first_interim_time = None
         self._last_speaking_time = None
         self._vad_speech_started = False
         self._user_turn_committed = False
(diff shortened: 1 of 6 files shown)


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1270,6 +1283,8 @@ async def _on_stt_event(self, ev: stt.SpeechEvent) -> None:
                 )
 
         elif ev.type == stt.SpeechEventType.INTERIM_TRANSCRIPT:
+            if self._first_interim_time is None and self._speech_start_time is not None:
```
Sanjays2402: This marks an interim as received before checking that it contains any text. Deepgram v1 forwards non-final `Results` even when `alt["transcript"]` is empty, so the first blank result wins the timestamp and the metric no longer measures first transcript latency; gate this on `ev.alternatives[0].text`.

PR #6539, opened 2026-07-24
feat(telemetry): include agent_name as an OTLP resource attribute

jmcclanahan:
Stamp lk.agent_name onto the telemetry resource when the job carries an agent name (explicit dispatch), so LiveKit Cloud agent insights can attribute per-agent metrics for self-hosted workers. Hosted agents also receive identity attrs via OTEL_RESOURCE_ATTRIBUTES; attributes passed to Resource.create() take precedence over the env var, and both sources derive the name from the same WorkerOptions value.

Confirmed working with session: https://cloud.livekit.io/projects/p_5zxtbufktnq/sessions/RM_NDwUfhhxaCGh

### DIFF
--- a/livekit-agents/livekit/agents/telemetry/traces.py
+++ b/livekit-agents/livekit/agents/telemetry/traces.py
@@ -236,6 +236,7 @@ def _setup_cloud_tracer(
     *,
     room_id: str,
     job_id: str,
+    agent_name: str = "",
     observability_url: str,
     enable_traces: bool = True,
     enable_logs: bool = True,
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -278,7 +279,15 @@ def __call__(self) -> dict[str, str]:
...
-    resource = Resource.create({SERVICE_NAME: "livekit-agents", **base_metadata})
+    resource_attributes: dict[str, AttributeValue] = {
+        SERVICE_NAME: "livekit-agents",
+        **base_metadata,
+    }
+    if agent_name:
+        # identifies the agent for LiveKit Cloud agent insights (explicit dispatch
+        # only; the default dispatch has no agent name)
+        resource_attributes[trace_types.ATTR_AGENT_NAME] = agent_name
+    resource = Resource.create(resource_attributes)
```
chenghao-mou: nit: we can add the agent name to `session_metadata` so it is available for both traces and logs (`resource` is only used in traces rn)

chenghao-mou (approved):

livekit-agents/livekit/agents/telemetry/traces.py
jmcclanahan: Yep, added

PR #6542, opened 2026-07-24
fix(speechify): restore caller header, close owned httpx client

luke-speechify:
Restores Speechify-Caller header, close owned httpx client

#6327's rewrite onto the official speechify-api SDK dropped the Speechify-Caller attribution header by mistake (previously sent manually via aiohttp as `x-caller`) and never closed the AsyncSpeechify client it constructs internally, per the pre-merge review finding that went unaddressed.

Construct the client with a dedicated httpx.AsyncClient carrying the Speechify-Caller default header (merges fine with the SDK's per-request Authorization/X-Fern-* headers), and close it in aclose() when this TTS instance owns it - mirroring the ownership pattern other plugins (openai) use for injected vs. self-constructed clients.


tinalenguyen (approved):
thank you for the follow up PR!!

PR #6549, opened 2026-07-26
fix(transcription): strip `***bold italic***` markdown emphasis

darrenori:
### Problem

`filter_markdown` strips markdown before text reaches TTS, but it only covers `*italic*` and `**bold**`. The combined `***text***` / `___text___` form is left untouched, so the delimiters leak through and get spoken.

```python
"This is ***very important*** text."  ->  "This is ***very important*** text."
"___Totally___ critical."             ->  "___Totally___ critical."
"Use ***both*** and **bold**"         ->  "Use ***both*** and bold"
```

Neither existing pattern can match it:

- **bold** — the body is `[^*\n]+?`, which stops at the third asterisk
- **italic** — the `(?!\s|\*)` guard rejects a `*` immediately after the opening delimiter

Same for the underscore variants. LLMs emit `***…***` fairly often for emphasis, so this shows up in practice.

### Fix

Add a triple-delimiter pattern ahead of each existing pair. The word/asterisk boundary assertions (`(?<![\w*])` / `(?![\w*])`) are carried over unchanged from the existing patterns.

### Tests

Added `test_bold_italic`, covering the four cases above across chunk sizes `1, 2, 3, 7, 50` so the streaming path is exercised too. All fail before the change, pass after.

Existing coverage is unaffected — the full `MARKDOWN_INPUT` document test, the punctuation-boundary cases, and the no-false-positive cases (`2 * 3 = 6`, `Use *.py files`, `x**2 + y**2 = z**2`, `__dunder_method__`) all still pass.

`ruff check` and `ruff format` are clean.

### DIFF
--- a/livekit-agents/livekit/agents/voice/transcription/filters.py
+++ b/livekit-agents/livekit/agents/voice/transcription/filters.py
@@ -15,11 +15,16 @@ INLINE_PATTERNS = [
     (re.compile(r"!\[([^\]]*)\]\([^)]*\)"), r"\1"),
     # links: keep text part [text](url) -> text
     (re.compile(r"\[([^\]]*)\]\([^)]*\)"), r"\1"),
+    # bold italic: remove asterisks from ***text*** (must run before the **/*
+    # patterns, which cannot match across the extra delimiter)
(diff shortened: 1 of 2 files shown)


longcw (approved):
thanks for the pr! could you sign the CLA before we merge

PR #6550, opened 2026-07-26
fix(interruption): keep an open overlap when a second agent speech segment starts

### LINKED ISSUE #6548: Adaptive interruption: a second agent speech segment (e.g. after a tool call) disarms an open overlap and silently swallows the interruption
### Summary

A second agent speech segment inside a single agent turn disarms an adaptive-interruption overlap that the user is still speaking through. From that moment the user's audio stops being sent to the bargein gateway, no verdict is ever emitted, and the interruption is silently dropped — the agent talks straight through the user.

The most common trigger is a **tool call**: a reply that produces tool calls deliberately skips `_on_end_of_agent_speech`, and the follow-up tool-response reply then raises `_AgentSpeechStartedSentinel` a second time with no end sentinel in between.

### Mechanism

`_reset_state()` clears `_overlap_started` unconditionally for *both* speech sentinels:

https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/inference/interruption.py#L588-L592

```python
case _AgentSpeechStartedSentinel() | _AgentSpeechEndedSentinel():
    await _reset_state()
    self._agent_speech_started = isinstance(input_frame, _AgentSpeechStartedSentinel)
    continue
```

`_reset_state()` (`interruption.py:573-582`) sets `_overlap_started = False`, resets the audio buffer, clears the cache and zeroes `_num_requests`.

The consequences, in order:

1. **User audio stops reaching the gateway.** The batch send is gated on `_overlap_started` (`interruption.py:649`):
   ```python
   case rtc.AudioFrame() if self._agent_speech_started:
       samples_written = self._audio_buffer.push_frame(input_frame)
       self._accumulated_samples += samples_written
       if self._accumulated_samples >= self._batch_size and self._overlap_started:
           output_ch.send_nowait(self._audio_buffer.read())
   ```
   `_agent_speech_started` is back to `True`, so frames are still buffered — they are just never classified.

2. **The overlap can never be re-armed.** `_overlap_started` is only set by `_OverlapSpeechStartedSentinel` (`interruption.py:594-597`), which is only sent from `AudioRecognition._on_start_of_speech` (`audio_recognition.py:527-533`) — i.e. a *fresh* VAD/STT speech onset. VAD does not re-announce speech that is already under way, so nothing re-opens the overlap for a user who was mid-sentence.

3. **No verdict is emitted at all.** At `_OverlapSpeechEndedSentinel` the whole body is behind `if self._overlap_started` (`interruption.py:621-640`), so `OverlappingSpeechEvent` is never sent and `AgentActivity.on_interruption` never runs.

4. **There is no VAD fallback during the gap** — see the trigger below: `_restore_interruption_by_audio_activity()` sits in the same branch that gets skipped, so VAD interruption stays disabled by the earlier `_disable_vad_interruption_soon()` (`agent_activity.py:3185-3186`).

### Trigger condition: a multi-segment agent turn

`_on_start_of_agent_speech` raises the sentinel unconditionally — there is no "already speaking" guard (`audio_recognition.py:464-465`):

```python
if self._adaptive_interruption_active:
    self._interruption_ch.send_nowait(_AgentSpeechStartedSentinel())
```

And a reply that produced tool calls **skips the end sentinel entirely** (`agent_activity.py:3333-3342`):

```python
if not speech_handle.interrupted and len(tool_output.output) > 0:
    if self._audio_recognition:
        self._audio_recognition._on_end_of_agent_speech(...)
    if self.interruption_enabled:
        self._restore_interruption_by_audio_activity()
```

Because `_on_end_of_agent_speech` is skipped, `AudioRecognition._agent_speaking` also stays `True` (it is only cleared inside that method, `audio_recognition.py:474`/`507`), so the overlap legitimately stays open across the tool call.

Then, after the tools finish, a **new** `_pipeline_reply_task` is created for the **same** `SpeechHandle` (`agent_activity.py:3454-3478`), and its first-audio-frame callback calls `_on_start_of_agent_speech` again (`agent_activity.py:3183-3184`) — the second start sentinel, with no end sentinel in between.

Concrete sequence:

1. Reply segment 1 starts speaking → `_AgentSpeechStartedSentinel`, `_agent_speech_started = True`.
2. User starts talking over it → VAD onset → `_OverlapSpeechStartedSentinel` → `_overlap_started = True`; audio starts flowing to the gateway.
3. Segment 1 ends having produced tool calls → state `"thinking"`, **no** `_AgentSpeechEndedSentinel`. The overlap is still open and the user is still talking.
4. Tools execute; the tool-response reply's first audio frame fires `_on_start_of_agent_speech` → **second** `_AgentSpeechStartedSentinel` → `_reset_state()` → `_overlap_started = False`. ← **bug**
5. The user is still mid-sentence. Nothing more reaches the classifier, and nothing can re-arm the overlap.
6. User stops → `_OverlapSpeechEndedSentinel` → `_overlap_started` is `False` → no `OverlappingSpeechEvent`, no `on_interruption`. The interrupt is silently gone.

The window is exactly the case adaptive interruption exists to handle: the user has begun interrupting but the classifier has not yet returned a verdict. (If a bargein *had* already been detected, `speech_handle.interrupted` would be `True` and line 3333 would take the `elif`, sending the end sentinel normally — so the bug specifically eats interruptions that are still being evaluated.)

The same shape exists in `_tts_task_impl` (`agent_activity.py:2722` start / `2839-2844` end, guarded by `agent_state == "speaking"`), so a queued `say()` in one turn can do it too.

### Suggested fix

Skip the reset when the stream already believes the agent is speaking *and* an overlap is open — i.e. treat the sentinel as a continuation of the same turn rather than a new one. Something like:

```python
case _AgentSpeechStartedSentinel() if self._agent_speech_started and self._overlap_started:
    # a later segment of the same agent turn (tool reply, queued say()):
    # resetting here would strand an overlap the user is still in the middle of
    continue
case _AgentSpeechStartedSentinel() | _AgentSpeechEndedSentinel():
    await _reset_state()
    ...
```

The guard has to be narrow: a genuine new turn must still wipe the overlap, the cache and the counters.

### Reference implementation (JS) — currently a deliberate divergence

The Node port has the same bug and it is fixed in https://github.com/livekit/agents-js/pull/2117, which adds exactly this guard in `agents/src/inference/interruption/interruption_stream.ts` plus regression tests that drive the pipeline through a second speech segment mid-overlap and assert user audio keeps reaching the (mocked) gateway and that a bargein verdict still surfaces.

**That PR is knowingly a divergence from Python and should probably be ported back here.** (The other half of that JS PR — overlap state lost across a transport retry — does *not* apply to Python: the flags live on `self` and `_run()` reconnects in place on the same stream object, so Python never loses them.)

### Verification notes

This was verified by reading the current code at `9accd2a2` (`main`) and tracing the call graph; I did not build a Python runtime reproduction. The JS equivalent of steps 1-6 above *is* covered by a passing/failing regression test in the linked PR, which is what established the mechanism.

biztex:
﻿Fixes #6548

## Problem

One agent turn can span several speech segments: a reply that produces tool calls deliberately skips `_on_end_of_agent_speech`, so the follow-up tool-response reply (or a queued `say()`) raises `_AgentSpeechStartedSentinel` a **second** time with no end sentinel in between. `_forward_data` treated every start sentinel as a brand-new turn and called `_reset_state()`, clearing `_overlap_started` while the user was mid-interruption.

From that moment the interruption is silently swallowed:

- `_overlap_started` gates whether user audio is forwarded to the bargein gateway at all, so classification stops;
- the flag can only be raised again by `_OverlapSpeechStartedSentinel`, which only comes from a fresh VAD speech onset — and VAD does not re-announce speech that is already under way;
- at `_OverlapSpeechEndedSentinel` the event emission is behind `if self._overlap_started`, so no `OverlappingSpeechEvent` is emitted and `AgentActivity.on_interruption` never runs. The agent talks straight through the user.

## Fix

Skip the reset when the stream already believes the agent is speaking **and** an overlap is open — i.e. treat the sentinel as a continuation of the same turn. The guard is deliberately narrow: a genuine new turn (an end sentinel was seen, so `_agent_speech_started` is `False`) still wipes the overlap, the audio buffer, the cache and the counters.

This ports the stream-side guard from the JS fix (livekit/agents-js#2117). The other half of that PR — restoring overlap state on a transport-failover replacement stream — does **not** apply to Python: these flags live on `self` and `_run()` reconnects in place on the same stream object.

## Tests

New `tests/test_interruption/test_interruption_overlap_state.py` (unit, hermetic — fake gateway WS that answers every batch), ported from the JS PR's regression suite:

- `test_second_speech_segment_keeps_open_overlap` — user audio keeps reaching the gateway after a mid-overlap second segment, and the end-of-overlap event still surfaces with `num_requests > 0` *(fails on `main`)*
- `test_bargein_verdict_after_second_speech_segment` — a `bargein_detected` response after the second segment still produces `is_interruption=True` *(fails on `main`)*
- `test_new_agent_turn_still_resets_overlap_state` — a genuine new turn still resets: audio stays gated until a fresh overlap opens *(passes on `main` and with the fix — pins the guard's narrowness)*

Verification: full `--unit` suite passes (1274 passed), `ruff format --check` / `ruff check` clean, `mypy` strict clean.

### DIFF
--- a/livekit-agents/livekit/agents/inference/interruption.py
+++ b/livekit-agents/livekit/agents/inference/interruption.py
@@ -585,6 +585,18 @@ class InterruptionStreamBase(ABC):
             match input_frame:
                 case _FlushSentinel():
                     continue
+                case _AgentSpeechStartedSentinel() if (
+                    self._agent_speech_started and self._overlap_started
+                ):
+                    # One agent turn can span several speech segments — the reply that
+                    # follows a tool call, or a queued say() — and the later segments
+                    # arrive here with no end sentinel in between. Resetting on those
+                    # would strand an overlap the user is still in the middle of:
+                    # _overlap_started gates whether their audio reaches the gateway at
+                    # all, and only a fresh VAD speech onset — which never comes for
+                    # speech already under way — can raise it again.
+                    logger.trace("agent speech continued into a new segment, keeping open overlap")
+                    continue
                 case _AgentSpeechStartedSentinel() | _AgentSpeechEndedSentinel():
                     await _reset_state()
                     self._agent_speech_started = isinstance(
--- a/tests/test_interruption/test_interruption_overlap_state.py
+++ b/tests/test_interruption/test_interruption_overlap_state.py
@@ -1,0 +1,251 @@
+"""Regression tests: overlap state must survive events that are not a new agent turn.
+
+``_overlap_started`` is the gate that lets user audio reach the bargein gateway at all,
+and the only thing that ever raises it is an ``_OverlapSpeechStartedSentinel``, which in
+turn only comes from a VAD start-of-speech. VAD does not re-announce speech that is
+already under way, so once the flag is cleared mid-overlap nothing can re-arm it: every
+remaining frame of the interruption is dropped, no verdict is emitted, and the agent
+talks straight through the user.
+
+One agent turn can span several speech segments — the reply that follows a tool call, or
+a queued ``say()`` — and those later segments raise ``_AgentSpeechStartedSentinel`` again
+with no end sentinel in between. Treating that as a new turn used to reset the overlap
+the user was in the middle of (see issue #6548, ported from livekit/agents-js#2117).
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import struct
+import time
+from unittest.mock import AsyncMock
+
+import aiohttp
+import numpy as np
+import pytest
+
+from livekit import rtc
+from livekit.agents.inference.interruption import (
+    AdaptiveInterruptionDetector,
+    InterruptionWebSocketStream,
+    OverlappingSpeechEvent,
+    _AgentSpeechEndedSentinel,
+    _AgentSpeechStartedSentinel,
+    _OverlapSpeechEndedSentinel,
+    _OverlapSpeechStartedSentinel,
+)
+from livekit.agents.types import APIConnectOptions
+
+pytestmark = [pytest.mark.unit, pytest.mark.concurrent]
+
+CONN_OPTIONS = APIConnectOptions(max_retry=0, retry_interval=0.0, timeout=1.0)
+BATCH_SAMPLES = 1600  # detection_interval (0.1s) * sample_rate (16000)
+
+
+def _make_audio_frame(*, num_samples: int = BATCH_SAMPLES) -> rtc.AudioFrame:
+    return rtc.AudioFrame(
+        data=np.zeros(num_samples, dtype=np.int16).tobytes(),
+        sample_rate=16000,
+        num_channels=1,
+        samples_per_channel=num_samples,
+    )
+
+
+class _FakeGatewayWS:
+    """Stand-in for the bargein gateway: answers every audio batch promptly.
+
+    Without prompt answers the transport's own inference timeout would tear the
+    stream down mid-test. Set ``bargein_next`` to answer the next batch with a
+    ``bargein_detected`` verdict instead of ``inference_done``.
+    """
+
+    def __init__(self) -> None:
+        self.closed = False
+        self.close_code: int | None = None
+        self.sent_audio: list[bytes] = []
+        self.bargein_next = False
+        self._rx: asyncio.Queue[aiohttp.WSMessage] = asyncio.Queue()
+        self._push_json({"type": "session.created", "default_threshold": 0.5})
+
+    def _push_json(self, payload: dict) -> None:
+        self._rx.put_nowait(
+            aiohttp.WSMessage(type=aiohttp.WSMsgType.TEXT, data=json.dumps(payload), extra=None)
+        )
+
+    async def send_str(self, data: str) -> None:  # session.create / session.close
+        pass
+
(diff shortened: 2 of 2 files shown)


biztex:
**TL;DR for reviewers**

One agent turn can span several speech segments (the reply after a tool call, or a queued `say()`). The second segment's start sentinel used to hit `_reset_state()` mid-overlap, which silently dropped an in-progress user barge-in: audio stopped reaching the classifier, the flag could never re-arm (VAD does not re-announce ongoing speech), and no verdict was emitted — the agent talked straight through the user.

The change is a narrow continuation guard in `_forward_data` (mirrors the JS fix in livekit/agents-js#2117), plus three hermetic regression tests: audio keeps flowing across a second segment, a bargein verdict still surfaces afterwards, and a genuine new turn still resets everything.

**Status:** all checks green, CLA signed, no open review threads. Happy to adjust anything.

biztex:
Superseded by #6564, which fixes the root cause at the signal source (real speech boundaries around the thinking gap, restarting overlap inference on resume) and also covers the backchannel-metrics and remote-session angles. Closing in its favor.

PR #6551, opened 2026-07-26
fix(llm): keep falsy tool return values in make_function_call_output

Osamaali313:
### Summary

`make_function_call_output` (`livekit/agents/llm/utils.py`) builds the tool result sent back to the LLM with:

```python
output=str(output or ""),
```

`_is_valid_function_output` in the same file explicitly accepts `int`, `float`, `bool`, and empty collections as valid tool return values, but `output or ""` collapses every falsy value to `""`. So a tool that legitimately returns `0`, `False`, `0.0`, `[]`, or `{}` has its result replaced by an empty string before it reaches the model.

```python
@function_tool
async def items_in_cart() -> int:
    return 0   # the model receives "" instead of "0"
```

Falsy-return cases (verbatim logic slice):

| tool returns | current | fixed |
|---|---|---|
| `0` | `""` | `"0"` |
| `False` | `""` | `"False"` |
| `0.0` | `""` | `"0.0"` |
| `None` | `""` | `""` |
| `5` | `"5"` | `"5"` |

`FunctionCallOutput.output` is serialized to providers unchanged (OpenAI `tool` content, Anthropic `tool_result`, …), so the empty string reaches the model and it loses the answer.

### Fix

```python
output="" if output is None else str(output),
```

Preserves the original intent (map `None` → `""`) while correctly stringifying `0`, `False`, `0.0`, and empty collections. Non-falsy outputs are unchanged.

I verified the behavior with a standalone slice of the function; I didn't add a unit test because the change sits in the live tool-execution path and the suite needs the full agents runtime — happy to add one if you'd like a spot for it.


longcw (approved):
looks good to me!

can you also fix the same idiom in [voice/events.py](https://github.com/livekit/agents/blob/48c179329522a44a979c97f58865773badf5241f/livekit-agents/livekit/agents/voice/events.py?plain=1#L286)

Osamaali313:
Thanks! Done — pushed the same fix to `voice/events.py:286` (the stub fallback in `RunContext`), so a falsy non-`None` tool return is preserved there too.

PR #6552, opened 2026-07-27
feat(deps): relax opentelemetry constraint to >=1.39.0,<2

### LINKED ISSUE #6544: relax opentelemetry version
### Feature Type

Would make my life easier

### Feature Description

`livekit-agents` pin `opentelemetry` to `~=1.39`, if no specific incompatible, could relax it so our application could use newer opentelemetry

_No response_

### Additional Context

_No response_

biztex:
﻿Fixes #6544

## Problem

`livekit-agents` pins `opentelemetry-api`/`-sdk`/`-exporter-otlp` to `~=1.39.0` (the 1.39.x patch series). Any application that depends on `livekit-agents` therefore cannot install opentelemetry 1.40–1.44, which conflicts with instrumentation packages and other libraries that track current OTel releases.

## Why the pin exists — and why it can be relaxed now

The pin was introduced in #4162 (previously `>=1.34`) when OTel 1.39 changed the private logs-SDK surface `telemetry/traces.py` depends on: `LogData` → `ReadWriteLogRecord` and the switch to kwargs-style `Logger.emit()`. Pinning to the patch series was the defensive response.

I verified that surface against **every release in the proposed range** — `1.39.1`, `1.40.0`, `1.41.1`, `1.42.1`, `1.43.0`, `1.44.0` (latest) — with a smoke test that exercises each private import `traces.py` uses (`opentelemetry.sdk._logs`, `opentelemetry.exporter.otlp.proto.http._log_exporter`, `opentelemetry.util._decorator._agnosticcontextmanager`, `opentelemetry._logs`), instantiates a `LogRecordProcessor` with the `emit`/`on_emit` hooks, performs the kwargs-style `Logger.emit()` call from `_upload_session_report`, and imports `telemetry.traces` + `telemetry.otel_metrics`. All six versions pass. The full `--unit` suite (1274 tests) also passes with opentelemetry **1.44.0** installed, and again on the locked 1.39.1.

## Change

- `~=1.39.0` → `>=1.39.0,<2` for the three opentelemetry packages, matching the constraint style used elsewhere in this file (`livekit-api>=1.2.0,<2`, `pydantic>=2.0,<3`).
- `uv.lock`: only the `requires-dist` specifiers change — the locked versions remain 1.39.x, so the dev environment and CI resolve exactly as before. The range only widens what downstream applications may install.

If a future OTel minor breaks the private `sdk._logs` surface again, renovate (weekly, 2-week release age) plus CI will surface it; happy to tighten the bound to a tested window (e.g. `<1.45`) instead if you prefer that tradeoff.


livekit-agents/pyproject.toml
```diff
@@ -45,9 +45,9 @@ dependencies = [
     "numpy>=1.26.0",
     "pydantic>=2.0,<3",
     "nest-asyncio>=1.6.0",
-    "opentelemetry-api~=1.39.0",
-    "opentelemetry-sdk~=1.39.0",
-    "opentelemetry-exporter-otlp~=1.39.0",
+    "opentelemetry-api>=1.39.0,<2",
```
longcw: should we use `>=1.39.0,<1.45` since only 1.44 was tested?

biztex: Done in 1a8cbfbed - switched to `>=1.39.0,<1.45`.

For the record: each minor in the range (1.39.1, 1.40.0, 1.41.1, 1.42.1, 1.43.0, 1.44.0) passed a smoke test of the private surface `telemetry/traces.py` depends on (the `sdk._logs` imports, `LogRecordProcessor.emit`/`on_emit` hooks, kwargs-style `Logger.emit`), and the full unit suite ran on 1.39.1 and 1.44.0. But agreed - a tested ceiling is the safer default given the private `_logs` dependency.

biztex:
**TL;DR for reviewers**

Relaxes the `opentelemetry-*` pin from the `~=1.39.0` patch series to `>=1.39.0,<1.45` so applications depending on `livekit-agents` can use newer OTel releases. The original pin (#4162) guarded the private `sdk._logs` surface `telemetry/traces.py` uses; that surface was smoke-tested on every release in the new range (1.39.1 through 1.44.0) and the full unit suite passes on both 1.39.1 and 1.44.0. `uv.lock` still resolves to 1.39.x, so CI and the dev environment are unchanged.

**Status:** review feedback from @longcw (cap at the tested window instead of `<2`) is applied in 1a8cbfbed and the thread is resolved; checks green. When OTel 1.45 ships I can follow up with the same verification to bump the cap.

longcw (approved):

PR #6554, opened 2026-07-27
feat(bland): add Bland TTS plugin

rv-blandai:
## Summary

Adds a TTS plugin for [Bland](https://bland.ai/), backed by [`POST /v2/tts`](https://docs.bland.ai/api-v2/post/tts). The voice UUID picks the model; `expressiveness` and `stability` are optional `0.0`–`1.0` controls.

```python
from livekit.plugins import bland

tts = bland.TTS(voice_id="f04af0e5-1a80-48a9-b02d-52f30d417cfa")
```

The API key is read from `BLAND_API_KEY`.

## Tests

`tests/test_plugin_bland_tts.py` — 31 tests, including the response path end to end against a local server. Also registered in the cross-provider `SYNTHESIZE_TTS` suite.


livekit-plugins/livekit-plugins-bland/livekit/plugins/bland/tts.py
```diff
@@ -0,0 +1,185 @@
...
+                json=body,
+                timeout=aiohttp.ClientTimeout(total=30, sock_connect=self._conn_options.timeout),
+            ) as resp:
+                resp.raise_for_status()
+
+                output_emitter.initialize(
+                    request_id=utils.shortuuid(),
+                    sample_rate=SAMPLE_RATE,
+                    num_channels=NUM_CHANNELS,
+                    mime_type="audio/wav",
```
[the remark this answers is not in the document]
rv-blandai: The assumption here is wrong; `pcm_44100` on this endpoint will still be encapsulated with a WAV header. Clarifying with a comment

livekit-plugins/livekit-plugins-bland/livekit/plugins/bland/models.py
```diff
@@ -0,0 +1,3 @@
+from typing import Literal
+
+TTSModels = Literal["BTTS_V2", "BTTS_V3"]
```
tinalenguyen: it seems that this isn't imported or used anywhere, maybe we can just remove it since it's voice id based

livekit-plugins/livekit-plugins-bland/livekit/plugins/bland/tts.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class TTS(tts.TTS):
+    def __init__(
+        self,
+        *,
+        voice_id: str = DEFAULT_VOICE_ID,
+        language: str = "en",
+        consistency: NotGivenOr[int] = NOT_GIVEN,
+        boost: NotGivenOr[bool] = NOT_GIVEN,
```
tinalenguyen: small question: i see that `boost` is an `int` on the docs, would in-between values ever be allowed?

tinalenguyen (commented):
hi, thank you for the PR! i left a few comments, i also noticed that the audio seemed to be choppy often. did you see this as well?

also, was there a reason why  the `v1/speak` endpoint was used over the `v1/speak/stream` endpoint? i wonder if that could be related to the above

livekit-plugins/livekit-plugins-bland/livekit/plugins/bland/tts.py
rv-blandai: We're doing a redesign to standardize some of the API conventions, that will replace the /v1/speak endpoint with /v2/tts (as it stands now). Updated to better reflect the current /v1/speak parameters 👍

rv-blandai: sorry for the delay - spent time making a more standardized (and far more performant) /v2/ pair of endpoints.

rv-blandai: few of the benefits:
- normalized scalar values to a simpler 0-1 range.
- just one HTTP endpoint (and one WS) instead of the different streaming/non-streaming endpoints.
- opinionated about expecting voice * uuids *, and fixed inconsistencies/weird logic with wav headers and such.

theomonnom (approved):

PR #6557, opened 2026-07-27
fix: transcription_delay reported as ~0 with the default VAD

davidzhao:
### Problem

`transcription_delay` is (wall clock when the final transcript arrived) − (wall clock when the user stopped talking). The second term is the `_last_speaking_time` anchor, which a VAD sets from the local clock — correctly and unconditionally, the session's auto-loaded default VAD included.

Since 61b14fdc (#4722), the anchor fallback also fired when `_using_default_vad`, so any session on the default VAD — i.e. most sessions — had its VAD anchor **overwritten on every transcript** by the provider-derived estimate. That estimate collapses to the transcript-arrival instant whenever the provider sends no timestamps, reporting a ~0 delay.

That commit introduced `_using_default_vad` as a compatibility shim when it started auto-loading a VAD: sites that previously asked "did the user configure a VAD?" were mechanically rewritten to treat the default VAD as absent. For the anchor sites that translation is wrong — the same commit *wires* the default VAD into `AudioRecognition`, where it genuinely produces anchors.

### Change

The three duplicated anchor-fallback conditions (final, preflight, stt end-of-speech) collapse into one `use_stt_speaking_time`, computed once per event. Three identical copies are how #4722 managed to break all three in the same way.

```python
use_stt_speaking_time = (
    self._vad is None
    or self._last_speaking_time is None
    or (self._turn_detection_mode == "stt" and has_stt_end_time)
)
```

- **Any wired VAD's anchor outranks the provider estimate again**, the default VAD included. This is the fix for the reported bug.
- **The provider estimate stays the fallback** for no-VAD sessions and for segments the VAD missed (`_last_speaking_time is None`).
- **`turn_detection="stt"` prefers a real provider timestamp**, since the provider endpoints the turn in that mode. Only a real one: a transcript that simply arrives without `end_time` says nothing about the boundary, and the estimate there is nothing but `now`, which would reproduce the ~0 delay this PR is fixing.
- **The stt-mode `END_OF_SPEECH` branch always anchors**, since that event *is* the provider stating the boundary. It prefers `ev.speech_end_time` — the provider's wall-clock speech end, which `StreamAdapter` builds as `now - silence_duration - inference_duration`, the same formula behind the VAD anchor — and falls back to the word-timestamp estimate otherwise.

All three anchor paths now bound the value at `now`. `speech_end_time` was the last unbounded one; the metrics tolerate a future anchor (`_compute_end_of_turn_metrics` clamps both delays at 0) but `extra_sleep` does not — it adds `last_speaking_time - time.time()` to the endpointing delay, so a provider clock running ahead would delay the turn commit by the skew.

The stt-mode preference is the one behavior change against `main` beyond the bug fix, and it is narrow. On `main`, `_using_default_vad` is `True` for any session that omits `vad=` regardless of `turn_detection` (`agent_session.py:605`), so default-VAD stt sessions already used the provider timestamp — this preserves that. The only configuration that moves is **explicit VAD + `turn_detection="stt"`**, which now defers to a real provider timestamp where it previously kept the VAD anchor:

| config | mode | main | this PR |
|---|---|---|---|
| `vad=None` (opted out) | any | provider | provider |
| explicit VAD | `vad` | VAD | VAD |
| explicit VAD | `stt` | VAD | provider *(if timestamped)* |
| default silero VAD | `vad` | provider | **VAD** ← the fix |
| default silero VAD | `stt` | provider | provider *(if timestamped)* |

A skewed provider timestamp that clamps to `now` is indistinguishable from a good one, so stt mode can still under-report the delay. Accepted: without a separate VAD signal there is no more accurate answer available, and `test_skewed_provider_timestamp` documents it rather than leaving it silent.

Provider timestamps themselves were verified to pass through correctly for both plugins and inference (provider-relative time + `start_time_offset` + `input_started_at`, clamped to `now`) — no changes needed there.

### Tests

`tests/test_transcription_delay_anchor.py`, parametrized over both turn detection modes so the shared rules are asserted under each:

- the default-VAD anchor survives a timestamp-less transcript (with the composed `transcription_delay` value asserted) and a skewed provider timestamp;
- a usable provider timestamp replaces the anchor in stt mode only;
- the provider estimate still applies without a VAD, and when the VAD missed the segment;
- preflight transcripts follow the same rules;
- the stt-mode `END_OF_SPEECH` branch prefers `speech_end_time` over conflicting word timestamps, clamps a future `speech_end_time` to `now`, anchors on a real word timestamp when there is no `speech_end_time`, and still anchors on arrival time when the event carries no alternatives at all (the common shape) — that last one paired against the untimestamped-transcript test, so the distinction between "the provider said now" and "the provider said nothing" is pinned by tests.

Mutation-checked in every direction: dropping `has_stt_end_time` fails 3 tests, dropping the stt clause entirely fails 4, ignoring `ev.speech_end_time` fails 1, and dropping the `min(..., now)` clamp fails 1.

### Also in this PR

- **`speaking=` on the transcript hooks** (`731cd69e`): the same `_using_default_vad` over-application gated the speaking state passed to `on_final_transcript` / `on_interim_transcript` — `speaking is False` is what arms the false-interruption resume timer, so default-VAD sessions never auto-resumed a transcript-triggered pause. `self._vad` is the resolved, wired VAD (default included) and drives `_speaking` either way, so the state is now reported whenever a VAD is wired. `AudioRecognition` no longer reads VAD provenance at all (the `using_default_vad` parameter is removed); the flag stays in `AgentActivity`, where its four remaining uses arbitrate real policy.
- **audio_eot suite repair + CI** (`b0b308ae`): fixed the 8 rotted tests in `test_audio_recognition_turn_detection.py`, and `make unit-tests` now runs `--unit --audio_eot`, putting the suite in the unit-test CI gate.


livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -493,20 +492,16 @@ async def stt_node(
...
             conn_options = activity.session.conn_options.stt_conn_options
             async with wrapped_stt.stream(conn_options=conn_options) as stream:
-                _audio_input_started_at: float = (
-                    activity._audio_recognition._input_started_at
+                # provider timestamps are relative to this stream's first frame; the
+                # offset places them on the input's audio timeline, which is how far
+                # that timeline had already advanced when the stream was created.
+                # Measured in audio delivered, not wall clock: the two only agree
+                # while audio flows at exactly realtime.
+                stream.start_time_offset = (
```
longcw: here changed the `start_time_offset` to a relative timestamp to the input audio starts, but in https://github.com/livekit/agents/blob/48c179329522a44a979c97f58865773badf5241f/livekit-agents/livekit/agents/stt/stt.py?plain=1#L463 and https://github.com/livekit/agents/blob/48c179329522a44a979c97f58865773badf5241f/livekit-agents/livekit/agents/stt/fallback_adapter.py?plain=1#L362 it's still wall clock time.

livekit-agents/livekit/agents/voice/audio_recognition.py
longcw: should we remove `_using_default_vad` here or add a comment on why it keeps here but removed for `_last_speaking_time`?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1208,8 +1205,12 @@ def _process_stt_event(self, ev: stt.SpeechEvent) -> None:
...
 
-            if self._vad is None or self._using_default_vad or self._last_speaking_time is None:
-                # vad disabled or missed a speech, use stt timestamp
+            if self._vad is None or self._last_speaking_time is None:
+                # no vad, or vad missed this speech: fall back to the provider's
+                # estimate. Any vad anchor — the session's default vad included —
+                # is measured against the local clock and outranks it; the
+                # estimate collapses to `now` when the provider sends no
+                # timestamps, which reports a ~0 transcription_delay
                 self._last_speaking_time = stt_last_speaking_time
```
chenghao-mou: This also means we change how endpointing works for `turn_detection='stt'` + implicit vad in a turn:

1. vad speech (`_last_speaking_time` updated)
2. vad false negative + STT new SOS + final (`_last_speaking_time` anchored to the SOS)
   1. STT EOS no longer update `_last_speaking_time` 
3. turn commit uses the  `_last_speaking_time` from STT SOS to choose delay while before, it chooses the value from STT EOS.

chenghao-mou (approved):

chenghao-mou (approved):
lgtm

PR #6561, opened 2026-07-27
bithuman: skip the SDK dependency on Python 3.14

longcw:
Gate the bithuman SDK behind a `python_version < '3.14'` marker since it publishes no build for 3.14 and breaks `uv sync` there, and print the reason at both import sites so the resulting failure isn't just a bare `ModuleNotFoundError`.


chenghao-mou:
I reached out to them about missing packages last week and they said they are going to publish new ones. We should also gate it > 3.10 as well. Both 3.10 and 3.14 releases are missing.

davidzhao (approved):

PR #6563, opened 2026-07-27
fix: preserve adaptive interruption across tool calls

chenghao-mou:
**Bug:** Adaptive interruption could silently drop a user overlap when agent playout resumed after a tool call because the interruption stream received mismatched speech boundaries.

Treat thinking gaps as real speech boundaries and restart overlap inference when playout resumes mid-utterance. Agent-ended overlaps remain inconclusive and no longer count as backchannels; remote forwarding is noted until the protocol carries that marker.

Fixes AGT-3180.


chenghao-mou:
Superseded by #6564, which contains the same commit on the correctly named branch.

PR #6564, opened 2026-07-27
fix: preserve adaptive interruption across tool calls

### LINKED ISSUE #6548: Adaptive interruption: a second agent speech segment (e.g. after a tool call) disarms an open overlap and silently swallows the interruption

chenghao-mou:
**Bug:** Adaptive interruption could silently drop a user overlap when agent playout resumed after a tool call because the interruption stream received mismatched speech boundaries.

Treat thinking gaps as real speech boundaries and restart overlap inference when playout resumes mid-utterance. Agent-ended overlaps remain inconclusive and no longer count as backchannels; remote forwarding is noted until the protocol carries that marker.

Fixes livekit/agents#6548. Fixes livekit/agents#6548.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2067,8 +2067,7 @@ class AgentActivity(RecognitionHooks):
                 self._session._update_agent_state("listening")
                 if self._audio_recognition:
                     self._audio_recognition._on_end_of_agent_speech(
-                        ignore_user_transcript_until=ignore_user_transcript_until or time.time(),
-                        paused=True,
+                        ignore_user_transcript_until=ignore_user_transcript_until or time.time()
                     )
                 if self.interruption_enabled:
                     self._restore_interruption_by_audio_activity()
@@ -2192,8 +2191,7 @@ class AgentActivity(RecognitionHooks):
         # flush held transcripts again if possible
         if self._audio_recognition:
             self._audio_recognition._on_end_of_agent_speech(
-                ignore_user_transcript_until=ev.overlap_started_at or ev.detected_at,
-                paused=self._paused_speech is not None,
+                ignore_user_transcript_until=ev.overlap_started_at or ev.detected_at
             )
 
     def on_interim_transcript(self, ev: stt.SpeechEvent, *, speaking: bool | None) -> None:
@@ -3468,7 +3466,7 @@ class AgentActivity(RecognitionHooks):
             self._session._update_agent_state("thinking")
             if self._audio_recognition:
                 self._audio_recognition._on_end_of_agent_speech(
-                    ignore_user_transcript_until=time.time(), paused=True
+                    ignore_user_transcript_until=time.time()
                 )
             if self.interruption_enabled:
                 self._restore_interruption_by_audio_activity()
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -4386,9 +4386,7 @@ class AgentActivity(RecognitionHooks):
                     otel_context=self._paused_speech.handle._agent_turn_context,
                 )
                 if self._audio_recognition and self._paused_speech.agent_state == "speaking":
-                    self._audio_recognition._on_start_of_agent_speech(
-                        started_at=time.time(), resumed=True
-                    )
+                    self._audio_recognition._on_start_of_agent_speech(started_at=time.time())
                 if self.interruption_enabled:
                     self._disable_vad_interruption_soon()
                 audio_output.resume()
@@ -4458,13 +4456,6 @@ class AgentActivity(RecognitionHooks):
 
         if not self._paused_speech:
             return
-
-        # the pause withheld end-of-agent-speech for a resume; interrupting ends the turn
-        # instead. the audio stopped when it was paused, so no playout is left to wait for
-        if interrupt and self._audio_recognition:
-            self._audio_recognition._on_end_of_agent_speech(
-                ignore_user_transcript_until=time.time()
-            )
 
         if (
             interrupt
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2608,8 +2608,9 @@ class AgentActivity(RecognitionHooks):
 
     def _on_pipeline_reply_done(self, _: asyncio.Task[None]) -> None:
         if not self._speech_q and (not self._current_speech or self._current_speech.done()):
+            was_speaking = self._session.agent_state == "speaking"
             self._session._update_agent_state("listening")
-            if self._audio_recognition:
+            if self._audio_recognition and was_speaking:
                 self._audio_recognition._on_end_of_agent_speech(
                     ignore_user_transcript_until=time.time()
                 )
@@ -3331,7 +3332,14 @@ class AgentActivity(RecognitionHooks):
             current_span.set_attribute(trace_types.ATTR_RESPONSE_TEXT, forwarded_text)
 
         if not speech_handle.interrupted and len(tool_output.output) > 0:
+            was_speaking = self._session.agent_state == "speaking"
             self._session._update_agent_state("thinking")
+            if self._audio_recognition and was_speaking:
+                self._audio_recognition._on_end_of_agent_speech(
+                    ignore_user_transcript_until=time.time()
+                )
+            if self.interruption_enabled and was_speaking:
+                self._restore_interruption_by_audio_activity()
         elif self._session.agent_state == "speaking":
             self._session._update_agent_state("listening")
             if self._audio_recognition:
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -3466,6 +3466,12 @@ class AgentActivity(RecognitionHooks):
 
         if not speech_handle.interrupted and len(tool_output.output) > 0:
             self._session._update_agent_state("thinking")
+            if self._audio_recognition:
+                self._audio_recognition._on_end_of_agent_speech(
+                    ignore_user_transcript_until=time.time(), paused=True
+                )
+            if self.interruption_enabled:
+                self._restore_interruption_by_audio_activity()
         elif self._session.agent_state == "speaking":
             self._session._update_agent_state("listening")
             if self._audio_recognition:
--- a/livekit-agents/livekit/agents/voice/remote_session.py
+++ b/livekit-agents/livekit/agents/voice/remote_session.py
@@ -696,6 +696,7 @@ class SessionHost:
             overlap_started_at = Timestamp()
             overlap_started_at.FromNanoseconds(int(event.overlap_started_at * 1e9))
 
+        # TODO(AGT-3180): Forward agent_ended when the remote-session protocol supports it.
         pb = agent_pb.AgentSessionEvent.OverlappingSpeech(
             is_interruption=event.is_interruption,
             detection_delay=event.detection_delay,
--- a/livekit-agents/livekit/agents/voice/remote_session.py
+++ b/livekit-agents/livekit/agents/voice/remote_session.py
@@ -595,6 +595,7 @@ class SessionHost:
             overlap_started_at = Timestamp()
             overlap_started_at.FromNanoseconds(int(event.overlap_started_at * 1e9))
 
+        # TODO(AGT-3180): Forward agent_ended when the remote-session protocol supports it.
         pb = agent_pb.AgentSessionEvent.OverlappingSpeech(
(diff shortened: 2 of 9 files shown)


biztex:
This looks like the right fix — treating the thinking gap as a real speech boundary and restarting overlap inference when playout resumes addresses the root cause at the signal source, where my #6550 only made the stream tolerate the redundant start sentinel. The metrics and remote-session coverage go beyond what mine attempted, too.

Closing #6550 in favor of this. If the hermetic stream-level regression tests from it would be useful here (a fake bargein gateway driving `InterruptionWebSocketStream` through a mid-overlap second segment), happy to adapt them to the new sentinel semantics — just say the word.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2608,8 +2608,9 @@ def _to_instructions(v: Instructions | str) -> Instructions:
 
     def _on_pipeline_reply_done(self, _: asyncio.Task[None]) -> None:
         if not self._speech_q and (not self._current_speech or self._current_speech.done()):
+            was_speaking = self._session.agent_state == "speaking"
```
longcw: question: what is the difference between `session.agent_state` and `self._agent_speaking`?

biztex: (Chiming in from #6550, since this distinction was the heart of #6548 — hope it helps.)

They live at different layers and deliberately diverge mid-turn:

- `session.agent_state` is the session-level lifecycle (`listening` / `thinking` / `speaking`), updated via `AgentActivity._update_agent_state()`. A reply that produces tool calls flips it to `"thinking"` between segments; it is what applications observe through `agent_state_changed`.

- `AudioRecognition._agent_speaking` is a private recognition-side flag meaning "an agent playout window is open for interruption/endpointing purposes". It is set by `_on_start_of_agent_speech()` (first audio frame of playout) and cleared only inside `_on_end_of_agent_speech()` — which is intentionally *not* called when the reply produced tool calls, so it stays `True` straight through the thinking gap. It gates overlap tracking in `_on_start_of_speech()`, the transcript-ignore window, and `_should_ignore_input_audio()`.

So during a tool-call gap the two disagree by design: `agent_state == "thinking"` while `_agent_speaking == True`. `was_speaking` as written captures "playout was audibly active at reply-done" (session view) rather than "the recognition-side speech window is still open" — which one is wanted here is of course the author's call.

longcw:
from claude code:

#6662 section (b) is the same bug with a different trigger: a second agent-speech boundary inside one turn resets the detector stream while an overlap is open. It corrected the false-interruption pause. It did not touch the tool branch:

https://github.com/livekit/agents/blob/59c2fae92ae2180e6d77bcd610c9b7a6dbb123f9/livekit-agents/livekit/agents/voice/agent_activity.py#L3443-L3452

The tool-response reply is a new `_pipeline_reply_task` on the same `speech_handle`, so its playout start reaches the same call site as a first segment, where `resumed` defaults to `False`. Second start sentinel, `_reset_state()`, verdict gone. `_restore_interruption_by_audio_activity()` is in the skipped `elif`, so there is no VAD fallback either.

Two edits instead:

1. Call `_on_end_of_agent_speech(..., paused=True)` and `_restore_interruption_by_audio_activity()` in the tool branch. A pause keeps `_overlap_open` and sends no reset, so user audio keeps reaching the gateway while the tools run.
2. Add `resumed = resumed or self._overlap_open` in `_on_start_of_agent_speech`. `_overlap_open` is set only while an overlap is unresolved, so a new turn still resets. This covers the queued `say()` case too.

Then the synthesized `_on_start_of_speech`, the `_agent_speaking` guard, both `was_speaking` gates and the `self._speaking and` condition can all go. Three of them have side effects: the synthesized onset overwrites `_utterance_started_at` with the resume time, the guard lets `_turn_tracker` reset mid-utterance, and `speech_duration=0.0` with `_overlap_count = 0` shifts the buffer down to `_prefix_size`.

Keep the sentinel reorder. It fixes a separate bug, and it is real: the reset currently goes out before the `agent_ended=True` sentinel, so the stream drops that event and `audio_recognition.py:502` never emits. The `num_backchannels` change needs the reorder to reach the metrics at all.

livekit-agents/livekit/agents/voice/remote_session.py
```diff
@@ -599,6 +599,7 @@ def _on_overlapping_speech(self, event: OverlappingSpeechEvent) -> None:
             overlap_started_at = Timestamp()
             overlap_started_at.FromNanoseconds(int(event.overlap_started_at * 1e9))
 
+        # TODO(AGT-3180): Forward agent_ended when the remote-session protocol supports it.
```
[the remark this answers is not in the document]
biztex: Verified this against the branch head (217d25ffb) — the finding is real, and it is newly reachable behavior rather than a pre-existing gap:

- Before this PR, `_on_end_of_agent_speech` pushed `_AgentSpeechEndedSentinel` *before* `_on_end_of_overlap_speech`, so `_reset_state()` had already cleared `_overlap_started` and the detector's `_OverlapSpeechEndedSentinel` arm produced no event. With the reordering (`audio_recognition.py:509-514`), the detector now emits `OverlappingSpeechEvent(is_interruption=False, agent_ended=True)` (`interruption.py:621-641`) — an event that simply never fired before.
- `AgentActivity._on_overlap_speech_ended` forwards every event to the session unconditionally (`agent_activity.py:1887-1889`), so the remote forwarder sees them all.
- Both local consumers were taught the distinction (`interruption.py:670` excludes `agent_ended` from `num_backchannels`; `audio_recognition.py:1480` skips the backchannel latch), but the wire message cannot express it: `OverlappingSpeech` in livekit/protocol (`protobufs/agent/livekit_agent_session.proto`) carries only `is_interruption` / `overlap_started_at` / `detection_delay` / `detected_at`. A remote consumer therefore has no choice but to read these as confirmed backchannels.

Since the proto change lives in livekit/protocol (an `optional bool agent_ended = 5` would be backward-compatible there), the self-contained interim fix is to skip forwarding inconclusive verdicts, in line with the existing TODO:

```suggestion
        # TODO(AGT-3180): Forward agent_ended when the remote-session protocol supports it.
        if event.agent_ended:
            # an inconclusive verdict is indistinguishable from a confirmed backchannel
            # on the wire; skip it until the protocol can carry the marker
            return
```

Suppressing the event entirely matches pre-PR remote behavior (these overlaps produced no event at all), so nothing downstream regresses in the meantime, and the TODO keeps the protocol follow-up visible.

davidzhao (approved):

theomonnom (approved):

PR #6566, opened 2026-07-27
fix(evals): avoid overriding judge reasoning

inickt:
## Summary

- Stop the generic eval judge from injecting the Chat Completions-specific `reasoning_effort="none"` option.
- Preserve reasoning options configured on the supplied LLM.
- Add regression coverage for the options forwarded by the judge.

## Why

The judge accepts any LLM, but `reasoning_effort` is specific to Chat Completions. Passing it to an OpenAI Responses client fails because Responses uses the `reasoning` object instead.

Explicitly configured reasoning options continue to apply. LLMs without an explicit configuration will now use their provider or model default rather than having the judge force reasoning off.

## Testing

- Added focused unit coverage.
- Ran the full formatting, linting, and type-check suite.


tinalenguyen (commented):
thank you for the PR, this change makes sense. i think maybe we should set the default reasoning for inference GPT 5 models to `none` before applying this change so the judges don't suddenly start with medium effort. i can make that PR

tinalenguyen (approved):

PR #6571, opened 2026-07-27
(expressive mode): register inference replaces presets

cshape:
stacked on #6528 — the first two commits are that PR; review the last three.

our evals (~2,000 generations, 5 personas × serious/neutral/casual moments, 4 llms) suggest the agent persona and conversational context choose the appropriate register on their own, so this replaces the formal/casual presets with:

- a provider-neutral register rule in the shared preamble (composed when the moment is heavy, loose when it's casual, reassessed every turn), plus fish-specific guidance built conditionally so an opted-out option is never mentioned
- preset removal — on fish the two presets resolved to byte-identical instructions, and neither moved behavior the persona didn't already move
- `nonverbal_sounds` accepts a bool, and dicts are sparse opt-outs: `{"laughing": False}` removes laughter and nothing else

measured at the default (everything on): 0.00 laughs/turn at serious moments across all four llms, 0.67–0.83 at casual ones, judge fit 4.59/5 overall. `{"nonverbal_sounds": False, "disfluencies": False}` produced 0 sounds and 1 filler across 360 generations.

one semantic change to sign off on: `NonverbalOptions()` (empty dict) now means "no opt-outs" — everything stays on; `nonverbal_sounds=False` is the all-off form.

full findings + reproducible eval harness shared separately.


cshape:
rebased onto main as #6632

PR #6574, opened 2026-07-27
fix(examples): pass LemonSlice avatar options via extra_kwargs

shawnfeldman:
## Problem

`examples/avatar/inference_agent.py` passes `image_url` and `prompt` directly to `inference.AvatarSession`, but that class has no such parameters — provider options go through `extra_kwargs` (typed as `LemonSliceOptions`), exactly as the [class docstring](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/inference/avatar.py#L107-L114) shows.

Running the example fails immediately:

```
TypeError: AvatarSession.__init__() got an unexpected keyword argument 'image_url'. Did you mean 'base_url'?
```

The example and the class landed in the same commit (538884e65, #6492), so the example was never executed.

## Fix

```python
avatar = inference.AvatarSession(
    "lemonslice",
    extra_kwargs=inference.LemonSliceOptions(
        image_url=avatar_image_url,
        prompt="Be expressive in your movements and use your hands while talking.",
    ),
)
```

`LemonSliceOptions` is already exported from `livekit.agents.inference` (lazily via `__getattr__`, and listed in `__all__`), so no import change is needed.

## Verification

- Reproduced the original `TypeError`, and confirmed the new form constructs successfully with `extra_kwargs` reaching `AvatarSession._extra_kwargs` intact (both keys map onto first-class gateway payload fields via `_OPTION_TO_PAYLOAD_FIELD`).
- `examples/avatar/inference_agent.py` now imports end to end.
- `ruff format --check` and `ruff check` pass; `tests/test_inference_avatar.py --unit` passes (29 passed).
- Grepped the repo for other `inference.AvatarSession` callers — this example was the only incorrect one; `tests/test_inference_avatar.py` already uses `extra_kwargs`.

Examples-only change, so no version bump.

### DIFF
--- a/examples/avatar/README.md
+++ b/examples/avatar/README.md
@@ -9,11 +9,12 @@ Try it in the [LiveKit Playground](https://agents.livekit.io/?example=avatar).
 
 > **Inference variant:** [`inference_agent.py`](./inference_agent.py) is a
 > minimal version that provisions the avatar through **LiveKit Inference**
-> instead of the BYOK plugin — the agent needs only `LIVEKIT_API_KEY` /
-> `LIVEKIT_API_SECRET` (no `LEMONSLICE_API_KEY`); the gateway creates the
+> instead of the BYOK plugin — no `LEMONSLICE_API_KEY` is needed; the agent
+> authenticates with your LiveKit credentials and the gateway creates the
 > provider session with LiveKit's wholesale key. Requires the
 > `avatar_lemonslice` feature flag on your project. Run with
-> `python inference_agent.py dev` and set `LEMONSLICE_IMAGE_URL`.
+> `python inference_agent.py dev` and set `LIVEKIT_URL`, `LIVEKIT_API_KEY`,
+> `LIVEKIT_API_SECRET`, and `LEMONSLICE_IMAGE_URL`.
 
 ## What's in here
 
--- a/examples/avatar/inference_agent.py
+++ b/examples/avatar/inference_agent.py
@@ -1,13 +1,19 @@
 """Avatar agent provisioned through LiveKit Inference (no provider key).
 
 Unlike agent.py (which uses the BYOK lemonslice plugin with a LEMONSLICE_API_KEY),
-this starts the avatar with inference.AvatarSession: the agent authenticates only
-with LIVEKIT_API_KEY / LIVEKIT_API_SECRET, and the Inference gateway creates the
-LemonSlice session using LiveKit's wholesale key. Media and lip-sync still flow
-in-room over DataStream, exactly as the BYOK path does.
+this starts the avatar with inference.AvatarSession: no provider key is needed —
+the agent authenticates with its LiveKit credentials (LIVEKIT_API_KEY /
+LIVEKIT_API_SECRET, or LIVEKIT_INFERENCE_API_KEY / LIVEKIT_INFERENCE_API_SECRET
+when set), and the Inference gateway creates the LemonSlice session using
+LiveKit's wholesale key. Agent audio and the playback RPCs still flow in-room
+over DataStream and the avatar publishes its video as a normal track, exactly as
+the BYOK path does.
 
 Requires the `avatar_lemonslice` feature flag to be enabled for your project on
-the Inference gateway.
+the Inference gateway; without it the gateway returns HTTP 403 "Inference Avatar
+is not enabled for this project".
+
+Env: LIVEKIT_URL, LIVEKIT_API_KEY, LIVEKIT_API_SECRET, LEMONSLICE_IMAGE_URL.
 
 Run:
     python inference_agent.py dev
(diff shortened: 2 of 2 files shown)


shawnfeldman (commented):
## PR Review Summary

Reviewed the 4-line change against both the SDK (`inference/avatar.py`) and the **agent-gateway** wire contract (`livekit/agent-gateway`). **The fix is correct end-to-end — no blocking issues.** One suggestion below is directly implicated by the diff; the rest are pre-existing and out of scope.

### Gateway contract verification ✅

Traced the request the example now produces all the way to LemonSlice:

| Stage | Result |
|---|---|
| SDK `_create_session` | `image_url` and `prompt` are both in `_OPTION_TO_PAYLOAD_FIELD`, so they are promoted to **top-level payload fields** — they do *not* land in the forwarded `extra_kwargs` blob |
| POST target | `{base_url}/avatar/sessions` → `https://agent-gateway.livekit.cloud/v1/avatar/sessions`, matching the route at `pkg/deps/deps.go:87` |
| Gateway `avatarRequest` | `image_url` and `prompt` are declared first-class fields (`pkg/handler/avatar.go:59-60`) ✅ |
| `unknownExtraKwargs` allowlist | Not exercised — the emitted payload carries **no** `extra_kwargs` key at all, since both options are known ✅ |
| LemonSlice adapter `Validate` | `hasID=false, hasImage=true` → passes the XOR rule (`lemonslice.go:101-111`) ✅ |
| Adapter payload mapping | `image_url`→`agent_image_url`, `prompt`→`agent_prompt` (`lemonslice.go:259-264`) ✅ |

**This is worth calling out explicitly:** the fix works *because* both keys are in `_OPTION_TO_PAYLOAD_FIELD`. LemonSlice's `allowed_extra_kwargs` in the gateway config is `[model, aspect_ratio, simulcast, response_done_timeout]` — it deliberately does **not** include `image_url`/`prompt`. Had these keys taken the generic forwarding path instead, the gateway would have rejected the request with `400 unknown extra_kwargs keys`. The chosen form is the only one that works.

### Suggestions

**1. The catalog-agent-id hint now leads into a `ValueError`** — [`examples/avatar/inference_agent.py:51-52`]

```python
# Pass a catalog agent id instead of an image with
# inference.AvatarSession("lemonslice/<agent_id>", ...).
```

This read cleanly *before* this diff: `image_url` was a flat kwarg, so "swap the model string, drop that kwarg" was self-evident and `...` stood in for the remaining flat kwargs. Now the `...` elides an `extra_kwargs=inference.LemonSliceOptions(...)` wrapper, and the reader has to know to remove `image_url` *from inside the TypedDict* while keeping `prompt` there.

A reader who follows the hint literally — copying the call directly below and changing only the model string — hits a hard failure in `__init__` (`avatar.py:173-179`):

```
ValueError: pass either a catalog id in the model string ('lemonslice/<avatar_id>') or image_url, not both
```

The gateway rejects the same combination independently (`lemonslice.go:106`), so there is no path where it works. Suggest making the alternative concrete rather than elided, and moving it adjacent to the call:

```python
# For a pre-built LemonSlice agent, pass the catalog id in the model string
# and drop image_url from the options (the two are mutually exclusive):
#   inference.AvatarSession(
#       "lemonslice/<agent_id>",
#       extra_kwargs=inference.LemonSliceOptions(prompt="..."),
#   )
```

### Observations (pre-existing — not introduced by this PR, listed for a follow-up)

- **`LemonSliceOptions` can't express the gateway's allowlisted knobs.** The TypedDict covers `image_url`/`prompt`/`idle_prompt`/`idle_timeout`, but the gateway separately allowlists `model`, `aspect_ratio`, `simulcast`, and `response_done_timeout` for forwarding. A user who wants `model="pro"` must fall back to a plain `dict` (accepted via the `LemonSliceOptions | dict[str, Any]` union) and thereby loses TypedDict key checking for *all* the options. Worth considering adding those four to the TypedDict.
- **`inference_agent.py:6` — "Media and lip-sync still flow in-room over DataStream."** DataStream carries the outbound agent *audio* plus three playback RPCs (`lk.clear_buffer`, `lk.playback_finished`, `lk.playback_started`); the avatar's **video is a normal published WebRTC track** — `avatar.py:346` waits on `wait_remote_track=rtc.TrackKind.KIND_VIDEO` to prove it. Lip-sync is generated provider-side and does not traverse DataStream at all. The SDK's own class docstring says "Media and **RPC**" — the example substituted "lip-sync".
- **The "only `LIVEKIT_API_KEY` / `LIVEKIT_API_SECRET`" claim** (lines 4-5, restated at 47-48, and again in `examples/avatar/README.md:12-13`) overstates slightly: `LIVEKIT_URL` is also hard-required (`start()` raises without it, `avatar.py:255-263`), and `LIVEKIT_INFERENCE_API_KEY`/`_SECRET` take precedence for gateway auth when set. The substantive point — *no provider key* — is correct.
- **Lines 47-49 restate lines 3-7** almost verbatim in a 72-line file, and the two copies have already drifted ("over DataStream" vs. "media stays in-room").

### Strengths

- Matches the SDK's own canonical example verbatim (`avatar.py:109-112`), so the example and the API reference now agree.
- Uses the `LemonSliceOptions` TypedDict constructor rather than a bare dict, which buys mypy key/type checking. Other `extra_kwargs` sites in `examples/` use plain dict literals — this one is the better pattern, not the outlier to fix.
- No stale copies of the broken form left anywhere: the only other `inference.AvatarSession(...)` construction with options is the class docstring, which was already correct.
- Correctly scoped — examples-only, so no version bump, and `examples/` is outside `scripts/check_types.py`'s mypy scope so nothing else is gated on it.
- Not a leak despite no explicit `aclose()`: `BaseAvatarSession.start()` registers `job_ctx.add_shutdown_callback(self.aclose)` (`voice/avatar/_types.py:90`), so the gateway session is terminated on job shutdown.

<sub>Verified: `ruff format --check` + `ruff check` pass; `tests/test_inference_avatar.py --unit` → 29 passed; example imports end-to-end; original `TypeError` reproduced and confirmed fixed.</sub>

tinalenguyen (approved):

PR #6579, opened 2026-07-28
fix: order delayed realtime transcripts by speech start

chenghao-mou:
**Why:** Late OpenAI realtime transcripts can appear after the reply they prompted in session reports.
Correlate transcripts with speech-start timestamps for report chronology while preserving realtime server item order in the model context.
Bound correlation state and cover back-to-back delayed transcripts plus the server/report ordering split.

Fixes AGT-3186


davidzhao (approved):
nice fix!

chenghao-mou:
Closed in favor of [https://github.com/livekit/agents/pull/6616](<https://github.com/livekit/agents/pull/6616>)

PR #6582, opened 2026-07-28
fix(livekit-agents): forward prewarm to the primary LLM in the fallback adapter

### LINKED ISSUE #6572: LLM FallbackAdapter doesn't pass through prewarm to contained LLM instances
### Bug Description

Some LLM clients (i.e. google) contain logic to be prewarmed (and are triggered by [AgentSession.__init__](https://github.com/livekit/agents/blob/11989dc889a526e1ec171b61feaab650b6d1d5a2/livekit-agents/livekit/agents/voice/agent_session.py#L490)). The LLM FallbackAdapter currently swallows this. Compare to, for instance, the [TTS fallback adapter](https://github.com/livekit/agents/blob/11989dc889a526e1ec171b61feaab650b6d1d5a2/livekit-agents/livekit/agents/tts/fallback_adapter.py#L126), which passes prewarm through to the first TTS option.

### Expected Behavior

The LLM is prewarmed.

### Reproduction Steps

```bash
1. 
2.
3.
...
```

### Operating System

macOS

### Models Used

Deepgram Flux / Gemini / Eleven v2.5 flash

### Package Versions

```bash
livekit-agents==1.6.7
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
Add a LLM FallbackAdapter.prewarm that implements a similar pattern to TTS FallbackAdapter prewarm.
```

### Additional Context

_No response_

_No response_

fxhxdxd:
Fixes livekit/agents#6572.

### Problem

`LLM.prewarm()` short-circuits unless the concrete class overrides `_prewarm_impl` (`livekit-agents/livekit/agents/llm/llm.py:162-163`). `llm.FallbackAdapter` doesn't override it, so every `prewarm()` call it receives — from `AgentSession.__init__` (`voice/agent_session.py:490`) and from `AgentActivity` (`voice/agent_activity.py:636,721`) — returns immediately without ever reaching the wrapped instances. Providers that do implement prewarming are therefore never warmed when used behind the adapter, so the first reply pays the full DNS + TLS setup cost.

`tts.FallbackAdapter` already handles this correctly (`tts/fallback_adapter.py:126-128`); the LLM adapter is the gap.

`stt.FallbackAdapter` has the same gap — happy to cover it in this PR or a follow-up, whichever you prefer.

### Fix

Override `prewarm()` to delegate to the first instance, mirroring the TTS adapter. Only the primary is prewarmed: the remaining instances are not expected to serve traffic unless it fails, and warming them would open connections to providers that may never be used.

The `loop` argument is forwarded, which is the one place this differs from the TTS version — `LLM.prewarm()` accepts `loop` (the TTS and STT variants don't) and `AgentSession` passes the session loop explicitly, so dropping it would schedule the prewarm task on the wrong loop.

### Tests

New `tests/test_llm_fallback.py`:

* `test_prewarm_forwarded_to_primary_llm` — the primary's `_prewarm_impl` runs and the fallback instance stays cold.
* `test_prewarm_forwards_event_loop` — the prewarm task is scheduled on the loop supplied by the caller.

Both fail on `main` and pass with this change. `ruff check`, `ruff format --check` and `mypy --strict` are clean on the touched files, and `test_llm_fallback.py`, `test_tts_fallback.py` and `test_stt_fallback.py` pass together (15 passed).

### DIFF
--- a/tests/test_llm_fallback.py
+++ b/tests/test_llm_fallback.py
@@ -1,0 +1,58 @@
+from __future__ import annotations
+
+import asyncio
+
+import pytest
+
+from livekit.agents.llm import FallbackAdapter
+
+from .fake_llm import FakeLLM
+
+pytestmark = [pytest.mark.unit]
+
+
+class PrewarmableLLM(FakeLLM):
+    """FakeLLM that opts into prewarming by overriding ``_prewarm_impl``."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.prewarmed = asyncio.Event()
+
+    async def _prewarm_impl(self) -> None:
+        self.prewarmed.set()
+
+
+async def test_prewarm_forwarded_to_primary_llm() -> None:
+    primary = PrewarmableLLM()
+    fallback = PrewarmableLLM()
+
+    fallback_adapter = FallbackAdapter([primary, fallback])
+    try:
+        fallback_adapter.prewarm()
+
+        await asyncio.wait_for(primary.prewarmed.wait(), timeout=5)
+        assert not fallback.prewarmed.is_set(), (
+            "expected only the primary LLM to be prewarmed, the fallbacks should stay cold"
+        )
+    finally:
+        await fallback_adapter.aclose()
+        await primary.aclose()
+        await fallback.aclose()
+
+
+async def test_prewarm_forwards_event_loop() -> None:
+    primary = PrewarmableLLM()
+
+    fallback_adapter = FallbackAdapter([primary])
+    try:
+        loop = asyncio.get_running_loop()
+        fallback_adapter.prewarm(loop=loop)
+
+        assert primary._prewarm_task is not None, "expected the primary LLM to be prewarmed"
+        assert primary._prewarm_task.get_loop() is loop, (
+            "expected the prewarm task to be scheduled on the provided event loop"
+        )
+        await asyncio.wait_for(primary.prewarmed.wait(), timeout=5)
+    finally:
+        await fallback_adapter.aclose()
+        await primary.aclose()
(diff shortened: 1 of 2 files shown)


tests/test_llm_fallback.py
```diff
@@ -0,0 +1,58 @@
...
+        await fallback.aclose()
+
+
+async def test_prewarm_forwards_event_loop() -> None:
+    primary = PrewarmableLLM()
+
+    fallback_adapter = FallbackAdapter([primary])
+    try:
+        loop = asyncio.get_running_loop()
+        fallback_adapter.prewarm(loop=loop)
```
chenghao-mou: nit: even if we drop `loop=loop`, it will pass as it defaults to the running loop. We can test it with a new loop instead: 

```
class RecordingLLM(FakeLLM):
    def __init__(self) -> None:
        super().__init__()
        self.prewarm_loop: asyncio.AbstractEventLoop | None = None

    def prewarm(self, *, loop: asyncio.AbstractEventLoop | None = None) -> None:
        self.prewarm_loop = loop

...
supplied_loop = asyncio.new_event_loop()
try:
    fallback_adapter.prewarm(loop=supplied_loop)
    assert primary.prewarm_loop is supplied_loop
finally:
    supplied_loop.close()
```

chenghao-mou (approved):
lgtm. One small test nit.

tests/test_llm_fallback.py
fxhxdxd: Good catch, you're right that the original assertion held either way, since the wrapped LLM defaults to the running loop when `loop` is omitted.

Switched to recording the loop the primary is asked to prewarm on and supplying one distinct from the running loop, per your suggestion. Verified it now fails if `loop=loop` is dropped from the adapter, which the previous version did not catch. Pushed in `aeb0b79`.

longcw (approved):

PR #6591, opened 2026-07-28
fix(llm): reposition mid-conversation instructions for Gemini gateway and Mistral

L65FREAD:
## Summary

`generate_reply(instructions=...)` appends the per-turn instructions as a trailing system message. The Mistral Conversations serializer maps every system message onto the `instructions` request field, so the trailing per-turn message silently overwrote the agent's base prompt.

The mistralai serializer now routes through the SDK's existing `convert_mid_conversation_instructions` (already used by the google/anthropic/aws serializers), which keeps the first system message as the base preamble and rewrites later ones as position-preserving `<instructions>`-wrapped user turns.

Helper hardening while touching it:

- A system message only counts as the preamble when no conversation content precedes it — previously, with no base system message anywhere, a trailing per-turn message stayed a trailing system message.
- Text-less mid-conversation system messages are dropped instead of leaking through (they could shadow the base prompt on Mistral).

The LiveKit Inference gateway path (`google/` models served via the openai wire format) was originally repositioned client side in `inference/llm.py`; that change has been dropped per review — the gateway already handles it for `gemini`, and the remaining `gemma` gap is being fixed server side.

## Testing

- New unit tests: mistral serializer conversion, openai serializer pass-through, the no-preamble edge, and the empty-system-message edge.
- `tests/test_chat_ctx.py`: 31 passed, 1 skipped; ruff check and format clean.

### DIFF
--- a/livekit-agents/livekit/agents/inference/llm.py
+++ b/livekit-agents/livekit/agents/inference/llm.py
@@ -22,7 +22,6 @@ from typing_extensions import TypedDict
 from .. import llm
 from .._exceptions import APIConnectionError, APIStatusError, APITimeoutError
 from ..llm import ToolChoice, utils as llm_utils
-from ..llm._provider_format.utils import convert_mid_conversation_instructions
 from ..llm.chat_context import ChatContext
 from ..llm.tool_context import Tool
 from ..log import logger
@@ -76,9 +75,6 @@ _REASONING_EFFORT_TOOL_INCOMPATIBLE_PREFIXES: set[str] = {"gpt-5.2", "gpt-5.4"}
 _MODEL_THINK_TAGS = {
     "google/gemma-4-31b-it": ("<|channel>thought", "<channel|>"),
 }
-
-# providers that reject system messages after the start of the conversation
-_REPOSITION_INSTRUCTIONS_MODEL_PREFIXES: tuple[str, ...] = ("google/",)
 
 
 def drop_unsupported_params(
@@ -360,9 +356,6 @@ class LLM(llm.LLM):
             tools=tools or [],
             conn_options=conn_options,
             extra_kwargs=extra,
-            reposition_instructions=self._opts.model.startswith(
-                _REPOSITION_INSTRUCTIONS_MODEL_PREFIXES
-            ),
         )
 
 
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -398,7 +407,12 @@ async def _run(self) -> None:
         retryable = True
 
         try:
-            chat_ctx, _ = self._chat_ctx.to_provider_format(format=self._provider_fmt)
+            ctx = (
+                convert_mid_conversation_instructions(self._chat_ctx)
```
longcw: should this be handled in the gateway instead? cc @adrian-cowham

L65FREAD: the gateway is probably the right long-term fix for this, fixing all sdks

however, the Mistral fix, the helper hardening, and the AIPlatformLLM opt-in all bypass the gateway. AIPlatformLLM needs the fix either way, since self-deployed Vertex endpoints reject non-leading system roles
happy to drop the gateway bit from this pr

adrian-cowham: We do this already for `gemini` models, the gap is our own hosted (i.e. not on Vertex) `gemma` models. I will address this server side in the gateway for `gemma`. @L65FREAD would you mind dropping the `Inference/llm.py` changes from this PR? I'll get the server side changes out by the end of the week.

livekit-agents/livekit/agents/llm/_provider_format/utils.py
```diff
@@ -17,38 +17,40 @@ def convert_mid_conversation_instructions(
...
-            )
-        else:
-            if item.type == "message" and item.role in ("system", "developer"):
-                first_system_seen = True
-            items.append(item)
+            continue
+
+        if item.type in ("message", "function_call", "function_call_output"):
+            preamble_allowed = False
+        items.append(item)
```
longcw: I'd like to drop the rule that "no conversation content (messages, tool calls)
    precedes it", it's not wrong but also not fully validated. maybe like
```python
if item.type == "message" and item.role in ("system", "developer"):
    if not first_system_seen:
        first_system_seen = True
        items.append(item)
    elif text := item.raw_text_content:
        items.append(llm.ChatMessage(...))  # positioned <instructions> turn
    continue                                # text-less: dropped
items.append(item)
```

longcw (approved):

PR #6587, opened 2026-07-28
(xai stt): expose vad_threshold, smart_turn, smart_turn_timeout, and keyterm params

ajayarora1235:
Fixes #6588.

Follows the pattern of #5493 (which exposed `endpointing`): adds the remaining xAI STT websocket parameters documented at https://docs.x.ai/developers/model-capabilities/audio/speech-to-text as constructor args / `update_options` on `xai.STT`:

- `vad_threshold` (0.0–1.0) — built-in VAD sensitivity; xAI's websocket default is 0.08
- `smart_turn` (0.0–1.0) — confidence threshold for xAI's ML end-of-turn ("Smart Turn") prediction
- `smart_turn_timeout` (1–5000 ms) — silence backstop when Smart Turn predicts the speaker is continuing
- `keyterm` (list, max 100 terms) — vocabulary biasing; sent as repeated query params, so the websocket `params` dict became a list of pairs

All four are `NotGivenOr` and omitted from the query string when unset, so existing behavior is unchanged — the default connection sends the identical query as before. Verified with `ruff check` / `ruff format`; param names/ranges are taken from the xAI docs above.

Motivation: these are the main turn-taking tunables for using xai/stt-1 with `turn_detection="stt"` in an AgentSession — today the plugin (and the Inference gateway's `xai.ExtraSettings`, which rejects these keys) only exposes `endpointing`, so smart-turn thresholds and keyword boosting are unreachable without patching the plugin.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6594, opened 2026-07-28
(cartesia stt): expose ink-2 turn-detection thresholds and keyterm params

ajayarora1235:
Sibling of #6587 (same pattern for xAI STT): exposes the turn-detection tuning parameters documented for Cartesia's `/stt/turns/websocket` (https://docs.cartesia.ai/api-reference/stt/turns/websocket) on `cartesia.STT`:

- `turn_start_threshold` (0.5–0.9, default 0.8) — likelihood above which a turn starts
- `turn_eager_end_threshold` (0.3–0.6, default 0.4) — likelihood below which `turn.eager_end` fires
- `turn_end_threshold` (0.05–0.5, default 0.2) — likelihood below which the turn ends
- `turn_end_timeout_ms` (640–11200, default 5600) — silence backstop for ending a turn
- `keyterm` (list, up to 100 terms / 1200 chars) — recall biasing; sent as repeated query params, so the turns-websocket `params` dict became a list of pairs

These are the primary tuning levers when running ink-2 with `turn_detection="stt"` in an `AgentSession` — today the plugin connects to the turns endpoint with no way to adjust them. All five are `NotGivenOr` and omitted when unset, so the default connection sends the identical query string as before. Passing any of them with a non-turn-detecting model (ink-whisper → the legacy endpoint) raises `ValueError`, mirroring the model-gated params in the AssemblyAI plugin.

Live-verified against `wss://api.cartesia.ai/stt/turns/websocket` (real key, today):
- in-range values (e.g. `turn_start_threshold=0.85`, repeated `keyterm`) → `{"type":"connected", ...}`
- out-of-range (`turn_start_threshold=2.0`) → error frame `"turn_start_threshold must be between 0.5 and 0.9."` + close — so the server parses and validates exactly these param names (unknown params are silently ignored, which is why the client-side `ValueError` guard is worth having)

Verified with `ruff check` / `ruff format`.

### DIFF
--- a/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/_recognize_streams/auto_finalize_recognize_stream.py
+++ b/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/_recognize_streams/auto_finalize_recognize_stream.py
@@ -216,6 +216,11 @@ class AutoFinalizeRecognizeStream(CartesiaRecognizeStream):
         ws_base_url: str,
         session: aiohttp.ClientSession,
         language: LanguageCode,
+        turn_start_threshold: NotGivenOr[float] = NOT_GIVEN,
+        turn_eager_end_threshold: NotGivenOr[float] = NOT_GIVEN,
+        turn_end_threshold: NotGivenOr[float] = NOT_GIVEN,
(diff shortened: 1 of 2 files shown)


tinalenguyen (approved):
thank you for the PR!

PR #6596, opened 2026-07-29
fix(inference): tolerate null token counts in streamed usage

### LINKED ISSUE #6595: CompletionUsage validation crashes session when provider returns usage.completion_tokens=None mid-stream (OpenAI-compatible path)
### Bug Description

When using the OpenAI-compatible LLM path (`livekit.plugins.openai.llm.LLM`, which inherits its streaming `_run` loop from `livekit.agents.inference.llm.LLMStream`), if the upstream provider sends a final streaming chunk where `chunk.usage` is a non-null object but one of its fields (e.g. `completion_tokens`) is `null`, the SDK crashes.

The code at `livekit-agents/livekit/agents/inference/llm.py` only guards on `if chunk.usage is not None:` before constructing:
```python
usage=llm.CompletionUsage(
    completion_tokens=chunk.usage.completion_tokens,
    prompt_tokens=chunk.usage.prompt_tokens,
    total_tokens=chunk.usage.total_tokens,
    ...
)
```
`CompletionUsage.completion_tokens` is a required `int` (no `Optional`), so a `null` value raises a pydantic `ValidationError`, which gets wrapped and re-raised as `APIConnectionError(retryable=False)`. Because it's marked non-retryable, this eventually surfaces as an unrecoverable `llm_error` that terminates the whole `AgentSession` — killing an in-progress voice call — even though the actual issue is a malformed/partial usage payload, not a real connection failure.

We're hitting this against an OpenAI-compatible proxy in front of a custom model; the proxy occasionally omits `completion_tokens` in the usage object on stream completion. This is analogous to #5404, which fixed a similar "usage present but incomplete on final chunk" issue for the Google/Gemini plugin — but the equivalent guard was never added to the shared OpenAI-compatible streaming path.

### Expected Behavior

A `null`/missing field inside `chunk.usage` should not raise an unhandled, non-retryable exception that ends the session. At minimum, missing numeric usage fields should default to `0` (or the chunk should be treated as retryable) rather than crashing.

### Reproduction Steps

1. Configure `livekit.plugins.openai.LLM` against an OpenAI-compatible endpoint (e.g. a proxy in front of a non-OpenAI model).
2. Have that endpoint return a streamed response whose final SSE chunk includes a `usage` object with `completion_tokens: null` (rather than omitting `usage` entirely).
3. Observe the `ValidationError` for `CompletionUsage`, wrapped into `APIConnectionError(retryable=False)`, terminating the session.

### Operating System

Linux (GKE)

### Models Used

OpenAI-compatible proxy in front of a custom-hosted model

### Package Versions

```
livekit==1.1.13
livekit-agents==1.6.5
livekit-plugins-openai==1.6.5
```

### Proposed Solution

```python
if chunk.usage is not None:
    usage = chunk.usage
    usage_chunk = llm.ChatChunk(
        id=chunk.id,
        usage=llm.CompletionUsage(
            completion_tokens=usage.completion_tokens or 0,
            prompt_tokens=usage.prompt_tokens or 0,
            prompt_cached_tokens=cached_tokens or 0,
            total_tokens=usage.total_tokens or 0,
            service_tier=getattr(chunk, "service_tier", None),
        ),
    )
    self._event_ch.send_nowait(usage_chunk)
```

### Additional Context

Traceback:
```
  File ".../livekit/agents/inference/llm.py", line 445, in _run
    usage=llm.CompletionUsage(
  File ".../pydantic/main.py", line 250, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for CompletionUsage
completion_tokens
  Input should be a valid integer [type=int_type, input_value=None, input_type=NoneType]

livekit.agents._exceptions.APIConnectionError: Connection error.
```

VihaanAgarwal:
Fixes #6595

## What breaks

On the OpenAI-compatible path (`livekit.agents.inference.llm.LLMStream`), a provider can send a final streaming chunk where `usage` is present but individual counts are `null`. We only guard on `if chunk.usage is not None:`, then pass the fields straight into `CompletionUsage`, where `completion_tokens`, `prompt_tokens` and `total_tokens` are all required `int`.

The pydantic `ValidationError` gets caught by the generic `except Exception` at the bottom of `_run` and re-raised as `APIConnectionError(retryable=False)`. Because it's non-retryable it surfaces as an unrecoverable `llm_error` and terminates the `AgentSession`, so a malformed usage payload drops a live call.

The reporter hits this behind a proxy in front of a custom model that occasionally omits `completion_tokens`.

## Fix

Coerce the three counts with `or 0`. That's what `prompt_cached_tokens` on the adjacent line already does, so this keeps the block internally consistent rather than introducing a new pattern. Complete payloads are unaffected.

Same shape as #5404, which fixed the other half of this (usage present but not reached).

## Tests

`tests/test_inference_llm_usage.py` drives `LLMStream` with a fake OpenAI stream, so it needs no network or credentials:

- a chunk with `completion_tokens=None` yields usage with `0` and does not raise
- a chunk with `prompt_tokens`/`total_tokens` null is coerced the same way
- a fully-populated chunk passes through untouched

The first two fail on `main` with `APIConnectionError` and pass with the change.

Full unit gate: 1175 passed, 4 skipped. `make type-check`, `make lint` and `make format-check` are clean. The 9 errors in the pytest summary are a pre-existing OTLP metrics-exporter teardown issue that reproduces identically on a clean checkout.

### DIFF
--- a/livekit-agents/livekit/agents/inference/llm.py
+++ b/livekit-agents/livekit/agents/inference/llm.py
@@ -454,10 +454,10 @@ class LLMStream(llm.LLMStream):
                         usage_chunk = llm.ChatChunk(
                             id=chunk.id,
                             usage=llm.CompletionUsage(
-                                completion_tokens=chunk.usage.completion_tokens,
-                                prompt_tokens=chunk.usage.prompt_tokens,
+                                completion_tokens=chunk.usage.completion_tokens or 0,
+                                prompt_tokens=chunk.usage.prompt_tokens or 0,
                                 prompt_cached_tokens=cached_tokens or 0,
-                                total_tokens=chunk.usage.total_tokens,
+                                total_tokens=chunk.usage.total_tokens or 0,
                                 service_tier=getattr(chunk, "service_tier", None),
                             ),
                         )
--- a/tests/test_inference_llm_usage.py
+++ b/tests/test_inference_llm_usage.py
@@ -1,0 +1,108 @@
+from __future__ import annotations
+
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from livekit.agents.inference.llm import LLMStream
+from livekit.agents.llm import ChatContext
+from livekit.agents.types import APIConnectOptions
+
+pytestmark = pytest.mark.unit
+
+
+class _FakeOpenAIStream:
+    """Stands in for the openai streaming response: async context manager + async iterator."""
+
+    def __init__(self, chunks: list[SimpleNamespace]) -> None:
+        self._chunks = chunks
+
+    async def __aenter__(self) -> _FakeOpenAIStream:
+        return self
+
+    async def __aexit__(self, *args: object) -> None:
+        return None
+
+    async def __aiter__(self):
+        for chunk in self._chunks:
+            yield chunk
+
+
+def _usage_chunk(
+    *,
+    completion_tokens: int | None,
+    prompt_tokens: int | None,
+    total_tokens: int | None,
+) -> SimpleNamespace:
+    return SimpleNamespace(
+        id="chunk-id",
+        choices=[],
+        usage=SimpleNamespace(
+            completion_tokens=completion_tokens,
+            prompt_tokens=prompt_tokens,
+            total_tokens=total_tokens,
+            prompt_tokens_details=None,
+        ),
+    )
+
+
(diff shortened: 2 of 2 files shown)


VihaanAgarwal:
Closing, #6763 landed and covers this.

PR #6597, opened 2026-07-29
(inference): type Cartesia turn-detection params in CartesiaOptions

ajayarora1235:
Companion to #6594 (which exposed these on the direct plugin): the Inference gateway **already accepts and validates** Cartesia's turn-detection tuning params via `extra_kwargs` — they're just missing from the `CartesiaOptions` TypedDict, so users can't discover them.

Verified live against the gateway today:
- `extra_kwargs={"turn_start_threshold": 0.6}` on `cartesia/ink-2` → session created and runs normally.
- `extra_kwargs={"turn_start_threshold": 2.0}` → session creation fails with `cartesiaturns: extra settings validation failed: turn_start_threshold must be <= 0.9, got 2` — i.e. the gateway's cartesia-turns adapter knows these params and enforces Cartesia's documented ranges.

This PR only adds the keys to the TypedDict (typing/discoverability — `extra_kwargs` already passes them through verbatim) and annotates the two existing keys as ink-whisper-only, per Cartesia's API reference. Ranges/defaults from https://docs.cartesia.ai/api-reference/stt/turns/websocket.

Verified with `ruff check` / `ruff format`.

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -68,8 +68,15 @@ InworldModels = Literal["inworld/inworld-stt-1",]
 
 
 class CartesiaOptions(TypedDict, total=False):
-    min_volume: float  # default: not specified
-    max_silence_duration_secs: float  # default: not specified
+    min_volume: float  # ink-whisper only; default: not specified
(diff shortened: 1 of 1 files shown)


tinalenguyen (approved):
thanks for the PR!

Issue #4087, opened 2025-11-25
Elevenlabs Scribe v2 never commits the turn


IngLP:
### Bug Description

The Agent start responding only after 
"Received message type committed_transcript"
but 11labs Scribe yields that message VERY late. Many seconds later the user stopped talking, I tried in a room with absolute silence too.

### Expected Behavior

Silero VAD commits the turn after detecting user silence.
Agent respond to the user message already timely received with
"Received message type partial_transcript:"

### Reproduction Steps

```bash
1. Create an Agent with SileroVAD enabled and use Elevenlabs STT Sribe v2 with streaming=True
2. Start a call, talk, the Agent will never answer (or only after 10-15sec)
```

### Operating System

Docker on K8s

### Package Versions

```bash
livekit-agents==1.3.4
```

### Proposed Solution

```python

```

longcw:
see comments here https://github.com/livekit/agents/pull/3909#pullrequestreview-3481227767

and a potential fix is to use local VAD with their manual commit strategy https://github.com/livekit/agents/pull/4043 but according to test it's still not very stable since https://github.com/livekit/agents/pull/4043#issuecomment-3562607911

IngLP:
Ok thanks, this means we have to wait for #4043 to be fixed and merged?

IngLP:
Hi, any news?

longcw:
#4043 still cannot solve the issue perfectly because the scribe v2 will  generate text no matter it's silent or even just noise (random texts in this case).

for now you can try to adjust the scribe v2 server-side VAD parameters following https://github.com/livekit/agents/blob/livekit-agents@1.3.6/examples/other/elevenlab_scribe_v2.py

IngLP:
Hi @longcw , thanks for the answer. In my experiments i saw that when server vad of 11labs is off, transcriptions are correct, they just don't commit. What about leaving off 11labs VAD and relying on our VAD to commit speech?

longcw:
>  What about leaving off 11labs VAD and relying on our VAD to commit speech?

that's exactly the #4043 trying to do, but the local VAD also has false positives that may commit when there is no user speaking, and in that case the 11labs STT will return some random texts that interrupt the agent.

IngLP:
Hi @longcw , I double checked the code: I am leaving server_vad to NotGiven, and looking at the code this means leaving the 11labs VAD active (!!).
In such case, as said, I see no false positives at all.
[...]

PR #6598, opened 2026-07-29
fix(stt): send the turn commit when a flush yields no buffered audio

chenghao-mou:
`AudioByteStream.flush()` returns `[]` when its buffer happens to be empty, roughly 1 in 5 flushes with 10ms input frames repacked into 50ms chunks. Any finalize gated on `for frame in frames:` is then skipped, and `has_ended` leaks into the next utterance so the message fires mid-way through the following turn.

**elevenlabs:** Scribe v2 realtime never finalized at all without a `commit`, so send an empty `commit: true` chunk, hoisted out of the frame loop.
**deepgram:** hoist `Finalize` out of the frame loop (`stt_v2` gets the same dedent for its metrics flush).

**Tests:** both plugins get coverage that drives the real `SpeechStream._run` loop against a fake socket, asserting the commit lands after a flush with an empty repack buffer (the regression) and after the leftover audio when a partial chunk is pending (ordering). Both fail with the fix reverted.

**Scope:** the default `stt_node` only calls `push_frame`, and the pipeline commits a turn by pushing silence, so nothing flushes a streaming STT stream in a normal `AgentSession`. Verified by patching `RecognizeStream.flush`/`end_input` across the unit suite: 0 calls from the voice pipeline. This hardens direct `stt.stream()` + `flush()` use; it does not on its own fix AGT-2264.

Gladia has the same shape and is deliberately left alone: `stop_recording` is session-terminating there (see `gladia/stt.py:388`), so making it fire reliably per turn would be worse than the current miss.

Related to AGT-2264 (#4087)

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -628,10 +628,10 @@ class SpeechStream(stt.SpeechStream):
                         self._audio_duration_collector.push(frame.duration)
                         await ws.send_bytes(frame.data.tobytes())
 
-                        if has_ended:
-                            self._audio_duration_collector.flush()
-                            await ws.send_str(SpeechStream._FINALIZE_MSG)
-                            has_ended = False
+                    if has_ended:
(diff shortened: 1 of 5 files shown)


davidzhao (approved):
great catch!

PR #6599, opened 2026-07-29
fix: defer interim interruptions to local VAD

chenghao-mou:
Local VAD should own barge-in detection and apply the configured minimum speech duration. STT partials can be stale or punctuation-only, so treating both as interruption signals can stop valid agent playback.

Sessions without VAD retain interim-based interruption, and final transcripts remain a fallback.

Fixes AGT-3202


u9g (approved):

u9g:
Tested with cue-cli.

davidzhao (approved):

Issue #6569, opened 2026-07-27
Realtime (Gemini Live): speech marked played-out at generation completion when the model calls a tool mid-utterance — buffered audio dropped mid-word; tool response can be lost, deadlocking the session


blake-newman:
## Versions

- livekit-agents 1.6.6 (reproduced); still present as of 1.6.7 — the `livekit-agents@1.6.6...livekit-agents@1.6.7` compare touches neither `plugins/google/realtime/realtime_api.py` nor the `voice/generation.py` / speech-handle / transcript-synchronizer machinery. #6445 fixes an adjacent failure in the same family (realtime turn dying silently on a racy chat-ctx push) but explicitly scopes out the tool-output path, and our trace shows no `RealtimeError` — the tool response send was never attempted.
- livekit-plugins-google 1.6.6
- google-genai 2.12.1
- Python 3.11, `AgentSession` with `RealtimeModel` (Gemini Live over Vertex, native audio), server-side turn detection (`turn_handling=TurnHandlingOptions(turn_detection="realtime_llm")` equivalent — automatic activity detection enabled)

## Summary

Gemini Live native-audio models routinely emit a `toolCall` while their generation is still
streaming/playing (they do not stop speaking to call a tool). When that happens, the speech is
treated as played-out at **generation** completion rather than **playout** completion:

1. The still-buffered remainder of the utterance is dropped mid-word. The transcript
   synchronizer logs `_SegmentSynchronizerImpl.playback_finished called before text/audio
   input is done` at the moment it happens. Gemini streams audio ~5x faster than realtime, so
   the gap between "generation complete" and "playout complete" is large (14 of 22 seconds in
   our repro).
2. `SpeechHandle.wait_for_playout()` / downstream `playback_finished` consumers observe the
   speech as finished early (our mic-gating that unmutes "when the greeting finishes playing"
   unmuted 18s early).
3. In a timing subset, the tool's `functionResponse` is **never sent** to Gemini (the tool
   executed and returned; no `tool_response` frame ever went out on the websocket). Gemini then
   waits indefinitely for the response and the session is deadlocked — the agent never speaks
   again; subsequent `generate_reply()` calls produce no generation.

## Repro evidence (websocket-level, timestamps from one session)

```
15:53:18.4   intro audio starts playing to the room (22s of audio; fully streamed by ~15:53:22)
15:53:23.111 < toolCall  (model calls a tool ~5s into playout)
15:53:23.114 livekit.agents executing tool
[...]

PR #6600, opened 2026-07-29
fix(voice): deliver tool results to realtime session on interruption

longcw:
Gemini Live blocks a turn until every tool call it emitted is answered, and offers no way to cancel one. If playout is interrupted from our side while a tool result is still owed, dropping that result leaves the server waiting forever: it stops responding and subsequent `generate_reply()` calls produce no generation. Reported in livekit/agents#6569.

## When it happens

Only when the interrupt is client-initiated — a programmatic `session.interrupt()`, a handoff or close, false-interruption resume, or local turn detection firing alone. A voice barge-in makes Gemini send `interrupted`, which releases its wait on its own.

## Fix

The realtime path now mirrors what livekit/agents#6349 did for the pipeline path: commit the outputs of tools that finished despite the interruption, handoffs excluded so they stay retryable. On top of that it syncs those outputs to the realtime session, which the pipeline has no need to do.

Delivering a result makes Gemini reply, which is unwanted right after an interruption, so it goes out with `SILENT` scheduling. Gemini ignores `scheduling` on `BLOCKING` declarations (the default), so with blocking tools the reply is unavoidable — unblocking the turn is the better trade.


livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -657,10 +670,20 @@ async def update_chat_ctx(self, chat_ctx: llm.ChatContext) -> None:
                 append_ctx.items.append(item)
 
         if append_ctx.items:
+            scheduling = self._opts.tool_response_scheduling
+            if (
+                not self._opts.vertexai  # scheduling is not supported by vertex
```
u9g: Redundant now that `create_function_response` skips `scheduling` for vertexai — this check can go.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -3983,6 +3983,27 @@ def _create_assistant_message(
...
+                sanitized_out.fnc_call_out
+                for sanitized_out in tool_output.output
+                if sanitized_out.fnc_call_out is not None and sanitized_out.agent_task is None
+            ]
+            if interrupted_fnc_outputs:
+                self._agent._chat_ctx.insert(interrupted_fnc_outputs)
+                self._session._tool_items_added(interrupted_fnc_outputs)
+
+                # unlike the pipeline, a realtime model holds the call open server-side
+                chat_ctx = self._rt_session.chat_ctx.copy()
```
u9g, line 4113: `chat_ctx` already returns `self._chat_ctx.copy()`, so this copies twice.

longcw: I'd keep `copy()` here to make it explicit since the `rt_session.chat_ctx` is implemented per plugin and not documented to must be a copy.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -4098,6 +4098,27 @@ def _create_assistant_message(
...
+
+            # commit results of tools that finished despite the interruption, similar to the pipeline task
+            interrupted_fnc_outputs = [
+                sanitized_out.fnc_call_out
+                for sanitized_out in tool_output.output
+                if sanitized_out.fnc_call_out is not None and sanitized_out.agent_task is None
+            ]
+            if interrupted_fnc_outputs:
+                self._agent._chat_ctx.insert(interrupted_fnc_outputs)
+                self._session._tool_items_added(interrupted_fnc_outputs)
```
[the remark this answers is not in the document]
longcw: it's not introduced in this pr.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -819,6 +841,9 @@ def start_user_activity(self) -> None:
             )
 
     def interrupt(self) -> None:
+        # recorded locally since this cannot reach Gemini under automatic activity detection
+        self._playout_interrupted = True
```
chenghao-mou, line 845: Q: if `_start_new_generation` clears the flag and but immediately sets it to `True` with `_handle_input_speech_started`, we would have silent tools whenever the generation is not triggered by "generate_reply"?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -4098,6 +4098,27 @@ def _create_assistant_message(
...
+
+                # unlike the pipeline, a realtime model holds the call open server-side
+                chat_ctx = self._rt_session.chat_ctx.copy()
+                chat_ctx.items.extend(interrupted_fnc_outputs)
+                try:
+                    await self._rt_session.update_chat_ctx(chat_ctx)
```
chenghao-mou, line 4116: seems update might be handled differently, from Sol:

```
Ultravox converts that output to `agent_reaction="speaks"`, while Phonic starts a new assistant turn. Thus an interrupted tool can make non-Google agents speak again. Silent interrupted-result delivery needs provider-specific semantics or an explicit capability.
```

longcw:
Superseded by #6785, which broadens this to every tool call rather than just Gemini's.

PR #6607, opened 2026-07-29
fix: disable AEC warmup for outbound SIP calls

chenghao-mou:
Outbound SIP callers can speak before the greeting, but the default AEC warmup suppresses those initial interruptions. Detect outbound SIP participants through the empty sip.ruleID attribute and disable warmup only when the option is omitted; explicit settings and the three-second default for other calls remain unchanged.

Fixes AGT-3199


chenghao-mou:
Superseded by #6608 after correcting the head branch name.

PR #6608, opened 2026-07-29
fix: disable AEC warmup for outbound SIP calls

chenghao-mou:
Outbound SIP callers can speak before the greeting, but the default AEC warmup suppresses those initial interruptions. Detect outbound SIP participants through the empty sip.ruleID attribute and disable warmup only when the option is omitted; explicit settings and the three-second default for other calls remain unchanged.

Tested with a SIP outbound call.

Fixes AGT-3199

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -95,6 +95,10 @@ if TYPE_CHECKING:
     from .transcription.text_transforms import TextTransforms
 
 
+_SIP_RULE_ID_ATTR = "sip.ruleID"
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -95,6 +95,10 @@
     from .transcription.text_transforms import TextTransforms
 
 
+_SIP_RULE_ID_ATTR = "sip.ruleID"
```
theomonnom, line 98: We really need to have a better abstraction for SIP in general.

theomonnom (approved):

davidzhao (approved):

PR #6612, opened 2026-07-30
fix: scope the foreground idle hold to one activity

longcw:
`ctx.foreground()` took a session-wide idle hold, so it also blocked `wait_for_idle` on activities it had nothing to do with.

When the foregrounded work is an `AgentTask`, any async tool the task's model calls deadlocks: the tool's deferred reply waits for the floor to go quiet, the floor is only released when the task finishes, and the task only finishes once the model has reacted to that reply. 

The hold now lives on the activity that had the floor rather than on the session. A held parent activity still keeps sibling tool replies out for the whole block — its object survives pause/resume, so the protection is intact when the task returns — while the task's own activity delivers its replies as usual.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -560,12 +560,6 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         self._user_turn_released: asyncio.Event = asyncio.Event()
         self._user_turn_released.set()
 
-        # count of active `_wait_for_idle_and_hold` scopes; while > 0, non-holder
-        # `wait_for_idle` callers block until release. holder bypasses via contextvar.
-        self._idle_holds: int = 0
-        self._idle_released: asyncio.Event = asyncio.Event()
-        self._idle_released.set()
-
         self._global_run_state: RunResult | None = None
         # TODO(theomonnom): need a better way to expose early assistant metrics
         self._early_assistant_metrics: MetricsReport | None = None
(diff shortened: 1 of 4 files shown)


livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1495,20 +1486,24 @@ async def wait_for_idle(self) -> AgentActivity:
...
     async def _wait_for_idle_and_hold(self) -> AsyncIterator[AgentActivity]:
-        """Wait for idle, then block other ``wait_for_idle`` callers until exit."""
+        """Wait for idle, then block other ``wait_for_idle`` callers on that activity.
+
+        Per-activity, not per-session: work handed to another agent keeps its own floor,
+        or its replies would wait on the work that is waiting for them.
+        """
         from .agent_activity import _IdleHoldContextVar
 
         activity = await self.wait_for_idle()
```
chenghao-mou, line 1496: One thing noticed from Codex:

 **A foreground caller can acquire the hold on a paused activity.**

If `foreground()` starts waiting on activity A, then an `AgentTask` pauses A and starts B, `wait_for_idle()` can still return A. The new code holds A while subsequent session speech uses B. Other idle callers therefore proceed during the foreground block. Recheck the current activity after waiting and retarget before acquiring the hold. [[agent_session.py:1496](https://github.com/livekit/agents/blob/f7a07cbb82edd651ab4869ae4c199f5d092ebc5b/livekit-agents/livekit/agents/voice/agent_session.py#L1496-L1498)](https://github.com/livekit/agents/blob/f7a07cbb82edd651ab4869ae4c199f5d092ebc5b/livekit-agents/livekit/agents/voice/agent_session.py#L1496-L1498)

PR #6614, opened 2026-07-30
bithuman: fail closed on Python versions the SDK has no build for

u9g:
Rescoped: #6561 landed the marker on the plugin's own dependency while this was open, so the root `override-dependencies` approach here is redundant. Rebased onto main; what's left is the metadata fix.

`requires-python` was `>=3.10` and the classifiers claimed 3.10 only, while `bithuman` 2.6+ requires `>=3.11,<3.14`. The wide dep range (`>=0.5.25,<3`) let a 3.10 install backtrack to `bithuman==2.3.10` (which ships cp310 wheels) and succeed against a stale SDK rather than being refused, so the floor moves into `requires-python` and the classifiers list the versions that actually work.

The `<3.14` bound stays a marker: the virtual workspace root intersects `requires-python` over all members, so an upper bound here would cap the whole workspace.

Verified: the built wheel reports `Requires-Python: >=3.11` and `Requires-Dist: bithuman<3,>=0.5.25; python_version < '3.14'`; `uv sync --python 3.14 --all-extras --dev` resolves with `bithuman` absent and the plugin still installed; `--python 3.13` installs `bithuman==2.7.0`; `uv lock` is a no-op.

Not addressed: there are no Windows or macOS-x86_64 `bithuman` wheels, so installs fail there on every Python version. The `>=0.5.25` floor is also stale — it is what makes deep backtracks possible — but raising it is a user-facing compat change and belongs in its own PR.

### DIFF
--- a/livekit-plugins/livekit-plugins-bithuman/pyproject.toml
+++ b/livekit-plugins/livekit-plugins-bithuman/pyproject.toml
@@ -8,7 +8,7 @@ dynamic = ["version"]
 description = "Agent Framework plugin for services from BitHuman Avatar Rendering"
 readme = "README.md"
 license = "Apache-2.0"
-requires-python = ">=3.10.0"
+requires-python = ">=3.11"
 authors = [{ name = "LiveKit", email = "support@livekit.io" }]
 keywords = ["voice", "ai", "realtime", "audio", "video", "livekit", "webrtc"]
 classifiers = [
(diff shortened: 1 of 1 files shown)


livekit-plugins/livekit-plugins-bithuman/pyproject.toml
```diff
@@ -8,7 +8,7 @@ dynamic = ["version"]
 description = "Agent Framework plugin for services from BitHuman Avatar Rendering"
 readme = "README.md"
 license = "Apache-2.0"
-requires-python = ">=3.10.0"
+requires-python = ">=3.11"
```
longcw: could you explain a bit why `requires-python = ">=3.11"` is needed?

u9g: current bithuman ([2.6.0](https://pypi.org/pypi/bithuman/2.6.0/json)/[2.7.0](https://pypi.org/pypi/bithuman/2.7.0/json)/[2.8.0](https://pypi.org/pypi/bithuman/2.8.0/json)) only have wheels for: cp311, cp312, cp313

[2.3.10](https://pypi.org/pypi/bithuman/2.6.0/json) includes wheels for: cp310, cp311, cp312, cp313, cp314

then starting >=2.3.11, they only ship cp311–cp313. idk what the difference is of if we are okay with 2.3.10, if we are I guess we can instead do python 3.10-3.14

livekit-plugins/livekit-plugins-bithuman/pyproject.toml
```diff
@@ -18,14 +18,18 @@ classifiers = [
...
 ]
 
-# bithuman builds only for Python 3.11 through 3.13
+# bithuman builds only for Python 3.11 through 3.13. requires-python holds the lower
+# bound; the upper bound has to be a marker here, since the virtual workspace root
+# intersects requires-python over all members and would cap the whole workspace.
 dependencies = [
     "livekit-agents>=1.6.7",
-    "bithuman>=0.5.25,<3; python_version >= '3.11' and python_version < '3.14'",
+    "bithuman>=0.5.25,<3; python_version < '3.14'",
```
u9g: I made `PIL`, `loguru` and `cv2` explicit deps, rather than implicit as they were before.

longcw (approved):

PR #6617, opened 2026-07-30
fix(sarvam): saaras:v3 does not support the config/prompt message

### LINKED ISSUE #6606: livekit-plugins-sarvam: saaras:v3 marks supports_prompt=True but the endpoint it connects to (/speech-to-text/ws) has no documented config/prompt message
## Summary

`SarvamSTTOptions.prompt` / the `MODEL_CONFIGS` table marks `supports_prompt=True` for `saaras:v3`, and the streaming client sends an initial `{"type": "config", "prompt": ...}` message on connect for any model where `_model_supports_prompt()` is true (`stt.py`, `_send_initial_config`). But `saaras:v3` has `use_translate_endpoint=False`, so it connects to `wss://api.sarvam.ai/speech-to-text/ws` — and Sarvam's own AsyncAPI spec for that exact endpoint documents only two client message types: an audio-data message and a flush signal. There is no `config`/`prompt` message anywhere in its schema.

The `config` + `prompt` message *is* real and documented — but only on a different endpoint: the legacy `wss://api.sarvam.ai/speech-to-text-translate/ws` (used only when `use_translate_endpoint=True`, i.e. only for `saaras:v2.5`). That endpoint's own spec says: *"For the latest model (saaras:v3), use the `/speech-to-text` endpoint"* — confirming `saaras:v3` traffic never touches the endpoint that actually supports the config/prompt message.

## Why this matters

We use `model="saaras:v3"` with a fixed `prompt` (a hotword/contextual-biasing list of property names + number words, meant to reduce misrecognition of short spoken numbers) in a production telephony voice agent. We hit a real, reproducible failure today: a caller answering "how many guests" with the same word three turns running got transcribed as "P", then "Hmm", before landing on the correct word the third time — despite that exact word being in the `prompt` hotword list on every call. Given the endpoint mismatch above, we now believe the `prompt` message has likely never had any effect for `saaras:v3` traffic, silently dropped server-side (no error surfaced, consistent with what we observed).

## Ask

Could you confirm whether the `prompt`/`config` message does anything at all when sent to `/speech-to-text/ws` for `saaras:v3`? If it doesn't:
- `MODEL_CONFIGS["saaras:v3"].supports_prompt` should probably be `False` (or the plugin should document that `prompt` is a no-op / silently ignored for `v3`), so callers don't build a biasing feature on a parameter that isn't wired to anything.
- If Sarvam does have an equivalent for `v3` (a different message shape, a query param, etc.), pointing us at the right mechanism would help — we couldn't find one in Sarvam's own `/speech-to-text` docs.

## Environment

- `livekit-plugins-sarvam==1.5.17`
- Relevant source: `stt.py` — `MODEL_CONFIGS` (~line 147), `_get_urls_for_model` (~line 332), `_send_initial_config` (~line 1053)
- Sarvam docs referenced: `https://docs.sarvam.ai/api-reference/speech-to-text/transcribe/ws` (current, no config message documented) vs. `https://docs.sarvam.ai/api-reference/speech-to-text-translate/translate/ws` (legacy, documents the config/prompt message, v2.5-only)

biztex:
Fixes #6606.

`MODEL_CONFIGS["saaras:v3"]` declared `supports_prompt=True`, so the plugin sent a config/prompt message on a connection that ignores it — callers could set `prompt` for hotword biasing and it never reached anything.

Sets the flag to `False`. The existing streaming guard already skips the config frame when prompt support is false, so no control-flow change is needed.


dhruvladia-sarvam:
@biztex Thanks for finding this. Let’s keep the fix scoped to the reported capability mismatch:
Change MODEL_CONFIGS["saaras:v3"].supports_prompt from True to False.

The existing streaming guard already avoids sending the config frame when prompt support is false, so no further control-flow change is needed. Sarvam silently ignores a raw config/prompt frame on the v3 endpoint, it does not return a provider error, and since saaras:v2.5 is on its way to be deprecated, the LiveKit docs reflect no support for a prompt parameter. Please drop the unknown-model behaviour change, warning logs, and expanded endpoint/protocol commentary.

biztex:
Thanks — done in 27c8dc3. It's just the flag now: `supports_prompt=False` for `saaras:v3`, nothing else. Dropped the unknown-model change, the warnings and the endpoint commentary, and left the streaming guard alone.

Kept a two-line test asserting v3 is False and v2.5 is still True, happy to drop that too if you'd rather have it as a pure one-liner.

livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/stt.py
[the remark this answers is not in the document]
biztex: Leaving this one as is. @dhruvladia-sarvam asked to keep the change to the reported capability mismatch and to drop the unknown-model behaviour change specifically, so re-adding it here would undo that.

On the impact: he confirmed Sarvam silently ignores a raw config/prompt frame on the v3 endpoint rather than returning an error, so an unknown `saaras*` model doesn't get its connection rejected — worth a separate change if the fallback should be tightened, with his input on what unknown models should default to.

The stale line in the description is already gone; it was rewritten when the scope was trimmed.

tests/test_sarvam_stt_options.py
davidzhao, line 1: let's drop the test as well, it's not useful in the current state

biztex: Dropped in 5b532ea — the PR is just the one-line flag change now.

tinalenguyen (approved):

PR #6618, opened 2026-07-30
feat(google): accept a credentials object on STT and TTS

### LINKED ISSUE #6586: ivekit-google feature request — credentials param on STT & TTS (issue-template format)
### Feature Type

Would make my life easier

### Feature Description

google.STT and google.TTS (Google/Vertex Speech-to-Text and Text-to-Speech) only support authentication via `credentials_info` (a service-account-key dict),`credentials_file` (a path on disk), or `google.auth.default()` (environment-based ADC). There's no way to pass a google.auth.credentials.Credentials` object directly.

This is a problem for setups using Workload Identity Federation with custom credential providers, where credentials are obtained in memory and don't exist as files on disk.
RealtimeModel already accepts a `credentials` parameter, and every other Google client library (Secret Manager, Storage, Cloud SQL, etc.) accepts one too. STT and TTS should match — especially since both already build their underlying client with `credentials=<object>` internally (`STT._create_client`,
`TTS._ensure_client`); they just derive that object from a file first. The ask is to allow that object to be supplied directly.

Proposed change: add an optional `credentials` parameter to `google.STT` and `google.TTS` and pass it through to the underlying `SpeechAsyncClient` /`TextToSpeechAsyncClient`. Defaults to NOT_GIVEN, so no behavior change for existing callers.

Subclassing the plugin to override client creation, or serializing the in-memory credentials to a temporary file — both fragile and version-coupled.

### Additional Context

Same request as the one already accepted for `RealtimeModel` (add a `credentials` parameter) — this extends that parity to the STT and TTS plugins in the same `livekit-plugins-google` package.

biztex:
﻿Fixes #6586

## Problem

`google.STT` and `google.TTS` only authenticate via `credentials_info` (a service-account dict), `credentials_file` (a path on disk), or Application Default Credentials. Workload Identity Federation setups with custom credential providers hold a `google.auth.credentials.Credentials` object in memory that never exists as a file — and had no way to pass it, despite both classes already constructing their clients with `credentials=<object>` internally (they just insisted on deriving that object from a file first).

## Change

Adds an optional `credentials` parameter to both `google.STT` and `google.TTS`, passed straight through to `SpeechAsyncClient` / `TextToSpeechAsyncClient`. It takes precedence over `credentials_info` / `credentials_file`, and STT skips its ADC-availability check when the object is supplied. Signature and type match the `credentials` parameter already accepted by `RealtimeModel`, `LLM`, and the aiplatform LLM in this same plugin — this just completes the parity, exactly as requested in the issue. Default is `None`, so existing callers are unaffected.

## Tests

New hermetic `tests/test_google_credentials.py` (unit-marked; uses `AnonymousCredentials`, no network/files/env): constructing STT without ADC now works with a credentials object, both clients receive the exact supplied object, and the object wins over a (nonexistent) `credentials_file`.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
@@ -315,6 +315,14 @@ class STT(stt.STT):
         if self._location != "global":
             client_options = ClientOptions(api_endpoint=f"{self._location}-speech.googleapis.com")
         if self._credentials is not None:
+            if self._project_id is None:
+                # in-memory credentials (e.g. Workload Identity Federation) may
+                # carry the project directly; resolve it here so _get_recognizer
+                # doesn't fall back to Application Default Credentials, which
+                # such setups typically don't have
+                self._project_id = getattr(self._credentials, "project_id", None) or getattr(
+                    self._credentials, "quota_project_id", None
+                )
             client = client_cls(credentials=self._credentials, client_options=client_options)
         elif is_given(self._credentials_info):
             client = client_cls.from_service_account_info(
@@ -345,7 +353,16 @@ class STT(stt.STT):
         except AttributeError:
             from google.auth import default as ga_default
 
-            _, project_id = ga_default()
+            try:
+                _, project_id = ga_default()
+            except DefaultCredentialsError as e:
+                raise APIConnectionError(
+                    "google stt: could not determine the GCP project id: the supplied "
+                    "credentials expose no project_id/quota_project_id and Application "
+                    "Default Credentials are unavailable. Pass credentials that carry a "
+                    "project (e.g. with a quota_project_id) or use credentials_info / "
+                    "credentials_file."
+                ) from e
         return f"projects/{project_id}/locations/{self._location}/recognizers/_"
 
     def _sanitize_options(self, *, language: NotGivenOr[str] = NOT_GIVEN) -> STTOptions:
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
@@ -27,6 +27,7 @@ from typing import cast, get_args
 from grpc.aio import StreamStreamCall
 
 import google.auth
+import google.auth.credentials
 from google.api_core.client_options import ClientOptions
 from google.api_core.exceptions import DeadlineExceeded, GoogleAPICallError
 from google.auth import default as gauth_default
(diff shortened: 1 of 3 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -239,9 +246,14 @@ def __init__(
...
         self._credentials_file = credentials_file
+        self._credentials = credentials
         self._project_id: str | None = None
 
-        if not is_given(credentials_file) and not is_given(credentials_info):
+        if (
+            credentials is None
+            and not is_given(credentials_file)
+            and not is_given(credentials_info)
+        ):
```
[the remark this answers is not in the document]
biztex: ﻿Addressed in aed08c979. When a credentials object is supplied, the recognizer project is now resolved from the object itself (`project_id` / `quota_project_id`) at client creation, so WIF-style setups never touch ADC; and if the project still cannot be determined, `_get_recognizer` raises a clear could-not-determine-project error pointing at the fix, instead of leaking a confusing default-credentials failure at transcription time. Two new tests pin both paths (project-carrying credentials resolve the recognizer; no project + no ADC produces the explicit error).

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
```diff
@@ -158,6 +159,7 @@ def __init__(
         ] = NOT_GIVEN,
         credentials_info: NotGivenOr[dict] = NOT_GIVEN,
         credentials_file: NotGivenOr[str] = NOT_GIVEN,
+        credentials: google.auth.credentials.Credentials | None = None,
```
davidzhao: let's keep the `NotGivenOr[]` pattern here

livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -85,6 +86,7 @@ def __init__(
         audio_encoding: texttospeech.AudioEncoding = texttospeech.AudioEncoding.PCM,  # type: ignore
         credentials_info: NotGivenOr[dict] = NOT_GIVEN,
         credentials_file: NotGivenOr[str] = NOT_GIVEN,
+        credentials: google.auth.credentials.Credentials | None = None,
```
davidzhao: same here

davidzhao (commented):
lg, just one nit

livekit-plugins/livekit-plugins-google/livekit/plugins/google/stt.py
biztex: Done in f45bd7b — `NotGivenOr[google.auth.credentials.Credentials]` with `NOT_GIVEN`, matching `credentials_info`/`credentials_file` next to it. Call sites use `is_given()` and the docstrings were updated.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
biztex: Same change applied here in f45bd7b.

tinalenguyen (approved):

PR #6619, opened 2026-07-30
fix(aws): strip temperature and retry when Bedrock rejects it

### LINKED ISSUE #6581: Bedrock rejects 'temperature' params on newer models
### Bug Description

When using 'livekit.pluging.aws.LLm' with newer Antropic Models on AWS Bedrock, returns a 'ValidationException'.

Bedrock discontinued the temperature parameter for some models

### Expected Behavior

To treat validation. The AWS Bedrock plugin should not send the temperature parameter for models that reject them

### Reproduction Steps

```bash
1. 
2.
3.
...
```

### Operating System

Windows 11

### Models Used

_No response_

### Package Versions

```bash
livekit-agents==1.6.6
livekit-plugins-aws==1.6.6
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

biztex:
﻿Fixes #6581

## Problem

Bedrock has discontinued the `temperature` inference parameter for some newer models; sending it fails the whole `ConverseStream` request with a `ValidationException`. The plugin wrapped that as a retryable `APIConnectionError` and retried with **identical** parameters — every attempt failing the same way, so sessions configured with a `temperature` simply cannot use those models.

## Change

The exact set of rejecting models is a moving target (and differs per region), so instead of maintaining a model list, the stream now reacts to the provider's own signal:

- On a `ValidationException` whose message names `temperature` while the request carried one, the parameter is stripped from the request options and the error is re-raised as retryable — the framework's existing retry immediately re-runs the request without it.
- The rejection is remembered on the `LLM` instance (`_temperature_rejected`), so every subsequent `chat()` omits `temperature` outright — the fallback costs one extra round-trip once per LLM instance, not per turn.
- A warning is logged once with the model id and the provider message. Validation errors about anything else, and models that accept `temperature`, behave exactly as before.

## Tests

New hermetic `tests/test_aws_bedrock_temperature.py` (unit-marked, `_run` stubbed, no network): strip-and-flag on a matching ValidationException, no action on unrelated validation errors / when temperature was not sent / on non-ClientError exceptions, and `chat()` omitting the parameter after a rejection.

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
@@ -20,6 +20,7 @@ from typing import TYPE_CHECKING, Any
 
 from aiobotocore.session import AioSession  # type: ignore
 from botocore.config import Config  # type: ignore
+from botocore.exceptions import ClientError  # type: ignore
 
 from livekit.agents import APIConnectionError, APIStatusError, llm
 from livekit.agents.llm import ChatContext, FunctionToolCall, ToolChoice
@@ -119,6 +120,9 @@ class LLM(llm.LLM):
             cache_system=cache_system,
             cache_tools=cache_tools,
         )
+        # set once Bedrock rejects the temperature parameter for this model
+        # (newer models discontinue it); later requests then omit it entirely
+        self._temperature_rejected = False
 
     @property
     def model(self) -> str:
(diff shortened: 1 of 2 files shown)


tinalenguyen (commented):
hi, thanks for the pr!

i think it makes more sense to guard against sending `temperature` to models that don't accept it rather than reacting to an error

PR #5519, opened 2026-04-22
fix: don't swallow _ExitCli during shutdown

lawrence3699:
Closes #4664

## What was happening
`_ExitCli` is raised from the CLI signal handlers to stop the worker. Because it inherited from `Exception`, broad `except Exception` blocks could catch it before the shutdown path reached `drain()` / `aclose()`.

`SupervisedProc._memory_monitor_task()` is one concrete path: if `SIGTERM` lands while `psutil.Process(...)` is running, the task logs and swallows `_ExitCli` instead of letting the worker drain.

## What changed
- make `_ExitCli` inherit from `BaseException`
- add a regression test covering `_memory_monitor_task()`

## Why this is correct
`_ExitCli` is internal control flow used only for process shutdown, and the CLI already catches it explicitly at the top level. Making it bypass generic `except Exception` handlers lets shutdown propagate without changing the existing explicit `_ExitCli` handling.

## Validation
- `uv run pytest tests/test_drain_timeout.py -q`
- `uv run ruff check livekit-agents/livekit/agents/cli/cli.py tests/test_drain_timeout.py`

PR #6624, opened 2026-07-30
fix(cli): derive _ExitCli from SystemExit so loop callbacks cannot swallow it

### LINKED ISSUE #5856: _ExitCli still swallowed by asyncio Handle._run after PR #5519 (callback-frame race)
PR #5519 (released in v1.5.7) changed `_ExitCli(Exception)` → `_ExitCli(BaseException)` so generic `except Exception:` blocks no longer swallow it during shutdown. Thanks for that fix — it closes one swallow path. There's a second swallow path it doesn't address that I wanted to flag.

## The remaining swallow path

`_handle_exit` in `livekit/agents/cli/cli.py` raises `_ExitCli()` synchronously from inside a Python signal handler. Python's signal-via-wakeup-fd mechanism on Linux delivers SIGTERM into the asyncio event loop's next iteration — but the exact frame the signal-handler runs in is whatever the loop happened to be executing at that microsecond. When SIGTERM lands while the loop is mid-callback (e.g., inside `_set_result_unless_cancelled` resolving a Future), the `_ExitCli` raise happens inside an `asyncio.events.Handle._run` frame.

CPython's `Handle._run`:

```python
def _run(self):
    try:
        self._context.run(self._callback, *self._args)
    except (SystemExit, KeyboardInterrupt):
        raise                       # ← propagates
    except BaseException as exc:
        ...
        self._loop.call_exception_handler(context)   # ← swallows
```

After PR #5519, `_ExitCli` is `BaseException`, which matches the second `except`. So it gets logged as `"Exception in callback ... _handle_exit raise _ExitCli"` and the loop **keeps running**. The drain path is never reached.

This is the exact production SEV I documented in [prodigal-tech/agent-orchestrator#330](https://github.com/prodigal-tech/agent-orchestrator/issues/330) — a pod kept advertising itself to LiveKit Cloud through its full 15-minute `terminationGracePeriodSeconds`, accepted 29 new dispatches during that window, then got SIGKILL'd at grace expiry. We were on 1.3.12 at the time, but the bug is identical on 1.5.7 — the BaseException change doesn't help here.

## Empirical reproduction

```python
import asyncio

class ExitCliBase(BaseException):
    \"\"\"Mirrors v1.5.7 _ExitCli.\"\"\"
    pass

class ExitCliKbd(KeyboardInterrupt):
    \"\"\"Alternative.\"\"\"
    pass

def trial(label, cls):
    loop = asyncio.new_event_loop()
    main = loop.create_task(asyncio.sleep(10))
    loop.call_later(0.05, lambda: (_ for _ in ()).throw(cls()))
    try:
        loop.run_until_complete(main)
        print(f\"{label}: SWALLOWED — drain never reached\")
    except cls:
        print(f\"{label}: PROPAGATED → drain runs\")
    finally:
        loop.close()

trial(\"_ExitCli(BaseException)\", ExitCliBase)   # SWALLOWED
trial(\"_ExitCli(KeyboardInterrupt)\", ExitCliKbd)  # PROPAGATED
```

Output on Python 3.11:
```
Exception in callback ... ExitCliBase
_ExitCli(BaseException): SWALLOWED — drain never reached
_ExitCli(KeyboardInterrupt): PROPAGATED → drain runs
```

## Suggested fixes (one of)

1. **Change `_ExitCli` to inherit from `KeyboardInterrupt`** instead of `BaseException`. `Handle._run` re-raises `KeyboardInterrupt` explicitly; everything else still works (it's still `BaseException` via inheritance, still bypasses `except Exception:`). Smallest change.
2. **Stop using exceptions for signal-driven shutdown.** Use `loop.add_signal_handler(SIGTERM, callback)` instead of `signal.signal(...)`. The callback can do `main_task.cancel()` or set a `stopping_event` that the main coroutine awaits. This avoids raising from arbitrary frames entirely.

Option 2 is more invasive but architecturally cleaner. Option 1 is a one-character change.

## Workaround we're carrying

For now we monkey-patch `_cli._ExitCli` to a `KeyboardInterrupt` subclass at startup (in our `main.py` before `agents.cli.run_app`). Code: [agent_orchestrator/patches/livekit_exit_cli.py on this branch](https://github.com/prodigal-tech/agent-orchestrator/blob/feature/livekit-agents-1.4.6-upgrade/agent_orchestrator/patches/livekit_exit_cli.py). The harness that validated end-to-end recovery is under `tests/manual/network_blip/docker/`.

Happy to send a PR if option 1 is acceptable.

biztex:
Fixes livekit/agents#5856

## Problem

`_handle_exit` raises `_ExitCli` synchronously from a signal handler, and the frame the raise lands in is whatever the event loop happened to be executing at that moment. When SIGTERM arrives while the loop is mid-callback, the raise happens inside asyncio's `Handle._run` — which re-raises only `SystemExit`/`KeyboardInterrupt` and reports every other `BaseException` to the loop exception handler. The exit is logged ("Exception in callback ... \_handle_exit") and the loop keeps running: the drain/shutdown path is never reached. [#5519](<https://github.com/livekit/agents/issues/5519>) (`Exception` → `BaseException`) closed the generic `except Exception:` swallow path; this closes the remaining callback-frame path identified in the issue.

## Change

`_ExitCli` now derives from `SystemExit` (in both `cli.py` and `_legacy.py`) — the exception class the loop machinery re-raises *by design*, which is exactly the semantic `_ExitCli` wants. Safety of the switch:

* `SystemExit` is still a `BaseException`, so every existing handler that caught `_ExitCli` continues to catch it; the `except _ExitCli:` sites are unchanged.
* There are no `except SystemExit` sites anywhere in `livekit-agents` that could change behavior (verified by search).

## Tests

New hermetic `tests/test_exit_cli_propagation.py` driving the exact failure: a raise from inside a loop callback must propagate out of `run_forever` rather than reach the loop exception handler. Verified red on the old class (the loop swallows the exit and runs until a safety stop) and green with the fix; also pins the subclass relationship for both CLI variants and that `except _ExitCli` still catches.

## Also fixes

* livekit/agents#6724 — SIGTERM ignored when `request_fnc` blocks. Same cause one boundary over: the raise lands in a coroutine, and `Task.__step` re-raises only `SystemExit`/`KeyboardInterrupt` too, so anything else is stored on the task and the loop keeps running. Covered by `test_propagates_out_of_a_task`.
* Under uvloop this stops being a race and becomes deterministic: signal delivery for `signal.signal()` handlers is routed through `Loop._invoke_signals` as a scheduled callback, so the raise *always* lands in `Handle._run` (`cbhandles.pyx`, same `(KeyboardInterrupt, SystemExit)` rule). A worker idle on I/O — the normal state at deploy time — never drains. Reported in-thread against a 3-worker production deployment.

### DIFF
--- a/tests/test_exit_cli_propagation.py
+++ b/tests/test_exit_cli_propagation.py
@@ -55,3 +55,33 @@ class TestExitCliPropagation:
         except _ExitCli:
             caught = True
         assert caught
+
+    def test_propagates_out_of_a_task(self) -> None:
+        # #6724: SIGTERM landing while the main thread is inside a job-request task
+        # (a request_fnc doing blocking work) puts the raise in a coroutine, not a
+        # callback. Task.__step re-raises only SystemExit/KeyboardInterrupt out of the
+        # loop; any other BaseException is stored on the task ("Task exception was never
+        # retrieved") and the loop keeps running, so the worker never drains.
+        loop = asyncio.new_event_loop()
+        try:
+
+            async def _job_request_task() -> None:
+                raise _ExitCli()
+
+            async def _main() -> None:
+                await asyncio.sleep(30)  # the worker's main task, still running
+
+            main_task = loop.create_task(_main())
+            loop.call_soon(lambda: loop.create_task(_job_request_task()))
+
+            swallowed: list[dict] = []
+            loop.set_exception_handler(lambda _loop, ctx: swallowed.append(ctx))
+            loop.call_later(0.5, loop.stop)  # safety stop for the broken case
+
+            with pytest.raises(_ExitCli):
+                loop.run_until_complete(main_task)
+
+            assert not swallowed, "exit was left on the task instead of reaching the loop"
+        finally:
+            main_task.cancel()
+            loop.close()
--- a/tests/test_exit_cli_propagation.py
+++ b/tests/test_exit_cli_propagation.py
@@ -1,0 +1,57 @@
+"""Regression tests for #5856: _ExitCli must escape asyncio's Handle._run.
+
+``_handle_exit`` raises ``_ExitCli`` synchronously from a signal handler; the
+frame that raise lands in is whatever the event loop happened to be executing.
+When that frame is a loop callback, CPython's ``Handle._run`` re-raises only
+``SystemExit``/``KeyboardInterrupt`` and reports every other ``BaseException``
+to the loop exception handler — logging it and leaving the loop running, so the
+CLI's drain/shutdown path is never reached.
+"""
+
+import asyncio
+
+import pytest
+
+from livekit.agents.cli import _legacy
+from livekit.agents.cli.cli import _ExitCli
+
+pytestmark = pytest.mark.unit
+
+
+class TestExitCliPropagation:
+    def test_is_systemexit(self) -> None:
+        # the only exception types Handle._run re-raises are SystemExit and
+        # KeyboardInterrupt; _ExitCli must be one of them to survive a raise
+        # that lands inside a loop callback
+        assert issubclass(_ExitCli, SystemExit)
+        assert issubclass(_legacy._ExitCli, SystemExit)
+
+    def test_propagates_out_of_a_loop_callback(self) -> None:
+        # simulates SIGTERM landing while the loop is mid-callback: the raise
+        # happens inside Handle._run. Before the fix the loop swallowed it and
+        # kept running until the 0.5s stop; now run_forever raises immediately.
+        loop = asyncio.new_event_loop()
+        try:
+
+            def _raise_exit() -> None:
+                raise _ExitCli()
+
+            swallowed: list[dict] = []
+            loop.set_exception_handler(lambda _loop, ctx: swallowed.append(ctx))
+            loop.call_soon(_raise_exit)
+            loop.call_later(0.5, loop.stop)  # safety stop for the broken case
+
+            with pytest.raises(_ExitCli):
+                loop.run_forever()
+
+            assert not swallowed, "exit was reported to the exception handler instead of raised"
+        finally:
+            loop.close()
+
+    def test_still_caught_by_except_exit_cli(self) -> None:
+        # _run_worker's `except _ExitCli:` handling must keep working
+        try:
+            raise _ExitCli()
+        except _ExitCli:
+            caught = True
+        assert caught
--- a/livekit-agents/livekit/agents/cli/_legacy.py
+++ b/livekit-agents/livekit/agents/cli/_legacy.py
@@ -75,7 +75,9 @@ class _ToggleMode(Exception):
     pass
 
 
-class _ExitCli(BaseException):
+class _ExitCli(SystemExit):
+    # SystemExit so asyncio's Handle._run re-raises it instead of swallowing it
+    # when the signal-handler raise lands inside a loop callback (#5856).
     pass
 
 
--- a/livekit-agents/livekit/agents/cli/cli.py
+++ b/livekit-agents/livekit/agents/cli/cli.py
@@ -26,7 +26,13 @@ HANDLED_SIGNALS = (
 )
 
 
-class _ExitCli(BaseException):
+class _ExitCli(SystemExit):
+    # Derives from SystemExit, not BaseException: _handle_exit raises this from a
+    # signal handler, and the frame the raise lands in is whatever the event loop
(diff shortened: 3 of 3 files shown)


chenghao-mou:
Thanks for the PR! Can you also add a test case for #6724, I believe they can be fixed by the same change.

biztex:
Added in 327cd1b — and confirmed: same cause, one asyncio boundary over.

#5856 is the raise landing in a loop callback, where `Handle._run` re-raises only `SystemExit`/`KeyboardInterrupt`. #6724 is it landing in a coroutine instead, where `Task.__step` applies the same rule — anything else is stored on the task ("Task exception was never retrieved") and the loop keeps running, so the worker never drains. Deriving from `SystemExit` satisfies both.

`test_propagates_out_of_a_task` drives the reported shape: a job-request task raises while the worker's main task is still awaiting, and the exit has to reach `run_until_complete` rather than sit on the task. It fails with `_ExitCli(BaseException)` and passes with the fix.

SFGMorales:
We hit this exact issue in production today (2026-08-06) — two consecutive deploys where all 3 of our worker processes failed to respond to SIGTERM from docker compose up -d recreation, forcing Docker to wait the full stop_grace_period (630s) and SIGKILL them.

Root cause confirmed: with uvloop active (asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())), signal delivery for signal.signal()-registered handlers always goes through Loop._invoke_signals as a scheduled callback, which uvloop's Handle._run wraps in a try/except that only re-raises SystemExit/KeyboardInterrupt — so _ExitCli (even as a BaseException per #5519) is swallowed 100% of the time when the loop is idle waiting on I/O, which is the typical state at deploy time.

Deterministic repro (no active job, no Docker needed): start a worker (... dev), wait for "registered worker", kill -TERM <top-level CLI pid>. Log shows Exception in callback Loop._read_from_self → _invoke_signals → _handle_exit → _ExitCli, and the process never exits on its own — confirmed hung 25s+ later, no drain/aclose/process exiting in the logs, only recoverable via SIGKILL.

Production impact: two deploys today, ~15min and ~12min of total capacity loss respectively (all 3 workers down simultaneously, required manual docker start to recover — docker compose up -d alone doesn't retry).

This fix (_ExitCli → SystemExit) looks correct — confirmed against uvloop's source (cbhandles.pyx, Handle._run only excepts (KeyboardInterrupt, SystemExit)) that it resolves this specific case too. Happy to help validate against our repro once it's ready to test — this affects a fairly large production deployment (~6k calls/day across 3 worker processes).

biztex:
Thanks, that's a really useful data point — the uvloop detail explains why you get it every time rather than occasionally. On stock asyncio it depends where the raise happens to land, so it's a race; with uvloop routing `signal.signal` handlers through `_invoke_signals`, the raise is always inside a callback, and an idle worker is exactly the state at deploy time. That matches your 3/3.

The branch is pushed and CI is green, so it's ready whenever you want to try it. If you can run it against your repro that would be great — a confirmation on a real deployment is worth more than my unit tests here.

I didn't add a uvloop test: it isn't a dependency of the repo and I'm on Windows so I can't run it locally, and I'd rather not ship a test I can't verify. The two tests cover both boundaries on stock asyncio (`Handle._run` and `Task.__step`), and uvloop applies the same rule in cbhandles.pyx as you found.

Also noted your stop_grace_period point — 630s of a deploy stuck per incident is a nasty way to find this.

tests/test_exit_cli_propagation.py
```diff
@@ -0,0 +1,87 @@
...
+            loop.call_soon(lambda: loop.create_task(_job_request_task()))
+
+            swallowed: list[dict] = []
+            loop.set_exception_handler(lambda _loop, ctx: swallowed.append(ctx))
+            loop.call_later(0.5, loop.stop)  # safety stop for the broken case
+
+            with pytest.raises(_ExitCli):
+                loop.run_until_complete(main_task)
+
+            assert not swallowed, "exit was left on the task instead of reaching the loop"
```
chenghao-mou, line 84: This test leaves both tasks improperly finalized, and `main_task` is cancelled without being awaited, so garbage collection reports `Task exception was never retrieved` and `Task was destroyed but it is pending!`. 

Can we retain both tasks, cancel any pending task, and gather them with `return_exceptions=True` before closing the loop? we also don't need the assertion if we already have the raises check.

davidzhao:
closing in favor of #7034

PR #6625, opened 2026-07-30
fix(llm): tolerate list-shaped streaming delta content in strip_thinking_tokens

### LINKED ISSUE #6323: strip_thinking_tokens crashes on list-shaped streaming delta content (AttributeError: 'list' object has no attribute 'find')
**Versions:** livekit-agents 1.6.3 (the code is unchanged on current `main`), livekit-plugins-openai, Python 3.12.

**Setup:** `openai.LLM` plugin pointed at an OpenAI-compatible provider (Mistral, `base_url=https://api.mistral.ai/v1`), wrapped in `llm.FallbackAdapter`.

**What happens:** the OpenAI-compatible streaming schema allows `delta.content` to be either a `str` or a list of typed content parts (`[{"type": "text", "text": "…"}]`). Mistral has started emitting the list form on some streamed chunks. `llm/utils.py::strip_thinking_tokens` assumes `str | None` and calls `content.find(...)`, raising `AttributeError: 'list' object has no attribute 'find'`. The plugin wraps this as a retryable `APIConnectionError`; under a `FallbackAdapter` the failure lands mid-stream ("failed after sending chunk, skip retrying"), so the adapter silently moves the whole session to the fallback leg — a hard-to-diagnose degradation for voice agents.

**Repro (no network):**
```python
import asyncio
from livekit.agents.llm.utils import strip_thinking_tokens
strip_thinking_tokens([{"type": "text", "text": "Hallo"}], asyncio.Event())
# AttributeError: 'list' object has no attribute 'find'
```

**Suggested fix:** flatten list-shaped content to its concatenated text parts before the tag scan — see the linked PR.

biztex:
Fixes livekit/agents#6323

## Problem

`strip_thinking_tokens` assumes `delta.content` is `str | None`. The OpenAI-compatible ecosystem allows it to be a **list of typed content parts** (`[{"type": "text", "text": ...}]`) — Mistral emits this shape on some streamed chunks — and the openai SDK constructs stream models without validation (`construct()`), so when `openai.LLM` points at such a provider via `base_url`, the raw list reaches the filter and crashes the session. Wrapped as a retryable `APIConnectionError` mid-stream, a `FallbackAdapter` then silently fails the whole session over to its fallback leg — a hard-to-diagnose degradation.

Re: the open question on the issue about the source of the claim ("the OpenAI streaming spec only documents string"): correct for OpenAI's own service — the authority here is **Mistral's** chat-completions API, whose assistant/delta message `content` is typed `string | ContentChunk[]` in their reference and SDK. Since the plugin explicitly supports OpenAI-*compatible* providers through `base_url`, and the SDK's unvalidated stream models pass whatever the provider sends, the list shape is reachable in practice (that's how the original reporter hit it in production). Answered in more detail on the issue thread.

## Change

Flatten list-shaped content to its concatenated text parts before the tag scan — dict parts (`{"text": ...}`), plain strings, and typed objects with a `.text` attribute. Unknown content types are dropped with a warning instead of crashing. String/None behavior is byte-identical.

## Tests

Five new cases in `tests/test_llm_utils.py`: the issue's exact repro (`[{"type": "text", "text": "Hallo"}]`), mixed part lists, thinking-tag stripping across list-shaped chunks, non-text parts ignored, and typed objects with `.text`.

## Relationship to [#6324](<https://github.com/livekit/agents/issues/6324>)

This supersedes the stalled [#6324](<https://github.com/livekit/agents/issues/6324>) (same diagnosis, by @KSerProject — credit where due): that PR has had merge conflicts and an unsigned CLA for four weeks. Happy to close this one if the original author returns.

### DIFF
--- a/livekit-agents/livekit/agents/llm/utils.py
+++ b/livekit-agents/livekit/agents/llm/utils.py
@@ -628,9 +628,37 @@ def _partial_marker_length(content: str, markers: tuple[str, ...]) -> int:
     )
 
 
+def _flatten_delta_content(content: Any) -> str | None:
+    """Flatten list-shaped streaming delta content to its concatenated text.
+
+    The OpenAI-compatible ecosystem allows ``delta.content`` to be a list of
+    typed content parts (``[{"type": "text", "text": ...}]``) instead of a
+    plain string — Mistral emits this shape on some streamed chunks — and the
+    openai SDK constructs stream models without validation, so the list
+    reaches us as-is when the plugin points at such a provider (#6323).
+    """
+    if content is None or isinstance(content, str):
+        return content
+    if isinstance(content, list):
+        parts: list[str] = []
+        for part in content:
+            if isinstance(part, str):
+                parts.append(part)
+                continue
+            text = part.get("text") if isinstance(part, dict) else getattr(part, "text", None)
+            if isinstance(text, str):
+                parts.append(text)
+        return "".join(parts)
+    logger.warning(
+        "unexpected streaming delta content type %s; dropping the chunk", type(content).__name__
+    )
+    return None
+
+
 def strip_thinking_tokens(
-    content: str | None, state: ThinkingTokenFilter, *, final: bool = False
+    content: str | list[Any] | None, state: ThinkingTokenFilter, *, final: bool = False
 ) -> str | None:
+    content = _flatten_delta_content(content)
     if content is not None:
         state._buffer += content
 
--- a/livekit-agents/livekit/agents/llm/utils.py
+++ b/livekit-agents/livekit/agents/llm/utils.py
@@ -648,7 +648,9 @@ def _flatten_delta_content(content: Any) -> str | None:
             text = part.get("text") if isinstance(part, dict) else getattr(part, "text", None)
             if isinstance(text, str):
                 parts.append(text)
-        return "".join(parts)
+        # no text part at all carries no content, unlike a part that carries an
+        # empty string - that one is kept, like a plain "" delta
+        return "".join(parts) if parts else None
     logger.warning(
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -783,9 +628,37 @@ def _partial_marker_length(content: str, markers: tuple[str, ...]) -> int:
...
+    if isinstance(content, list):
+        parts: list[str] = []
+        for part in content:
+            if isinstance(part, str):
+                parts.append(part)
+                continue
+            text = part.get("text") if isinstance(part, dict) else getattr(part, "text", None)
+            if isinstance(text, str):
+                parts.append(text)
+        return "".join(parts)
```
chenghao-mou: nit: maybe we can return None here if parts is empty?

chenghao-mou (approved):
lgtm, thanks for the PR. One small nit.

livekit-agents/livekit/agents/llm/utils.py
biztex: Thanks for the review! Done in 56d7e80 — a list with nothing textual in it now reports no content, which also lines it up with the `None` and unexpected-type paths above.

I used `if parts` rather than `or None` so that an empty *text* part is still distinguishable: `[{"type": "text", "text": ""}]` keeps returning `""`, the same as a plain `""` delta, while `[]` and `[{"type": "image_url", ...}]` return `None`. That preserves the `content == ""` branch at the end of `strip_thinking_tokens`, which exists to pass a genuinely empty delta through.

No behaviour change for the emitted chunks either way — `_parse_choice` gates on `if not delta.content and not delta_extra`, so both were already falsy. Added tests for both shapes.

PR #6629, opened 2026-07-30
fix(inference): don't claim word alignment for models that send none

russellmartin-livekit:
## The bug

`inference.STT` hardcoded `aligned_transcript="word"` for every model:

```python
capabilities=stt.STTCapabilities(
    streaming=True,
    interim_results=True,
    diarization=diarization_enabled,
    aligned_transcript="word",   # <- regardless of model
    ...
)
```

That flag is not cosmetic. It is the `can_gatekeep` test in
`AgentActivity._resolve_interruption_detection`:

```python
can_gatekeep = (
    self.stt is not None
    and self.stt.capabilities.aligned_transcript
    and self.stt.capabilities.streaming
)
```

When `can_gatekeep` is true and a VAD is present, adaptive interruption is
eligible, and it is enabled by default on LiveKit Cloud (`is_hosted()`) and in
dev mode (`lk agent dev`, `console`). Enabling it flips the fast path off: once
the `backchannel_boundary` window expires, `_disable_vad_interruption_soon` sets
`_interruption_by_audio_activity_enabled = False`, handing barge-in from VAD to
the adaptive detector. `backchannel_boundary` defaults to `(1.0, 1.0)`.

So for a model with no word timings we traded a VAD barge-in for a transcript
gate that cannot fire, and the user pays about a second of extra barge-in
latency. `_should_hold_stt_event` requires `start_time > 0`, and the gateway's
Ink-2 translator pins `Start` to the audio-time offset (0 until a reconnect) and
`Words` to `[]`, so the gate is inert. After a reconnect it does engage, but on
`end_time - start_time` derived from billed audio bytes rather than the
utterance span, which is worse than not engaging.

Cartesia Ink-2 is the case that surfaced this. Its Turns API sends only
`{type, transcript, turn_id, request_id}` per turn — verified against the live
service on both the version the plugin pins and the one the gateway pins, and
neither `add_timestamps` nor `timestamp_granularities[]` changes it. The
`livekit-plugins-cartesia` plugin reports `aligned_transcript=False` for the
same model, so the two integrations disagreed about identical audio, and only
the gateway path was slow. That matches the reports of Ink-2 barge-in being
roughly a second late through inference versus the plugin or raw VAD.

## The fix

Derive the capability per model, and recompute it in `update_options` alongside
`keyterms` and `chat_context` (it was previously never recomputed, so switching
models left a stale claim).

`False` for models known to send no word timings (`cartesia/ink-2`, `inworld/`,
`google/`), and for `"auto"`, where the provider is resolved server-side per
language and so cannot be known client-side. Erring toward `False` is the
cheaper mistake: it forgoes adaptive interruption, whereas a false `"word"`
disables a working VAD path in favour of one that cannot replace it.

Models that genuinely forward words (`cartesia/ink-whisper`, `deepgram/nova-3`,
`assemblyai/universal-streaming`, ...) keep `"word"` and are unaffected.

## Tests

`tests/test_inference_stt_aligned_transcript_claim.py`, with a
`TurnScriptedSTT` fake in `tests/fake_turn_stt.py` that replays a
turn-detecting provider's event order.

Three tests fail on `main` and pass here:

- `test_inference_and_plugin_agree_about_ink2` — both integrations report the
  same alignment for the same model
- `test_models_that_do_send_words_keep_their_claim` — the aligned providers are
  untouched
- `test_alignment_is_recomputed_when_the_model_changes` — `update_options`
  refreshes it

The rest document the mechanism, and hold before and after: that the capability
flag alone decides whether adaptive interruption runs, that the gateway's Ink-2
payload carries no words, that `_should_hold_stt_event` can't act on it, that
plain production is unaffected, and that `interruption.mode="vad"` remains a
working workaround on either integration.

## Workaround for anyone hitting this before it ships

```python
interruption=InterruptionOptions(mode="vad", min_duration=0.2)
```

`mode="vad"` short-circuits `_resolve_interruption_detection` before the
hosted/dev check, so it keeps the VAD path armed regardless of the STT.
`min_duration` then governs barge-in speed; the default is `0.5`. This does not
affect Ink-2's own turn detection, which continues to drive endpointing.

## Notes

- Complementary to #6599, which stopped interim transcripts from interrupting
  when a VAD is present. Different path, same symptom area.
- The prefix list is a client-side guess about server behaviour. If the gateway
  starts forwarding word timings for these providers, this needs updating —
  worth considering whether the gateway should advertise alignment instead.

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -260,18 +260,34 @@ def _supports_agent_context_carryover(model: NotGivenOr[STTModels | str]) -> boo
     return is_given(model) and isinstance(model, str) and model in _ASSEMBLYAI_CARRYOVER_MODELS
 
 
-# Models whose transcripts carry no word timings. Cartesia's Turns API sends only
-# {type, transcript, turn_id, request_id} per turn, verified against the live service.
-_UNALIGNED_MODEL_PREFIXES = (
-    "cartesia/ink-2",
-    "inworld/",
-    "google/",
+# Models verified to carry word timings. Unknown models stay disabled until verified.
+_WORD_ALIGNED_MODELS = frozenset(
+    {
+        "deepgram/nova-3",
+        "deepgram/nova-3-medical",
+        "deepgram/nova-2",
+        "deepgram/nova-2-medical",
+        "deepgram/nova-2-conversationalai",
+        "deepgram/nova-2-phonecall",
+        "deepgram/flux-general",
+        "deepgram/flux-general-en",
+        "deepgram/flux-general-multi",
+        "cartesia/ink-whisper",
+        "assemblyai/universal-streaming",
+        "assemblyai/universal-streaming-multilingual",
+        "assemblyai/u3-rt-pro",
+        "assemblyai/universal-3-5-pro",
+        "elevenlabs/scribe_v2_realtime",
+        "xai/stt-1",
+        "speechmatics/enhanced",
+        "speechmatics/standard",
+    }
 )
 
 
 def _aligned_transcript_for_model(
     model: NotGivenOr[STTModels | str],
-) -> Literal["word"] | bool:
+) -> Literal["word", False]:
     """Word-level alignment, which adaptive interruption relies on to gatekeep transcripts.
 
     False when the model sends no word timings, and for "auto", where the provider is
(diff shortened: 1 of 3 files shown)


chenghao-mou (approved):

livekit-agents/livekit/agents/inference/stt.py
```diff
@@ -579,7 +605,7 @@ def __init__(
                 streaming=True,
                 interim_results=True,
                 diarization=diarization_enabled,
-                aligned_transcript="word",
+                aligned_transcript=_aligned_transcript_for_model(model),
```
chenghao-mou, line 622: nit: we should check the fallback models too

PR #6631, opened 2026-07-30
(expressive mode): add fish audio

cshape:
rebase of #6528 onto main (the original was auto-closed when tina/refactor-expressive-mode merged in #6406).

adds fish audio (s2 models) to expressive mode, same shape as inworld/xai/cartesia:

- emotion tags (discrete vocabulary: regretful, hopeful, happy, excited, curious, surprised, sad, empathetic, sarcastic) -> `[very EMOTION]`
- breaks -> `[break]` / `[long-break]` by duration
- emphasis, the only wrapping prosody -> prefix form: `are you [emphasis] sure`
- sounds: `[laughing]`, `[chuckling]`, `[clear throat]`, governed by the `laughing` / `reflex_sounds` steering fields; disabled sounds are dropped from the advertised vocabulary and examples like the other providers
- disfluent few-shot examples render only while `disfluencies` is enabled, so examples never contradict the delivery guidelines

no `_MAX_INPUT_LEN` entry on purpose: fish's markers are sentence-scoped (every sentence carries its own emotion tag), so we stick with normal per-sentence synthesis instead of batching longer expressive chunks — no steering is lost and time-to-first-audio stays low.

adapted to the post-#6406 architecture during the rebase: fish registers in the new `_PROVIDER_MARKUP` (plain tag-name list), and transcript tests use `split_all_markup` — bracket cues are handled at the aligned-transcript source by `drop_bracket_cues`, matching the other providers.

verified live through the inference gateway (`fishaudio/s2.1-pro` + gemma-4-31b-it) and with offline sweeps across several llms checking tag placement, conversion, transcript stripping, and steering filters. 1624 unit tests + strict mypy green on the rebased stack.

### DIFF
--- a/livekit-agents/livekit/agents/tts/_provider_format.py
+++ b/livekit-agents/livekit/agents/tts/_provider_format.py
@@ -11,6 +11,7 @@ Provider docs:
 - Inworld: https://docs.inworld.ai/tts/best-practices/prompting-for-tts-2
 - xAI: https://docs.x.ai/developers/model-capabilities/audio/text-to-speech
 - xAI: https://docs.x.ai/developers/model-capabilities/audio/voice
+- Fish Audio: https://docs.fish.audio/developer-guide/core-features/emotions
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


tinalenguyen (approved):
thank you!

PR #6633, opened 2026-07-30
fix: account for STT fallback transcript alignment

chenghao-mou:
Inference STT can switch server-side to a fallback whose transcripts do not carry word timings. Only claim word alignment when the primary and every fallback support it, keeping adaptive interruption from relying on unavailable timing data.

Fixes AGT-3210

### DIFF
--- a/livekit-agents/livekit/agents/inference/stt.py
+++ b/livekit-agents/livekit/agents/inference/stt.py
@@ -365,6 +365,20 @@ def _normalize_fallback(
     return [_make_fallback(fallback)]
 
 
+def _aligned_transcript_for_models(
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
lg! what happens when it's in automatic fallback mode? where we don't specify the model but inference picks one?

chenghao-mou:
> lg! what happens when it's in automatic fallback mode? where we don't specify the model but inference picks one?

Good question. I believe a stt model's fallback only contains its deployments, not other models or vendors. cc @livekit/inference-gateway

chenghao-mou:
^ confirmed correct on Slack

PR #6634, opened 2026-07-31
plugins/phonic: add configs_for_tools for per-tool behavior overrides

arunwpm-work:
## What

Adds `configs_for_tools` to the Phonic realtime model — per-tool behavior overrides forwarded to Phonic's tool config. One entry per tool (keyed by `name`); each may set `require_speech_before_tool_call`, `forbid_speech_after_tool_call`, and `forbid_tool_call_after_speech`. Omitted fields fall back to the plugin defaults; tools with no entry are unchanged. `wait_for_speech_before_tool_call` (on) and `allow_tool_chaining` (off) remain fixed by the plugin.

This generalizes the existing `forbid_speech_after_tool_call: list[str]` option into a single per-tool config object, and adds `forbid_tool_call_after_speech` (drop a tool call if the agent already spoke that turn). See the plugin README for the field reference.


livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -76,7 +86,7 @@ class _RealtimeOptions:
     no_input_poke_text: NotGivenOr[str]
     no_input_end_conversation_sec: NotGivenOr[float]
     additional_params: NotGivenOr[dict[str, typing.Any]]
-    forbid_speech_after_tool_call: NotGivenOr[list[str]]
```
tinalenguyen, line 79: removing this would break existing code, let's throw a warning to switch and pass it into PhonicToolConfig

arunwpm-work: should be fixed now! thanks for flagging :)

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -56,6 +56,16 @@
 TOOL_CALL_OUTPUT_TIMEOUT_MS = 60000
 
 
+class PhonicToolConfig(TypedDict, total=False):
```
tinalenguyen, line 59: should we export this in the realtime init file?

tinalenguyen (approved):
lgtm, left a small comment!

livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
arunwpm-work: yes! added it in

PR #6642, opened 2026-07-31
fix(openai realtime): treat create_response=False as client-side turn taking

### LINKED ISSUE #6635: RealtimeModel forces `allow_interruptions=True` / `response.cancel` even when OpenAI `interrupt_response=False` (server VAD + manual responses)
### Bug Description

We have configured `gpt-realtime` to disallow user interruptions (`allow_interruptions=False`), but LiveKit forces all detected user speech to send a `response.cancel` because it forces `interrupt_response=True` / `allow_interruptions=True` for all realtime sessions that use server-side turn detection.

This is highly problematic: if agent speech ever overlaps with user speech, the agent is cut off. We use a turn-taking-by-button-click model, so in those situations the agent appears unresponsive when in fact LiveKit has instructed it to stop — which leads to user bug reports.

**Significance:** We will be required to maintain our own fork of the LiveKit codebase to fix this source of bug reports. Although it may seem niche, it is quite important for us to resolve. The only reason we did not notice this sooner is that, given our turn-taking model, agent and user speech rarely overlap (we mute the user microphone during the agent turn). We previously did not have verbose logging of `response.cancel`, so the root cause was tricky to spot. Even with microphone muting, the issue cannot be fully worked around — for example, we cannot reliably mute around speech triggered by function-call output.

I am not fluent in the other realtime models livekit supports, but the deep wiki says that this bug also applies to Google Gemini Live and XAI https://deepwiki.com/search/-title-realtimemodel-forces-al_e19bb7f1-7e3b-46b7-b175-10a3f22fb5f1?mode=fast .

## Context

We use OpenAI's `gpt-realtime` model in a turn-taking approach following the officially documented pattern: [Keep VAD but disable automatic responses](https://developers.openai.com/api/docs/guides/realtime-conversations#keep-vad-but-disable-automatic-responses).

That is, we set `turn_detection.interrupt_response` and `turn_detection.create_response` to `false`, and we use server-side VAD to commit messages to context. This gives us full control over when to commit data (at end of user turn) while still allowing OpenAI to manage server-side transcript updates during the user turn.

We do **not** use LiveKit's `"manual"` turn detection without VAD, because we want server-side VAD to auto-commit user messages during the user turn.

```python
openai.realtime.RealtimeModel(
    model="gpt-realtime",
    voice="marin",
    api_key=cls._config.openai.api_key,
    turn_detection=create_turn_detection(  # auto commit audio via VAD
        detection_type="server_vad",
        prefix_padding_ms=300,
        silence_duration_ms=500,
        threshold=0.5,
        create_response=False,  # agent must not trigger response by VAD
        interrupt_response=False,
    ),
    temperature=0.6,
    input_audio_noise_reduction=(
        params.openai_realtime_device_noise_reduction.value
        if params.openai_realtime_device_noise_reduction is not None
        else NOT_GIVEN
    ),
    max_session_duration=55 * 60,  # timeout to reconnect to OAI
    conn_options=conn_options,
    tracing=cls._get_openai_tracing_configuration(),
)
```

This configuration generally works well for us.

> **Note — why not LiveKit `"manual"` turn detection with no VAD?**
>
> We want OpenAI server-side transcription because each user turn is very long (huge paragraphs of text). If we instead committed only at end of each turn as in [LiveKit manual turn detection](https://docs.livekit.io/agents/logic/turns/#manual), we would hit audio buffer limits and see high end-of-turn latency as OpenAI processes a large amount of data.
>
> We could use a different transcription engine, but we found many issues with external STT + OpenAI `gpt-realtime`. Aside from adding complexity we do not need, the recent GA of `gpt-realtime` assumes server-side Whisper transcription for cost optimisation — it automatically replaces audio tokens with text tokens from the transcript, and we do not believe there is a way to commit external STT data to regain this feature. Official OpenAI server-side STT is therefore a requirement for us.

The LiveKit framework currently detects if `turn_detection` is not `None` or `"manual"`, and in those cases forces LiveKit `allow_interruptions` to be `True`. We would like `allow_interruptions=False` because we are in a turn-taking scenario.

The LiveKit assumption is generally safe — users typically want interruptions so that when a user speaks, ongoing generation stops. However, per the [official OpenAI configuration](https://developers.openai.com/api/docs/guides/realtime-conversations#keep-vad-but-disable-automatic-responses) for `gpt-realtime`, we want: *"if the user speaks during agent speech, let the agent keep speaking"*.

> `interrupt_response`: optional boolean  
> Whether or not to automatically interrupt any ongoing response with output to the default conversation (i.e. `conversation` of `auto`) when a VAD start event occurs.

We have set this to `False` on the OpenAI side. The **problem** is that even so, when a user speaks during the agent turn, LiveKit still sends `session.interrupt()` (a `response.cancel`) — which is not what we want.

### Expected Behavior

# Expected Behaviour

When a RealtimeModel is configured with server-side VAD and `interrupt_response=False` / `create_response=False` (Keep VAD but disable automatic responses), LiveKit should:

1. Allow `allow_interruptions=False` without raising / overriding it.
2. Not call `session.interrupt()` / send `response.cancel` when user speech is detected during agent speech.
3. Let the agent continue speaking when user audio overlaps with agent output.

In other words, LiveKit's interruption behaviour should be consistent with the model's `interrupt_response` setting.

### Reproduction Steps

```bash
# Reproduction Steps

1. Create an `AgentSession` / voice agent using OpenAI `RealtimeModel` with server-side VAD turn detection, `create_response=False`, and `interrupt_response=False` (as in the configuration above).
2. Attempt to set `allow_interruptions=False` for a turn-taking (button-click) UX.
3. Observe that LiveKit rejects or overrides this (forces interruptions on) when `RealtimeModel.capabilities.turn_detection` is enabled.
4. Start an agent reply / speech turn.
5. Produce user speech that overlaps with agent speech (e.g. unmute briefly, or speech after a function-call-triggered reply where muting cannot be applied reliably).
6. Observe that LiveKit emits `session.interrupt()` / `response.cancel` and cuts off agent speech, even though OpenAI `interrupt_response` is `False`.
```

### Operating System

Ubuntu Linux

### Models Used

gpt-realtime

### Package Versions

```bash
livekit-agents==1.6.6.
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

The simple fix for us is to allow LiveKit `allow_interruptions` to be `False` for `gpt-realtime` when it is configured with `interrupt_response=False`, so the two settings stay consistent.

I am aware LiveKit aims to be a *model-agnostic* framework and cannot make a special-case exception only for OpenAI. A more generalisable approach might be to add a RealtimeModel `capability` representing the ability to disable interruptions/allow client to handle turns, set for OpenAI (and other providers) when appropriate. This issue [seems to be relevant to other realtime models](https://deepwiki.com/search/-title-realtimemodel-forces-al_e19bb7f1-7e3b-46b7-b175-10a3f22fb5f1?mode=fast ).  For most users (default `interrupt_response=True`), this capability could be disabled by default.

Code map explaining the issue: https://deepwiki.com/search/when-using-a-realtime-model-li_79d76135-3e85-4304-b4a4-d126112d7c89

### Point A — unconditional interrupt on input speech started

In `agent_activity.py`, LiveKit emits interrupt unconditionally regardless of `allow_interruptions` settings. When the user speaks during the agent turn, agent speech is stopped. This should depend on `allow_interruptions` (it already raises when `allow_interruptions=False` — maybe that is already sufficient?).

```python
def _on_input_speech_started(self, _: llm.InputSpeechStartedEvent) -> None:
    if self.vad is None or self.using_default_vad:
        self._session._update_user_state("speaking")
        if self._audio_recognition:
            self._audio_recognition._on_start_of_speech(
                started_at=time.time(),
                user_speaking_span=self._session._user_speaking_span,
            )

    # self.interrupt() is going to raise when allow_interruptions is False, llm.InputSpeechStartedEvent is only fired by the server when the turn_detection is enabled.  # noqa: E501
    # When using the server-side turn_detection, we don't allow allow_interruptions to be False.
    try:
        self.interrupt()  # input_speech_started is also interrupting on the serverside realtime session  # noqa: E501
    except RuntimeError:
        logger.exception(
            "RealtimeAPI input_speech_started, but current speech is not interruptable, this should never happen!"  # noqa: E501
        )
```

### Point B — invalid assumption that server-side VAD + `allow_interruptions=False` is always invalid

LiveKit assumes it is invalid to set server-side VAD turn detection on a `RealtimeModel` with `allow_interruptions=False`. That combination is a **valid** OpenAI configuration for server VAD with manual response control: [Keep VAD but disable automatic responses](https://developers.openai.com/api/docs/guides/realtime-conversations#keep-vad-but-disable-automatic-responses).

One solution would be to only raise this `ValueError` when the OpenAI session is configured with the defaults `turn_detection.interrupt_response=True` and `turn_detection.create_response=True`:

`livekit-agents/livekit/agents/voice/agent_activity.py`

```python
if (
    isinstance(self.llm, llm.RealtimeModel)
    and self.llm.capabilities.turn_detection
    and not self.allow_interruptions
):
    raise ValueError(
        "the RealtimeModel uses a server-side turn detection, "
        "allow_interruptions cannot be False, disable turn_detection in "
        "the RealtimeModel and use VAD on the AgentSession instead"
    )
```

### Point C — `generate_reply` overrides `allow_interruptions=False`

Also in ``livekit-agents/livekit/agents/voice/agent_activity.py`:

```python
if (
    isinstance(self.llm, llm.RealtimeModel)
    and self.llm.capabilities.turn_detection
    and allow_interruptions is False
):
    logger.warning(
        "the RealtimeModel uses a server-side turn detection, allow_interruptions cannot be False when using VoiceAgent.generate_reply(), "  # noqa: E501
        "disable turn_detection in the RealtimeModel and use VAD on the AgentTask/VoiceAgent instead"  # noqa: E501
    )
    allow_interruptions = NOT_GIVEN
```

This overrides the choice of `allow_interruptions` from `False` back to `NOT_GIVEN` (which acts as `True`). Similarly, this `if` should be narrowed to only the case where it is relevant (`turn_detection.interrupt_response=True` and `turn_detection.create_response=True`).

### Additional Context

- We would be grateful if the LiveKit team can consider this feature request / bug fix.
- Related OpenAI docs: [Keep VAD but disable automatic responses](https://developers.openai.com/api/docs/guides/realtime-conversations#keep-vad-but-disable-automatic-responses)
- Related LiveKit docs: [Manual turn detection](https://docs.livekit.io/agents/logic/turns/#manual)
- DeepWiki code map: https://deepwiki.com/search/when-using-a-realtime-model-li_79d76135-3e85-4304-b4a4-d126112d7c89

_No response_

longcw:
Server VAD with `create_response=False` still commits and transcribes the user audio, but the reply is the client's call, so the session no longer reports server-side turn detection and `allow_interruptions=False` is honored.

Fixes livekit/agents#6635

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -453,10 +453,13 @@ class RealtimeModel(llm.RealtimeModel):
             )
 
         modalities = modalities if is_given(modalities) else ["text", "audio"]
+        resolved_turn_detection = to_turn_detection(turn_detection)
         super().__init__(
             capabilities=llm.RealtimeCapabilities(
                 message_truncation=True,
-                turn_detection=turn_detection is not None,
+                # create_response=False leaves the reply to the client: client-side turn taking
+                turn_detection=resolved_turn_detection is not None
+                and resolved_turn_detection.create_response is not False,
                 can_disable_turn_detection=not is_given(turn_detection),
                 user_transcription=input_audio_transcription is not None,
                 auto_tool_reply_generation=False,
@@ -501,7 +504,7 @@ class RealtimeModel(llm.RealtimeModel):
             modalities=modalities,
             input_audio_transcription=to_audio_transcription(input_audio_transcription),
             input_audio_noise_reduction=to_noise_reduction(input_audio_noise_reduction),
-            turn_detection=to_turn_detection(turn_detection),
+            turn_detection=resolved_turn_detection,
             api_key=api_key,
             base_url=base_url_val,
             is_azure=is_azure,
@@ -2244,6 +2247,12 @@ class RealtimeSession(
 
     def _handle_error(self, event: RealtimeErrorEvent) -> None:
         if event.error.message.startswith("Cancellation failed"):
+            return
+
+        if event.error.code == "input_audio_buffer_commit_empty" and (
+            self._opts.turn_detection is not None
+        ):
+            # the server VAD commits each segment itself, ours lands on an emptied buffer
             return
 
         provider_label = self._realtime_model._provider_label
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -454,6 +454,16 @@ class RealtimeModel(llm.RealtimeModel):
 
         modalities = modalities if is_given(modalities) else ["text", "audio"]
         resolved_turn_detection = to_turn_detection(turn_detection)
+        if (
+            resolved_turn_detection is not None
+            and resolved_turn_detection.create_response is False
+            and resolved_turn_detection.interrupt_response is not False
+        ):
+            logger.warning(
+                "create_response=False hands turn taking to the client, but the server still "
+                "cancels its response on user speech, pass interrupt_response=False as well"
+            )
+
         super().__init__(
             capabilities=llm.RealtimeCapabilities(
                 message_truncation=True,
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1882,14 +1882,15 @@ class AgentActivity(RecognitionHooks):
                     user_speaking_span=self._session._user_speaking_span,
                 )
 
-        # self.interrupt() is going to raise when allow_interruptions is False, llm.InputSpeechStartedEvent is only fired by the server when the turn_detection is enabled.  # noqa: E501
-        # When using the server-side turn_detection, we don't allow allow_interruptions to be False.
         try:
             self.interrupt()  # input_speech_started is also interrupting on the serverside realtime session  # noqa: E501
         except RuntimeError:
-            logger.exception(
-                "RealtimeAPI input_speech_started, but current speech is not interruptable, this should never happen!"  # noqa: E501
-            )
+            # only out of sync when the server cancelled its own response, with client-side turn
+            # taking an uninterruptible speech is expected
+            if self._rt_turn_detection_enabled:
+                logger.exception(
+                    "RealtimeAPI input_speech_started, but current speech is not interruptable, this should never happen!"  # noqa: E501
+                )
 
     def _on_input_speech_stopped(self, ev: llm.InputSpeechStoppedEvent) -> None:
         if self.vad is None or self.using_default_vad:
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/utils.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/utils.py
@@ -115,6 +115,8 @@ def to_turn_detection(
                 kwargs["silence_duration_ms"] = turn_detection.silence_duration_ms
             if turn_detection.create_response is not None:
                 kwargs["create_response"] = turn_detection.create_response
+            if turn_detection.interrupt_response is not None:
+                kwargs["interrupt_response"] = turn_detection.interrupt_response
             return realtime.realtime_audio_input_turn_detection.ServerVad(**kwargs)
         elif turn_detection.type == "semantic_vad":
             kwargs["type"] = "semantic_vad"
--- a/tests/test_realtime/test_openai_realtime_model.py
+++ b/tests/test_realtime/test_openai_realtime_model.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from types import SimpleNamespace
 from typing import cast
 
@@ -80,6 +81,29 @@ def test_create_response_false_reports_client_side_turn_taking() -> None:
     assert auto_reply.capabilities.turn_detection is True
 
     assert RealtimeModel(api_key="fake").capabilities.turn_detection is True
+
+
+def test_create_response_false_warns_when_the_server_still_interrupts(
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    # interrupt_response is a separate switch, left on the server keeps cancelling its response
+    # on user speech while the client believes it owns interruptions
+    with caplog.at_level(logging.WARNING, logger="livekit.plugins.openai"):
+        RealtimeModel(
+            api_key="fake",
+            turn_detection=ServerVad(type="server_vad", create_response=False),
+        )
+    assert "pass interrupt_response=False" in caplog.text
+
+    caplog.clear()
+    with caplog.at_level(logging.WARNING, logger="livekit.plugins.openai"):
+        RealtimeModel(
+            api_key="fake",
+            turn_detection=ServerVad(
+                type="server_vad", create_response=False, interrupt_response=False
+            ),
+        )
+    assert caplog.text == ""
 
 
 def test_legacy_turn_detection_keeps_interrupt_response() -> None:
--- a/tests/test_realtime/test_openai_realtime_model.py
+++ b/tests/test_realtime/test_openai_realtime_model.py
@@ -4,6 +4,8 @@ from types import SimpleNamespace
 from typing import cast
 
 import pytest
+from openai.types.beta.realtime.session import TurnDetection as BetaTurnDetection
+from openai.types.realtime.realtime_audio_input_turn_detection import ServerVad
 
 from livekit.agents import llm
 from livekit.agents._exceptions import APIError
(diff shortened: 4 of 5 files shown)


biztex:
Treating `create_response=False` as client-side turn taking is a nicer framing than what I had in #6638 — it fixes #6635 without touching the interruption paths at all.

One thing worth pairing with it: once an uninterruptible speech is legitimately possible here, `_on_input_speech_started` calls `self.interrupt()` on every barge-in, and that call raises from the *middle* of `AgentActivity.interrupt()` — after `_cancel_preemptive_generation()` and `_interrupt_background_speeches()` have already run, but before the queued speeches, `rt_session.interrupt()` and the future wiring. The `except RuntimeError` here swallows the error, but the background speeches are already interrupted and the rest of the sequence is skipped, so each barge-in leaves a half-applied interruption. The same thing happens for a queued `say(..., allow_interruptions=False)`, since the `_speech_q` loop doesn't filter protected handles the way `_interrupt_background_speeches` does.

Both are independent of which approach fixes #6635 (they reproduce today with a pipeline agent and `say(allow_interruptions=False)` + `session.interrupt()`), so I put them in #6644 rather than adding noise here.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -453,10 +453,23 @@ def __init__(
...
+            )
+
         super().__init__(
             capabilities=llm.RealtimeCapabilities(
                 message_truncation=True,
-                turn_detection=turn_detection is not None,
+                # create_response=False leaves the reply to the client: client-side turn taking
+                turn_detection=resolved_turn_detection is not None
+                and resolved_turn_detection.create_response is not False,
                 can_disable_turn_detection=not is_given(turn_detection),
```
bml1g12, line 473: turn_detection is given when using server_vad ([create_response=False,  interrupt_response=False](https://developers.openai.com/api/docs/guides/realtime-conversations#keep-vad-but-disable-automatic-responses),) VAD, but its still client driven turn taking, so I think can_disable_turn_detection should be True not False when create_response=False?

longcw: `can_disable_turn_detection` is true when `turn_detection` is not specified, it's not used if `turn_detection` is already False.

bml1g12: We have turn_detection = ServerVad() , i.e. specified, and have client controller turn detection (c.f. on https://github.com/livekit/agents/issues/6635#issue-5025296203 "Note — why not LiveKit "manual" turn detection with no VAD" for why) 

Which means I think can_disable_turn_detection = not isGiven(True) = False 

> can_disable_turn_detection: bool = False
>    """Whether server-side turn detection can be disabled for a session so the client drives
>    turn-taking. Set by plugins that implement ``session(turn_detection_disabled=True)``."""

Is `can_disable_turn_detection` = False problematic for addressing https://github.com/livekit/agents/issues/6635 I wonder? I do not fully understand that flag, but I desire our own application code to control turn taking via RPC events, not the OpenAI server deciding when to generate replies and interrupt, so it sounds like I want it to be True to me?

longcw: `can_disable_turn_detection` means if the framework can automatically disable the server turn detection **fully** if a client turn detection is specified.

in your case, `turn_detection = ServerVad()` means we shouldn't change your config. and in this pr with `create_response` True, the `cap.turn_detection` is already set to False. nothing needed from `can_disable_turn_detection`

bml1g12: I see

biztex: @bml1g12 since you mentioned the flag is hard to place, here is your exact config traced through this PR — it should do what you want, and `can_disable_turn_detection=False` is the right value for you rather than a problem.

With `turn_detection=ServerVad(create_response=False, interrupt_response=False)`:

1. `capabilities.turn_detection` becomes **False**, because this PR makes it `resolved_turn_detection is not None and resolved_turn_detection.create_response is not False`. (Small slip in the comment above — it's `create_response` **False** that flips the capability off, not True.)
2. `AgentActivity._resolve_rt_turn_detection_enabled()` returns on its first branch when `caps.turn_detection` is False, so `can_disable_turn_detection` is never read. That's what @longcw means by "nothing needed from it" — it only gates whether the framework may switch the server's turn detection *off* for you, and there is nothing left to switch off.
3. `_rt_turn_detection_enabled` is therefore False, so the two places that used to reject your setup no longer fire: the `ValueError` at construction (`agent_activity.py:236`) and the `say()` override that forced `allow_interruptions` back to `NOT_GIVEN` (`agent_activity.py:1410`). `allow_interruptions=False` is honoured, and `say(..., allow_interruptions=False)` keeps its value.

And because `can_disable_turn_detection` is False, the framework leaves the `ServerVad` config you passed exactly as you set it — with it True the framework would consider itself free to turn your server VAD off, which is the opposite of what you want, since you rely on the server VAD to keep committing and transcribing while your own code decides when to reply.

Worth keeping the `interrupt_response=False` in there as this PR's new warning suggests: with `create_response=False` alone the server still cancels its own response on user speech, which would cut a reply short even though your app never asked for it.

bml1g12: I see thanks for the detailed explanation, I misunderstood what  the parameter can_disable_turn_detection meant

davidzhao (approved):

PR #6643, opened 2026-07-31
fix(google): release the genai client when the realtime session closes

biztex:
## Problem

`google.realtime.RealtimeSession` constructs its own `genai.Client` (`realtime_api.py:488`), but `aclose()` only tears down the websocket and the pending futures — the client is never closed. Its http clients therefore stay alive until the garbage collector reaches them, and both `google/genai/client.py:186` and `google/genai/_api_client.py:2234` implement `__del__` as:

```python
def __del__(self) -> None:
    try:
    except Exception:
        pass
```

So the close is scheduled as a task on whatever event loop happens to be running when the collector fires — not necessarily the loop the client was created on, and possibly during shutdown.

The visible symptom is in CI: the `unit-tests` job intermittently fails the `fail_on_leaked_tasks` check with pending `BaseApiClient.aclose` / `AsyncClient.aclose` / httpx `AsyncClient.aclose` tasks, attributed to whichever test happened to be finishing when the collector ran (I hit it on an unrelated AMD-only PR, #6639).

## Fix

`RealtimeSession.aclose()` now awaits `self._client.aio.aclose()` after the rest of its teardown, so the session releases what it allocated. Failures there are logged rather than raised, since aclose is a teardown path.

`tests/test_plugin_google_realtime.py` builds sessions through an async context manager that closes them, instead of leaving them to the collector — otherwise the tests keep leaking clients regardless of the plugin fix. Added a test asserting the close actually reaches the genai client.

## Verification

`ruff format --check`, `ruff check`, `check_types.py` (mypy strict) and the full `pytest --unit` suite all pass locally.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -858,6 +858,15 @@ class RealtimeSession(llm.RealtimeSession):
         if self._current_generation:
             self._mark_current_generation_done()
 
+        # release the genai http clients owned by this session. Without this
+        # they stay open until the garbage collector runs `AsyncClient.__del__`,
+        # which schedules `aclose()` on whatever event loop happens to be
+        # running at that moment.
+        try:
+            await self._client.aio.aclose()
+        except Exception:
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
lg

PR #6644, opened 2026-07-31
fix(voice): stop the interrupt walk at a queued speech that disallows interruptions

biztex:
A bug in `AgentActivity.interrupt()`, independent of how #6635 gets fixed — found while working on #6638, which was closed in favour of #6642. It isn't realtime-specific.

## Problem

`SpeechHandle.interrupt()` raises `RuntimeError` when the handle disallows interruptions, and the queue loop calls it on every queued speech without handling that. With an interruptible speech playing and a protected one queued behind it, the raise escapes from the middle of `interrupt()`: the remaining queued speeches are never interrupted and the returned future never resolves.

Reachable today without a realtime model: `session.say("...", allow_interruptions=False)` queued behind another speech, then any `session.interrupt()` from user code.

## Fix

The queue walk stops at the first speech that disallows interruptions, logging a warning that names the handle and points at `force=True`. Everything behind it is going to play, so interrupting past it would leave a gap in the conversation — skipping per handle would drop the middle speech of `[protected, interruptible, protected]` and play the other two.

The queue is walked sorted by `(priority, sequence)`, since `_speech_q` is a heap and its list order is not the order it pops in.

Behaviour that is deliberately unchanged: the preemptive generation and the background speeches are cancelled as before, `rt_session.interrupt()` is still called unconditionally (nothing maps the streaming realtime response to a `SpeechHandle`, so there is no sound way to decide it belongs to a speech that was kept), a protected speech that is currently *playing* still raises, and `force=True` interrupts the whole chain.

## Verification

`tests/test_interrupt_protected_speech.py` covers stopping mid-queue with the speeches behind untouched and the warning emitted, the `[protected, interruptible, protected]` no-hole case, a priority-ordered case where heap order differs from pop order, the playing-protected speech still raising, `force=True`, and the ordinary all-interruptible path.

`ruff format --check`, `ruff check`, `check_types.py` (mypy strict) and the full `pytest --unit` suite pass locally.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1506,66 +1506,48 @@ class AgentActivity(RecognitionHooks):
 
         return interrupted_speeches
 
-    def _playout_ordered_speeches(self) -> list[SpeechHandle]:
-        """The playing speech followed by the queued ones, in playout order.
-
-        ``_speech_q`` is a heap, so its list order is not the order the
-        speeches will be popped in.
-        """
-        speeches = [self._current_speech] if self._current_speech is not None else []
-        speeches.extend(
-            speech for _, _, speech in sorted(self._speech_q, key=lambda item: (item[0], item[1]))
-        )
-        return speeches
-
     def interrupt(self, *, force: bool = False) -> asyncio.Future[None]:
         """Interrupt the current speech generation and any queued speeches.
 
-        The playing speech and the queued ones are interrupted in playout
-        order, stopping at the first speech that disallows interruptions:
-        everything behind it still plays, so interrupting past it would leave
-        a gap in the conversation. ``force`` interrupts them regardless.
+        A queued speech that disallows interruptions keeps playing, along with the ones
+        behind it, unless ``force`` is set.
 
         Returns:
             An asyncio.Future that completes when the interruption is fully processed
             and chat context has been updated
+
+        Raises:
+            RuntimeError: If the speech currently playing disallows interruptions and
+                ``force`` is False.
         """
-        # independent of the playing speech, so always cancelled
         self._cancel_preemptive_generation()
 
         future = asyncio.Future[None]()
 
         interrupted_speeches = self._interrupt_background_speeches(force=force)
 
-        kept_next_speech = False
-        stopped_a_live_speech = False
-        for speech in self._playout_ordered_speeches():
-            # a done or already interrupted handle lingers in the queue (and
-            # _current_speech) until the scheduling task drops it, but it is
-            # never going to play
-            stale = speech.done() or speech.interrupted
+        if self._current_speech is not None:
+            self._current_speech.interrupt(force=force)
+            interrupted_speeches.append(self._current_speech)
 
-            if not force and not speech.allow_interruptions:
-                if stale:
-                    # it cannot leave a gap behind it, so keep walking
-                    continue
+        if self._rt_session is not None:
+            self._rt_session.interrupt()
 
-                # SpeechHandle.interrupt() would raise here, and the speeches
-                # behind this one are going to play, so stop the walk instead
-                kept_next_speech = not stopped_a_live_speech
+        # _speech_q is a heap, so its list order is not the order it pops in
+        for _, _, speech in sorted(self._speech_q, key=lambda item: (item[0], item[1])):
+            try:
+                speech.interrupt(force=force)
+            except RuntimeError:
+                # the speeches behind this one are going to play, so stopping
+                # here keeps the conversation contiguous
+                logger.warning(
+                    "a queued speech does not allow interruptions and will play after the "
+                    "interruption, use interrupt(force=True) to interrupt it as well",
+                    extra={"speech_id": speech.id},
+                )
                 break
 
-            speech.interrupt(force=force)
             interrupted_speeches.append(speech)
-            if not stale:
-                stopped_a_live_speech = True
-
-        # a realtime response still streaming belongs to the first speech that
-        # is actually going to play - the playing one, or the head of the queue
-        # while it waits to be promoted - so cancelling it while that speech is
-        # kept would truncate it. Anything else must still stop the provider.
-        if self._rt_session is not None and not kept_next_speech:
-            self._rt_session.interrupt()
 
         if not interrupted_speeches:
             future.set_result(None)
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -1506,53 +1506,50 @@ class AgentActivity(RecognitionHooks):
 
         return interrupted_speeches
 
+    def _playout_ordered_speeches(self) -> list[SpeechHandle]:
+        """The playing speech followed by the queued ones, in playout order.
+
+        ``_speech_q`` is a heap, so its list order is not the order the
+        speeches will be popped in.
+        """
+        speeches = [self._current_speech] if self._current_speech is not None else []
+        speeches.extend(
+            speech for _, _, speech in sorted(self._speech_q, key=lambda item: (item[0], item[1]))
+        )
+        return speeches
+
     def interrupt(self, *, force: bool = False) -> asyncio.Future[None]:
         """Interrupt the current speech generation and any queued speeches.
 
-        Speeches that disallow interruptions are left running unless ``force``
-        is set; the playing one is reported instead of being skipped silently.
+        The playing speech and the queued ones are interrupted in playout
+        order, stopping at the first speech that disallows interruptions:
+        everything behind it still plays, so interrupting past it would leave
+        a gap in the conversation. ``force`` interrupts them regardless.
 
         Returns:
             An asyncio.Future that completes when the interruption is fully processed
             and chat context has been updated
-
-        Raises:
-            RuntimeError: If the playing speech disallows interruptions and
-                ``force`` is False. Nothing is interrupted in that case.
         """
-        # checked up-front: SpeechHandle.interrupt() raises for a protected
-        # handle, and letting that happen mid-sequence would leave the
-        # background speeches interrupted, the queue untouched, the realtime
-        # session untold and the returned future unresolved
-        if (
-            not force
-            and self._current_speech is not None
-            and not self._current_speech.allow_interruptions
-        ):
-            raise RuntimeError(
-                "the current speech does not allow interruptions, "
-                "use interrupt(force=True) to interrupt it anyway"
-            )
-
+        # independent of the playing speech, so always cancelled
         self._cancel_preemptive_generation()
 
         future = asyncio.Future[None]()
 
         interrupted_speeches = self._interrupt_background_speeches(force=force)
 
-        if self._current_speech is not None:
-            self._current_speech.interrupt(force=force)
-            interrupted_speeches.append(self._current_speech)
+        current_speech = self._current_speech
+        for speech in self._playout_ordered_speeches():
+            if not force and not speech.allow_interruptions:
+                # SpeechHandle.interrupt() would raise here, and the speeches
+                # behind this one are going to play, so stop the walk instead
+                break
 
-        # skip queued handles that disallow interruptions, like the background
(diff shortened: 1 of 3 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1581,10 +1509,31 @@ def _interrupt_background_speeches(self, force: bool = False) -> list[SpeechHand
...
+        # checked up-front: SpeechHandle.interrupt() raises for a protected
+        # handle, and letting that happen mid-sequence would leave the
+        # background speeches interrupted, the queue untouched, the realtime
+        # session untold and the returned future unresolved
+        if (
+            not force
+            and self._current_speech is not None
+            and not self._current_speech.allow_interruptions
+        ):
+            raise RuntimeError(
```
longcw: thanks!  the second one looks like a bug to me:

1. The preemptive generation and the background speeches don't depend on the playing speech, so I'd keep cancelling both unconditionally.
2. The second is real: with an interruptible current speech and a protected one queued behind it, playout stops but the loop raises before `rt_session.interrupt()`, so the server keeps generating. But filtering the queue per handle punches a hole in it (`[protected, interruptible, protected]` drops the middle one and plays the rest).

rather than filtering per handle I'd walk [current] + queue in playout order and stop at the first protected speech. (note `_speech_q` is a heap, so the existing loop order may not be the playout order). and if current is None or interrupted, we make sure `self._rt_session.interrupt()` is called.

biztex: Both points taken — implemented in 0a6109d.

1. Reverted the up-front raise. You're right that the preemptive generation and the background speeches don't depend on the playing speech, so cancelling them isn't partial application; they're unconditional again as before.

2. The queue now walks in playout order and stops at the first protected speech, instead of filtering per handle — your `[protected, interruptible, protected]` case is exactly the hole I'd have punched. Added `_playout_ordered_speeches()`, which yields the playing speech followed by `sorted(self._speech_q, key=(priority, seq))` — thanks for the heap note, the raw list order is not pop order and I'd have gotten that wrong too. The realtime cancel is now sent when the current speech is `None` or ended up interrupted, so the "playout stopped but the server keeps generating" case is covered while a surviving protected speech isn't truncated.

One consequence worth flagging since it's a public behaviour change: with the walk stopping instead of raising, `AgentActivity.interrupt()` (and therefore `AgentSession.interrupt()`) no longer raises `RuntimeError` when the playing speech is protected — it interrupts nothing and returns a future. The docstrings say so now. That follows from "stop at the first protected speech", but if you'd rather keep the raise for that one case I'm happy to add it back after the walk.

Tests updated to the new semantics: the protected head stopping the walk while background speeches and the preemptive generation are still cancelled, stopping mid-queue with the speeches behind untouched, your three-speech no-hole case, a priority-ordered case where heap order differs from playout order, the realtime cancel when nothing is playing, and `force=True` taking the whole chain.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1506,28 +1506,50 @@ def _interrupt_background_speeches(self, force: bool = False) -> list[SpeechHand
...
 
-        for _, _, speech in self._speech_q:
             speech.interrupt(force=force)
             interrupted_speeches.append(speech)
 
-        if self._rt_session is not None:
+        # a realtime response still streaming belongs to the playing speech;
+        # cancelling it while that speech survives would truncate it
+        if self._rt_session is not None and (current_speech is None or current_speech.interrupted):
             self._rt_session.interrupt()
```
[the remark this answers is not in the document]
biztex: Both windows check out, and keying the guard off the head of the walk is simpler than what I had — fixed in 1ed7452.

Verified each: `_schedule_speech` pushes to `_speech_q` and `_current_speech` is only assigned when the scheduling task pops it, so with a realtime generation waiting to be promoted the walk could keep a protected queue head and still cancel the response that belongs to it. And `SpeechHandle._cancel()` returns early once `done()`, so `interrupted` stays `False` for a done-but-not-yet-cleared current speech — the provider would have been left generating even though nothing was protected.

The cancel now depends on whether the walk kept the *first* speech in playout order (`kept_next_speech = index == 0`), which covers both: an empty playout order or a break further down still cancels, and only a kept head suppresses it. Two tests added — a protected queue head with nothing playing keeps the response, and a done current speech still cancels.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1506,28 +1506,53 @@ def _interrupt_background_speeches(self, force: bool = False) -> list[SpeechHand
...
-        if self._current_speech is not None:
-            self._current_speech.interrupt(force=force)
-            interrupted_speeches.append(self._current_speech)
+        kept_next_speech = False
+        for index, speech in enumerate(self._playout_ordered_speeches()):
+            if not force and not speech.allow_interruptions:
+                # SpeechHandle.interrupt() would raise here, and the speeches
+                # behind this one are going to play, so stop the walk instead
+                kept_next_speech = index == 0
+                break
```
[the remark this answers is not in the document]
biztex: Right on both counts — a handle that will never play can't leave a gap, so it shouldn't stop the walk. Fixed in e28ca1a. (Confirmed the lingering paths: `_current_speech` is only cleared after `_wait_for_generation()` returns, and the scheduling loop's own "skip done speech (interrupted when it's in the queue)" comment shows force-interrupted handles sitting in `_speech_q`.)

I took the `continue` but not the `index == 0` that came with it, because once the walk can skip entries, position stops tracking what actually plays next. With `[stale-protected, live-protected]` the break lands at index 1, so `kept_next_speech` would be `False` and we'd cancel the response belonging to the live protected speech — the same truncation as before, one slot over. Same shape in reverse for `[stale-interruptible, live-protected]`, where the stale handle would be "interrupted" (a no-op) and again make the live one look unshielded.

So the guard now tracks whether the walk actually stopped a speech that was going to play, and `kept_next_speech = not stopped_a_live_speech`. That's what "the response belongs to the first speech that will really play" needs, and it subsumes the earlier `index == 0` version.

Three tests added: a finished protected `_current_speech` no longer shields the queue (and the realtime cancel is sent), a force-interrupted protected queue entry likewise, and a live protected speech behind a stale one keeps its streaming response.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1578,28 +1506,65 @@ def _interrupt_background_speeches(self, force: bool = False) -> list[SpeechHand
...
-        if self._rt_session is not None:
+            if not stale:
+                stopped_a_live_speech = True
+
+        # a realtime response still streaming belongs to the first speech that
+        # is actually going to play - the playing one, or the head of the queue
+        # while it waits to be promoted - so cancelling it while that speech is
+        # kept would truncate it. Anything else must still stop the provider.
+        if self._rt_session is not None and not kept_next_speech:
             self._rt_session.interrupt()
```
longcw: `kept_next_speech` can't work as intended: nothing maps the active realtime response to a `SpeechHandle`, and queue position doesn't stand in for it. so guard the `rt_session` with `kept_next_speech` actually won't always correct.

I think that moving the `rt_session.interrupt` to after `current_speech` interrupted is safer choice. 

```suggestion
        A queued speech that disallows interruptions keeps playing, along with the ones
        behind it, unless ``force`` is set.

        Returns:
            An asyncio.Future that completes when the interruption is fully processed
            and chat context has been updated
        """
        self._cancel_preemptive_generation()

        future = asyncio.Future[None]()

        interrupted_speeches = self._interrupt_background_speeches(force=force)

        if self._current_speech is not None:
            self._current_speech.interrupt(force=force)
            interrupted_speeches.append(self._current_speech)

        if self._rt_session is not None:
            self._rt_session.interrupt()

        for _, _, speech in sorted(self._speech_q, key=lambda x: (x[0], x[1])):
            try:
                speech.interrupt(force=force)
            except RuntimeError:
                logger.warning(
                    "a queued speech does not allow interruptions and will play after the "
                    "interruption, use interrupt(force=True) to interrupt it as well",
                    extra={"speech_id": speech.id},
                )
                break

            interrupted_speeches.append(speech)
```

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1578,28 +1506,65 @@ def _interrupt_background_speeches(self, force: bool = False) -> list[SpeechHand
 
         return interrupted_speeches
 
+    def _playout_ordered_speeches(self) -> list[SpeechHandle]:
```
longcw: no need to be a helper function.

livekit-agents/livekit/agents/voice/agent_activity.py
biztex: You're right, and I checked before changing it: `response_id` only ever reaches `_realtime_spans` and the metrics, so there genuinely is no handle the streaming response can be traced back to. Queue position was a guess that happens to be right in the common case, and when it's wrong it truncates a reply silently — not a good trade for a fix that's supposed to be about the queue. Taken as suggested in 5602c51: the cancel is unconditional again, right after the current speech, exactly as before this PR.

I kept the `try/except RuntimeError` shape you wrote rather than pre-checking `allow_interruptions`, since the warning naming the handle is more useful than a silent skip.

One small thing I dropped along with `kept_next_speech`, just so it's a deliberate choice and not an oversight: a queued handle that was already force-interrupted still raises here, so it stops the walk even though it can never play. It looked harmless in practice — a handle only becomes done inside the queue via `interrupt(force=True)`, which cancels everything behind it in the same pass — so I left it out rather than adding a branch for it. Happy to add the guard if you'd rather be strict about it.

livekit-agents/livekit/agents/voice/agent_activity.py
biztex: Done — inlined the `sorted(...)` into the loop and dropped the helper in 5602c51.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -1523,13 +1530,25 @@ def interrupt(self, *, force: bool = False) -> asyncio.Future[None]:
             self._current_speech.interrupt(force=force)
             interrupted_speeches.append(self._current_speech)
 
-        for _, _, speech in self._speech_q:
-            speech.interrupt(force=force)
-            interrupted_speeches.append(speech)
-
         if self._rt_session is not None:
             self._rt_session.interrupt()
```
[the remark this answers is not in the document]
biztex: Thanks — the observation is accurate, but this is deliberate rather than something to change, so I'm resolving with the reasoning rather than acting on it.

The unguarded `self._current_speech.interrupt(force=force)` is not introduced here: it is the same on `main` today, in the same position, after `_cancel_preemptive_generation()` and `_interrupt_background_speeches()`. This PR narrows its scope to the *queued* speeches, which is where the raise did real damage — it escaped mid-loop and left the rest of the queue scheduled with the future unresolved.

The suggested alternative — fold the playing speech into the walk and make the realtime cancel conditional — is exactly what this PR did two revisions ago, and the maintainer rejected it in review: nothing maps the streaming realtime response back to a `SpeechHandle` (`response_id` only reaches `_realtime_spans` and the metrics), so any guard keyed on position or on which handle was kept is guesswork, and when it guesses wrong it truncates a reply silently. The unconditional cancel is the known-safe behaviour.

On the specific consequences:

- Cancelling the preemptive generation and the background speeches before the raise is intended — they don't depend on the playing speech, so stopping them is correct regardless of what happens next, not collateral damage.
- Not reaching `rt_session.interrupt()` on that path is the outcome we want: the protected speech carries on playing, and cancelling the provider underneath it would truncate exactly the speech the developer marked uninterruptible.
- The half of this that *was* real is fixed: `rt_session.interrupt()` now runs before the queue walk, so the `break` can no longer skip it. On `main` it sat after the loop, which is how the queued raise stranded it.

The description/docstring mismatch you noted was against the previous revision; both were rewritten after that push and now state that a protected *playing* speech still raises, which matches the code.

bml1g12:
@longcw  thanks for review here, it looks like its ready for re-review

longcw (approved):

PR #6645, opened 2026-07-31
fix(livekit-agents): keep plugin-specific realtime handlers across child swaps

### LINKED ISSUE #6556: Should automatically resubscribe those plugin specific rt_session event handlers to new child rt_session when realtime model is fallbacked to the same plugin-type model
### Feature Type

### Feature Description

According to deepwiki [https://deepwiki.com/search/can-fallbackrealtimesession-im_15e6ca5e-6280-4091-b3d6-65ddfb72aedf?mode=deep](https://deepwiki.com/search/can-fallbackrealtimesession-im_15e6ca5e-6280-4091-b3d6-65ddfb72aedf?mode=deep),
for `livekit.agents.llm.RealtimeModelAapteri`, **openai specific** realtime session event handlers are attached to its **child** realtime session and handlers for **general** reatime session event in [https://github.com/livekit/agents/blob/87ef1b15f11239fa1e4cec9f6eace72b71904830/livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py#L58](https://github.com/livekit/agents/blob/87ef1b15f11239fa1e4cec9f6eace72b71904830/livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py#L58) ) are forwarded to attached to its **parent** realtime session.

When model fallback happens, its  **old** child realtime session are closed and discarded immediately before the **new** one is created and its **parent** realtime session sill keep those handlers of those **general** realtime session event. As a result, those **openai specific** realtime session event handlers original attached to the child session will **lose**. Therefore, we need to **resubscribe** those **openai specific** realtime session event handlers to the new child realtime session. It has to be done **manually** every time when we have **openai specific** realtime session event handlers to attach.

Can you modify the source codes so that the **resubscribe** is implemented **automatically** when realtime model is fallbacked to the same plugin-type model (eg. openai -> openai,  google -> google)?

Since we have set realtime model fallback from openai to Azure openai as below:
```python
llm=RealtimeModelFallbackAdapter(
 [openai.realtime.RealtimeModel(),
openai.realtime.RealtimeModel.with_azure()]
)
```
,
we need to manually **resubscribe** those **openai specific** realtime session event handlers to the new child realtime session through the codes like below:
```python
def attach_oai_realtime_event_handlers(self, rt_session: RealtimeSession) -> None:
  if not isinstance(rt_session, livekit.plugins.openai.realtime.realtime_model.RealtimeSession:
     raise TypeError(f"Expected OpenAI RealtimeSession, got {type(rt_session).__name__}")

    @rt_session.on("openai_server_event_received")
    def on_openai_server_event_received(event: dict[str, Any]) -> None:
          pass

if type(rt_session).__name__ == "_FallbackRealtimeSession":
  current_child_session = rt_session._active
  @rt_session.on("session_reconnected")
  def on_session_reconnected(_: RealtimeSessionReconnectedEvent) -> None:
      nonlocal current_child_session
      new_child_session = rt_session._active 
      # check whether the new child session is different from the current child session
      # to make sure the received 'session_reconnected' event is emitted by model fallback
      if new_child_session is not current_child_session:
          LOGGER.info(
              "Resubscribe OpenAI realtime event handlers when model fallback happens "
              "for RealtimeFallbackAdapter"
          )
          attach_oai_realtime_event_handlers(new_child_session)
          current_child_session = new_child_session
```

### Additional Context

_No response_

### LINKED ISSUE #6559: all attached plugin-specific realtime session event handlers lost when retart_session with the same model.
### Bug Description
According to deepwiki [https://deepwiki.com/search/can-fallbackrealtimesession-im_15e6ca5e-6280-4091-b3d6-65ddfb72aedf?mode=deep](https://deepwiki.com/search/can-fallbackrealtimesession-im_15e6ca5e-6280-4091-b3d6-65ddfb72aedf?mode=deep),
for `livekit.agents.llm.RealtimeModelAapter`,  **openai specific** realtime session event handlers are attached to its **child** realtime session and handlers for **general** reatime session event in [https://github.com/livekit/agents/blob/87ef1b15f11239fa1e4cec9f6eace72b71904830/livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py#L58](https://github.com/livekit/agents/blob/87ef1b15f11239fa1e4cec9f6eace72b71904830/livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py#L58) ) are forwarded to attached to its **parent** realtime session.

According to [source codes](https://github.com/livekit/agents/blob/48c179329522a44a979c97f58865773badf5241f/livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py#L135), when implementing **llm.restart_session(switch_model=False)** , its old child realtime session are closed and discarded and its parent realtime session only keep those handlers of those general realtime session event, and thus **all attached plugin-specific realtime session event handlers (eg. openai specific realtime event handlers etc.) will lose**

We think this is not a right behavior since you mentioned `livekit.agents.llm.RealtimeModelAapter` will preserve the agent's bound event handlers in PR [https://github.com/livekit/agents/pull/6282](https://github.com/livekit/agents/pull/6282). 

Please correct the source code so that **all attached plugin-specific realtime session event handlers should be automatically forwarded to the new child realtime session if llm.restart_session(switch_model=False) is implemented.** 

### Expected Behavior

all attached plugin-specific realtime session event handlers should be **automatically forwarded** to the new child realtime session if `llm.restart_session(switch_model=False)` is implemented.

### Reproduction Steps

1. set 
```python
llm=RealtimeModelFallbackAdapter(
 [openai.realtime.RealtimeModel(),
openai.realtime.RealtimeModel.with_azure()]
)
```

2. attach openai-specific realtime session event handlers:
```python
   rt_session = agent.realtime_llm_session
   @rt_session._active.on("openai_server_event_received")
   def on_openai_server_event_received(event: dict[str, Any]) -> None:
        logger.info("openai_server_event_received handling"
```
3. implement 
```python
await agent.llm.restart_session(switch_model=False)
```
5. try to simulate to emit "openai_server_event_received" realtime session event
6. check whether the attached openai-specific realtime session event handlers still work

### Operating System

MacOS, linux

### Models Used

Openai gpt-realtime, Azure gpt-realtime

### Package Versions

```bash
livekit=1.1.13
livekit-agents=1.6.6
livekit-api=1.2.0
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

fxhxdxd:
Fixes #6559
Fixes #6556

Same root cause for both.

The fallback session only forwards the seven events in `_FORWARDED_EVENTS`. Provider events like `openai_server_event_received` aren't in that list, so they never reach the wrapper. People work around it by subscribing on `rt_session._active`, and those handlers get dropped when `_swap` replaces the child. That happens on `restart_session(switch_model=False)` and on a real fallback.

What I did: `on` / `off` / `once` on the wrapper now check the event name. Known events behave exactly as before. Anything else gets tracked and registered on the active child, and `_bind` / `_unbind` replay them onto every new child. So you subscribe on the session you already hold and `_active` stays internal.

Subscribing to an event the current child never emits is allowed, it just never fires. A fallback list can mix providers (openai to azure), so raising there felt wrong, but happy to change it if you'd rather.

Tests added to `tests/test_realtime_fallback.py`:

- handler survives `restart_session()` on the same model
- handler survives `restart_session(switch_model=True)` to another model
- handler reaches the current child before any swap
- handler stays gone after `off`
- `once` fires at most once across a swap
- known events still register on the wrapper, not as plugin handlers

Five of those six fail on main. ruff and `mypy --strict` are clean on the touched files, and the full `test_realtime_fallback.py` passes (42).

### DIFF
--- a/livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py
@@ -6,7 +6,7 @@ import time
 import weakref
 from collections.abc import AsyncIterable, Callable
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
 from livekit import rtc
 
@@ -65,6 +65,13 @@ _FORWARDED_EVENTS: tuple[EventTypes, ...] = (
     "metrics_collected",
     "remote_item_added",
 )
+
+# events the wrapper handles itself; anything else is plugin-specific and is
+# registered directly on the active child session
+_WRAPPER_EVENTS: frozenset[str] = frozenset(_FORWARDED_EVENTS) | {
+    "error",
+    "realtime_availability_changed",
+}
 
 
 def _merge_capabilities(models: list[RealtimeModel]) -> RealtimeCapabilities:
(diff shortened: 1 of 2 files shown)


fxhxdxd:
Closing this. #6628 landed first and solves the same problem, I missed it when I picked this up. Sorry for the duplicate noise.

For what it's worth, the two approaches differ: #6628 adds explicit `on_active_session` / `off_active_session`, this one overrode `on`/`off`/`once` so plugin events route to the child automatically. #6628 is the safer call since it doesn't change what an existing method does.

One gap that might be worth pulling across: #6559 is specifically about `restart_session(switch_model=False)`, not a provider failover. Both go through `_swap` so #6628 should fix it too, but its tests drive swaps via `_swap_task` and none of them call `restart_session()`. I have a test for that path plus one for `once` firing at most once across a swap. Happy to send those as a small follow-up PR on top of #6628 if that's useful to anyone.

PR #6568, opened 2026-07-27
fix(llm): preserve provider events on restart

Bortlesboat:
## Summary

- forward provider-specific realtime session events through the fallback wrapper
- rebind those event forwarders whenever the underlying realtime session is restarted
- keep subscriptions made during a restart detached from the closing child session

## Root cause

The fallback session only forwarded its fixed core event set. Provider-specific events such as `openai_server_event_received` were available only on the private active child, and that child is replaced by `restart_session()`.

Applications can now subscribe to provider-specific events on the stable fallback session instead:

```python
session.on("openai_server_event_received", handler)
```

## Testing

- `pytest tests/test_realtime_fallback.py -q` (38 passed)
- `ruff check livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py tests/test_realtime_fallback.py`
- `ruff format --check livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py tests/test_realtime_fallback.py`
- `mypy --config-file pyproject.toml --untyped-calls-exclude=smithy_aws_core -p livekit.agents.llm.realtime_fallback_adapter`

Fixes #6559

Issue #6553, opened 2026-07-27
Should make livekit.agents.llm._FallbackRealtimeSession as public class for implementing instance check easily


moz164164:
### Feature Type

Would make my life easier

### Feature Description

In recent released version v1.6.6, a new class `livekit.agents.llm._FallbackRealtimeSession` is the subclass of `livekit.agents.llm.RealtimeSession`. Also, a existing class `livekit.plugins.openai.realtime.realtime_model.RealtimeSession` is the subclass  of `livekit.agents.llm.RealtimeSession` as well.

Since **openai specific** realtime session event (eg. "openai_server_event_received", "openai_client_event_queued" etc. ) handlers are only allowed to be attached to `livekit.plugins.openai.realtime.realtime_model.RealtimeSession`, we have to check whether the realtime session object of agent is the instance of `livekit.plugins.openai.realtime.realtime_model.RealtimeSession` or `livekit.agents.llm._FallbackRealtimeSession`

If it is `livekit.plugins.openai.realtime.realtime_model.RealtimeModel.RealtimeSession`, we attach those **openai specific** realtime session event handlers to the realtime session directly.

If it is `livekit.agents.llm._FallbackRealtimeSession` and its **active child realtime session** is the instance of `livekit.plugins.openai.realtime.realtime_model.RealtimeSession`, we attach those **openai specific** realtime session event handlers to its **active child realtime session** instead as below:
```python
rt_session = agent.realtime_llm_session
if isinstance(rt_session, livekit.agents.llm._FallbackRealtimeSession) and isinstance(rt_session._active, livekit.plugins.openai.realtime.realtime_model.RealtimeSession):
      @rt_session._active.on("openai_server_event_received"):
       def on_openai_server_event_received(event: dict[str, Any]) -> None:
            ....
```

However, since `livekit.agents.llm._FallbackRealtimeSession` is a **private** class and thus cannot import it for instance check directly.

We want `livekit.agents.llm._FallbackRealtimeSession` become a **public** class `livekit.agents.llm.FallbackRealtimeSession` so that we can easily import it for instance check directly.

### Workarounds / Alternatives

We temporarily implement the instance check through 
```python
rt_session = agent.realtime_llm_session
if type(rt_session).__name__ == "_FallbackRealtimeSession" and isinstance(rt_session._active, livekit.plugins.openai.realtime.realtime_model.RealtimeSession):
      @rt_session._active.on("openai_server_event_received"):
[...]

PR #6646, opened 2026-07-31
fix(llm): keep plugin-specific realtime handlers across a session swap

### LINKED ISSUE #6556: Should automatically resubscribe those plugin specific rt_session event handlers to new child rt_session when realtime model is fallbacked to the same plugin-type model

### LINKED ISSUE #6559: all attached plugin-specific realtime session event handlers lost when retart_session with the same model.

priyam-garg:
`_FallbackRealtimeSession` only re-attached the fixed `_FORWARDED_EVENTS` allowlist to each new child, so handlers for provider-specific events (e.g. openai's `openai_server_event_received`) were silently dropped by `restart_session()` and by an availability fallback.

Forwarders are now created on demand: subscribing on the wrapper registers a forwarder for that event, and `_bind` replays every forwarder onto each new child, so plugin-specific subscribers survive a swap like the generic ones already did. A `_child_bound` flag keeps a subscription made mid-swap off the child that is being discarded.

Forwarders also became varargs, so a plugin event carrying more than one payload argument no longer loses all but the first.

Fixes #6559
Fixes #6556

### DIFF
--- a/livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py
@@ -6,7 +6,7 @@ import time
 import weakref
 from collections.abc import AsyncIterable, Callable
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
 from livekit import rtc
 
@@ -55,7 +55,9 @@ _SOFT_CAPABILITIES = (
     "can_disable_turn_detection",
 )
 
-# child events re-emitted on the wrapper
+# child events re-emitted on the wrapper. plugin-specific events (e.g. openai's
+# ``openai_server_event_received``) are not listed here -- the adapter cannot know them
+# ahead of time, so ``on()`` adds a forwarder for whatever a caller subscribes to.
 _FORWARDED_EVENTS: tuple[EventTypes, ...] = (
     "input_speech_started",
     "input_speech_stopped",
@@ -148,7 +150,9 @@ class RealtimeModelFallbackAdapter(
             await model.aclose()
 
 
-class _FallbackRealtimeSession(RealtimeSession[Literal["realtime_availability_changed"]]):
+# ``str`` rather than a Literal: the wrapper stands in for any provider session, and a
+# plugin's own event names are only known to that plugin.
+class _FallbackRealtimeSession(RealtimeSession[str]):
     """Bound once by AgentActivity; swaps the inner child session internally."""
 
     def __init__(
@@ -164,17 +168,15 @@ class _FallbackRealtimeSession(RealtimeSession[Literal["realtime_availability_ch
         self._tools: NotGivenOr[list[Tool]] = NOT_GIVEN
         self._tool_choice: NotGivenOr[ToolChoice | None] = NOT_GIVEN
 
-        # stable per-event forwarders so they can be detached on swap
-        def _make_forwarder(event: EventTypes) -> Callable[[object], None]:
-            # callbacks receive only the payload, so bind the event name per forwarder
-            def _forward(ev: object) -> None:
-                self.emit(event, ev)
-
-            return _forward
-
-        self._forwarders: dict[EventTypes, Callable[[object], None]] = {
-            event: _make_forwarder(event) for event in _FORWARDED_EVENTS
-        }
+        # stable per-event forwarders so they can be detached on swap. ``_bind`` replays
+        # this dict onto every new child, so anything in it survives a swap.
+        self._forwarders: dict[str, Callable[..., None]] = {}
+        # whether ``_forwarders`` are currently attached to ``self._active``. a forwarder
+        # added while unbound is picked up by the next ``_bind`` instead of being attached
+        # to the child that is on its way out.
+        self._child_bound = False
+        for event in _FORWARDED_EVENTS:
+            self._add_forwarder(event)
 
         # per-model availability, with a cooldown after a failure
         self._available = [True] * len(adapter._models)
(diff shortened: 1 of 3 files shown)


priyam-garg:
Some extra detail on the root cause and the one behavioral change reviewers will want to weigh in on.

### Root cause

`_forwarders` was built once, at construction, from a hardcoded tuple:

```python
self._forwarders: dict[EventTypes, Callable[[object], None]] = {
    event: _make_forwarder(event) for event in _FORWARDED_EVENTS
}
```

`_bind()` replays exactly that dict onto each new child, so only those seven generic events survive a swap. The adapter has no way to enumerate plugin event names up front — `openai_server_event_received`, `openai_client_event_queued`, and whatever the other plugins define — so a static allowlist can never be complete. That's why the fix moves forwarder creation to subscribe time rather than adding more names to the tuple.

### The one behavioral change

Handlers have to be bound on the wrapper rather than on the child. Binding to `_active` attaches to one specific child object, which a swap then discards — that is the reported bug, and no amount of forwarding on our side can rescue a handler the caller attached directly to the session being thrown away.

```python
rt_session = agent.realtime_llm_session

# before: bound to the child, lost on the next swap
rt_session._active.on("openai_server_event_received", handler)

# after: bound to the wrapper, re-attached to every new child
rt_session.on("openai_server_event_received", handler)
```

Worth calling out explicitly since it's a call-site change for anyone already working around this, though `_active` was never public API.

### Design notes

- **`error` is deliberately excluded** from the dynamic forwarders. `_on_child_error` already re-emits it, sometimes re-stamped as recoverable, and a second forwarder would duplicate every error. The existing error tests (`test_restart_emits_no_error`, `test_non_recoverable_error_forwarded_as_recoverable_while_fallback_remains`) now run through the overridden `on()` and would catch a regression here.
- **`_child_bound`** covers the window between `_unbind` of the outgoing child and `_bind` of the incoming one. Subscribing during that window adds to the dict only; the upcoming `_bind` picks it up, so the handler never attaches to the child being discarded.
- **Forwarders became varargs.** The old `_forward(ev)` silently truncated any plugin event carrying more than one payload argument.
- The class generic widens to `RealtimeSession[str]` since plugin event names are only known to the plugin. That `Literal["realtime_availability_changed"]` was already inaccurate — the event is emitted on the *adapter*, never on the session.

### Test evidence

Four of the five new tests fail on `main` and pass with the fix:

```
FAILED test_restart_preserves_plugin_event_subscribers
FAILED test_fallback_preserves_plugin_event_subscribers
FAILED test_plugin_event_subscribed_mid_swap_binds_to_new_child
FAILED test_forwards_multi_arg_plugin_events
4 failed, 37 passed
```

The fifth, `test_plugin_event_detached_from_discarded_child`, passes either way by construction — it's a regression guard against a fix that leaves the discarded child bound and double-delivers.

### Scope of #6556

Flagging this so the auto-close isn't a surprise: #6556 scopes the ask to resubscribing *when falling back to the same plugin-type model*. This implementation resubscribes on every swap, including across plugin types — an `openai_server_event_received` forwarder would attach to, say, a Gemini session and simply never fire. Harmless, but broader than what was asked. Happy to gate it on `type(old) is type(new)` if you'd prefer the narrower behavior.

### Possible follow-up

This makes the case for #6553 a bit stronger: callers now need `session.on(...)` on `_FallbackRealtimeSession`, a class whose name starts with an underscore. Left out of this PR to keep it focused, but happy to pick it up separately.

Verified with `ruff format --check`, `ruff check`, and `mypy --strict` on `livekit.agents.llm`. Note that the tests exercise the hermetic fakes in `tests/fake_realtime.py`; I haven't run the issue's exact repro against a live OpenAI realtime session, though the forwarding path is provider-agnostic.

longcw:
duplicated by https://github.com/livekit/agents/pull/6568

PR #6648, opened 2026-07-31
fix(elevenlabs): close TTS websocket on job shutdown

BhaskarKapri07:
## Summary

- register the ElevenLabs TTS `aclose()` method as a job shutdown callback when its reusable multi-stream WebSocket is first acquired
- register only once even when the WebSocket connection is reused
- close ElevenLabs-owned streams and the WebSocket without closing the job-owned `aiohttp.ClientSession`

## Problem

The ElevenLabs multi-stream WebSocket is intentionally reused across synthesis requests. `AgentSession.aclose()` stops active synthesis streams, but it does not close the configured TTS model, and `FallbackAdapter.aclose()` does not cascade cleanup to its child providers.

At job teardown, the shared HTTP context can therefore close while the ElevenLabs WebSocket is still open. The remote endpoint observes an abnormal `1006` closure because it never receives a WebSocket Close frame, even when synthesis completed successfully.

## Why the job shutdown callback

The worker awaits `AgentSession.aclose()` before running job shutdown callbacks, then closes the shared HTTP context after those callbacks. Registering the concrete ElevenLabs model in that window:

1. lets active synthesis streams stop first
2. gracefully closes the reusable WebSocket
3. preserves ownership of the shared HTTP session
4. avoids closing provider connections during agent handoffs

This is deliberately narrower than closing TTS models from `AgentActivity`, as discussed in #3825, and avoids the shared-session ownership issue identified in #4745.

## Tests

The regression test uses the real ElevenLabs plugin against a local `aiohttp` WebSocket server and verifies that:

- job shutdown produces close code `1000`
- repeated connection acquisition registers only one callback
- the supplied/shared HTTP session remains open

Commands run:

- `uv run pytest --plugin elevenlabs tests/test_plugin_elevenlabs_tts.py -q`
- `make check`

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
@@ -34,6 +34,7 @@ from livekit.agents import (
     APIError,
     APIStatusError,
     APITimeoutError,
+    JobContext,
     LanguageCode,
     get_job_context,
     tokenize,
@@ -212,7 +213,7 @@ class TTS(tts.TTS):
 
         self.__current_connection: _Connection | None = None
         self._connection_lock = asyncio.Lock()
-        self._job_context_cb_registered = False
+        self._registered_job_ctx_ref: weakref.ReferenceType[JobContext] | None = None
 
     @property
     def model(self) -> str:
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/tts.py
```diff
@@ -286,6 +289,16 @@ async def _current_connection(self) -> tuple[_Connection, float, bool]:
...
         async with self._connection_lock:
+            job_ctx = get_job_context(required=False)
+            registered_job_ctx = (
+                self._registered_job_ctx_ref() if self._registered_job_ctx_ref is not None else None
+            )
+            if job_ctx is not None and job_ctx is not registered_job_ctx:
+                # AgentSession closes active synthesis streams but does not own the TTS
+                # model. Close our reusable WebSocket before the job-owned HTTP session.
+                job_ctx.add_shutdown_callback(self.aclose)
+                self._registered_job_ctx_ref = weakref.ref(job_ctx)
```
Sanjays2402, line 300: the ref only tracks the last job ctx, so a TTS shared by concurrent jobs in one process (JobExecutorType.THREAD, or a module-level/prewarmed TTS) ping-pongs: job A registers, job B registers, then A calls `_current_connection` again and registers a second callback, and so on. worse, whichever job shuts down first runs `aclose()`, which also iterates `self._streams` and closes the *other* job's in-flight synthesis streams. is a shared TTS instance across simultaneous jobs considered supported here? if so this probably needs a set of ctx refs and a shutdown path that only drops the connection when the last one goes.

PR #6649, opened 2026-07-31
Make every example a uv workspace member

u9g:
Each example declared its dependencies in a `requirements.txt` that nothing in the root sync path read — `[tool.uv.workspace].members` covered only `livekit-plugins/*` and `livekit-agents`, and pytest ignores `examples`. So `uv sync` left example dependencies uninstalled, and an example crashed at import on its first missing one:

```
File "examples/hotel_receptionist/benchmark.py", line 20, in <module>
  import apsw
ModuleNotFoundError: No module named 'apsw'
```

`apsw>=3.46` had been declared in `examples/hotel_receptionist/requirements.txt` the whole time. The lk CLI also classified the examples as pip projects from that file, launched a bare interpreter resolved from the enclosing workspace, and has no missing-dependency gate on that path — the uv path fast-fails with `run uv sync`, the pip path does not.

## What changed

Every example that had a `requirements.txt` (9 of them) now declares its dependencies in a `pyproject.toml` and joins the workspace. Dependencies land in the root lock, so `uv sync --all-extras --dev` installs them, and `[tool.uv.sources]` keeps in-repo dependencies resolving to the local editable packages rather than PyPI.

A deploy build context is the example directory alone, where workspace sources cannot resolve, so `requirements.txt` is rendered from the declared dependencies at deploy time instead of being checked in. The pinning logic moved out of the `deploy-examples.yml` heredoc into `scripts/render_example_requirements.py` first, unchanged, so it can also be run by hand before `lk agent deploy`:

```bash
python scripts/render_example_requirements.py examples/hotel_receptionist --ref main
```

## Verified

- `uv sync --all-extras --dev --dry-run` now includes `apsw`, `aiocsv` and `duckduckgo-search`.
- `uv.lock` records each example as `source = { virtual = ... }` with `livekit-agents` as `editable = "livekit-agents"` — local SDK changes are still what the examples run against.
- `lk agent dockerfile` in an example now reports `Detected agent language [python.uv]`, was `python.pip`.
- The render script reproduces the previous CI output for all 9 examples.
- `uv lock --check`, ruff and `mypy --strict` clean. Ruff still resolves the root config for example files, since the new `pyproject.toml`s carry no `[tool.ruff]`.
- The deploy preflight still passes: the CLI's static version check prefers `pyproject.toml` over `requirements.txt`, and it reads only `.dockerignore`, never `.gitignore`, so the rendered file still uploads.

## Notes

- **No per-example `uv.lock`.** One would pin `livekit-agents` to a path source, so an isolated Docker build context still could not `uv sync --locked` — 9 lockfiles to keep in sync for no deploy benefit.
- **Dockerfiles keep `pip install -r requirements.txt`.** The upstream `python.uv` Dockerfile template does `COPY pyproject.toml uv.lock` + `uv sync --locked`, which needs a self-contained project; a workspace member is not one. All 8 stay byte-identical, with a comment pointing at the render script.
- **`livekit-agents[evals]` dropped** from avatar, frontdesk, homepage and hotel_receptionist. There is no `evals` extra in `livekit-agents` — pip silently ignored it, uv rejects it.
- `cd examples/foo && uv sync` no longer works standalone, and a first sync pulls the whole workspace. `examples/README.md` already documented root `uv sync --all-extras --dev` plus `uv run`; the three example READMEs that still said `pip install -r requirements.txt` now match.


examples/hotel_receptionist/Dockerfile
```diff
@@ -26,16 +27,21 @@ RUN apt-get update && apt-get install -y \
...
 
 WORKDIR /app
 USER appuser
 
-COPY requirements.txt ./
-RUN pip install --user --no-cache-dir -r requirements.txt
+# The example directory is the whole build context, so pyproject.toml has to
+# resolve on its own here — no workspace, no lockfile. A deploy from this repo
+# repoints the livekit-* dependencies at the ref being deployed first:
+#   python scripts/pin_example_to_ref.py examples/<name> --ref <git-ref>
```
theomonnom: Let's remove those comments, it's not relevant for users IMO 
```suggestion
```

.github/pin_example_to_ref.py
theomonnom, line 1: I would move this file inside: 
https://github.com/livekit/agents/tree/main/.github


Because the users shouldn't really worry about this script, it's more like an internal thing for us

theomonnom (approved):

.github/pin_example_to_ref.py
u9g: done

PR #6652, opened 2026-07-31
feat(xai): support session-managed STT keyterms

u9g:
## Summary

- advertise xAI STT keyterm support to the AgentSession context pipeline
- merge explicit user keyterms with framework-managed terms without losing either source
- enforce xAI limits of 100 terms and 50 characters, prioritizing explicit user terms
- defer session-keyterm reconnects until the active utterance ends

Follow-up to #6587, which exposed the raw xAI `keyterm` WebSocket option.

## Validation

- `pytest --plugin xai` (6 passed)
- `ruff format --check` and `ruff check` on changed files
- repository-wide strict mypy check (625 source files)

### DIFF
--- a/livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/stt.py
+++ b/livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/stt.py
@@ -81,15 +81,17 @@ def _merge_keyterms(user_keyterms: list[str], session_keyterms: list[str]) -> li
             _MAX_KEYTERM_LENGTH,
         )
 
-    remaining = _MAX_KEYTERMS - len(user_keyterms)
-    if len(valid_session_keyterms) > remaining:
+    # session keyterms arrive oldest-first, so the tail holds the freshest detections
+    overflow = len(valid_session_keyterms) - (_MAX_KEYTERMS - len(user_keyterms))
+    if overflow > 0:
         logger.warning(
-            "ignored %d xAI session keyterms beyond the %d-term limit",
-            len(valid_session_keyterms) - remaining,
+            "dropped the %d oldest xAI session keyterms beyond the %d-term limit",
+            overflow,
             _MAX_KEYTERMS,
         )
+        valid_session_keyterms = valid_session_keyterms[overflow:]
 
-    return [*user_keyterms, *valid_session_keyterms[:remaining]]
+    return [*user_keyterms, *valid_session_keyterms]
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/stt.py
```diff
@@ -244,6 +292,8 @@ def update_options(
             self._opts.smart_turn_timeout = smart_turn_timeout
 
         if is_given(keyterm):
+            self._user_keyterms = _validate_user_keyterms(keyterm)
```
longcw: move this validation to the top so the options are not partially updated?

livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/stt.py
```diff
@@ -49,6 +49,47 @@
...
+        raise ValueError(f"xAI keyterms must contain at most {_MAX_KEYTERM_LENGTH} characters each")
+    return unique
+
+
+def _merge_keyterms(user_keyterms: list[str], session_keyterms: list[str]) -> list[str]:
+    seen = set(user_keyterms)
+    valid_session_keyterms: list[str] = []
+    invalid_count = 0
+    for term in session_keyterms:
+        if len(term) > _MAX_KEYTERM_LENGTH:
```
longcw, line 71: this truncates the newest detected terms, we should drop the oldest instead.

u9g: done https://github.com/livekit/agents/pull/6652/changes/7113e2841054e2c45b8b323b1bf191bf159ec969

livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/stt.py
u9g: done https://github.com/livekit/agents/pull/6652/changes/7e1a08c0a24f516c2c6d2530ce5f1c05e2c3ee83

livekit-plugins/livekit-plugins-xai/livekit/plugins/xai/stt.py
```diff
@@ -259,6 +313,23 @@ def update_options(
                 keyterm=keyterm,
             )
 
+    def _update_session_keyterms(self, keyterms: list[str]) -> None:
+        if keyterms == self._session_keyterms:
+            return
+
+        self._session_keyterms = list(keyterms)
+        merged = _merge_keyterms(self._user_keyterms, self._session_keyterms)
```
longcw, line 321: `list(keyterms)` is `KeytermDetector.keyterms` which is `[*self._static_terms, *self._detected_terms]`. Slicing off the head therefore eats the user's declared static terms first. Maybe we need to reverse the `self._detected_terms` in the detector and eat the tail.

longcw (commented):
looks good to me, one issue with the order of the keyterms

PR #6662, opened 2026-08-01
fix(voice): avoid dropping turns and resuming early while a speech is paused

longcw:
Three fixes around a paused agent speech: the turn gate, the overlap the barge-in verdict comes from, and the resume timer.

### Turns silently dropped (realtime without STT)

**a. The gate treated a paused speech as still deciding.** `on_end_of_turn` dropped a turn whenever the agent had a live, uninterrupted speech and no interruption had been flagged, on the assumption that a verdict might still arrive. Once the speech is paused that assumption is permanently false — pausing silences the agent and ends the overlap, so no verdict is coming — and the user was ignored for a whole turn while the agent resumed mid-sentence.

Only a confirmed verdict suppresses a turn now. An unjudged overlap commits, so a barge-in verdict that lands late can interrupt the agent on what may have been a backchannel — an unwanted interruption is audible, so the user hears the agent stop and carries on, while a discarded turn is invisible and leaves them repeating themselves to an agent that never registered them.

**b. Pausing tore down the overlap that verdict comes from.** Pausing signalled end-of-agent-speech to the adaptive detector, which reset its stream and closed the overlap window. The verdict is emitted when the user's speech ends, so with the window gone none ever arrived: no late detection could land, and every paused backchannel would interrupt the agent.

A pause no longer ends the detector's overlap and a resume no longer restarts its stream. The user's speech ending closes it instead, which emits the verdict and stops inference.

### Resume racing the turn decision

The resume timer counts `false_interruption_timeout` from the VAD `END_OF_SPEECH` event, while the turn commits at `last_speaking_time + endpointing_delay` — measured from before the VAD silence window. When the detector reads the pause as mid-utterance that delay becomes `max_delay`, so with the shipped defaults the resume lands 0.25s before the commit and the agent is cut off right after resuming.

An open turn decision now owns the paused speech: the timer defers to it and resumes only once the turn is dropped, while a committed turn cancels the resume synchronously from `on_end_of_turn`. The elapsed timeout counts toward that wait, so a resume with no decision pending keeps its current latency.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -4333,6 +4333,13 @@ class AgentActivity(RecognitionHooks):
         if not self._paused_speech:
             return
 
+        # the pause withheld end-of-agent-speech for a resume; this ends the agent turn instead.
+        # the audio stopped when it was paused, so there is no playout left to wait for
+        if self._audio_recognition:
+            self._audio_recognition._on_end_of_agent_speech(
+                ignore_user_transcript_until=time.time()
+            )
+
         if (
             interrupt
             and not self._paused_speech.handle.interrupted
@@ -4345,12 +4352,6 @@ class AgentActivity(RecognitionHooks):
             if self._paused_speech.handle._generations:
                 await self._paused_speech.handle._wait_for_generation()
         self._paused_speech = None
-
-        # the pause withheld end-of-agent-speech for a resume; this ends the agent turn instead
-        if self._audio_recognition:
-            self._audio_recognition._on_end_of_agent_speech(
-                ignore_user_transcript_until=time.time()
-            )
 
         if (
             self._session.options.interruption["resume_false_interruption"]
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2319,24 +2319,18 @@ class AgentActivity(RecognitionHooks):
             self._cancel_preemptive_generation()
             return False
 
-        # avoid interruption if backchannel is detected with realtime model
+        # avoid interruption if backchannel is detected with realtime model. only a confirmed
+        # verdict suppresses a turn: an unjudged overlap is committed, since interrupting the
+        # agent is recoverable where silently discarding what the user said is not
         if (
             self.stt is None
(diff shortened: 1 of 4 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -4282,9 +4281,39 @@ def _on_false_interruption() -> None:
...
+            # or drops it, and only then is the interruption known to be false
+            eot_task = (
+                self._audio_recognition._end_of_turn_task if self._audio_recognition else None
+            )
+            if eot_task is not None and not eot_task.done():
+                self._false_interruption_pending = True
+                eot_task.add_done_callback(_on_turn_settled)
+                return
+
+            _on_false_interruption()
```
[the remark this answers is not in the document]
longcw: The description is accurate; the label "bug" is what's wrong.

Waiting for the eou task is the fix, not a side effect of it. Both pause kinds mean the same thing — we don't yet know whether the user took the floor — so neither should un-mute the agent before the decision lands.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -4318,6 +4346,12 @@ async def _cancel_speech_pause(
                 await self._paused_speech.handle._wait_for_generation()
```
chenghao-mou, line 4353: should we call `_on_end_of_agent_speech` before calling `_wait_for_generation()` here?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -567,7 +572,10 @@ def _on_end_of_overlap_speech(
         speaking (the user may still be talking), in which case the synthesized verdict
         is inconclusive and must not be treated as a confirmed backchannel.
         """
-        if not self._adaptive_interruption_active or not self._agent_speaking:
+        # an open overlap outlives a paused agent speech, so it can still be closed here
+        if not self._adaptive_interruption_active or not (
+            self._agent_speaking or self._overlap_in_current_turn
```
chenghao-mou: do we need `self._overlap_in_current_turn` here if there are multiple overlap speeches in a turn (backchannel 1, backchannel 2, and interruption), should we introduce another flag `_overlap_open` instead?

davidzhao (approved):
lgtm!

PR #6663, opened 2026-08-02
fix(aws): count cached tokens in Bedrock's prompt_tokens

arthi-arumugam-git:
## The contract

`CompletionUsage` states what its own fields mean:

```python
prompt_tokens: int
"""The number of input tokens used (includes cached tokens)."""
...
total_tokens: int
"""The total number of tokens used (completion + prompt tokens)."""
```

The Bedrock path met neither.

```python
completion_tokens=metadata["usage"]["outputTokens"],
prompt_tokens=metadata["usage"]["inputTokens"],      # net of the cache
total_tokens=metadata["usage"]["totalTokens"],       # counts the cache
prompt_cached_tokens=metadata["usage"]["cacheReadInputTokens"] ...
```

Bedrock reports `inputTokens` net of the cache and puts the cached tokens in their own
`cacheReadInputTokens` and `cacheWriteInputTokens` fields. Assigning `inputTokens` straight
through therefore leaves every cached token out of `prompt_tokens`, while `total_tokens` is
taken from the provider and does count them. The two disagree inside the same object.

`cache_creation_tokens` and `cache_read_tokens` were also never populated, though Bedrock
supplies both.

## The anthropic plugin already does this

Same models, served directly rather than through Bedrock:

```python
# https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#tracking-cache-performance
prompt_token = self._input_tokens + self._cache_creation_tokens + self._cache_read_tokens
```

So the correct behaviour is settled in this repo; the Bedrock path just diverges from it.
This brings the two in line.

## Why it is not marginal

In a voice agent the system prompt and the conversation history are re-read from cache on
every turn, so cache reads come to dominate the input count within a few turns. A turn billed
for 2100 input tokens with 2000 of them cached reported 100. Nothing raises; `ModelUsageCollector`
just sums a number that is quietly low.

## Also a latent crash

The cache fields were read with `in` and passed through unchanged, so a `cacheReadInputTokens`
of `null` reached `prompt_cached_tokens`, which is a non-optional `int`. They are now coerced
to `0`.

## Tests

`livekit-plugins/livekit-plugins-aws/tests/test_llm_usage_cached_tokens.py`, five tests, four
of which fail on `main`:

```
FAILED test_cache_reads_are_included_in_prompt_tokens
FAILED test_cache_writes_are_included_too
FAILED test_total_equals_prompt_plus_completion
FAILED test_null_cache_fields_are_treated_as_zero
```

The fifth pins that an uncached request is completely unchanged.

`ruff format` and `ruff check` clean at line-length 100.

## Note

I found this coming from the same fix in Pipecat, where the AWS and Anthropic services had
the identical shape: https://github.com/pipecat-ai/pipecat/pull/5163. Happy to also align
`total_tokens` handling across the other plugins if you want that as a follow-up, but I kept
this to the one provider that diverges from your own anthropic implementation.

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
@@ -287,18 +287,24 @@ class LLMStream(llm.LLMStream):
                 logger.warning(f"aws bedrock llm: unknown chunk type: {chunk}")
 
         elif "metadata" in chunk:
-            metadata = chunk["metadata"]
+            usage = chunk["metadata"]["usage"]
+            # Bedrock reports inputTokens net of the cache and puts the cached tokens in
+            # their own fields, so they have to be added back for prompt_tokens to mean what
+            # CompletionUsage says it means ("includes cached tokens"). This mirrors what the
+            # anthropic plugin already does for the same models served directly.
+            cache_read_tokens = usage.get("cacheReadInputTokens") or 0
+            cache_creation_tokens = usage.get("cacheWriteInputTokens") or 0
+            prompt_tokens = usage["inputTokens"] + cache_read_tokens + cache_creation_tokens
+            completion_tokens = usage["outputTokens"]
             return llm.ChatChunk(
                 id=request_id,
                 usage=llm.CompletionUsage(
-                    completion_tokens=metadata["usage"]["outputTokens"],
-                    prompt_tokens=metadata["usage"]["inputTokens"],
-                    total_tokens=metadata["usage"]["totalTokens"],
-                    prompt_cached_tokens=(
-                        metadata["usage"]["cacheReadInputTokens"]
-                        if "cacheReadInputTokens" in metadata["usage"]
-                        else 0
-                    ),
+                    completion_tokens=completion_tokens,
+                    prompt_tokens=prompt_tokens,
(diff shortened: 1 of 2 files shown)


arthi-arumugam-git:
Good catch from Devin Review, and it was right. `tests/conftest.py` scans every `test_*.py` for a `pytest.mark.<category>` and aborts collection for any module without one, so the new file would have failed the whole run rather than just itself. I had been running the file directly, which skips the root conftest, so I never saw it.

`pytestmark = pytest.mark.unit` added, matching the other plugin test modules. 5 passed, ruff clean.

davidzhao (approved):

arthi-arumugam-git:
A short follow-up, and a correction on my side.

The same shape was in four other codebases that share no code with this one, so I packaged the check as a tool. The first published version of it flagged this merged fix as a defect.

That was my bug, not yours. Bedrock spells cache creation `cacheWriteInputTokens` and the checker only knew Anthropic's `cache_creation_input_tokens`, so the line here that reads both counters correctly looked like it read only one. A checker that flags the fix is worse than no checker, so it now knows the alias.

`livekit/agents` at HEAD reports nothing across 736 files.

https://github.com/arthi-arumugam-git/cachecheck, if it is worth a CI job. `uvx cachecheck .`, no dependencies, exit code 1 on a finding.

Issue #6664, opened 2026-08-02
Post-Call CRM & Analytics Telemetry Collector for livekit/agents


abhishektangudu:
### Feature Type

Nice to have

### Feature Description

When building enterprise GTM voice agents (e.g., inbound sales discovery, lead qualification, or support bots), syncing post-call session data to systems of record (Salesforce, BigQuery) is a core requirement.

While LiveKit offers OpenTelemetry (for OTLP trace spans to SigNoz/Langfuse) and Server Webhooks (for room lifecycle events), there is currently a gap for developers who need a single, structured JSON document containing full transcripts + @function_tool execution records to send to business webhooks. Server webhooks do not contain transcripts or tool logs, and OpenTelemetry spans are split across micro-events designed for APM tools rather than CRM ingestion.

Describe the solution you'd like: Add a lightweight, opt-in PostCallTelemetryCollector and standardized PostCallReport schema to livekit.agents.beta.gtm_telemetry.

Placement note: The module lives in beta/ (the repo's established opt-in staging area for features like EndCallTool and workflows) rather than the top-level telemetry/ package, which is reserved for OTel/Cloud plumbing with private exports. This avoids expanding the stable public API surface before the feature graduates.

Key capabilities:

Automatically records transcript turns (via conversation_item_added), @function_tool executions (via tool_execution_updated + function_tools_executed — with self-timed latency, arguments, results, and errors), and session metrics (via metrics_collected).
Generates a clean, serializable PostCallReport JSON payload with float-epoch timestamps (matching the repo-wide created_at: float = Field(default_factory=time.time) convention — avoids the deprecated datetime.utcnow() and stays deterministic under the virtual-time test harness).
Built-in resilient webhook dispatcher with HMAC-SHA256 signing (X-LiveKit-Signature header) and exponential backoff (max 3 attempts, 5xx + network retries, no retry on 4xx).
Standardized adapters for HubSpot Engagements API & Salesforce Activity Tasks (pure payload builders — no network or CRM SDK dependencies).
How this complements existing LiveKit features:

Does not replace OpenTelemetry (OTel continues to trace micro-spans).
Does not replace Server Webhooks (Server webhooks continue handling room lifecycle).
Provides self-hosted and cloud users with an application-level GTM summary payload.
Works with or without a JobContext, on both the pipeline (STT→LLM→TTS) and real-time paths.
I have implemented a fully tested solution and am ready to submit a PR if this aligns with the framework roadmap!

[...]

PR #6666, opened 2026-08-02
Add beta post-call CRM/analytics telemetry collector (#6664)

samanyugoyal2010:
Adds livekit.agents.beta.gtm_telemetry: an opt-in collector that builds a deterministic PostCallReport (transcript, tool executions, metrics summary) from an AgentSession, a signed webhook dispatcher with bounded retries, and pure HubSpot/Salesforce payload adapters.

Reuses existing session internals (session.history, session.usage, session._recorded_events) instead of live event listeners, so attach() timing doesn't affect correctness and duplicate/out-of-order event delivery is a non-issue. Transcript and per-turn latency are sourced from ChatMessage/ChatMessage.metrics rather than the deprecated metrics_collected event.

### DIFF
--- a/examples/voice_agents/gtm_telemetry_agent.py
+++ b/examples/voice_agents/gtm_telemetry_agent.py
@@ -1,0 +1,105 @@
+import logging
+import os
+
+from dotenv import load_dotenv
+
+from livekit.agents import (
+    Agent,
+    AgentServer,
+    AgentSession,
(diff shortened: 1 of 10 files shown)


samanyugoyal2010:
@davidzhao if this PR could be merged or given feedback i believe everything is resolved let me know if their is something to do on my end

PR #6677, opened 2026-08-03
feat: add DuplexModel for full-duplex speech models

longcw:
A full-duplex speech model listens and speaks at the same time. Its audio streams continuously whether or not it is speaking, it decides for itself when to talk, and it can be neither interrupted nor truncated.

That does not fit `RealtimeModel`, which is built around the response: a bounded object the client triggers, cancels, truncates and commits against. A duplex model has no turn at all. Its output is one unbroken stream that neither begins nor ends with a response, so there is nothing for that surface to name. Server-side turn detection does not close the gap, because it only decides when a response starts.

`DuplexModel` / `DuplexSession` give plugin authors a surface carrying only what such a model can actually do. `DuplexRealtimeAdapter` finds the agent's turns in that stream and runs one inside an `AgentSession`.

Why an adapter:

- **It cuts a turnless stream into turns.** The adapter gates the continuous output against the model's own noise floor and presents each stretch of audible speech as a generation, which is the shape a realtime session needs. Output the model never transcribes — backchannels, breath, laughter — still plays, so nothing is discarded to fit a turn-shaped API.
- **After the adapter, the framework sees an ordinary realtime session.** Turns, transcripts and tool calls arrive as `GenerationCreatedEvent`s, so `SpeechHandle`, speech scheduling, `drain()` and `agent_state` need no duplex-specific path. The only change outside the adapter is one capability flag, `paced_audio_output`: the model's audio arrives at playback pace, so nothing is buffered on our side and the model's own stop ends playout. It defaults to current behaviour.
- **It leaves room for the duplex model.** Turn boundaries, transcript timing and frame tagging may vary between providers, and several are optional; the adapter degrades to whatever a given model reports, and the plugin surface stays honest either way.

`AgentSession(llm=...)` and `Agent(llm=...)` accept a `DuplexModel` directly and wrap it internally. `Agent.duplex_session` reaches the plugin's own session, where a provider puts what the abstraction does not carry.


livekit-agents/livekit/agents/llm/duplex_adapter.py
```diff
@@ -0,0 +1,514 @@
...
+        ]:
+            logger.warning(
+                "duplex context is append-only; the model keeps what it has been told",
+                extra={"item_ids": edited},
+            )
+        if new_items := [
+            item for _, item_id in diff.to_create if (item := chat_ctx.get_by_id(item_id))
+        ]:
+            await self._duplex._append_items(new_items)
+        self._chat_ctx = chat_ctx
```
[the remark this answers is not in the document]
longcw: Removal already triggers the warning "duplex context is append-only; the model keeps what it has been told", so the app is told at the point of misuse. Only remove-then-re-add triggers the duplicate issue, and it's already a misuse.

livekit-agents/livekit/agents/llm/duplex_adapter.py
```diff
@@ -0,0 +1,514 @@
...
+            self._fragments and self._audio_ms - self._waiting_since_ms >= _UNCLAIMED_TRANSCRIPT_MS
+        ):
+            logger.error(
+                "duplex transcript outlived the audio it describes",
+                extra={"text": "".join(f.text for f in self._fragments)},
+            )
+            burst = self._open_burst()
+            while self._fragments:
+                burst.attach(self._fragments.popleft())
+            self._close_burst()
```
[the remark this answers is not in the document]
longcw: Late transcripts are the designed fallback.

livekit-agents/livekit/agents/llm/duplex_adapter.py
```diff
@@ -0,0 +1,514 @@
...
+        rms = float(np.sqrt(np.mean(np.square(samples)))) / 32768.0 if samples.size else 0.0
+
+        self._history.append((rms, frame.duration))
+        self._history_duration += frame.duration
+        while self._history_duration > self._window and len(self._history) > 1:
+            self._history_duration -= self._history.popleft()[1]
+        floor = max(min(level for level, _ in self._history), _SILENCE_FLOOR)
+
+        if not self._open:
+            if rms > floor * self._open_ratio:
```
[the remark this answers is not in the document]
longcw: using FixedGate instead

livekit-agents/livekit/agents/llm/duplex_adapter.py
```diff
@@ -0,0 +1,629 @@
...
+            if self._quiet >= self._min_silence_duration:
+                self._open = False
+        else:
+            self._quiet = 0.0
+
+        return self._open
+
+
+@dataclass
+class _Burst:
```
chenghao-mou, line 181: nit: how about `_Utterance`?

livekit-agents/livekit/agents/llm/duplex_adapter.py
```diff
@@ -0,0 +1,629 @@
...
+                MessageGeneration(
+                    message_id=burst.id,
+                    text_stream=burst.text_ch,
+                    audio_stream=burst.audio_ch,
+                    modalities=modalities,
+                )
+            )
+        return burst
+
+    def _close_burst(self) -> None:
```
chenghao-mou, line 447: should we deactivate the gate here?

livekit-agents/livekit/agents/llm/duplex.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @abstractmethod
+    def push_audio(self, frame: rtc.AudioFrame) -> None: ...
+
+    def push_video(self, frame: rtc.VideoFrame) -> None:
+        """Feed a video frame to the model; ignored by models without video input."""
+        pass
+
+    @abstractmethod
+    async def aclose(self) -> None: ...
```
chenghao-mou: should we release `_configured` here?

livekit-agents/livekit/agents/llm/duplex_adapter.py
```diff
@@ -0,0 +1,629 @@
...
+        # the framework's reply task handles RealtimeError; a cancelled future would end it
+        if self._pending_reply is not None and not self._pending_reply.done():
+            self._pending_reply.set_exception(RealtimeError(reason))
+
+    async def aclose(self) -> None:
+        await aio.cancel_and_wait(self._segment_atask)
+        self._close_burst()
+        self._fail_pending_reply("the session closed before the model replied")
+        with contextlib.suppress(Exception):
+            await self._duplex.aclose()
```
chenghao-mou, line 636: should we unregister the adapter’s callbacks here?

livekit-agents/livekit/agents/llm/duplex.py
```diff
@@ -0,0 +1,222 @@
...
+            await self._update_instructions(instructions)
+
+        if is_given(chat_ctx):
+            await self._append_items(chat_ctx.items)
+
+        if is_given(tools):
+            await self._update_tools(tools)
+
+        # only a complete configuration releases a model that cannot be reconfigured later
+        self._configured.set()
```
chenghao-mou, line 209: from Codex:

> Configuration errors propagate out of startup before `_started` becomes true, so `AgentSession.aclose()` then skips cleanup.

chenghao-mou (approved):

tinalenguyen (approved):

Fyko:
yay

PR #6680, opened 2026-08-03
fix(tts): probe a skipped provider so the streamed path can recover it

### LINKED ISSUE #6678: tts.FallbackAdapter never recovers a failed provider on the streamed path
### Summary

On `tts.FallbackAdapter`, a provider marked unavailable is **never re-probed on the streamed path**. `FallbackSynthesizeStream._try_recovery` returns before starting a recovery probe, because the token list it needs is still empty at the point it is called. The chunked path recovers the same provider on the very next request, so a process that only uses `adapter.stream()` stays pinned to its fallback provider for the rest of its life after a single transient outage.

### Where

`FallbackSynthesizeStream._run()` creates the task that fills `_pushed_tokens`, then loops over providers:

```python
input_task = asyncio.create_task(_forward_input_task())   # fills self._pushed_tokens

try:
    for i, tts in enumerate(self._fallback_adapter._tts_instances):
        tts_status = self._fallback_adapter._status[i]
        if tts_status.available or all_failed:
            ...                       # the only branch containing an await
        self._try_recovery(tts)
```

and `_try_recovery` (line 444) begins:

```python
retry_text = self._pushed_tokens.copy()
if not retry_text:
    return
```

For a provider that is **already** unavailable, the `if tts_status.available or all_failed:` branch is skipped, so nothing between the `create_task` and the `_try_recovery(tts)` call awaits. `_forward_input_task` has not been scheduled yet, `_pushed_tokens` is empty, and the probe never starts.

The guard is unique to this method. The three siblings all start a probe unconditionally, gated only on the recovery slot:

| method | line | starts a probe unconditionally? |
| --- | --- | --- |
| `tts` `FallbackChunkedStream._try_recovery` | 181 | yes |
| `tts` `FallbackSynthesizeStream._try_recovery` | 444 | **no — bails on empty `_pushed_tokens`** |
| `stt` `FallbackRecognizeStream._try_recovery` | 192 | yes |
| `stt` streamed `_try_recovery` | 423 | yes |

### Reproduction

Run from the repo root (uses the existing `tests/fake_tts.py` harness):

```python
import asyncio

async def drive(adapter, text, streamed):
    if streamed:
        stream = adapter.stream()
        stream.push_text(text)
        stream.flush()
        stream.end_input()
        ctx = stream
    else:
        ctx = adapter.synthesize(text)
    try:
        async with ctx as s:
            async for _ in s:
                pass
    except Exception as exc:
        return type(exc).__name__
    return "ok"


async def scenario(streamed):
    kind = "streamed" if streamed else "chunked"
    primary = FakeTTS(fake_exception=APIConnectionError("transient outage"))
    backup = FakeTTS(fake_audio_duration=1.0, sample_rate=48000)
    adapter = FallbackAdapter([primary, backup], max_retry_per_tts=0)
    status = adapter._status[0]

    # Knock the primary out, so the adapter marks it unavailable.
    await drive(adapter, "first", streamed)
    assert status.available is False, "setup failed: the primary was never marked unavailable"

    # The outage is over: the primary is healthy from here on.
    primary.update_options(fake_exception=None, fake_audio_duration=1.0)

    # Let the recovery probe from the first request finish, so the slot is free.
    for _ in range(20):
        await asyncio.sleep(0.1)
        if status.recovering_task is not None and status.recovering_task.done():
            break
    assert status.recovering_task is not None and status.recovering_task.done(), (
        "setup failed: the first recovery probe never started or is still in flight"
    )

    # Give the adapter four more requests to notice the primary is back.
    for i in range(4):
        await drive(adapter, f"request {i + 2}", streamed)
        for _ in range(15):
            await asyncio.sleep(0.1)
            if status.available:
                break
        if status.available:
            print(f"{kind}: recovered on request {i + 2}")
            return
    print(f"{kind}: still unavailable after 4 healthy requests")


async def main():
    await scenario(streamed=True)
    await scenario(streamed=False)


asyncio.run(main())
```

Output on `bbf163f`:

```
streamed: still unavailable after 4 healthy requests
chunked: recovered on request 2
```

The two scenarios are identical apart from `adapter.stream()` vs `adapter.synthesize()`, and the setup assertions fail loudly if the provider was never marked unavailable or if the first probe is still occupying the slot — so the chunked line is a genuine control rather than a differently-configured run.

### Mechanism confirmation

Instrumenting `FallbackSynthesizeStream._try_recovery` over three streamed requests:

```
call 1: pushed_tokens=['first']  slot_before_done=None  started_probe=True
call 2: pushed_tokens=[]         slot_before_done=True  started_probe=False
call 3: pushed_tokens=[]         slot_before_done=True  started_probe=False
available at end: False
```

Call 1 is the request during which the provider was still `available`: it failed mid-stream, after `_forward_input_task` had run, so `_pushed_tokens` is populated and a probe starts. Calls 2 and 3 are the requests where the provider was already unavailable and got skipped — `_pushed_tokens` is empty and no probe starts. The recovery slot is free in both cases (`slot_before_done=True`), so the slot check is not what blocks them.

### Impact

Any deployment using `tts.FallbackAdapter` through `stream()` — which is what `AgentSession` does for streaming TTS — permanently loses its primary provider after one transient failure. There is no log line for it: the skipped recovery is silent, so the only visible symptom is that traffic never returns to the primary. `aclose()` and process restart are the only ways back.

### Note on `_TTSStatus.recovering_task`

Related but separate: both `_try_recovery` implementations write to the single `_TTSStatus.recovering_task` slot, whereas `_STTStatus` keeps `recovering_recognize_task` and `recovering_stream_task` apart for exactly this reason. With the empty-token bug fixed, a chunked recovery in flight would make the streamed path's slot check bail (and vice versa). I can address that in the same PR if you'd like, or leave it — it only becomes reachable once the above is fixed.

### Fix direction

Happy to open a PR. The narrow fix is for the streamed `_try_recovery` to not depend on `_pushed_tokens` being populated — a recovery probe does not need the caller's real text, and the STT streamed sibling probes with a fresh stream rather than replaying input. If you would rather keep replaying the pushed text when it is available, awaiting a scheduling step (or draining `_input_ch` into `_pushed_tokens` before the provider loop) would also close it, but the unconditional probe matches the other three adapters.

Environment: `bbf163f`, Python 3.12.10, Windows.

LHMQ878:
Fixes #6678.

### Problem

On `tts.FallbackAdapter`, a provider marked unavailable is never re-probed on the streamed path, so it can never recover. A single transient outage pins the adapter to its fallback provider for the life of the process, silently — the skipped recovery logs nothing.

`FallbackSynthesizeStream._run()` creates the task that fills `_pushed_tokens`, then loops over providers:

```python
input_task = asyncio.create_task(_forward_input_task())   # fills self._pushed_tokens

for i, tts in enumerate(self._fallback_adapter._tts_instances):
    tts_status = self._fallback_adapter._status[i]
    if tts_status.available or all_failed:
        ...                       # the only branch containing an await
    self._try_recovery(tts)
```

A provider that is **already** unavailable skips that branch, so nothing awaits between the `create_task` and the `_try_recovery(tts)` call. `_forward_input_task` has not been scheduled, `_pushed_tokens` is empty, and `_try_recovery`'s `if not retry_text: return` fires before it ever reaches the recovery-slot check.

### Fix

A recovery probe only needs to learn whether the provider answers; it does not need the caller's real text. So the empty case now probes with a short placeholder instead of bailing out.

This brings the method in line with its three siblings, all of which probe unconditionally and gate only on the recovery slot:

| method | line | probes unconditionally? |
| --- | --- | --- |
| `tts` `FallbackChunkedStream._try_recovery` | 181 | yes |
| `tts` `FallbackSynthesizeStream._try_recovery` | 444 | **no — fixed here** |
| `stt` `FallbackRecognizeStream._try_recovery` | 192 | yes |
| `stt` streamed `_try_recovery` | 423 | yes |

The `stt` streamed one is the closest precedent: it opens a fresh `stt.stream(...)` for the probe rather than replaying pushed input. Real pushed text is still preferred here when it is available, so the recovery request stays representative in the case that already worked.

### Verification

Measured on `bbf163f` with the reproduction from #6678 — a primary that fails once, then is healthy, driven identically through `stream()` and `synthesize()`:

```
before:  streamed: still unavailable after 4 healthy requests
         chunked:  recovered on request 2
after:   streamed: recovered on request 2
         chunked:  recovered on request 2
```

The chunked line is a control: same fakes, same timing, only the entry point differs. The reproduction also asserts that the provider really was marked unavailable and that the first probe has settled, so it fails loudly rather than passing vacuously if the setup drifts.

Instrumenting `_try_recovery` across three streamed requests pins the mechanism:

```
call 1: pushed_tokens=['first']  slot_before_done=None  started_probe=True
call 2: pushed_tokens=[]         slot_before_done=True  started_probe=False
call 3: pushed_tokens=[]         slot_before_done=True  started_probe=False
```

Call 1 is the request during which the provider was still available and failed mid-stream, after `_forward_input_task` had run. Calls 2 and 3 are the ones where it was already unavailable and got skipped. The slot is free in both (`slot_before_done=True`), so the slot check is not what blocks them.

- `test_tts_stream_recover` — the streamed counterpart of the existing chunked `test_tts_recover`, whose absence is why this survived. Confirmed failing on the unmodified tree (`fake1 should have recovered`) and passing with the fix.
- `tests/test_tts_fallback.py`, `tests/test_stt_fallback.py`, `tests/test_llm_fallback.py`: 16 passed.
- `ruff check` and `ruff format --check` clean on both changed files.
- `tests/test_tts.py` cannot be collected in my environment (`ModuleNotFoundError: No module named 'dotenv'`), unrelated to this change.

Environment: Python 3.12.10, Windows.

### Not included

Both `_try_recovery` implementations write to the single `_TTSStatus.recovering_task` slot, whereas `_STTStatus` deliberately keeps `recovering_recognize_task` and `recovering_stream_task` separate. With this fix in place, a chunked recovery in flight will make the streamed path's slot check bail, and vice versa. That is a real but distinct issue, only reachable once the above is fixed — noted in #6678, and I'm happy to add the second slot here or in a follow-up, whichever you prefer.

### DIFF
--- a/livekit-agents/livekit/agents/tts/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/tts/fallback_adapter.py
@@ -29,6 +29,10 @@ DEFAULT_FALLBACK_API_CONNECT_OPTIONS = APIConnectOptions(
     max_retry=0, timeout=DEFAULT_API_CONNECT_OPTIONS.timeout
 )
 
+# synthesized by a recovery probe when no real text is available to replay; the audio is discarded,
+# only whether the request succeeds matters
+_RECOVERY_PROBE_TEXT = "."
+
 
 @dataclass
 class _TTSStatus:
@@ -444,9 +448,12 @@ class FallbackSynthesizeStream(SynthesizeStream):
     def _try_recovery(self, tts: TTS) -> None:
         assert isinstance(self._tts, FallbackAdapter)
 
-        retry_text = self._pushed_tokens.copy()
-        if not retry_text:
-            return
+        # A provider that is already unavailable is skipped by the loop in `_run`, which reaches
+        # here without awaiting, so `_forward_input_task` has not run yet and `_pushed_tokens` is
+        # still empty. Bailing out in that case means such a provider is never probed again and
+        # can never recover. A probe does not need the caller's real text, so fall back to a short
+        # placeholder, matching the other adapters, which all probe unconditionally.
+        retry_text = self._pushed_tokens.copy() or [_RECOVERY_PROBE_TEXT]
 
         tts_status = self._tts._status[self._tts._tts_instances.index(tts)]
         if tts_status.recovering_task is None or tts_status.recovering_task.done():
--- a/tests/test_tts_fallback.py
+++ b/tests/test_tts_fallback.py
@@ -243,6 +243,48 @@ async def test_tts_recover() -> None:
     await fallback_adapter.aclose()
 
 
+async def test_tts_stream_recover() -> None:
+    # the streamed counterpart of test_tts_recover: a provider that is already unavailable is
+    # skipped by the provider loop, which reaches _try_recovery without awaiting, so
+    # _forward_input_task has not filled _pushed_tokens yet. Bailing out on the empty list meant
+    # such a provider was never probed again and could never recover.
+    fake1 = FakeTTS(fake_exception=APIConnectionError("fake1 failed"))
+    fake2 = FakeTTS(fake_audio_duration=5.0)
+
+    fallback_adapter = FallbackAdapterTester([fake1, fake2])
+
+    async with fallback_adapter.stream() as stream:
+        stream.push_text("hello test")
+        stream.end_input()
+
+        async for _ in stream:
+            pass
+
+    assert fake1.stream_ch.recv_nowait()
+    assert not fallback_adapter.availability_changed_ch(fake1).recv_nowait().available
+
+    fake1.update_options(fake_exception=None, fake_audio_duration=5.0)
+
+    # let the recovery probe started by the request above settle, so the next one is not skipped
+    # merely because the single recovering_task slot is still occupied
+    await asyncio.sleep(1.0)
+
+    # fake1 is healthy again, but the adapter only learns that by probing it. This second request
+    # is served by fake2 and must still start a probe for the skipped fake1.
+    async with fallback_adapter.stream() as stream:
+        stream.push_text("hello test")
+        stream.end_input()
+
+        async for _ in stream:
+            pass
+
+    assert (
+        await asyncio.wait_for(fallback_adapter.availability_changed_ch(fake1).recv(), 5.0)
+    ).available, "fake1 should have recovered"
+
+    await fallback_adapter.aclose()
+
+
 async def test_audio_resampled() -> None:
     fake1 = FakeTTS(sample_rate=48000, fake_exception=APIConnectionError("fake1 failed"))
     fake2 = FakeTTS(fake_audio_duration=5.0, sample_rate=16000)


LHMQ878:
Closing as requested: #6684 covers the same streamed recovery bug and now includes the post-close recovery guard. Thanks for coordinating the overlap.

PR #6683, opened 2026-08-03
fix(tts): give each fallback recovery path its own task slot

### LINKED ISSUE #6682: tts.FallbackAdapter: one shared recovering_task slot lets one path suppress the other's recovery probe
### Summary

`_TTSStatus` has a **single** `recovering_task` slot, but it is written by **both** of the TTS fallback adapter's recovery paths — `FallbackChunkedStream._try_recovery` and `FallbackSynthesizeStream._try_recovery`. Because each path guards on `if tts_status.recovering_task is None or tts_status.recovering_task.done()`, a probe in flight on one path silently suppresses probing on the other. A provider that is up again therefore stays marked unavailable for longer than it should — or, in a process that keeps one path continuously busy recovering, indefinitely.

`_STTStatus` deliberately does **not** do this: it keeps `recovering_recognize_task` and `recovering_stream_task` as separate fields, and `aclose()` cancels both. The TTS adapter is the odd one out.

### Where

```python
@dataclass
class _TTSStatus:
    available: bool
    recovering_task: asyncio.Task[None] | None   # <-- one slot, two writers
    needs_resampling: bool
```

Writers:

- `FallbackChunkedStream._try_recovery` — guard at :185, assignment at :201
- `FallbackSynthesizeStream._try_recovery` — guard at :452, assignment at :484

Compare `livekit-agents/livekit/agents/stt/fallback_adapter.py`:

```python
@dataclass
class _STTStatus:
    available: bool
    recovering_recognize_task: asyncio.Task[None] | None
    recovering_stream_task: asyncio.Task[None] | None
```

### Consequences

1. **Suppression** (confirmed below). While one path's probe is in flight, the other path's `_try_recovery` sees a non-done task in the shared slot and returns without probing. A recovered provider is not detected on that path.
2. **Overwrite / lost cancellation.** When the second path *does* pass the guard (the first task having completed), it overwrites the slot. `aclose()` only cancels whatever is in the slot at that moment, so any task the slot no longer references cannot be cancelled by the adapter. This is the same class of problem as #4921 on the LLM adapter, but reachable here without `aclose()` racing anything — the reference is simply dropped.

### Reproduction

Both scripts run against plain `bbf163f` with no other patches. **This does not depend on #6680** — the suppression is triggered entirely from the chunked path plus one `stream()` call.

`probe_slot.py`:

```python
import asyncio, logging
logging.basicConfig(level=logging.CRITICAL)

async def main() -> None:
    # fails without pushing audio, so _try_recovery is actually reached;
    # fake_exception_count must cover the in-stream retry or it is never marked unavailable
    fake1 = FakeTTS(fake_exception=APIConnectionError("fake1 down"), fake_exception_count=99)
    fake2 = FakeTTS(fake_audio_duration=1.0)
    adapter = FallbackAdapter([fake1, fake2], max_retry_per_tts=0)
    status = adapter._status[0]

    try:
        async with adapter.synthesize("hello one") as s:
            async for _ in s: pass
    except Exception: pass

    assert not status.available, "PROBE INVALID: fake1 was never marked unavailable"

    # make the next chunked probe hang, so it is unambiguously in flight
    fake1.update_options(fake_exception=None, fake_timeout=100.0)
    try:
        async with adapter.synthesize("hello two") as s:
            async for _ in s: pass
    except Exception: pass
    hanging = status.recovering_task
    assert hanging is not None and not hanging.done(), "PROBE INVALID: no in-flight probe"

    # streamed request on the SAME adapter and provider index
    try:
        async with adapter.stream() as st:
            st.push_text("hello three"); st.end_input()
            async for _ in st: pass
    except Exception: pass

    print("slot unchanged (streamed path did not probe):", status.recovering_task is hanging)

asyncio.run(main())
```

Output:

```text
after chunked request : recovering_task = <Task finished ...FallbackChunkedStream._try_recovery...>
in-flight chunked task: <Task pending ...fallback_adapter.py:189> done = False
after streamed request: recovering_task = <Task pending ...fallback_adapter.py:189>
>>> (B) SUPPRESSION: streamed path skipped probing because the chunked
    probe still occupies the single shared slot.
```

The slot still holds the *chunked* task (`fallback_adapter.py:189`) after the streamed request — the streamed path never created one.

### Control: the STT adapter, with separate slots, behaves differently

Same scenario against `stt.FallbackAdapter` (`probe_slot_stt.py`, using `FakeSTT`; note it has no `fake_exception_count`, and with `max_retry_per_stt=0` it does not need one):

```text
available after r1     : False
recovering_recognize   : pending=True
recovering_stream      : None
after streamed request :
   recovering_recognize_task = pending
   recovering_stream_task    = <Task pending ...stt/fallback_adapter.py:442>
>>> STT DID probe on the streamed path while the recognize probe was in flight.
```

So the behaviour is not a global policy that recovery should be serialized per provider — the sibling adapter explicitly allows the two paths to probe concurrently. TTS differs only because of the shared field.

### Suggested fix

Mirror `_STTStatus`: split the field into `recovering_synthesize_task` and `recovering_stream_task`, have each `_try_recovery` guard and assign its own, and cancel both in `aclose()`. That keeps the existing "one probe per path per provider" behaviour while removing the cross-path interference, and makes the two adapters structurally identical.

I mentioned this while filing #6678/#6680 as a follow-up I'd hold back rather than fold in; opening it separately now that I've confirmed it reproduces on its own at `bbf163f`. Happy to send the patch — it is a small, self-contained change, and it does not conflict with #6680 (which touches `retry_text`, not the slot). Let me know if you'd rather have it as part of that PR.

### Environment

`livekit-agents` at `bbf163f`, Python 3.12, Windows. Probes use the in-repo `tests/fake_tts.py` / `tests/fake_stt.py`; no network and no API keys.

LHMQ878:
Fixes #6682.

### Problem

`_TTSStatus` carried a single `recovering_task` slot, but it was written by **both** of the TTS fallback adapter's recovery paths — `FallbackChunkedStream._try_recovery` and `FallbackSynthesizeStream._try_recovery`. Each guards on `if tts_status.recovering_task is None or tts_status.recovering_task.done()`, which gives two problems:

1. **Suppression.** While one path's probe is in flight, the other path's `_try_recovery` finds a non-done task in the shared slot and returns without probing. A provider that is healthy again stays marked unavailable.
2. **Lost cancellation.** When the second path does pass the guard it overwrites the slot, so `aclose()` — which only cancels what the slot currently references — can no longer cancel the dropped task.

`_STTStatus` already keeps `recovering_recognize_task` and `recovering_stream_task` as separate fields and cancels both in `aclose()`. The TTS adapter was the odd one out; this makes the two structurally identical.

### Change

Split the field into `recovering_synthesize_task` / `recovering_stream_task`, point each `_try_recovery` at its own slot, and cancel both in `aclose()`. Existing behaviour — at most one recovery probe per path per provider — is unchanged; only the cross-path interference goes away.

### Test

`test_recovery_is_not_suppressed_across_paths` drives the direction that actually isolates the slot: the **streamed** path leaves a probe in flight (gated on an `asyncio.Event` rather than a timer, so nothing can race it to completion), then a **chunked** request must still probe the same provider. The discriminating assertion is provider-level rather than field-level — did `fake1` receive a chunked `synthesize()` call — so it fails for the right reason at the parent commit:

```text
bbf163f  FAILED  assert await asyncio.wait_for(fake1.synthesize_ch.recv(), 1.0)
                 "chunked path did not probe fake1 while the streamed probe was in flight"
this PR  PASSED
```

It also asserts `aclose()` cancels the still-gated streamed probe after the chunked path has written its own slot, covering problem 2.

`tests/test_tts_fallback.py tests/test_stt_fallback.py tests/test_llm_fallback.py`: 15 passed at the parent commit, 16 with this change (the one added test). `ruff check` and `ruff format --check` clean.

One thing to flag: I originally reproduced this in the chunked → streamed direction, and that reproduction was **vacuous** — the streamed path bails at `if not retry_text: return` before it ever consults the slot, so the fix doesn't change that scenario. I've corrected the issue accordingly. Notably, **#6680 removes that guard**, which un-masks the second direction; so after #6680 lands the shared slot is reachable from both paths rather than one. The two changes remain independent (#6680 touches `retry_text`, this touches the slot) and don't conflict textually — happy to sequence them in whichever order you prefer, or to fold this in if you'd rather have one PR.

Branched off `bbf163f` deliberately rather than off my #6680 branch, so the two can be reviewed and landed independently.

### DIFF
--- a/livekit-agents/livekit/agents/tts/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/tts/fallback_adapter.py
@@ -33,7 +33,11 @@ DEFAULT_FALLBACK_API_CONNECT_OPTIONS = APIConnectOptions(
 @dataclass
 class _TTSStatus:
     available: bool
-    recovering_task: asyncio.Task[None] | None
+    # One slot per recovery path, mirroring `_STTStatus`. A single shared slot let a probe in
+    # flight on one path suppress probing on the other, and let the second path to run drop the
+    # first path's task so `aclose()` could no longer cancel it.
+    recovering_synthesize_task: asyncio.Task[None] | None
+    recovering_stream_task: asyncio.Task[None] | None
     needs_resampling: bool
 
 
@@ -100,7 +104,12 @@ class FallbackAdapter(
                 logger.info(f"resampling {t.label} from {t.sample_rate}Hz to {sample_rate}Hz")
 
             self._status.append(
-                _TTSStatus(available=True, recovering_task=None, needs_resampling=needs_resampling)
+                _TTSStatus(
+                    available=True,
+                    recovering_synthesize_task=None,
+                    recovering_stream_task=None,
+                    needs_resampling=needs_resampling,
+                )
             )
 
             t.on("metrics_collected", self._on_metrics_collected)
@@ -132,8 +141,11 @@ class FallbackAdapter(
 
     async def aclose(self) -> None:
         for tts_status in self._status:
-            if tts_status.recovering_task is not None:
-                await aio.cancel_and_wait(tts_status.recovering_task)
+            if tts_status.recovering_synthesize_task is not None:
+                await aio.cancel_and_wait(tts_status.recovering_synthesize_task)
+
+            if tts_status.recovering_stream_task is not None:
+                await aio.cancel_and_wait(tts_status.recovering_stream_task)
 
         for t in self._tts_instances:
             t.off("metrics_collected", self._on_metrics_collected)
@@ -182,7 +194,8 @@ class FallbackChunkedStream(ChunkedStream):
         assert isinstance(self._tts, FallbackAdapter)
 
         tts_status = self._tts._status[self._tts._tts_instances.index(tts)]
-        if tts_status.recovering_task is None or tts_status.recovering_task.done():
+        recovering_task = tts_status.recovering_synthesize_task
+        if recovering_task is None or recovering_task.done():
 
             async def _recover_tts_task(tts: TTS) -> None:
                 try:
@@ -198,7 +211,7 @@ class FallbackChunkedStream(ChunkedStream):
                 except Exception:  # exceptions already logged inside _try_synthesize
                     return
 
-            tts_status.recovering_task = asyncio.create_task(_recover_tts_task(tts))
+            tts_status.recovering_synthesize_task = asyncio.create_task(_recover_tts_task(tts))
 
     async def _run(self, output_emitter: AudioEmitter) -> None:
         assert isinstance(self._tts, FallbackAdapter)
@@ -449,7 +462,8 @@ class FallbackSynthesizeStream(SynthesizeStream):
             return
 
         tts_status = self._tts._status[self._tts._tts_instances.index(tts)]
-        if tts_status.recovering_task is None or tts_status.recovering_task.done():
+        recovering_task = tts_status.recovering_stream_task
+        if recovering_task is None or recovering_task.done():
 
             async def _recover_tts_task(tts: TTS) -> None:
                 try:
@@ -481,4 +495,4 @@ class FallbackSynthesizeStream(SynthesizeStream):
                 except Exception:
                     return
 
-            tts_status.recovering_task = asyncio.create_task(_recover_tts_task(tts))
+            tts_status.recovering_stream_task = asyncio.create_task(_recover_tts_task(tts))
--- a/tests/test_tts_fallback.py
+++ b/tests/test_tts_fallback.py
@@ -10,10 +10,10 @@ from livekit import rtc
 from livekit.agents import APIConnectionError, APIConnectOptions, APIError, utils
 from livekit.agents.tts import TTS, AvailabilityChangedEvent, FallbackAdapter
 from livekit.agents.tts.tts import SynthesizedAudio, SynthesizeStream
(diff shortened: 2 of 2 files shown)


LHMQ878:
Verified the two PRs compose, since #6680 removes the `retry_text` guard this one sits next to:

```text
trial merge of #6680 into this branch: auto-merges cleanly in both files
  fallback_adapter.py -> retry_text guard replaced by `or [_RECOVERY_PROBE_TEXT]` (from #6680),
                         two separate slots preserved (from this PR)
tests/test_tts_fallback.py tests/test_stt_fallback.py tests/test_llm_fallback.py
  bbf163f  : 15 passed
  this PR  : 16 passed
  merged   : 17 passed   (both new tests)
```

So they are independent in both senses — no textual conflict and no behavioural interference. Land them in either order; the merge was validated in a throwaway worktree and discarded, so nothing here is rebased.

One caveat on my own numbers, for the record: my first attempt at this validation ran the merged *tests* against unmerged *source*, because the editable install in my venv resolves `livekit.agents` to the primary checkout rather than the worktree. That produced a bogus `test_tts_stream_recover` failure. The 17-passed figure above is from re-running with the merge applied in the checkout the install actually points at.

longcw (approved):

PR #3533, opened 2025-09-29
Support passing credentials to the LLM constructor

mokeko:

PR #6686, opened 2026-08-03
fix(google): release the genai client when the LLM or Gemini TTS closes

LHMQ878:
## Summary

#6643 released the genai http clients when the google **realtime session** closes. Two other components in this plugin construct their own `genai.Client` and were left out:

| component | owns a `genai.Client` | releases it on `aclose()` |
|---|---|---|
| `realtime.RealtimeSession` | yes | yes — #6643 |
| `llm.LLM` | yes (`llm.py:259`) | **no — no `aclose()` at all** |
| `beta.gemini_tts.TTS` | yes (`beta/gemini_tts.py:146`) | **no — no `aclose()` at all** |
| `LLMStream` | no, borrows it | n/a |

Both base classes implement `__aexit__ -> aclose()`, so `async with google.LLM(...)` is the documented lifecycle — and it currently leaves the clients open. `llm.LLM.aclose()` only cancels a prewarm task, and `tts.TTS.aclose()` is an empty default rather than abstract, so a plugin that owns a client and doesn't override it leaks silently with nothing to flag it.

The consequence is the one #6643 describes: the clients live until the collector runs `AsyncClient.__del__`, which does `asyncio.get_running_loop().create_task(self.aclose())` — creating pending tasks on whatever event loop happens to be running at that moment.

Measured on `bbf163fe`, driving the documented `async with` path:

```
                    aio.aclose called   _async_httpx_client
google.LLM          False               OPEN
gemini_tts.TTS      False               OPEN
```

and with this change:

```
google.LLM          True                closed
gemini_tts.TTS      True                closed
```

The convention already exists on both sides: the openai plugin's `LLM.aclose()` does `await self._client.close()`. Unlike that one, neither google class accepts an injected client — both always construct their own — so there is no `_owns_client` gate to add here. (#3533 passes *credentials*, not a client, so it doesn't change that.)

I matched #6643's `try/except` as well: `aclose()` runs on the shutdown path, so a provider-side failure there has to stay contained.

**One thing I deliberately did not change.** `aio.aclose()` closes only the *async* httpx client; `genai.Client` also holds a sync `_httpx_client` (closed by the separate, synchronous `Client.close()`), which stays open:

```
after aio.aclose() -> async closed: True | sync closed: False
```

The merged realtime fix has the same residual, so I matched it rather than diverging. If you'd like the sync client released too, that's a small follow-up across all three call sites and I'm happy to do it — but it seemed wrong to silently make these two behave differently from the realtime session.

## Test plan

New `tests/test_plugin_google_client_release.py`, modelled on #6643's `test_session_close_releases_the_genai_client`:

* `test_llm_close_releases_the_genai_client` — spies `aio.aclose`, drives `async with`, asserts the async httpx client is closed.
* `test_gemini_tts_close_releases_the_genai_client` — same for the beta TTS.
* `test_llm_close_survives_a_failing_client_close` — a client that raises on close must not break `aclose()`. **This one passes on `main` too** (there is no release call there to fail), so it guards the `try/except` against future regression rather than proving the bug — flagging that rather than implying all three are regression tests.

A/B of the first two: **2 failed on `bbf163fe` → 3 passed** with the change.

Wider verification, each compared against `main`:

* `pytest -m unit` (all collectible): **1601 passed / 4 failed** with the change vs **1599 passed / 6 failed** on `main` — the same 4 pre-existing failures either way (`test_ivr_activity` tf-idf ×3, `test_tokenizer_xml_markup` tiktoken), plus my 2 new tests failing at base. Nothing regressed.
* `tests/test_plugin_google_realtime.py`, `test_google_thought_signatures.py`, `test_tts_fallback.py` + the new file: 48 passed.
* `ruff check` and `ruff format --check` on the three touched files: clean.
* `mypy --platform linux -p livekit.plugins.google`: 1 error, in `stt.py` (a missing `types-grpcio` stub), **identical on `main`** — 0 in either file this PR touches.

Two environment notes, since they affect what I could run: `uv run` can't resolve here (`bithuman==2.7.0` has no win_amd64 wheel), so I ran `ruff`/`mypy`/`pytest` from the project venv directly; and `make type-check` wants all ~70 plugins plus the `typing` stub group installed, which this environment doesn't have — hence the plugin-scoped mypy run above. 30 test files fail collection here purely from uninstalled optional plugins (`livekit.plugins.openai`, `anthropic`, `mcp`, …), identical before and after the change.

## Issue number

Follows up #6643.


longcw (approved):
looks good to me. could you address the comment and maybe delete the tests/test_plugin_google_client_release.py?

LHMQ878:
Addressed in 2d62bd0a4: both new aclose methods now have API docstrings, and tests/test_plugin_google_client_release.py has been removed as requested. Focused ruff check and format check pass. The full uv environment still cannot resolve on Windows because bithuman 2.7.0 has no win_amd64 wheel.

PR #6690, opened 2026-08-03
fix: fallback adapters report the active instance's model/provider

Fyko:
## What

Adds a `metrics_metadata` property to the STT/TTS/LLM/RealtimeModel base classes (returns own `model`/`provider`), overridden by the fallback adapters to report the instance that most recently served a request. The three `lk.agents.turn.*` label sites in `agent_activity.py` now read it. Adapter `model`/`provider` stay `"FallbackAdapter"`/`"livekit"` for spans, logs, and error events.

## Why

Turn histograms labeled from the top-level component collapse into one `model_name="FallbackAdapter"` series with a fallback adapter — you can't tell your primary's ttfb from your fallback's. Discussed with the LiveKit team in slack; surfacing the actual provider was agreed on. Scoped to `metrics_metadata` per @longcw's review, keeping the adapter's own identity stable everywhere else.

Semantics: the primary before any traffic, the last instance to produce output after that. Recovery probes never relabel. Tests cover all four adapters, including probe-doesn't-relabel and fresh-session-resets-to-primary.

https://claude.ai/code/session_01AkHCx6Qfu5i2Z6h8Pa82gE

### DIFF
--- a/livekit-agents/livekit/agents/llm/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/fallback_adapter.py
@@ -77,16 +77,21 @@ class FallbackAdapter(
             _LLMStatus(available=True, recovering_task=None) for _ in self._llm_instances
         ]
 
+        # the instance that most recently served a request; used to label metrics & traces
+        self._active_instance: LLM = self._llm_instances[0]
+
         for llm_instance in self._llm_instances:
             llm_instance.on("metrics_collected", self._on_metrics_collected)
 
     @property
     def model(self) -> str:
-        return "FallbackAdapter"
+        """The model of the instance that most recently served a request (the primary before any traffic)."""  # noqa: E501
+        return self._active_instance.model
 
     @property
     def provider(self) -> str:
-        return "livekit"
+        """The provider of the instance that most recently served a request (the primary before any traffic)."""  # noqa: E501
+        return self._active_instance.provider
 
     def chat(
         self,
@@ -194,6 +199,7 @@ class FallbackLLMStream(LLMStream):
                     if should_set_current:
                         should_set_current = False
                         self._current_stream = stream
+                        self._fallback_adapter._active_instance = llm
                     yield chunk
 
         except asyncio.TimeoutError:
--- a/livekit-agents/livekit/agents/llm/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/fallback_adapter.py
@@ -10,7 +10,7 @@ from typing import Any, ClassVar, Literal
 from .._exceptions import APIConnectionError, APIError
 from ..log import logger
 from ..types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, APIConnectOptions, NotGivenOr
-from .chat_context import ChatContext
+from .chat_context import ChatContext, MetricsMetadata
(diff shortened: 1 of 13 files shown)


livekit-agents/livekit/agents/llm/realtime_fallback_adapter.py
```diff
@@ -120,13 +120,18 @@ def __init__(
...
     def model(self) -> str:
-        return "RealtimeModelFallbackAdapter"
+        """The model name of the instance currently serving sessions (the primary until a swap)."""
+        return self._active_instance.model
 
     @property
     def provider(self) -> str:
-        return "livekit"
+        """The provider of the instance currently serving sessions (the primary until a swap)."""
+        return self._active_instance.provider
```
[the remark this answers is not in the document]
Fyko: _This is an AI generated response by the `/pr` skill_

----

Good catch — a fresh session always comes up on `models[0]`, but the adapter kept the stale failover label. Fixed in 172e76396: `_FallbackRealtimeSession.__init__` now resets `_active_instance` to the primary, with a regression test (`test_new_session_resets_active_model_to_primary`) covering failover → new session → label back on primary.

livekit-agents/livekit/agents/llm/fallback_adapter.py
```diff
@@ -77,16 +77,21 @@ def __init__(
...
+        self._active_instance: LLM = self._llm_instances[0]
+
         for llm_instance in self._llm_instances:
             llm_instance.on("metrics_collected", self._on_metrics_collected)
 
     @property
     def model(self) -> str:
-        return "FallbackAdapter"
+        """The model of the instance that most recently served a request (the primary before any traffic)."""  # noqa: E501
+        return self._active_instance.model
```
longcw: The nested `llm_request` span already carries the real `gen_ai.request.model`, so `model` can stay `"FallbackAdapter"`.

The only gap is `lk.agents.turn.*`, whose labels come from `MetricsReport` at three sites ([agent_activity:3380](https://github.com/livekit/agents/blob/4a187d164dfdb9b552846d7529652a7100918704/livekit-agents/livekit/agents/voice/agent_activity.py#L3380), [:3385](https://github.com/livekit/agents/blob/4a187d164dfdb9b552846d7529652a7100918704/livekit-agents/livekit/agents/voice/agent_activity.py#L3385), [:4457](https://github.com/livekit/agents/blob/4a187d164dfdb9b552846d7529652a7100918704/livekit-agents/livekit/agents/voice/agent_activity.py#L4457)). Maybe add a single `metrics_metadata` property, read at those three sites, splits the histograms and keeps `model`, `ErrorEvent`, and the `user_turn` span stable.

```python
if TYPE_CHECKING:
    from ..llm.chat_context import MetricsMetadata

@property
def metrics_metadata(self) -> MetricsMetadata:  # base classes return their own model/provider
    return self._active_instance.metrics_metadata
```

Fyko: ah, didn't know of that. good idea. updated in [refactor: label turn metrics via metrics_metadata, keep adapter model…](https://github.com/livekit/agents/pull/6690/commits/e222fc12cb93a1e37e7f0c604b9f36058bb4d7d5)

longcw (approved):

PR #6694, opened 2026-08-04
fix(voice): keep the run open while a tool holds the floor

### LINKED ISSUE #6658: Test framework issues when updating agent from tool
### Bug Description

Hi, I'm trying to test a multi-agent application but I'm facing issues when trying to test it following this guide: 

https://docs.livekit.io/agents/start/testing/test-framework/#overview

The app starts at "Agent 1" which uses a tool that updates the session agent  inside it in the following way:

```python
#Agent 1

@function_tool
async def tool(ctx: RunContext):
     await ctx.update("")
     forward_agent = Agent2()
     async with ctx.foreground():
        session = ctx.session
        session.update_agent(forward_agent)
     
     return None

class Agent2(Agent):
   async def on_enter(self):
      await self.session.generate_reply("Tell a joke")
```

and the test script is something like this:

```python
#test_script.py 

result.expect.next_event().is_function_call(
    name = "tool",
    arguments = ...
)

result.expect.next_event().is_function_call_output(
    output = ""
)

await result.expect.next_event().is_message(role = "assistant").judge(
    llm,
    intent = ...
)
```

But when I run the test script, I'm getting the following error:

Image

Debugging the test, I noticed that the tool invocation only emits these two events:

FunctionCallEvent
FunctionCallOutputEvent

The MessageEvent generated by Agent2 never reaches the test framework, even though the application behaves correctly outside the testing environment.

In production, after session.update_agent(forward_agent) is called, Agent2.on_enter() is executed, generate_reply() is called, and the assistant responds as expected.

My questions are: 

- Is there a supported way to continue listening for events emitted by the new agent after update_agent()?0
- Or is there a different recommended way to test multi-agent handoffs where the second agent immediately generates a reply?

Any guidance would be greatly appreciated.

### Expected Behavior

-

### Reproduction Steps

```bash
1. 
2.
3.
...
```

### Operating System

Linux

### Models Used

Gemini-3.5-flash-lire

### Package Versions

```bash
livekit==1.1.13
livekit-agents==1.6.5
livekit-api<2
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

longcw:
Fixes livekit/agents#6658

## The bug

Under `session.run()`, a tool that calls `update_agent()` inside `ctx.foreground()` records no handoff and nothing after it.

`ctx.foreground()` waits for the activity to become idle. The activity becomes idle only after the speech handle of the turn completes, and that same completion ends the `RunResult`. The run finishes while the tool still waits for the floor. `update_agent()` then finds a run that is already done, so `_watch_handle` does nothing.

## The fix

* `_wait_for_idle_and_hold` registers a guard on the active run before it waits for the floor, and releases it on exit. A run cannot complete while a tool waits for the floor or holds it.
* `AgentTask` suspends that guard with the blocked tasks it already suspends. Without this step, `await AgentTask()` inside `ctx.foreground()` deadlocks. The run waits for the guard, the guard waits for the task, and the task waits for a user turn. That turn only arrives after `run()` returns.
* `_watch_handle` and the methods next to it accept `asyncio.Future`. `asyncio.Task` is a subclass of `Future`, so every current caller stays valid.

## Scope

A tool that returns early with `ctx.update()` and then works in the background never takes the floor. It gets no guard, so the run keeps its own timing.

A tool that returns early, waits, and then takes the floor much later is still not recorded. Nothing holds the floor in that window, and the later work belongs to a new turn.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent.py
+++ b/livekit-agents/livekit/agents/voice/agent.py
@@ -960,7 +960,7 @@ class AgentTask(Agent, Generic[TaskResult_T]):
 
         # TODO(theomonnom): could the RunResult watcher & the blocked_tasks share the same logic?
         self.__inactive_ev.clear()
-        suspended_handles: list[SpeechHandle | asyncio.Task[Any]] = []
+        suspended_handles: list[SpeechHandle | asyncio.Future[Any]] = []
         pending_on_enter_task: asyncio.Task[None] | None = None
         try:
             # use wait_on_enter=False to avoid deadlock: on_enter may spawn nested
@@ -991,13 +991,14 @@ class AgentTask(Agent, Generic[TaskResult_T]):
 
             # now unwatch the parent speech handle and blocked tasks that belong to the
             # old activity — they can't complete while this AgentTask is running, and
-            # keeping them watched would block RunResult from completing.
+            # keeping them watched would block RunResult from completing. A foreground
+            # hold waiting on this task is in the same position, so its guard suspends too.
             if run_state and not run_state.done():
                 if speech_handle and run_state._unwatch_handle(speech_handle):
                     suspended_handles.append(speech_handle)
-                for task in blocked_tasks:
-                    if run_state._unwatch_handle(task):
-                        suspended_handles.append(task)
+                for blocked in [*blocked_tasks, *session._foreground_guards]:
+                    if run_state._unwatch_handle(blocked):
+                        suspended_handles.append(blocked)
                 if suspended_handles:
                     run_state._mark_done_if_needed(None)
         except Exception:
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -679,6 +679,8 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         self._idle_released.set()
 
         self._global_run_state: RunResult | None = None
+        # guards keeping an active run open across `_wait_for_idle_and_hold` scopes
+        self._foreground_guards: set[asyncio.Future[None]] = set()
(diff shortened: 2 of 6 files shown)


biztex:
I'd misdiagnosed this one on the issue — I reproduced without `ctx.foreground()` and blamed the pre-1.6.7 `on_enter` gap. Reaching idle being what completes the run is the part I missed. I've corrected that on #6658 so the reporter isn't left with my workaround.

Independently verified the branch, since I had the context loaded:

- Full `pytest --unit`: 1708 passed. (One unrelated failure, `test_ipc.py::test_proc_pool`, is a local native-DLL memory error under full-suite load — passes in isolation.)
- A/B'd the guard by stubbing only its registration in `_wait_for_idle_and_hold` and leaving everything else identical: `2 failed, 4 passed`, failing on exactly the two handoff tests. So they're testing the guard rather than passing incidentally.
- `guard.set_result(None)` in the `finally` looked like it could hit `InvalidStateError` if anything cancelled a watched handle, but nothing in `RunResult` ever cancels what it watches — the only mutations are `add`/`discard` — so it holds.
- Registering before the `wait_for_idle()` await, with the wait moved inside the `try`, is what makes the `ActivityClosedError` path safe; `test_guard_released_when_the_floor_never_arrives` pins it.

One observation, not a request. In the `AgentTask` restore path the suspended handles are re-watched into `session._global_run_state` read *after* `self.__fut`, which may be a newer run than the one they were suspended from. That's pre-existing for speech handles and blocked tasks, and for a guard it seems right anyway — the tool still holds the floor, so the new run's turn genuinely isn't finished. Noting it only because guards now ride that path too, and the comment above it explains the unwatch but not the re-watch target.

chenghao-mou (approved):

PR #6697, opened 2026-08-04
fix(inference): keep a stream retryable until it emits generation

u9g:
## What

`retryable` was cleared by any chunk reaching the caller, including chunks that carry no output — a usage block, or provider metadata such as the inference gateway's deployment/tier stamp or a Gemini thought signature.

The gateway stamps `extra_content.livekit` onto every delta it forwards, including the leading contentless one, and `_parse_choice` returns a real `ChatChunk` for a delta whose only payload is `extra_content`. So **every streamed response through the gateway went unretryable from its first chunk.** Only failures landing before that first chunk could still recover.

The consequence, from a simulation run: a mid-stream stall past the inter-chunk timeout failed the turn outright, with `partial=''` — nothing generated, nothing a retry could have duplicated, and the caller had heard nothing. Two other turns in the same run stalled identically but *before* their first chunk, retried, and recovered silently (their TTFTs decompose exactly as `10.0s timeout + 0.1s first-retry interval + a fast second attempt`, against a p50 of 0.44s).

## Change

`retryable` is now cleared on text or a tool call — the output a retry would actually repeat. Two commits:

1. The `retryable` guard itself. Also drops the `retryable = False` on the usage chunk: token counts aren't output either. In practice that line rarely mattered (with `include_usage`, usage arrives last, after content has already cleared the flag), but it's the same defect.
2. A `httpx.TimeoutException` clause. Only the request call runs inside the openai client's error mapping, so a timeout waiting on the *stream body* arrived as a raw httpx exception, fell through to the catch-all, and was reported as `APIConnectionError` — which is what made the original failure read as a connection problem. `APITimeoutError` subclasses `APIConnectionError`, so no behavior changes; errors are just named correctly.

## Tests

`tests/test_inference_llm_retry.py` — hermetic, via a mock transport that yields SSE chunks then stalls. The metadata chunk in the fixtures is the literal `extra_content.livekit` payload observed in production.

- a stall after metadata alone exhausts its retries (3 attempts at `max_retry=2`)
- a stall after real text does not retry (1 attempt)
- a stalled stream body surfaces as `APITimeoutError`

Each confirmed failing without its corresponding fix. Full `--unit` suite: 1737 passed.

## Related

A gateway-side stopgap and a TypeScript port are going up alongside this; I'll cross-link them here once they exist.

### DIFF
--- a/livekit-agents/livekit/agents/inference/llm.py
+++ b/livekit-agents/livekit/agents/inference/llm.py
@@ -465,11 +465,16 @@ class LLMStream(llm.LLMStream):
                     for choice in chunk.choices:
                         chat_chunk = self._parse_choice(chunk.id, choice, thinking_filter)
                         if chat_chunk is not None:
-                            retryable = False
+                            # only generation makes a retry duplicate what the caller
+                            # already saw; provider metadata (a gateway deployment
+                            # stamp, a thought signature) is not output
+                            if chat_chunk.delta and (
+                                chat_chunk.delta.content or chat_chunk.delta.tool_calls
+                            ):
+                                retryable = False
                             self._event_ch.send_nowait(chat_chunk)
 
                     if chunk.usage is not None:
-                        retryable = False
                         tokens_details = chunk.usage.prompt_tokens_details
                         cached_tokens = tokens_details.cached_tokens if tokens_details else 0
                         usage_chunk = llm.ChatChunk(
--- a/livekit-agents/livekit/agents/inference/llm.py
+++ b/livekit-agents/livekit/agents/inference/llm.py
@@ -465,7 +465,7 @@ class LLMStream(llm.LLMStream):
                     for choice in chunk.choices:
                         chat_chunk = self._parse_choice(chunk.id, choice, thinking_filter)
                         if chat_chunk is not None:
-                            if chat_chunk.carries_generation():
+                            if chat_chunk.has_response():
                                 retryable = False
                             self._event_ch.send_nowait(chat_chunk)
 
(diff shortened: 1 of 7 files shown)


toubatbrian (approved):

livekit-agents/livekit/agents/inference/llm.py
```diff
@@ -465,11 +476,11 @@ async def _run(self) -> None:
                     for choice in chunk.choices:
                         chat_chunk = self._parse_choice(chunk.id, choice, thinking_filter)
                         if chat_chunk is not None:
-                            retryable = False
+                            if _carries_generation(chat_chunk):
```
longcw: nit: `_carries_generation` can be an inline condition

longcw (approved):

livekit-agents/livekit/agents/inference/llm.py
u9g: done

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -81,6 +81,15 @@ class ChatChunk(BaseModel):
     delta: ChoiceDelta | None = None
     usage: CompletionUsage | None = None
 
+    def carries_generation(self) -> bool:
```
theomonnom: maybe?
```suggestion
    def has_response(self) -> bool:
```

theomonnom (approved):

livekit-agents/livekit/agents/llm/llm.py
u9g: done

PR #6699, opened 2026-08-04
fix(deepgram): use punctuated_word for word-level transcripts

mdylan2:
## Summary

`SpeechData.text` and `SpeechData.words` are built from different Deepgram fields, so the two disagree on the same object.

`text` comes from the transcript, which honours the `punctuate` option. The word list was built from the raw per-word field, which is lowercase and unpunctuated regardless of any option:

```python
text=alt["transcript"],                    # honours `punctuate`
...
words=[
    TimedString(
        text=word.get("word", ""),         # never punctuated
```

Deepgram returns `punctuated_word` alongside `word`, in the very dict that comprehension iterates:

```
keys : ['confidence', 'end', 'punctuated_word', 'start', 'word']
word            : hello my name is alex
punctuated_word : Hello. My name is Alex.
```

Field selection across the four cases:

```python
_word_text({"word": "alex", "punctuated_word": "Alex."}, punctuate=True)   # -> "Alex."
_word_text({"word": "alex"},                             punctuate=True)   # -> "alex"
_word_text({"word": "alex", "punctuated_word": ""},      punctuate=True)   # -> "alex"
_word_text({"word": "alex", "punctuated_word": "Alex."}, punctuate=False)  # -> "alex"
```

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -882,9 +882,11 @@ def _word_text(word: dict, *, punctuate: bool) -> str:
     honours `punctuate`, so the word list follows the same option to stay consistent with
     the text it belongs to. Falls back to `word` if the key is missing.
     """
-    if punctuate:
-        return word.get("punctuated_word") or word.get("word", "")
-    return word.get("word", "")
+    raw: str = word.get("word", "")
+    if not punctuate:
+        return raw
+    punctuated: str = word.get("punctuated_word") or raw
+    return punctuated
 
 
 def live_transcription_to_speech_data(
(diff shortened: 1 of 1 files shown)


chenghao-mou (approved):
lgtm. thanks for the PR

livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
```diff
@@ -285,6 +285,7 @@ async def _recognize_impl(
                 return prerecorded_transcription_to_speech_event(
                     config.language,
                     await res.json(),
+                    punctuate=config.punctuate,
```
chenghao-mou: nit: smart_format also applies punctuate so we should use OR here:

```python
use_punctuated_word=config.punctuate or config.smart_format
```

and later

```
use_punctuated_word=self._opts.punctuate or self._opts.smart_format
```

mdylan2: Good catch, fixed

PR #6700, opened 2026-08-04
feat(baseten): add Qwen3-ASR STT and Qwen3-TTS support, and language_options for Whisper STT

A-K-Erol:
## Summary

Baseten hosts **Qwen3-ASR Streaming** and **Qwen3-TTS** alongside the Whisper and Orpheus models this plugin already supports. They speak different wire protocols, so the existing `STT`/`TTS` classes can't reach them — pointing either at a Qwen3 endpoint connects and then produces nothing.

This adds `Qwen3STT` and `Qwen3TTS` as **separate classes**. The existing Orpheus/Whisper paths are untouched, so this is non-breaking.

| | `STT` / `TTS` | `Qwen3STT` / `Qwen3TTS` |
|---|---|---|
| STT audio | raw binary PCM | base64 `input_audio_buffer.append` |
| STT results | `message_type` / `transcript` | `type: "transcription"` / `segments[].text` |
| TTS transport | `{prompt, voice, …}`, or WS + `__END__` sentinel | `session.config` → `input.text` → `input.done` |
| TTS voices | preset names (`tara`) | registered voice clones |

```python
session = AgentSession(
    stt=baseten.Qwen3STT(model_id="your-qwen3-asr-model-id"),
    tts=baseten.Qwen3TTS(model_id="your-qwen3-tts-model-id", voice="your-voice"),
)
```

Both accept `model_endpoint`, `model_id`, or `chain_id` with the same precedence as `STT` (extracted into `_endpoint.py`).

## Notes on the design

A few protocol details drove decisions that aren't obvious from the diff:

- **`input.done` is a flush, not a close.** The session config stays in effect, so `Qwen3TTS` keeps one warm socket across turns. Re-dialing per utterance would add a connect plus a config round trip to every agent response.
- **Parked sockets need an application-level keepalive.** The server has a 30s idle timeout that protocol pings don't reset, so an idle socket reads as OPEN long after the server has given up. An empty `input.done` answers `session.done` with zero sentences and proves the *session* is alive.
- **Interrupted sockets are discarded, not parked.** Closing the socket is what stops in-flight generation; a graceful `session.close` would keep the GPU busy producing audio nobody hears.
- **One emitter segment per `SynthesizeStream`.** `push_text()` after a flush is dropped by the framework and `_main_task` raises on a segment-count mismatch, so a mid-stream flush means "synthesize what's buffered", never "start a new segment".
- **Qwen3-ASR reports a language *name*** (`"English"`), so `Qwen3STT` maps the common ones to ISO codes rather than passing a name where a code is expected.

## Voices

Qwen3-TTS Base ships no built-in speakers — there's no `tara` equivalent. `voice` names a registered clone, and `register_voice`/`list_voices` are exported to manage them. Worth knowing: the server stores uploaded voices on the container's local disk, so a runtime-registered voice lives on one replica and is lost on restart. The README documents baking the reference into the deployment instead, or passing `ref_audio`/`ref_text` to clone inline per session.

## Also: `language_options` for the existing Whisper `STT`

Bundled here because it is the same plugin and came out of the same customer
conversation. Baseten's streaming transcription API has accepted a
`language_options` list since Whisper runtime v0.5.0, which scopes detection to
the languages an agent actually supports. The plugin only ever sent a single
`audio_language`, forcing a choice between a fixed tag that mistranscribes the
other language and `auto`, which detects across all 99 and is unreliable on the
one- to two-second utterances typical of telephony.

```python
stt = baseten.STT(model_id="...", language="auto", language_options=["en", "de"])
```

Only added to the handshake when non-empty — `StreamingWhisperInput` uses
`extra="forbid"`, so sending it unconditionally would break anyone on an older
runtime. Also wired through `update_options` on both `STT` and `SpeechStream`.

Verified against a live Whisper Large V3 Turbo streaming deployment, with a
negative control: `language_options: ["en", "de"]` is accepted and transcribes
normally, while a deliberately misspelled field name closes the socket with
1011 — so acceptance confirms the field name rather than showing it was
silently ignored.

## Testing

Developed against `livekit-agents` 1.6.8 with a mock server implementing the Qwen3 protocols, driven through the real framework machinery (`AudioEmitter`, `RecognizeStream`, the retry loop) and through a real `AgentSession` with a real-time audio sink. Covered:

- TTS: token-by-token `push_text`, socket reuse without config resend, keepalive on a parked socket, barge-in discarding the socket, transient error retried / persistent error propagating, word timestamps rebased across sentence boundaries
- STT: handshake shape, `start_of_speech` → interim → final → `end_of_speech`, one-shot `recognize()`, partials disabled via `interim_results=False`
- `AgentSession`: TTS audio out with reuse across turns and `interrupt()` mid-playout; STT mic audio in surfacing interim + final user turns over consecutive VAD-bounded turns

I'm happy to contribute those as pytest suites if you'd like them in-tree — I left them out to keep the diff focused and avoid adding scripts your CI would try to collect.

## Live validation

Since opening this, both adapters have been run end to end against real Baseten
deployments of the same model-registry trusses they target (Qwen3-ASR streaming
on RTX Pro 6000, Qwen3-TTS Base on RTX Pro 6000), using real speech with ground
truth rather than synthetic audio.

**STT** — a mu-bench en-US utterance, streamed at 100ms frames:

```
partial: I want to get a higher limit on my
FINAL:   Hi, I want to get a higher limit on my credit card. | lang: en
truth:   Hi. I want to get a higher limit on my credit card.
```

100% word overlap, both through `Qwen3STT.stream()` directly and through a real
`AgentSession` (`user_input_transcribed`, 6 interims + 1 final). Confirms the
handshake, the base64 append frames, `type: "transcription"` parsing, the
`language_code: "English"` -> `en` mapping, and clean termination on the
commit-triggered final.

**TTS** — `voice.list` returns `{"voices": [], ...}` on the Base checkpoint,
confirming it ships no built-in speakers; `voice.add` cloning from a 14s
reference works; synthesis returns real 24kHz PCM; the second turn reuses the
warm socket (TTFA 1081ms vs a cold first turn).

**Round trip** — feeding the live TTS output back into the live STT transcribes
at 100% word overlap, so the synthesized audio is genuinely intelligible speech
and not just well-formed bytes.

One operational note worth stating: on a cold replica the first synthesis
exceeded the 60s session timeout and was retried by the framework before
succeeding. That is cold-start behavior rather than an adapter issue, but
production voice agents should keep `min_replica >= 1`.

`ruff check` and `ruff format` are clean.

### DIFF
--- a/livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/_endpoint.py
+++ b/livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/_endpoint.py
@@ -17,8 +17,12 @@
 from __future__ import annotations
 
 import os
+from urllib.parse import urlparse
+
+from .log import logger
 
 _TRUSS_URL_TEMPLATE = "wss://model-{model_id}.api.baseten.co/environments/production/websocket"
+_LOOPBACK = {"localhost", "127.0.0.1", "::1"}
 _CHAIN_URL_TEMPLATE = "wss://chain-{chain_id}.api.baseten.co/environments/production/websocket"
 
 
@@ -41,5 +45,14 @@ def resolve_endpoint(
         raise ValueError(
             f"This model is served over WebSocket only; got {endpoint!r}. Endpoints "
             "look like wss://model-<id>.api.baseten.co/environments/production/websocket"
+        )
+    if endpoint.startswith("ws://") and urlparse(endpoint).hostname not in _LOOPBACK:
+        # ws:// stays allowed for local proxies and tests, but the API key rides in
+        # an Authorization header and the audio is unencrypted, so plaintext to
+        # anything off-host is worth shouting about.
+        logger.warning(
+            "endpoint %r is plaintext ws://: the Baseten API key and all audio will be "
+            "sent unencrypted. Use wss:// for any non-local host.",
+            endpoint,
         )
     return endpoint
--- a/livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/_endpoint.py
+++ b/livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/_endpoint.py
@@ -1,0 +1,45 @@
+# Copyright 2025 LiveKit, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Shared endpoint resolution for the Qwen3 WebSocket models."""
+
+from __future__ import annotations
+
+import os
+
+_TRUSS_URL_TEMPLATE = "wss://model-{model_id}.api.baseten.co/environments/production/websocket"
+_CHAIN_URL_TEMPLATE = "wss://chain-{chain_id}.api.baseten.co/environments/production/websocket"
+
+
+def resolve_endpoint(
+    model_endpoint: str | None, model_id: str | None, chain_id: str | None, env_var: str
+) -> str:
+    """Resolve a WebSocket endpoint, mirroring `STT`'s precedence rules."""
+    endpoint = model_endpoint
+    if not endpoint and model_id:
+        endpoint = _TRUSS_URL_TEMPLATE.format(model_id=model_id)
+    if not endpoint and chain_id:
+        endpoint = _CHAIN_URL_TEMPLATE.format(chain_id=chain_id)
+    endpoint = endpoint or os.environ.get(env_var)
+    if not endpoint:
+        raise ValueError(
+            f"An endpoint is required: pass model_endpoint, model_id, or chain_id, "
+            f"or set {env_var}."
+        )
+    if not endpoint.startswith(("wss://", "ws://")):
+        raise ValueError(
+            f"This model is served over WebSocket only; got {endpoint!r}. Endpoints "
+            "look like wss://model-<id>.api.baseten.co/environments/production/websocket"
+        )
+    return endpoint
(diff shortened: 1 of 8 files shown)


livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/_endpoint.py
```diff
@@ -0,0 +1,58 @@
...
+
+from __future__ import annotations
+
+import os
+from urllib.parse import urlparse
+
+from .log import logger
+
+_TRUSS_URL_TEMPLATE = "wss://model-{model_id}.api.baseten.co/environments/production/websocket"
+_LOOPBACK = {"localhost", "127.0.0.1", "::1"}
```
iancarrasco-b10: ooc what is reason for this

iancarrasco-b10 (approved):

livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/_endpoint.py
A-K-Erol: It's the exemption list for the plaintext-`ws://` warning further down in `resolve_endpoint`:

```python
if endpoint.startswith("ws://") and urlparse(endpoint).hostname not in _LOOPBACK:
    logger.warning("endpoint %r is plaintext ws://: the Baseten API key and all audio "
                   "will be sent unencrypted. Use wss:// for any non-local host.", endpoint)
```

We send `Authorization: Api-Key ...` on that connection, so a `ws://` endpoint to a remote host puts the key and the audio in cleartext. Loopback is exempt because local proxies and the tests legitimately use `ws://127.0.0.1` and nothing leaves the machine.

Warning rather than rejecting was deliberate — hard-failing would break those local setups, and the existing `TTS`/`STT` in this plugin accept `ws://` with no check at all today.

Fair callout though: it was wedged between the two URL templates with no context, which is exactly why it read as arbitrary. Moved it next to the logic it serves and added a comment in 843eac0.

tinalenguyen (commented):
hi, thanks for the PR! i wonder if we could fold in `Qwen3TTS` and `Qwen3STT` into the existing `baseten.STT` and `baseten.TTS` classes based on the models specified. cartesia does something similar as well for their STT and it feels clean, what do you think? [example snippet](https://github.com/livekit/agents/blob/62d527d4603c9b0eaafdce272a80d42de4efae9e/livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/stt.py#L291-L322)

tinalenguyen (approved):

PR #6705, opened 2026-08-05
feat(openai): support context hints in STT

longcw:
`gpt-transcribe` and `gpt-live-transcribe` accept context hints: `keywords` for literal terms expected in the audio, and a plural `languages` list for code-switched speech.

- `language` now takes a `str` or a `list[str]`; more than one language raises on models that accept only one, and codes are normalized to ISO-639-1 because the API rejects regional tags such as `en-US`.
- `keywords` gives the plugin somewhere to put terms, so `stt_context_options` and automatic keyterm detection now reach OpenAI STT. Detected terms merge behind the user's own and apply with a `session.update` on the open connection; only a change of `model` reconnects.
- `gpt-live-transcribe` rejects any `turn_detection` config and emits no `speech_started`/`speech_stopped`, so it joins `gpt-realtime-whisper` on the client-commit path.
- Requires `openai>=2.50`, where `keywords`/`languages` became typed on both transcription APIs.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
@@ -108,6 +108,13 @@ def _as_languages(language: str | list[str]) -> list[str]:
     return [code for code in language if code]
 
 
+def _as_turn_detection(
+    turn_detection: SessionTurnDetection,
+) -> RealtimeTranscriptionSessionAudioInputTurnDetection:
+    """Validate a turn-detection dict, filling in the `type` the union is tagged on."""
+    return _TURN_DETECTION.validate_python({"type": "server_vad", **turn_detection})
+
+
 def _validate_context(model: str, languages: list[str], keywords: list[str]) -> None:
     """Raise when keywords or several languages go to a model that takes neither."""
     if _supports_context_hints(model):
@@ -129,7 +136,7 @@ class _STTOptions:
     model: STTModels | str
     languages: list[str]
     detect_language: bool
-    turn_detection: SessionTurnDetection
+    turn_detection: RealtimeTranscriptionSessionAudioInputTurnDetection
     keywords: list[str] = field(default_factory=list)
     prompt: NotGivenOr[str] = NOT_GIVEN
     noise_reduction_type: NotGivenOr[str] = NOT_GIVEN
@@ -163,7 +170,7 @@ def _session_update(opts: _STTOptions) -> SessionUpdateEvent:
     )
     # leave the field unset for models that reject it; the rest get server-side VAD
     if not _is_realtime_only(opts.model):
-        audio_input.turn_detection = _TURN_DETECTION.validate_python(opts.turn_detection)
+        audio_input.turn_detection = opts.turn_detection
 
     if opts.noise_reduction_type:
         audio_input.noise_reduction = NoiseReduction(type=opts.noise_reduction_type)
@@ -275,7 +282,7 @@ class STT(stt.STT):
             model=model,
             prompt=prompt,
             keywords=resolved_keywords,
-            turn_detection=turn_detection,
+            turn_detection=_as_turn_detection(turn_detection),
             temperature=temperature,
         )
         if is_given(noise_reduction_type):
@@ -434,9 +441,6 @@ class STT(stt.STT):
         if is_given(language):
             opts.languages = _as_languages(language)
             _validate_context(opts.model, opts.languages, opts.keywords)
-            if opts.languages != self._opts.languages:
-                # the pool would otherwise hand this stream a socket carrying another language
-                self._pool.invalidate()
         stream = SpeechStream(
             stt=self,
             pool=self._pool,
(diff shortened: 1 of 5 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -62,33 +79,112 @@
...
+    return [code for code in language if code]
+
+
+def _validate_context(model: str, languages: list[str], keywords: list[str]) -> None:
+    """Raise when keywords or several languages go to a model that takes neither."""
+    if _supports_context_hints(model):
+        return
+    if keywords:
+        raise ValueError(
+            f"keywords are only supported by gpt-transcribe and gpt-live-transcribe, not {model}"
```
chenghao-mou: nit: should we put the model names into constants?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -143,19 +243,23 @@ def __init__(
...
                 except ImportError as e:
                     raise ImportError(
-                        "livekit-plugins-silero is required for the gpt-realtime-whisper model "
+                        f"livekit-plugins-silero is required for {model} "
                         "(no server-side endpointing). Pass `vad=None` to opt out and drive "
                         "`input_audio_buffer.commit` manually."
                     ) from e
                 vad = SileroVAD.load()
 
         super().__init__(
```
chenghao-mou, line 257: should we change use_realtime to NOT_GIVEN so we can set it to True when using the live model?

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -143,19 +243,23 @@ def __init__(
                     from livekit.plugins.silero import VAD as SileroVAD
                 except ImportError as e:
                     raise ImportError(
-                        "livekit-plugins-silero is required for the gpt-realtime-whisper model "
+                        f"livekit-plugins-silero is required for {model} "
```
chenghao-mou: we should probably use the inference.VAD instead

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -332,33 +447,64 @@ def update_options(
...
             turn_detection: When using realtime, this controls how model detects the user is done speaking.
             noise_reduction_type: Type of noise reduction to apply. "near_field" or "far_field"
             temperature: Sampling temperature between 0 and 1. Not supported for realtime transcription.
         """  # noqa: E501
-        if is_given(model):
-            self._opts.model = model
-        if is_given(language):
-            self._opts.language = LanguageCode(language)
+        # resolve first: an unsupported combination must raise before anything is applied
+        resolved_model = model if is_given(model) else self._opts.model
```
chenghao-mou, line 483: Q: if the user changes from the live model to a different one and also has a VAD running before the change, we would have two client-commit owners, right?

chenghao-mou (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
longcw: now skip the vad commit if the new model doesn't need it.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -449,37 +573,42 @@ async def _recognize_impl(
...
+                prompt=transcription.prompt or openai.omit,
                 response_format=format,
                 temperature=self._opts.temperature
                 if is_given(self._opts.temperature)
                 else openai.omit,
                 timeout=httpx.Timeout(30, connect=conn_options.timeout),
             )
 
-            sd = stt.SpeechData(text=resp.text, language=self._opts.language)
+            sd = stt.SpeechData(text=resp.text, language=_transcript_language(self._opts.languages))
```
chenghao-mou, line 644: nit: we should use the detected language/languages

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/stt.py
```diff
@@ -62,100 +80,185 @@
...
-                settings. Pass `vad=None` to opt out of the auto-load and drive
+                does not support server-side turn detection (`gpt-realtime-whisper`,
+                `gpt-live-transcribe`).
+                When not provided and the model requires it, the bundled Silero VAD is used with
+                default settings. Pass `vad=None` to opt out and drive
                 `input_audio_buffer.commit` yourself.
         """  # noqa: E501
 
+        if not is_given(use_realtime):
+            use_realtime = _requires_client_commit(model)
```
chenghao-mou: nit: we should warn if it is False while the model is streaming.

davidzhao (approved):

theomonnom (approved):

PR #6714, opened 2026-08-05
fix(voice): defer false-interruption resume while a transcript is pending

ajayarora1235:
## Problem

With `resume_false_interruption` enabled, stale held agent audio resumes into the caller's pause. Observed in production (STT turn detection, Cartesia Ink-2): an agent reply held pre-playout behind a long caller narration resumed at each of the caller's breaths, emitted one word ("I'm" … then, on the next held reply, "I"), and was killed by the turn commit moments later — heard by the caller as the agent repeatedly butting in at every pause they took. (Details anonymized; the repro below is fully synthetic.)

## Mechanism

The EOS-to-final gap. With STT-based turn detection, a speech window's `END_OF_SPEECH` can arrive before its `FINAL_TRANSCRIPT` (a long utterance produces several SOS/EOS windows; the committing final lags). In that gap:

- `on_end_of_speech` arms the false-interruption timer with the pause's **own** timeout — which is **0** for a pre-playout pause taken by `on_start_of_speech` ("resume immediately when user stops speaking") — so the resume is instantaneous at the caller's first breath;
- `_on_timeout`'s deferral (from #6662) only hooks an **open** `_end_of_turn_task`, and no bounce is open until the final arrives — so nothing defers the resume even though word-bearing interims prove live speech and a turn decision is imminent.

## Repro

Two deterministic tests (`tests/test_stale_resume_stt_windows.py`), built on the `test_false_interruption_resume.py` idiom — **red on `main` at the first commit** (check out the test-only commit to see them fail):

1. pre-playout pause (timeout=0) + bare EOS → resume fires instantly into the breath;
2. mid-playout pause (configured timeout) + inter-window EOS → resume fires in the gap between two speech windows of one utterance. (Distinct from `test_resume_is_immediate_when_no_turn_decision_is_open`: there the EOS is a lone noise glitch; here a real speech window with pending interims precedes it.)

## Fix

Defer the resume while word-bearing interims are in flight, mirroring the existing eot-task deferral: the coming final will start (or refresh) the turn decision that owns the pause (and the final path already interrupts it via `_cancel_speech_pause`). The deferral re-checks at 100ms and is **bounded** (`_PENDING_TRANSCRIPT_MAX_DEFERRAL`, 2s): if the transcript never materializes (e.g. the STT stream died), behavior degrades to the plain timeout — the pre-fix fallback. A third test pins the bound.

**Design question for maintainers:** would you prefer hooking the transcription-timeout event instead of the bounded re-check? The re-check is self-contained in `_start_false_interruption_timer`; the event hook would avoid polling but couples the timer to the recognition lifecycle. Happy to rework.

All 8 existing tests in `tests/test_false_interruption_resume.py` still pass (plus `test_agent_session.py`, `test_agent_task_close_race.py` — 91 total across the four files). Extends #6662.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -98,6 +98,11 @@ from .turn import (
     _StreamingTurnDetectorStream,
 )
 
+# false-interruption resume: re-check cadence and cap while a speech window's
+# transcript is still in flight (see _start_false_interruption_timer)
+_PENDING_TRANSCRIPT_RECHECK = 0.1
+_PENDING_TRANSCRIPT_MAX_DEFERRAL = 2.0
+
 if TYPE_CHECKING:
     from ..llm import mcp
     from .agent_session import AgentSession, ExpressiveOptions
@@ -4287,6 +4292,10 @@ class AgentActivity(RecognitionHooks):
     def _start_false_interruption_timer(self, timeout: float) -> None:
         self._cancel_false_interruption_timer()
 
+        # set when resume is first deferred on a pending transcript; a dead STT
+        # stream with interims never finalizing degrades to plain timeout behavior
+        transcript_wait_deadline: float | None = None
+
         def _on_false_interruption() -> None:
             if self._paused_speech is None or (
                 self._current_speech and self._current_speech is not self._paused_speech.handle
@@ -4341,6 +4350,7 @@ class AgentActivity(RecognitionHooks):
             _on_false_interruption()
 
         def _on_timeout() -> None:
+            nonlocal transcript_wait_deadline
             self._false_interruption_timer = None
 
             # an open turn decision owns the paused speech: it either commits and interrupts it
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -4353,6 +4363,22 @@ def _on_timeout() -> None:
...
+            if self._audio_recognition and getattr(
+                self._audio_recognition, "_audio_interim_transcript", ""
+            ):
+                if transcript_wait_deadline is None:
+                    transcript_wait_deadline = time.monotonic() + _PENDING_TRANSCRIPT_MAX_DEFERRAL
+                if time.monotonic() < transcript_wait_deadline:
+                    self._false_interruption_timer = self._session._loop.call_later(
+                        _PENDING_TRANSCRIPT_RECHECK, _on_timeout
+                    )
+                    return
```
chenghao-mou, line 4378: If the deadline expires while the interim transcript is still non-empty, we fall through and resume the paused speech. 
This creates vendor-dependent behavior: ghost interims unnecessarily extend the pause, while finals delayed beyond the window can still reproduce the stale-resume issue. 

Should the maximum deferral be configurable, preferably as a duration rather than a boolean, so integrations can control how long interim transcripts are trusted?

Issue #6415, opened 2026-07-14
turn-detector-v1: high eot variance on short utterances — guidance on tuning for short affirmations?


Propconnect-ai:
### Context

I run a production voice agent for lead qualification (currently on another platform) and was evaluating LiveKit. While reading the debug logs from two `console` sessions I noticed something I'd like to understand better before building on it.

This is a question, not a bug report — I may well be misreading the design intent.

### What I observed

`turn-detector-v1`, English, `unlikely_threshold: 0.56`, no overrides.

**The same word, spoken twice, in two sessions — same microphone, same room:**

| Utterance | eot probability | threshold | endpointing_delay |
|---|---|---|---|
| `"Hello?"` (session 1) | **0.477** | 0.56 | **2.5 s** |
| `"Hello?"` (session 2) | **0.650** | 0.56 | **0.3 s** |

Identical input, opposite sides of the threshold, ~8× difference in wait before the agent responded.

**Two semantically identical affirmations, same session:**

| Utterance | eot probability | endpointing_delay |
|---|---|---|
| `"Yeah."` | **0.127** | **2.5 s** |
| `"Yes."` | **0.649** | **0.3 s** |

`"Yeah."` scored 0.127 — the model was ~87% confident the turn had *not* ended, when it plainly had. It waited 2.5 seconds.

**And one that cleared by a rounding error:**

| Utterance | eot probability | threshold | margin |
|---|---|---|---|
| `"I'm not sure."` | **0.5613** | **0.560** | **+0.0013** |

### Why I'm asking

A qualification call is largely short affirmations — *yeah, okay, sure, right, correct, no*. If the score distribution for those straddles the threshold, a meaningful fraction of turns will fall below it and trigger the full 2.5 s endpointing delay.

On a phone call with a stranger, 2.5 s of dead air after they say "yeah" doesn't read as thoughtfulness — it reads as the line dropping. Callers start talking over the agent or hang up.

I'm not claiming the model is wrong. Short utterances genuinely carry less prosodic and lexical signal, and I don't know what the training distribution looked like. But the variance appears highest on the shortest and most frequent utterances in a real call, which is where it matters most.

### Questions

1. Is this variance expected on very short utterances, or does it suggest something wrong with my setup?
2. The config exposes `"overrides": null` alongside the per-language thresholds. Is per-language (or per-use-case) override the intended lever here? I couldn't find guidance on when or why to reach for it.
[...]

PR #6715, opened 2026-08-05
test(voice): repair and re-enable turn-detection tests in CI

brandonbodily:
## Summary

Four test modules are marked `pytest.mark.audio_eot`, covering **83 tests** of turn-detection behaviour:

| Module | Tests |
|---|---|
| `tests/test_audio_recognition_turn_detection.py` | 29 |
| `tests/test_audio_turn_detector_fallback.py` | 40 |
| `tests/test_turn_detection_fsm.py` | 11 |
| `tests/test_turn_detection_cloud_stream.py` | 3 |

CI runs `make unit-tests` → `pytest --unit`, which **deselects all of them**, and no workflow passes `--audio_eot`:

```
$ pytest tests/test_audio_recognition_turn_detection.py --unit
29 deselected in 0.06s

$ grep -rn "audio_eot" .github/
(no matches)
```

None of these 83 tests has been running in CI.

## Consequence

`test_audio_recognition_turn_detection.py` drifted out of sync with the source. On a clean checkout of `main`, **8 of its 29 fail**:

| # | Test expects | Source has |
|---|---|---|
| 1 | `AudioRecognition.update_turn_detector` | `_update_turn_detector` (renamed private) |
| 2 | `AudioRecognition.clear_user_turn` | `_clear_user_turn` (renamed private) |
| 3 | `_make_full_recognition_for_eou` stub | missing `_turn_backchannel_over_agent`, `_stt_consumer_atask` |
| 4 | log assertion on `"already flushed"` | warning now reads `"transcript arrives after turn has been committed"` |

Each is a one-line repair. The other three modules were already green — they simply never ran.

## Changes

- Repair the four drift points.
- Add `--audio_eot` to the `unit-tests` target so these run alongside `--unit`.

No library code is touched — only the test module and the make target.

## Verification

```
$ pytest --audio_eot          # before
8 failed, 75 passed

$ pytest --audio_eot          # after
83 passed
```

`pytest --unit` and `pytest --unit --audio_eot` produce **identical** results across every other test in the suite.

## Why this matters beyond the repair

This is the coverage for turn-detection commit policy, which is where several open correctness issues live (#6504, #3515). With the suite green, adding a single assertion to the existing `test_late_stt_final_after_flush_short_circuits` reproduces #6504 deterministically:

```python
ar._hooks.on_end_of_turn.assert_not_called()
# AssertionError: Expected 'on_end_of_turn' to not have been called. Called 2 times.
```

That test already drives the exact #6504 scenario — a late STT final after `_turn_detector_flushed` is set — and asserts on `predict`, on `on_eot_prediction`, and on the warning, but never on whether a turn is **committed**. The sibling test two functions below (`test_predict_timeout_signals_fallback_and_drops_future`) does assert `on_end_of_turn.assert_called_once()`, so the pattern is established in this file.

I've deliberately left that assertion **out** of this PR — it would fail until #6504 is fixed, and PR #6516 is already open against it. This PR is only the CI repair, so it lands independently and green.

## Two other markers look dark, but I didn't touch them

While tracing this I noticed `plugin` (26 modules) and `docs` (1) are also never selected by any workflow — they're only reachable through `tests/Makefile:unit-tests` (`--unit --plugin --evals --docs`), and CI's `make unit-tests` step runs from the repo root, so it hits the root `makefile` target instead.

To be clear about what still *does* work: the plugin matrix isn't idle — `tests/Makefile:test` runs `pytest --tts`, and `test_tts.py` reads the `PLUGIN` env var (lines 286, 551) to filter its parametrized provider list, so each matrix job genuinely exercises that provider's TTS. What it doesn't run are the 26 dedicated `test_plugin_*` modules, including the `*_stt.py` ones.

I've left those out of this PR on purpose — re-enabling 26 provider-integration modules is a credentials-and-CI-cost decision for maintainers, not something to slip into a test repair. Happy to open a separate issue if that's useful.

### DIFF
--- a/makefile
+++ b/makefile
@@ -82,9 +82,9 @@ check: format-check lint type-check ## Run all checks (format, lint, type-check)
 
 fix: format lint-fix ## Run format and lint checks and fix issues automatically (format, lint)
 
-unit-tests: ## Run unit tests (modules marked with `pytestmark = pytest.mark.unit`)
+unit-tests: ## Run unit tests (modules marked `pytest.mark.unit` or `pytest.mark.audio_eot`)
 	@echo "$(BOLD)$(CYAN)Running unit tests...$(RESET)"
-	PYTHONPATH="$$PWD" uv run pytest --unit $(PYTEST_ARGS)
+	PYTHONPATH="$$PWD" uv run pytest --unit --audio_eot $(PYTEST_ARGS)
 
 # ============================================
 # Development Workflows
--- a/tests/test_audio_recognition_turn_detection.py
+++ b/tests/test_audio_recognition_turn_detection.py
@@ -85,6 +85,8 @@ def _make_full_recognition_for_eou() -> AudioRecognition:
     ar._turn_detector_stream = stream_mock
     ar._turn_detector_prediction_fut = None
     ar._turn_detector_flushed = False
+    ar._turn_backchannel_over_agent = False
+    ar._stt_consumer_atask = None
     ar._turn_detector_late_prediction_warned = False
     ar._agent_speaking = False
     ar._interruption_enabled = False
@@ -349,7 +351,7 @@ class TestEotPredictionDedup:
         ar._turn_detector_stream.flush = MagicMock()
         ar.update_stt = MagicMock()  # type: ignore[method-assign]
 
-        ar.clear_user_turn()
+        ar._clear_user_turn()
 
         assert ar._last_emitted_prediction is None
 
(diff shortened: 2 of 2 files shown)


Propconnect-ai:
Nice — re-enabling these is overdue, and test_late_stt_final_after_flush_short_circuits is exactly the path #6504 and #6415 exercise. I reproduced both on LiveKit Cloud (1.6.5) with audio + support access, linked on those issues — the email-input case ("It's john at" commits via VAD, then "example dot com" arrives after the flush) triggers it deterministically.

Once this lands and #6516 fixes the commit path, happy to open the follow-up that adds the on_end_of_turn.assert_not_called() assertion to that test so #6504 stays covered as a regression — you've basically written the blueprint for it above. Let me know if that split works for you.

PR #6720, opened 2026-08-06
fix(aws): gate audio frames during text block sends to prevent Nova Sonic ValidationException

lanazhang:
Summary<br>Fixes a mid-session race condition in the Nova Sonic realtime plugin where generate_reply() and \_process_audio_input run on independent async tasks with no coordination. When an audioInput frame arrives at the server between a TEXT contentStart and TEXT contentEnd, Nova Sonic throws:

ValidationException: Chat history should be sent completely before streaming audio.<br>This affects \~5% of generate_reply() calls under real traffic.

Problem<br>The plugin sends audio and text on separate async tasks:

\_process_audio_input (Task A):  audioInput → audioInput → audioInput → ...<br>\_send_text_message   (Task B):  contentStart → textInput → contentEnd<br>Without coordination, the server can receive events in this order:

TEXT  contentStart (interactive=True)<br>TEXT  textInput<br>AUDIO audioInput   ← text block still open → ValidationException<br>TEXT  contentEnd   ← too late<br>The server-side validation does not distinguish interactive=True (cross-modal input allowed during audio) from interactive=False (chat history that must complete before audio). Both trigger the same rejection.

Fix<br>Added an asyncio.Event (\_text_block_gate) that pauses audio frame transmission while a text block is in-flight:

**init**: Initialize \_text_block_gate (starts set/open — audio flows freely)<br>\_process_audio_input: await self.\_text_block_gate.wait() before each audioInput send<br>\_send_text_message: gate.clear() before contentStart, gate.set() after contentEnd (in try/finally)<br>This ensures TEXT contentEnd always arrives before the next audioInput frame.

Latency Impact<br>\~22ms per text block (the duration of the 3 sends + two 10ms sleeps already present). At most 1-2 audio frames queue in the channel and send immediately when the gate re-opens. Well below the human perception threshold. Same pattern as the existing \_stream_ready gate from PR [#5626](<https://github.com/livekit/agents/issues/5626>).

Relation to PR [#5626](<https://github.com/livekit/agents/issues/5626>)<br>PR [#5626](<https://github.com/livekit/agents/issues/5626>) fixed a session-initialization race where audio started before the stream was ready. This PR fixes a mid-session race where the stream is established and audio is already flowing but text interleaves with it. Different trigger point, different code path, same class of bug.

Testing<br>Reproduced deterministically with a script that sends events in the failing order. Verified the fix eliminates the error across 15 concurrent text+audio injection attempts (750+ audio frames, 0 errors).


livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
```diff
@@ -2189,14 +2203,23 @@ async def _send_text_message(self, text: str, interactive: bool = True) -> None:
...
-        await self._send_raw_event(self._event_builder.create_content_end_event(content_name))
+        # Gate audio frames while the text block is in-flight.
+        # This ensures TEXT contentEnd arrives before the next audioInput frame,
+        # preventing the "Chat history should be sent completely before streaming audio"
+        # ValidationException from Nova Sonic.
+        self._text_block_gate.clear()  # Pause audio frame transmission
+        try:
+            # Send event sequence: contentStart → textInput → contentEnd
+            await self._send_raw_event(event)
+            await asyncio.sleep(0.01)
```
theomonnom: Are the sleep really needed?

lanazhang: Good catch. The sleeps were pre-existing (not added by this PR), but you're right they're unnecessary. The events are sequential on the same stream so ordering is guaranteed without them. Removing them also reduces the gate hold time from ~22ms to <3ms, which is a nice bonus.

I'll remove them in the next push.

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
```diff
@@ -2189,14 +2203,37 @@ async def _send_text_message(self, text: str, interactive: bool = True) -> None:
...
+            content_started = False
+            try:
+                await self._send_raw_event(event)
+                content_started = True
+                await self._send_raw_event(
+                    self._event_builder.create_text_content_event(content_name, text)
+                )
+                await self._send_raw_event(
+                    self._event_builder.create_content_end_event(content_name)
+                )
```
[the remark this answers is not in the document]
lanazhang: resolved

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
lanazhang: resolved

theomonnom (approved):

PR #6731, opened 2026-08-06
fix(room_io): key user transcript segments on the transcription item_id

### LINKED ISSUE #6710: room_io: user transcript segments ignore item_id, so an in-place item revision renders as N separate segments
## Environment

- `livekit-agents` **1.6.5**
- `livekit-plugins-openai` **1.6.5**, `livekit-plugins-xai` **1.6.5**
- Realtime model: xAI `grok-voice-think-fast-2.0`
- Observed: **2026-08-05**, production traffic

## Summary

When a realtime provider **finalizes the same input-audio transcription item more than once** — emitting repeated `input_audio_transcription_completed` events for a single `item_id`, each carrying a longer transcript — `RoomIO` renders each one as an **independent transcription segment**. The result is N stacked transcript segments where each contains all of its predecessors, rather than one segment updated in place.

`ev.item_id` is present on the event and would disambiguate this, but the forwarding path does not use it.

## Where

`livekit/agents/voice/room_io/room_io.py`, `RoomIO._forward_user_transcript`:

```python
async for ev in event_ch:
    if self._user_tr_output is None:
        continue
    await self._user_tr_output.capture_text(ev.transcript)
    if ev.is_final:
        self._user_tr_output.flush()
```

`ev.item_id` is never read. Combined with `livekit/agents/voice/room_io/_output.py`, where `flush()` calls `_reset_state()` and `_reset_state()` assigns a **new** `self._current_id = utils.shortuuid("SG_")`, every `is_final=True` starts a fresh segment id — including a second final for an item that already had one.

Downstream, `AgentActivity._on_input_audio_transcription_completed` does use the id correctly:

```python
msg = llm.ChatMessage(role="user", content=[ev.transcript], id=ev.item_id)
self._agent._chat_ctx._upsert_item(msg)
```

so the **chat context** collapses the revisions into one message (correct), while the **rendered transcript** does not. The two views of the same conversation diverge.

## Measured evidence

One production session, agent-side instrumentation logging `item_id`, `is_final` and transcript length on every `user_input_transcribed` event. Two items were re-finalized:

| item_id (abridged) | finals | transcript length at each final | span |
|---|---:|---|---|
| `7051abbb…` | **4** | 30 → 78 → 132 → 177 | 28.8 s |
| `b366576e…` | **3** | 15 → 46 → 66 | 17.2 s |

Every other item in that session produced exactly one final.

A second session the same day, after we added a client-side mitigation, shows the same shape — one item finalized at 17, then 40, then 86 characters.

Client-side, each of those finals arrived as a **different** `TranscriptionSegment.id`, producing a new transcript bubble each time, each containing the full accumulated text.

## Suggested fix

Key the user transcript segment on `item_id`, so a repeated final for a known item updates that segment rather than opening a new one. Roughly:

```python
async for ev in event_ch:
    if self._user_tr_output is None:
        continue
    if ev.item_id is not None and ev.item_id != self._current_user_item_id:
        self._current_user_item_id = ev.item_id
        # start a new segment for a genuinely new item
    await self._user_tr_output.capture_text(ev.transcript)
    if ev.is_final:
        self._user_tr_output.flush()
```

with `_ParticipantTranscriptionOutput` retaining its `_current_id` when the item is unchanged. `item_id` is documented as `"Provider-specific ID for the transcribed input item, when available"` and is `str | None`, so the `None` case must keep today's behaviour.

## Why we think this is worth fixing here rather than only upstream of it

The provider behaviour that triggers it is being reported separately to xAI. But `item_id` exists on the event precisely to identify the item, `AgentActivity` already relies on it for exactly this purpose, and the divergence between the chat context and the rendered transcript is invisible to an application author — the transcript looks like the user genuinely repeated themselves. A one-line keying change makes `RoomIO` robust to a class of provider behaviour it currently has no defence against.

## Adjacent

**#6109** — same subsystem, transcription segment lifecycle. Raising this separately because the mechanism here is specifically the unused `item_id`, but they may share a fix.

## What we are NOT reporting here

Two other behaviours we observed are provider-side and are going to xAI, not here:

1. repeated `status="completed"` transcription events for one item, including a case where an utterance ~18 seconds later was folded into the previous item;
2. a duplicate assistant response generated for a single, cleanly-finalized user item.

Neither is a `livekit-agents` defect as far as we can tell.

RGB-loop:
A realtime provider may finalize the same input-audio transcription item more than once, each time with a longer revision of the transcript. RoomIO ignored ev.item_id when forwarding user transcripts, so every final opened a fresh segment id and clients rendered N stacked segments, each containing all of its predecessors — while the chat context correctly collapsed the revisions into one message via _upsert_item.

Derive the segment id from the item id when present, so a repeated final for the same item reuses the segment id and clients revise the segment in place. Events without an item_id (the STT path) keep the previous behavior. An open segment with no visible text yet (e.g. the empty interim emitted on input_speech_stopped) is re-keyed in place instead of flushed, so no empty orphan segment is published.

Fixes #6710

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/_output.py
+++ b/livekit-agents/livekit/agents/voice/room_io/_output.py
@@ -219,6 +219,7 @@ class _ParticipantLegacyTranscriptionOutput:
         self._room.on("local_track_published", self._on_local_track_published)
         self._flush_task: asyncio.Task[None] | None = None
         self._closed = False
+        self._next_segment_id: str | None = None
 
         self._reset_state()
         self.set_participant(participant)
@@ -257,6 +258,21 @@ class _ParticipantLegacyTranscriptionOutput:
         self._capturing = False
         self._pushed_text = ""
 
+    def set_segment_id(self, segment_id: str) -> None:
+        """Key the next segment on an upstream id (e.g. a realtime item_id), so a
+        repeated final for the same item reuses the segment id and clients revise
+        the segment in place instead of rendering a new one."""
+        if self._capturing:
+            if self._current_id == segment_id:
+                return
+            if self._pushed_text:
+                self.flush()
+            else:
+                # nothing visible was published yet: re-key the open segment
+                self._current_id = segment_id
+                return
+        self._next_segment_id = segment_id
+
     @utils.log_exceptions(logger=logger)
     async def capture_text(self, text: str) -> None:
         if self._participant_identity is None or self._track_id is None:
@@ -267,6 +283,9 @@ class _ParticipantLegacyTranscriptionOutput:
 
         if not self._capturing:
             self._reset_state()
+            if self._next_segment_id is not None:
+                self._current_id = self._next_segment_id
+                self._next_segment_id = None
             self._capturing = True
 
         if self._is_delta_stream:
@@ -385,6 +404,7 @@ class _ParticipantStreamTranscriptionOutput:
         self._room.on("local_track_published", self._on_local_track_published)
         self._flush_atask: asyncio.Task[None] | None = None
         self._closed = False
+        self._next_segment_id: str | None = None
 
         self._reset_state()
         self.set_participant(participant)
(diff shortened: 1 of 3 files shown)


longcw:
closing in favor of https://github.com/livekit/agents/pull/6729

PR #6734, opened 2026-08-06
feat(fishaudio): enable quality-guard on all TTS requests

cshape:
Adds `features: ["quality-guard"]` to every Fish Audio TTS request body, on both the HTTP (`/v1/tts`) and WebSocket (`/v1/tts/live`) paths.

Quality guard is a server-side reliability feature that increases the stability of generations by catching errors and automatically retrying — built for our enterprise users. It has no drawback, so it's enabled always by default rather than exposed as an option.

- Verified live against both endpoints: the field is accepted and synthesis is unaffected.
- Unit test pins the flag on every request.


davidzhao (approved):
lgtm!

theomonnom (approved):

PR #6735, opened 2026-08-06
fix(cartesia): narrow aligned_transcript when timestamps unsupported (#6493)

### LINKED ISSUE #6493: Is capabilities.aligned_transcript meant to reflect requested vs. actually-delivered alignment?
**Bug Description**

We're frequently seeing this warning with default TTS configs on Cartesia and ElevenLabs:

```
`use_tts_aligned_transcript` is enabled but no agent transcript was returned from tts
```

Before filing this as a confirmed bug, we want to check whether our reading of the code is correct, since we may be misunderstanding something:

1. `cartesia.TTS(word_timestamps=True)` and `elevenlabs.TTS(sync_alignment=True)` (both defaults) appear to set `capabilities.aligned_transcript=True` unconditionally at construction, independent of whether the specific model/language combination in use actually returns timing data. Cartesia's own plugin has an internal check for this (`tts.py:184-197`, warns when the language/model combo doesn't support `word_timestamps`), but as far as we can tell it doesn't clear the capability flag in that case.
2. In `generation.py`, `timed_texts_fut` looks like it resolves to a real (truthy) `Chan` object as soon as the TTS node is a streaming `AsyncIterable` — before any timed-word data has actually arrived:

   ```python
   if not isinstance(tts_node, AsyncIterable):
       timed_texts_fut.set_result(None)
       return False
   timed_text_ch = aio.Chan[io.TimedString]()
   timed_texts_fut.set_result(timed_text_ch)
   ```
3. In `agent_activity.py`, this truthiness (not actual data arrival) is what flips `read_transcript_from_tts = True` for a segment:

   ```python
   if (
       segment.tts is not None
       and use_aligned_transcript
       and (timed_texts := await segment.tts.timed_texts_fut)
   ):
       transcript = timed_texts
       read_transcript_from_tts = True
   ```

So `tts.capabilities.aligned_transcript` seems to answer "was alignment requested at construction?" rather than "will this specific call return alignment data?" — is that an accurate read of the intent here, or is there a capability-narrowing/fallback mechanism elsewhere that we're missing?

We initially also suspected this could cause the affected segment's text to be dropped from the recorded transcript (via `forwarded_text`/`chat_ctx`), but we checked our own persisted session data across several occurrences of this warning and the transcript content was intact every time — so that particular theory doesn't hold up, and we're not raising it here. We mention it only so you know we looked and are not assuming a downstream data-loss impact we haven't actually observed.

**Expected Behavior**

Unclear to us — hence asking. If the capability flag is meant to reflect "was this requested" rather than "will this be delivered," is the current warning-only behavior (with no other visible effect we could detect) the intended handling? Or should the capability be narrowed to combinations known to work?

**Reproduction Steps**

1. Create an `AgentSession` with `tts=cartesia.TTS()` (defaults) or `tts=elevenlabs.TTS()` (defaults).
2. Run a voice session.
3. Observe the warning firing on a subset of utterances, with no other apparent effect on transcript content that we could detect.

**Operating System**: Linux (GKE container)

**Models Used**: ElevenLabs `eleven_multilingual_v2`; also reproduces with Cartesia `sonic-2` (both default configs)

**Package Versions**

```
livekit-agents==1.5.18
livekit-plugins-elevenlabs==<pinned version>
livekit-plugins-cartesia==<pinned version>
```

Confirmed the same defaults/mechanism are still present on `main` as of writing.

**Session/Room/Call IDs**

```
roomID: sim-55bcf056-c1a9-48ca-a3a6-31176914e54a
jobID: AJ_aSVNf4qQNTSD
```

**Proposed Solution**

None — we didn't want to guess at a fix before confirming whether our reading of the mechanism above is even correct.

**Additional Context**

Related: livekit/agents#3667 (identical warning reported Oct 2025, closed without a root-cause conclusion); [#3206](<https://github.com/livekit/agents/issues/3206>)/[#3423](<https://github.com/livekit/agents/issues/3423>)/[#3409](<https://github.com/livekit/agents/issues/3409>) (history of `use_tts_aligned_transcript` auto-enable, revert, and the warning's addition).

dorukdumlu:
﻿## Summary
- Narrow Cartesia `capabilities.aligned_transcript` to model/language combos that can actually deliver word timestamps (en/de/es/fr on sonic models; all languages on preview models), instead of mirroring the `word_timestamps` request flag.
- Re-evaluate the capability on `update_options` when model/language changes, and skip `add_timestamps` in the WS payload when unsupported.
- Warn with an explicit "model configuration does not support aligned transcript" message (#6493).

Fixes #6493

## Notes
- ElevenLabs `sync_alignment` has no published model/language restriction matrix comparable to Cartesia's, so its capability stays tied to the request flag. Happy to extend if there's a known unsupported combo.
- Delivery-time empty alignment streams (requested + supported but no `TimedString` arrives) are out of scope here — that needs a separate fallback path in `generation.py` / `agent_activity.py`.

## Test plan
- [x] `pytest tests/test_plugin_cartesia_tts.py --plugin cartesia` (11 passed)
- [ ] Confirm unsupported language (e.g. `language="ja"`, `sonic-3`, `word_timestamps=True`) no longer triggers `use_tts_aligned_transcript is enabled but no agent transcript was returned from tts` when aligned transcripts are enabled on the session
- [ ] Confirm supported language (`en`) still requests timestamps and keeps `aligned_transcript=True`


dorukdumlu:
Rebased onto current main. Resolved add/add conflict in `tests/test_plugin_cartesia_tts.py` by keeping both the API-key redaction tests from main and the aligned-transcript capability tests from this PR. All 13 cartesia plugin tests pass locally.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -302,6 +304,30 @@ def update_options(
...
+            return
+        # https://docs.cartesia.ai/api-reference/tts/compare-tts-endpoints
+        logger.warning(
+            "model configuration does not support aligned transcript "
+            "(word_timestamps); disabling capabilities.aligned_transcript. "
+            "Supported: languages en, de, es, and fr with `sonic` models, "
+            "or all languages with `preview` models. model=%s language=%s",
+            self._opts.model,
+            self._opts.language.language if self._opts.language else None,
+        )
```
longcw: style: collapse them into one method? for example
```python
def _sync_aligned_transcript(self) -> None:
    """Resolve word_timestamps for the current model/language and mirror it on the capability."""
    supported = _supports_word_timestamps(self._opts.model, self._opts.language)
    if is_given(self._word_timestamps):
        enabled = self._word_timestamps
        if enabled and not supported:
            # https://docs.cartesia.ai/api-reference/tts/compare-tts-endpoints
            logger.warning(
                "word_timestamps was requested for model %s with language %s, which is not known "
                "to return them, aligned transcripts may be empty",
                self._opts.model,
                self._opts.language.language if self._opts.language else None,
            )
    else:
        enabled = supported

    self._opts.word_timestamps = enabled
    self._capabilities.aligned_transcript = enabled
```

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -127,10 +141,13 @@ def __init__(
             base_url (str, optional): The base URL for the Cartesia API. Defaults to "https://api.cartesia.ai".
         """  # noqa: E501
 
+        language_code = LanguageCode(language) if language else None
+        aligned_transcript = word_timestamps and _supports_word_timestamps(model, language_code)
```
longcw: maybe make `word_timestamps` default to `NOT_GIVEN` and skip the gate when user set it to True, and auto decide when it's not given.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
dorukdumlu: Done in 9485828. Collapsed both into _sync_aligned_transcript, used your version.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
dorukdumlu: Done in 9485828. Default is NOT_GIVEN now: auto-decides from the model/language combo, explicit True skips the gate and only warns, explicit False disables. add_timestamps reuses the resolved value so the gate in _to_cartesia_options is gone too.

livekit-plugins/livekit-plugins-cartesia/livekit/plugins/cartesia/tts.py
```diff
@@ -59,6 +59,20 @@
     _is_sonic_3,
 )
 
+# Word timestamps on non-preview sonic models are limited to these languages.
+# Preview models support all languages. See Cartesia TTS docs / #6493.
+_WORD_TIMESTAMP_LANGUAGES = frozenset({"en", "de", "es", "fr"})
```
chenghao-mou: Is there a link to this limitation? I couldn't find anything about language-based limit on their doc page.

chenghao-mou:
I've reached out to Cartesia about their documentation as the [current documentation](https://docs.cartesia.ai/use-the-api/compare-tts-endpoints) only says that model and language limits apply. Its [WebSocket reference](https://docs.cartesia.ai/api-reference/tts/websocket) does not provide the actual matrix.

dorukdumlu:
Thanks for checking with them. The en/de/es/fr set came from what the docs said back when I was digging into #6493, plus the combos that were failing in that issue. You're right that the current pages don't publish the matrix anymore, so whatever Cartesia tells you wins, it's a one line frozenset update either way.

One thing that makes this low risk: if the set is too narrow we just fall back to the plain transcript for a combo that could have had timestamps, and if it's too wide that combo behaves the same as main does today (timestamps requested, nothing arrives). Preview models and unset language are treated as supported, so unknown cases fail open.

Also rebased onto current main, the test file had grown some pooled-websocket tests in the meantime. All 18 cartesia plugin tests pass locally.

If Cartesia won't publish a matrix at all, I could instead flip the capability at runtime when the first chunk comes back without timestamps. That's the delivery-time fallback I marked out of scope in the description, but happy to go that way if you'd rather not hardcode anything.

dorukdumlu:
Devin's right here, took the suggestion in c97d591. My original thinking was that an explicit True should act as an escape hatch for a stale matrix, but the two concerns split cleanly: the flag keeps controlling the wire request (add_timestamps still goes out, so if Cartesia starts delivering for a combo we have wrong, that's observable), while the capability only claims combos known to deliver. Advertising alignment that never arrives would recreate exactly the empty-transcript path this PR is closing, and the reverse mistake just degrades to the plain transcript. Updated the warning text and the two tests that pinned the old behavior.

PR #6745, opened 2026-08-07
fix: hold pending replies at playout launch

chenghao-mou:
**Why:** A pending reply can launch into resumed user speech when forwarding clears an SOS hold or SOS precedes the current handle.

**Compared with #6733:** Use the playout silence gate instead of lagging `user_state`, preserve an existing handle hold, and require reconciliation in `say()`, pipeline, and realtime paths.

**Credit:** This supersedes and replaces #6733 by @ajayarora1235, who is retained as a commit co-author.

Fixes AGT-3240


livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -511,12 +514,14 @@ async def _audio_forwarding_task(
...
     out: _AudioOutput,
+    *,
+    prepare_playout: Callable[[], None],
 ) -> None:
     resampler: rtc.AudioResampler | None = None
 
     cancelled = False
     try:
-        audio_output.resume()
+        prepare_playout()
```
theomonnom: If the audio output is paused here? When is the current speech ever starting again?

chenghao-mou: `on_end_of_speech` or turn commit resolves the hold. 

It is the same hold as STT SOS. If the speech already existed at SOS, this preserves that hold; if SOS arrived before the speech existed, this applies the hold later at playout launch.

ajayarora1235 (approved):

livekit-agents/livekit/agents/voice/generation.py
theomonnom: Oh, can you add a comment, it isn't super obvious, like the playout is being paused from random places. 

It's hard to track what can block, what can resume the playout. Feels kinda all over the place rn IMO

chenghao-mou: Expanded the comments in e913b30f9. The four `AgentActivity` pause sites now connect pause creation, preservation, deferred application, and release.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -4275,6 +4286,43 @@ def _pause_enabled(self) -> bool:
             and self._session.output.audio.can_pause
         )
 
+    def _reconcile_playout_pause(self, speech_handle: SpeechHandle) -> None:
+        """Preserve, apply, or release a speech pause before forwarding audio."""
+        audio_output = self._session.output.audio
```
longcw, line 4292: check `output.audio_enabled`?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -4275,6 +4286,43 @@ def _pause_enabled(self) -> bool:
...
+        if (
+            pause_is_allowed
+            and self._session.agent_state != "speaking"
+            and not self._user_silence_event.is_set()
+        ):
+            assert audio_output is not None
+            # SOS arrived before this handle became current so we pause here
+            # EOS/transcripts/turn commit will resolve the pause eventually
+            self._update_paused_speech(speech_handle, timeout=0)
+            audio_output.pause()
```
longcw, line 4321: this also blocks `session.say()` when user is speaking?

chenghao-mou: I would assume that's expected behaviour and one needs `allow_interruptions=False` to force a speech, unless there is something I am missing.

longcw (approved):

PR #6746, opened 2026-08-07
fix(elevenlabs): send keyterms on the realtime STT WebSocket

pawlowskipawel:
## Problem

`keyterms` is accepted by `elevenlabs.STT(...)` but only ever sent on the **batch**
request. It is dropped on the **realtime** WebSocket path, so `scribe_v2_realtime`
sessions get no keyterm biasing even when the caller passes them.

ElevenLabs does support it there, as repeated `keyterms` query params on the connect URL:

- [Realtime STT API reference](https://elevenlabs.io/docs/api-reference/speech-to-text/v-1-speech-to-text-realtime) lists `keyterms` — *"List of keyterms the model is biased towards."*
- [Keyterm prompting guide](https://elevenlabs.io/docs/eleven-api/guides/how-to/speech-to-text/batch/keyterm-prompting) — *"also available for the realtime Speech to Text WebSocket API."* Batch allows 1000 keyterms × 50 chars, realtime 50 × 20.

The docstring claims the opposite. It was accurate when written, before ElevenLabs
shipped realtime keyterms — `no_verbatim`, released alongside them, **was** wired into
`_connect_ws`, but `keyterms` was missed and the docstring never updated.

## Changes

- `_connect_ws` appends repeated `keyterms=` query params, percent-encoded.
- `SpeechStream.update_options` accepts `keyterms` and triggers a reconnect, mirroring
  `no_verbatim`. `STT.update_options` forwards them to active streams.
- Docstring corrected with the real per-mode limits.
- 5 tests, following the existing `enable_logging` / `no_verbatim` patterns.

Keyterms are free-form text, unlike the other params in that URL (model ids, enums,
numbers, language codes), so they need encoding: verified against `yarl.URL` — the parser
`aiohttp.ws_connect` uses — an unescaped `&` splits `Smith & Sons` into `keyterms='Smith '`
plus a bogus `' Sons'` param, and `#` truncates `C#` to `C`.
`test_connect_ws_escapes_query_delimiters_in_keyterms` pins this via
`URL(...).query.getall("keyterms")`.

## Verification

`make check` clean (ruff, mypy over 629 files). `pytest tests/test_plugin_elevenlabs_stt.py` — 25 passed.

Those mock the socket, so they only prove the URL is built correctly. Checked separately
against the live endpoint: one TTS-synthesized clip per term, transcribed twice over
`scribe_v2_realtime` — **identical audio bytes** in both arms, so any difference isolates
the keyterm.

| term | without keyterms | with keyterms |
|---|---|---|
| `Sonic3000` | `Sonic 3000` | **`Sonic3000`** |
| `GPT4o` | `GPT-4o` | **`GPT4o`** |
| `K2Turbo` | `K2 Turbo` | **`K2Turbo`** |
| `nova-3` | `Nova 3` | **`nova-3`** |
| `scribe-v2` | `Scribe V2` | **`scribe-v2`** |

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
@@ -22,6 +22,7 @@ import os
 import weakref
 from dataclasses import dataclass
 from typing import Any, Literal, TypedDict
+from urllib.parse import quote
 
 import aiohttp
 
@@ -137,9 +138,9 @@ class STT(stt.STT):
                 be selected based on parameters provided.
             model_id (ElevenLabsSTTModels | str): Deprecated alias for `model`. Use `model` instead.
             keyterms (NotGivenOr[list[str]]): A list of keywords or phrases to bias the transcription towards.
-                Each keyterm can contain at most 5 words and must be less than 50 characters.
-                Maximum of 100 keyterms. Only supported for Scribe v2 batch recognition
-                (not realtime streaming). Usage incurs additional costs.
+                Supported for both Scribe v2 (batch) and Scribe v2 realtime. Batch accepts up to
+                1000 keyterms of at most 50 characters each; realtime accepts up to 50 keyterms of
+                at most 20 characters each. Usage incurs additional costs.
             no_verbatim (NotGivenOr[bool]): When True, the model removes filler words, false starts
                 and disfluencies from the transcript, producing cleaner output. Supported for both
                 Scribe v2 (batch) and Scribe v2 realtime. Default is False.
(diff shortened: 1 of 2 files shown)


longcw (approved):

livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
longcw, line 188: maybe out of scope of this pr, but worth to mention we can set the `STTCapabilities.keyterms` to True if the stt model support keyterms, and customize `STT._update_session_keyterms` in the plugin to support the [automatic keyterm detection](https://github.com/livekit/agents/pull/6039).

PR #6749, opened 2026-08-07
feat(spatius): default audio format to ogg_opus

3DRX:
Change the default audio format for the Spatius avatar plugin from pcm_s16le to ogg_opus. This is the audio format sent from livekit-plugins-spatius to Spatius via WebSocket. With ogg_opus, it saves bandwidth and also reduces latency for this hop.

This only changes the fallback default. Existing behavior is preserved for anyone who:
- passes audio_format= explicitly
- sets the SPATIUS_AUDIO_FORMAT environment variable (e.g. SPATIUS_AUDIO_FORMAT=pcm_s16le)


3DRX:
One caveat regarding the sample rate: libopus only natively supports 8000/12000/16000/24000/48000 Hz in audio sample rate, but Azure TTS, CambAI, Bland, ElevenLabs might produce audio in 22050/44100 Hz.

What do you think is the best solution here? Perhaps to add a resample layer for sample rates that are not natively supportted by libopus?

longcw:
> One caveat regarding the sample rate: libopus only natively supports 8000/12000/16000/24000/48000 Hz in audio sample rate, but Azure TTS, CambAI, Bland, ElevenLabs might produce audio in 22050/44100 Hz.
> 
> What do you think is the best solution here? Perhaps to add a resample layer for sample rates that are not natively supportted by libopus?

you can specify the sample rate in AudioOutput and the framework will resample it to the target sample rate.

longcw (approved):

PR #6751, opened 2026-08-07
fix(voice): keep transcript pacing speed nonzero for syllable-free segments

kevtae:
## Problem

We hit this in production (livekit-agents 1.6.5, still reproducible on current main):

```
ZeroDivisionError: float division by zero
  _main_task in livekit/agents/voice/transcription/synchronizer.py, line 417
```

When a segment's visible text hyphenates to **zero syllables** — e.g. the turn is markup-only (`<expr .../>`), punctuation, or emoji — while TTS still produced audio, `_reestimate_speed()` computes:

```python
self._speed = pushed_hyphens / self._audio_data.pushed_duration  # 0 / duration -> 0.0
```

Every subsequent pacing division in `_main_task` then raises (in Python even `0.0 / 0.0` raises `ZeroDivisionError`):

```python
delay = max(0.0, word_hyphens - d_hyphens) / self._speed
```

The task dies mid-segment and the rest of that segment's transcript is never forwarded to the room.

## Fix

Only overwrite `self._speed` when the segment actually has visible syllables (`pushed_hyphens > 0`); otherwise keep the default `STANDARD_SPEECH_RATE`-based estimate from `__init__`. The `_speed_on_speaking_unit` path is already safe — a `0.0` value is falsy and skips that branch.

## Test

Added a regression test in `tests/test_transcript_sync_markup.py`: 1s of audio with a bare `<expr type="sound" label="laugh"/>` as the entire turn. Without the fix it reproduces the exact crash above; with the fix the segment forwards all raw tokens and `_speed` stays positive.

One subtlety: the markup-only turn must have no trailing whitespace — with `retain_format=True` the tokenizer keeps a trailing space as a token, which hyphenates to one "syllable" and masks the repro.

`test_transcript_sync_markup.py`, `test_transcription_filter.py`, and `test_transcription_timeout.py` pass locally (366 tests).

https://claude.ai/code/session_01HcKtvDBg6Kp5JUNjvuw2Sg

### DIFF
--- a/livekit-agents/livekit/agents/voice/transcription/synchronizer.py
+++ b/livekit-agents/livekit/agents/voice/transcription/synchronizer.py
@@ -300,8 +300,11 @@ class _SegmentSynchronizerImpl:
         # pace against the visible text only so expressive tags don't inflate the speed
         clean_pushed_text = strip_all_markup(self._text_data.pushed_text)
         pushed_hyphens = len(self._calc_hyphens(clean_pushed_text))
-        # hyphens per second
-        if self._audio_data.pushed_duration > 0:
+        # hyphens per second. A segment whose visible text has no syllables at all
+        # (markup-only, punctuation, emoji) would estimate a speed of 0 and every
+        # pacing division in _main_task would raise ZeroDivisionError — keep the
+        # default estimate instead.
+        if pushed_hyphens > 0 and self._audio_data.pushed_duration > 0:
             self._speed = pushed_hyphens / self._audio_data.pushed_duration
 
         # hyphens per speaking unit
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/voice/transcription/synchronizer.py
longcw: `pushed_hyphens` is used below for the same purpose, maybe fix it there as well

```suggestion
        if pushed_hyphens > 0:
            # hyphens per second
            if self._audio_data.pushed_duration > 0:
                self._speed = pushed_hyphens / self._audio_data.pushed_duration

            # hyphens per speaking unit
            if (
                units := self._audio_data.estimated_rate.accumulate_to(
                    self._audio_data.pushed_duration
                )
            ) > 0:
                self._speed_on_speaking_unit = pushed_hyphens / units

```

longcw (approved):
looks good to me! another place may also need the fix

kevtae:
Friendly ping — this has been approved since Aug 11 (review follow-up pushed as c499253, guard hoisted to cover _speed_on_speaking_unit per the review). Anything else needed from my side to get it merged? We're carrying the ZeroDivision transcript loss in production until a release includes this.

PR #6752, opened 2026-08-07
feat(bland): BTTS ws integration

rv-blandai:
- Websocket-based realtime streaming through `/v2/tts/ws` endpoint
- Proper input streaming (ex. from an LLM text stream) is supported now
- Significant improvements to latency all around

### DIFF
--- a/livekit-plugins/livekit-plugins-bland/livekit/plugins/bland/tts.py
+++ b/livekit-plugins/livekit-plugins-bland/livekit/plugins/bland/tts.py
@@ -38,7 +38,7 @@ from livekit.agents.utils import is_given
 from .log import logger
 
 DEFAULT_BASE_URL = "https://api.bland.ai/v2"
-DEFAULT_VOICE_ID = "f04af0e5-1a80-48a9-b02d-52f30d417cfa"
+DEFAULT_VOICE_ID = "29158307-9893-4149-8a75-bc9ce313d64e"
 DEFAULT_SAMPLE_RATE = 48000
 NUM_CHANNELS = 1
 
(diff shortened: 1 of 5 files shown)


livekit-plugins/livekit-plugins-bland/livekit/plugins/bland/tts.py
```diff
@@ -79,7 +114,7 @@ def __init__(
             raise ValueError(f"sample_rate must be one of {SAMPLE_RATES}, got {sample_rate}")
 
         super().__init__(
-            capabilities=tts.TTSCapabilities(streaming=False),
+            capabilities=tts.TTSCapabilities(streaming=True),
```
tinalenguyen: should we expose a parameter, `streaming: bool = True` so the user can choose if they want to use the WS impl?

tests/test_tts.py
```diff
@@ -490,6 +490,13 @@ async def test_tts_synthesize_retry_after_partial_audio():
         },
         id="cartesia",
     ),
+    pytest.param(
```
tinalenguyen: let's remove the changes to this file, we don't have an api key set up to test as of yet

livekit-plugins/livekit-plugins-bland/livekit/plugins/bland/tts.py
rv-blandai: I think we should, I'll make that change. As of now the WS endpoint is almost strictly better but making it an option will be easier now than later.

livekit-plugins/livekit-plugins-bland/livekit/plugins/bland/tts.py
```diff
@@ -188,6 +318,222 @@ async def _run(self, output_emitter: tts.AudioEmitter) -> None:
...
+                            await asyncio.wait_for(
+                                cancel_and_drain(ws),
+                                timeout=min(self._conn_options.timeout, 2.0),
+                            )
+                        except BaseException:
+                            try:
+                                await ws.close()
+                            except Exception:
+                                pass
+                            raise
```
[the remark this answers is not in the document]
tinalenguyen: @rv-blandai could you verify if this is a real issue?

tests/test_tts.py
tinalenguyen: bump on this

tinalenguyen (approved):
thanks for the pr!

PR #6759, opened 2026-08-09
fix: respect non-retryable STT errors

chenghao-mou:
**Why:** Cartesia wraps failed WebSocket handshakes as `APIStatusError`. The STT retry loop ignored `retryable=False`, so 401 responses retried four times and ended as `APIConnectionError`.

Preserve the original status error and fail immediately. Trim change-narrative comments from the credential-redaction patch.

Fixes AGT-3238


davidzhao (approved):
lgtm

PR #6762, opened 2026-08-10
Add Gandr TTS plugin

AALG123:
Adds `livekit-plugins-gandr`, a TTS plugin for [Gandr](https://gandr.ai).

The package is on PyPI as `livekit-plugins-gandr`. Layout and pyproject follow the existing plugins.

```python
from livekit.plugins import gandr

session = AgentSession(
    tts=gandr.TTS(voice="gandr-mia"),
)
```

Verified in a clean virtual environment against `livekit-agents`: it imports as `livekit.plugins.gandr`, constructs, is a real `livekit.agents.tts.TTS`, reports 24 kHz mono, and exposes `prewarm` and `update_options`.

Two behaviours worth a reviewer's attention, both chosen for live calls:

- A request refused because capacity is busy is retried once, so the turn stays alive rather than the caller hearing silence.
- A stream the API marks as truncated raises instead of playing a partial sentence. That hands the turn to the agent retry, and to `FallbackAdapter` where one is configured. A visible retry is better than a silent half sentence when the sentence carries a dose or an address.

Voices are `gandr-mia`, `gandr-ava`, `gandr-jenny`, `gandr-dane`, `gandr-leo`, `gandr-lewis`, or a `gnd:` clone id, swappable mid-session via `update_options`. The API key is read from `GANDR_API_KEY`.

Happy to adjust anything to match your conventions.


AALG123:
Pushed a fix set that addresses the Devin review findings and gets the local gates green:

- prewarm is now a sync method that schedules the awaitable work, so the framework's fire-and-forget call actually wakes the worker instead of dropping a coroutine (matches the base TTS contract and the agent_activity call sites).
- A second consecutive cold refusal in _run is converted to a retryable APIStatusError instead of leaking the internal exception to the stream consumer.
- The package now registers a Plugin (following the rime plugin), so it participates in plugin discovery.
- ChunkedStream pins the concrete GandrTTS type, which is what the internal _prewarm_async and _ensure_session calls need for mypy.
- Removed the trailing whitespace and ran the repo's own ruff format.

Verified locally: ruff check, ruff format --check, py_compile, and mypy on livekit.plugins.gandr all pass. The two remaining mypy failures in the full repo run are pre-existing in livekit-plugins-bithuman (missing cv2/loguru stubs), unrelated to this package.

PR #6765, opened 2026-08-10
fix: support adaptive interruption without word timestamps

chenghao-mou:
Adaptive interruption currently requires aligned word timestamps, which excludes STT providers such as Ink-2. Gate transcripts by local event creation time and active VAD speech instead. Held transcripts flush in provider order before normal interruption thresholds apply.

Fixes AGT-3235

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent.py
+++ b/livekit-agents/livekit/agents/voice/agent.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import asyncio
 import time
 from collections.abc import AsyncGenerator, AsyncIterable, Coroutine, Generator
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from typing import TYPE_CHECKING, Any, Generic, Literal, TypeVar
 
 from livekit import rtc
@@ -526,6 +526,17 @@ class Agent:
                 forward_task = asyncio.create_task(_forward_input())
                 try:
                     async for event in stream:
+                        if (
+                            event.speech_end_time is None
+                            and wrapped_stt.capabilities.aligned_transcript
+                            and event.alternatives
+                            and event.alternatives[0].end_time > 0
+                        ):
+                            speech_end_time = (
+                                _audio_input_started_at + event.alternatives[0].end_time
+                            )
+                            if speech_end_time <= event.created_at:
+                                event = replace(event, speech_end_time=speech_end_time)
                         yield event
                 finally:
                     await utils.aio.cancel_and_wait(forward_task)
--- a/livekit-agents/livekit/agents/voice/agent.py
+++ b/livekit-agents/livekit/agents/voice/agent.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import asyncio
 import time
 from collections.abc import AsyncGenerator, AsyncIterable, Coroutine, Generator
-from dataclasses import dataclass, replace
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Generic, Literal, TypeVar
 
 from livekit import rtc
@@ -526,17 +526,6 @@ class Agent:
                 forward_task = asyncio.create_task(_forward_input())
                 try:
                     async for event in stream:
-                        if (
-                            event.speech_end_time is None
-                            and wrapped_stt.capabilities.aligned_transcript
-                            and event.alternatives
-                            and event.alternatives[0].end_time > 0
-                        ):
-                            speech_end_time = (
-                                _audio_input_started_at + event.alternatives[0].end_time
-                            )
-                            if speech_end_time <= event.created_at:
-                                event = replace(event, speech_end_time=speech_end_time)
                         yield event
                 finally:
                     await utils.aio.cancel_and_wait(forward_task)
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -302,6 +302,7 @@ class AgentActivity(RecognitionHooks):
         )
         self._interruption_detection_enabled: bool = self._interruption_detector is not None
         self._pending_interruption: inference.OverlappingSpeechEvent | None = None
+        self._audio_activity_interruption_in_progress: bool = False
 
         # this allows taking over audio interruption temporarily until interruption is detected
         # by default it is true unless turn_detection is manual or realtime_llm
@@ -2080,6 +2081,16 @@ class AgentActivity(RecognitionHooks):
 
     def _interrupt_by_audio_activity(self) -> None:
         """Interrupt the current speech or generation from detected audio activity."""
+        if self._audio_activity_interruption_in_progress:
+            return
+
+        self._audio_activity_interruption_in_progress = True
+        try:
+            self._interrupt_by_audio_activity_once()
+        finally:
+            self._audio_activity_interruption_in_progress = False
+
+    def _interrupt_by_audio_activity_once(self) -> None:
         if not self._interruption_by_audio_activity_enabled:
             return
 
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -301,8 +301,7 @@ class AgentActivity(RecognitionHooks):
             self._resolve_interruption_detection()
         )
         self._interruption_detection_enabled: bool = self._interruption_detector is not None
-        self._pending_interruption: inference.OverlappingSpeechEvent | None = None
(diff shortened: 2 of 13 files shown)


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1420,6 +1287,10 @@ async def _on_vad_event(self, ev: vad.VADEvent) -> None:
             with trace.use_span(self._ensure_user_turn_span()):
                 self._hooks.on_end_of_speech(ev)
 
+            self._active_vad_speech_started_at = None
+            if self._agent_speaking:
+                flush_start = self._transcript_flush_start(now=time.time())
+                self._trim_held_transcripts(flush_start=flush_start)
```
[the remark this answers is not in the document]
chenghao-mou: Good catch. The exact ordering is slightly different: the EOS sentinel produces a false verdict, while a true verdict can already be queued before EOS and consumed after the synchronous trim. The resulting race is the same—the early transcript can be removed before the true verdict flushes it.

Fixed in eb6f73e45 by keeping events held until the terminal verdict. A true verdict flushes from its original `overlap_started_at`; a false verdict trims from its event-time backchannel boundary. The regression tests cover the delayed true-verdict path, false-verdict cleanup, and provider order.

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1445,6 +1315,10 @@ async def _on_overlap_speech_event(self, ev: inference.OverlappingSpeechEvent) -
         # every verdict is terminal for its overlap, including one the cooldown then ignores
         self._overlap_open = False
 
+        if not ev.is_interruption and self._agent_speaking:
+            flush_start = self._transcript_flush_start(now=ev.detected_at)
+            self._trim_held_transcripts(flush_start=flush_start)
+
         if self._backchannel_boundary_active and not ev.is_interruption:
             logger.trace(
                 "ignoring backchannel event during backchannel boundary cooldown, falling back to vad"
```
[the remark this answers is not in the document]
chenghao-mou: Confirmed. During an active backchannel boundary, the false adaptive verdict is ignored and VAD remains authoritative, so the held transcripts must remain available to the fallback path. Fixed in f03ad0a27 by moving the false-verdict trim below the boundary early return. Added a regression with an asymmetric `(3.0, 0.5)` boundary to cover the destructive case.

longcw:
question: what is the difference between this pr and how we support adaptive interruption for realtime model without STT? can we make the logic shared for these two kinds of use cases?

chenghao-mou:
> question: what is the difference between this pr and how we support adaptive interruption for realtime model without STT? can we make the logic shared for these two kinds of use cases?

I don't think we can clear audio for streaming STT, and if we withhold audio from STT, we then could trigger inactivity timeout/reconnection.

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -639,147 +640,59 @@ async def _wait_for_user_silence(self) -> None:
...
+        end_boundary = self._backchannel_boundary[1] if self._backchannel_boundary else 0.0
+        flush_start = now - end_boundary
+        if vad_speech_started_at is not None:
+            flush_start = min(flush_start, vad_speech_started_at)
+        if self._agent_speech_started_at is not None:
+            flush_start = max(flush_start, self._agent_speech_started_at)
+        return flush_start
+
+    def _trim_held_transcripts(self, *, flush_start: float) -> None:
+        while self._transcript_buffer and (self._transcript_buffer[0].created_at < flush_start):
```
longcw: `created_at` can't separate "audio spoken during agent speech" from "audio spoken before agent speech whose transcript arrived late." it may drop the transcripts when STT is slow?

should we keep the stt timestamp test as an additional condition when timestamps are usable, arrival time as fallback?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2044,9 +2046,13 @@ def _interrupt_by_audio_activity(
 
         if self._rt_turn_detection_enabled:
             # ignore if realtime model has turn detection enabled
+            self._pending_interruption = None
             return
 
         interruption_options = self._session.options.interruption
+        if self._audio_recognition is not None:
+            self._audio_recognition._release_transcripts_for_audio_activity()
```
longcw: should we run this after the `interruption_options["min_words"]` check?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -4628,12 +4623,8 @@ def _resolve_interruption_detection(self) -> inference.AdaptiveInterruptionDetec
...
             can_gatekeep = not self._rt_turn_detection_enabled
         else:
-            # the STT pipeline gatekeeps by holding and flushing transcripts
-            can_gatekeep = (
-                self.stt is not None
-                and self.stt.capabilities.aligned_transcript
-                and self.stt.capabilities.streaming
-            )
+            # Local arrival time plus VAD makes transcript alignment unnecessary.
+            can_gatekeep = self.stt is not None
```
longcw, line 4705: question: does this also work for non-streaming STT now?

livekit-agents/livekit/agents/voice/audio_recognition.py
chenghao-mou: Good point. Initially, I thought we can get rid of the old timing logic and rely on end pointing delay + backchannel boundary, but late STT transcript can still arrive after all that. Even if we get an overlapping speech, a late transcript can arrive after that.

I am going to restore that timestamp and use created_at as a fallback.

livekit-agents/livekit/agents/voice/agent_activity.py
chenghao-mou: The idea is that we handle it over to VAD completely when barge-in returns positive. Even if we keep running, we would be wasting inference on known positive audio window, so it becomes equivalent to VAD interruption really.

livekit-agents/livekit/agents/voice/agent_activity.py
chenghao-mou: Yes. And with the original timestamp restored, we can track the STT event time for non-streaming ones too.

livekit-agents/livekit/agents/voice/agent.py
```diff
@@ -527,6 +526,17 @@ async def _forward_input() -> None:
...
+                            event.speech_end_time is None
+                            and wrapped_stt.capabilities.aligned_transcript
+                            and event.alternatives
+                            and event.alternatives[0].end_time > 0
+                        ):
+                            speech_end_time = (
+                                _audio_input_started_at + event.alternatives[0].end_time
+                            )
+                            if speech_end_time <= event.created_at:
+                                event = replace(event, speech_end_time=speech_end_time)
```
longcw: this lives in `Agent.default.stt_node` may be ignored if user customzied the `stt_node`?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -563,11 +568,26 @@ def _on_start_of_overlap_speech(
...
+            start_boundary > 0
+            and self._agent_speech_started_at is not None
+            and started_at <= self._agent_speech_started_at + start_boundary
+        )
+        if (
+            self._hooks.interruption_by_audio_activity_enabled
+            or self._backchannel_boundary_active
+            or started_in_boundary
+        ):
+            self._hooks.interruption_by_audio_activity_enabled = True
```
longcw, line 587: from claude code:
`_on_start_of_overlap_speech` sets `interruption_by_audio_activity_enabled = True`, and that setter cancels the backchannel boundary — which discards the only callback that turns VAD interruption back off. All four `_disable_vad_interruption_soon` call sites fire once per agent-speech interval, so nothing re-arms it. so one overlap inside the start boundary switches the rest of that agent speech to VAD interruption.

chenghao-mou: yes, this is intentional. If user speech is already active when agent playout starts, or its onset falls within `boundary[0]`, the current audible agent-speech interval stays on the VAD/STT interruption path. 

Because the final transcript for that utterance can arrive later, and STT events do not have a stable utterance ID. A non-empty final transcript received while agent audio is playing and no adaptive transcript gate is active (e.g. no overlap VAD speech after boundary) also selects the VAD/STT path. 

In theory, we could use the speech_end_time, but it feels heavy to bring more timestamp filtering logic to support it while vad interruption fallback feels safer and simpler. wdyt?

longcw (approved):

PR #6776, opened 2026-08-10
fix(sarvam): validate TTS parameters against the live Bulbul API

### LINKED ISSUE #6774: sarvam: TTS parameter validation diverges from the live Bulbul API
### Bug Description

---

### Expected Behavior

   recognize are not offered.

### Reproduction Steps

```bash

```

21lakshh:
Fixes #6774

The Sarvam TTS plugin's client-side validation disagreed with the Bulbul API in
three places. Each fix is a separate commit.

### 1. Speaker table for `bulbul:v3`

The plugin listed 30 speakers; the API and the docs both accept 37. Since
`validate_model_speaker_compatibility` raises, `anand, tarun, sunny, mani,
gokul, vijay, mohit, rehan, soham` were unusable despite synthesizing fine.
`amelia` and `sophia` were offered but rejected by the API.

v3 and v3-beta accept the same set, so they now share one frozenset. The
per-model `female`/`male` keys were dropped — only `"all"` was ever read, and
the split couldn't be verified against the API.

### 2. REST-only sample rates when streaming

32000/44100/48000 Hz are REST-only, but the constructor accepted them
unconditionally, so `synthesize()` worked while `stream()` failed on every call
with `Speech sample rate can only be 8000, 16000, 22050, 24000 Hz`. An
`AgentSession` always streams, so this hit the first utterance.

`SynthesizeStream` now fails fast with a clear local error before taking a
pooled connection; the constructor warns at config time. `synthesize()` still
accepts the full range.

### 3. `pace` range per model

v3/v3-beta accept 0.5–2.0, v2 accepts 0.3–3.0. The plugin applied the v2 range
to both, so `pace=2.5` on v3 passed validation then 400'd. The range is now
resolved per model, and `update_options(model=...)` re-checks the pace already
in effect.

### Verification

All findings were confirmed against the live API (REST and WebSocket) with a
real key, and cross-checked with the Bulbul docs — the docs and the API agree on
the 37-speaker list exactly. Boundary values were exercised on both endpoints;
`pace` output length scales inversely with the value at both limits, confirming
the parameter takes effect rather than being ignored.

New tests in `tests/test_plugin_sarvam_tts.py` (36, hermetic — dummy key, no
network).

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
@@ -99,6 +99,13 @@ _CODEC_TO_MIME: dict[str, str] = {
 
 _TELEPHONY_CODECS: frozenset[str] = frozenset({"mulaw", "alaw"})
 
+# 32000 / 44100 / 48000 Hz are documented as REST-only; the WebSocket endpoint
+# is capped at 24 kHz. bulbul:v3 / v3-beta reject the higher rates with an error
+# frame, so streaming is restricted to STREAMING_SAMPLE_RATES for every model.
+# See https://docs.sarvam.ai/api/getting-started/models/bulbul
+ALLOWED_SAMPLE_RATES: frozenset[int] = frozenset({8000, 16000, 22050, 24000, 32000, 44100, 48000})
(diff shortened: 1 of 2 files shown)


21lakshh (commented):
lgtm, devin checks passed top

Issue #6761, opened 2026-08-09
Contributing a SonexLabs TTS plugin — existing PyPI package name conflict


ayushi-agarwall:
Hi team,

I maintain SonexLabs's TTS API and would like to contribute a `livekit-plugins-sonex` plugin to this monorepo so it can be listed on the [TTS plugins page](https://docs.livekit.io/agents/models/tts/).

Before submitting a PR, I wanted to flag one thing: I already publish a standalone package named `livekit-plugins-sonex` on PyPI under my own account (source: https://github.com/sonexlabs/livekit-plugins-sonex), following the pattern I used before learning about the monorepo contribution process.

Since plugins merged into this repo are published to PyPI by LiveKit's own release automation under the same `livekit-plugins-<name>` naming convention, I want to check the right path forward before opening a PR:

1. Should I transfer/yield the existing `livekit-plugins-sonex` PyPI name to LiveKit's publishing account, or
2. Would you prefer a different package name for the monorepo version, or
3. Is there a preferred process for this situation?

Happy to follow whatever convention works best. Thanks\!

tinalenguyen:
hi @ayushi-agarwall, thanks for reaching out! we'd prefer if you open a PR with the implementation first, and once that's merged in you can transfer ownership to LiveKit

ayushi-agarwall:
Opened #6777 with the plugin implementation, per your suggestion above. CLA is signed. Also just pushed a follow-up commit addressing the automated review feedback (sample-rate reporting, Python 3.10 timeout handling, ruff format/lint). Happy to make any further changes needed — and can transfer the `livekit-plugins-sonex` PyPI package name to LiveKit once this is merged.

PR #6777, opened 2026-08-11
feat(sonex): add SonexLabs TTS plugin

ayushi-agarwall:
## What

Adds `livekit-plugins-sonex`, a text-to-speech plugin for [SonexLabs](https://sonexlabs.com)' Panini TTS API.

## Why

Follow-up to #6761 — per @tinalenguyen's guidance, opening this PR with the implementation first; happy to transfer the `livekit-plugins-sonex` PyPI package name to LiveKit once this is merged.

## Details

- Implements `tts.TTS` / `tts.ChunkedStream`. SonexLabs' `/v1/speech/stream` endpoint already streams synthesized *audio* out in chunks as it's generated (used here for low time-to-first-audio), but there's no endpoint that accepts *incremental/partial text* input over a persistent connection (the way Cartesia/ElevenLabs' websocket APIs do), so `TTSCapabilities.streaming=False` and only `synthesize()` is implemented — same shape as the `upliftai` plugin.
- Reuses the shared LiveKit Agents `aiohttp.ClientSession` via `utils.http_context` for connection pooling.
- Parses the WAV response directly to raw PCM (no `av`/`codecs` extra dependency required).
- Registered in root `pyproject.toml` (`tool.uv.sources`) and `uv.lock` regenerated.

## Testing

- Verified equivalent code against a live SonexLabs account in the standalone `sonexlabs/livekit-plugins-sonex` package (published to PyPI, in production use).
- `python -m py_compile` passes on all new files in this monorepo copy.
- CI green: ruff, type-check (3.10/3.13), unit-tests, blockguard-tests.


ayushi-agarwall:
Hi @livekit team, just following up on this PR. The SonexLabs TTS plugin implementation is ready and the CI checks are passing. I’ve also addressed the review feedback from the initial pass.

Would appreciate it if someone from the team could take another look when you get a chance. Happy to make any additional changes needed for it to be merged, and I can also transfer the `livekit-plugins-sonex` PyPI package to LiveKit as discussed in the related issue.

PR #6665, opened 2026-08-02
fix(voice): prevent dropped run_input acks in remote sessions

samanyugoyal2010:
Serialize SessionHost outbound events and responses through a single writer so concurrent transport writes and shutdown cancellation cannot drop a completed run_input acknowledgement after conversation events were already delivered (spurious "Agent did not respond within 60.0s").

Also harden RemoteSession pending-request handling (register-before-send, finally cleanup, late/unknown response warnings, clear close errors) and make room/TCP transports fail loudly on send errors.

addresses #6661

Issue #6661, opened 2026-08-01
lk agent simulate: spurious "Agent did not respond within 60.0s" on turns the transcript shows were answered (text-mode, Beta simulator)


Herzpraxis:
## Summary

Text-mode simulations (`lk agent simulate --scenarios …`) intermittently fail
with `Agent did not respond within 60.0s on turn N` even though the captured
transcript for the same run shows a complete, well-formed conversation —
including the agent's reply on the allegedly silent turn, the final booking
confirmation, and the farewell. The agent-side log shows no pending work at
that moment: the session closes with `reason: user_initiated`, and the session
report uploads normally.

## Environment

- livekit-agents 1.6.7 (Python 3.14), plugins: openai/azure 1.6.7
- LiveKit CLI 2.18.2, LiveKit Cloud project (eu)
- Agent architecture: single `Agent`, all tools are strict single-turn
  function tools (no AgentTasks, nothing outlives a turn); LLM is Azure
  OpenAI chat with `parallel_tool_calls=False`
- Simulations run in text mode (no STT/TTS/VAD)

## Observed behaviour

Across ~20 runs on two days (12-scenario and 15-scenario suites, run both
locally spawned and cloud-to-cloud):

- 3–11 of the jobs per run fail with `Agent did not respond within 60.0s on
  turn N`; N varies uniformly (turns 3–11) and is not correlated with any
  scenario, tool, or prompt. Worst observed: 11 of 15 jobs failed this way
  in a single run with `--concurrency 2` and zero provider rate-limit
  errors, while every one of the 15 transcripts was complete and every
  end-state assertion passed.
- For every such failure, the transcript in the same output contains the
  agent's response on turn N and the conversation continues to a clean end.
- The agent worker log for the affected room shows normal
  `conversation_item_added` events, then `session closed
  {"reason": "user_initiated"}` and a successful report upload — no error,
  no hung task, no retry.
- The same suites pass their end-state assertions (bookings/requests created
  via `on_simulation_end`) in every completed conversation.

## What we ruled out

- Long-running tools: our final architecture was rebuilt so no tool spans a
  turn, after reading `remote_session.py` — we understand the contract as
  "the simulator calls `interrupt(force=True)` before each user turn and
  waits for `session.run()` to complete". Our agent satisfies this.
- Model latency: the failures also occur on turns whose reply is visible in
  the transcript within seconds.
- Rate limits: runs without any 429s still show the same distribution.

## Expected

[...]

PR #6781, opened 2026-08-11
fix: stop dropping session host responses

theomonnom:
Alternative to [#6665](<https://github.com/livekit/agents/issues/6665>), addressing livekit/agents#6661. Diagnosis is @samanyugoyal2010's — commits are co-authored to them, happy to close in favour of theirs.

## What was happening

A `run_input` response that never reached the client looked exactly like a hung agent: the caller blocked for its full 60s timeout while the transcript was complete and the agent log showed a clean shutdown.

Three paths dropped one silently:

* `RoomSessionTransport.send_message` swallowed send failures into a warning and returned.
* The same method returned early when the room was gone.
* `SessionHost.aclose` cancelled in-flight handlers, discarding the response of a request that had already done its work.

Nothing serialized writes either, and every inbound message spawned a task to read it — which could finish out of order and deliver messages swapped.

## The fix

* Transports raise instead of dropping, and serialize sends with a lock.
* `aclose` drains handlers before cancelling, and logs the request types it gave up on.
* Reads and event writes each get one long-lived drainer instead of a task per message.

Responses stay direct — a handler awaits the transport and gets the failure — so no per-message future is needed to hand results back. Events are the only thing that needs a queue.

14 tests, each verified to fail without its change.

## Caveats

* Doesn't guarantee delivery: a mid-conversation send failure still ends in a caller timeout, now with a log line. Worth checking against livekit/agents#6661's 25–40% rate before closing it.
* Sequential reads trade concurrency for ordering; a bounded pool is the alternative if you'd rather keep per-stream isolation.

### DIFF
--- a/livekit-agents/livekit/agents/voice/remote_session.py
+++ b/livekit-agents/livekit/agents/voice/remote_session.py
@@ -132,6 +132,7 @@ class RoomSessionTransport(SessionTransport):
         with contextlib.suppress(utils.aio.ChanClosed):
             self._incoming_ch.send_nowait(reader)
 
+    @utils.log_exceptions(logger=logger)
     async def _read_loop(self) -> None:
         # sequential, so messages reach _recv_ch in the order they were sent
         async for reader in self._incoming_ch:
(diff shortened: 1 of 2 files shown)


davidzhao (approved):

samanyugoyal2010:
looks good to merge @theomonnom feel free to do so thanks for including me!

theomonnom:
@samanyugoyal2010 Thanks for reporting the issue!

PR #6782, opened 2026-08-11
fix(ivr): keep loop detection enabled without scikit-learn

### LINKED ISSUE #6778: IVR loop detection silently disabled: TfidfLoopDetector requires undeclared scikit-learn dependency
### Summary

`IVRActivity`'s loop detection is silently disabled out of the box because its only detector, `TfidfLoopDetector`, depends on `scikit-learn`, which is **not** a declared dependency of `livekit-agents` (neither a core requirement nor an extra). On a stock install, looping IVR menus are never detected — the code only falls back to the ~5s silence nudge.

Version: `livekit-agents==1.5.17`

### Details

`IVRActivity` wires `TfidfLoopDetector` as the sole loop detector:

```python
# livekit/agents/voice/ivr/ivr_activity.py
class IVRActivity:
    def __init__(self, session, *, max_silence_duration: float = 5.0) -> None:
        ...
        self._loop_detector = TfidfLoopDetector()
```

`TfidfLoopDetector.check_loop_detection()` imports sklearn lazily and, on `ImportError`, logs a warning and returns `False`:

```python
def check_loop_detection(self) -> bool:
    try:
        from sklearn.feature_extraction.text import TfidfVectorizer
        from sklearn.metrics.pairwise import cosine_similarity
    except ImportError:
        logger.warning(
            "TfidfLoopDetector: sklearn is not installed; loop detection is disabled. "
            "Please install the 'scikit-learn' package to enable loop detection."
        )
        return False
    ...
```

But `scikit-learn` appears nowhere in the package metadata:

```
$ grep -iE "sklearn|scikit|numpy" livekit_agents-1.5.17.dist-info/METADATA
Requires-Dist: numpy>=1.26.0
Requires-Dist: numpy>=1.26.0; extra == 'codecs'
```

So `numpy` (imported at the top of `ivr_activity.py`) is a hard dependency, but the sibling package the same module needs for its default behavior is not — there isn't even an extra like `livekit-agents[ivr]` to opt into it. The result is that anyone using IVR navigation gets loop detection silently no-op'd unless they happen to already have `scikit-learn` in their environment. The warning is only emitted at the first transcribed chunk, so it's easy to miss.

Introduced in #3656.

### Reproduce

1. `pip install livekit-agents==1.5.17` in a clean env (no `scikit-learn`).
2. Run an agent that enters IVR navigation (`AgentSession._start_ivr_detection` / `IVRActivity`).
3. Feed it a looping menu.
4. Observe: `check_loop_detection()` always returns `False`; the only warning is the sklearn one. No loop is ever detected.

### Expected

IVR loop detection should work on a default install, or the dependency should be explicit.

### Suggested fixes (either works)

- **Make it work out of the box:** reimplement the TF-IDF similarity using `numpy` only (already a hard dep). The logic is a small bag-of-words TF-IDF + cosine similarity over a ≤20-chunk window — no need for the full scikit-learn stack.
- **Or make the dependency explicit:** declare `scikit-learn` as an extra (e.g. `livekit-agents[ivr]`) and document it, so installs that use IVR fail loudly / opt in rather than silently degrading.

Happy to send a PR for the numpy-only detector if that direction is preferred.

uuzzrm:
## Summary

`TfidfLoopDetector` currently imports scikit-learn at runtime, but `scikit-learn` is only in the repository's development dependency group. On a normal `livekit-agents` install, IVR loop detection therefore logs a warning and always returns `False`.

This replaces the small TF-IDF and cosine-similarity calculation with NumPy, which is already a runtime dependency. The existing thresholds and detector state behavior are unchanged.

Fixes #6778

## Validation

- `uv run pytest tests/test_ivr_activity.py --unit -q`
- `uv run ruff check livekit-agents/livekit/agents/voice/ivr/ivr_activity.py tests/test_ivr_activity.py`
- `uv run ruff format --check livekit-agents/livekit/agents/voice/ivr/ivr_activity.py tests/test_ivr_activity.py`
- Compared the NumPy TF-IDF matrix and cosine similarities against scikit-learn across representative corpora.

The repository-wide unit suite passed 1,907 tests when run sequentially; 9 room tests could not start because `livekit-server` is not installed in the local environment. The repository type-check command also reports three pre-existing missing stubs for optional `cv2`, `loguru`, and `boto3` integrations; the changed files produce no type errors.

## AI assistance

AI assistance was used during implementation and review. I reviewed the resulting changes and test output and am responsible for the contribution.

### DIFF
--- a/livekit-agents/livekit/agents/voice/ivr/ivr_activity.py
+++ b/livekit-agents/livekit/agents/voice/ivr/ivr_activity.py
@@ -139,6 +139,8 @@ class TfidfLoopDetector:
         if len(self._transcribed_chunks) < 2:
             return False
 
+        # Rebuild the matrix for each chunk so the vocabulary and IDF reflect the
+        # current window. This is bounded by the small window size above.
         doc_matrix = self._tfidf_matrix(self._transcribed_chunks)
         last_chunk_similarity = doc_matrix[:-1] @ doc_matrix[-1]
 
--- a/livekit-agents/livekit/agents/voice/ivr/ivr_activity.py
+++ b/livekit-agents/livekit/agents/voice/ivr/ivr_activity.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from typing import TYPE_CHECKING
 
 import numpy as np
@@ -8,6 +9,8 @@ from livekit.agents import llm
 
 from ...log import logger
 from ...utils.aio.debounce import Debounced
+
+_TOKEN_PATTERN = re.compile(r"(?u)\b\w\w+\b")
 
 if TYPE_CHECKING:
     from ..agent_session import AgentSession
@@ -132,26 +135,12 @@ class TfidfLoopDetector:
             self._transcribed_chunks = self._transcribed_chunks[-self._window_size :]
 
     def check_loop_detection(self) -> bool:
-        try:
-            from sklearn.feature_extraction.text import TfidfVectorizer  # type: ignore
-            from sklearn.metrics.pairwise import cosine_similarity  # type: ignore
-        except ImportError:
-            logger.warning(
-                "TfidfLoopDetector: sklearn is not installed; loop detection is disabled. Please install the 'scikit-learn' package to enable loop detection."
-            )
-            return False
-
-        vectorizer = TfidfVectorizer()
-
         # Need at least two chunks to compute similarity against the last chunk
         if len(self._transcribed_chunks) < 2:
             return False
 
-        # NOTE: currently this is O(n^2) in the number of chunks, let's figure out a more efficient
-        # way if this become a bottleneck later.
-        doc_matrix = vectorizer.fit_transform(self._transcribed_chunks)
-        doc_similarity = cosine_similarity(doc_matrix)
-        last_chunk_similarity = doc_similarity[-1][:-1]
+        doc_matrix = self._tfidf_matrix(self._transcribed_chunks)
+        last_chunk_similarity = doc_matrix[:-1] @ doc_matrix[-1]
 
         if (
             last_chunk_similarity.size > 0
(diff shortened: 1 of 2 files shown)


uuzzrm:
The current `unit-tests` failure does not reach the IVR change. The run reports `1,917 passed, 5 skipped`, then fails during teardown of the unrelated `tests/test_plugin_google_realtime.py::test_output_streams_close_on_generation_complete` test because two `google.genai` close tasks (`AsyncClient.aclose` and `BaseApiClient.aclose`) remain pending. The log also shows a background OpenTelemetry exporter `TypeError` after the test session.

The changed IVR files are not in this failure path, and the focused IVR tests, Ruff, type checks, and other CI jobs are green. Could a maintainer rerun the unit job and/or track the Google Realtime fixture teardown separately?

uuzzrm:
After merging the current `main` base (`7477ec25`) into the branch, the unit job was rerun on head `c8f4b8fa` and produced the same unrelated teardown failure: `1,917 passed, 5 skipped`, with `tests/test_plugin_google_realtime.py::test_output_streams_close_on_generation_complete` leaking `google.genai` `AsyncClient.aclose` and `BaseApiClient.aclose` tasks. The IVR tests pass, and ruff/type-check/blockguard/release/CLA remain green. The job also logs a separate OpenTelemetry exporter `TypeError`. This still looks independent of the two-file IVR diff; could the Google Realtime teardown be tracked separately or the unit job be rerun once that environment issue is addressed?

tests/test_ivr_activity.py
```diff
@@ -51,6 +53,32 @@ def test_tfidf_detects_loop_on_repeated_user_speech() -> None:
...
+    """Detects loops on a default install without the optional scikit-learn package."""
+
+    real_import = builtins.__import__
+
+    def import_without_scikit_learn(name: str, *args: object, **kwargs: object) -> object:
+        if name == "sklearn" or name.startswith("sklearn."):
+            raise ImportError("scikit-learn is not installed")
+        return real_import(name, *args, **kwargs)
+
+    monkeypatch.setattr(builtins, "__import__", import_without_scikit_learn)
```
eashwar-mp, line 67: nit: since the implementation no longer imports sklearn at all, this monkeypatch never actually fires, the test passes the same without it. It still works as a guard against reintroducing the import, but maybe worth a one-line comment saying that's the intent, otherwise it reads as a near-duplicate of the test above.

livekit-agents/livekit/agents/voice/ivr/ivr_activity.py
```diff
@@ -132,26 +135,12 @@ def add_chunk(self, chunk: str) -> None:
...
         # Need at least two chunks to compute similarity against the last chunk
         if len(self._transcribed_chunks) < 2:
             return False
 
-        # NOTE: currently this is O(n^2) in the number of chunks, let's figure out a more efficient
-        # way if this become a bottleneck later.
-        doc_matrix = vectorizer.fit_transform(self._transcribed_chunks)
-        doc_similarity = cosine_similarity(doc_matrix)
-        last_chunk_similarity = doc_similarity[-1][:-1]
+        doc_matrix = self._tfidf_matrix(self._transcribed_chunks)
```
eashwar-mp, line 144: nit: the old code had a NOTE here that this recomputes similarity across all chunks each call. This version rebuilds the whole TF matrix from scratch on every chunk too, so that cost is still there. Fine at window=20, but wondering if it's worth keeping the note for whoever widens the window later.

eashwar-mp (approved):
Nice, this is a faithful numpy port. I checked the math against sklearn's defaults (token pattern, smooth_idf, l2 norm, cosine = dot of normalized rows) and it lines up, so the thresholds behave the same. Two small nits inline, neither blocking. Thanks for picking this up!

uuzzrm:
Addressed both non-blocking nits in e6b5ea8f: added a short comment explaining that the no-sklearn monkeypatch is an import-regression guard, and documented why the TF-IDF matrix is intentionally rebuilt for each bounded window so the vocabulary/IDF stay current. No behavior changes. git diff --check and Python compile checks pass locally; the full pytest collection is unavailable in this Windows checkout because the livekit runtime package is not installed. The existing unit-tests failure remains the unrelated Google Realtime teardown issue described above.

uuzzrm:
Thanks for the review. I addressed both non-blocking nits in commit `e6b5ea8f` on the current head:

- added a short comment explaining the no-sklearn monkeypatch as an import-regression guard;
- documented why the bounded-window TF-IDF matrix is intentionally rebuilt so its vocabulary and IDF stay current.

There are no behavior changes in this follow-up. The current IVR-focused checks, Ruff, type checks, blockguard, release gate, and CLA are green. The remaining unit-job failure is the unrelated Google Realtime teardown issue described earlier. Could you re-review the current head when convenient?

eashwar-mp (approved):
Looks good!

PR #6785, opened 2026-08-11
fix(voice): answer every tool call

### LINKED ISSUE #6569: Realtime (Gemini Live): speech marked played-out at generation completion when the model calls a tool mid-utterance — buffered audio dropped mid-word; tool response can be lost, deadlocking the session

longcw:
**Problem:** A tool call could end with no output at all — dropped when the user interrupted, or never produced when a tool raised `StopResponse` or returned an invalid value. A realtime model holds each call it emitted open until it is answered, so Gemini Live stopped responding and later `generate_reply()` calls produced no generation (livekit/agents#6569). A pipeline LLM re-issues the call instead and runs its side effects a second time.

**Fix:** Every call now gets exactly one output. An interrupted turn commits the results of tools that finished and delivers them to the realtime session; a handoff answers as an error, since the interruption left it unapplied.

**Behavior:** `FunctionCallOutput.reply_required` says when no reply is wanted, and each realtime plugin honours it as its provider allows — Gemini with `SILENT` scheduling, Ultravox with `agent_reaction="listens"`, Phonic by not opening a turn — or warns that it cannot. That replaces the blanket warning about models which generate tool replies on their own.

**Breaking:** `FunctionToolsExecutedEvent.function_call_outputs` is `list[FunctionCallOutput]`, no longer optional. A tool raising `StopResponse` now leaves a call and an empty output in history where it previously left nothing.

Replaces livekit/agents#6600. Fixes livekit/agents#6569


livekit-agents/livekit/agents/voice/events.py
```diff
@@ -282,17 +282,7 @@ def _make_update_pair(
...
-            fnc_call_out = FunctionCallOutput(
-                name=fnc_call.name,
-                call_id=fnc_call.call_id,
-                output=str(message or ""),
-                is_error=False,
-            )
-        else:
-            fnc_call_out = tool_output.fnc_call_out
-        return (fnc_call, fnc_call_out)
+        return (fnc_call, tool_output.fnc_call_out)
```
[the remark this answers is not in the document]
longcw: `_is_valid_function_output` return False only the value type is not supported and it should be an error.

livekit-agents/livekit/agents/voice/events.py
```diff
@@ -431,21 +421,19 @@ class FunctionToolsExecutedEvent(BaseModel):
...
+    the same ``call_id``. Every call has one output, even one whose tool raised
+    ``StopResponse``; such an output asks for no reply with ``reply_required``.
     """
 
     type: Literal["function_tools_executed"] = "function_tools_executed"
     function_calls: list[FunctionCall]
-    function_call_outputs: list[FunctionCallOutput | None]
+    function_call_outputs: list[FunctionCallOutput]
     created_at: float = Field(default_factory=time.time)
     _reply_required: bool = PrivateAttr(default=False)
```
chenghao-mou: should we replace this attribute with something like 

```python
def cancel_tool_reply(self) -> None:
    if not self.function_call_outputs:
        return
    for output in self.function_call_outputs:
        output.reply_required = False

@property
def has_tool_reply(self) -> bool:
    return any(output.reply_required for output in self.function_call_outputs)
```

livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -884,13 +884,12 @@ async def _traceable_fnc_tool(
...
                         output = make_tool_output(fnc_call=fnc_call, output=None, exception=e)
 
-                    if fnc_call_out := output.fnc_call_out:
-                        current_span.set_attribute(
-                            trace_types.ATTR_FUNCTION_TOOL_OUTPUT, fnc_call_out.output
-                        )
-                        current_span.set_attribute(
-                            trace_types.ATTR_FUNCTION_TOOL_IS_ERROR, fnc_call_out.is_error
-                        )
+                    current_span.set_attribute(
```
chenghao-mou, line 905: Q: is it possible we are storing stale output data if we update them in `_interrupted_tool_output` later?

chenghao-mou (approved):

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -4115,6 +4112,27 @@ def _create_assistant_message(
...
+            # commit results of tools that finished despite the interruption, as the pipeline
+            # task does. the calls are already recorded, so each one answers or the model waits
+            interrupted_fnc_outputs = [
+                _interrupted_tool_output(sanitized_out) for sanitized_out in tool_output.output
+            ]
+
+            if interrupted_fnc_outputs:
+                self._agent._chat_ctx.insert(interrupted_fnc_outputs)
+                self._session._tool_items_added(interrupted_fnc_outputs)
+
```
chenghao-mou, line 4146: this might need the same treatment in https://github.com/livekit/agents/pull/6823

livekit-agents/livekit/agents/voice/generation.py
longcw: The span attributes are snapshotted the moment the tool returns and the span closes there; `_interrupted_tool_output` rewrites the output later, only an interrupted handoff diverges where trace says success, history says "the agent handoff was interrupted". I think this is the honest trace as the tool did run and return an Agent.

PR #6792, opened 2026-08-11
fix(google): handle unsupported tool_choice values in the realtime plugin

### LINKED ISSUE #4770: tool_choice is not supported by the Google Realtime API.
### Bug Description

When generate_reply() is invoked inside a function tool while using Gemini Realtime, the agent fails to speak the generated message. The logs display the error: "tool_choice is not supported by the Google Realtime API." This occurs regardless of whether tool_choice is set to "auto" or "NOT_GIVEN".

### Expected Behavior

When generate_reply() is called within a function tool while using Gemini Realtime, the agent should successfully generate and speak the response message without logging any "tool_choice is not supported by the Google Realtime API." errors. The behavior should be consistent with normal message generation outside of tool execution.

### Reproduction Steps

```bash
1. Use gemini realtime as LLM
2. Define a function tool with generate_reply() inside it
3. Call the tool using agent
```

### Operating System

Linux

### Models Used

gemini-live-2.5-flash-native-audio

### Package Versions

```bash
Previous: 1.3.11 then updated to 1.4.1
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

tarikermis:
Closes #4770

The Google Realtime API does not support a `tool_choice` parameter. Session-level `none` is already emulated by rejecting tool calls, but two other paths were misleading:

- `generate_reply(tool_choice=...)` accepted the option and silently ignored it. It now warns, matching the existing warning for per-response `tools`.
- `update_options(tool_choice=...)` warned that unsupported values would fall back to `auto`, but still stored the unsupported value. It now normalizes the stored state to `auto` as well.

I added credential-free unit tests covering the warnings, the existing `none` emulation, and the invariant that no `tool_choice` value reaches `LiveConnectConfig`.

I used Kimi to help with the implementation and testing. I am responsible for the PR and happy to fix anything I missed.

Best Regards, Tarik

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -579,7 +579,6 @@ class RealtimeSession(llm.RealtimeSession):
         if is_given(tool_choice):
             # no per-response tool_choice on Gemini; "none" is emulated by rejecting any tool
             # call emitted during the turn (see _reject_tool_calls).
-            self._opts.tool_choice = tool_choice
             if tool_choice == "none":
                 logger.warning(
                     "the Google Realtime API has no tool_choice='none'; tool calls emitted "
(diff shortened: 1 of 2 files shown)


longcw:
unsupported tool_choice values is not the root cause, it's the gemini live doesn't support `generate_reply` in a tool call by default.

tarikermis:
Thanks for the context - that makes sense, generate_reply inside a tool call is the real gap then. I'll close my side here. Best Regards, Tarik

PR #6796, opened 2026-08-11
feat(hotel_receptionist): add say_goodbye_and_close_call

u9g:
Ports the `say_goodbye_and_close_call` tool out of #6567 so it can land on its own. That PR also rewrites the example onto a capability-loading architecture; none of that is here.

Today the hotel receptionist has no way to end a call — it says goodbye in prose and the line stays open. This adds the tool that owns the wind-down.

### The tool

The agent never says goodbye itself. `say_goodbye_and_close_call` returns the farewell instruction and then closes the line, and the instructions say so explicitly.

The hang-up is deferred rather than immediate. Callers routinely answer a farewell ("okay, bye!"), and a session torn down under that reply leaves their turn hanging — that showed up in simulations as 60s turn timeouts. So the tool does what a receptionist does: say goodbye, let the caller hang up first, and only close once the line stays quiet. Caller speech postpones the close by the full grace period rather than cancelling it, so a caller who keeps answering the farewell can't hold the line open for the rest of the call.

A repeat close (the caller answered the goodbye and the model called the tool again) uses a shorter quiet period and does *not* deliver a second farewell — without that guard it produced a "Goodbye!" / "Take care!" / "Goodbye!" loop.

### The pre-hangup policy audit

`end_call_check.py` is one LLM call at the moment of goodbye: it re-reads the receptionist's own standing policy against the transcript and names at most one concrete action the policy still requires. The verdict comes back as a forced `may_end`/`missing` tool call, so there is nothing to parse.

It is high-precision by construction and fails **open** on every failure path — no verdict, empty action, unparsable arguments, an error, a timeout, or a realtime model all let the call end. A caller stuck on a line that won't hang up is worse than a missed offer. It also fires at most once per call, so the agent can never get wedged unable to close.

### Verification

- Tool registers with a clean no-arg schema; the agent constructs and imports.
- `ruff format --check` and `ruff check` pass.
- mypy `--strict` output is byte-identical to `main`'s baseline for these files (5 pre-existing errors, none new).
- Ad-hoc checks over the five fail-open paths, the once-per-call nudge, system-prompt exclusion from the audit transcript, one-farewell-per-call, and the watchdog's fire/postpone/supersede semantics.

Example-only and Python-only — no library changes, and there's no equivalent example in `agents-js`.


examples/hotel_receptionist/end_call_check.py
```diff
@@ -0,0 +1,146 @@
...
+    function_tool(_missing, name="missing"),
+]
+
+
+def _verdict_from_tool_calls(tool_calls: list[llm.FunctionToolCall]) -> str | None:
+    """First may_end/missing call wins; anything unusable fails open (may end)."""
+    for call in tool_calls:
+        if call.name == "may_end":
+            return None
+        if call.name == "missing":
```
longcw: should we implement the tool body inside the function?

longcw (commented):
why not using the EndCall toolset, or add the features like find missing actions to it?

examples/hotel_receptionist/end_call_check.py
u9g: is this what you were suggesting? https://github.com/livekit/agents/commit/6acfccfcf

ShayneP (approved):

PR #6798, opened 2026-08-11
soniox: add reduce_silence option to TTS

razvanvoica:
## Summary

Soniox TTS v2 (`tts-rt-v2`) supports a `reduce_silence` parameter that shortens the pauses between words, independently of `speed` ([docs](https://soniox.com/docs/tts/concepts/reduce-silence)). This PR exposes it on `soniox.TTS`:

- new `reduce_silence: bool = False` constructor argument
- supported in `update_options()` as well
- the flag is only included in the WebSocket config message when enabled, so existing configurations produce byte-identical config messages

The server validates model support itself (`supports_silence_reduction` in the models listing) and returns `invalid_request` on models that lack it, so no client-side model gating is needed.

## Test plan

- `test_reduce_silence_propagates_to_options` covers the constructor default, `update_options`, and the constructor opt-in
- existing Soniox TTS tests pass unchanged
- verified manually against the Soniox EU REST endpoint (which shares the config schema per the API reference) with `tts-rt-v2`: `reduce_silence: true` accepted and audibly effective

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
@@ -88,6 +88,7 @@ class TTS(tts.TTS):
         sample_rate: int = DEFAULT_SAMPLE_RATE,
         bitrate: int | None = None,
         speed: float = DEFAULT_SPEED,
+        reduce_silence: bool = False,
         api_key: str | None = None,
         websocket_url: str = WEBSOCKET_URL,
         http_session: aiohttp.ClientSession | None = None,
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
```diff
@@ -88,6 +88,7 @@ def __init__(
         sample_rate: int = DEFAULT_SAMPLE_RATE,
```
chenghao-mou, line 101: > Only supported on models with supports_silence_reduction: true - see [List models](https://soniox.com/docs/api-reference/tts/get_tts_models). Setting it to true on any other model returns an [invalid_request](https://soniox.com/docs/api-reference/errors#invalid-request) error. See [Reduce silence](https://soniox.com/docs/tts/concepts/reduce-silence).

we should validate the model option here when it is True.

chenghao-mou (commented):
thanks for the PR! left one data validation comment.

livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
razvanvoica: Done in da9b084 - added a `_validate_reduce_silence()` check that raises `ValueError` from both the constructor and `update_options` (validated against the effective model/flag combination there, before any state is mutated). Implemented as a blocklist of the models that report `supports_silence_reduction: false` in the List models API (`tts-rt-v1`, `tts-rt-v1-preview`), so future models keep working without a plugin update and the server stays the final validator for unknown models. Covered by a new test, and updated the existing one which was enabling `reduce_silence` on the default v1-preview model.

razvanvoica:
Rebased on current main (no conflicts). The option is a single field in the stream config, validated against models that report supports_silence_reduction; tests included. We run this in production on tts-rt-v2 via a fork pinned to this patch and would love to drop the fork. Could a maintainer take a look?

PR #6799, opened 2026-08-11
fix(hotel_receptionist): pair near-miss rows in the expected-state diff

u9g:
## Summary

The expected-state grader reports a row the agent got *almost* right twice — once as `missing`, once as `unexpected` — each echoing every compared column. The one field that actually differs is buried in two near-identical dicts.

Greedily pair each missing row with its nearest unexpected row (fewest differing fields) and report only the changed fields. Unpaired rows still print as plain `missing` / `unexpected`.

```
- hotel_bookings: missing 1x {'guest_name': 'Ana Ruiz', 'nights': 2, 'room_type_view': 'king/ocean', ...}
- hotel_bookings: unexpected 1x {'guest_name': 'Ana Ruiz', 'nights': 3, 'room_type_view': 'king/ocean', ...}
+ hotel_bookings: row differs on nights: 2 != 3
```

Multiset differences are now expanded with `Counter.elements()` so pairing works per row; the `Nx` multiplicity prefix is gone and duplicates print one line each. Sorting uses a `repr` key because rows mix `None` with `str`/`int` in the same column, so tuples aren't directly orderable.

Extracted from #6567, which carries this alongside unrelated hotel-scenario work.

## Testing

- `ruff format --check` / `ruff check` clean; `mypy --strict` reports nothing in `benchmark.py`
- Exercised `diff_databases` against synthetic apsw DBs: single-field near miss pairs into one `row differs` line; a genuinely absent row still reports as `missing`; 2-expected/1-actual reports one pair plus one leftover; identical states and empty tables both return `[]`; `None`-vs-`str` columns sort without raising

### DIFF
--- a/examples/hotel_receptionist/benchmark.py
+++ b/examples/hotel_receptionist/benchmark.py
@@ -99,6 +99,37 @@ def _rows(
     return cols, counter
 
 
+def _changed_fields(
+    cols: list[str], expected_row: tuple[Any, ...], actual_row: tuple[Any, ...]
+) -> list[str]:
+    return [
+        f"{col}: {want!r} != {got!r}"
+        for col, want, got in zip(cols, expected_row, actual_row, strict=True)
+        if want != got
+    ]
+
+
+def _pair_rows(
+    cols: list[str], missing: list[tuple[Any, ...]], unexpected: list[tuple[Any, ...]]
+) -> list[tuple[tuple[Any, ...], tuple[Any, ...]]]:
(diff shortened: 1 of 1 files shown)


ShayneP:
I agree with Devin's comment, I think the original grader is probably still better? I’d prefer output that is verbose but more exact rather than output that looks better but is inferred.

If we want better readability, maybe we could add a nondestructive “possible near match” hint while keeping the original output.

PR #6800, opened 2026-08-11
fix(hotel_receptionist): stop the restaurant flow implying a table it never reserved

u9g:
A caller who won't give a phone number ended `BookRestaurantTask` with a bare `ToolError`, and `start_restaurant_booking` let it propagate. The receptionist read the failed tool call as a booked table and spoke a confirmation for a reservation that was never written.

Cherry-picked out of #6567, which bundles this with a lot of unrelated hotel-receptionist work.

## Changes

**Library** — `decline_phone_number_capture` completed the task with a bare `ToolError`, indistinguishable from any other failure, so a caller that wants to react to a refusal (rather than to a crash) had only the message text to match on. `PhoneNumberCaptureDeclinedError` subclasses `ToolError` with the identical message, so existing callers are unaffected.

**Example** — `RestaurantReservationNotCreatedError` is what the restaurant flow now completes with when the phone dialog is declined. `open_phone_dialog` catches the declined error and returns wording that states the outcome; `start_restaurant_booking` catches it and reports a non-reservation instead of falling through to the success string. The phone dialog also gets instructions to decline rather than stall when the caller has no number.

## Testing

- Full unit gate: 1942 passed, 5 skipped.
- ruff format + check clean; mypy clean on the changed library file.

### DIFF
--- a/examples/hotel_receptionist/book_restaurant.py
+++ b/examples/hotel_receptionist/book_restaurant.py
@@ -25,6 +25,16 @@ Each tool's return ends with a directive for the next action (e.g. "next: call o
 Never speak the same question twice in a row. If a field was just captured ("name recorded", "time recorded"), it is DONE - asking for it again stalls the call; the only valid next move is the directive in the last tool return.
 """
 
+_RESTAURANT_PHONE_INSTRUCTIONS = """\
+Caller cannot provide the phone number or does not have it handy: call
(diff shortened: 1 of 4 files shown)


u9g:
Duplicate of #6801, which carries the same change plus test coverage. Closing in favor of that one.

PR #6803, opened 2026-08-11
fix(hotel_receptionist): give card capture a no-usable-card exit

u9g:
## Summary

A caller whose only card is expired had no way out of `GetCardTask`. The expired-date `ToolError` told the agent to ask for another card, and `decline_card_capture` was documented as refusal-only, so the agent kept re-asking a caller who had already said they had nothing else to give.

Three changes route that dead end to the existing exit, kept to what the sub-task can actually observe:

- task instructions: if the caller refuses, or has no other card after theirs came back expired, don't keep asking - call `decline_card_capture` with the reason
- the expired-date `ToolError` first asks to confirm the date (a past date in voice is usually a misheard year), and names `decline_card_capture` only if the card really is expired and there's no other card
- `decline_card_capture`'s docstring widens from "the caller explicitly refuses" to also cover the expired-with-no-other-card case

The sub-task does not speak any reassurance about the booking: it is also opened for a new booking that isn't written yet, so "the booking stays held" would be false there. The caller-facing script for the card-update path stays where it already lives, in the receptionist's instructions.

Follow-up: #7177 (stacked) gives `BookRoomTask` a directive when the capture ends without a card, so the loop doesn't move up a level.

Cherry-picked from #6567 (2a5175745), `get_card.py` only, then narrowed as above.

## Testing

- `pytest examples/hotel_receptionist/ tests/test_tools.py --allow-uncategorized` (114 passed)
- `ruff format --check` and `ruff check` clean
- Scenario 15 ("Caller pushing about a declined card") simulated with this change plus #7177: [passed](https://cloud.livekit.io/projects/p_6hxckww6414/simulations/runs/SR_Jek87RtxnTMM). It reaches `GetCardTask` through `start_card_update` and the new wording did not misfire on "I don't have another card on me right now".

### DIFF
--- a/examples/hotel_receptionist/get_card.py
+++ b/examples/hotel_receptionist/get_card.py
@@ -22,6 +22,8 @@ Expect noisy voice transcription: digits read aloud ('four' -> 4, 'oh'/'zero' ->
 Never read the full card number or the security code back to the caller; refer to the card by its last four digits only. If a tool rejects a value, ask the caller to repeat just that detail - don't start the whole card over. If the caller switches cards mid-way, just record the new values; recording a field again replaces it.
 
 If the caller refuses to provide the card, call decline_card_capture.
+
+Caller has no usable card right now (their card is expired or keeps failing and they have no other card to give): don't keep asking for another card. Reassure them there's no pressure - the booking stays held - and call decline_card_capture with the reason; the receptionist arranges the retry from there.
 """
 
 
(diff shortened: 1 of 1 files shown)


ShayneP (approved):

u9g:
Pushed a second commit narrowing the wording after review: dropped "the booking stays held" (false on the new-booking path, where nothing is written until `confirm_booking`) and "keeps failing / won't validate" (the sub-task never authorizes a charge), and the expired-date error now asks to confirm the date before offering the exit. PR body updated. The BookRoom side of the loop is in #7177, stacked on this.

chenghao-mou (approved):

PR #6816, opened 2026-08-12
fix(silero): update bundled VAD checkpoint to v6.2

### LINKED ISSUE #6815: Bundled Silero VAD checkpoint misses short utterances compared with Silero 6.2.1
### Description

The Silero VAD model bundled with `livekit-plugins-silero` appears to use the original Silero v6 checkpoint. It performs significantly worse than the updated checkpoint in `silero-vad` 6.2.1 for short utterances in my testing.

### Environment

- `livekit-agents==1.5.1`
- `livekit-plugins-silero==1.5.1`
- Sample rate: 16 kHz
- Activation threshold: `0.5`
- Comparison model: `silero-vad==6.2.1`

I also checked `livekit-plugins-silero==1.6.9`, which appears to bundle the same checkpoint.

### Model verification

The bundled model in both LiveKit plugin versions:

```text
livekit-plugins-silero 1.5.1:
SHA256 597d30b3ec076608d059477bb14cfeffdf951bf5cae370d38f65d33bbfe82004

livekit-plugins-silero 1.6.9:
SHA256 597d30b3ec076608d059477bb14cfeffdf951bf5cae370d38f65d33bbfe82004
```

This matches the default `silero_vad.onnx` distributed with `silero-vad` 6.0.0 and 6.1.0.

The default model in `silero-vad` 6.2.1 is different:

```text
SHA256 1a153a22f4509e292a94e67d6f9b85e8deb25b4988682b7e174c65279d8788e3
```

### Reproduction results

| Utterance | Window | Bundled LiveKit model | Silero 6.2.1 | Old detected | 6.2.1 detected |
|---|---:|---:|---:|---|---|
| `yes` | 37.2–39.0 s | 0.213 | 0.970 | No | Yes |
| `hello` | 47.5–50.5 s | 0.440 | 0.983 | No | Yes |
| `再见` | 14.4–16.6 s | 0.444 | 0.999 | No | Yes |

Both models were tested using the same audio, sample rate, framing, context handling, recurrent-state handling, and activation threshold.

### Actual behavior

Clear short utterances can be missed completely by the bundled model because their speech probabilities do not cross the default activation threshold of `0.5`.

### Expected behavior

Clear short utterances should reliably trigger speech detection under the default configuration.

### Suggested resolution

Would you consider updating the bundled `silero_vad.onnx` checkpoint to the model distributed with `silero-vad` 6.2.1?

Alternatively, the plugin could:

1. Bundle both checkpoints and expose a model-version option.
2. Document how to use the newer checkpoint through `onnx_file_path`.
3. Add short-utterance regression tests before changing the default model.

### Expected Behavior

Clear short utterances should reliably trigger speech detection using the default activation threshold.

The Silero VAD model bundled with LiveKit should provide detection quality comparable to the current upstream Silero checkpoint. Ideally, the bundled model should be updated to Silero VAD 6.2.1, or the plugin should provide a documented option for selecting the newer checkpoint.

### Reproduction Steps

```bash
1. Install livekit-agents==1.5.1 and livekit-plugins-silero==1.5.1.
2. Load the bundled Silero VAD model with a 16 kHz sample rate and activation_threshold=0.5.
3. Run the the same short-utterance audio samples ("yes", "hello", and "再见") through the VAD.
4. Record the maximum speech probability and whether speech is detected.
5. Repeat the test using the default silero_vad.onnx model from silero-vad==6.2.1.
6. Use identical audio input, framing, context handling, recurrent-state initialization/reset, and threshold for both models.
7. Compare the results. The bundled LiveKit model does not detect these utterances, while the 6.2.1 model detects all of them.
```

### Operating System

Ubuntu 22.04

### Models Used

_No response_

### Package Versions

```bash
livekit-agents==1.5.1
livekit-plugins-silero==1.5.1
silero-vad==6.2.1
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

_No response_

longcw:
**Problem**

The plugin bundled the v6.1 Silero checkpoint, while `inference.VAD` already ran v6.2 through `livekit-local-inference`. Each ships the model separately, so the two drifted apart. v6.1 misses short utterances that v6.2 detects.

**Fix**

The bundled ONNX file now holds the v6.2 model: [https://github.com/snakers4/silero-vad/releases/tag/v6.2](<https://github.com/snakers4/silero-vad/releases/tag/v6.2>). The signature is unchanged, so no code changes. A new test asserts the plugin and `inference.VAD` agree, so they cannot drift again.

`VAD.load()` still accepts `onnx_file_path` to stay on v6.1.

Fixes livekit/agents#6815

### DIFF
--- a/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/resources/silero_vad.onnx
+++ b/livekit-plugins/livekit-plugins-silero/livekit/plugins/silero/resources/silero_vad.onnx
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:597d30b3ec076608d059477bb14cfeffdf951bf5cae370d38f65d33bbfe82004
+oid sha256:1a153a22f4509e292a94e67d6f9b85e8deb25b4988682b7e174c65279d8788e3
 size 2327524
--- a/tests/test_vad.py
+++ b/tests/test_vad.py
@@ -2,10 +2,13 @@ import asyncio
 from concurrent.futures import Future, ThreadPoolExecutor
 from typing import Any
 
+import numpy as np
 import pytest
 
 from livekit.agents import vad
 from livekit.agents.inference import VAD as InferenceVAD
+from livekit.local_inference import VAD as NativeVAD, VAD_WINDOW_SAMPLES
+from livekit.plugins.silero import onnx_model
 
 from . import utils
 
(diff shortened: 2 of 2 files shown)


chenghao-mou:
There are some differences between the inference and the silero one:

- `silero.VAD` applies an EMA with `alpha=0.35`: `p = 0.35 × previous + 0.65 × current` before emitting or thresholding it.
- `inference.VAD` uses `_native_vad.predict()` directly, with no Python-side EMA

so maybe we should not add the test now?

chenghao-mou (approved):
lgtm, one nit about the test

longcw:
> There are some differences between the inference and the silero one:
> 
> * `silero.VAD` applies an EMA with `alpha=0.35`: `p = 0.35 × previous + 0.65 × current` before emitting or thresholding it.
> * `inference.VAD` uses `_native_vad.predict()` directly, with no Python-side EMA
> 
> so maybe we should not add the test now?

the test compares the raw model output directly, EMA lives in `_VADStream._main_task` and is not included.

PR #6818, opened 2026-08-12
fix(openai): fail generate_reply fast on conversation_already_has_act…

### LINKED ISSUE #6817: RealtimeSession.generate_reply() races server-VAD responses → conversation_already_has_active_response; server error never correlated to the pending future (dies as 10s timeout)
### Bug Description

With the OpenAI Realtime model and server-side turn detection, two independent writers can start a response: (1) the server itself after VAD end-of-speech, and (2) the client via RealtimeSession.generate_reply(). Two distinct problems:

generate_reply() (realtime_model.py, ~L1604) sends response.create unconditionally — it never consults has_active_generation (defined ~25 lines below but unused on this path). If the server has just auto-created a response (client hasn't processed response.created yet), the server rejects with conversation_already_has_active_response. Window is widest right after a tool-reply generation starts or during a user turn ending.
When the server rejects, the error body carries event_id='response_create_*' — the exact client event id of the losing generate_reply. _handle_error (~L2211) ignores it: it only logs and emits realtime_model_error. The pending future in _response_created_futures is never failed, so the reply dangles until the 10s _on_timeout and dies as a generic "generate_reply timed out." The agent's intended speech is silently dropped.

error=APIError('OpenAI Realtime API returned an error', body=RealtimeError(message='Conversation already has an active response in progress: resp_EBlPemAbVfWVITtKcdTmQ. Wait until the response is finished before creating a new one.', type='invalid_request_error', code='conversation_already_has_active_response', event_id='response_create_c94357505d06', param=None), retryable=True) recoverable=True

### Expected Behavior

Client-initiated response.create should not race server-created responses — either wait for/cancel the active generation before sending, or on rejection fail the matching future immediately (correlating error.event_id) so callers can react, instead of a 10s timeout and a dropped reply.

### Reproduction Steps

```bash
1.AgentSession with openai.realtime.RealtimeModel (server turn detection on, create_response default).
2.Register a function_tool that takes ~15–20s and returns an error string.
3.During a call, ask something that triggers the tool; while it runs, say a short utterance ("no") right as the tool completes.
4.The tool-reply generation (server side) and the end-of-turn reply race → error above. Also reproducible by calling session.generate_reply() from any event handler (e.g. SIP DTMF) while the user is finishing a turn.
```

### Operating System

macOS

### Models Used

_No response_

### Package Versions

```bash
livekit-agents==1.6.7
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
Minimal: in _handle_error, when error.code == "conversation_already_has_active_response" and error.event_id matches a key in _response_created_futures, fail (or transparently retry after the active generation's _done_fut) that future immediately.
Fuller: have generate_reply() consult has_active_generation and defer/queue the response.create until the current generation completes (mirroring how _scheduling_task already serializes client-side speech).
```

### Additional Context

_No response_

_No response_

ByteMaster-1:
Fixes #6817

### Problem
OpenAI Realtime allows one active response per conversation. When a client
`response.create` races an already-active response (typically a server-VAD-created
one), the server rejects it with `conversation_already_has_active_response`. The
rejection arrives as an `error` frame whose `error.event_id` is the client event id —
**not** a `response.created`, and **no** `response.done` follows.

`generate_reply()` registers `_response_created_futures[event_id]` and arms a 10s
timeout. `_handle_error()` logged/emitted the error but never touched that future, so
it orphaned until the timeout fired — and then resolved as a generic
`"generate_reply timed out"`, indistinguishable from a real timeout. Downstream the
`SpeechHandle` stayed unset for ~10s and `SpeechHandle.exception()` raised
`InvalidStateError` until then.

### Fix (before → after)
- **Before:** collision → future orphaned ~10s → generic `"generate_reply timed out"`.
- **After:** collision → `_handle_error` correlates `error.event_id` to the pending
  future and fails it *immediately* with a typed `llm.RealtimeError` whose `.code` is
  `conversation_already_has_active_response`. `_realtime_reply_task` already
  `except llm.RealtimeError` and routes it onto the `SpeechHandle` via `_mark_done`.

**Both surfaces fire:** the exception (via `handle.exception()`) **and** the existing
recoverable `"error"` event.

### Scope (maintainer-aligned)
Detect → correlate → fail fast → surface typed + event. **No auto-retry and no queue** —
retry is client-side. Docstring on `AgentSession.generate_reply` shows the recipe
(`await handle` never raises; check `handle.exception()`; on the code, await
`current_speech.wait_for_playout()` then retry).

**Future work (not in this PR):** an opt-in SDK-side serialization of `response.create`
to avoid the collision entirely.

### Tests
- Positive: a fake `error` whose `event_id` matches a pending future fails it
  immediately with the typed error/code, and the error event still emits.
- Negative: an error with an unknown/absent `event_id` touches no future.
- `pytest tests/test_realtime/` (45 passed), `ruff`, and mypy all pass.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1468,10 +1468,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
 
         Note:
             ``await handle`` waits for the reply to finish and never raises; check
-            ``handle.exception()`` for the failure instead. With a realtime model, a reply
-            that races an already-active response fails fast with an ``llm.RealtimeError``
-            whose ``code`` is ``conversation_already_has_active_response``, so callers can
-            catch it and retry rather than waiting out a timeout.
+            ``handle.exception()`` for the failure instead.
         """  # noqa: E501
         if self._activity is None:
             raise RuntimeError("AgentSession isn't running")
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1467,23 +1467,11 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
             SpeechHandle: A handle to the generated reply.
 
         Note:
-            ``await handle`` waits for the reply to finish and never raises; inspect
+            ``await handle`` waits for the reply to finish and never raises; check
             ``handle.exception()`` for the failure instead. With a realtime model, a reply
-            that races an already-active response (server-VAD created) fails fast with an
-            ``llm.RealtimeError`` whose ``code`` is ``conversation_already_has_active_response``
-            rather than stalling until a timeout. The retry policy is yours to choose::
-
-                handle = session.generate_reply(user_input="...")
-                await handle
-                err = handle.exception()
-                if isinstance(err, llm.RealtimeError) and (
-                    err.code == "conversation_already_has_active_response"
-                ):
-                    # let the in-flight response play out, then retry
-                    if session.current_speech is not None:
-                        await session.current_speech.wait_for_playout()
-                    handle = session.generate_reply(user_input="...")
-                    await handle
+            that races an already-active response fails fast with an ``llm.RealtimeError``
+            whose ``code`` is ``conversation_already_has_active_response``, so callers can
+            catch it and retry rather than waiting out a timeout.
         """  # noqa: E501
         if self._activity is None:
             raise RuntimeError("AgentSession isn't running")
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1465,6 +1465,25 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
 
         Returns:
             SpeechHandle: A handle to the generated reply.
+
+        Note:
+            ``await handle`` waits for the reply to finish and never raises; inspect
+            ``handle.exception()`` for the failure instead. With a realtime model, a reply
+            that races an already-active response (server-VAD created) fails fast with an
+            ``llm.RealtimeError`` whose ``code`` is ``conversation_already_has_active_response``
+            rather than stalling until a timeout. The retry policy is yours to choose::
+
+                handle = session.generate_reply(user_input="...")
+                await handle
+                err = handle.exception()
+                if isinstance(err, llm.RealtimeError) and (
+                    err.code == "conversation_already_has_active_response"
+                ):
+                    # let the in-flight response play out, then retry
+                    if session.current_speech is not None:
+                        await session.current_speech.wait_for_playout()
+                    handle = session.generate_reply(user_input="...")
+                    await handle
         """  # noqa: E501
         if self._activity is None:
             raise RuntimeError("AgentSession isn't running")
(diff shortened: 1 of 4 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
```diff
@@ -2282,6 +2282,17 @@ def _handle_error(self, event: RealtimeErrorEvent) -> None:
             if not _is_fatal_error(event.error):
                 return
 
+        # a rejected response.create (e.g. the conversation already has an active response) draws
+        # an error, not a response.created, so nothing else settles the future generate_reply
+        # handed out. Fail it now with the provider code attached, instead of orphaning it until
+        # the 10s timeout turns it into a generic "generate_reply timed out". Fall through so the
+        # error still surfaces as an "error" event.
+        if (event_id := event.error.event_id) and (
```
longcw: maybe merge this and the above check under a single `if (event_id := event.error.event_id)` block and make the comments shorter.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1465,6 +1465,25 @@ def generate_reply(
...
+                await handle
+                err = handle.exception()
+                if isinstance(err, llm.RealtimeError) and (
+                    err.code == "conversation_already_has_active_response"
+                ):
+                    # let the in-flight response play out, then retry
+                    if session.current_speech is not None:
+                        await session.current_speech.wait_for_playout()
+                    handle = session.generate_reply(user_input="...")
+                    await handle
```
longcw: simplify the docstring? `handle.exception()` is worth to mention but maybe not the example here.

longcw (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
ByteMaster-1: Done — merged both waiter checks under a single if event_id := event.error.event_id:. Since the chat-ctx and response.create event-ids are separate namespaces (chat_ctx_* vs response_create_*) they can't collide, so the second check is now an elif, and the comments are trimmed to one line each. The response.create branch intentionally falls through instead of returning, so the error still hits the existing emit path (recoverable) or the fatal _is_fatal_error raise (terminal) — same reconnect-stopping behavior as before.

livekit-agents/livekit/agents/voice/agent_session.py
ByteMaster-1: Trimmed to a short Note: — kept the handle.exception() guidance and the conversation_already_has_active_response code so callers know what to branch on, and dropped the full retry example.

ByteMaster-1:
@longcw  can we please check this if possible ,
thanks.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1465,6 +1465,13 @@ def generate_reply(
 
         Returns:
             SpeechHandle: A handle to the generated reply.
+
+        Note:
+            ``await handle`` waits for the reply to finish and never raises; check
+            ``handle.exception()`` for the failure instead. With a realtime model, a reply
+            that races an already-active response fails fast with an ``llm.RealtimeError``
+            whose ``code`` is ``conversation_already_has_active_response``, so callers can
+            catch it and retry rather than waiting out a timeout.
```
longcw:
```suggestion
            ``await handle`` waits for the reply to finish and never raises; check
            ``handle.exception()`` for the failure instead.
```

longcw (approved):

PR #6826, opened 2026-08-12
fix: migrate http clients to httpx2

chenghao-mou:
**Behavior:** Upgrade OpenAI to 3, MCP to 2, and repository-owned HTTP transports to HTTPX2. Use one OpenAI client policy and test real MCP SSE and streamable HTTP requests.

**Compatibility:** Keep legacy `httpx.Timeout` inputs working through 1.x. Warnings point to application call sites. Remove support in 2.0. Keep HTTPX 1 adapters only for upstream SDKs that still require them.

Addresses AGT-3286

<details>

**Model:** GPT-5.6

> we need to add httpx because the new openai client dropped it and we did not declare it
>
> change of plan, can we upgrade to httpx2 instead?
>
> there is a migration guide: https://github.com/openai/openai-python/blob/main/httpx2.md as well, we should double check.
>
> we need a backward compatible change for the entire code base, not just inference... we need to deprecate the httpx version and warn the user that we are dropping them in 2.0.
>
> not just openai sdk, plain httpx usage should be updated too
>
> then we can stick with DefaultAsyncHttpx2Client
>
> i thought mcp can support httpx2 already?
>
> yeah, we should fix all three

</details>

### DIFF
--- a/livekit-agents/livekit/agents/cli/log.py
+++ b/livekit-agents/livekit/agents/cli/log.py
@@ -16,6 +16,8 @@ from ..plugin import Plugin
 NOISY_LOGGERS = [
     "httpx",
     "httpcore",
+    "httpx2",
+    "httpcore2",
     "openai",
     "watchfiles",
     "anthropic",
--- a/livekit-agents/livekit/agents/inference/llm.py
+++ b/livekit-agents/livekit/agents/inference/llm.py
@@ -4,7 +4,7 @@ import os
 from dataclasses import dataclass
 from typing import Any, Literal, cast
 
-import httpx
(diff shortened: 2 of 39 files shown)


longcw:
the plugins still depend on the httpx, should we add httpx to the agent instead?

chenghao-mou:
> the plugins still depend on the httpx, should we add httpx to the agent instead?

good point. Now it's upgrading the whole repo with deprecation warning and backward compatibility until 2.0.

chenghao-mou:
Closing this for now to avoid stale PRs and we can revisit this again later.

PR #6833, opened 2026-08-13
fix(telemetry): log a warning when the audio recording can't be read at upload time

### LINKED ISSUE #6832: telemetry: failed audio-recording read is silently swallowed during session-report upload (no log, audio part just missing)
When a session report is uploaded to the observability endpoint with `recording_options["audio"] = True`, a failure to read the local recording file is caught and discarded with no logging:

[https://github.com/livekit/agents/blob/bc5f3df/livekit-agents/livekit/agents/telemetry/traces.py#L683-L689](<https://github.com/livekit/agents/blob/bc5f3df/livekit-agents/livekit/agents/telemetry/traces.py#L683-L689>)

```python
audio_bytes = b""
if has_audio and report.audio_recording_path:
    try:
        async with aiofiles.open(report.audio_recording_path, "rb") as f:
            audio_bytes = await f.read()
    except Exception:
        audio_bytes = b""
```

`_build_multipart()` then attaches the `audio` part only `if audio_bytes:` (L709), so the upload proceeds without it and returns 200. The result: from the operator's side, "audio recording was never enabled" and "we tried to read the recording and failed" are indistinguishable. Any read failure at upload time (file deleted or cleaned up early, permission error, evicted ephemeral disk, truncated write) disappears without a trace -- literally, since this is the tracing path.

**Repro.** The script below builds a real `SessionReport` with `audio=True` and a nonexistent recording path, mocks only the OTel provider and the HTTP POST, and captures every log record plus the outgoing multipart body:

```
Upload completed without raising. HTTP POST was called: True
Multipart form parts sent to the observability endpoint:
  - form-data; name="header"; filename="header.binpb"
  - form-data; name="chat_history"; filename="chat_history.json"

audio part present in upload: False
Log records mentioning 'audio' or 'recording' during the whole call: 0
```

<details><summary>Full repro script (runs offline, no keys)</summary>

```python
"""
Proof: livekit-agents' _upload_session_report() in telemetry/traces.py silently
drops the audio recording from a Cloud observability upload when the local
recording file can't be read, with zero logging and zero indication to the
caller. File at:
  livekit-agents/livekit/agents/telemetry/traces.py:683-689

    audio_bytes = b""
    if has_audio and report.audio_recording_path:
        try:
            async with aiofiles.open(report.audio_recording_path, "rb") as f:
                audio_bytes = await f.read()
        except Exception:
            audio_bytes = b""          # <-- swallowed, no log

Later, _build_multipart() only attaches the "audio" part `if audio_bytes:`
(line 709), so a failed read is indistinguishable from "no audio requested".

Proof strategy: call the real _upload_session_report with
audio_recording_path pointing at a file that does not exist (simulating any
read failure: deleted-before-upload, permission error, evicted ephemeral
disk, etc — all hit the same bare `except Exception`). Capture:
  1) every log record emitted during the call (root logger + the module
     logger), to show NONE mentions the audio failure.
  2) the actual outgoing multipart POST body, to show the "audio" part is
     silently absent even though recording_options["audio"] = True.
The upload is mocked to return 200 so we isolate this behavior from network.
"""

import asyncio
import logging
import types
from pathlib import Path
from unittest.mock import AsyncMock, MagicMock, patch

from livekit.agents.observability import Tagger
from livekit.agents.telemetry import traces
from livekit.agents.voice.agent_session import RecordingOptions
from livekit.agents.voice.report import SessionReport


async def main():
    # 1. Build a minimal but real SessionReport with recording_options audio=True
    #    and a path that provably does not exist.
    missing_path = Path("/tmp/definitely-does-not-exist-livekit-recording-9f3a.ogg")
    assert not missing_path.exists()

    recording_options: RecordingOptions = {
        "audio": True,
        "traces": False,
        "logs": False,
        "transcript": True,
    }

    from livekit.agents.llm import ChatContext as RealChatContext

    options = types.SimpleNamespace(recording_options=recording_options)

    report = SessionReport(
        job_id="job-proof",
        room_id="room-proof",
        room="room-proof",
        options=options,
        events=[],
        chat_history=RealChatContext.empty(),
        audio_recording_path=missing_path,
        audio_recording_started_at=1000.0,
        timestamp=2000.0,
    )

    tagger = Tagger()

    # 2. Capture ALL log output during the call.
    captured_records: list[logging.LogRecord] = []

    class CaptureHandler(logging.Handler):
        def emit(self, record: logging.LogRecord) -> None:
            captured_records.append(record)

    handler = CaptureHandler()
    handler.setLevel(logging.DEBUG)
    logging.getLogger().addHandler(handler)
    logging.getLogger().setLevel(logging.DEBUG)

    # 3. Mock the OTel logger provider so _log() calls don't need a real
    #    OTel pipeline configured.
    fake_otel_logger = MagicMock()
    fake_logger_provider = MagicMock()
    fake_logger_provider.get_logger = MagicMock(return_value=fake_otel_logger)

    # 4. Mock the HTTP session/response for the multipart upload and
    #    CAPTURE the actual multipart body that gets sent.
    posted = {}

    fake_resp = AsyncMock()
    fake_resp.status = 200

    class FakePostCtx:
        def __init__(self, data, headers):
            posted["data"] = data
            posted["headers"] = headers

        async def __aenter__(self):
            return fake_resp

        async def __aexit__(self, *a):
            return False

    http_session = MagicMock()
    http_session.post = MagicMock(side_effect=lambda url, data, headers: FakePostCtx(data, headers))

    with patch("livekit.agents.telemetry.traces.get_logger_provider", return_value=fake_logger_provider):
        with patch("livekit.agents.telemetry.traces.api.AccessToken") as fake_token_cls:
            fake_token_cls.return_value.with_observability_grants.return_value.with_ttl.return_value.to_jwt.return_value = (
                "fake.jwt.token"
            )
            await traces._upload_session_report(
                agent_name="proof-agent",
                observability_url="https://example-observability.invalid",
                report=report,
                tagger=tagger,
                http_session=http_session,
            )

    logging.getLogger().removeHandler(handler)

    # ---- Assertions / findings ----
    print(f"Upload completed without raising. HTTP POST was called: {http_session.post.called}")

    mp = posted["data"]
    # aiohttp.MultipartWriter exposes its parts via ._parts (list of (Payload, headers, hlen))
    part_names = []
    for part_tuple in mp._parts:
        payload = part_tuple[0]
        cd = payload.headers.get("Content-Disposition", "")
        part_names.append(cd)
    print("Multipart form parts sent to the observability endpoint:")
    for p in part_names:
        print(f"  - {p}")

    audio_part_present = any('name="audio"' in p for p in part_names)
    print(f"\naudio part present in upload: {audio_part_present}")

    audio_related_logs = [
        r for r in captured_records
        if "audio" in r.getMessage().lower() or "recording" in r.getMessage().lower()
    ]
    print(f"Log records mentioning 'audio' or 'recording' during the whole call: {len(audio_related_logs)}")
    for r in audio_related_logs:
        print(f"  [{r.levelname}] {r.getMessage()}")

    assert not audio_part_present, "expected audio part to be silently missing"
    assert not audio_related_logs, "expected NO log record about the missing audio"
    print(
        "\nPROVEN: recording_options['audio']=True was requested, the recording "
        "file could not be read, and the upload silently proceeded WITHOUT the "
        "audio part and WITHOUT any warning/error log."
    )


asyncio.run(main())
```

</details>

**Possibly related:** livekit/agents#6264 reports audio missing from the dashboard with `record={"audio": True, ...}` and no error anywhere. I can't say this swallow is that issue's root cause (the reporter sees it consistently, which a nondeterministic read failure wouldn't fully explain), but as long as this bare `except` is here, anyone debugging livekit/agents#6264-shaped symptoms has no way to rule a failed read in or out from the logs.

**Proposed fix:** keep the graceful degradation (upload without audio rather than failing the whole report), but log a warning naming the path and the exception, in the style of the existing retry warning a few lines down (L740). PR incoming.

dubbl-a:
Fixes #6832.

A failed read of the local audio recording during session-report upload is currently swallowed bare, so the upload succeeds with the `audio` multipart part silently missing and nothing in the logs. This keeps the existing behavior (the report still uploads without audio rather than failing outright) and adds a warning naming the path and the exception, matching the style of the retry warning in the same function.

A runnable offline repro demonstrating the silent path is in #6832; re-running it against this branch shows the warning firing with the path and the underlying errno. Happy to add a unit test asserting the warning fires if you'd like one in-tree.

### DIFF
--- a/livekit-agents/livekit/agents/telemetry/traces.py
+++ b/livekit-agents/livekit/agents/telemetry/traces.py
@@ -689,9 +689,9 @@ async def _upload_session_report(
             audio_bytes = b""
             logger.warning(
                 "failed to read audio recording for session report upload, "
-                "uploading without the audio part (path=%s): %s",
+                "uploading without the audio part (path=%s)",
                 report.audio_recording_path,
-                e,
+                exc_info=e,
             )
 
     url = f"{observability_url}/observability/recordings/v0"
--- a/livekit-agents/livekit/agents/telemetry/traces.py
+++ b/livekit-agents/livekit/agents/telemetry/traces.py
@@ -685,8 +685,14 @@ async def _upload_session_report(
         try:
             async with aiofiles.open(report.audio_recording_path, "rb") as f:
                 audio_bytes = await f.read()
-        except Exception:
+        except Exception as e:
             audio_bytes = b""
+            logger.warning(
+                "failed to read audio recording for session report upload, "
+                "uploading without the audio part (path=%s): %s",
+                report.audio_recording_path,
+                e,
+            )
 
     url = f"{observability_url}/observability/recordings/v0"
 


livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -685,8 +685,14 @@ def _log(
...
             async with aiofiles.open(report.audio_recording_path, "rb") as f:
                 audio_bytes = await f.read()
-        except Exception:
+        except Exception as e:
             audio_bytes = b""
+            logger.warning(
+                "failed to read audio recording for session report upload, "
+                "uploading without the audio part (path=%s): %s",
+                report.audio_recording_path,
+                e,
```
longcw: can you use `exc_info=e` instead of `%s`?

longcw (approved):
something nit:

livekit-agents/livekit/agents/telemetry/traces.py
dubbl-a: Done, thanks. Switched to `exc_info=e` and kept the path in the message, so the traceback comes through now instead of just `str(e)`.

PR #6850, opened 2026-08-13
feat(ultravox): support server-side external TTS voices via externalVoice

ssgutierrez42:
## Summary

- Add `external_voice` parameter to `RealtimeModel` to pass an external TTS provider configuration (e.g. ElevenLabs, Cartesia, MiniMax) through Ultravox's `/calls` API via the `externalVoice` field
- When `external_voice` is provided, the `externalVoice` key is sent in the `/calls` payload instead of `voice`, keeping the model in voice mode so native barge-in is preserved
- The parameter accepts `dict[str, Any]` to accommodate all provider variants without coupling to Ultravox's evolving schema

## Why

Ultravox's `output_medium="text"` path (text-out half-cascade) cannot support barge-in — the plugin's interruption signal (`playback_clear_buffer`) only fires in voice mode, and Ultravox exposes no input-side speech-onset event. Server-side `externalVoice` solves this by having Ultravox call the TTS provider directly, keeping the session in voice mode with working interruption.

## Changes

- `_UltravoxOptions`: added `external_voice: NotGivenOr[dict[str, Any]]`
- `RealtimeModel.__init__`: added `external_voice` keyword argument (defaults to `NOT_GIVEN`), documented as mutually exclusive with `voice`
- `/calls` payload construction: `voice` is now conditional — when `external_voice` is given, `externalVoice` is sent instead

## References

- Ultravox `/calls` API: https://docs.ultravox.ai/api-reference/calls/calls-post
image


livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -139,6 +141,8 @@ def __init__(
             The Ultravox model to use.
         voice : str | UltravoxVoice
             The voice to use for TTS.
+        external_voice : dict[str, Any], optional
+            The Ultravox configuration for an external TTS provider. Mutually exclusive with ``voice``.
```
longcw, line 145: should we log a warning when both `voice` and `external_voice` are given?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
```diff
@@ -634,6 +638,11 @@ async def _main_task(self) -> None:
                     "selectedTools": parse_tools(list(self._tools.function_tools.values())),
                 }
 
+                if is_given(self._realtime_model._opts.external_voice):
```
longcw, line 653: should we skip a given but empty `external_voice` or raise an error, or fallback to `voice`?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
longcw: `external_voice` cannot work with `output_medium="text"` too?

longcw (approved):
looks good to me. something nit:

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
ssgutierrez42: @longcw 
good catch !
voice always has a default value ("Mark"), so we may need a bigger change to distinguish “explicitly set” from “default.” Do you think we should log a warning when `voice != DEFAULT_VOICE and is_given(external_voice)` ?

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
ssgutierrez42: will update to raise an error 📝 

I realize that silently falling back to `voice` may mask the bug (caller thinks they’re using an external voice but hears the default “Mark” instead), and sending it through would produce a confusing 400 from Ultravox.

livekit-plugins/livekit-plugins-ultravox/livekit/plugins/ultravox/realtime/realtime_model.py
ssgutierrez42: > `external_voice` cannot work with `output_medium="text"` too?

@longcw not as documented currently - because `externalVoice` tells UV to route its audio output through an external TTS provider, but in `output_medium="text"` the model emits text tokens only (no audio for `external_voice` to synthesize)

ssgutierrez42:
@longcw ready for re-review

longcw (approved):

longcw:
looks good to me! could you sign the CLA before we merge?

ssgutierrez42:
@longcw CLA ready

PR #6851, opened 2026-08-13
fix(voice): run on_user_turn_completed for text input

u9g:
## The bug

`Agent.on_user_turn_completed` only ever ran on the audio end-of-turn path. Text
input went through `generate_reply()`, which never reaches it:

```python
async def _default_text_input_cb(sess, ev):
    async with sess._claim_user_turn():
        await sess.interrupt()
        sess.generate_reply(user_input=ev.text)   # hook never runs
```

So a turn the user *typed* skipped the hook entirely, while the same turn
*spoken* ran it. Two consequences:

- The docs point to `on_user_turn_completed` as the seam for editing or dropping
  a user message before it reaches the LLM. A guard wired there — PII redaction,
  a safety check, `StopResponse` — was silently bypassed whenever the user typed
  instead of spoke.
- Text-only simulation runs never fired it at all, so any agent holding state in
  that hook behaves differently under `lk agent simulate` than in production.
  A tool-gated flow of ours stalled in all 16 simulated calls for exactly this
  reason: a gate cleared in `on_user_turn_completed` never opened, and every
  downstream tool call was refused.

Demonstrated by driving one agent two ways — identical agent, identical message,
only the modality differs:

| turn driven through | `on_user_turn_completed` |
|---|---|
| audio pipeline | fires |
| `_default_text_input_cb` | never fires (LLM still replies) |

## The fix

Route the default text input callback through the same hook before generating,
so `StopResponse` and chat-context edits hold for a typed turn the way they do
for a spoken one.

`generate_reply()` is synchronous and can't await an async hook inline, so the
text path gets a small async seam on `AgentSession` rather than a change to
`generate_reply` itself.

Two commits: a pure refactor extracting the hook invocation out of
`_user_turn_completed_task`, then the functional change reusing it.

## Scope

Only the **default** callback routes through the hook. A custom `text_input_cb`
calling `sess.generate_reply(user_input=...)` still bypasses it, and
`generate_reply` is deliberately left alone — it's a generic "make the agent
reply" API, and `on_user_turn_exceeded`'s own default calls it, so firing the
hook there would be wrong. If custom callbacks should be covered too, the seam
wants to be public API instead of `_`-prefixed; happy to do that here.

## Tests

`tests/test_user_turn_hook_modality.py` pairs each text case against a spoken
control: the hook firing, chat-context edits reaching the LLM, and
`StopResponse` dropping the turn. Against `main` the three text cases fail and
the spoken control passes; with this change all four pass.

Full unit gate: 1949 passed, 5 skipped.

## Cross-SDK parity

`agents-js` has the identical gap at `agents/src/voice/room_io/room_io.ts:49`:

```ts
export const DEFAULT_TEXT_INPUT_CALLBACK: TextInputCallback = (sess, ev) => {
  sess.interrupt();
  sess.generateReply({ userInput: ev.text });
};
```

Not fixed here. It's a larger change there — the callback is synchronous and
doesn't await `interrupt()`, so awaiting the hook means making
`TextInputCallback` async. Happy to open the matching PR.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2399,6 +2399,32 @@ class AgentActivity(RecognitionHooks):
         )
         return True
 
+    async def _run_user_turn_hook(
+        self, user_message: llm.ChatMessage
+    ) -> tuple[llm.ChatContext, float] | None:
+        """Run ``on_user_turn_completed`` for a completed user turn.
+
+        Returns the chat context the hook may have edited, along with how long it
+        took, or None when the turn is to be dropped - the hook raised
+        StopResponse, or raised at all.
+        """
+        # create a temporary mutable chat context to pass to on_user_turn_completed
+        # the user can edit it for the current generation, but changes will not be kept inside the
+        # Agent.chat_ctx
+        temp_mutable_chat_ctx = self._agent.chat_ctx.copy()
+        start_time = time.perf_counter()
+        try:
+            await self._agent.on_user_turn_completed(
+                temp_mutable_chat_ctx, new_message=user_message
+            )
+        except StopResponse:
+            return None
+        except Exception:
+            logger.exception("error occurred during on_user_turn_completed")
+            return None
+
+        return temp_mutable_chat_ctx, time.perf_counter() - start_time
+
     @utils.log_exceptions(logger=logger)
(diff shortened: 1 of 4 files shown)


longcw:
this is expected, user should use a custom `text_input_cb` instead of `on_user_turn_completed` to customize the process for text input.

PR #6854, opened 2026-08-14
feat(room_io): mix audio from every participant into one AgentSession

### LINKED ISSUE #6795: Support multiple participants in a single AgentSession with a shared chat context
# Feature Request: Support multiple participants in a single `AgentSession`

## Problem

The current SDK assumes a single active participant per `AgentSession`. While the existing workarounds are sufficient for transcription, they don't support a single-session, multi-participant conversation with a shared chat context.

## Use case

We're building **AI-conducted interviews with human takeover**.

1. An AI agent interviews a candidate in a 1:1 conversation.
2. A human interviewer joins the room and takes over questioning.
3. The agent switches to a silent "observer" mode—it stops speaking but continues listening.
4. While silent, it must continue hearing **both** the interviewer and the candidate, appending both to the **same chat context**.
5. When the interviewer leaves or hands control back, the agent resumes the interview with the complete conversation history.

The requirement is therefore:

> **N participants → 1 AgentSession → 1 shared chat context → 1 LLM/TTS**, with the agent's speaking ability toggleable independently of its listening.

## Why the current approaches don't work

* **`RoomIO.set_participant()` / `toggle_io.py`**

  * Only one participant can be attached at a time.
  * Participants that are not currently attached are not included in the conversation history, resulting in a one-sided context.

* **Multiple `AgentSession`s**

  * This creates one session per participant, which also means one chat context per participant.
  * Our use case requires a **single shared context** and a **single LLM/TTS pipeline**.

## SDK limitation (agents 1.6.7)

From our investigation, the current implementation appears to make two single-participant assumptions:

* `_participant_available_fut` latches onto the first participant (`room_io.py:378-396`).
* `_ParticipantInputStream` maintains a single `_stream` / `_publication` (`_input.py:202-204`), so a second participant's track replaces the existing one instead of being added alongside it.

## Expected behavior

It should be possible for a single `AgentSession` to receive audio from multiple participants while maintaining a single conversation context. Each completed user turn should also include the speaker identity so the application can attribute transcripts correctly while preserving one shared chat history.

This would enable use cases such as:

* AI interviews with human takeover
* Meeting copilots
* Customer support escalations
* Collaborative conversations where the agent needs to listen to multiple participants while speaking selectively

Darshak-Jankat:
Closes #6795

## Problem

`AgentSession` can only listen to one participant at a time: `_ParticipantInputStream` keeps a single `_stream`/`_publication`, so a second participant's track replaces the first instead of joining it. Use cases like an AI interview with human takeover need N participants → 1 session → 1 shared chat context.

## Approach

Opt-in mixing. With `AudioInputOptions.mix_participants=True`, `RoomIO` subscribes to the microphone of every accepted participant and mixes them with `rtc.AudioMixer` (the same primitive `BackgroundAudio` uses) into the single audio input the session already consumes. One STT/LLM/TTS pipeline, one chat context, no API surface beyond the flag:

```python
session.start(
    room=ctx.room,
    room_options=RoomOptions(audio_input=AudioInputOptions(mix_participants=True)),
)
```

## Changes

- **`voice/room_io/_input.py`** — `_ParticipantAudioInputStream` keeps a `_MixedSource` per participant (channel + track stream + forward task) and feeds each channel into an `rtc.AudioMixer`; the mixed output becomes the session's input. The only change to the shared single-participant path is a `_sink(participant)` hook in the base class, which still returns `_data_ch` — video and non-mixed audio behave exactly as before.
- **`voice/room_io/room_io.py`** — every accepted participant is added to the mix on connect and removed on disconnect. Kind filtering is unchanged, and the agent's own avatar worker (`ATTRIBUTE_PUBLISH_ON_BEHALF`) is excluded so the agent can't hear itself. The linked participant is still the first one and only drives the outputs (audio, transcription, chat text).
- **`voice/room_io/types.py`** — the `mix_participants` option.

Notes:
- AGC runs once on the mixed output rather than per stream (one shared `AudioProcessingModule` interleaved across speakers would be wrong).
- A noise-cancellation *selector* now builds a processor per participant, owned by that participant's stream; a directly-passed `FrameProcessor` instance keeps today's shared lifetime.
- `set_participant()` is a no-op for the input in this mode — listening covers everyone; toggle it with the existing audio-input enable/disable.

## Not included

Per-turn speaker attribution, also mentioned in the issue. Mixing collapses everyone into one stream, so a single STT cannot label who spoke — that needs per-participant recognition, which is a separate change.

## Tests

`tests/test_room_io.py`:
- `test_mix_participants_sums_every_participant_audio` — two participants' frames come out of the input summed into one frame.
- `test_mix_participants_tracks_room_membership` — RoomIO mixes both humans, skips its own avatar worker, links only the first, and drops a participant on disconnect.

`uv run pytest --unit` passes (the pre-existing `test_sent_tokenizer` failure and `test_room.py` errors are environment-related and reproduce on `main`).

https://claude.ai/code/session_01XQ72uVTX79ocrm3KJ3LKYh

### DIFF
--- a/livekit-agents/livekit/agents/voice/room_io/_input.py
+++ b/livekit-agents/livekit/agents/voice/room_io/_input.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 from abc import ABC, abstractmethod
 from collections.abc import AsyncIterator, Iterable
+from dataclasses import dataclass
 from typing import Any, Generic, TypeVar, cast
 
 from typing_extensions import override
@@ -148,15 +150,20 @@ class _ParticipantInputStream(Generic[T], ABC):
             "source": rtc.TrackSource.Name(publication.source),
         }
         logger.debug("start reading stream", extra=extra)
+        sink = self._sink(participant)
         async for event in stream:
             if not self._attached:
                 # drop frames if the stream is detached
                 continue
             frame = cast(T, event.frame)
             self._process_frame(frame)
-            await self._data_ch.send(frame)
+            await sink.send(frame)
 
         logger.debug("stream closed", extra=extra)
+
+    def _sink(self, participant: rtc.RemoteParticipant) -> aio.Chan[T]:
+        """The channel a participant's frames are forwarded to."""
+        return self._data_ch
 
     def _process_frame(self, frame: T) -> None:
         """Hook for subclasses to process frames in-place before forwarding."""
@@ -227,6 +234,16 @@ class _ParticipantInputStream(Generic[T], ABC):
                 return
 
 
+@dataclass
+class _MixedSource:
+    """One participant's contribution to the mixed audio input."""
+
+    chan: aio.Chan[rtc.AudioFrame]
+    stream: rtc.AudioStream | None = None
+    task: asyncio.Task[None] | None = None
+    publication_sid: str | None = None
+
+
 class _ParticipantAudioInputStream(_ParticipantInputStream[rtc.AudioFrame], AudioInput):
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


Darshak-Jankat:
Superseded by #6855, reopened from the correct account. Same commit.

PR #6857, opened 2026-08-14
tests: stop room tests hanging on macOS

longcw:
**Problem**

`tests/test_room.py` is marked `concurrent`, so its nine tests open every room at the same time. When the file descriptor soft limit is large, the native layer on macOS never completes `Room.connect` and `pytest --unit` hangs forever. Nothing bounds `Room.connect`, and per-item timeouts do not arm inside a concurrent group, so the run stops with no test name.

**Fix**

`pytest_configure` now caps the file descriptor soft limit at 8192 on macOS, and only ever lowers it, so a machine already below that value is unaffected. `Room.connect` now runs under `asyncio.wait_for`, so a wedged connection fails in 15 seconds with a test name. Linux is untouched, so CI behavior does not change.

The 8192 value is measured on one macOS machine: 8192 passes on three trials, 12288 and above deadlock, and 1024 and below fail on descriptor exhaustion.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,7 @@ import asyncio
 import dataclasses
 import logging
 import re
+import sys
 from collections.abc import Iterator
 from functools import cache
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


chenghao-mou (approved):
lgtm

PR #6862, opened 2026-08-14
feat(agents): serialize the inline AgentTasks awaited from one turn

u9g:
Inline `AgentTask`s awaited from one turn's parallel tool calls now queue for a slot on the activity they pause and take it in turn, instead of overwriting each other's handoffs and leaving the session wedged until close times out; scoping the slot to the activity is what keeps a nested task, which pauses a different one, from waiting on its parent. Supersedes #6860.


theomonnom (approved):
lgtm!

cc @longcw for a second review

longcw (approved):

PR #6864, opened 2026-08-14
fix(sarvam): align TTS validation with the Bulbul API

### LINKED ISSUE #6774: sarvam: TTS parameter validation diverges from the live Bulbul API

uuzzrm:
Fixes #6774

The sarvam plugin's TTS option validation drifted from what the Bulbul
endpoints actually accept, so three combos that should work were being
rejected up front and three that can't work were only failing once the
request hit the API.

**Speakers.** bulbul:v3 has 37 documented voices. Nine were missing from
the compatibility map (anand, tarun, sunny, mani, gokul, vijay, mohit,
rehan, soham), while amelia/sophia were listed even though the API
rejects them with "Speaker is not recognized". Verified against
https://docs.sarvam.ai/api/getting-started/models/bulbul and the REST
reference.

**Sample rates.** 32/44.1/48 kHz are REST-only (bulbul:v3). `stream()`
now raises up front for those instead of erroring mid-session over the
socket, and `synthesize()` rejects them for non-v3 models.

**pace.** Now validated per model in both the constructor and
`update_options`: 0.5-2.0 for v3/v3-beta, 0.3-3.0 for v2.

Added `tests/test_tts_validation.py` (unit category) covering the speaker
map, pace ranges, and the streaming/REST sample-rate gates. `uv run
pytest --unit livekit-plugins/livekit-plugins-sarvam` passes and
`ruff check` / `ruff format` are clean on the changed files.

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
@@ -99,6 +99,14 @@ _CODEC_TO_MIME: dict[str, str] = {
 
 _TELEPHONY_CODECS: frozenset[str] = frozenset({"mulaw", "alaw"})
 
+# Sample rates accepted by the Sarvam streaming (WebSocket) API. The higher
+# rates (32/44.1/48 kHz) exist for the REST API only (bulbul:v3), so a stream
+# built with one of them would only fail once a connection is open.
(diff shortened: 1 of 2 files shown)


uuzzrm:
Addressed the stale-pace finding in 94829cc: update_options now re-validates the stored pace against the new model's range when only the model changes, clamping with a warning (same policy as pitch) instead of letting an out-of-range value reach the API. Added regression coverage for the clamp and for the in-range case.

Issue #5158, opened 2026-03-19
Intermittent truncation of initial audio on PCM path (AudioByteStream ->   AudioSource.capture_frame())


satruin:
### Bug Description

When using a custom TTS plugin that outputs raw PCM/WAV audio through the PCM path (`AudioByteStream -> AudioSource.capture_frame()`), 
the first few tens to hundreds of milliseconds of audio are intermittently truncated on the receiving end.

In our tests, this was observed only on the PCM path. 
We did not observe the same issue when using the encoded audio path (for example, MP3 via `AudioStreamDecoder`).

The issue is intermittent and occurs roughly 1 in 3-5 attempts under the same conditions.

What we verified:
- The raw audio bytes from the TTS provider are intact when dumped and played locally.
- Waveform analysis shows the initial samples are valid and do not contain abnormal spikes.
- We added debug logging in the Python SDK path and verified that every frame reaches `AudioSource.capture_frame()` successfully.
- We were not able to find any frame loss before `capture_frame()`.

In our tests, this appears to happen somewhere below that handoff point.

### Expected Behavior

All audio frames passed to `AudioSource.capture_frame()` should be delivered to the remote participant without truncation, 
including the beginning of speech.

### Reproduction Steps

```bash
1. Generate short utterances using a custom TTS plugin that returns WAV/PCM audio at 24kHz mono.
2. Feed the PCM bytes into the agents PCM path (`AudioByteStream`) and forward every emitted frame to `AudioSource.capture_frame()`.
3. Publish the audio track to a LiveKit room and receive it in a browser client.
4. Repeat playback of the same utterance multiple times under the same conditions.
5. Intermittently, the beginning of speech is truncated on the receiving side;
approximately the first 50-300ms is lost.

We can provide a minimal reproducer or captured PCM sample if needed.
```

### Operating System

Linux (production), macOS (development)

### Package Versions

```bash
livekit-agents==1.4.2
livekit==1.1.2
Python 3.13
```

### Proposed Solution

```python
No concrete fix proposed yet. 
We are mainly reporting the issue and sharing the investigation results. 
If there is a recommended configuration or known workaround for the PCM path, we would like to try it.
```

### Additional Context

Symptoms:
- A brief "pop" sound at the start of speech
- The first syllable or word is partially or fully cut off
- The same audio may play correctly on one attempt and be truncated on another

Workarounds attempted:
- Insert delay (30/50/100ms) before the first frame: no improvement
[...]

PR #6876, opened 2026-08-16
fix(utils): keep whole samples when flushing a partial trailing sample

### LINKED ISSUE #6874: AudioByteStream.flush() discards all buffered audio when the buffer ends on a partial sample
### Bug Description

`AudioByteStream.flush()` throws away every buffered sample when the buffer ends on a partial sample, instead of returning the complete samples and dropping only the incomplete tail.

The guard in `livekit-agents/livekit/agents/utils/audio.py` is:

```python
if len(self._buf) % self._bytes_per_sample != 0:
    logger.warning("AudioByteStream: incomplete frame during flush, dropping")
    return []
```

So a single stray byte discards the whole buffer, which can be just under a full frame of audio. On an 8 kHz mono 16-bit stream with a 200 ms target frame, one odd byte costs up to 200 ms of speech. The warning says "incomplete frame", but what gets dropped is all the complete samples too.

The buffer is also left untouched on that path. `self._buf.clear()` only runs on the success path, so the misaligned bytes stay in the stream and every later push keeps the same one byte offset. Once the byte count realigns, the frames that come out are shifted by a byte, which reads as noise rather than as missing audio.

This is reachable whenever a source can hand over an odd number of bytes, which for us is a telephony PCM feed at 8 kHz. It also interacts with #5158, where truncation on the PCM path is being tracked from the other end.

### Expected Behavior

`flush()` should emit the complete samples it holds and drop only the trailing partial sample, and it should clear the buffer either way so that a partial sample cannot shift the alignment of everything that follows. The warning should say how many bytes were discarded.

### Reproduction Steps

```bash
1. Buffer 1500 whole samples plus one stray byte, then flush:

s = AudioByteStream(sample_rate=8000, num_channels=1, samples_per_channel=1600)
s.push(b"\x11\x22" * 1500 + b"\x33")   # 1500 samples + 1 partial byte
frames = s.flush()

print("frames:", len(frames))
print("samples returned:", sum(f.samples_per_channel for f in frames))
print("bytes still buffered:", len(s._buf))

2. Observed:

frames: 0
samples returned: 0
bytes still buffered: 3001

1500 samples, 188 ms of audio, are gone, and the buffer still holds them in a
state that will misalign later frames.

3. Expected: one frame of 1500 samples, and an empty buffer.
```

### Operating System

macOS 27.0 (arm64), also reproduced on Linux

### Models Used

Not model specific, this is in the PCM path before any provider

### Package Versions

```bash
livekit-agents==1.6.10
livekit==1.1.14
Python 3.13.7
```

Also on main at 06c71f9.

rahul2002m:
Fixes #6874.

`AudioByteStream.flush()` bailed out with an empty list whenever the buffer did not end on a sample boundary, so one stray byte took every complete sample with it. On an 8 kHz mono stream with a 200 ms target frame that is up to 200 ms of speech, and the warning called it an incomplete frame while what it actually dropped was the whole buffer.

The buffer was also left in place on that path, since only the success path called `clear()`. So the misaligned byte stayed in the stream and offset every later frame by a byte, which comes out as noise rather than as missing audio. That second part is the one that worried me more, because it is silent.

The fix drops only the trailing bytes that cannot form a sample, returns the samples before them, and leaves the buffer empty on both paths. The warning now says how many bytes went.

I ran into this on a telephony feed at 8 kHz where the source can hand over an odd number of bytes. It may also be worth a look next to #5158, which tracks truncation on the PCM path from the other end, though I have not confirmed they are the same thing.

Tests are in a new `tests/test_utils_audio.py`, since `AudioByteStream` did not have a module of its own. Ten cases: whole samples surviving a partial tail, only the tail being dropped, the buffer ending up empty, the warning naming the byte count, a lone partial byte, an empty buffer staying quiet, flush after whole frames were emitted, idempotence, and the stereo case where a sample is 4 bytes. Eight of the ten fail on main; the other two are guards for behaviour this does not change.

`uv run pytest --unit` gives 1566 passed against 1556 on main, which is the ten new ones and no regressions. `make check` is clean, formatting, ruff and mypy. There are 9 errors in the unit run on my machine from the OTLP metrics exporter at teardown, but they are identical on a clean checkout of main so they are not from this change.

### DIFF
--- a/livekit-agents/livekit/agents/utils/audio.py
+++ b/livekit-agents/livekit/agents/utils/audio.py
@@ -179,10 +179,12 @@ class AudioByteStream:
             list[rtc.AudioFrame]: A list containing any remaining `AudioFrame` objects.
 
         This method processes any remaining data in the buffer that does not
-        fill a complete frame. If the remaining data forms a partial frame
-        (i.e., its size is not a multiple of the expected sample size), a warning is
-        logged and an empty list is returned. Otherwise, it returns the final
-        `AudioFrame` containing the remaining data.
+        fill a complete frame, and returns it as the final `AudioFrame`.
+
+        If the buffer does not end on a sample boundary, the trailing incomplete
+        sample cannot be represented and is dropped with a warning; the complete
+        samples before it are still returned. The buffer is always left empty, so
+        a partial sample can never offset the samples of a later push.
 
         Use this method when you have no more data to push and want to ensure
         that all buffered audio data has been processed.
@@ -190,8 +192,16 @@ class AudioByteStream:
         if len(self._buf) == 0:
             return []
 
-        if len(self._buf) % self._bytes_per_sample != 0:
-            logger.warning("AudioByteStream: incomplete frame during flush, dropping")
+        remainder = len(self._buf) % self._bytes_per_sample
+        if remainder:
+            logger.warning(
+                "AudioByteStream: buffer does not end on a sample boundary during "
+                "flush, dropping the trailing %d byte(s)",
+                remainder,
+            )
+            del self._buf[len(self._buf) - remainder :]
+
+        if len(self._buf) == 0:
             return []
 
         frames = [
--- a/tests/test_utils_audio.py
+++ b/tests/test_utils_audio.py
@@ -1,0 +1,103 @@
+import logging
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


longcw:
closing because of https://github.com/livekit/agents/issues/6874#issuecomment-5312809737

PR #6499, opened 2026-07-21
fix(voice): deadline-based user_away_timeout ignores transcript-less noise (#6030)

dorukdumlu:
## Summary
- Track an absolute `_user_away_deadline` for `user_away_timeout` so transcript-less user `speaking ↔ listening` flips re-arm with *remaining* time (telephony noise can no longer defer "away" indefinitely) — #6030.
- Add `user_away_on`: `"presence"` (default, backcompat) vs `"conversation"` (ignore user VAD speaking; only pause while the agent is active / refresh on final transcript). Prefer `conversation` for noisy SIP + `turn_detection="stt"`.
- Drop redundant away-recovery `_set_user_away_timer(reset=True)` that could arm the timer while the agent is speaking (Devin review).

## Test plan
- [x] `test_noise_user_state_flips_do_not_defer_away_deadline`
- [x] `test_final_transcript_and_agent_activity_refresh_away_deadline`
- [x] `test_away_recovery_transcript_does_not_arm_during_agent_speech`
- [x] `test_conversation_user_away_ignores_user_speaking`
- [x] `test_conversation_user_away_still_cancels_on_agent_speaking`
- [x] Existing `test_final_transcript_resets_away_timer_when_not_speaking`

PR #6880, opened 2026-08-17
feat(voice): add user_away_signal for transcript-based away tracking

### LINKED ISSUE #6030: user_away_timeout getting triggered for SIP participant on telephony calls even when user is not speaking and taking background noise and conversations as user state speaking.

longcw:
**Problem**

The `away` user state follows detected speech. On a noisy phone line, VAD and Deepgram `SpeechStarted` report speech that never becomes text, and each report re-arms the full `user_away_timeout`. Noise also ends `away` as soon as it starts, which stops the "are you still there?" flows that the state exists for.

**Fix**

This change adds `user_away_signal`. The default value `"audio"` keeps the current behavior, so a session that does not set the option is unchanged. The value `"transcript"` trusts only transcribed speech: activity that produces no text does not hold off `away`, and only a transcript ends `away`.

The `"transcript"` value reads interim transcripts, so a long answer does not trip `away` before its final arrives. This needs a streaming STT, because `StreamAdapter` emits no interim transcripts.

Fixes livekit/agents#6030. Supersedes livekit/agents#6499, and thanks to @dorukdumlu, whose analysis in that issue found the root cause.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -235,6 +235,8 @@ DEFAULT_EXPRESSIVE_OPTIONS: ExpressiveOptions = ExpressiveOptions(
     speech_steering=DEFAULT_SPEECH_STEERING_OPTIONS,
 )
 
+UserAwaySignal = Literal["audio", "transcript"]
+
 
 def _append_instructions(template: Instructions | str, extra: str) -> Instructions:
     # concatenate the *raw* template text so any {placeholders} survive until render()
@@ -285,6 +287,7 @@ class AgentSessionOptions:
     """sparse endpointing keys the user provided explicitly"""
     max_tool_steps: int
     user_away_timeout: float | None
+    user_away_signal: UserAwaySignal
     transcription_timeout: float | None
     min_consecutive_speech_delay: float
     use_tts_aligned_transcript: bool | None
@@ -389,6 +392,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
         aec_warmup_duration: NotGivenOr[float | None] = NOT_GIVEN,
         ivr_detection: bool = False,
         user_away_timeout: float | None = 15.0,
+        user_away_signal: UserAwaySignal = "audio",
         transcription_timeout: float | None = None,
         session_close_transcript_timeout: float = 2.0,
         # Runtime settings
@@ -478,6 +482,10 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
             user_away_timeout (float, optional): If set, set the user state as
                 "away" after this amount of time after user and agent are silent.
                 Defaults to ``15.0`` s, set to ``None`` to disable.
+            user_away_signal (Literal["audio", "transcript"], optional): Which user
+                activity holds off the "away" state. ``"audio"`` (default) trusts any
+                detected speech. ``"transcript"`` trusts only transcribed speech, so
+                noise that never becomes text is ignored; it needs a streaming STT.
             transcription_timeout (float, optional): If set, emit a
                 ``user_transcription_timeout`` event when VAD detects user speech
                 during the user's turn but no non-empty final transcript arrives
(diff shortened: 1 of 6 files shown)


davidzhao (commented):
this lg, but I think we should not use an additional toggle, but automatically count any presence of user transcript as activity. if we can make dealing with VAD signals more resilient, then we could simplify the amount of toggles a user has to figure out and tune.

longcw:
> automatically count any presence of user transcript as activity

this has been added, the problem is that the noise causes the VAD to false alarm and therefore never recognizes if the user is `away` or not. perhaps the new behavior should be the default, only a user turn with transcriptions can be delay the `away` state.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -478,6 +482,10 @@ def __init__(
             user_away_timeout (float, optional): If set, set the user state as
                 "away" after this amount of time after user and agent are silent.
                 Defaults to ``15.0`` s, set to ``None`` to disable.
+            user_away_signal (Literal["audio", "transcript"], optional): Which user
+                activity holds off the "away" state. ``"audio"`` (default) trusts any
+                detected speech. ``"transcript"`` trusts only transcribed speech, so
+                noise that never becomes text is ignored; it needs a streaming STT.
```
chenghao-mou, line 488: We should mention the audio signal can come from either vad or STT.

chenghao-mou, line 488: This would also suppress the timer for some STTs that send ghost transcripts. Observed with 11labs Scribe v2 from https://github.com/livekit/agents/pull/4043

dorukdumlu: Same failure on Deepgram SIP, not only Scribe v2. Noise often yields non-empty interims (uh, the, punctuation). If those re-arm the full away window, a drip every few seconds never reaches away.

Separate hole on the same path: if away swallows VAD start and end, an interim promotes to speaking and a later final does not demote. The user stays speaking until the next timeout.

Issue #6881, opened 2026-08-17
Flaky unit-tests: a genai client finalizer lands in an unrelated test's leak check


helin-mktech:
`tests/test_plugin_openai_stt_context.py::test_use_realtime_defaults_to_the_only_transport_a_model_has[gpt-realtime-whisper-True]` intermittently fails with:

```
Failed: Test leaked tasks:
Coroutine : BaseApiClient.aclose    google/genai/_api_client.py:2213
Coroutine : AsyncClient.aclose      google/genai/client.py:152
```

That test is synchronous and never touches google. The tasks come from `tests/test_plugin_google_realtime.py` — the only module in the unit suite that constructs genai clients.

**Cause.** `AsyncClient.__del__` and `BaseApiClient.__del__` both do this unconditionally, with no check for a client that was already closed:

```python
# google/genai/client.py:186, _api_client.py:2248
def __del__(self) -> None:
    try:
      asyncio.get_running_loop().create_task(self.aclose())
    except Exception:
      pass
```

`RealtimeSession.aclose()` closes `client.aio` (`realtime_api.py:898`), but that does not disarm the finalizer — the object stays garbage until the collector reaches it, which can be several modules later. `fail_on_leaked_tasks` diffs `asyncio.all_tasks()` around each test, so the task is charged to whatever test happens to be running then.

**Why it surfaces now.** It reproduces only on the CPython build CI uses — 3.12.13 fails every run, 3.12.3 passes every run, on the same machine with the same locked dependencies. Which test gets blamed depends on GC timing, so anything that shifts allocation can move it. It currently lands on #6340, which adds a test module and nothing else related.

**Fix.** Settle the finalizers inside the test that owns them, so they cannot escape into a later module.

helin-mktech:
Fix: #6882

PR #6882, opened 2026-08-17
fix(tests): settle genai client finalizers in the test that owns them

helin-mktech:
Fixes the flaky `Test leaked tasks` failure described in #6881.

`AsyncClient.__del__` schedules `aclose()` on whatever event loop is running when the collector reaches it, even for a client the test already closed. Collected late, those tasks surface as leaked tasks in an unrelated test in a later module — usually `tests/test_plugin_openai_stt_context.py`, which never touches google.

This adds an autouse fixture to the one module that builds genai clients, so the finalizers run and finish while that test still owns the loop.

Verified on the CPython build CI uses (3.12.13), full unit suite:

| | `unit-tests` |
| --- | --- |
| before | fails, every run |
| after | passes, every run |

Reverting the fixture brings the failure straight back.

### DIFF
--- a/tests/test_plugin_google_realtime.py
+++ b/tests/test_plugin_google_realtime.py
@@ -5,6 +5,7 @@ import gc
 import logging
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
+from typing import Any
 
 import pytest
 from google.genai import types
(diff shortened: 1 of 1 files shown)


tests/test_plugin_google_realtime.py
```diff
@@ -14,6 +16,30 @@
...
+    unrelated test in a later module.
+    """
+    yield
+    # gc.collect() runs the finalizers synchronously, so anything they schedule
+    # is already registered by the time it returns.
+    gc.collect()
+    if pending := [
+        task
+        for task in asyncio.all_tasks()
+        if not task.done() and "aclose" in (getattr(task.get_coro(), "__qualname__", "") or "")
```
longcw: maybe also check if it's from genai
```python
and (getattr(task.get_coro(), "__module__", "") or "").startswith("google.genai")
```

longcw (approved):

tests/test_plugin_google_realtime.py
helin-mktech: Good catch — aclose on its own is too loose. Done, keyed on the coroutine's defining module now.

One correction on the mechanics: coroutine objects don't carry __module__ at all, so getattr(coro, "__module__", "") is always '' and that check would never match — the fixture would have gone quietly inert.
coro.cr_frame.f_globals["__name__"] gives the module.

helin-mktech:
Hi @longcw , I was wondering if this fix could be merged so that I can rebase the other PR based on this.

PR #6891, opened 2026-08-18
fix(voice): await _commit_user_turn on session close so trailing STT transcript commits (#6889)

### LINKED ISSUE #6889: AgentSession close doesn't wait for trailing STT transcript to commit (unawaited _commit_user_turn is cancelled by activity.aclose())
### Bug Description

On session close triggered by the remote SIP party hanging up, `AgentSession` intends to flush any in-flight STT recognition so the caller's final utterance is captured before teardown — but the flush is cancelled before it can run, so the transcript for that last utterance is silently lost.

We hit this in production on an outbound telephony flow: the agent was mid-way through IVR navigation when it reached the target line's voicemail system, and the voicemail system responded "the mailbox is full and cannot accept new messages at this time" and then hung up (SIP BYE) immediately after finishing that sentence. VAD (`user_speaking`) on our side shows ~13 continuous seconds of speech matching that message, ending less than a second before the remote participant disconnects. Every earlier utterance in the same call session finalized into a `user_turn` transcript within ~1s of speech ending — but this final one never does. From the app's perspective, transcript shows the call simply ending, even though the recording captures the entire speech that was played out.

We traced this to `AgentSession._aclose_impl()`:

https://github.com/livekit/agents/blob/02569a40794645195bd92003431e5197ea413922/livekit-agents/livekit/agents/voice/agent_session.py#L1242-L1251

```python
if (
    reason != CloseReason.ERROR
    and (audio_recognition := activity._audio_recognition) is not None
):
    # wait for the user transcript to be committed
    audio_recognition._commit_user_turn(
        audio_detached=True,
        transcript_timeout=self._opts.session_close_transcript_timeout,
    )

await activity.aclose()
```

The comment says "wait for the user transcript to be committed," but `_commit_user_turn(...)` is never awaited — it returns an `asyncio.Future` (https://github.com/livekit/agents/blob/02569a40794645195bd92003431e5197ea413922/livekit-agents/livekit/agents/voice/audio_recognition.py#L1038-L1044) and only starts a background task:

https://github.com/livekit/agents/blob/02569a40794645195bd92003431e5197ea413922/livekit-agents/livekit/agents/voice/audio_recognition.py#L1120-L1128

```python
if self._commit_user_turn_atask is not None:
    self._commit_user_turn_atask.cancel()

self._commit_user_turn_atask = asyncio.create_task(_commit_user_turn())
self._commit_user_turn_atask.add_done_callback(_on_task_done)
return fut
```

That task is meant to inject up to 2s of silence to flush the STT buffer and then wait (up to `session_close_transcript_timeout`, default 2.0s — see https://github.com/livekit/agents/blob/02569a40794645195bd92003431e5197ea413922/livekit-agents/livekit/agents/voice/agent_session.py#L392 and #L494-L496) for the STT provider to emit a final transcript for the trailing utterance.

But the very next line in `_aclose_impl`, `await activity.aclose()`, walks straight into `AgentActivity._close_session()`:

https://github.com/livekit/agents/blob/02569a40794645195bd92003431e5197ea413922/livekit-agents/livekit/agents/voice/agent_activity.py#L1321

```python
await self._audio_recognition._aclose()
```

which immediately cancels the very task that was just started:

https://github.com/livekit/agents/blob/02569a40794645195bd92003431e5197ea413922/livekit-agents/livekit/agents/voice/audio_recognition.py#L793-L796

```python
async def _aclose(self) -> None:
    self._closing.set()
    if self._commit_user_turn_atask is not None:
        await aio.cancel_and_wait(self._commit_user_turn_atask)
```

So the flush task is cancelled within microseconds of being created — nowhere near its own `transcript_timeout` budget, let alone the STT provider's own end-of-turn timeout (e.g. Deepgram Flux's `eot_timeout_ms`, which we have configured at 3000ms). In our production traces the entire `_aclose_impl` → `activity.aclose()` sequence completes in under 2ms.

### Expected Behavior

On a clean (non-error) close, `_aclose_impl` should actually wait for the `_commit_user_turn(...)` future it creates — up to `transcript_timeout` — before calling `activity.aclose()` (which cancels that same task). That way, an utterance that's still being recognized when the session starts closing (e.g. because the remote SIP party hung up right after finishing speaking) gets a chance to finalize and be committed as a transcript/conversation item before teardown, instead of being silently discarded.

### Reproduction Steps

1. Start an `AgentSession` on a SIP call using a streaming STT (we use Deepgram Flux, `turn_detection="stt"`).
2. Have the remote party speak a multi-second utterance and then immediately disconnect (SIP BYE) right as they finish speaking — e.g., an IVR/voicemail system that plays a message and hangs up as soon as playback ends, with no pause between the end of audio and the hangup.
3. Observe: `participant_disconnected` fires, the app's disconnect handler calls `session.aclose()` (directly or via `JobContext.shutdown()` → the job runner's `session.aclose()`), which invokes `_aclose_impl()`.
4. Check `stt.SpeechEvent`/transcript output, or an APM trace of `_commit_user_turn_atask` / `_aclose_impl`: no `FINAL_TRANSCRIPT`/turn-committed event is produced for the trailing utterance, and the `_commit_user_turn_atask` is cancelled almost immediately after creation rather than being awaited.

We reproduced this via distributed tracing (Sentry) rather than a minimal script — happy to help build a minimal repro against a mock STT that delays its final result by ~1s past `audio_detached=True`, since that's exactly the timing window this bug loses.

### Operating System

Linux (Cloud Run container, production)

### Models Used

STT: Deepgram Flux (`deepgram.STTv2`, `turn_detection="stt"`, `eot_timeout_ms=3000`). LLM/TTS not relevant to this bug — it's purely in the STT turn-commit-on-close path.

### Package Versions

```
livekit==1.1.14
livekit-agents==1.6.9
livekit-api==1.2.0
livekit-protocol==1.1.22
livekit-plugins-deepgram==1.6.9
```

Also confirmed still present in `main` as of writing (same unawaited-future / immediate-cancel pattern at the current HEAD of `agent_session.py`/`audio_recognition.py`).

### Proposed Solution

```python
# agent_session.py, inside AgentSession._aclose_impl()
if (
    reason != CloseReason.ERROR
    and (audio_recognition := activity._audio_recognition) is not None
):
    # wait for the user transcript to be committed
    await audio_recognition._commit_user_turn(
        audio_detached=True,
        transcript_timeout=self._opts.session_close_transcript_timeout,
    )

await activity.aclose()
```

i.e. simply `await` the future that `_commit_user_turn(...)` already returns, so `activity.aclose()` (and its `_audio_recognition._aclose()` → `cancel_and_wait(self._commit_user_turn_atask)`) can't race ahead of it. `_commit_user_turn`'s own internals already bound this by `transcript_timeout`/`asyncio.wait_for`, so this shouldn't introduce an unbounded hang — worst case it adds up to `session_close_transcript_timeout` (default 2.0s) to session close on this code path, which seems like the intended behavior given the existing comment and timeout option.

### Additional Context

This came out of a production investigation (Spara — outbound sales/support voice agent). We first thought this was a Deepgram Flux limitation (audio genuinely never reaching the STT before hangup), but tracing showed the audio track was live and VAD was actively detecting speech for the full duration of the utterance, right up until the participant disconnected — so the audio was available, it's the transcript-commit-on-close path that never gets a chance to run.

Happy to open a PR with the one-line `await` fix above if that's the right direction — wanted to file the issue first in case there's context we're missing about why it's currently fire-and-forget (e.g. a prior incident with `_aclose_impl` hanging that this was a deliberate workaround for).

jonahlipsky:
## Summary

Fixes livekit/agents#6889.

On a clean (non-error) `AgentSession` close, `_aclose_impl()` intends to flush any in-flight STT recognition so the caller's final utterance is committed as a transcript before teardown. The comment says "wait for the user transcript to be committed," but `_commit_user_turn(...)` was called **without** `await` — it returns an `asyncio.Future` and only starts a background task. The very next line, `await activity.aclose()`, runs `AgentActivity._close_session()` → `_audio_recognition._aclose()`, which immediately `cancel_and_wait`s that same task. So the flush is cancelled within microseconds of being created, nowhere near its `transcript_timeout` budget, and the trailing utterance's transcript is silently lost.

This bites in production on telephony flows where the remote SIP party hangs up (SIP BYE) immediately after finishing speaking (e.g. an IVR/voicemail system that plays a message and disconnects with no trailing pause).

## Change

`await` the future that `_commit_user_turn(...)` already returns, so `activity.aclose()` can't race ahead of it:

```python
# wait for the user transcript to be committed
await audio_recognition._commit_user_turn(
    audio_detached=True,
    transcript_timeout=self._opts.session_close_transcript_timeout,
)
```

`_commit_user_turn` already bounds itself with `transcript_timeout` / `asyncio.wait_for`, so this doesn't introduce an unbounded hang — worst case it adds up to `session_close_transcript_timeout` (default 2.0s) to session close on this path, which matches the intent of the existing comment and timeout option.

## Test plan

- [ ] Utterance still being recognized when the session starts closing (remote party hangs up right as they finish speaking) is now finalized and committed as a transcript/conversation item instead of being discarded.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1249,7 +1249,7 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
                     and (audio_recognition := activity._audio_recognition) is not None
                 ):
                     # wait for the user transcript to be committed
-                    audio_recognition._commit_user_turn(
+                    await audio_recognition._commit_user_turn(
                         audio_detached=True,
                         transcript_timeout=self._opts.session_close_transcript_timeout,
                     )
--- a/livekit-agents/livekit/agents/voice/agent_session.py
+++ b/livekit-agents/livekit/agents/voice/agent_session.py
@@ -1248,11 +1248,22 @@ class AgentSession(rtc.EventEmitter[EventTypes], Generic[Userdata_T]):
                     reason != CloseReason.ERROR
                     and (audio_recognition := activity._audio_recognition) is not None
                 ):
-                    # wait for the user transcript to be committed
-                    await audio_recognition._commit_user_turn(
+                    # Best-effort: wait for the trailing user transcript to be
+                    # committed before teardown. This must never abort the rest of
+                    # the close, so isolate the flush's own failure or cancellation
+                    # (e.g. a racing commit_user_turn() cancels the in-flight flush
+                    # task) from the close path. asyncio.wait() still propagates
+                    # cancellation of the close task itself.
+                    commit_fut = audio_recognition._commit_user_turn(
                         audio_detached=True,
                         transcript_timeout=self._opts.session_close_transcript_timeout,
                     )
+                    await asyncio.wait([commit_fut])
+                    if not commit_fut.cancelled() and (exc := commit_fut.exception()):
+                        logger.warning(
+                            "error while committing the final user turn on close",
+                            exc_info=exc,
+                        )
 
                 await activity.aclose()
             self._activity = None
--- a/tests/test_session_close_commit_user_turn.py
+++ b/tests/test_session_close_commit_user_turn.py
@@ -1,0 +1,153 @@
+"""Regression tests for AgentSession close waiting on the trailing STT transcript.
+
+On a clean (non-error) close, ``AgentSession._aclose_impl`` flushes any in-flight
+STT recognition via ``audio_recognition._commit_user_turn(...)`` so the caller's
+final utterance is committed before teardown. ``_commit_user_turn`` only *starts*
+a background task and returns a future; the very next step, ``activity.aclose()``,
+runs ``AudioRecognition._aclose()`` which cancels that same task. If the future is
+not awaited first, the trailing transcript is silently discarded (issue #6889).
+
+These tests pin the invariant that the commit future is awaited to completion
+before ``activity.aclose()`` gets a chance to cancel it.
+"""
+
+from __future__ import annotations
+
+import asyncio
+
+import pytest
+
+from livekit.agents import Agent
+from livekit.agents.voice.audio_recognition import AudioRecognition
+from livekit.agents.voice.transcription.synchronizer import _SyncedAudioOutput
+
+from .fake_io import FakeAudioInput
+from .fake_session import FakeActions, create_session
+
+pytestmark = [pytest.mark.unit, pytest.mark.virtual_time, pytest.mark.no_concurrent]
+
+
+async def _aclose_session(session: object, activity: object) -> None:
+    """Close the session and the transcription synchronizer that ``create_session``
+    wires onto ``session.output.audio`` (``run_session`` does this cleanup for us
+    normally, but these tests drive close by hand)."""
+    audio_output = session.output.audio  # type: ignore[attr-defined]
+    transcription_sync = (
+        audio_output._synchronizer if isinstance(audio_output, _SyncedAudioOutput) else None
+    )
+    await session.aclose()  # type: ignore[attr-defined]
+    if transcription_sync is not None:
+        await transcription_sync.aclose()
+
+
+async def _start_session_mid_utterance() -> tuple[object, AudioRecognition]:
+    """Start a session and return (activity, audio_recognition) while the user is
+    still being recognized (audio pushed, close not yet requested)."""
+    actions = FakeActions()
+    # A long trailing utterance whose final transcript lands well after we begin
+    # closing (end_time 2.5s + stt_delay 3.0s), mirroring a SIP party that hangs
+    # up right as they finish speaking.
+    actions.add_user_speech(
+        0.5,
+        2.5,
+        "the mailbox is full and cannot accept new messages at this time",
+        stt_delay=3.0,
+    )
+    session = create_session(
+        actions,
+        extra_kwargs={"session_close_transcript_timeout": 2.0},
+    )
+    await session.start(Agent(instructions="You are a helpful assistant."))
+
+    audio_input = session.input.audio
+    assert isinstance(audio_input, FakeAudioInput)
+    audio_input.push(0.1)
+    # let the activity + audio recognition spin up and begin recognizing
+    await asyncio.sleep(0.1)
+
+    activity = session._activity
+    assert activity is not None
+    audio_recognition = activity._audio_recognition
+    assert audio_recognition is not None
+    return activity, audio_recognition
+
+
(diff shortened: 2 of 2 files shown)


jonahlipsky:
@chenghao-mou Here's the PR!

chenghao-mou (approved):
lgtm, Devin's comment looks valid too.

jonahlipsky:
### Re: the Devin review on the trailing-transcript flush

Devin correctly flagged that awaiting `_commit_user_turn()` unguarded during close is a regression: if that flush future is **cancelled** (a lock-free `commit_user_turn()` racing the close cancels the in-flight flush task, which cancels the future) or **carries an exception** (e.g. EOU detection raising), it propagates out of `_aclose_impl` — which is wrapped in `@log_exceptions` (re-raises) — and abandons the rest of teardown (`activity.aclose()`, span ends, forward-audio/video cancel, `emit("close")`, `room_io.aclose()`), leaking the room connection with `_started` still `True`. Good catch — the pre-#6889 fire-and-forget code didn't have this exposure.

**Devin's suggested fix:**
```python
with contextlib.suppress(asyncio.CancelledError, Exception):
    await audio_recognition._commit_user_turn(...)
```

**What we went with instead:**
```python
commit_fut = audio_recognition._commit_user_turn(...)
await asyncio.wait([commit_fut])
if not commit_fut.cancelled() and (exc := commit_fut.exception()):
    logger.warning("error while committing the final user turn on close", exc_info=exc)
```

**Why we deviated:** `contextlib.suppress(asyncio.CancelledError, ...)` swallows `CancelledError` unconditionally — including a *genuine* cancellation of the close task itself — which breaks cooperative cancellation (a caller cancelling `aclose()` would be silently ignored). We only want to isolate the *flush's own* failure/cancellation from the close path, not make the close un-cancellable.

`asyncio.wait([commit_fut])` does exactly that: it waits for the flush to finish without adopting its outcome, so a cancelled/failed flush future doesn't raise into the close path, but a cancellation of the close task itself still propagates normally. Verified across all four cases (normal resolve / racing-cancel of the future / exception on the future → teardown continues; cancel of the close task → propagates). It also surfaces the failure via a logged warning rather than swallowing it silently.

Covered by two regression tests in `tests/test_session_close_commit_user_turn.py` (cancelled flush future and failed flush future must both still complete teardown), both confirmed to fail on the unguarded `await` and pass with this change.

chenghao-mou:
Going to close this one in favor of #6897 (you are added as a co-author). There was a same regression issue with the other task. 

Note on the cancellation part, it is true swallowing them might delay the aclose timeout,  I will create a new PR to handle it instead.

PR #6899, opened 2026-08-19
feat(maya): add Maya Research TTS plugin

saicherry93479:
## Summary

- Adds `livekit-plugins-maya` (`livekit-plugins/livekit-plugins-maya/`), a WebSocket TTS provider covering ten Indian languages plus Indian English, with every voice speaking all eleven. Connections come from `utils.ConnectionPool`, so one socket carries a whole conversation and the handshake is paid once rather than per utterance.
- Maya's v2 protocol needs a handshake per connection, so `_connect_ws` sends `{"type": "start", "v2": true, ...}` and waits for the `metadata` reply before returning the socket — turns sent before that are rejected rather than served in the older frame shape, so the pool must never hand out a socket that hasn't completed it. A rejected handshake raises `APIError` instead of yielding one that would fail every turn.
- Each agent turn maps to one Maya context: `SynthesizeStream` mints a `context_id`, sends each tokenized sentence under it with `continue: true`, and closes with an empty `continue: false` frame, yielding exactly one `end`. Sentences are not gated on the previous one's audio, so a multi-sentence reply streams continuously.
- A turn that ends without its terminator — the caller interrupted, or the task was cancelled — is dropped with a targeted `cancel` before the socket returns to the pool. Without it the server keeps generating and the next borrower of that connection receives audio from a dead turn.
- Voice, language and model are passed through as given rather than checked against a fixed list; Maya validates them and answers with an `error` frame, so values it adds later work without a release here.
- `ChunkedStream` runs over the same websocket rather than Maya's HTTP endpoint, so the plugin has one transport and one auth path.
- Registers the workspace member, the `maya` optional dependency, and an entry in the cross-provider `SYNTHESIZE_TTS` suite. `version.py` tracks the `livekit-agents` version (1.6.10), matching the sibling plugins so the `livekit-plugins-maya>=1.6.10` pin resolves.

## Example usage

```python
from livekit.plugins import maya

tts = maya.TTS(voice="Ananya", language="hi")
```

The API key is read from `MAYA_API_KEY`.

## Test plan

- [x] `uv run pytest tests/test_plugin_maya_tts.py` — 25 tests against a local server speaking the v2 protocol, covering the handshake contents, the turn model, connection reuse across turns, cancel-on-interrupt, no cancel for a turn that finished, a rejected handshake, and error frames
- [x] `uv run ruff check && uv run ruff format --check`
- [x] `uv run mypy livekit-plugins/livekit-plugins-maya/livekit/plugins/maya/`
- [x] Live against the production endpoint: one-shot synthesis, multi-sentence turns sharing one context, consecutive turns reusing the socket, and repeated barge-in followed by a clean turn with no leftover audio

Happy to adjust anything to match house style.


dheemutech:
## Maya Research update: tested contribution ready for review

Pushed `fd724714bc2f0dff6653578b9f979905be9c16f6` to this existing PR, preserving the original contribution and review fixes. It merges current upstream main (`e9a3422f`) and resolves the optional-dependency/lockfile conflicts without rebasing or force-pushing.

### What changed

- The integration is **Maya Research TTS**, not tied to one model generation. The current public API documents **Maya Calyx**, so the plugin now explicitly defaults to that model with `Aarav`. Model and voice strings still pass through to the service for future supported releases.
- The handshake now rejects incompatible sample rate, channel count or encoding before any text is sent, and closes the socket. Previously, a 48 kHz response only produced a warning while the emitter still advertised 24 kHz. That regression was reproduced before the fix.
- Added regression coverage for both streaming and one-shot metadata failures, changing connection settings, and preserving every sample in a final partial audio frame.
- Added the public [Maya Research Cookbook](https://github.com/MayaResearch/maya-cookbook): API reference, TTS quickstarts, LiveKit and Pipecat agents, streaming/cancellation, a from-scratch agent, and coding-agent instructions. The plugin README clearly distinguishes this contribution from an already-published upstream package.

Current example:

```python
from livekit.plugins import maya

tts = maya.TTS(model="Maya Calyx", voice="Aarav", language="hi")
```

### Verification

- **61 Maya plugin tests passed** on each of Python 3.10.21, 3.12.14 and 3.13.15. The non-3.12 compatibility checks used isolated plugin/core environments because the unrelated browser workspace member requires Python 3.12.
- **87 tests passed** in the combined Maya, audio-emitter and connection-pool suite on Python 3.12.
- Repository-wide Ruff lint/format passed; focused strict Maya plugin typing passed, including 3.10 and 3.13.
- **Eight real Maya checks passed**: Hindi/Telugu/English one-shot and streamed speech, reuse after completed turns, active-turn cancellation, and a new context after cancellation. Seven valid non-silent WAVs were generated privately, not added to the repository.
- **One fresh full synthetic agent turn passed** with the candidate Maya plugin and the cookbook's locked released LiveKit 1.8.0/provider environment: synthetic audio → Soniox → OpenRouter → Maya → recorded audio sink, with no agent errors. The separate source-tree agent setup could not load its unfetched Silero Git LFS asset, so that is not claimed as a full-agent pass.
- The 13-file PR diff was checked for credentials used during testing; zero matches. No credentials, customer recordings or private model assets are in the contribution.

Reproduce the keyless checks from the repository root:

```sh
uv sync --package livekit-plugins-maya --group dev --no-group typing --python 3.12 --locked
uv run --no-sync pytest tests/test_plugin_maya_tts.py tests/test_audio_emitter.py tests/test_connection_pool.py --plugin maya --unit -q
uv run --no-sync ruff check .
uv run --no-sync ruff format --check .
uv run --no-sync mypy --follow-imports=silent livekit-plugins/livekit-plugins-maya/livekit/plugins/maya
```

Live tests used synthetic text and approved local credentials. They are not human-listening scores, physical microphone/speaker checks, LiveKit room-transport certification, a latency SLA, or proof of an upstream release. GitHub checks and maintainer review remain separate.

Could the maintainers review this updated revision for merge and confirm any required packaging/release adjustments? Happy to address integration feedback on this same PR.

**Post-push verification:** [LiveKit lint and typing CI](https://github.com/livekit/agents/actions/runs/34261740466) and [the tests workflow](https://github.com/livekit/agents/actions/runs/34261740448) completed successfully. The provider-matrix and eval jobs were skipped by that workflow; the actual Maya checks above ran separately. [Cookbook CI](https://github.com/MayaResearch/maya-cookbook/actions/runs/34262115456) also passed at `f9bd349`, which pins this exact plugin revision. The PR is mergeable but still awaits review; no upstream merge or package release is claimed.

PR #6901, opened 2026-08-19
fix(runway): warn and surface session status when avatar leaves before max_duration

### LINKED ISSUE #6703: Runway plugin: maxDuration silently clamped to ~300s on realtime sessions (documented API limit is 30 min)
### Bug Description

**Plugin**: livekit-plugins-runway 1.6.7 (Python)

**Summary**: Sessions created by the plugin via POST /v1/realtime_sessions with `maxDuration: 1800` are consistently terminated server-side at ~300 seconds, recorded as normal COMPLETED with no error. Runway's documented limit for API integrations is up to 30 minutes, and Runway support confirms no documented tier clamp (account is API tier 1) — but can't explain the behavior.

**Evidence** (Runway's own GET /v1/realtime_sessions/{id} for two sessions created by the plugin with maxDuration 1800):

{ "id": "110817be-b717-4bf7-990b-243cf072e28a", "createdAt": "2026-08-04T02:06:32.079Z", "status": "COMPLETED", "duration": 317 }
{ "id": "75b5cb68-b40b-4657-a8ee-770e90ce34de", "createdAt": "2026-08-04T02:25:31.230Z", "status": "COMPLETED", "duration": 315 }

**Verified**: the plugin does forward the value (`body["maxDuration"] = self._max_duration` in `_create_session`), with `X-Runway-Version: 2024-11-06`. The creation response contains only `{ "id": ... }` — no echoed/normalized duration — so the applied limit is invisible client-side at both creation and termination.

**Impact**: the avatar participant dies mid-session (in our case, mid-sentence during long agent turns) with no error surfaced to the agent; the AgentSession continues without video. Anyone building >5-minute avatar sessions on this plugin hits this wall blind.

**Questions**: Is the ~300s limit expected on the LiveKit integration path? If it's server-side at Runway, could the plugin at least surface the effective limit (or the discrepancy) instead of silently proceeding? Happy to provide more session IDs or logs.

### Expected Behavior

Sessions honor the requested maxDuration up to the documented 30-minute API limit for Runway API integrations — or, if a server-side limit applies on the LiveKit path, the plugin/API surfaces the effective limit (at session create, or as a warning when the avatar session ends far short of the requested max_duration) instead of accepting 1800 and silently applying ~300.

### Reproduction Steps

```bash
1. Create an AgentSession (any STT/LLM/TTS) and a runway.AvatarSession with a custom avatarId and max_duration=1800; await avatar.start(session, room=ctx.room).
2. Run a session with a participant in the room for more than 5 minutes.
3. At ~300s the avatar participant leaves the room with no error surfaced to the agent; the AgentSession continues without video.
4. GET https://api.dev.runwayml.com/v1/realtime_sessions/{id} for that session returns {"status": "COMPLETED", "duration": 315-317} despite the requested 1800.
```

### Operating System

Windows 11 (worker, Python 3.14.6); rooms on LiveKit Cloud

### Models Used

LiveKit Inference STT (deepgram/nova-3), Anthropic claude-sonnet-4-6, ElevenLabs TTS, Runway gwm1_avatars (custom avatar)

### Package Versions

```bash
livekit==1.1.13
livekit-agents==1.6.7
livekit-plugins-runway==1.6.7
livekit-plugins-anthropic==1.6.7
livekit-plugins-elevenlabs==1.6.7
```

### Session/Room/Call IDs

Runway realtime session IDs: 110817be-b717-4bf7-990b-243cf072e28a (duration 317), 75b5cb68-b40b-4657-a8ee-770e90ce34de (duration 315)
LiveKit Cloud room: RM_V8ySdrYFqGYY

### Proposed Solution

```python
If the server applies a limit different from the requested maxDuration, surface it: e.g. log a warning at avatar session end when elapsed << max_duration, and/or expose the session's terminal status (the GET endpoint returns COMPLETED/duration and failureCode on failures) through an event so agents can react to the avatar dying instead of continuing blind.
```

### Additional Context

_No response_

_No response_

piyushrajyadav:
## What this fixes

Closes #6703

When using `runway.AvatarSession` with `max_duration=1800`, the Runway server silently terminates sessions at ~300 s. The plugin correctly forwards `maxDuration` in the creation request but the response contains only an `{"id": ...}` — no echoed duration — so the applied server-side cap is invisible client-side. The avatar participant then leaves the room mid-session with **no warning**, and the AgentSession continues without video.

### `livekit-plugins/livekit-plugins-runway/livekit/plugins/runway/avatar.py`

**Tracking + flag** (minimal state, no API surface change):
- `_session_started_at: float | None` — `time.monotonic()` snapshot captured right after `_create_session` returns.
- `_session_ending: bool` — set to `True` in both `_on_agent_session_close` and `_ensure_end_session_task()` so agent-initiated teardowns are distinguishable from unexpected server-side disconnects.

**`participant_disconnected` listener** registered on the room in `start()`:
- Filters to the avatar participant identity only.
- No-ops if `_session_ending` is `True` (normal teardown path).
- Logs a `WARNING` with `elapsed_seconds`, `session_id`, and (when `max_duration` was set and the gap is >60 s) a human-readable note that the server may have applied a lower cap.
- Schedules `_log_session_status()` as a fire-and-forget task.

**`_log_session_status(session_id)`** (new async helper):
- Calls `GET /v1/realtime_sessions/{id}` using the already-initialised HTTP session.
- Logs the Runway-reported `status`, `duration`, and `failureCode` fields at `WARNING` so operators can see what Runway recorded without reading the issue tracker.
- Errors are caught and logged at `DEBUG` — never raises, never blocks the session.

## Why this approach

The issue's proposed solution asks for two things:
1. A warning when elapsed << max_duration → covered by the `participant_disconnected` handler.
2. Exposing the terminal status (`status`, `duration`, `failureCode`) from the GET endpoint → covered by `_log_session_status`.

Both are best-effort: they surface information for operators without changing any error paths or adding retry logic, keeping the change minimal and safe.

## No-API-change guarantee

- No new public methods or classes.
- No new runtime dependencies (`time` is stdlib).
- Existing teardown paths (`_ensure_end_session_task`, `_end_runway_realtime_session`, `_cancel_runway_realtime_session`) are unchanged in their async logic; only `_session_ending = True` is added before the task is created.

## How it was tested

- `ruff check --fix` → `All checks passed!`
- `ruff format` → `1 file reformatted` (only `asyncio.ensure_future(...)` call condensed to one line by the formatter), then re-checked → `All checks passed!`
- The Runway plugin has no unit tests in the repo. The logic is straightforward event-listener and async-HTTP code following the same patterns as `_cancel_runway_realtime_session`. I have flagged that dedicated unit tests could mock the `participant_disconnected` event and the `aiohttp` GET call — happy to add them if the maintainers want to introduce a test harness for this plugin.

> **Flagging for maintainer review**: The 60-second threshold for the "early end" warning is a judgment call. If Runway ever legitimately ends a session within 60 s of the requested limit (e.g. for very short sessions), this could produce a misleading warning. An alternative would be a percentage-based threshold (e.g. warn if elapsed < 50% of max_duration). I went with 60 s as an absolute floor because the repro shows 1500 s remaining, which is unambiguously wrong, but I have noted this in the code comment.

### DIFF
--- a/livekit-plugins/livekit-plugins-runway/livekit/plugins/runway/avatar.py
+++ b/livekit-plugins/livekit-plugins-runway/livekit/plugins/runway/avatar.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import json
 import os
+import time
 from typing import Any
 
 import aiohttp
@@ -83,6 +84,15 @@ class AvatarSession(BaseAvatarSession):
         self._room: rtc.Room | None = None
         self._realtime_session_id: str | None = None
         self._end_session_task: asyncio.Task[None] | None = None
+        self._session_started_at: float | None = None
+        # Set to True once we intentionally initiate the end-of-session flow so
+        # that the participant_disconnected handler can tell the difference
+        # between an agent-initiated teardown and an unexpected server-side
+        # termination.
+        self._session_ending: bool = False
+        # Strong references to in-flight _log_session_status tasks so they
+        # cannot be garbage-collected before they emit their log lines.
+        self._pending_status_tasks: set[asyncio.Task[None]] = set()
 
     @property
     def avatar_identity(self) -> str:
@@ -133,10 +143,14 @@ class AvatarSession(BaseAvatarSession):
 
         logger.debug("starting Runway avatar session")
         await self._create_session(livekit_url, livekit_token, room.name)
+        self._session_started_at = time.monotonic()
 
         @agent_session.on("close")
         def _on_agent_session_close(_: Any) -> None:
+            self._session_ending = True
             self._ensure_end_session_task()
+
+        room.on("participant_disconnected", self._on_participant_disconnected)
 
         agent_session.output.replace_audio_tail(
             DataStreamAudioOutput(
@@ -146,6 +160,52 @@ class AvatarSession(BaseAvatarSession):
                 sample_rate=SAMPLE_RATE,
             ),
         )
+
+    def _on_participant_disconnected(self, participant: rtc.RemoteParticipant) -> None:
+        """Handle avatar participant leaving the room unexpectedly.
+
+        When the avatar participant disconnects without the agent having
+        initiated the teardown, log a warning and fetch the Runway session
+        status so operators can see why the session ended early.
+        """
+        if participant.identity != self._avatar_participant_identity:
+            return
+        # If we initiated the teardown ourselves, the disconnect is expected.
+        if self._session_ending:
+            return
+        elapsed = (
+            time.monotonic() - self._session_started_at
+            if self._session_started_at is not None
+            else None
+        )
+        extra: dict[str, object] = {
+            "session_id": self._realtime_session_id,
+            "elapsed_seconds": round(elapsed, 1) if elapsed is not None else None,
+        }
+        if is_given(self._max_duration) and elapsed is not None:
+            remaining = self._max_duration - elapsed
+            extra["requested_max_duration"] = self._max_duration
+            extra["remaining_seconds"] = round(remaining, 1)
+            if remaining > 60:  # ended more than 60 s before the requested limit
+                logger.warning(
+                    "Runway avatar participant left unexpectedly before max_duration; "
+                    "the server may have applied a lower limit than requested "
+                    "(%.0fs elapsed, %.0fs requested). "
+                    "The AgentSession will continue without video.",
+                    elapsed,
+                    self._max_duration,
+                    extra=extra,
+                )
+                if self._realtime_session_id is not None:
+                    self._schedule_log_session_status(self._realtime_session_id)
+                return
+        logger.warning(
+            "Runway avatar participant left unexpectedly. "
+            "The AgentSession will continue without video.",
+            extra=extra,
+        )
+        if self._realtime_session_id is not None:
+            self._schedule_log_session_status(self._realtime_session_id)
 
     async def _create_session(self, livekit_url: str, livekit_token: str, room_name: str) -> None:
         body: dict[str, object] = {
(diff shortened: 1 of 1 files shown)


piyushrajyadav:
Thanks for the review, Devin — that's a legitimate catch.

Fixed in the force-push to the branch (7050285):

- Added _pending_status_tasks: set[asyncio.Task[None]] to track all in-flight status lookups with a **strong reference**, preventing GC mid-flight.
- Replaced both bare syncio.ensure_future(...) calls with a new _schedule_log_session_status() helper that adds the task to the set and registers a done_callback to discard it on completion — the same bookkeeping pattern as _end_session_task.
- In close(), added wait asyncio.gather(*self._pending_status_tasks, return_exceptions=True) so any status fetch started near shutdown completes and emits its log line before the session tears down.


uff check + 
uff format both pass cleanly on the updated diff.

piyushrajyadav:
Addressed in e804311 (force-pushed). Three issues fixed in one pass:

**1. Handler lifetime (this finding):**
- Converted the inline @room.on('participant_disconnected') closure to a proper named method _on_participant_disconnected on the class (matching the base class pattern where _on_connection_state_changed / _on_conversation_item_added are named methods registered/deregistered symmetrically).
- In close(), call self._room.off('participant_disconnected', self._on_participant_disconnected) **before** super().aclose() clears self._room — so no late events can spawn new tasks after deregistration.

**2. Late task creation during shutdown window:**
- close() now sets self._session_ending = True as its very first action, so even if a disconnect event somehow fires during the syncio.shield(end_session_task) await, the handler will no-op.

**3. syncio.ensure_future -> syncio.create_task (anticipated finding):**
- Replaced with syncio.create_task(..., name=f'runway_log_session_status_{session_id}') to match the existing codebase convention (line 269 already uses create_task) and provide a descriptive task name for debug output.

**Ordering in close():**
1. _session_ending = True — handler no-ops on any late event
2. 
oom.off(...) — handler deregistered, no new invocations possible
3. _ensure_end_session_task() — end-session lifecycle runs
4. gather(*_pending_status_tasks) — drain any in-flight lookups (set is now stable since the handler can't add more)
5. _pending_status_tasks.clear() — clean state
6. super().aclose() — base class teardown


uff check + 
uff format both pass clean.

PR #6905, opened 2026-08-19
Sunset saaras and saarika v2.5

dhruvladia-sarvam:
Sunset saarika:v2.5 and saaras:v2.5.
Please move to saaras:v3 or saaras:v4


tinalenguyen (commented):
i think we should throw a warning if these model strings are passed, could you also fix the merge conflicts

dhruvladia-sarvam:
> i think we should throw a warning if these model strings are passed, could you also fix the merge conflicts

Resolved the merge conflicts, thanks for pointing it out. Agreed, added sunset warning for now, it has been intentionally advertised that those models are being actively deprecated.

tinalenguyen (approved):

PR #6908, opened 2026-08-19
fix(voice): keep the agent thinking while a tool is still running

### LINKED ISSUE #6904: Filler speech during a long tool call resets `agent_state` to `listening`, arming the user-away timer mid-tool
### Bug Description

`with_filler` exists to cover long-running tool calls. Using it during one causes the session to mark the user `away` while the tool is still running, which fires whatever inactivity handling the app has attached — in our case a "are you still there?" re-prompt spoken over a lookup the agent had itself just announced.

The user-away timer is armed only when both sides are idle:

```python
# agent_session.py (main) — _update_agent_state at L1966, _update_user_state at L2010
if state == "listening" and self._agent_state == "listening":
    self._set_user_away_timer()
else:
    self._cancel_user_away_timer()
```

A running tool is supposed to be protected by the transition to `thinking`:

```python
# agent_activity.py (main, ~L4197)
# wait for the tool execution to complete
tool_output.first_tool_started_fut.add_done_callback(
    lambda _: self._session._update_agent_state("thinking")
)
```

But `thinking` is asserted **once**, when the first tool starts. When a filler plays, the session goes to `speaking`, and at the end of that speech two separate paths put it back to `listening`, neither aware that a tool is still running:

```python
# 1. agent_activity.py, in _tts_task_impl        (main L3002 / 1.6.9 L2948)
if self._session.agent_state == "speaking":

# 2. agent_activity.py, _on_pipeline_reply_done  (main L2769 / 1.6.9 L2719)
#    registered as the done callback for the say() task itself
if not self._speech_q and (not self._current_speech or self._current_speech.done()):
```

Nothing restores `thinking` afterwards — the next transition to `thinking` (main L3498 / 1.6.9 L3448) requires `len(tool_output.output) > 0`, i.e. the tool has already finished. So for the remainder of the tool call the session sits in `listening`, the away timer arms, and after `user_away_timeout` the user is marked `away` even though the agent is demonstrably busy.

For a non-realtime session, `say()` dispatches to `_tts_task` → `_tts_task_impl` (main L1490-1502), so filler speech reaches both paths above. There is no tool-in-flight state anywhere in `agent_activity.py` for either path to consult.

Sequence:

1. Tool starts → `agent_state = "thinking"` → away timer cancelled. Correct.
2. Filler plays → `agent_state = "speaking"`.
3. Filler ends → `agent_state = "listening"` (L3002).
4. Both sides `listening` → `_set_user_away_timer()` **arms**.
5. Tool still running → `user_away_timeout` elapses → `user_state = "away"`.

The result is that the feature built for long tool calls disables the state protection that makes long tool calls safe. An agent that says nothing during the tool is protected; an agent that uses `with_filler` as documented is not.

Real transcript shape:

```
Agent:  "One moment. I'm checking."
        ...lookup runs, >5s...
Agent:  "Hello? Can you hear me?"
```

### Expected Behavior

A tool call in flight keeps the user-away timer disarmed, whether or not filler speech played during it. Either:

- filler speech restores the previous `agent_state` when it ends, rather than hardcoding `listening`; or
- the away timer is not armed while any tool is in flight, independently of `agent_state`.

### Reproduction Steps

```bash
1. Create an AgentSession with user_away_timeout set (we use 5.0; the default 15.0
   reproduces with a correspondingly slower tool).
2. Attach a user_state_changed handler that logs new_state == "away".
3. Define a function tool that wraps an operation slower than user_away_timeout in
   ctx.with_filler(...).
4. Trigger the tool and stay silent.
5. The filler plays; user_state flips to "away" while the tool is still running.
   Without the with_filler block, "away" does not fire — the session stays in
   "thinking" for the whole tool call.

- Minimal example -

session = AgentSession(..., user_away_timeout=5.0)

def _on_user_state(ev):
    print("user_state:", ev.new_state, "agent_state:", session.agent_state)

class Assistant(Agent):
    @function_tool()
    async def slow_lookup(self, ctx: RunContext, q: str) -> str:
        """A tool slower than the away timeout."""
        async with ctx.with_filler("Just a second.", delay=0.6):
            await asyncio.sleep(10)
        return "done"

# Logs: user_state: away agent_state: listening   <-- while the tool is still running
```

### Operating System

Linux 6.6.114.1-microsoft-standard-WSL2 (WSL2 on Windows 11)

### Models Used

_No response_

### Package Versions

```bash
livekit==1.1.12
livekit-agents==1.6.4
livekit-api==1.1.1
livekit-plugins-soniox==1.6.4
python==3.14.4
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
# Option 1 — restore the state the session held before the filler spoke, instead of
# assuming a filler always returns the session to listening.
# In _tts_task_impl, capture agent_state before speaking and restore it after, or skip
# the transition entirely when a tool is still in flight:

if self._session.agent_state == "speaking":
    if self._tools_in_flight():          # tool still running -> go back to thinking
    else:

# Option 2 — decouple the away timer from agent_state for this case: do not arm it
# while any tool is in flight, so no speech during a tool call can arm it.
```

### Additional Context

_No response_

_No response_

VihaanAgarwal:
`with_filler` speech during a long tool call knocks the session back to `listening`, which arms the user-away timer while the tool is still running. With `user_away_timeout` set, the user gets marked `away` mid-lookup and the app's inactivity handling fires over the agent's own tool call.

Two paths reset the state when a speech ends, and neither knows a tool is in flight:

- `_tts_task_impl` sets `listening` when the filler finishes playing
- `_on_pipeline_reply_done` does the same as the say task's done callback

The pipeline reply has the same gap without any filler: at playout end it only picks `thinking` when a tool already finished (`len(tool_output.output) > 0`), so a slow tool drops the state to `listening` until its result comes back.

The fix consults in-flight tool executions at those three spots. `_background_speeches` already tracks speeches that are awaiting their tools, so the two speech-end paths hand back `thinking` when it's non-empty. The reply task also stays in `thinking` when `exe_task` hasn't completed yet.

Added a test that runs a slow tool with a filler and a short `user_away_timeout`. On main the filler ends in `listening` and the user flips to `away` mid-tool. With the fix both speech ends return to `thinking` and the away timer stays disarmed. Unit suite, ruff and mypy are green.

Fixes #6904

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -2766,7 +2766,11 @@ class AgentActivity(RecognitionHooks):
 
     def _on_pipeline_reply_done(self, _: asyncio.Task[None]) -> None:
         if not self._speech_q and (not self._current_speech or self._current_speech.done()):
-            self._session._update_agent_state("listening")
+            # a speech awaiting its tool executions keeps the agent busy: stay in
+            # "thinking" so the user-away timer isn't armed mid-tool (#6904)
+            self._session._update_agent_state(
+                "thinking" if self._background_speeches else "listening"
+            )
             if self._audio_recognition:
                 self._audio_recognition._on_end_of_agent_speech(
                     ignore_user_transcript_until=time.time()
@@ -2999,7 +3003,9 @@ class AgentActivity(RecognitionHooks):
             self._session._conversation_item_added(msg)
 
         if self._session.agent_state == "speaking":
-            self._session._update_agent_state("listening")
+            self._session._update_agent_state(
+                "thinking" if self._background_speeches else "listening"
+            )
             if self._audio_recognition:
                 self._audio_recognition._on_end_of_agent_speech(
                     ignore_user_transcript_until=time.time()
@@ -3495,7 +3501,7 @@ class AgentActivity(RecognitionHooks):
             speech_handle._item_added([msg])
             current_span.set_attribute(trace_types.ATTR_RESPONSE_TEXT, forwarded_text)
 
-        if not speech_handle.interrupted and len(tool_output.output) > 0:
+        if not speech_handle.interrupted and (len(tool_output.output) > 0 or not exe_task.done()):
             self._session._update_agent_state("thinking")
             if self._audio_recognition:
                 self._audio_recognition._on_end_of_agent_speech(
--- a/tests/test_agent_session.py
+++ b/tests/test_agent_session.py
@@ -23,6 +23,7 @@ from livekit.agents import (
     MetricsCollectedEvent,
     ModelSettings,
     NotGivenOr,
+    RunContext,
     TurnHandlingOptions,
     UserInputTranscribedEvent,
     UserStateChangedEvent,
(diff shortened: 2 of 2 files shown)


livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -3495,7 +3501,7 @@ async def _next_segment() -> _SpeechSegment | None:
             speech_handle._item_added([msg])
             current_span.set_attribute(trace_types.ATTR_RESPONSE_TEXT, forwarded_text)
 
-        if not speech_handle.interrupted and len(tool_output.output) > 0:
+        if not speech_handle.interrupted and (len(tool_output.output) > 0 or not exe_task.done()):
```
longcw, line 3504: this branch also calls `_on_end_of_agent_speech` and `_restore_interruption_by_audio_activity`

longcw (commented):
looks good to me, but one thing I am not sure is

livekit-agents/livekit/agents/voice/agent_activity.py
longcw: maybe move it to the other branch, but may need to take care if there is no tool reply after exe_task, the agent state should be back to listening
```python
    # a tool still running keeps the agent busy: "listening" here would arm the
    # user-away timer mid-tool (#6904)
    still_running = not speech_handle.interrupted and not exe_task.done()
    self._session._update_agent_state("thinking" if still_running else "listening")

longcw:
closing in favor of https://github.com/livekit/agents/pull/6937, which is based on this pr but fixed the realtime path and avoided user away state when async tool is running. @VihaanAgarwal is listed as a co-author.

PR #6914, opened 2026-08-19
fix: place Gemini realtime user turns before replies

chenghao-mou:
Gemini can deliver the final user transcript after its reply, which makes session reports show the answer before the question. Anchor interim and final transcript events to the response generation time so the user turn stays before the reply it prompted.

Addresses AGT-3186

<details>

**Model:** GPT-5.6

> I got a customer report that gemini realtime model gave misaligned (timestamp wise) transcript.
>
> okay, can you open a PR?

</details>


davidzhao (approved):
nice fix!

PR #6921, opened 2026-08-20
fix(voice): place recorded audio where it actually played

longcw:
**Problem**

RecorderIO rebuilds the agent channel by inference. It collects the frames it forwarded, places them from the wall-clock moment `playback_finished` arrives, and pads the difference with silence. Nothing measures where the audio really went.

Four conditions break that inference: a sink that runs dry inside a segment, a late flush, audio discarded on `pause()`, and a microphone that stops delivering. Each one moves the audio, and none of them records the gap. A slow TTS or a muted participant may hit it.

**Fix**

A sink that owns its playback device now reports where its audio went. `AudioOutput` gains `playback_progressed` next to `playback_started` and `playback_finished`, and `_ParticipantAudioOutput` reports one run at a time off its own queue. Each report is past tense, so it is never revised, and a jump in its offset describes audio the sink discarded rather than played.

RecorderIO places both channels on one absolute timeline and leaves unwritten time silent. When a sink reports nothing, the recorder uses its segment endpoints instead, which is all a remote avatar worker can know. That removes the pause reconstruction, the end-anchored padding, and the `playback_position` truncation.

### DIFF
--- a/livekit-agents/livekit/agents/voice/io.py
+++ b/livekit-agents/livekit/agents/voice/io.py
@@ -124,11 +124,28 @@ class PlaybackStartedEvent:
 
 
 @dataclass
+class PlaybackProgressedEvent:
+    """A stretch of the current segment that has played.
+
+    Reported once the audio can no longer be discarded, so it is never revised.
+    """
+
+    started_at: float
+    """The timestamp (time.time()) at which this stretch began to play"""
+    offset: float
+    """Where it starts in the audio captured for the current segment"""
+    duration: float
+    """How much of it played"""
+
+
+@dataclass
 class AudioOutputCapabilities:
     pause: bool
 
 
-class AudioOutput(ABC, rtc.EventEmitter[Literal["playback_finished", "playback_started"]]):
+class AudioOutput(
+    ABC, rtc.EventEmitter[Literal["playback_finished", "playback_started", "playback_progressed"]]
+):
     def __init__(
         self,
         *,
@@ -167,6 +184,7 @@ class AudioOutput(ABC, rtc.EventEmitter[Literal["playback_finished", "playback_s
         if next_in_chain is not None:
             next_in_chain.on("playback_finished", self._forward_next_playback_finished)
             next_in_chain.on("playback_started", self._forward_next_playback_started)
+            next_in_chain.on("playback_progressed", self._forward_next_playback_progressed)
 
     def _forward_next_playback_finished(self, ev: PlaybackFinishedEvent) -> None:
         self.on_playback_finished(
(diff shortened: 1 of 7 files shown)


davidzhao (approved):

livekit-agents/livekit/agents/voice/recorder_io/recorder_io.py
```diff
@@ -17,18 +17,123 @@
...
+# beyond it keeps a drifting capture clock from sliding the channel
+RESYNC_TOLERANCE = 0.1
+
+
+@dataclass
+class _Captured:
+    channel: int
+    started_at: float
+    frame: rtc.AudioFrame
+    ends_run: bool
```
chenghao-mou: QQ: is there a place where we assign True?

livekit-agents/livekit/agents/voice/io.py
```diff
@@ -336,6 +370,7 @@ def set_next_in_chain(self, new: AudioOutput) -> None:
         if old is not None:
             old.off("playback_finished", self._forward_next_playback_finished)
             old.off("playback_started", self._forward_next_playback_started)
+            old.off("playback_progressed", self._forward_next_playback_progressed)
             if self._pending_playback_count > 0:
                 # stop audio still playing on the old sink
                 old.clear_buffer()
```
chenghao-mou, line 376: QQ: should we call `clear_buffer()` first before un-register the hooks since `_interrupted_event` can trigger event emits where there is an active `_wait_for_playout` task?

livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -181,10 +198,18 @@ async def _wait_buffered_audio() -> None:
...
+            self._report_run(
+                offset=self._source_pushed_duration - queued_duration, ended_at=time.time()
+            )
             self._audio_source.clear_queue()
             wait_for_playout.cancel()
         else:
             wait_for_interruption.cancel()
+            # the source drained, so everything pushed has played
+            self._report_run(offset=self._source_pushed_duration, ended_at=time.time())
+
```
chenghao-mou, line 215: should we use the dry at time if `flush()` occurs after the source has drained?

livekit-agents/livekit/agents/voice/recorder_io/recorder_io.py
longcw: removed and add flush to resampler now

livekit-agents/livekit/agents/voice/io.py
longcw: `clear_buffer` may not report progress syncly, I'll create a fix in a separate pr maybe add a force report method.

longcw: https://github.com/livekit/agents/pull/6967

PR #6926, opened 2026-08-20
feat(speechify): stream with timestamps via /v1/audio/stream/with-timestamps

luke-speechify:
## Summary

`SynthesizeStream` now streams audio with word-level marks from the real streaming endpoint instead of issuing one batch request per sentence.

For streaming-native models (`simba-3.0`, `simba-3.2`, or no model given, which the server resolves to `simba-3.0`) the stream uses `/v1/audio/stream/with-timestamps` via the SDK's `stream_with_timestamps`. Each `speech.chunk` event pushes its PCM audio as it arrives and converts its `speech_marks` into a timed transcript, so audio and aligned word timestamps are emitted while the audio is still landing. `speech.done` ends the sentence, `speech.error` surfaces as `APIStatusError`.

The legacy `simba-english` and `simba-multilingual` models do not serve that route, the server rejects them with 400 `speech_marks_unsupported`. They keep the per-sentence `/v1/audio/speech` path unchanged.

## Changes

- `tts.py`: route selection in `_synthesize`, `_stream_request_kwargs` (`output_format=pcm_24000`), `_supports_streaming_marks`, shared `_marks_to_timed`, rewired docstring
- `pyproject.toml` + `uv.lock`: `speechify-api` floor `>=3.0.1` to `>=4.0.0` (first release with `stream_with_timestamps`)
- `tts.py`: warning when a curated simba-3.2 voice (id suffix `_32`, e.g. `dominic_32`) is paired with any other model, including `NOT_GIVEN`
- `tests/test_plugin_speechify.py`: new hermetic unit tests

## Why it's not breaking

- Public API unchanged: no constant, function, or `TTS` method removed or altered in signature
- Legacy models issue byte-identical requests on the wire, verified by spying the raw SDK call from both the unchanged `synthesize()` path and the new `stream()` legacy branch
- Output contract identical: same 24 kHz mono PCM frames, same `USERDATA_TIMED_TRANSCRIPT` userdata
- The `speechify-api` floor bump only resolves forward; the methods the plugin uses exist in both versions

## Test plan

- `mypy` strict: clean on plugin and tests
- `ruff` format + lint: clean
- `pytest tests/test_plugin_speechify.py`: 7 tests, covering the streaming route, legacy fallback, `NOT_GIVEN` model, marks conversion, error propagation, and the voice/model warning
- Live E2E against the real API: simba-3.2 streamed 3.13s of audio with 7 word marks in order, legacy fallback 2.96s with marks, `synthesize()` multi-sentence offset accumulation correct


luke-speechify:
picked this up. CLA status noted, work is underway.

---
session: 108

tinalenguyen (approved):

PR #6944, opened 2026-08-22
ElevenLabs STT: add secondary_languages support

### LINKED ISSUE #6920: "secondary_languages" parameter in ElevenLabs STT plugin
### Feature Type

Would make my life easier

### Feature Description

ElevenLabs added new "secondary_languages" parameter to their STT engine. It allows to predict speech from limited subset of languages and improves accuracy for bilingual applications. 
https://elevenlabs.io/docs/api-reference/speech-to-text/v-1-speech-to-text-realtime

_No response_

### Additional Context

_No response_

rrfunde:
## Summary
- Adds a `secondary_languages: NotGivenOr[list[str]]` option to the ElevenLabs STT plugin, letting bilingual applications constrain speech prediction to a limited set of languages alongside `language_code`.
- Forwarded as repeated form fields for batch requests and repeated `secondary_languages` query params for Scribe v2 realtime sessions (mirroring how `keyterms` are handled).
- Supported in the constructor, `update_options()` (with reconnect propagation to live realtime streams), and both `STTOptions`.

Fixes #6920

## Testing
- New unit tests in `tests/test_plugin_elevenlabs_stt.py`: connect URL includes repeated `secondary_languages` params, omitted when NOT_GIVEN, defaults, and option updates.
- `uv run pytest tests/test_plugin_elevenlabs_stt.py` — 29 passed

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
+++ b/livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py
@@ -95,6 +95,7 @@ class STTOptions:
     sample_rate: STTRealtimeSampleRates
     server_vad: NotGivenOr[VADOptions | None]
     keyterms: NotGivenOr[list[str]]
+    secondary_languages: NotGivenOr[list[str]]
     no_verbatim: bool
     enable_logging: bool
     previous_text: str | None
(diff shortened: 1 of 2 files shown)


tinalenguyen (commented):
hi, thanks for the pr! could we gate this param on the realtime path and throw a warning for the batch path?

rrfunde:
Thanks for the review — agreed, gating on the realtime path is right. The batch endpoint does not accept `secondary_languages`, so the current diff silently sends a param the API ignores; the docstring already says "Supported for Scribe v2 realtime" and the code should enforce it.

I noticed you pushed `tina/11labs-secondary-languages` (a113ddc) doing exactly that — warn-and-drop at construction and in `update_options`, never added to the batch form. Happy either way: if you would rather land that one, feel free to close this in its favour. Otherwise I will apply the same gating here and push to this branch — just let me know which you prefer so we do not duplicate the work.

tinalenguyen:
closing this PR since i took over in #7086 which has been merged! this will be available in the next release

PR #6945, opened 2026-08-22
speechmatics: expose end_of_turn_config and vad_config options

### LINKED ISSUE #6924: speechmatics: `EndOfTurnConfig` / `VoiceActivityConfig` are not tunable
### Feature Type

Would make my life easier

## Problem

In the Speechmatics plugin, `ADAPTIVE` (and `SMART_TURN`) turn detection computes the effective end-of-turn wait as `end_of_utterance_silence_trigger × EndOfTurnConfig penalty multipliers`, minus recent TTFB, floored at `min_end_of_turn_delay` (`speechmatics.voice._client.VoiceAgentClient._calculate_finalize_delay`). With the SDK defaults:

- The preset-enabled client-side Silero VAD reports `speech_ended` after **0.18s** of silence (`VoiceActivityConfig.silence_duration` default).
- With Smart Turn not enabled (the `ADAPTIVE` preset), every VAD stop matches the `VAD_STOPPED + SMART_TURN_INACTIVE` penalty: **×0.2**.
- A fragment the STT itself punctuated as sentence-final matches `ENDS_WITH_FINAL + ENDS_WITH_EOS`: **×0.5**, and simultaneously loses the ×2.0 "not end-of-sentence" stretch.

At `end_of_utterance_silence_trigger=0.6`, the post-VAD wait is `0.6 × 0.2 × 0.5 = 0.06s` (punctuated) or `0.6 × 0.2 × 2.0 = 0.24s` (unpunctuated), floored as low as `0.01s` after TTFB subtraction. Total pause tolerance is roughly **0.25-0.45s** regardless of the configured trigger (even 1.5s yields ≤0.6s, since the trigger only scales the base linearly).

On phone calls this splits natural mid-sentence pauses (a caller recalling a date of birth or spelling out an email address) into multiple committed user turns, and the agent replies to each fragment.

## Why it can't be fixed from the integration today

The plugin's `STT` constructor does not accept `end_of_turn_config` or `vad_config`, and neither is in `_prepare_config()`'s `advanced_params` whitelist, so the penalty table, `min_end_of_turn_delay`, `base_multiplier`, and the VAD's `silence_duration`/`threshold` are unreachable. No combination of the exposed parameters (`end_of_utterance_silence_trigger`, `end_of_utterance_max_delay`, `max_delay`, `punctuation_overrides`, …) changes the ×0.2 fast path.

## Proposal

Expose both objects through the existing preset-override mechanism, like the other advanced parameters:

```python
stt = speechmatics.STT(
    turn_detection_mode=TurnDetectionMode.ADAPTIVE,
    end_of_utterance_silence_trigger=0.6,
    end_of_turn_config=EndOfTurnConfig(min_end_of_turn_delay=0.15, use_forced_eou=True),
    vad_config=VoiceActivityConfig(enabled=True, silence_duration=0.3),
)
```

- `end_of_turn_config: NotGivenOr[EndOfTurnConfig]` and `vad_config: NotGivenOr[VoiceActivityConfig]` constructor kwargs (both types are already public exports of `speechmatics-voice`, which the plugin already depends on).
- Added to `STTOptions` and to the `advanced_params` whitelist, so a provided value replaces the preset's object and an omitted kwarg changes nothing; there is no behavior change for existing users.

Granular kwargs (e.g. `vad_silence_duration`) would also work if the config objects are considered too SDK-coupled; open to either shape.

## Workarounds / Alternatives

**Tuning the exposed parameters** (implemented, partial): raising `end_of_utterance_silence_trigger` only scales the base linearly; with the ×0.2 VAD-stop multiplier hardcoded, even a 1.5s trigger yields ≤0.6s of pause tolerance. Adding `punctuation_overrides={"sensitivity": 0.2}` reduces false sentence-final periods at hesitations so fewer turns hit the ×0.5 fast branch; it helps, but cannot lift the ×0.2 ceiling either.

**Switching turn-detection modes** (evaluated):

- `EXTERNAL` (passing a `vad`) bypasses the penalty machinery entirely, but endpointing degrades to a plain silence window and the EOU parameters become inert.
- `FIXED` honors the trigger literally with no penalties, but loses the fast path for completed speech, so every turn pays the full trigger as flat latency.
- `SMART_TURN` removes the blind ×0.2 (the penalty requires `SMART_TURN_INACTIVE`), but requires the ONNX model on disk (a missing model **silently falls back to ADAPTIVE**), runs synchronous inference on the event loop per end-of-speech, and its incomplete-verdict stretches added noticeable latency in our testing.

**Subclassing `STT` and overriding `_prepare_config()`** (current workaround): mutating `config.end_of_turn_config` / `config.vad_config` after the preset loads works, but it depends on a private method and SDK internals, so it's fragile across plugin releases, which is exactly what this issue asks to make unnecessary.

### Additional Context

_No response_

rrfunde:
## Summary
- Adds `end_of_turn_config: NotGivenOr[EndOfTurnConfig]` and `vad_config: NotGivenOr[VoiceActivityConfig]` constructor kwargs to the Speechmatics STT plugin (both types are already public exports of `speechmatics-voice`).
- Both are stored in `STTOptions` and added to the `advanced_params` override whitelist in `_prepare_config()`, so a provided value replaces the preset's object and an omitted kwarg changes nothing — no behavior change for existing users.

This makes adaptive endpointing tunable again: with the preset defaults, effective pause tolerance is ~0.25–0.45s regardless of `end_of_utterance_silence_trigger` (the ×0.2 VAD-stop penalty and `min_end_of_turn_delay` floor were unreachable), which splits natural mid-sentence pauses on phone calls into multiple turns. Previously the only workaround was subclassing `STT` and mutating private internals.

Fixes #6924

## Testing
- New unit tests in `tests/test_plugin_speechmatics_stt.py`: presets are used when kwargs omitted, provided objects replace the preset's, options stored.
- `uv run pytest tests/test_plugin_speechmatics_stt.py` — 6 passed

### DIFF
--- a/livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
+++ b/livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
@@ -38,10 +38,12 @@ from speechmatics.voice import (
     AdditionalVocabEntry,
     AgentServerMessageType,
     AudioEncoding,
+    EndOfTurnConfig,
     OperatingPoint,
     SpeakerFocusConfig,
     SpeakerFocusMode,
     SpeakerIdentifier,
+    VoiceActivityConfig,
     VoiceAgentClient,
     VoiceAgentConfig,
     VoiceAgentConfigPreset,
@@ -109,6 +111,8 @@ class STTOptions:
     max_delay: float | None = None
     end_of_utterance_silence_trigger: float | None = None
     end_of_utterance_max_delay: float | None = None
+    end_of_turn_config: EndOfTurnConfig | None = None
+    vad_config: VoiceActivityConfig | None = None
     punctuation_overrides: dict | None = None
     include_partials: bool | None = None
 
@@ -135,6 +139,8 @@ class STT(stt.STT):
         max_delay: NotGivenOr[float] = NOT_GIVEN,
         end_of_utterance_silence_trigger: NotGivenOr[float] = NOT_GIVEN,
         end_of_utterance_max_delay: NotGivenOr[float] = NOT_GIVEN,
+        end_of_turn_config: NotGivenOr[EndOfTurnConfig] = NOT_GIVEN,
+        vad_config: NotGivenOr[VoiceActivityConfig] = NOT_GIVEN,
         additional_vocab: NotGivenOr[list[AdditionalVocabEntry]] = NOT_GIVEN,
         punctuation_overrides: NotGivenOr[dict] = NOT_GIVEN,
         speaker_sensitivity: NotGivenOr[float] = NOT_GIVEN,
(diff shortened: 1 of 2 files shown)


rrfunde:
Good catch — fixed in d481273: the constructor now raises `ValueError` when `vad_config` with `enabled=True` is combined with `turn_detection_mode=EXTERNAL`, since the external VAD (or manual finalize()) already controls turn boundaries. Documented in the docstring, with regression tests.

rrfunde:
Fixed in cf4bf51: the vad_config conflict is now validated before the Silero auto-load, so callers get the configuration error instead of an unrelated Silero ImportError (or a needlessly loaded model).

tinalenguyen (approved):
thank you for the PR!

PR #6946, opened 2026-08-22
sarvam: align TTS validation with the live Bulbul API

### LINKED ISSUE #6774: sarvam: TTS parameter validation diverges from the live Bulbul API

rrfunde:
## Summary
Three fixes for divergences between the plugin's validation and what the Bulbul API actually accepts ([docs](https://docs.sarvam.ai/api/getting-started/models/bulbul)):

1. **`bulbul:v3` speakers**: accept the 9 documented voices that were previously rejected by `validate_model_speaker_compatibility` (`anand, tarun, sunny, mani, gokul, vijay, mohit, rehan, soham`) and stop offering `amelia`/`sophia`, which the API rejects with "Speaker 'amelia' is not recognized" (they remain available on `v3-beta`, where they work).
2. **REST-only sample rates**: 32000/44100/48000 Hz are REST-only per the docs; `stream()` now fails locally with a clear error instead of an opaque server error frame mid-session. `synthesize()` still accepts them.
3. **Per-model `pace` range**: validated as 0.5–2.0 on v3/v3-beta and 0.3–3.0 on v2 (both in the constructor and `update_options()\)), instead of applying the v2 range everywhere and letting `pace=2.5` on v3 fail with a 400 at request time.

Fixes #6774

## Testing
- New unit tests in `tests/test_plugin_sarvam_tts.py` covering all three behaviors.
- `uv run pytest tests/test_plugin_sarvam_tts.py` — 8 passed

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
@@ -794,6 +794,20 @@ class TTS(tts.TTS):
         output_audio_codec: str | None = None,
     ) -> None:
         """Update TTS options with validation."""
+        # Validate the effective model/pace combination before mutating
+        # anything, so a rejected update never leaves partial state behind and
+        # an atomic `update_options(model=..., pace=...)` is accepted when the
+        # supplied pace is valid for the new model.
+        if model is not None or pace is not None:
+            effective_model = model if model is not None else self._opts.model
+            effective_pace = pace if pace is not None else self._opts.pace
+            low, high = _pace_range_for_model(effective_model)
+            if not low <= effective_pace <= high:
+                raise ValueError(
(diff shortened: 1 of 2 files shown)


rrfunde:
Good catch — fixed in 49a8600: `update_options()` now revalidates the retained pace whenever the model changes and raises with a clear local error (e.g. `Pace 2.5 is invalid for bulbul:v3: must be between 0.5 and 2.0`) instead of letting the next request fail at the API. Regression test included.

rrfunde:
Both fixed in c65f527: update_options() now validates the effective model/pace combination upfront — an atomic `update_options(model="bulbul:v3", pace=1.0)` from a v2 instance with pace=2.5 succeeds, and any rejected combined update leaves no partial state behind. Tests cover both cases.

PR #6947, opened 2026-08-22
aws: omit temperature/topP for Bedrock models that reject them

### LINKED ISSUE #6581: Bedrock rejects 'temperature' params on newer models

rrfunde:
## Summary

Claude Opus 4.7/4.8 and Sonnet 5 on Bedrock reject any Converse request whose `inferenceConfig` contains `temperature` or `topP` with:

```
ValidationException: `temperature` is deprecated for this model.
```

The plugin now detects these models (case-insensitive substring match, so region-prefixed IDs like `us.anthropic.claude-opus-4-7` and inference profile ARNs containing the model name are covered) and omits both sampling parameters, logging a warning when values were explicitly configured. Other models are unaffected.

Fixes livekit/agents#6581

## Testing

* New unit tests in `tests/test_plugin_aws_llm.py`: temperature/topP sent for supporting models, omitted for opus-4-7/4-8 and sonnet-5 across ID forms, default model unaffected.
* `uv run pytest tests/test_plugin_aws_llm.py` — 4 passed
* `ruff format/check` clean; `mypy -p livekit.plugins.aws` shows only a pre-existing missing-import note on clean main (unrelated file)

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
@@ -82,6 +82,7 @@ class LLM(llm.LLM):
         additional_request_fields: NotGivenOr[dict[str, Any]] = NOT_GIVEN,
         cache_system: bool = False,
         cache_tools: bool = False,
+        supports_sampling_params: NotGivenOr[bool] = NOT_GIVEN,
         session: AioSession | aioboto3.Session | None = None,
     ) -> None:
         """
@@ -107,6 +108,11 @@ class LLM(llm.LLM):
             additional_request_fields (dict[str, Any], optional): Additional request fields to send to the AWS Bedrock Converse API. Defaults to None.
             cache_system (bool, optional): Caches system messages to reduce token usage. Defaults to False.
             cache_tools (bool, optional): Caches tool definitions to reduce token usage. Defaults to False.
+            supports_sampling_params (bool, optional): Explicit override for whether the model accepts
+                'temperature'/'top_p'. By default the plugin detects known-rejecting models from the
+                model ID, which cannot cover application inference-profile ARNs that hide the
+                underlying model name — set False for those profiles to have sampling parameters
+                dropped instead of triggering a ValidationException. Defaults to NOT_GIVEN (auto-detect).
             session (AioSession, optional): Optional aiobotocore session to use. Passing a legacy
                 aioboto3.Session is deprecated but still accepted.
         """  # noqa: E501
@@ -136,6 +142,11 @@ class LLM(llm.LLM):
             additional_request_fields=additional_request_fields,
             cache_system=cache_system,
             cache_tools=cache_tools,
+        )
+        self._supports_sampling_params = (
+            supports_sampling_params
+            if is_given(supports_sampling_params)
+            else not _model_rejects_sampling_params(bedrock_model)
         )
 
     @property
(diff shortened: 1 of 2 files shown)


rrfunde:
Fixed in 45e168d: the warning is now emitted once per `LLM` instance (`_sampling_params_warned` flag), with a regression test asserting exactly one warning across three turns.

rrfunde:
Addressed all three in 091984f:
- **Opaque inference profiles**: added a `supports_sampling_params` constructor override so users with application inference-profile ARNs that hide the model name can explicitly declare capability; auto-detection remains the default.
- **Redaction**: the warning message no longer interpolates the model ID — it's static, with the value moved to a structured `lk.pii.model` attribute.
- **CI coverage**: test module re-marked as `pytest.mark.unit` so it runs in the `--unit` gate.

rrfunde:
Addressed in f647f79: auto-detection now skips application inference profiles entirely — their names are user-defined, so substring matching could misclassify a profile merely named after a rejecting model (e.g. `.../claude-opus-4-7-prod`) and silently drop configured sampling params. Application profiles default to sending sampling params; the explicit `supports_sampling_params=False` override covers profiles targeting rejecting models. Test added for the misclassification case.

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
```diff
@@ -39,6 +39,28 @@
...
 DEFAULT_TEXT_MODEL = "amazon.nova-2-lite-v1:0"
 
+# Model IDs that reject ``temperature``/``topP`` in the Converse API's
+# ``inferenceConfig`` with a ValidationException ("... is deprecated for this
+# model"). Matched as case-insensitive substrings so region-prefixed IDs and
+# inference profile ARNs containing the model name are covered too.
+_MODELS_REJECTING_SAMPLING_PARAMS = (
+    "claude-opus-4-7",
+    "claude-opus-4-8",
+    "claude-sonnet-5",
```
longcw: missing opus 5?

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
```diff
@@ -39,6 +39,28 @@
...
+)
```
longcw, line 52: should "claude-opus-5" be included?

longcw (approved):
thanks for the pr! could you sign the CLA, clean up the comments and something nit:

livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/llm.py
rrfunde: Yes — good catch. Opus 5 has the same removal as 4.7/4.8 and Sonnet 5: `temperature`/`top_p`/`top_k` are no longer part of the request surface, so Converse rejects them the same way. The accepting/rejecting line is really "4.6 and older accept, 4.7 and newer reject".

Added `claude-opus-5`, and `claude-fable-5` for the same reason (harmless as a substring entry if it is not reachable through Converse yet), plus updated the `temperature` docstring which still implied Opus 5 was fine. Substring matching stays safe here — none of these are prefixes of an older accepting ID (`claude-opus-4-6`, `claude-sonnet-4-6`, `claude-opus-4-5` all miss).

PR #6949, opened 2026-08-22
fix(langchain): report token usage on ChatChunk

### LINKED ISSUE #6948: LangChain LLMAdapter never reports token usage, so llm_node_tps is always absent
### Summary

The LangChain plugin's `LLMAdapter` never populates `usage` on the `ChatChunk` objects it emits. As a consequence, for LangGraph-backed agents:

- `llm_node_tps` is always absent from per-turn latency metrics
- there is no token usage to aggregate for session-level usage tracking

`llm_node_ttft`, `tts_node_ttfb` and `playback_latency` are reported normally, so this is specific to anything derived from token counts.

### Environment

- `livekit-agents` 1.7.0
- `livekit-plugins-langchain` 1.7.0
- `langchain` 1.3.16 / `langchain-core` 1.6.0 / `langgraph` 1.2.11
- Python 3.12.8, macOS (arm64)

### Steps to reproduce

1. Build a minimal LangGraph workflow and use it as the session LLM via `langchain.LLMAdapter(graph=...)`.
2. Log per-turn metrics from `ChatMessage.metrics`:

```python
@session.on("conversation_item_added")
def on_conversation_item_added(ev: ConversationItemAddedEvent) -> None:
    metrics = getattr(ev.item, "metrics", None)
    if metrics:
        logger.info("metrics [%s] %s", ev.item.role, metrics)
```

3. Have a few turns of conversation.

### Actual behavior

`llm_node_tps` never appears, across every turn:

```
metrics [assistant] llm_node_ttft=0.750  tts_node_ttfb=1.263  playback_latency=0.095
metrics [assistant] llm_node_ttft=0.597  tts_node_ttfb=1.057  playback_latency=0.000
metrics [assistant] llm_node_ttft=0.581  tts_node_ttfb=1.123  playback_latency=0.036
```

### Root cause

`llm_node_tps` is computed in `livekit/agents/voice/generation.py` and requires `usage` to have been seen on the stream:

```python
if (
    usage is not None
    and first_content_at is not None
    and (streaming_window := last_content_at - first_content_at) > 0
):
    data.tps = usage.completion_tokens / streaming_window
```

`usage` is collected from the chunks themselves:

```python
if chunk.usage is not None:
    usage = chunk.usage
```

But `_to_chat_chunk` in `livekit/plugins/langchain/langgraph.py` constructs `ChatChunk` without `usage`:

```python
return llm.ChatChunk(
    id=message_id,
    delta=llm.ChoiceDelta(
        role="assistant",
        content=content,
    ),
)
```

`usage` does not appear anywhere in the `livekit-plugins-langchain` package, so the condition can never be satisfied. The other two conditions are met — the adapter does stream multiple chunks, and `llm_node_ttft` is reported — so the only missing piece is the token count.

### Suggested fix

LangChain's `AIMessageChunk` already carries the information as `usage_metadata: UsageMetadata | None`, with `input_tokens` / `output_tokens` / `total_tokens`, which maps onto `CompletionUsage` directly:

```python
usage = None
if isinstance(msg, BaseMessageChunk) and (um := getattr(msg, "usage_metadata", None)):
    usage = llm.CompletionUsage(
        completion_tokens=um["output_tokens"],
        prompt_tokens=um["input_tokens"],
        total_tokens=um["total_tokens"],
    )

return llm.ChatChunk(
    id=message_id,
    delta=llm.ChoiceDelta(role="assistant", content=content),
    usage=usage,
)
```

One caveat worth noting: some providers only emit `usage_metadata` when explicitly asked (for example `ChatOpenAI(stream_usage=True)`), so the chunk carrying usage may have empty content. The current implementation returns `None` early for any chunk without text content, which would drop a usage-only chunk even after the mapping above is added.

Happy to open a PR if this direction looks right.

leilei3167:
Fixes #6948.

`LLMAdapter._to_chat_chunk` was building `ChatChunk` without `usage`, so LangGraph-backed sessions never had token counts. That means `llm_node_tps` stays missing even though `llm_node_ttft` / TTS timings show up normally.

This maps `AIMessageChunk.usage_metadata` (`input_tokens` / `output_tokens` / `total_tokens`) onto `CompletionUsage`. Some providers only attach usage on a later chunk with empty content (e.g. `ChatOpenAI(stream_usage=True)`), so those usage-only chunks are no longer dropped as `None`.

Note: providers that require an explicit stream-usage flag still need that flag set, or there is nothing to copy.

## Test plan
- [x] `uv run pytest tests/test_langgraph.py --unit` (16 passed)
- [x] `ruff check --fix` and `ruff format` on the touched files

### DIFF
--- a/livekit-plugins/livekit-plugins-langchain/livekit/plugins/langchain/langgraph.py
+++ b/livekit-plugins/livekit-plugins-langchain/livekit/plugins/langchain/langgraph.py
@@ -234,6 +234,7 @@ def _extract_message_chunk(item: Any) -> BaseMessageChunk | str | None:
 def _to_chat_chunk(msg: str | Any) -> llm.ChatChunk | None:
     message_id = utils.shortuuid("LC_")
     content: str | None = None
+    usage: llm.CompletionUsage | None = None
 
     if isinstance(msg, str):
         content = msg
@@ -241,6 +242,15 @@ def _to_chat_chunk(msg: str | Any) -> llm.ChatChunk | None:
         content = msg.text
         if getattr(msg, "id", None):
             message_id = msg.id  # type: ignore
+        if um := getattr(msg, "usage_metadata", None):
+            # Some providers only attach usage_metadata on a later, empty-content
+            # chunk (e.g. ChatOpenAI(stream_usage=True)). Keep those so voice
+            # generation can compute llm_node_tps.
+            usage = llm.CompletionUsage(
+                completion_tokens=um["output_tokens"],
+                prompt_tokens=um["input_tokens"],
+                total_tokens=um["total_tokens"],
+            )
     elif isinstance(msg, dict):
         raw = msg.get("content")
         if isinstance(raw, str):
@@ -250,7 +260,7 @@ def _to_chat_chunk(msg: str | Any) -> llm.ChatChunk | None:
         if isinstance(raw, str):
             content = raw
 
-    if not content:
+    if not content and usage is None:
         return None
 
     return llm.ChatChunk(
@@ -258,5 +268,8 @@ def _to_chat_chunk(msg: str | Any) -> llm.ChatChunk | None:
         delta=llm.ChoiceDelta(
             role="assistant",
             content=content,
-        ),
+        )
+        if content
+        else None,
+        usage=usage,
     )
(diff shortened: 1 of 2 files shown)


leilei3167:
Hi — just a friendly ping on this PR whenever someone with merge rights has a moment. Happy to address any feedback. Thanks!

PR #6963, opened 2026-08-24
fix(realtime): separate server turn detection from server replies

longcw:
**Problem**

`capabilities.turn_detection` tells the framework who owns the input audio buffer. The openai plugin sets it from `create_response`, which says who answers the turn.

With `turn_detection=ServerVad(create_response=False)` the server still segments and commits every turn, but the framework reads the capability as client-side turn taking and commits as well. Each turn logs `input_audio_buffer_commit_empty`, which #6642 suppressed rather than removed.

Seven reads of this capability want the "the server segments" meaning, so barge-in gatekeeping and the default VAD were wrong for this configuration too.

**Fix**

`turn_detection` goes back to its first meaning: the server detects turns, and it segments and commits the input audio itself.

The half it absorbed becomes `auto_turn_reply_generation`, which is `True` by default. Only the openai plugin needs a change, and a plugin that does not know the field behaves as before.

Five of the seven wrong reads become correct with no edit. `AgentActivity` gains `_rt_server_reply_enabled` for the six places that want "the server answers", and only the two `commit_audio()` calls needed an explicit gate.

Note: with an explicit `create_response=False`, endpointing now comes from the server's own speech events, so a client `turn_detection="vad"` is ignored with a warning. Pass `turn_detection=None` to keep full client control. This ordering is what closes the gap where `response.create` could otherwise overtake the server's commit and answer a turn the conversation does not hold yet.

### DIFF
--- a/livekit-agents/livekit/agents/llm/realtime.py
+++ b/livekit-agents/livekit/agents/llm/realtime.py
@@ -61,7 +61,8 @@ class RealtimeCapabilities:
     message_truncation: bool
     """Whether generated assistant messages can be truncated after interruption"""
     turn_detection: bool
-    """Whether the model emits server-side speech start and stop events for turn taking"""
+    """Whether the model detects turns server-side, segmenting and committing the input audio
+    itself. The client must not commit or clear that buffer."""
     user_transcription: bool
     """Whether the model emits user audio transcription events"""
     auto_tool_reply_generation: bool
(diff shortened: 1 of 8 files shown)


davidzhao (approved):

longcw:
closing and create a fix in openai realtime instead https://github.com/livekit/agents/pull/6965

PR #6967, opened 2026-08-24
fix(voice): report played audio across an audio sink swap

longcw:
**Problem:** `replace_audio_tail` can swap the audio sink while a segment still plays. Each sink counts playback from its own first frame, not from the start of the segment, so the recorder placed the audio of a new sink at the segment start and dropped what the detached sink played.

**Fix:** Room I/O reports its playhead when it clears its buffer, and the proxy maps every report onto the segment timeline. A sink that reports no runs of its own, such as a remote avatar worker, is reported on its behalf.

Follow-up of https://github.com/livekit/agents/pull/6921#discussion_r3835990289

<details>

> **How to see it**
>
> `tests/test_audio_sink_proxy.py` is a new file and every case in it describes behaviour main does not have. The recorder-side effect is covered by the new cases in `tests/test_recorder_io.py`.
>
> **Blast radius**
>
> `set_next_in_chain` has two callers: the proxy's own constructor, and `replace_audio_tail` at `io.py:731`. `replace_audio_tail` has one in-tree caller, `inference/avatar.py:318`, so the swap this fixes is the inference avatar handover. Every other sink chain reaches the untouched path.
>
> **The three fields and when they reset**
>
> `_offset_base` is where the current sink's zero sits inside the segment, `_sink_reported` records whether that sink ever emitted progress of its own, and `_sink_started_at` is when it began. All three reset on the first frame of a new segment in `capture_frame`; only `_offset_base` advances on a swap.
>
> **Why a detached sink's run is capped by elapsed time**
>
> `min(self._pushed_duration - self._offset_base, time.time() - self._sink_started_at)`. A sink can be handed more audio than it had time to play, and once it is detached it cannot be asked, so the smaller of the two is the most that can be claimed.
>
> **Why the old sink is flushed on the clear**
>
> A sink ends a segment when it is flushed, and nothing flushes it once detached. A clear alone would leave it open and holding audio.

</details>

### DIFF
--- a/livekit-agents/livekit/agents/voice/io.py
+++ b/livekit-agents/livekit/agents/voice/io.py
@@ -385,11 +385,22 @@ class _AudioSinkProxy(AudioOutput):
             # progress only observes, so the clear is still the sink's last word
             old.off("playback_progressed", self._forward_next_playback_progressed)
 
-            if self._capturing:
-                # it cannot be asked once detached, so all it was given counts as played
-                if not self._sink_reported:
-                    self._report_run(self._pushed_duration - self._offset_base)
+            # it cannot be asked once detached: what it was given counts as played, capped by
+            # how long it played, unless the recorder has the whole segment to place
+            if (
+                self._pending_playback_count > 0
+                and not self._sink_reported
+                and (self._capturing or self._offset_base)
+                and self._sink_started_at is not None
+            ):
+                self._report_run(
+                    min(
+                        self._pushed_duration - self._offset_base,
+                        time.time() - self._sink_started_at,
+                    )
+                )
 
+            if self._capturing:
                 # the new sink counts from its own zero, this far into the segment
                 self._offset_base = self._pushed_duration
                 self._sink_reported = False
@@ -412,16 +423,15 @@ class _AudioSinkProxy(AudioOutput):
             self.on_playback_finished(playback_position=self._pushed_duration, interrupted=True)
 
     def _report_run(self, duration: float) -> None:
-        """Report a run the current sink played but never reported itself."""
-        if duration <= 0:
+        """Report a run the current sink played but never reported itself.
+
+        A sink that never said playback started has nothing to place: nothing reached it.
+        """
+        if duration <= 0 or self._sink_started_at is None:
             return
 
         self.on_playback_progressed(
-            started_at=self._sink_started_at
-            if self._sink_started_at is not None
-            else time.time() - duration,
-            offset=self._offset_base,
-            duration=duration,
+            started_at=self._sink_started_at, offset=self._offset_base, duration=duration
         )
 
     def _forward_next_playback_started(self, ev: PlaybackStartedEvent) -> None:
@@ -456,8 +466,9 @@ class _AudioSinkProxy(AudioOutput):
             self._sink_started_at = None
 
         await super().capture_frame(frame)
-        await self.next_in_chain.capture_frame(frame)
+        # counted before the handover, so a swap during it keeps the frame on the old sink
         self._pushed_duration += frame.duration
+        await self.next_in_chain.capture_frame(frame)
 
     def flush(self) -> None:
         super().flush()
--- a/livekit-agents/livekit/agents/voice/io.py
+++ b/livekit-agents/livekit/agents/voice/io.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
 import asyncio
+import time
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -142,6 +142,12 @@ def clear_buffer(self) -> None:
 
         if not self._pushed_duration:
             return
+
+        # a detaching sink never reaches the playout wait, so the playhead is reported here
+        self._report_run(
+            offset=self._source_pushed_duration - self._audio_source.queued_duration,
+            ended_at=time.time(),
+        )
```
chenghao-mou, line 150: QQ: should we clear the queued audio here too, in case of a mid-capture swap?

longcw: nice catch. fixed by flush before clear buffer in `set_next_in_chain`

livekit-agents/livekit/agents/voice/io.py
```diff
@@ -368,12 +373,27 @@ def set_next_in_chain(self, new: AudioOutput) -> None:
...
-            old.off("playback_progressed", self._forward_next_playback_progressed)
             if self._pending_playback_count > 0:
-                # stop audio still playing on the old sink
+                # a sink ends a segment when it is flushed, and nothing flushes this one
+                # once it is detached, so a clear alone leaves it open and holding audio
+                if self._capturing:
+                    old.flush()
                 old.clear_buffer()
+            # progress only observes, so the clear is still the sink's last word
+            old.off("playback_progressed", self._forward_next_playback_progressed)
```
chenghao-mou, line 386: Codex reported some less serious issues that appear to be valid:

**\[P2\] The proxy still misses audio played between clear and the asynchronous queue clear — [io.py:385](<https://github.com/livekit/agents/blob/4b2507c2a34b14bd081e7a1cc7a544acd84bb129/livekit-agents/livekit/agents/voice/io.py#L385-L386>)**

The new test correctly shows that `_wait_for_playout()` can report a second run after `clear_buffer()`: queued audio continues playing until the waiter calls `clear_queue()`.

However, the proxy removes its progress listener immediately after `clear_buffer()`. In my repro:

* The sink reported the initial run and another `0.05` seconds.
* The proxy forwarded only the initial run.

Therefore, the recording still omits audio that played during that scheduling window.

**\[P2\]** `_sink_reported` still conflates “played zero” with “cannot report” — [io.py:388](<https://github.com/livekit/agents/blob/4b2507c2a34b14bd081e7a1cc7a544acd84bb129/livekit-agents/livekit/agents/voice/io.py#L388-L395>)

Both original failures remain:

* A Room sink with `_source_pushed_duration == 0` emitted no progress event. The proxy reported the full `0.5` seconds as played.
* After one sink reported two seconds, a remote sink received five seconds and was replaced after `flush()`. The recording retained only the first two seconds.

Event absence cannot identify whether a sink played nothing or cannot report progress. Model that distinction explicitly.

**\[P2\] Replacement still races with an in-flight capture — [io.py:394](<https://github.com/livekit/agents/blob/4b2507c2a34b14bd081e7a1cc7a544acd84bb129/livekit-agents/livekit/agents/voice/io.py#L394-L394>)**

`_offset_base` snapshots `_pushed_duration`, but `capture_frame()` increments that duration only after awaiting the old sink.

If replacement occurs during that await, the in-flight frame is omitted from the base. My two-frame repro still maps the new sink’s frame to offset `0.0` instead of `1.0`.

longcw: the last two should be fixed, the first one requires more changes comparing to what it solved, leave it for now

livekit-agents/livekit/agents/voice/io.py
```diff
@@ -391,6 +422,33 @@ def set_next_in_chain(self, new: AudioOutput) -> None:
         if old is not None and self._pending_playback_count > 0 and not self._capturing:
             self.on_playback_finished(playback_position=self._pushed_duration, interrupted=True)
```
chenghao-mou, line 423: should we also use some kind of assumed played duration here?

Like the

```
min(
        self._pushed_duration - self._offset_base,
        max(0.0, time.time() - self._sink_started_at),
    )
``` used above. we can calculate it once and reuse it here.

chenghao-mou (approved):
lgtm, left one comment above

PR #6975, opened 2026-08-25
test(google): disarm genai client finalizers in realtime tests

### LINKED ISSUE #6881: Flaky unit-tests: a genai client finalizer lands in an unrelated test's leak check

rkfshakti:
Fixes #6881

## Problem

`tests/test_plugin_openai_stt_context.py::test_use_realtime_defaults_to_the_only_transport_a_model_has[gpt-realtime-whisper-True]` intermittently fails with:

```
Failed: Test leaked tasks:
Coroutine : BaseApiClient.aclose    google/genai/_api_client.py:2213
Coroutine : AsyncClient.aclose      google/genai/client.py:152
```

That test is synchronous and never touches google. The tasks come from `tests/test_plugin_google_realtime.py` — the only module in the unit suite that constructs genai clients.

## Root cause

`AsyncClient.__del__` and `BaseApiClient.__del__` both do this unconditionally, with no check for a client that was already closed:

```python
def __del__(self) -> None:
    try:
    except Exception:
      pass
```

`RealtimeSession.aclose()` closes `client.aio`, but that does not disarm the finalizer — the object stays garbage until the collector reaches it, which can be several modules later. `fail_on_leaked_tasks` diffs `asyncio.all_tasks()` around each test, so the task is charged to whatever test happens to be running then.

## Fix

Disarm both finalizers at module import in `tests/test_plugin_google_realtime.py` — the only module that constructs genai clients. Sessions already close their clients explicitly, so the finalizers are pure redundancy.

Verified:
- Without the disarm, dropping a client schedules 2 pending `AsyncClient.aclose` tasks on the running loop
- With the disarm, zero tasks leak
- All 18 realtime tests pass; full unit suite passes (1735 passed)

### DIFF
--- a/tests/test_plugin_google_realtime.py
+++ b/tests/test_plugin_google_realtime.py
@@ -5,7 +5,7 @@ from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 
 import pytest
-from google.genai import types
+from google.genai import _api_client, client as genai_client, types
 
 from livekit.agents import llm, utils
 from livekit.plugins.google.realtime.api_proto import ClientEvents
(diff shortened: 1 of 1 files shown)


rkfshakti:
Closing as duplicate — this exact fix (settle genai client finalizers in the test that owns them) was already merged upstream as #6882 (commit ebf9c9125, merged 2026-08-27). The merged version uses the same autouse `_settle_genai_finalizers` fixture approach. Thanks for the review!

rkfshakti:
Duplicate of merged #6882 — same fix already in main.

PR #6978, opened 2026-08-25
feat(rime): support Coda streaming in LiveKit

naszzz:
## Summary

- add the Rime Coda v1 WebSocket transport and LiveKit streaming adapter
- select Coda streaming behavior by endpoint and support raw or tokenized text input
- document the integration and add unit coverage for request controls and WebSocket behavior

## Interface and data flow

```mermaid
flowchart TB
    subgraph MAIN["Before this PR"]
        M_TTS["rime.TTS"]
        M_MODE{"use_websocket"}
        M_HTTP["HTTP synthesis"]
        M_WS3["Legacy WS3<br/>JSON messages"]

        M_TTS --> M_MODE
        M_MODE -->|"false"| M_HTTP
        M_MODE -->|"true"| M_WS3
    end

    subgraph BRANCH["This PR"]
        TTS["rime.TTS<br/>typed rime_tts_event"]
        SELECT{"Endpoint configuration"}

        TTS --> SELECT

        SELECT -->|"base_url"| LEGACY{"Legacy mode"}
        LEGACY -->|"HTTP URL"| HTTP["HTTP synthesis<br/>existing path"]
        LEGACY -->|"WebSocket URL or<br/>use_websocket=true"| WS3["LegacyWebSocketAdapter<br/>WS3 JSON + aligned timestamps"]

        SELECT -->|"websocket_url"| MODEL{"Model resolution"}

        MODEL -->|"/coda/ws"| CODA["model = coda"]
        MODEL -->|"/mist/ws"| MIST["model = mistv3"]
        MODEL -->|"/ws"| DEDICATED["Dedicated endpoint<br/>explicit model required"]

        CODA --> V1
        MIST --> V1
        DEDICATED --> V1

        V1["WebSocketV1Adapter<br/>connection pool + tokenizer"]
        V1 --> PROTOCOL{"websocket_protocol"}

        PROTOCOL -->|"binary"| BINARY["rime.v1.binary<br/>protobuf frames"]
        PROTOCOL -->|"json"| JSON["rime.v1.json<br/>proto3 JSON"]

        BINARY --> WIRE
        JSON --> WIRE

        WIRE["WebSocket v1 lifecycle<br/>ready → start → text* → end or cancel<br/>audio* → done"]
    end

    subgraph MODEL_STATE["Endpoint and model safety"]
        UPDATE["update_options"]
        UPDATE --> IDENTITY{"Same normalized<br/>model endpoint?"}

        IDENTITY -->|"yes + model changed"| REJECT["Reject update<br/>socket is still bound to old model"]
        IDENTITY -->|"no"| ALLOW["Accept endpoint and model change"]

        UPDATE --> URL{"Full connection URL changed?"}
        URL -->|"yes"| POOL["Replace connection pool"]
        URL -->|"no"| KEEP["Keep connection pool"]

        NORMAL["Model endpoint identity ignores:<br/>query, fragment, trailing slash<br/><br/>It normalizes:<br/>scheme, host, effective port"]
    end

    subgraph STREAM_STATE["Per-stream historical state"]
        CURRENT["TTS._opts.model<br/>current endpoint model"]
        CURRENT --> NEW["Create SynthesizeStream"]
        NEW --> SNAPSHOT["_metrics_model<br/>snapshot for this stream"]
        SNAPSHOT --> METRICS["Metrics keep the correct model<br/>after later TTS updates"]
    end

    V1 -. uses .-> UPDATE
    TTS -. owns .-> CURRENT
```

## Testing

- `uv run pytest tests/test_plugin_rime_tts.py tests/test_plugin_rime_websocket_v1.py`
- 59 passed


livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/_websocket_v1_adapter.py
```diff
@@ -0,0 +1,343 @@
...
+class CodaV1SynthesisOptions:
+    """Coda settings before conversion to the WebSocket v1 wire format."""
+
+    speaker: str
+    language: NotGivenOr[str] = NOT_GIVEN
+    sampling_rate: NotGivenOr[int] = NOT_GIVEN
+    repetition_penalty: NotGivenOr[float] = NOT_GIVEN
+    temperature: NotGivenOr[float] = NOT_GIVEN
+    top_p: NotGivenOr[float] = NOT_GIVEN
+    max_tokens: NotGivenOr[int] = NOT_GIVEN
```
ryanli: These five parameters are no longer used in Coda.

livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
```diff
@@ -143,20 +146,64 @@ def _check_time_scale_factor_supported(
...
         model: NotGivenOr[TTSModels | str] = NOT_GIVEN,
         speaker: NotGivenOr[str] = NOT_GIVEN,
         lang: TTSLangs | str = "eng",
+        repetition_penalty: NotGivenOr[float] = NOT_GIVEN,
+        temperature: NotGivenOr[float] = NOT_GIVEN,
+        top_p: NotGivenOr[float] = NOT_GIVEN,
+        max_tokens: NotGivenOr[int] = NOT_GIVEN,
+        time_scale_factor: NotGivenOr[float] = NOT_GIVEN,
+        speed_alpha: NotGivenOr[float] = NOT_GIVEN,
+        sample_rate: int = 22050,
```
ryanli: I think we should remove the default here — and use the default provided by the TTS service.

livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
```diff
@@ -143,20 +146,64 @@ def _check_time_scale_factor_supported(
...
+        *,
+        websocket_url: str,
+        speaker: str = "astra",
+        lang: TTSLangs | str = "eng",
+        repetition_penalty: NotGivenOr[float] = NOT_GIVEN,
+        temperature: NotGivenOr[float] = NOT_GIVEN,
+        top_p: NotGivenOr[float] = NOT_GIVEN,
+        max_tokens: NotGivenOr[int] = NOT_GIVEN,
+        time_scale_factor: NotGivenOr[float] = NOT_GIVEN,
+        sample_rate: int = 22050,
```
ryanli: I think we should remove the default here — and use the default provided by the TTS service.

ryanli (commented):
Oops, thought I sent the review earlier.

livekit-plugins/livekit-plugins-rime/README.md
```diff
@@ -13,3 +13,69 @@ pip install livekit-plugins-rime
...
+tts = rime.TTS(
+    websocket_url="wss://api.rime.ai/coda/ws",
+    speaker="astra",
+    api_key=os.environ["RIME_API_KEY"],
+)
+```
+
+Pass the active model WebSocket endpoint explicitly. The presence of `websocket_url` selects
+WebSocket v1 streaming. The plugin reads the model from the path segment before `/ws`, such as
+`coda` in `/coda/ws` or `mistv3` in `/mistv3/ws`. Do not pass `model` with `websocket_url`.
```
ryanli: We don't support `/mistv3` yet — and it will be `/mist` instead of `/mistv3`.

livekit-plugins/livekit-plugins-rime/README.md
```diff
@@ -13,3 +13,69 @@ pip install livekit-plugins-rime
...
+You can send more text after `flush()`. The same Rime context remains active, and an input pause
+needs no wire message. Only `end_input()` ends normal input and causes Rime to send `done`.
+Each outgoing `text` value must be a stable sentence-sized unit. Calling `flush()` drains any
+buffered fragment, so call it only after a complete sentence or a stable clause.
+
+The v1 implementation has these limits:
+
+- It requests raw `audio/pcm` data.
+- It does not provide aligned word timestamps.
+- `speed_alpha` is not supported. Use `time_scale_factor` to control speed.
```
ryanli: I'd say this is not a limit but rather the intended behaviour with the new interface.

livekit-plugins/livekit-plugins-rime/README.md
```diff
@@ -13,3 +13,69 @@ pip install livekit-plugins-rime
...
+| `stream.aclose()` | Stop active synthesis. | Send `cancel` when needed. |
+
+You can send more text after `flush()`. The same Rime context remains active, and an input pause
+needs no wire message. Only `end_input()` ends normal input and causes Rime to send `done`.
+Each outgoing `text` value must be a stable sentence-sized unit. Calling `flush()` drains any
+buffered fragment, so call it only after a complete sentence or a stable clause.
+
+The v1 implementation has these limits:
+
+- It requests raw `audio/pcm` data.
```
ryanli: This can be changed to all audio codecs that Rime supports?

livekit-agents/livekit/agents/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "1.7.1"
+__version__ = "1.7.2"
```
tinalenguyen: are the changes here and in tts.py intentional?

livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "1.7.1"
+__version__ = "1.7.5"
```
tinalenguyen: this bump as well

tinalenguyen (commented):
are the tts.py and version.py changes intentional?

chasef07:
The timeout currently starts after the `start` or `cancel` write finishes. If a WebSocket write stalls, synthesis or `aclose()` can hang despite the configured timeout.

I reproduced both with a blocked write: a 30 ms timeout was still pending after 200 ms. Could we include the writes in the timeout and add a regression test?

chasef07:
Optional cleanup: the legacy and v1 adapters duplicate the pool reference counting, retirement, and close-task handling. Could we share that lifecycle code in one small private helper so future fixes only need to happen in one place?

livekit-plugins/livekit-plugins-rime/livekit/plugins/rime/tts.py
```diff
@@ -116,52 +135,115 @@ def _model_params(opts: _TTSOptions) -> dict[str, object]:
...
+            raise ValueError("model is required when websocket_url ends with /ws")
+        return model
+    if is_given(model):
+        raise ValueError("model is derived from websocket_url; omit model")
+    if endpoint_model == "mist":
+        return MODEL_MIST_V3
+    return endpoint_model
+
+
+class TTS(tts.TTS[Literal["rime_tts_event"]]):
```
tinalenguyen: i don't think we need the `rime_tts_event`, it doesn't seem to be used anywhere but in the tests

tests/fixtures/rime_websocket_v1.json
```diff
@@ -0,0 +1,200 @@
+[
```
tinalenguyen: is this file needed? also would be great if we could combine all the tests into one file

livekit-agents/livekit/agents/tts/tts.py
```diff
@@ -717,7 +718,9 @@ def _emit_metrics() -> None:
                 streamed=True,
                 acquire_time=self._acquire_time,
                 connection_reused=self._connection_reused,
-                metadata=Metadata(model_name=self._tts.model, model_provider=self._tts.provider),
+                metadata=Metadata(
```
tinalenguyen, line 725: i think these changes were unintentionally introduced again

naszzz: Intentional: keeps existing Rime streams’ metrics tied to their original model after updates. Now scoped to Rime. Other providers retain prior behavior.

tinalenguyen (approved):
tested and it lgtm, few findings below. also let's scope the changes to the rime plugin and remove the changes to `tts.py`

**1. Default speaker depends on transport instead of model** — `tts.py:296`

```python
model_is_explicit = resolved_model != MODEL_CODA
```

Explicitness is inferred from the resolved model rather than from what the caller passed, so a `websocket_url` ending in `/coda/ws` resolves to `coda`, sets this `False`, and takes the `if not model_is_explicit` branch at `tts.py:334` → `"astra"`, a Mist voice on a Coda endpoint. `DefaultCodaVoice` (`tts.py:338`) is unreachable from the v1 path.

- `TTS(model="coda")` → `lyra`
- `TTS(websocket_url=".../coda/ws", model="coda")` → `astra`

Suggest tracking `model_was_given = is_given(model)`, then picking one default and applying it to both transports.

**2. ws3 is the only path that doesn't declare its sample rate** — `tts.py:422-424`

| Path | Line | Sends rate? |
|---|---|---|
| HTTP | `tts.py:655` | always |
| v1 WS | `tts.py:467` | always |
| ws3 | `tts.py:422-424` | only if `is_given(sample_rate)` |

`main` always sent it on ws3. Now when the caller omits `sample_rate`, the URL carries no `samplingRate` but `_legacy_synthesis_options()` still passes `_default_sample_rate(model)` to `output_emitter.initialize`. If Rime's server default for that model differs, PCM is reinterpreted at the wrong rate — pitch/speed shift, no error. `tests/test_plugin_rime_tts.py:146` locks the omission in, and commit `b9b290062` ("send HTTP sample rate explicitly") rejected this same reasoning for HTTP.

Suggest sending it unconditionally to match the other two paths — or, if the intent is to let the server pick, stop declaring a rate to the emitter and read it from the stream.

Minor, same area: v1 over-declares in the opposite direction (`tts.py:467` always sends), so an explicit `audio_format="audio/pcmu"` yields a 24000 Hz G.711 request; `validate_audio_format` (`_websocket_v1.py:93`) only checks set membership, not rate compatibility.

PR #6985, opened 2026-08-26
feat: surface agent backchannel opportunity as a public event

AinaLuc:
## Summary
- Promotes the previously-internal `_AgentBackchannelOpportunityEvent` to a public `AgentBackchannelOpportunityEvent`, emitted on `AgentSession` (`agent_backchannel_opportunity`) and dispatched to a new `Agent.on_backchannel_opportunity()` hook, so agents can react to mid-turn pauses with short acknowledgment phrases ("mm-hmm", "right", etc).
- Adds an `end_of_turn_margin` property to the event as a risk axis for picking an appropriate phrase.
- Fixes a race in `AudioRecognition.update_vad()`: the previous VAD task was cancelled *after* the new one was already created, so a rapid `update_vad` call could let a stale task's cleanup clobber the new stream/channel. Adds a generation counter so a superseded VAD task's events and cleanup are ignored.

## Test plan
- [ ] `make check` (format, lint, type-check)
- [ ] `uv run pytest --unit`
- [ ] Manual test: agent session with turn detection enabled, verify `on_backchannel_opportunity` fires and `update_vad` swap doesn't leave the session in a stuck speaking state

### DIFF
--- a/livekit-agents/livekit/agents/__init__.py
+++ b/livekit-agents/livekit/agents/__init__.py
@@ -79,6 +79,7 @@ from .types import (
 from .version import __version__
 from .voice import (
     Agent,
+    AgentBackchannelOpportunityEvent,
     AgentEvent,
     AgentFalseInterruptionEvent,
     AgentSession,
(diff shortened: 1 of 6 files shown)


davidzhao:
this feature is not ready yet.. we'll expose it along with docs when it is

Issue #5117, opened 2026-03-16
Google TTS Plugin Timeout with Gemini Model & Chirp_3 Streaming Error (livekit-plugins-google==1.4.4)


sahil-axl:
### Bug Description

Hi Team,

I'm facing an issue with the Google TTS plugin while using LiveKit Agents.

I am currently using:

livekit-plugins-google==1.4.4
Issue 1 – chirp_3 Model Not Working with Streaming

When I try to use the chirp_3 model for TTS, I receive the following error:

{"message": "failed to synthesize speech: Streaming for this Gemini-TTS model is currently not supported. Please use another Gemini-TTS model or Chirp3: HD voices for streaming., retrying in 0.1s",
"level": "WARNING",
"name": "livekit.agents",
"tts": "livekit.plugins.google.tts.TTS",
"attempt": 1,
"streamed": true,
"timestamp": "2026-03-13T10:47:24.540323+00:00"}

From the message, it appears that streaming is not supported for this model. However, the error message itself suggests using Chirp3 HD voices for streaming, which is confusing because the chirp_3 model fails when used.

Issue 2 – Gemini Default Model Random Timeout

If I switch to the default Gemini TTS model, the agent initially starts speaking correctly. However, after some time, it randomly fails with a timeout error:

{"message": "failed to synthesize speech, retrying in 0.1s
Traceback (most recent call last):
File "/app/.venv/lib/python3.12/site-packages/livekit/agents/tts/tts.py", line 456, in _main_task
await self._run(output_emitter)
File "/app/.venv/lib/python3.12/site-packages/livekit/plugins/google/tts.py", line 398, in _run
await asyncio.gather(*tasks)
File "/app/.venv/lib/python3.12/site-packages/livekit/plugins/google/tts.py", line 391, in _run_segments
await self._run_stream(input_stream, output_emitter, streaming_config)
File "/app/.venv/lib/python3.12/site-packages/livekit/plugins/google/tts.py", line 443, in _run_stream
raise APITimeoutError() from None
livekit.agents._exceptions.APITimeoutError: Request timed out. (body=None, retryable=True)",
"level": "WARNING",
"name": "livekit.agents",
"tts": "livekit.plugins.google.tts.TTS",
"attempt": 1,
"streamed": true
}

### Expected Behavior

the agent should speak seamlessly without getting any timeout issues with default gemini model and chirp_3 model shoulf also work. Please correct me if i'm wrong

### Reproduction Steps

```bash
# -------------------------
[...]

Issue #2951, opened 2025-07-20
Google Cloud TTS isn't working while running


sherlockholmes211:
Getting the below error when I am trying to run it in the console 

```
 File "/Users/vishnuteja/Documents/WorkSpace/ava/venv/lib/python3.12/site-packages/livekit/plugins/google/tts.py", line 325, in _run_segments
    await self._run_stream(input_stream, output_emitter, streaming_config)
  File "/Users/vishnuteja/Documents/WorkSpace/ava/venv/lib/python3.12/site-packages/livekit/plugins/google/tts.py", line 375, in _run_stream
    await input_gen.aclose()
RuntimeError: aclose(): asynchronous generator is already running
```

I am using the below versions
python  - 3.12
```
livekit                                  1.0.12
livekit-agents                           1.2.1
livekit-api                              1.0.3
livekit-blingfire                        1.0.0
livekit-plugins-assemblyai               1.2.1
livekit-plugins-google                   1.2.1
livekit-plugins-noise-cancellation       0.2.5
livekit-plugins-openai                   1.2.1
livekit-plugins-silero                   1.2.1
livekit-plugins-turn-detector            1.2.1
livekit-protocol                         1.0.4
```

longcw:
can you share a reproducible example or the agent session configuration?

nidhi-canopas:
I get similar error.

It happens randomly, not always.

For details, please have a look at below images.

<img width="783" height="417" alt="Image" src="https://github.com/user-attachments/assets/a8d53a18-2026-492c-b4f1-757262689dd0" />

<img width="779" height="320" alt="Image" src="https://github.com/user-attachments/assets/d1908126-b259-49cf-bdec-67d97c046eb3" />

Here,s is the code for **tts_node**.

```
async def tts_node(self, text: AsyncIterable[str], model_settings: ModelSettings):
       # Handle streaming text
       full_message_parts = []

        # Define an async generator to stream the collected parts for TTS processing *and* collect them
        async def stream_and_collect_generator():
             async for chunk in text:  # text is the original LLM output stream
                    full_message_parts.append(chunk)
                    yield chunk

        # Process the stream for TTS first using the new generator
         processed_tts_stream = Agent.default.tts_node(
                self,
                process_stream(
                    stream_and_collect_generator(), self.ctx, self.participant
                ),
                model_settings,
            )

         # Yield from the processed TTS stream
         async for frame in processed_tts_stream:
               yield frame

          print(full_message_parts)
```

[...]

PR #6986, opened 2026-08-26
fix(google): don't cut streaming TTS off at the connect timeout

### LINKED ISSUE #3347: Google TTS (LiveKit Plugin) stops mid-way while LLM continues generating text
Description

I am using LiveKit with the Google TTS plugin to build a workflow where:

The LLM generates text continuously.

The TTS speaks out the LLM output in real-time.

The problem is that Google TTS stops producing audio in the middle, even though the LLM continues generating the rest of the text. At the end, I receive:

Full text output from the LLM

Only partial audio output from Google TTS

So, TTS does not complete speaking the entire LLM response.

please explain and address this issue

biztex:
**Problem**

`SynthesizeStream._run_stream` passes `conn_options.timeout` (10 s by default, documented as the *connect* timeout) as `timeout=` of `streaming_synthesize()`. For a gapic streaming method that is the gRPC deadline of the whole bidi call (`google.api_core` → `grpc.aio.StreamStreamMultiCallable(timeout=...)`, "duration of time in seconds to allow for the RPC"), and the request generator keeps the call open for as long as the LLM produces text for the segment. Any reply that takes longer than 10 s to generate and synthesize is ended with `DeadlineExceeded` → `APITimeoutError` mid-utterance, and because audio was already emitted the base class doesn't retry ("TTS failed after partial audio was already sent to the user, skip retrying") — the rest of the turn is dropped. `use_streaming=True` is the default, so this hits every long reply.

The deadline has a second effect: grpc only cancels its request-poller task in `cancel()`, never when the deadline fires. So when the LLM is still producing text at that moment, the poller is mid-`__anext__` on `input_gen` and the `finally: await input_gen.aclose()` fails with `RuntimeError: aclose(): asynchronous generator is already running` (the traceback in #2951), leaking the poller.

**Fix**

- Open the RPC without a deadline (like `stt.py`'s `streaming_recognize`), and use `conn_options.timeout` only to bound the wait for the first audio chunk (`asyncio.wait_for` on the first `anext`), still surfacing it as `APITimeoutError`. Subsequent chunks are unbounded, since the gaps between them follow the LLM's pace.
- Let grpc own the request generator's lifetime: it finalizes it when the call completes or is cancelled (which is what a first-chunk timeout does), so the explicit `aclose()` that raced with grpc's poller is gone.

`ChunkedStream` is untouched: there the kwarg is a per-request deadline for a single unary `synthesize_speech`, which is what it should be.

**Tests**

`tests/test_plugin_google_tts.py` drives `tts.stream()` against a fake client that mirrors grpc.aio (request iterator consumed by a background task, `timeout` enforced as a whole-call deadline, cancelling a read cancels the RPC and its poller):
- `test_streaming_not_cut_off_by_connect_timeout`: three sentences pushed over ~0.45 s with `timeout=0.2` — every sentence reaches the fake and every audio byte is delivered. On `main` it fails with the `aclose()` RuntimeError above and a leaked poller task.
- `test_streaming_first_response_timeout`: a server that never answers while the input is still open raises `APITimeoutError` within the timeout, with no leaked tasks.

ruff / mypy (strict) clean.

Fixes #3347
Fixes the timeout part of #5117 (its chirp_3 voice-name error is a separate config/model-support question).
Also fixes the `aclose(): asynchronous generator is already running` failure reported in #2951.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
@@ -451,15 +451,27 @@ class SynthesizeStream(tts.SynthesizeStream):
             except Exception:
                 logger.exception("an error occurred while streaming input to google TTS")
 
-        input_gen = input_generator()
+        # grpc.aio consumes the request generator from its own task and finalizes it once the RPC
+        # completes or is cancelled; closing it from here would race with that task ("aclose():
+        # asynchronous generator is already running") when the RPC ends while the LLM is still
+        # producing text.
         try:
-            stream = await self._tts._ensure_client().streaming_synthesize(
-                input_gen, timeout=self._conn_options.timeout
-            )
+            # no RPC deadline: the call stays open for as long as the LLM keeps producing text,
+            # the connect timeout only bounds the wait for the first audio chunk
+            stream = await self._tts._ensure_client().streaming_synthesize(input_generator())
             output_emitter.start_segment(segment_id=utils.shortuuid())
 
-            async for resp in stream:
+            resp_iter = aiter(stream)
+            try:
+                resp = await asyncio.wait_for(
+                    anext(resp_iter, None), timeout=self._conn_options.timeout
+                )
+            except asyncio.TimeoutError:
+                raise APITimeoutError() from None
+
+            while resp is not None:
                 output_emitter.push(resp.audio_content)
+                resp = await anext(resp_iter, None)
 
             output_emitter.end_segment()
 
@@ -467,8 +479,6 @@ class SynthesizeStream(tts.SynthesizeStream):
             raise APITimeoutError() from None
         except GoogleAPICallError as e:
             raise APIStatusError(e.message, status_code=e.code or -1, body=f"{e.details}") from e
-        finally:
-            await input_gen.aclose()
 
 
 def _gender_from_str(gender: str) -> SsmlVoiceGender:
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -451,24 +451,34 @@ async def input_generator() -> AsyncGenerator[
...
+            try:
+                resp = await asyncio.wait_for(
+                    anext(resp_iter, None), timeout=self._conn_options.timeout
+                )
+            except asyncio.TimeoutError:
+                raise APITimeoutError() from None
+
+            while resp is not None:
                 output_emitter.push(resp.audio_content)
+                resp = await anext(resp_iter, None)
```
longcw: After the first chunk, `resp = await anext(resp_iter, None)` has no timeout.  a server that answers once and then goes quiet hangs the segment forever with no error

livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
```diff
@@ -451,24 +451,34 @@ async def input_generator() -> AsyncGenerator[
...
-            async for resp in stream:
+            resp_iter = aiter(stream)
+            try:
+                resp = await asyncio.wait_for(
+                    anext(resp_iter, None), timeout=self._conn_options.timeout
+                )
+            except asyncio.TimeoutError:
+                raise APITimeoutError() from None
+
+            while resp is not None:
```
longcw, line 472: nit: maybe pushing the first chunk and then `async for resp in resp_iter`?

longcw (approved):

PR #6991, opened 2026-08-26
(gemini stt): add gemini-3.5-transcribe-live to inference and plugin

tinalenguyen:


livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,352 @@
...
+                            )
+                        )
+        except APIError as e:
+            if getattr(e, "code", None) == 1000 or "1000" in str(e):
+                logger.debug(f"Gemini ASR session closed normally: {e}")
+            else:
+                logger.warning(f"Gemini ASR receive error: {e}")
+                raise
+        except Exception as e:
+            logger.debug(f"Gemini ASR receive loop ended: {e}")
```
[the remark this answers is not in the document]
tinalenguyen: Fixed in 755d2228d — the three log calls now use static messages with the exception text in an `lk.pii.error` attribute, matching `llm/utils.py:623`.

Left `raise APIConnectionError(f"... {e}") from e` as-is: that's the established pattern across plugins (35 other call sites) and isn't a log/span/event body.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,352 @@
...
+                is_enterprise = False
+            elif creds or project_id or self._opts.location:
+                is_enterprise = True
+
+        client_kwargs: dict[str, Any] = {
+            "api_key": self._api_key,
+            "http_options": self._http_options,
+        }
+        if is_enterprise:
+            client_kwargs["enterprise"] = True
```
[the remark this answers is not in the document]
tinalenguyen: I don't think this one holds — `enterprise` is the current kwarg on `google.genai.Client`, and `vertexai` is the legacy alias.

From the `google-genai` version this repo pins in `uv.lock` (2.11.0), `client.py` `class Client` `__init__`:

```
enterprise: Optional[bool] = None,
vertexai: Optional[bool] = None,
...
  enterprise (bool): Indicates whether the client should use the Gemini
    Enterprise API ...
  vertexai (bool): Legacy flag for `enterprise`.
...
resolved_vertexai = enterprise if enterprise is not None else vertexai
```

Confirmed against both 2.11.0 (pinned) and 2.19.0 (installed) — `inspect.signature(Client.__init__)` lists `enterprise` in both, so `Client(**client_kwargs)` doesn't raise `TypeError` and `_api_client.vertexai` does get set, so language hints are applied.

`llm.py` uses `vertexai=` because it predates the rename; both work. Leaving as-is.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,395 @@
...
+        except (ClientError, ServerError, APIError) as e:
+            raise APIStatusError(
+                message=str(e),
+                status_code=getattr(e, "code", -1),
+                body=None,
+            ) from e
+        except asyncio.CancelledError:
+            raise
+        except Exception as e:
+            raise APIConnectionError(f"Gemini STT connection failed: {e}") from e
```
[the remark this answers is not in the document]
tinalenguyen: Good catch, this one is real — I'd waved it off on the earlier thread as "not a log body", but it does become one: `stt/stt.py:490` interpolates the exception straight into its retry log.

```python
logger.warning(
    f"failed to recognize speech: {e}, retrying in {retry_interval}s", ...
)
```

Fixed in b76d448f9. Both raises now carry a static message with only the exception type, the detail goes to `lk.pii.error`, and they use `from None` so the cause chain doesn't carry the text back in either. `status_code` is still passed through, so `APIStatusError` resolves `retryable` exactly as before.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,406 @@
...
+)
+from livekit.agents.types import (
+    NOT_GIVEN,
+    NotGivenOr,
+)
+from livekit.agents.utils import is_given
+
+from ..log import logger
+
+DEFAULT_MODEL = "gemini-2.5-flash"
```
chenghao-mou: I don't think this model supports Live endpoint. We should just use the 3.5 model: https://ai.google.dev/gemini-api/docs/models/gemini-2.5-flash

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,406 @@
...
+                        lang = LanguageCode(code)
+
+                    if text := transcription.text:
+                        if not committed:
+                            # Gemini opens a transcript with a leading space
+                            text = text.lstrip()
+                        committed += text
+                        self._event_ch.send_nowait(
+                            self._speech_event(
+                                stt.SpeechEventType.INTERIM_TRANSCRIPT, committed, lang
```
chenghao-mou: we don't have to emit this as interim if they already have `interim_input_transcription`

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,406 @@
+# Copyright 2023 LiveKit, Inc.
```
chenghao-mou, line 1: we might need to bump the genai version to support the new features. from Codex:

> package metadata permits `google-genai==1.68.0`, but this code requires newer SDK fields. Version 1.68 rejects `enterprise=`, `language_hints`, and `language_auto`. Those transcription fields arrived in 2.9. Raise the minimum to at least `google-genai>=2.9`.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,406 @@
...
+        language: LanguageCode | str | None = "en-US",
+        language_hints: list[str] | None = None,
+        sample_rate: int = DEFAULT_SAMPLE_RATE,
+        api_key: NotGivenOr[str] = NOT_GIVEN,
+        vertexai: NotGivenOr[bool] = NOT_GIVEN,
+        credentials: Any | None = None,
+        credentials_path: str | None = None,
+        project: NotGivenOr[str] = NOT_GIVEN,
+        location: NotGivenOr[str] = NOT_GIVEN,
+        http_options: Any | None = None,
```
chenghao-mou, line 92: should we support the adaptation_phrases here too?

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,431 @@
...
+# after the input channel closes. Bounded so teardown can't hang on a silent session.
+FINALIZE_TIMEOUT = 2.0
+
+
+@dataclass
+class _STTOptions:
+    model: str
+    language: LanguageCode | str | None
+    language_hints: list[str] | None
+    adaptation_phrases: list[str] | None
```
chenghao-mou: I think they renamed these two in newer versions:

Image

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,431 @@
...
+                        self._event_ch.send_nowait(
+                            self._speech_event(stt.SpeechEventType.INTERIM_TRANSCRIPT, text, lang)
+                        )
+
+                if (
+                    getattr(transcription, "finished", False)
+                    or getattr(sc, "turn_complete", False)
+                    or getattr(sc, "generation_complete", False)
+                ):
+                    flush()
```
chenghao-mou: looking at the documentation, which isn't super clear, it looks like we can emit finals whenever we have input_transcription:

> input_transcription: the finalized transcript emitted when the speaker pauses, the turn completes, or speech is finalized. Once emitted, this text represents the model's authoritative transcription of that speech segment. In smart transcription mode, this will include the cleaned, formatted response.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
```diff
@@ -0,0 +1,431 @@
...
+        )
+
+        config = types.LiveConnectConfig(
+            response_modalities=["TEXT"],
+            input_audio_transcription=input_audio_transcription,
+        )
+
+        try:
+            async with client.aio.live.connect(model=self._opts.model, config=config) as session:
+                send_task = asyncio.create_task(self._send_loop(session))
```
chenghao-mou, line 274: I think we need to handle reconnection/away as well:

> Session duration: Live transcription sessions support continuous streaming for up to 10 minutes.

chenghao-mou: we reconnected automatically with retry at a higher level, so not a huge issue: 


```
2026-08-26 21:05:05,668 - DEBUG livekit.agents - conversation_item_added {"role": "assistant", "lk.pii.text": "Glad you think so! Anything else you want to explore or chat about?", "room": "console-5e9c3e0f", "pid": 23385, "job_id": "AJ_FPrqgSDiUVnc"}
2026-08-26 21:05:14,019 - WARNING livekit.plugins.google - Gemini ASR receive error {"lk.pii.error": "1008 None. Connection aborted because the client failed to close the connection after receiving a GoAway signal once the session durat", "room": "console-5e9c3e0f", "pid": 23385, "job_id": "AJ_FPrqgSDiUVnc"}
2026-08-26 21:05:14,021 - WARNING livekit.plugins.google - Gemini STT connection failed {"error_type": "ConnectionClosedError", "lk.pii.error": "received 1008 (policy violation) Connection aborted because the client failed to close the connection after receiving a GoAway signal once the session durat; then sent 1008 (policy violation) Connection aborted because the client failed to close the connection after receiving a GoAway signal once the session durat", "room": "console-5e9c3e0f", "pid": 23385, "job_id": "AJ_FPrqgSDiUVnc"}
2026-08-26 21:05:14,022 - WARNING livekit.agents - failed to recognize speech: Gemini STT connection failed (ConnectionClosedError), retrying in 0.1s {"stt": "livekit.plugins.google.beta.gemini_stt.STT", "attempt": 0, "streamed": true, "room": "console-5e9c3e0f", "pid": 23385, "job_id": "AJ_FPrqgSDiUVnc"}
````

chenghao-mou (approved):
LGTM. make sure you update the uv lock file.

livekit-plugins/livekit-plugins-google/livekit/plugins/google/beta/gemini_stt.py
sbichenko: Does that lead to a 9-second pause in STT's activity on 10th minute?

chenghao-mou: No. It still emits events until the actual closure. Then reconnection happens 0.1s after that.

PR #6913, opened 2026-08-19
fix: prevent wait_for_idle cancellation from poisoning user turns

Excelius-Wang:
Fixes #6911.

## Summary

- Shield the end-of-turn and user-turn-completion tasks observed by `wait_for_idle()`.
- Wait for a predecessor user-turn task without propagating cancellation in either direction.
- Preserve non-cancellation exceptions raised by the predecessor.
- Add regression tests for cancellation isolation and successor recovery.

## Why

`wait_for_idle()` is an observer of SDK-owned tasks. Cancelling the observer currently cancels the task it is awaiting. A later user turn then awaits that cancelled predecessor and is cancelled before processing, so one observer cancellation can poison subsequent turns.

## Testing

- `pytest tests/test_wait_for_idle_cancellation.py -q` — 3 passed
- `pytest tests/test_tools.py --unit -q` — 114 passed
- `ruff format --check livekit-agents/livekit/agents/voice/agent_activity.py tests/test_wait_for_idle_cancellation.py`
- `ruff check livekit-agents/livekit/agents/voice/agent_activity.py tests/test_wait_for_idle_cancellation.py`
- `mypy -p livekit.agents.voice.agent_activity`

PR #7011, opened 2026-08-27
fix: keep VAD from cancelling a manually committed user turn

### LINKED ISSUE #7010: Manual mode: VAD START_OF_SPEECH cancels a committed user turn and strands _user_turn_committed
`commit_user_turn()` resolves after scheduling the end-of-turn bounce, which sleeps the endpointing delay before delivering the turn. With `turn_detection="manual"`, a VAD `START_OF_SPEECH` inside that window cancels the pending task unconditionally (`audio_recognition.py`, SOS handler). Audio tail around a push-to-talk release (last syllable, breath, button noise) is enough to raise it.

Two problems compound:

1. The cancel is silent. No reply is generated and nothing is logged. In manual mode the eou task always belongs to a turn the application already committed, so the VAD is overriding an explicit commit.
2. The cancelled bounce skips its turn-scoped cleanup, so `_user_turn_committed` stays set and `_on_stt_event` discards every later transcript until the next commit flushes the stale buffer.

Observed in production (push-to-talk voice app, default injected Silero VAD): a user asked a question, got silence, repeated it; the repeat's audio was also discarded, and the eventual reply was generated from the stale buffer of the first attempt.

Repro: manual mode, commit a turn, deliver a VAD `START_OF_SPEECH` during the endpointing sleep. The turn never delivers and later finals are dropped. A failing test is included in the linked PR.

Same cancellation-poisoning class as #6913 and #6897. PR with fix and regression tests: https://github.com/livekit/agents/pull/7011.

vignesh-koruverse:
Fixes #7010.

## Summary

- In manual turn detection, VAD `START_OF_SPEECH` no longer cancels the pending end-of-turn task. An eou task in manual mode always belongs to a turn the application already committed.
- A cancelled eou task now resets the turn-scoped flags via a done callback, unless a newer task superseded it. Stranded, `_user_turn_committed` makes `_on_stt_event` discard every later transcript until the next commit.
- Add regression tests for both behaviors and for supersession safety.

## Why

`commit_user_turn()` resolves after scheduling the end-of-turn bounce, which sleeps the endpointing delay before delivery. Audio tail around a push-to-talk release can raise `START_OF_SPEECH` inside that window. The cancel is silent: no reply, no log, and the skipped cleanup leaves the session discarding all later transcripts. Same cancellation-poisoning class as #6913 and #6897.

## Testing

- `pytest tests/test_audio_recognition_turn_detection.py --audio_eot -q`: the two behavior tests fail without the fix and pass with it
- `ruff check` and `ruff format --check` on both changed files
- `mypy -p livekit.agents.voice.audio_recognition`: clean

### DIFF
--- a/livekit-agents/livekit/agents/voice/audio_recognition.py
+++ b/livekit-agents/livekit/agents/voice/audio_recognition.py
@@ -1362,7 +1362,11 @@ class AudioRecognition:
             self._turn_detector_prediction_fut = None
             self._turn_detector_flushed = False
 
-            if self._end_of_turn_task is not None:
+            # a manually committed turn is authoritative: VAD activity must not
+            # cancel its pending end-of-turn
+            if self._end_of_turn_task is not None and not (
+                self._turn_detection_mode == "manual" and self._user_turn_committed
+            ):
                 self._end_of_turn_task.cancel()
 
             if self._session.amd is not None:
@@ -1756,6 +1760,16 @@ class AudioRecognition:
                 self._user_turn_start,
             )
         )
+        self._end_of_turn_task.add_done_callback(self._on_eou_task_done)
+
+    def _on_eou_task_done(self, task: asyncio.Task[None]) -> None:
+        # a cancelled eou task skips the turn-scoped resets at the end of
+        # _bounce_eou_task; left set, _user_turn_committed makes _on_stt_event
+        # discard every later transcript. skip if a newer task took over.
+        if task.cancelled() and self._end_of_turn_task is task:
+            self._turn_backchannel_over_agent = False
+            self._overlap_in_current_turn = False
+            self._user_turn_committed = False
 
     def _check_user_turn_limit(self, transcript: str) -> None:
         """Check if the user turn exceeds configured limits.
--- a/tests/test_audio_recognition_turn_detection.py
+++ b/tests/test_audio_recognition_turn_detection.py
@@ -235,6 +235,77 @@ def _end_of_speech() -> vad.VADEvent:
     )
 
 
+class TestManualCommitSurvivesVadActivity:
+    """In ``turn_detection="manual"`` the application owns turn boundaries: an
+    eou task only exists for a turn the app already committed, so VAD activity
+    (a breath or click in the audio tail after commit) must neither cancel it
+    nor, when some other site cancels it, leave ``_user_turn_committed`` set —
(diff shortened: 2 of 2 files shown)


Rehansanjay (commented):
Not a maintainer — reviewing as another contributor because this sat twelve days and the class of bug is one I have been working through elsewhere in the repo (#7139 and the plugin fixes around it). I checked both halves against `main` at `2cc38f4`.

**The done-callback mirrors the normal path exactly.** `_bounce_eou_task` ends with

```python
# reset turn-scoped barge-in state once per logical turn (commit or drop)
self._turn_backchannel_over_agent = False
self._overlap_in_current_turn = False
self._user_turn_committed = False
```

and `_on_eou_task_done` resets those three and nothing else. So the cancel path now converges on the same state as the commit path, which is the right shape — a cancelled cleanup should not be a *different* cleanup.

**The identity guard holds for the supersede case, and for a reason worth stating.** `_run_eou_detection` cancels the previous task and reassigns `self._end_of_turn_task` on the next statement. `add_done_callback` defers to a later loop iteration, so by the time the old task's callback runs the attribute already points at the new task and `self._end_of_turn_task is task` is False. It is correct, but it rests on that ordering rather than on anything explicit — `test_superseded_bounce_leaves_the_new_commit_intact` is doing real work here and is the test I would keep if any of them were trimmed.

**One question on the first half.** Skipping the cancel leaves the rest of the SOS handler running: `_speaking = True`, `_vad_speech_started = True`, `_speech_start_time` set, transcription timeout cancelled, turn-detector inference cancelled. The surviving bounce then reaches

```python
if self._last_speaking_time == last_speaking_time:
    self._speech_start_time = None
    self._vad_speech_started = False
    self._last_speaking_time = None
```

and SOS does not touch `_last_speaking_time` — only `INFERENCE_DONE` does, and only when `raw_accumulated_speech > 0`. So on a bare SOS with no inference behind it, that branch still fires and clears `_vad_speech_started` while `_speaking` stays `True`. It looks benign, since the next SOS re-sets `_speech_start_time` through `if not self._vad_speech_started` and EOS clears `_speaking` — but it is a state pair that the pre-cancel behaviour never produced. Was that considered, or is it worth having the guard cover the turn-detector cancellation immediately above it as well?

Nothing blocking from me. The diagnosis in #7010 matches what the code does, the tests fail without the fix, and the third one pins the subtle case rather than the obvious one.

PR #7014, opened 2026-08-27
fix(gradium): update parameters

timpratim:
- Remove "LLM" from __all__: the plugin exports no LLM, so 'from livekit.plugins.gradium import *' raised AttributeError.
- Make GRADIUM_MODEL_ENDPOINT effective for TTS and STT: the parameter's non-None default meant the documented env var was never consulted. Default endpoints are unchanged when neither is set.
- Allow switching voices via update_options(voice_id=...): the existing voice parameter is a fallback that the voice_id default always shadowed.
- Change the default TTS voice to 4SZHfMpw-p46Ywgs (Harper), a voice from the public flagship catalog, matching the published LiveKit docs.


tinalenguyen (approved):
thank you for the PR!

PR #7024, opened 2026-08-28
fix(openai): handle incomplete Responses events

### LINKED ISSUE #7013: OpenAI Responses WebSocket hangs after response.incomplete terminal event
### Bug description

`livekit.plugins.openai.responses.LLM` can hang indefinitely when the OpenAI Responses WebSocket ends a request with the documented terminal event `response.incomplete`.

`_ResponsesWebsocket.generate_response()` currently yields that event but only returns for:

```python
["response.completed", "response.failed", "error"]
```

`response.incomplete` is also absent from `_parse_ws_event()`. The event is therefore ignored by the parser, and `generate_response()` waits for another WebSocket frame even though OpenAI has already ended the response. No further frame arrives, so the agent remains stuck.

This is present in `livekit-plugins-openai==1.6.2` and still appears in the current `main` implementation:

https://github.com/livekit/agents/blob/ebf9c9125d1b870a52b1ce1b2b27304acf74a03d/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py#L157-L161

### Reproduction

We encountered this with a function tool whose JSON Schema contains a first-character negative lookahead:

```json
{
  "type": "function",
  "name": "append_sheet_row",
  "description": "Append a row to a spreadsheet",
  "parameters": {
    "type": "object",
    "properties": {
      "row": {
        "type": "array",
        "items": {
          "type": "string",
          "pattern": "^(?![=+\\-@\\\\]).*"
        }
      }
    },
    "required": ["row"]
  }
}
```

For the same minimal request:

1. Direct OpenAI Responses HTTP completes.
2. Direct OpenAI Responses WebSocket emits `response.incomplete` with `incomplete_details.reason=max_output_tokens`, then emits no further event.
3. LiveKit yields and ignores `response.incomplete`, then remains blocked waiting for the next frame.
4. Replacing only the pattern with the semantically equivalent `^$|^[^=+\-@\\].*` makes the WebSocket response complete normally.

The schema-specific OpenAI behavior is a separate issue, but a valid terminal event should not make the LiveKit client hang regardless of what caused it.

### Expected behavior

`response.incomplete` should terminate the WebSocket response loop. LiveKit should surface an explicit failure containing `response.incomplete_details.reason` (or otherwise return control to the caller) rather than waiting indefinitely.

At minimum, the terminal-event list should include `response.incomplete`. Ideally `_parse_ws_event()` should also parse it so callers receive the provider's reason.

### Environment

- `livekit-agents==1.6.2`
- `livekit-plugins-openai==1.6.2`
- Python 3.11
- Reproduced through the default OpenAI Responses WebSocket transport

### Related issue

https://github.com/livekit/agents/issues/5092 covers a different Responses WebSocket error-handling path, but similarly demonstrates that provider errors can be masked or leave an agent unresponsive.

dawNotPoi:
## Summary

- treat `response.incomplete` as a terminal Responses WebSocket event
- parse the SDK event and surface its reason as a non-retryable provider error for both HTTP and WebSocket streams
- add hermetic regression coverage for transport termination, retry behavior, and response-context preservation

Fixes #7013

## Testing

- `uv run pytest tests/test_plugin_openai_responses.py --plugin openai -q`
- `uv run ruff format --check livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py tests/test_plugin_openai_responses.py`
- `uv run ruff check livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py tests/test_plugin_openai_responses.py`
- `uv run mypy --platform linux -p livekit.plugins.openai`
- `git diff --check`

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/responses/llm.py
@@ -32,6 +32,7 @@ from openai.types.responses import (
     ResponseCreatedEvent,
     ResponseErrorEvent,
     ResponseFailedEvent,
+    ResponseIncompleteEvent,
     ResponseInputParam,
     ResponseOutputItemDoneEvent,
     ResponseOutputMessage,
@@ -156,7 +157,12 @@ class _ResponsesWebsocket:
 
                 event = json.loads(raw_msg.data)
                 yield event
-                if event["type"] in ["response.completed", "response.failed", "error"]:
+                if event["type"] in [
+                    "response.completed",
+                    "response.failed",
+                    "response.incomplete",
+                    "error",
+                ]:
                     completed = True
                     return
         finally:
@@ -567,6 +573,8 @@ class LLMStream(llm.LLMStream):
             return ResponseCompletedEvent.model_validate(event)
         elif event_type == "response.failed":
             return ResponseFailedEvent.model_validate(event)
+        elif event_type == "response.incomplete":
+            return ResponseIncompleteEvent.model_validate(event)
         return None
 
     def _process_event(self, event: ResponseStreamEvent | None) -> llm.ChatChunk | None:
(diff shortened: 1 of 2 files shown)


gabriele-dialpad:
Thanks for fixing this!

chenghao-mou (approved):
thanks for the PR!

PR #7028, opened 2026-08-28
Migrate Speechmatics STT plugin to Agent STT

Smiljanic19A:
Moves `livekit-plugins/livekit-plugins-speechmatics` off the legacy `speechmatics-voice` SDK onto
`speechmatics-agent-stt`, so the plugin works against the Speechmatics **Agent STT** endpoint
(`/v2/agent`). Trims the config to what Agent STT accepts, validates the values it cannot send,
surfaces errors that were silently swallowed, keeps the previous public API importable behind
deprecation warnings, and adds the first test suite.

---

## Breaking

**The endpoint changes, and there is no fallback.**

The plugin now speaks only the Agent STT protocol — a singular `segment`, a top-level `turn_config`,
`StartOfTurn`/`EndOfTurn` — where plain `/v2` sends `AddTranscript`. The default `base_url` moves
from `wss://eu2.rt.speechmatics.com/v2` to `wss://eu2.rt.speechmatics.com/v2/agent`: same host and
region, different service. Pointing `base_url` or `SPEECHMATICS_RT_URL` back at `/v2` does not
restore the old behaviour — the session connects and then returns no transcripts, so the plugin
warns at construction when the resolved URL is not an Agent STT endpoint. **Anyone who set
`SPEECHMATICS_RT_URL` keeps that value across the upgrade**, which is the case most likely to look
like a hang rather than a misconfiguration.

Everything else:

- **Speaker focus is gone** — `focus_speakers`, `ignore_speakers`, `focus_mode`, `SpeakerFocusMode`,
  `update_speakers()`. Agent STT has no equivalent: a segment carries a speaker label and nothing
  routes on it. All of it still resolves and warns (see Back-compatibility), but the feature does
  nothing. Diarization is unaffected.
- **Config fields Agent STT does not accept are ignored:** `max_delay`, `punctuation_overrides`,
  `end_of_utterance_silence_trigger`, `end_of_utterance_max_delay`, `end_of_utterance_mode`,
  `end_of_turn_config`, `vad_config`, `speaker_passive_format`, and the pre-existing legacy
  `audio_settings`, `chunk_size`, `http_session`, `transcription_config`. Accepted so construction
  does not break, warned with the reason, and dropped before the config or the wire.
- **`speaker_active_format` is renamed `speaker_format`.** The old name paired with
  `speaker_passive_format` under speaker focus; with no passive case, "active" distinguished
  nothing. Old name still accepted, with a warning.
- **`sample_rate` must be `16000` and `audio_encoding` must be `PCM_S16LE`**, both now a `ValueError`
  at construction. Defaults are unchanged, so only an explicit other value is affected. The spec
  pins the rate; the encoding check is deliberately *narrower* than the spec, because `pcm_f32le` is
  legal on the wire but LiveKit frames are int16 forwarded unconverted — declaring it would have the
  service read 2-byte samples as 4-byte floats. Plausible garbage, no error.
- **`AdditionalVocabEntry` is a dataclass**, not a pydantic model. Construction and field access are
  unchanged and either SDK's class is accepted as input, but `.model_dump()` / `.model_validate()`
  are gone from the exported type.
- **A server error now raises.** Sessions that previously logged a rejection and hung until timeout
  now fail with an `APIError`. Better, but it is a new exception on a path that used to be silent.

## Back-compatibility

Per review: the public names this plugin exposed before Agent STT still work, each with a warning,
**until 2026-10-05**. Four mechanisms:

| Mechanism | Covers | Behaviour |
|---|---|---|
| Module `__getattr__` in `__init__.py` | removed exported *names* | imports, warns once per process, returns the voice SDK's own object |
| `_DROPPED_ARGS` (15 names) | kwargs with no equivalent | warned with the reason, then discarded |
| `_MIGRATED_ARGS` (3 names) | renamed kwargs | warned and carried over to the replacement |
| `_resolve_model` / `_resolve_turn_detection_mode` | values whose meaning changed | reconciled once at construction |

Symbols need the `__getattr__` route because they fail at *import*, before any `STT` exists, so the
kwarg handling never runs:

```python
from livekit.plugins.speechmatics import OperatingPoint   # import time  — the shim
STT(operating_point="enhanced")                           # construction — the kwarg path
```

What each old name does now:

- **`OperatingPoint`** imports and warns; it is `speechmatics.rt.OperatingPoint`, the identical class
  object callers already hold, so `isinstance` still passes.
- **`SpeakerFocusMode`** imports and warns. The feature is gone and `focus_mode` is ignored; it is
  expected to be reintroduced in a future release.
- **`TurnDetectionMode.FIXED` / `.ADAPTIVE` / `.SMART_TURN`** keep their original values and resolve
  to the default mode with a warning. Each named a service-side endpointing strategy Agent STT does
  not distinguish; `FIXED` warns twice, since its `end_of_utterance_silence_trigger` timing is the
  one behaviour genuinely lost rather than renamed.
- **`operating_point="enhanced"` / `"standard"`** resolve to `linden-1` with a warning instead of
  raising. These are RT operating points, not Agent STT models, and the previous README's own
  examples used `enhanced` — so this is the line most existing code will hit.
- **`STT.update_speakers()`** is a no-op that warns. It only ever updated speaker focus.
- **`AdditionalVocabEntry`** is accepted from either SDK and copied onto the Agent STT type. The two
  classes hold the same fields, so nothing is lost and nothing is warned. Left alone, the old class
  survived construction and then failed the `StartRecognition` JSON encode.

Unrecognised keyword arguments are now reported too, so `STT(langauge="en")` no longer leaves you
with English-by-default and no explanation.

`pyproject.toml` therefore carries `speechmatics-voice>=0.2.8` beside
`speechmatics-agent-stt~=0.1.0`. Three things about it:

- It is the **base package**, not `speechmatics-voice[smart]` as before — `[smart]` is onnxruntime +
  transformers for local smart-turn inference, which this plugin no longer does.
- It adds **no new distributions** to anyone's tree: it needs `speechmatics-rt` (already pinned
  higher by `agent-stt` — `>=1.1.1` against voice's `>=0.5.3`), plus `pydantic` and `numpy`, both
  already required by `livekit-agents` core. It does raise two floors: `pydantic` to `>=2.10.6`
  (core allows `>=2.5`) and `numpy` to `>=1.26.4` (core allows `>=1.26.0`).
- It is **never on the runtime path.** The imports live inside the `__getattr__` branches, so
  importing the plugin, constructing `STT`, building the config and passing deprecated kwargs all
  leave `speechmatics.voice` out of `sys.modules`. Only touching a deprecated *symbol* loads it,
  which a subprocess test pins.

Remove the dependency, the shims and the deprecated tables together after 2026-10-05.

## Non-breaking

- **Segment shape** now uses `Segment.from_message` (singular `segment` + message-level metadata).
  This is what made every session transcribe nothing.
- **New `model` parameter**, preferred over `operating_point`, resolved by a pure `_resolve_model()`.
  It never raises: an unsupported name — a typo or an old operating point — warns and falls back to
  the default model, and when `model` and `operating_point` disagree, `model` wins with a warning.
- **Turn detection on the wire** is a top-level `turn_config`, not a `transcription_config` field.
- **`EXTERNAL` stays the default**, for parity with the previous plugin, and Silero is still
  auto-loaded when no `vad` is passed so a bare `STT()` closes turns. Unlike before, a missing
  `livekit-plugins-silero` no longer raises `ImportError` — construction succeeds and the plugin
  warns that nothing will finalize. `VAD` mode is the alternative, where the service endpoints
  itself; pair it with `turn_detection="stt"` on `AgentSession`, which otherwise ignores the
  end-of-speech events this plugin emits.
- **Server errors surfaced** as `APIError` on the stream — non-retryable for a rejected session,
  retryable `APIConnectionError` for connect failures. Warnings still just log.
- **Dropped connections detected.** The SDK swallows `TransportError` inside `send_audio`, closes its
  audio gate and drops every later frame, so a mid-call socket drop stopped transcription with no
  error. The plugin now reads the gate after each send and raises a retryable `APIConnectionError`.
- **Usage counted only for audio the service accepted**, so `RECOGNITION_USAGE` no longer
  over-reports frames the SDK dropped.
- **Diarization config is actually sent** — `max_speakers`, `speaker_sensitivity`,
  `prefer_current_speaker` and `known_speakers` were accepted from the caller and never put on the
  wire. Sent only when diarization is on and a knob is set.
- **Partials on by default.** The service enables them, so `interim_results` is `True` unless
  `include_partials=False`.
- **Capabilities mirror the options** they describe (`interim_results`, `diarization`), computed
  after deprecated arguments are migrated so an alias is reflected too.
- **Deprecated-arg handling is table-driven:** one map of dropped names to the reason they are gone,
  one migration map. Adding a deprecation is one line. `operating_point` moved off
  `warnings.warn(DeprecationWarning)`, which Python hides by default, onto the plugin logger like
  everything else.
- **Result language** is labelled with the stream's own configured language, so a per-stream
  `stream(language=...)` override no longer diverges from the label.
- **`session_id` public property** on `SpeechStream`, and `Model` exported.
- **PII-safe logging** — message payloads moved into the redactable `lk.pii.message` attribute.
- **First test suite** — `tests/test_plugin_speechmatics_stt.py`: 30 tests, 51 cases, fully offline.
  Covers the config-to-wire contract (including that `transcription_config` stays inside the spec's
  whitelist, which is what used to reject every session), the message-to-event translation, and the
  whole back-compatible surface above.


livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -409,127 +336,42 @@ def _validate_stt_options(self) -> list[str]:
...
-        # Speaker configuration
-        config.speaker_config = SpeakerFocusConfig(
-            focus_speakers=opts.focus_speakers,
-            ignore_speakers=opts.ignore_speakers,
-            focus_mode=opts.focus_mode,
+        config = TranscriptionConfig(
+            language=language if is_given(language) else opts.language,
+            model=opts.model,
+            turn_detection_mode=_handle_turn_detection_mode(opts.turn_detection_mode),
+            diarization="speaker" if opts.enable_diarization else None,
```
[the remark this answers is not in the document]
Smiljanic19A: Resolved

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -33,46 +36,62 @@
     NotGivenOr,
 )
 from livekit.agents.utils import AudioBuffer, is_given
-from speechmatics.rt import ClientMessageType
-from speechmatics.voice import (
+
+# Everything comes from the Agent STT SDK now — the plugin no longer depends on the
```
LArmstrongDev: I think we can remove this comment. The MR explains the migration from Voice SDK to Agent STT, and the packages show the story, don't think the comment is needed.

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -33,46 +36,62 @@
...
-    (Silero is auto-loaded if no `vad` is provided).
+    The values mirror the Agent STT SDK's own turn-detection modes so the two never
+    drift; only the member names differ (DEFAULT=VAD)
     """
 
-    EXTERNAL = "external"
-    FIXED = "fixed"
-    ADAPTIVE = "adaptive"
-    SMART_TURN = "smart_turn"
+    DEFAULT = AgentTurnDetectionMode.VAD.value
```
LArmstrongDev: The agreed upon naming is `VAD` for VAD mode, and `EXTERNAL` for EXTERNAL mode.
For consistency, let's use the same naming here.

Regarding default, the previous plugin defaulted to EXTERNAL mode. What's the decision behind changing this to default to VAD mode?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -125,51 +137,45 @@ def __init__(
...
-        ignore_speakers: NotGivenOr[list[str]] = NOT_GIVEN,
-        focus_mode: SpeakerFocusMode = SpeakerFocusMode.RETAIN,
         known_speakers: NotGivenOr[list[SpeakerIdentifier]] = NOT_GIVEN,
         sample_rate: int = 16000,
         audio_encoding: AudioEncoding = AudioEncoding.PCM_S16LE,
         vad: NotGivenOr[vad.VAD | None] = NOT_GIVEN,
         **kwargs: Any,
     ):
-        """Create a new instance of Speechmatics STT using the Voice SDK.
+        """Create a new instance of Speechmatics STT.
```
LArmstrongDev:
```suggestion
        """Create a new instance of Speechmatics STT using Agent STT SDK.
```

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -125,51 +137,45 @@ def __init__(
...
 
         Args:
             api_key: Speechmatics API key. Can be set via `api_key` argument
                 or `SPEECHMATICS_API_KEY` environment variable.
 
             base_url: Custom base URL for the API. Can be set via `base_url`
-                argument or `SPEECHMATICS_RT_URL` environment variable. Optional.
+                argument or `SPEECHMATICS_RT_URL` environment variable. Falls back to
+                `DEFAULT_BASE_URL` (the Agent STT endpoint) when neither is set.
```
LArmstrongDev: What's meant by "the Agent STT endpoint" here? We have multiple endpoints.

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -125,51 +137,45 @@ def __init__(
...
+                from an external VAD). Defaults to `TurnDetectionMode.DEFAULT`.
 
-            turn_detection_mode: Controls how the STT engine detects end of speech
-                turns. Use `EXTERNAL` when turn boundaries are controlled manually,
-                for example via an external VAD or the `finalize()` method. Use
-                `ADAPTIVE` for simple VAD or `SMART_TURN` for ML-based endpointing.
-                `FIXED` uses a fixed amount of silence, as determined by the
-                `end_of_utterance_silence_trigger` parameter.
-                Defaults to `TurnDetectionMode.EXTERNAL`.
+            model: The transcription model (operating point) to use, e.g. `"linden-1"`.
```
LArmstrongDev:
```suggestion
            model: The transcription model to use, e.g. `"linden-1"`.
```

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/utils.py
```diff
@@ -19,7 +17,7 @@ def get_tts_url(base_url: str, voice: str, sample_rate: int) -> str:
     query_params = {}
     query_params["output_format"] = f"pcm_{sample_rate}"
     query_params["sm-sdk"] = f"livekit-plugins-{lk_version}"
```
LArmstrongDev, line 21: So, with both `sm-sdk` and `sm-app`, these will track the LiveKit version used. What about the Agent STT SDK version?

livekit-plugins/livekit-plugins-speechmatics/pyproject.toml
```diff
@@ -22,7 +22,13 @@ classifiers = [
     "Programming Language :: Python :: 3 :: Only",
 ]
 
-dependencies = ["livekit-agents>=1.7.1", "speechmatics-voice[smart]>=0.2.8"]
+dependencies = [
+    "livekit-agents>=1.6.10",
```
LArmstrongDev: Previously, the dependency was 1.7.1. Why has it lowered to 1.6.10?

LArmstrongDev (commented):
Partial review - some comments.

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
Smiljanic19A: Removed

Smiljanic19A: Removed

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
Smiljanic19A: Cool - did not realize we stuck on VAD as a name.

As for using VAD as the DEFAULT - it just makes more sense to me. Agent-transcription-service already solves endpointing - so i dont see why we would make users implement their own VAD by default. Having said that - i have now flipped it so that the default is EXTERNAL.

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
Smiljanic19A: Thanks!

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
Smiljanic19A: You are right - not really clear, updated it

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
Smiljanic19A: Thanks again

livekit-plugins/livekit-plugins-speechmatics/pyproject.toml
Smiljanic19A: My branch predated the lockstep bump to 1.7.1 on livekit/main, so it still showed 1.6.10. Bumped it to >=1.7.1 to match

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/utils.py
Smiljanic19A: Using sdk_version for sm-app, this line had been changed to test something out.

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -759,11 +638,18 @@ async def _process_audio(self) -> None:
...
                 if self._client:
                     for frame in frames:
                         self._speech_duration += frame.duration
-                        await self._client.send_audio(frame.data.tobytes())
+                        try:
+                            await self._client.send_audio(frame.data.tobytes())
+                        except (SMConnectionError, SMTimeoutError, TransportError) as e:
+                            raise APIConnectionError(
+                                f"lost connection to Speechmatics while sending audio: {e}"
+                            ) from e
```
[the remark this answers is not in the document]
Smiljanic19A: This is a pattern established throughout the repo - everyone interpolates e into APIConnectionError

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -644,49 +507,65 @@ async def _run(self) -> None:
...
+        # APIError so it reaches the caller instead of dying in a log; transient transport
+        # failures surface as a (retryable) APIConnectionError.
+        try:
+            await self._client.connect()
+        except (TranscriptionError, SessionError) as e:
+            raise APIError(
+                f"Speechmatics rejected the session: {e}", retryable=False
+            ) from e
+        except (SMConnectionError, SMTimeoutError, TransportError) as e:
+            raise APIConnectionError(f"failed to connect to Speechmatics: {e}") from e
```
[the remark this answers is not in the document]
Smiljanic19A: Established throughout the repo - everyone interpolates e into APIConnectionError

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -644,49 +507,65 @@ async def _run(self) -> None:
...
-            config=self._config,
+            transcription_config=self._config,
+            turn_config=TurnConfig(
+                turn_detection_mode=_handle_turn_detection_mode(self._turn_detection_mode)
+            ),
+            audio_format=AudioFormat(
+                encoding=self._stt._audio_encoding,
+                sample_rate=self._stt._sample_rate,
+                chunk_size=DEFAULT_CHUNK_SIZE,
+            ),
```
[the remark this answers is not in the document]
Smiljanic19A: expected

LArmstrongDev:
@Smiljanic19A - please can you use the top level for commit names to describe the changes you're making? As I'm reviewing, it's quite hard to follow when the names for the commits are all the same and non-descriptive, having to expand each out for context.

livekit-plugins/livekit-plugins-speechmatics/pyproject.toml
```diff
@@ -22,7 +22,7 @@ classifiers = [
     "Programming Language :: Python :: 3 :: Only",
 ]
 
-dependencies = ["livekit-agents>=1.8.0", "speechmatics-voice[smart]>=0.2.8"]
+dependencies = ["livekit-agents>=1.8.0", "speechmatics-agent-stt>=0.1.0"]
```
LArmstrongDev: This should be capped to avoid pulling in next major version automatically, as this could bring breaking changes.

livekit-plugins/livekit-plugins-speechmatics/README.md
```diff
@@ -10,76 +10,84 @@ See [https://docs.livekit.io/agents/integrations/stt/speechmatics/](https://docs
 pip install livekit-plugins-speechmatics
 ```
```
LArmstrongDev, line 12: There's no mention of the `linden-1` model here - feels like it should be mentioned.

livekit-plugins/livekit-plugins-speechmatics/README.md
```diff
@@ -10,76 +10,84 @@ See [https://docs.livekit.io/agents/integrations/stt/speechmatics/](https://docs
...
-    min_endpointing_delay=0.3,
-    max_endpointing_delay=5.0,
     ...
 )
 ```
 
-## Usage (Speechmatics end of utterance detection and speaker ID)
+## Diarization
+
+Speechmatics can attribute words to individual speakers. Diarization is enabled by default
```
LArmstrongDev: Agent STT attributes per segment, not per word for speakers. Segment carries one speaker, no per-word data.

livekit-plugins/livekit-plugins-speechmatics/README.md
```diff
@@ -10,76 +10,84 @@ See [https://docs.livekit.io/agents/integrations/stt/speechmatics/](https://docs
...
+Pass a `vad` to the plugin; its end-of-speech drives `finalize()`:
 
 ```python
-from livekit.agents import AgentSession, inference
-from livekit.agents.inference import TurnDetector
-from livekit.plugins import speechmatics
+from livekit.agents import AgentSession
+from livekit.plugins import silero, speechmatics
 
 agent = AgentSession(
```
LArmstrongDev, line 64: Because `AgentSession` fills in its own `inference.VAD` when none is given (`agent_session.py:605-607`), this runs a local Silero model in the plugin and a hosted inference VAD in the session over the same audio. 
Instead, the previous README passed `vad=inference.VAD()` to the session.

livekit-plugins/livekit-plugins-speechmatics/README.md
```diff
@@ -10,76 +10,84 @@ See [https://docs.livekit.io/agents/integrations/stt/speechmatics/](https://docs
...
+```python
+from livekit.agents import AgentSession
+from livekit.plugins import speechmatics
 
-The default `EXTERNAL` mode pairs naturally with LiveKit's turn detector. The format for the output text needs to be adjusted to not include any extra content at the end of the utterance. Using `[Speaker S1] ...` as the `speaker_active_format` should work well. You may need to adjust your system instructions to inform the LLM of this format for speaker identification. You must also include the listener for when the VAD has detected the end of speech.
+agent = AgentSession(
+    stt=speechmatics.STT(
+        turn_detection_mode=speechmatics.TurnDetectionMode.VAD,
+    ),
+    ...
```
LArmstrongDev, line 48: I don't think this example works as described. Speechmatics will close turns on its side and the plugin will emit an end-of-speech event, but the session ignores that event unless you also set `turn_detection="stt"` (see the check at `audio_recognition.py:1302`).

And since `AgentSession` quietly supplies its own VAD and turn detector when you don't pass any (in `agent_session.py:541` and `:605`), what actually decides the user's turn here is LiveKit's turn detector, not Speechmatics. So the mode gets configured but never used, I think?

livekit-plugins/livekit-plugins-speechmatics/pyproject.toml
Smiljanic19A: 👍

livekit-plugins/livekit-plugins-speechmatics/README.md
Smiljanic19A: Multiple things in the README seem off - added a model section that documents the valid options (linden-1 currently) - will have some follow ups shortly

livekit-plugins/livekit-plugins-speechmatics/README.md
Smiljanic19A: Fixed

livekit-plugins/livekit-plugins-speechmatics/README.md
Smiljanic19A: One inference.VAD() instance now goes to both the plugin and the session, and the silero import is gone.

livekit-plugins/livekit-plugins-speechmatics/README.md
Smiljanic19A: Added turn_detection="stt" to the example, fixed the "nothing to wire up" line, and noted the requirement in the mode list.

livekit-plugins/livekit-plugins-speechmatics/README.md
Smiljanic19A: updated

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/utils.py
LArmstrongDev: I'm a little confused here. I'd expect `sm-sdk` to point to the SDK version being used (Agent STT SDK), and then sm-app to point to the livekit version.

LArmstrongDev:
For the deprecated parameters, is there an error that appears if someone tries to use them? How will a user be warned?

LArmstrongDev:
Can you update the PR description to make sure it's up to date? E.g. we don't use DEFAULT here - the default mode is EXTERNAL, and we then have VAD as an option for turn detection mode.

LArmstrongDev:
What happens if someone tries to use EXTERNAL mode but doesn't configure any vad?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -759,11 +639,18 @@ async def _process_audio(self) -> None:
...
+                # Send audio frames. A transport drop mid-session surfaces as a (retryable)
+                # APIConnectionError so the base stream retries/reports instead of dying on a
+                # raw SDK error — matching how connect() failures are wrapped.
                 if self._client:
                     for frame in frames:
                         self._speech_duration += frame.duration
-                        await self._client.send_audio(frame.data.tobytes())
+                        try:
+                            await self._client.send_audio(frame.data.tobytes())
+                        except (SMConnectionError, SMTimeoutError, TransportError) as e:
```
LArmstrongDev: Is this reachable? I think the SDK would catch the TransportError itself and then drop frames once the session is marked not ready. Can you check to make sure it's handled?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -126,173 +130,118 @@ def __init__(
...
         speaker_sensitivity: NotGivenOr[float] = NOT_GIVEN,
         max_speakers: NotGivenOr[int] = NOT_GIVEN,
         speaker_active_format: NotGivenOr[str] = NOT_GIVEN,
-        speaker_passive_format: NotGivenOr[str] = NOT_GIVEN,
         prefer_current_speaker: NotGivenOr[bool] = NOT_GIVEN,
-        focus_speakers: NotGivenOr[list[str]] = NOT_GIVEN,
-        ignore_speakers: NotGivenOr[list[str]] = NOT_GIVEN,
-        focus_mode: SpeakerFocusMode = SpeakerFocusMode.RETAIN,
         known_speakers: NotGivenOr[list[SpeakerIdentifier]] = NOT_GIVEN,
         sample_rate: int = 16000,
```
LArmstrongDev, line 152: Do we protect against setting `sample_rate` to an invalid value here?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -126,173 +130,118 @@ def __init__(
...
+                "finalize() from end-of-speech, or use turn_detection_mode=VAD for service-side "
+                "endpointing."
+            )
+
         # Set default values for optional parameters
         super().__init__(
             capabilities=stt.STTCapabilities(
                 streaming=True,
-                interim_results=True,
+                # The service withholds partial segments unless asked, so interim results
```
LArmstrongDev: This is not true - partials are enabled by default in the Agent STT service.

Smiljanic19A:
> What happens if someone tries to use EXTERNAL mode but doesn't configure any vad? @LArmstrongDev 

There's a loud logger.warning at construction for exactly this combination - pointing at either passing a vad or switching to VAD mode. If we raise - we kill the manual finalize() path.

If someone ignores the warnings - they connect and transcribe nothing

Smiljanic19A:
> For the deprecated parameters, is there an error that appears if someone tries to use them? How will a user be warned? @LArmstrongDev 

They warn and keep working at construction.

Most of them are dropped - `_DROPPED_ARGS`.
There is also _MIGRATED_ARGS - for fields that have a new wire - where we populate the value from the old arg to the new wire.

Your comment did raise an issue though - upon inspection i noticed operating_point was the only one going through warnings.warn(DeprecationWarning) - so nothing was printed. That is fixed now

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -759,11 +639,18 @@ async def _process_audio(self) -> None:
                         self._vad_stream.push_frame(data)
                     frames = audio_bstream.write(data.data.tobytes())
 
-                # Send audio frames
+                # Send audio frames. A transport drop mid-session surfaces as a (retryable)
+                # APIConnectionError so the base stream retries/reports instead of dying on a
+                # raw SDK error — matching how connect() failures are wrapped.
                 if self._client:
                     for frame in frames:
                         self._speech_duration += frame.duration
```
LArmstrongDev, line 679: This counts the frames before the send - but the SDK drops frames when the session isn't ready. This means the usage would over-report after any interruption?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
```diff
@@ -803,50 +690,78 @@ def _handle_message(self, message: dict[str, Any]) -> None:
...
-        """Handle AddPartialSegment events."""
-        segments: list[dict[str, Any]] = message.get("segments", [])
-        if segments:
-            self._send_frames(segments, is_final=False)
+        # Parse the singular agent-STT segment
+        seg = Segment.from_message(message)
+
+        # Format the text with the active-speaker formatter
+        speaker_id = seg.speaker or "UU"
+        format_str = opts.speaker_active_format or "{text}"
```
LArmstrongDev: Is `speaker_active_format` leftover from speaker focus? What does it mean, when we have no speaker focus?

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
Smiljanic19A: Yeah - its dead we already solved this problem elsewhere. Dropped the dead except and check is_ready_for_audio after each send instead — closed gate with no session_error raises a retryable APIConnectionError.

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
Smiljanic19A: No - the specs would just reject invalid values.

That made sense to me as it would suck to have to update a guard list if the specs change in the future. 
I added 2 lists that get validated against input at construction: SUPPORTED_SAMPLE_RATE (16000, the spec's const) and SUPPORTED_AUDIO_ENCODINGS (PCM_S16LE). The spec allows pcm_f32le, but the plugin forwards LiveKit's 16-bit PCM frames unconverted, so declaring f32le would have been accepted by the server and read as garbage

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
Smiljanic19A: Moved the increment below the gate check so only accepted frames count

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
Smiljanic19A: Its a leftover yes. The formatter itself still does real work — it folds the speaker label into the transcript text — so I've renamed it to speaker_format and kept speaker_active_format as a deprecated alias in the migration table.

LArmstrongDev (changes requested):
Reviewed, left some comments.

Smiljanic19A:
> Reviewed, left some comments.

addressed - thanks @LArmstrongDev

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/stt.py
LArmstrongDev: Nice, the rename is neat - thanks.

LArmstrongDev:
> > Reviewed, left some comments.
> 
> addressed - thanks @LArmstrongDev

There's still the outstanding question on how the underlying SDK is tracked. Both `sm-sdk` and `sm-app` carry the LiveKit version, none carry the Agent STT SDK version.

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/utils.py
Smiljanic19A: These query params are appended to the tts url. We dont use an sdk tts - so i dropped that line completely.

For stt - the query params are added by the sdk (sdk_identifier) - so the version is tracked thorugh that.

As for the confusion - thats my bad my previous response was nonesense.

LArmstrongDev (approved):
Thanks for making the markups. LGTM

livekit-plugins/livekit-plugins-speechmatics/livekit/plugins/speechmatics/__init__.py
```diff
@@ -34,8 +33,7 @@
     "SpeechStream",
     "AdditionalVocabEntry",
     "AudioEncoding",
-    "OperatingPoint",
```
tinalenguyen, line 37: removing these could break user code, i think we should keep backward compatibility for the voice sdk and throw a warning until 10/5 (the deprecation date for inference as well)

Smiljanic19A: Really appreciate the input Tina.
 
We're in agreement on prioritising reducing the friction for current users on upgrade. That being said, supporting the Voice SDK and Agent STT simultaneously isn't a path we want to follow. This PR does mark a transition - from the Voice SDK to Agent STT - and Agent STT is a different product, with different configuration and features. Naturally, this means some features are not available on upgrade.
 
I have added in code paths to help users with the transition (e.g. for deprecated parameters like operating_point, I've added in a redirect to model and a warning that the parameter is deprecated) where possible. However, I have kept the MR to Agent STT only - removing Voice SDK entirely - to avoid adding complexity and maintaining a second path inside the plugin, which would require a translation layer for a service we're moving off.
 
The summary of protective changes I have made is:
 
| Surface | Old behaviour | Now |
|---|---|---|
| `OperatingPoint` | `ImportError` | imports, warns — the same `speechmatics.rt` class object, so `isinstance` still passes |
| `SpeakerFocusMode` | `ImportError` | imports, warns |
| `TurnDetectionMode.FIXED` / `.ADAPTIVE` / `.SMART_TURN` | `AttributeError` | kept with their original values, resolved to the default mode, warns |
| `operating_point="enhanced"` / `"standard"` | `ValueError` at construction | resolves to `linden-1`, warns |
| `STT.update_speakers()` | `AttributeError` | warns, no-op |
| `AdditionalVocabEntry` | failed the `StartRecognition` encode | either SDK's class accepted and converted; same fields, so nothing is lost |
| 15 removed kwargs (`focus_mode`, `max_delay`, `end_of_utterance_*`, …) | silently swallowed by `**kwargs` | warned, each with the reason it's gone |
| 3 renamed kwargs (`speaker_active_format`, `enable_partials`, `diarization_sensitivity`) | silently swallowed | warned and carried over to the replacement |
| An unrecognised kwarg | silently swallowed | warned, so a typo doesn't cost you a default you never chose |
 
The result is that none of it breaks. The previous README's own example — `OperatingPoint.ENHANCED`,
`TurnDetectionMode.ADAPTIVE`, `focus_mode=...` — now constructs and runs, with a warning per
deprecated part. There's a test that executes exactly that snippet, in
`tests/test_plugin_speechmatics_stt.py` (30 tests, offline).
 
Even with the above, there are consequences of it being a different product that cannot be covered over: the endpoint. The
default `base_url` moves from `…/v2` to `…/v2/agent` — same host and region, different service — and
pointing it back at `/v2` doesn't restore the old behaviour, since that endpoint speaks a protocol
this plugin no longer reads. The plugin warns at construction when the resolved URL isn't an Agent
STT endpoint, which is the most we can do from inside it. Worth knowing that `SPEECHMATICS_RT_URL`
survives an upgrade untouched, so that's the case that will present as a hang rather than a
misconfiguration.

tinalenguyen (approved):

PR #6740, opened 2026-08-07
fix(cartesia): redact API keys from websocket handshake errors

LHMQ878:
## Summary
- Wrap Cartesia TTS/STT websocket handshake failures as `APIStatusError`/`APIConnectionError` without chaining the aiohttp cause, so `X-API-Key` in `RequestInfo` cannot appear in exception/`Task` repr logs.
- Make `ConnectionPool.prewarm()` swallow connect failures and log with `%s` (str) instead of leaving an unretrieved task exception (the path that published Cartesia API keys).

## Test plan
- [x] `pytest tests/test_connection_pool.py tests/test_plugin_cartesia_tts.py -v --unit --plugin cartesia`
- [ ] Confirm a failed Cartesia TTS prewarm no longer logs `X-API-Key` / raw API key material

Fixes #6739

Issue #6739, opened 2026-08-07
never retrieved exception=WSServerHandshakeError leads to published Cartesia API Key in logs


AdemFr:
### Bug Description

On agent worker failures we saw

```
Task exception was never retrieved
exception=WSServerHandshakeError(RequestInfo(url=URL('wss://api.cartesia.ai/tts/websocket?cartesia_version=2025-04-16'), method='GET', headers=<CIMultiDictProxy('Host': 'api.cartesia.ai', 'User-Agent': 'LiveKit Agents Cartesia Plugin/1.6.6', 'X-API-Key': '<RAW_API_KEY>', 'Upgrade': 'websocket', 'Connection': 'Upgrade', 'Sec-WebSocket-Version': '13', 'Sec...
```

Not sure if the cartesia plugin, or more generally the livekit agents package should handle this better.

Here is an AI diagnostic of the issue:

1. LiveKit's ConnectionPool.prewarm() fires _prewarm_impl() as a task and keeps only a weakref — on failure nothing awaits it.
2. The task is GC'd unretrieved, so CPython's asyncio handler logs it, repr()-ing the future.
3. Task.__repr__ embeds the exception's repr.
4. aiohttp.ClientResponseError.__repr__ interpolates its RequestInfo, which is a NamedTuple containing the outgoing request headers.
5. The Cartesia plugin puts the key in those headers.

### Expected Behavior

No API Key should ever make it into logs from any third party.

### Reproduction Steps

```bash
1. 
2.
3.
...
- Sample code snippet, or a GitHub Gist link -
```

### Operating System

Linux

### Models Used

Cartesia Sonic 3.5

### Package Versions

```bash
livekit==1.1.13
livekit-agents==1.6.6
livekit-api==1.2.0
livekit-blingfire==1.1.0
livekit-local-inference==0.2.6
livekit-plugins-cartesia==1.6.6
livekit-plugins-deepgram==1.6.6
livekit-plugins-elevenlabs==1.6.6
livekit-plugins-noise-cancellation==0.2.5
livekit-plugins-openai==1.6.6
livekit-protocol==1.1.19
```

### Proposed Solution

```python

```

PR #7032, opened 2026-08-29
fix(plugins): redact API keys from rejected websocket handshakes

### LINKED ISSUE #7031: API keys leak into logs when a websocket upgrade is rejected (6 sites still unguarded after #6740)
### Bug

`WSServerHandshakeError` extends `aiohttp.ClientResponseError`, so it is neither an `aiohttp.ClientConnectorError` nor an `asyncio.TimeoutError`. Six credentialed `ws_connect` sites still wrap the connect in `except (aiohttp.ClientConnectorError, asyncio.TimeoutError)`, so a rejected upgrade (401/403 from a bad, expired or rate-limited key) matches neither handler and escapes uncaught. It surfaces as an unretrieved task exception whose repr embeds `RequestInfo`, including the auth header.

That is the same failure #6739 reported against cartesia and #6740 fixed there. The reporter on that issue wondered whether it was "the cartesia plugin, or more g[enerally]"; these are the sites where it is still reachable.

| plugin | file |
|---|---|
| deepgram | `stt.py`, `stt_v2.py` |
| simplismart | `stt.py` |
| smallestai | `stt.py` |
| xai | `stt.py`, `tts.py` |

Each sends the key on the websocket (deepgram, for example, `headers={"Authorization": f"Token {self._api_key}"}` in `stt.py`) and then catches only the two error types above. I checked the other credentialed `ws_connect` sites by walking each `try` block's handlers; the remaining ten already catch the response error or re-raise with `from None`.

### Steps to reproduce

Deepgram's exact call shape against a server that refuses the upgrade:

```python
import asyncio
import aiohttp
from aiohttp import web

FAKE_KEY = "dg_live_SUPERSECRET_abcdef0123456789"


async def reject(request):
    return web.Response(status=401, text="Unauthorized")


async def main():
    app = web.Application()
    app.router.add_get("/v1/listen", reject)
    runner = web.AppRunner(app)
    await runner.setup()
    await web.TCPSite(runner, "127.0.0.1", 8781).start()

    session = aiohttp.ClientSession()
    try:
        await asyncio.wait_for(
            session.ws_connect(
                "http://127.0.0.1:8781/v1/listen?model=nova-2",
                headers={"Authorization": f"Token {FAKE_KEY}"},
            ),
            10,
        )
    except (aiohttp.ClientConnectorError, asyncio.TimeoutError):
        print("caught by the plugin's handler")
    except Exception as e:
        print(type(e).__name__, "escaped;", "key in repr:", FAKE_KEY in repr(e))
        print(repr(e)[:220])
    finally:
        await session.close()
        await runner.cleanup()


asyncio.run(main())
```

```
WSServerHandshakeError escaped; key in repr: True
WSServerHandshakeError(RequestInfo(url=URL('http://127.0.0.1:8781/v1/listen?model=nova-2'), method='GET',
headers=<CIMultiDictProxy('Host': '127.0.0.1:8781', 'Authorization': 'Token dg_live_SUPERSECRET_abcdef0123456789',
```

A 401 is the ordinary case here, not a contrived one: any wrong, rotated or rate-limited key produces it, and the reconnect path retries, so it can repeat.

### Expected

No credential material in logs or exception output.

The shape #6740 used for cartesia ports directly: catch the response error separately and re-raise with `from None`, surfacing `type(e).__name__` rather than `str(e)` or `repr(e)`.

### Note

Filing publicly because #6739 was handled that way and there is no `SECURITY.md` in this repo or on the livekit org. Happy to move it if you would rather it went through a private channel. I have the patch and can open a PR.

### Version

main @ cdb37ad

aniketwaghh:
## Summary

`WSServerHandshakeError` subclasses `aiohttp.ClientResponseError`, so it is neither an `aiohttp.ClientConnectorError` nor an `asyncio.TimeoutError`. Six credentialed `ws_connect` sites still wrapped the connect in `except (aiohttp.ClientConnectorError, asyncio.TimeoutError)`, so a rejected upgrade matched neither handler and escaped uncaught, carrying `RequestInfo` and therefore the auth header.

Same failure as #6739, which #6740 fixed for cartesia. That reporter asked whether it was "the cartesia plugin, or more g[enerally]"; these are the sites where it was still reachable.

| plugin | file |
|---|---|
| deepgram | `stt.py`, `stt_v2.py` |
| simplismart | `stt.py` |
| smallestai | `stt.py` |
| xai | `stt.py`, `tts.py` |

Fixes #7031

## Approach

The shape from #6740, applied unchanged: handle `asyncio.TimeoutError` and `aiohttp.ClientResponseError` separately, re-raise with `from None`, and surface only the status and `type(e).__name__` rather than `str(e)` or `repr(e)`.

I scoped this by parsing every `try` block containing a credentialed `ws_connect` and reading which exception types its handlers actually name, rather than grepping. My first grep pass wrongly cleared `deepgram/stt.py`, which is why the count is from the AST walk. The other ten credentialed sites already catch the response error or re-raise with `from None`, and are untouched here.

## Verification

A local server refusing the upgrade with 401, called with deepgram's exact call shape.

Before:

```
WSServerHandshakeError escaped the handler
WSServerHandshakeError(RequestInfo(url=URL('http://127.0.0.1:8781/v1/listen?model=nova-2'),
method='GET', headers=<CIMultiDictProxy('Host': '127.0.0.1:8781',
'Authorization': 'Token dg_live_SUPERSECRET_abcdef0123456789', ...
API KEY PRESENT IN EXCEPTION REPR: True
```

After, the same request raises `APIStatusError: Invalid response status`, status_code 401, `__cause__` None, and the key appears in neither `str()` nor `repr()`.

Checked what actually reaches a log, since that was the reported symptom:

```
KEY in traceback.print_exc() output:   False
KEY in logger.exception() output:      False
```

One thing worth stating plainly: `raise ... from None` suppresses `__cause__` and sets `__suppress_context__`, so tracebacks and `logging.exception` omit the original error, which is the #6739 symptom. The original exception object does remain on `__context__`, so an error reporter that walks the context chain and ignores the suppress flag could still reach it. That is equally true of the cartesia fix in #6740; I have kept the same shape rather than diverge. Happy to go further if you would rather these paths scrub `RequestInfo` explicitly.

## Tests

`tests/test_ws_handshake_credential_redaction.py`, mirroring `tests/test_plugin_cartesia_tts.py` from #6740: construct a `WSServerHandshakeError` whose `RequestInfo` headers carry the key, drive the real `_connect_ws`, and assert `APIStatusError` with the status preserved, the key absent from `str` and `repr`, and `__cause__` None. Marked per plugin so they run under the existing `--plugin` filtering.

`ruff check` and `ruff format` clean on all seven files.

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -775,14 +775,6 @@ class SpeechStream(stt.SpeechStream):
                 ),
                 self._conn_options.timeout,
             )
-            self._report_connection_acquired(time.perf_counter() - t0, False)
-            ws_headers = {
-                k: v for k, v in ws._response.headers.items() if k.startswith("dg-") or k == "Date"
-            }
-            logger.debug(
-                "Established new Deepgram STT WebSocket connection:",
-                extra={"headers": ws_headers},
-            )
         except asyncio.TimeoutError:
             raise APIConnectionError("failed to connect to deepgram") from None
         except aiohttp.ClientResponseError as e:
@@ -795,6 +787,15 @@ class SpeechStream(stt.SpeechStream):
             raise APIConnectionError(
                 f"failed to connect to deepgram ({type(e).__name__})"
             ) from None
+
+        self._report_connection_acquired(time.perf_counter() - t0, False)
+        ws_headers = {
+            k: v for k, v in ws._response.headers.items() if k.startswith("dg-") or k == "Date"
+        }
+        logger.debug(
+            "Established new Deepgram STT WebSocket connection:",
+            extra={"headers": ws_headers},
+        )
         return ws
 
     def _on_audio_duration_report(self, duration: float) -> None:
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt.py
@@ -783,8 +783,18 @@ class SpeechStream(stt.SpeechStream):
                 "Established new Deepgram STT WebSocket connection:",
                 extra={"headers": ws_headers},
             )
-        except (aiohttp.ClientConnectorError, asyncio.TimeoutError) as e:
-            raise APIConnectionError("failed to connect to deepgram") from e
+        except asyncio.TimeoutError:
+            raise APIConnectionError("failed to connect to deepgram") from None
+        except aiohttp.ClientResponseError as e:
+            # RequestInfo carries the request headers, so chaining this error or
+            # formatting it puts the API key in the exception repr (#6739).
+            raise APIStatusError(
+                message=e.message, status_code=e.status, request_id=None, body=None
+            ) from None
+        except Exception as e:
+            raise APIConnectionError(
+                f"failed to connect to deepgram ({type(e).__name__})"
+            ) from None
         return ws
 
     def _on_audio_duration_report(self, duration: float) -> None:
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt_v2.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt_v2.py
@@ -580,8 +580,18 @@ class SpeechStreamv2(stt.SpeechStream):
                 "Established new Deepgram STT WebSocket connection:",
                 extra={"headers": ws_headers},
             )
-        except (aiohttp.ClientConnectorError, asyncio.TimeoutError) as e:
-            raise APIConnectionError("failed to connect to deepgram") from e
+        except asyncio.TimeoutError:
+            raise APIConnectionError("failed to connect to deepgram") from None
+        except aiohttp.ClientResponseError as e:
+            # RequestInfo carries the request headers, so chaining this error or
+            # formatting it puts the API key in the exception repr (#6739).
+            raise APIStatusError(
+                message=e.message, status_code=e.status, request_id=None, body=None
+            ) from None
+        except Exception as e:
+            raise APIConnectionError(
+                f"failed to connect to deepgram ({type(e).__name__})"
+            ) from None
         return ws
 
     def _on_audio_duration_report(self, duration: float) -> None:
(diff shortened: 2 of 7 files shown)


aniketwaghh:
Pushed faf17dd after a second look at the scope of the catch-all.

The first commit kept each plugin's existing `try` boundary and only swapped the handlers. In deepgram (`stt.py`, `stt_v2.py`) and smallestai that boundary also covers the statements that run after a successful connect, including a read of the private `ws._response` attribute. So an `AttributeError` there, from an aiohttp change rather than a connection problem, would have been converted into a retryable `APIConnectionError` and hidden behind a "failed to connect" message while the reconnect loop kept going. Before this PR it would have propagated. Narrowing each `try` to the connect call removes that and matches the scope cartesia guards in #6740, which wraps exactly one statement.

Two things I checked while I was in here, in case they come up in review.

`APIStatusError` already derives retryability from the status, so a 401 or 403 comes back non-retryable and the stream stops instead of reconnecting against a key that will keep failing. That is a change from today, where the handshake error escapes the handler entirely.

`ClientConnectorError` no longer has its own clause and now lands in the catch-all, so its message gains a `(ClientConnectorError)` suffix. Retryability is unchanged.

Redaction and lint are unchanged after the narrowing: 0 unguarded credentialed `ws_connect` sites, and the key is absent from both `traceback.print_exc()` and `logging.exception` output.

aniketwaghh:
On the header-logging flag at `stt.py:792-797`: that block predates this PR. It came in with #6444 and this branch only dedents it, moving it out of the `try` in faf17dd. Removed and added lines are identical once you strip indentation.

The convention point looks right to me though. `extra={"headers": ws_headers}` does not carry an `lk.pii.` prefix, so it is outside the redaction path that `lk.pii.dtmf_inputs` and `lk.pii.arguments` use elsewhere in the SDK. What lands there is Deepgram response headers filtered to `dg-*` plus `Date`, so request ids and model metadata rather than credentials or caller data, and it is behind `logger.debug`. Lower stakes than what this PR is about, and a different concern from the request-side leak.

Leaving it out to keep this PR to the credential path. Happy to open a separate one for the `lk.pii.` prefix if you want it, here and anywhere else `extra=` carries provider metadata.

luzmariatorres895-ops (changes requested):
Por

davidzhao (approved):

PR #7035, opened 2026-08-29
fix(mistral): inject dummy user when last turn is assistant

### LINKED ISSUE #7030: Mistral plugin fails when generating a reply with an assistant message as the last ChatContext item Description
### Bug Description

I am experiencing an issue with the Mistral LLM plugin in LiveKit Agents when the last message in the ChatContext has the assistant role.
The issue is reproducible in at least two different scenarios:
1. When a tool uses context.with_filler().
2. When calling session.generate_reply() without providing user_input.
In both cases, the Mistral API rejects the request with HTTP 400 / error code 3230.
The same behavior does not occur with the Ollama LLM plugin.

### Expected Behavior

session.generate_reply() should be able to generate a response even when the current ChatContext ends with an assistant message, provided that this is a valid LiveKit context.
For example:
user
assistant
followed by:
await session.generate_reply()
should result in a valid provider request.
Likewise, a filler message generated by context.with_filler() should not prevent the LLM from generating the response after a tool completes.

### Reproduction Steps

```bash
Reproduction 1: with_filler()
A simplified tool looks like this:
@function_tool
async def my_tool(context: RunContext):
    async with context.with_filler("One moment..."):
        result = await do_something()

    return result
The filler is added to the ChatContext as an assistant message.
After the tool returns, LiveKit attempts to generate the next LLM response.
At this point, the context can end with:
assistant: "One moment..."
The Mistral API then rejects the request:
Status 400

Expected last role User or Tool
(or Assistant with prefix True)
for serving but got assistant

type: invalid_request_error
code: 3230
Reproduction 2: generate_reply() without user_input
The issue can also be reproduced without with_filler().
For example, if the current ChatContext ends with an assistant message:
await session.generate_reply()
results in the same Mistral error:
Status 400

Expected last role User or Tool
(or Assistant with prefix True)
for serving but got assistant

type: invalid_request_error
code: 3230
However, adding a user_input makes the request valid:
await session.generate_reply(
    user_input="Please continue."
)
This suggests that the issue is related to generating a new response when the last message in the context is an assistant message.
```

### Operating System

macOS Tahoe 26.5.1

### Models Used

Deepgram/Mistral/ElevenLabs

### Package Versions

```bash
"livekit-api>=1.2.0",
    "livekit-plugins-anthropic>=1.7.1",
    "livekit-plugins-deepgram>=1.7.1",
    "livekit-plugins-elevenlabs>=1.7.1",
    "livekit-plugins-google>=1.7.1",
    "livekit-plugins-mistralai>=1.7.1",
    "livekit-plugins-openai>=1.7.1",
    "livekit-plugins-silero>=1.7.1",
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
Add a dummy user message as it is for the others providers that need it.
```

### Additional Context

_No response_

_No response_

leilei3167:
Mistral's Conversations API returns 400 / code 3230 when the last turn is an assistant message. That's a valid LiveKit context after `with_filler()` or `generate_reply()` with no `user_input`.

The other formatters already inject a dummy user turn for this. Mistral's path never took `inject_dummy_user_message`, so the request went up as-is.

Fixes #7030

## Test plan
- [x] `uv run pytest tests/test_chat_ctx.py --unit` (61 passed, 1 skipped)
- [x] `ruff check` / `ruff format` on the touched files

### DIFF
--- a/livekit-agents/livekit/agents/llm/_provider_format/mistralai.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/mistralai.py
@@ -16,8 +16,14 @@ class MistralFormatData:
 
 def to_conversations_ctx(
     chat_ctx: llm.ChatContext,
+    *,
+    inject_dummy_user_message: bool = True,
 ) -> tuple[list[dict], MistralFormatData]:
     """Convert ChatContext to Mistral Conversations API entry format.
+
+    Mistral requires the last serving turn to be User or Tool (HTTP 400 /
+    code 3230 otherwise). When ``inject_dummy_user_message`` is True, append
+    a dummy user turn if the context ends on an assistant message.
 
     Returns:
         A tuple of (entries, instructions) where instructions is the extracted
@@ -65,7 +71,20 @@ def to_conversations_ctx(
                 }
             )
 
+    if inject_dummy_user_message and not _last_serving_role_ok(entries):
+        entries.append({"type": "message.input", "role": "user", "content": "."})
+
     return entries, MistralFormatData(instructions=instructions)
+
+
+def _last_serving_role_ok(entries: list[dict[str, Any]]) -> bool:
+    """True when the last entry is a user message or a tool result."""
+    if not entries:
+        return False
+    last = entries[-1]
+    if last.get("type") == "function.result":
+        return True
+    return last.get("type") == "message.input" or last.get("role") == "user"
 
 
 def _to_entry(item: llm.ChatItem) -> dict[str, Any] | None:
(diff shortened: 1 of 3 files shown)


leilei3167:
@livekit/agent-devs small one for #7030 — Mistral was dropping `inject_dummy_user_message`, so a trailing assistant turn (filler / `generate_reply()` with no user_input) 400'd. Same dummy `.` user turn the Google formatter already uses.

davidzhao:
fixed in #7036

PR #7037, opened 2026-08-29
fix(mistralai): handle trailing assistant messages

### LINKED ISSUE #7030: Mistral plugin fails when generating a reply with an assistant message as the last ChatContext item Description

Shivakushwah143:
## Summary

Fixes #7030.

Mistral's Conversations API requires the conversation input to end with a user message or a tool result.

When a LiveKit `ChatContext` ended with an assistant message, the Mistral formatter produced a trailing `message.output`, causing Mistral to reject the request.

This change appends a minimal user input when the formatted Mistral conversation would otherwise end with an assistant message.

## Changes

- handle assistant-ending contexts in the Mistral provider formatter
- support `inject_dummy_user_message` for the Mistral provider format
- preserve existing behavior for user-ending and tool-result-ending contexts
- add regression tests covering:
  - trailing assistant message
  - trailing user message
  - trailing tool result
  - empty context

## Testing

- Mistral regression tests: **4 passed**
- `tests/test_chat_ctx.py`: **61 passed, 1 skipped**
- Ruff: **all checks passed**
- `git diff --check`: **passed**

The skipped test requires LiveKit API credentials and is unrelated to this change.

### DIFF
--- a/livekit-agents/livekit/agents/llm/_provider_format/mistralai.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/mistralai.py
@@ -16,8 +16,17 @@ class MistralFormatData:
 
 def to_conversations_ctx(
     chat_ctx: llm.ChatContext,
+    *,
+    inject_dummy_user_message: bool = True,
 ) -> tuple[list[dict], MistralFormatData]:
     """Convert ChatContext to Mistral Conversations API entry format.
+
+    Args:
+        chat_ctx: The chat context to convert.
+        inject_dummy_user_message: If ``True`` (the default), a dummy user message
+            (``"."``) is appended when the last entry is an assistant message.
+            The Mistral Conversations API requires the inputs list to end with a
+            ``message.input`` (user) or ``function.result`` (tool) entry.
 
     Returns:
         A tuple of (entries, instructions) where instructions is the extracted
@@ -64,6 +73,15 @@ def to_conversations_ctx(
                     "result": tool_output.output,
                 }
             )
+
+    # Mistral Conversations API requires the last entry to be a user message
+    # (message.input) or tool result (function.result), not an assistant message.
+    if (
+        inject_dummy_user_message
+        and entries
+        and entries[-1]["type"] == "message.output"
+    ):
+        entries.append({"type": "message.input", "role": "user", "content": "."})
 
     return entries, MistralFormatData(instructions=instructions)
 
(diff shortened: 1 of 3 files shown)


Shivakushwah143:
Hi! All CI checks are passing now. Happy to make any changes if needed — thanks for reviewing!

davidzhao:
fixed in #7036

PR #7039, opened 2026-08-29
fix(telemetry): prevent blocking shutdown offline

### LINKED ISSUE #7029: 1.7.1: telemetry export intermittently deadlocks offline processes (OTEL_SDK_DISABLED=true resolves)
### Description

After upgrading from livekit-agents 1.7.0 to 1.7.1, our offline test suite (pytest, no network egress for telemetry endpoints) began hanging **nondeterministically** — the same test selection would sometimes complete in ~7s and sometimes deadlock indefinitely. Bisection showed the hang was not tied to any specific test: individual tests always passed in isolation; only multi-file runs hung, and which combination hung varied between invocations.

Setting the standard OpenTelemetry opt-out resolves it completely and deterministically:

```
OTEL_SDK_DISABLED=true python -m pytest ...   # 98/98 in ~8.5s, every run
```

Without it, runs that previously took ~1s on 1.7.0 take ~6-7s when they complete at all, which looks like an exporter retry/backoff path, and occasionally never return.

### Environment
- livekit-agents 1.7.1 (plugins: google/deepgram/silero 1.7.1), Python 3.12.12, macOS 24.6.0 (arm64)
- No OTLP endpoint configured; no network egress for telemetry in the test context

### Expected
When no exporter/endpoint is configured (or export fails), telemetry should fail fast and silently — never block or add multi-second latency to the host process, and never require consumers to know about `OTEL_SDK_DISABLED`.

### Suggested direction
Bound telemetry export with a short timeout and drop-on-failure semantics, or default the SDK off unless an endpoint is explicitly configured.

Happy to provide more detail or run diagnostics against a branch.

Shivakushwah143:
## Summary

Fixes #7029.

OpenTelemetry providers register automatic `atexit` shutdown handlers by default.

When the OTLP endpoint is unavailable, those handlers can block process exit while waiting for exporter worker threads. LiveKit already has its own bounded telemetry shutdown path, so the SDK-owned `atexit` handlers are unnecessary.

This change disables automatic shutdown handlers for the tracer, logger, and meter providers created by LiveKit.

## Changes

- set `shutdown_on_exit=False` for `TracerProvider`
- set `shutdown_on_exit=False` for `LoggerProvider`
- set `shutdown_on_exit=False` for `MeterProvider`
- add regression coverage for:
  - providers not registering SDK-owned `atexit` handlers
  - offline process exiting without hanging

## Testing

- `tests/test_telemetry_shutdown.py`: 2 passed
- existing telemetry tests: 18 passed
- Ruff lint: passed
- Ruff format: passed
- `git diff --check`: passed

Normal telemetry exporting behavior is unchanged. The change only disables the SDK's automatic process-exit shutdown path.

### DIFF
--- a/livekit-agents/livekit/agents/telemetry/traces.py
+++ b/livekit-agents/livekit/agents/telemetry/traces.py
@@ -344,12 +344,18 @@ def _setup_cloud_tracer(
         # Check if a tracer provider is not set and set one up
         # below shows how the ProxyTracerProvider is returned when none have been setup
         # https://github.com/open-telemetry/opentelemetry-python/blob/0018c0030bac9bdce4487fe5fcb3ec6a542ec904/opentelemetry-api/src/opentelemetry/trace/__init__.py#L555
+        #
+        # shutdown_on_exit=False: the SDK registers atexit handlers that call
+        # provider.shutdown() → worker_thread.join(30s).  When the OTLP endpoint
+        # is unreachable (offline / no network), the worker threads may be stuck
+        # exporting and the join blocks process exit.  We manage shutdown
+        # explicitly via _shutdown_telemetry() with a bounded wall-clock timeout.
         tracer_provider: trace_api.TracerProvider
         if isinstance(
             tracer._tracer_provider,
             (trace_api.ProxyTracerProvider, trace_api.NoOpTracerProvider),
         ):
-            tracer_provider = trace_sdk.TracerProvider(resource=resource)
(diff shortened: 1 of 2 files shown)


Shivakushwah143:
All CI checks are passing now. The regression test is fully credential-independent, and the production fix remains unchanged. Ready for review — happy to adjust anything if needed.

longcw:
closing since already fixed in https://github.com/livekit/agents/pull/7033

PR #7041, opened 2026-08-29
fix(google): don't let closing the TTS input generator mask the real error

avionicharshit-byte:
`SynthesizeStream._run_stream` hands `input_gen` to the grpc client, which drives it from its own task. when the call fails, that task is usually still parked inside the generator waiting for the next sentence, so the `await input_gen.aclose()` in the `finally` raises

```
RuntimeError: aclose(): asynchronous generator is already running
```

and that replaces the `APITimeoutError` / `APIStatusError` the caller was meant to see. it shows up on interrupted turns and on deadline exceeded, so it reads as intermittent.

same race and same fix as #4766, which guarded the `aclose()` calls in `utils/aio/itertools.py`.

added `tests/test_plugin_google_tts_stream_close.py`: a fake client consumes `input_gen` from a background task, then raises `DeadlineExceeded`. it fails on main with the RuntimeError and passes with the guard.

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/tts.py
@@ -468,7 +468,12 @@ class SynthesizeStream(tts.SynthesizeStream):
         except GoogleAPICallError as e:
             raise APIStatusError(e.message, status_code=e.code or -1, body=f"{e.details}") from e
         finally:
-            await input_gen.aclose()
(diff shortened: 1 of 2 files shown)


avionicharshit-byte:
@longcw mind a look when you get a chance? small fix, guards `input_gen.aclose()` so it stops masking the real `APITimeoutError` / `APIStatusError`. same guard as #4766, regression test included. happy to rebase.

longcw (approved):
Thanks for the pr! looks good to me

PR #7043, opened 2026-08-30
fix: unblock avatar audio after lost stream trailer

mikemikimike:
## Summary

When an avatar audio stream loses its trailer, `DataStreamAudioReceiver` can remain blocked on that reader forever even after later audio streams arrive. This change closes the stale reader when a subsequent stream is received, allowing the receiver to emit the pending segment end and process queued audio.

## Testing

- `uv run --package livekit-agents --with pytest python -m pytest tests/test_datastream_io.py -q`
- `uv run --package livekit-agents ruff check livekit-agents/livekit/agents/voice/avatar/_datastream_io.py tests/test_datastream_io.py`
- `git diff --check`

### DIFF
--- a/livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
+++ b/livekit-agents/livekit/agents/voice/avatar/_datastream_io.py
@@ -420,7 +420,9 @@ class DataStreamAudioReceiver(AudioReceiver):
 
         # A new stream means the previous segment has ended. Close a reader whose
         # trailer was lost so its receive loop can advance to the new segment.
-        if self._current_reader is not None:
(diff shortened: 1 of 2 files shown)


mikemikimike:
Thanks for catching this. Fixed in 0a74af3 by closing the last queued reader when a new stream arrives, while preserving the active-reader fallback. Added a regression test covering an active reader followed by a queued reader and a new stream. Local ruff, format, and diff checks pass; the focused pytest dependency sync was blocked by slow wheel downloads, so CI is being used for the full test verification.

longcw:
closing since fixed in https://github.com/livekit/agents/pull/7007

PR #7045, opened 2026-08-30
fix: handle google/gemini-3.5-flash thought token leak in inference

### LINKED ISSUE #6942: Inference: `google/gemini-3.5-flash` leaks its thought part into content (no closing delimiter)
## Bug Description

Through LiveKit Inference, `google/gemini-3.5-flash` streams Gemini's **thought part as ordinary `delta.content`** at the start of an assistant turn, so it reaches `llm_node` → TTS / `transcription_node` / session history and is spoken to the caller. We have not seen it on the native `livekit-plugins-google` path with the same model, so it looks specific to the OpenAI-shaped gateway (`inference.LLM`).

This is the same class of leak as #6375 (Gemma 4), but it cannot be fixed the same way: **there is no closing delimiter**, so no `_MODEL_THINK_TAGS` start/end pair expresses it.

### Sampled grammar

We instrumented `llm_node` to record every turn's raw content chunks and ran a 41-case voice-agent regression suite (287 assistant turns with content). **27 turns (9.4%) opened with a leaked marker.** Verbatim shapes, with our own domain data replaced by placeholders:

```text
thought\n
thought\nI see you're calling from the number we have on file. Am I speaking with <name>?
thoughtI see you're calling from the number we have on file. Am I speaking with <name>?
thought[]I see you're calling from the number we have on file. Am I speaking with <name>?
thought() I can help you with that. Since you need a return visit ...
thought [] Muddy waters. I have the phone number <phone>. Under instructions, I must say "I see ...
thought absenteeism: None. The prior tool `<tool_name>` is just a mock tool or marker. I should ...
thought() Note: First attempt returned no slots. Let me ask the caller if they want to try a ...
thoughtLevel: 0 Elided Double Blueprints (skip checking)\n- Wait for their response before ...
thoughtLevel: 257766579840.0ms0.000sI'll get that scheduled for you. Let me check what we have ...
```

Two things make this unfixable from the app side:

1. **No stable delimiter after the `thought` token.** The speech follows directly, or after `\n`, or after an empty `[]` or `()`, with or without surrounding spaces. Since `thoughtI see you're calling` occurs, not even a word boundary can be required — which in turn means an app-side filter risks eating real speech that happens to begin with "thoughtful"/"thoughts".

2. **8 of the 27 leaks carried a non-empty thought body**, with no delimiter separating it from the speech that follows (see the last four lines above). No prefix rule can tell `thoughtLevel: 257766579840.0ms0.000s` + `I'll get that scheduled for you.` apart from content the model meant to say. **These are unconditionally spoken to the caller.**

### No signal to filter on

The leaking chunks carry no distinguishing metadata — `ChoiceDelta.extra` is `None` on every one of them. `extra` *is* populated on neighbouring chunks in the same stream, e.g.:

```text
{'google': {'thought_signature': '...'}}
{'livekit': {'inference_deployment': 'standard_vertex', 'inference_tier_requested': 'standard', ...}}
```

So there is no `reasoning` / `reasoning_content` field and no flag the SDK could key off; the only available lever is a text heuristic on spoken output.

Observed across all three deployments the gateway routed us to in this sweep: `standard_aistudio`, `standard_vertex`, `us_vertex`.

## Expected Behavior

The thought part should never reach `ChatChunk.delta.content`. Either:

- the gateway does not serialize the thought part into the content stream for `google/gemini-3.5-flash` (preferred — matches the native google plugin's behaviour), or
- it is delivered on a distinct field (`delta.reasoning_content`, or a flag in `delta.extra`) so `inference/llm.py` can drop it deterministically instead of pattern-matching prose.

A `_MODEL_THINK_TAGS` entry is not sufficient here: `ThinkingTokenFilter` strips a start/end **pair**, and this marker has no end tag.

## Reproduction Steps

```bash
1. AgentSession with `inference.LLM(model="google/gemini-3.5-flash")` and function tools registered.
2. Drive a multi-turn conversation in which the model calls a tool and then speaks.
3. Record the raw `delta.content` of each chunk in `llm_node` before it reaches TTS:

     class Probe(Agent):
         async def llm_node(self, chat_ctx, tools, model_settings):
             async for chunk in Agent.default.llm_node(self, chat_ctx, tools, model_settings):
                 if isinstance(chunk, llm.ChatChunk) and chunk.delta and chunk.delta.content is not None:
                     print(repr(chunk.delta.content))
                 yield chunk

4. Inspect turns that follow a tool call: a share of them open with a chunk whose
   content is exactly "thought", followed by the separator shapes listed above.
```

The rate was ~9% of assistant turns over 287 turns, so a few dozen turns are usually enough to see it. It is not deterministic per conversation.

## Workaround

App-side anchored prefix strip in `llm_node`, over the turn's buffered chunks (the token routinely arrives in its own chunk, so a per-chunk filter is not enough):

```python
_THOUGHT_PREFIX = re.compile(r"^thought\s*(?:\[[^\]]*\]|\([^)]*\))?\s*(?![a-z])")
```

Anchoring to the start of the turn, plus the "not a lowercase letter" lookahead, is what keeps ordinary speech safe — every leak in the sweep was followed by a capitalised sentence or by nothing, whereas a turn opening "thoughtful of you to call" is left alone.

This clears 19 of the 27 leaks. The 8 with a non-empty body still get spoken, which is why we are filing rather than only patching locally.

## Environment

- `livekit-agents` 1.6.9
- Model: `inference.LLM(model="google/gemini-3.5-flash")` via LiveKit Inference (no `provider` pin), `temperature=0.0`, `reasoning_effort` unset
- macOS 15 (Darwin 25.6.0), Python 3.13

## Related

- #6375 — Gemma 4 `<|channel>thought…<channel|>` markers via Inference (same class, but had a closing tag)
- #4190 — Gemini thought summaries in the native google plugin
- #6323 — `strip_thinking_tokens` list-delta crash

HoneyTyagii:
Add _MODEL_THINK_TAGS entry for google/gemini-3.5-flash with empty end tag to match the observed behavior where the thought token has no closing delimiter. The strip_thinking_tokens function already handles empty end tags correctly by treating the remainder as visible content after the start tag is found.

Fixes issue where thought tokens were leaked into delta.content and spoken to callers via TTS/transcription.

Closes #6942

### DIFF
--- a/livekit/agents/inference/llm.py
+++ b/livekit/agents/inference/llm.py
@@ -1,0 +1,606 @@
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass
+from typing import Any, Literal, cast
+
+import httpx
+import openai
+from openai.types.chat import (
+    ChatCompletionChunk,
+    ChatCompletionMessageParam,
+    ChatCompletionPredictionContentParam,
+    ChatCompletionToolChoiceOptionParam,
+    ChatCompletionToolParam,
+    completion_create_params,
+)
+from openai.types.chat.chat_completion_chunk import Choice
+from openai.types.shared.reasoning_effort import ReasoningEffort
+from openai.types.shared_params import Metadata
+from typing_extensions import TypedDict
+
+from .. import llm
+from .._exceptions import APIConnectionError, APIStatusError, APITimeoutError
+from ..llm import ToolChoice, utils as llm_utils
+from ..llm.chat_context import ChatContext
+from ..llm.tool_context import Tool
+from ..log import logger
+from ..types import DEFAULT_API_CONNECT_OPTIONS, NOT_GIVEN, APIConnectOptions, NotGivenOr
+from ..utils import is_given
+from ._utils import (
+    HEADER_INFERENCE_PROVIDER,
+    create_access_token,
+    extract_quota_usage,
+    get_default_inference_url,
+    get_inference_headers,
+)
+
+lk_oai_debug = int(os.getenv("LK_OPENAI_DEBUG", 0))
+
+# Reasoning models don't support sampling parameters.
+# See: https://platform.openai.com/docs/guides/reasoning
+_REASONING_UNSUPPORTED_PARAMS: set[str] = {
+    "temperature",
+    "top_p",
+    "presence_penalty",
+    "frequency_penalty",
+    "logit_bias",
+    "logprobs",
+    "top_logprobs",
+    "n",
+}
+
+# xAI reasoning models only restrict presence_penalty, frequency_penalty, stop.
+# They still support temperature and top_p.
+_XAI_REASONING_UNSUPPORTED_PARAMS: set[str] = {
+    "presence_penalty",
+    "frequency_penalty",
+    "stop",
+}
+
+# Model prefix -> set of param names that should be dropped
+_UNSUPPORTED_PARAMS: dict[str, set[str]] = {
+    "o1": _REASONING_UNSUPPORTED_PARAMS,
+    "o3": _REASONING_UNSUPPORTED_PARAMS,
+    "o4": _REASONING_UNSUPPORTED_PARAMS,
+    "gpt-5": _REASONING_UNSUPPORTED_PARAMS,
+    "grok-4-1-fast-reasoning": _XAI_REASONING_UNSUPPORTED_PARAMS,
+    "grok-4.20-0309-reasoning": _XAI_REASONING_UNSUPPORTED_PARAMS,
+    "grok-4.20-multi-agent": _XAI_REASONING_UNSUPPORTED_PARAMS,
+}
+
+# models that don't support reasoning_effort when function tools are present
+_REASONING_EFFORT_TOOL_INCOMPATIBLE_PREFIXES: set[str] = {"gpt-5.2", "gpt-5.4"}
+
+_MODEL_THINK_TAGS = {
+    "google/gemma-4-31b-it": ("<|channel>thought", "<channel|>"),
+        "google/gemini-3.5-flash": ("thought", ""),
+}
+
+
+def drop_unsupported_params(
+    model: str, params: dict[str, Any], tools: list[Any] | None = None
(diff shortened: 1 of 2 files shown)


STiFLeR7 (commented):
Re-checked this against the leak samples in #6942 and I think there's a residual gap in `strip_thinking_tokens`'s `end_tag == ""` branch:

```python
idx = state._buffer.find(state.start_tag)
if idx >= 0:
    visible.append(state._buffer[:idx])
    state._buffer = state._buffer[idx + len(state.start_tag) :]
    if state.end_tag == "":
        # For empty end tag, we do not look for an end tag: the rest is visible.
        visible.append(state._buffer)
        state._buffer = ""
        break
```

This strips only the literal 4-character marker `"thought"` and appends *everything after it* as visible — it doesn't consume the separator artifact that follows in the real leak shapes #6942 sampled:

```text
thought[]I see you're calling from the number we have on file...
thought() I can help you with that...
thought [] Muddy waters...
```

After this filter, those become `[]I see you're calling...`, `() I can help you with that...`, ` [] Muddy waters...` — the `"thought"` word is gone, but the `[]`/`()`/leading-space artifact immediately after it is still spoken/shown, since it's unconditionally appended along with the real content.

The issue body's own proposed app-side workaround already accounts for this shape:

```python
_THOUGHT_PREFIX = re.compile(r"^thought\s*(?:\[[^\]]*\]|\([^)]*\))?\s*(?![a-z])")
```

i.e. it strips `"thought"` *plus* an optional trailing `[...]`/`(...)` and whitespace, not just the bare word. Might be worth folding an equivalent pattern into the `end_tag == ""` branch here — e.g. matching `start_tag` plus an optional `\s*(?:\[[^\]]*\]|\([^)]*\))?\s*` immediately after it — so the artifact doesn't leak through even when this PR's fix is applied. (The 8/27 cases with a non-empty thought body and no separator at all are correctly called out in the issue as unfixable without a real delimiter — this is specifically about the empty-body cases where a separator artifact is present and could be stripped.)

longcw:
closing since it's fixed in gateway

PR #7050, opened 2026-08-31
fix(inworld): keep a reference to the prewarm task and cancel it on close

Rehansanjay:
## The bug

`TTS.prewarm()` throws away the task it creates:

```python
def prewarm(self) -> None:
    asyncio.create_task(self._prewarm_impl())
```

Two consequences.

**The task is only weakly referenced.** The event loop does not keep a strong reference to a bare task, so it can be garbage collected mid-execution and the prewarm silently never happens.

**The ordering one is worse.** `_prewarm_impl` calls `_get_pool`, which builds a fresh `_ConnectionPool` whenever `self._pool` is None:

```python
async def _get_pool(self) -> _ConnectionPool:
    async with self._pool_lock:
        if self._pool is None or self._pool._closed:
            self._pool = _ConnectionPool(...)
        return self._pool
```

and `aclose()` sets `self._pool = None` after closing it. So a prewarm still in flight when the TTS is closed **recreates the pool after close**, leaving a live websocket connection that nothing owns and nothing will ever close.

## The fix

Store the task and cancel it at the top of `aclose()`, before the pool is torn down.

This is the pattern already in the tree. `openai/tts.py` hand-rolls its prewarm the same way and holds the task:

```python
def prewarm(self) -> None:
    ...
    self._prewarm_task = asyncio.create_task(_prewarm())

    if self._prewarm_task:
        await aio.cancel_and_wait(self._prewarm_task)
```

and `utils.ConnectionPool.prewarm()` — used by the fifteen other plugins — keeps `self._prewarm_task` and checks whether it is still running before starting another.

Inworld was the only plugin calling `asyncio.create_task` in `prewarm()` without holding on to the result.

## Checks

`ruff check` and `ruff format --check` pass on the changed file. No behaviour change on the success path: prewarm still runs in the background and is still fire-and-forget from the caller's point of view.

---

Found with a small AST pass over the repo looking for `create_task` calls whose result is discarded. Happy to send the other instances separately if useful.

### DIFF
--- a/livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
+++ b/livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
@@ -42,7 +42,7 @@ from livekit.agents.types import (
     APIConnectOptions,
     NotGivenOr,
 )
-from livekit.agents.utils import is_given
+from livekit.agents.utils import aio, is_given
 from livekit.agents.voice.io import TimedString
 
 from .log import logger
@@ -964,6 +964,7 @@ class TTS(tts.TTS):
         self._idle_connection_timeout = idle_connection_timeout
         self._pool: _ConnectionPool | None = None
         self._pool_lock = asyncio.Lock()
+        self._prewarm_task: asyncio.Task[None] | None = None
         self._streams = weakref.WeakSet[SynthesizeStream]()
         self._sentence_tokenizer = (
             tokenizer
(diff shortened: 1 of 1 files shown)


Rehansanjay:
Bumping — this is the same leak you merged in #7023, one plugin over.

`inworld`'s `TTS.prewarm()` discards its task. Beyond the weak-reference hazard, there is an ordering bug next to it: `_prewarm_impl` calls `_get_pool`, which builds a new `_ConnectionPool` whenever `self._pool` is None, and `aclose()` sets it to None after closing. A prewarm still in flight when the TTS closes therefore rebuilds the pool afterwards, leaving a live connection that nothing owns and nothing will close.

The fix follows what is already in the tree: `openai/tts.py` stores its hand-rolled prewarm task and cancels it in `aclose()`, and `utils.ConnectionPool.prewarm()` does the same for the fifteen plugins that use it. inworld was the only plugin doing neither.

One file, +10/-2, all checks green, no open review threads.

cc @davidzhao — same shape as #7023.

davidzhao (approved):

PR #7051, opened 2026-08-31
fix(soniox): keep references to the prewarm and self-close tasks

Rehansanjay:
## The bug

Five `asyncio.create_task` calls in `soniox/tts.py` discard the task they create. The event loop only holds a **weak** reference to a bare task, so it can be garbage collected before it finishes.

### `_Connection` self-close (four sites)

The connection closes itself from `mark_non_current`, `unregister_stream`, and the `finally` of both `_send_loop` and `_recv_loop`:

```python
asyncio.create_task(self.aclose(), name="soniox-tts-conn-drain-close")
```

If that task is collected, `aclose()` never runs and the WebSocket stays open.

Worth noting the class already gets this right for its own loops — `_send_task`, `_recv_task` and `_keepalive_task` are all stored on the instance. The close tasks were the exception.

### `TTS.prewarm` — an ordering problem too

```python
async def _task() -> None:
    await self._current_connection(timeout=20.0)

asyncio.create_task(_task(), name="soniox-tts-prewarm")
```

`_current_connection` opens a new `_Connection` when there is none, and `aclose()` sets `__current_connection = None` after closing. So a prewarm still in flight when the TTS is closed **reconnects afterwards**, leaving a live WebSocket that nothing owns and nothing will close.

## The fix

The four self-close sites go through a small helper that keeps the task and skips scheduling a second close while one is still running:

```python
def _schedule_close(self, name: str) -> None:
    if self._close_task is not None and not self._close_task.done():
        return
    self._close_task = asyncio.create_task(self.aclose(), name=name)
```

`prewarm` stores its task, and `aclose()` cancels it before tearing the connection down — the same shape as `openai/tts.py`, and what `utils.ConnectionPool.prewarm()` does for the plugins that use it.

No behaviour change on the success path.

## Checks

`ruff check` and `ruff format --check` pass on the changed file.

---

Found with a small AST pass looking for `create_task` calls whose result is discarded — the same sweep that turned up #7050.

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
@@ -154,6 +154,7 @@ class TTS(tts.TTS):
 
         # One persistent connection shared across streams (see _Connection).
         self.__current_connection: _Connection | None = None
+        self._prewarm_task: asyncio.Task[None] | None = None
         self.__conn_lock = asyncio.Lock()
 
     @property
@@ -251,13 +252,20 @@ class TTS(tts.TTS):
                 logger.debug(f"Soniox TTS prewarm failed: {e}")
 
         try:
-            asyncio.create_task(_task(), name="soniox-tts-prewarm")
+            self._prewarm_task = asyncio.create_task(_task(), name="soniox-tts-prewarm")
         except RuntimeError:
             # No running event loop (e.g. called outside async context) — skip.
             pass
 
     async def aclose(self) -> None:
         """Close all streams and the persistent connection."""
+        # Cancel first: the prewarm task calls _current_connection, which opens a new
+        # connection when there is none. A prewarm still in flight here would otherwise
+        # reconnect after close, leaving a live WebSocket nothing owns.
+        if self._prewarm_task is not None:
+            await utils.aio.cancel_and_wait(self._prewarm_task)
+            self._prewarm_task = None
+
         for stream in list(self._streams):
             await stream.aclose()
         self._streams.clear()
@@ -579,6 +587,7 @@ class _Connection:
         self._send_task: asyncio.Task[None] | None = None
         self._recv_task: asyncio.Task[None] | None = None
         self._keepalive_task: asyncio.Task[None] | None = None
+        self._close_task: asyncio.Task[None] | None = None
         self._is_current = True
         self._closed = False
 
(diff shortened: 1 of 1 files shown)


Rehansanjay:
Both review points addressed in a1569b1.

**Repeated prewarm leaks a connection** — correct, and it undercut the point of the PR. `prewarm()` replaced `_prewarm_task` unconditionally, so a second call dropped the only reference to a task that was still running, and `aclose()` cancels just the latest one. It now starts a task only when there is no live one, which is how `utils.ConnectionPool.prewarm` guards the same case.

**Prewarm errors bypass log redaction** — moved to `exc_info`, matching the send and recv loops in this same file.

Rehansanjay:
On the remaining open thread — the `exc_info` one — I do not think it is actionable, and I would rather say so than quietly churn.

The earlier review on this series flagged the opposite problem: the prewarm failure interpolated the exception into the message body, where redaction cannot reach it. I moved it to `exc_info`, which is what the send and recv loops in this same file already do. The new finding says `exc_info` is itself the problem.

Taken literally that leaves nowhere to put the exception, and it would put this call site out of step with the rest of the codebase: `exc_info=` appears at 99 call sites across `livekit-agents` and `livekit-plugins` today. Dropping the exception entirely would mean a prewarm or a recycle can fail and leave an operator with a log line and no cause.

If the project does want a stricter rule for provider exceptions — a redacting formatter, or an `lk.pii.` structured field rather than `exc_info` — I am happy to follow it here and it would probably be worth applying repo-wide rather than to these two lines. Left as-is for now pending your view.

Rehansanjay:
Bumping — this is the same leak class as #7023, #7012 and #7050, all now merged.

`_Connection` self-closes from four places: `mark_non_current`, `unregister_stream`, and the `finally` of both the send and recv loops. Each discarded the task that closes the WebSocket, and the loop holds only a weak reference to a bare task, so a collected one leaves the socket open. The class already gets this right for its own `_send_task`, `_recv_task` and `_keepalive_task` — the close tasks were the exception. They now go through a `_schedule_close` helper that also skips scheduling a second close while one is in flight.

`TTS.prewarm` had an ordering problem alongside it: its task calls `_current_connection`, which opens a new connection when there is none, while `aclose()` sets `__current_connection` to None. A prewarm still in flight at close therefore reconnected afterwards, leaving a live WebSocket nothing owns. It is now cancelled at the top of `aclose()`, and `prewarm()` no longer replaces a task that is still running.

One file, +29/-6, all checks green.

To be upfront: there is one unresolved review thread, on logging the prewarm failure through `exc_info`. My reasoning is in the thread — `exc_info=` is used at 99 call sites across the repo, and the earlier review on this series asked me to move *to* it from interpolating the exception into the message body. Happy to follow a different convention if the project wants one, but I did not want to change it unilaterally.

cc @chenghao-mou — you built the connection rotation in #6434, and `mark_non_current` is one of the four sites here.

Rehansanjay:
Closing in favour of #7054, at @longcw's request — that PR carries this change unchanged and is being reviewed as one.

PR #7054, opened 2026-08-31
chore(lint): enable RUF006 so dangling asyncio tasks fail CI

Rehansanjay:
> **Stacked on #7050, #7051, #7052 and #7053.** Those four are the bug fixes; this PR is the one-line config change plus the last remaining fix. It carries their commits so CI is green here and now — as each of them merges, this diff shrinks on its own. If you would rather land the four first and take this as a two-file PR afterwards, that works too, just say so.

## Why

The event loop keeps only a **weak** reference to a task. A bare `asyncio.create_task(...)` or `asyncio.ensure_future(...)` whose result is thrown away can therefore be garbage collected before it finishes, and the failures are quiet — a websocket that never closes, a `session.update` the server never receives, a tool set that silently never applies.

Ruff has a rule for precisely this: **RUF006, `asyncio-dangling-task`**. The repo's select list is

```toml
select = ["E", "W", "F", "I", "B", "C4", "UP"]
```

so `RUF` is absent and the rule has never run here.

## What it found

Fifteen findings across the repo:

| Where | Count | Fixed in |
|---|---|---|
| `livekit-plugins-soniox/…/tts.py` | 5 | #7051 |
| `livekit-plugins-aws/…/realtime_model.py` | 2 | #7052 |
| `livekit-agents/…/inference/stt.py` | 1 | #7053 |
| `livekit-plugins-inworld/…/tts.py` | 1 | #7050 |
| `livekit-plugins-hamming/…/_plugin.py` | 1 | this PR |
| `examples/` (drive_thru, frontdesk, translation ×2) | 4 | this PR |
| `tests/test_ipc.py` | 1 | this PR |

The examples matter more than their line count suggests — people copy them, so an example that drops a task teaches the bug. Each now holds the task in a set that discards on completion, which is the pattern the asyncio docs ask for.

## Why `RUF006` and not `RUF`

Enabling the whole ruleset adds roughly 460 unrelated findings — `RUF100` unused-noqa alone is 251, `RUF022` unsorted-dunder-all is 77. That is a separate conversation and I did not want to smuggle it in here.

## Checks

- `ruff check` passes clean across both trees with the rule enabled.
- Verified the rule still fires when a dangling task is reintroduced, so this is enforcing rather than passing vacuously.
- `ruff format --check` clean on the changed file.

One note in case it is useful to others: `ensure_future` has the same weak-reference semantics as `create_task`, and RUF006 covers both. A hand-rolled grep or AST scan looking only for `create_task` misses them — that is how the core `inference/stt.py` case in #7053 stayed hidden.

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
@@ -553,6 +553,10 @@ class RealtimeSession(  # noqa: F811
         # Session recycling: proactively restart before credential expiry or 8-min limit
         self._session_start_time: float | None = None
         self._session_recycle_task: asyncio.Task[None] | None = None
+        # Held only so the loop's weak reference cannot collect them mid-flight.
+        # Their lifecycle - cancellation, coalescing, shutdown - is #7052's subject.
+        self._user_text_tasks: set[asyncio.Task[None]] = set()
+        self._deferred_tool_recycle_tasks: set[asyncio.Task[None]] = set()
         self._last_audio_output_time: float = 0.0  # Track when assistant last produced audio
         self._audio_end_turn_received: bool = False  # Track when assistant finishes speaking
         self._pending_generation_fut: asyncio.Future[llm.GenerationCreatedEvent] | None = None
@@ -1865,7 +1869,11 @@ class RealtimeSession(  # noqa: F811
                                 if self._pending_generation_fut is fut:
                                     self._pending_generation_fut = None
 
-                        asyncio.create_task(_send_user_text())
+                        task = asyncio.create_task(
+                            _send_user_text(), name="RealtimeSession._send_user_text"
+                        )
+                        self._user_text_tasks.add(task)
+                        task.add_done_callback(self._user_text_tasks.discard)
 
                     self._sent_message_ids.add(item.id)
                     self._chat_ctx.items.append(item)
@@ -1931,7 +1939,11 @@ class RealtimeSession(  # noqa: F811
                 f"[SESSION] Tools changed (added={new_tools - old_tools}, "
                 f"removed={old_tools - new_tools}), scheduling deferred session recycle"
             )
-            asyncio.create_task(self._deferred_tool_recycle())
+            task = asyncio.create_task(
+                self._deferred_tool_recycle(), name="RealtimeSession._deferred_tool_recycle"
+            )
+            self._deferred_tool_recycle_tasks.add(task)
+            task.add_done_callback(self._deferred_tool_recycle_tasks.discard)
         else:
             logger.debug("Tool list updated locally")
 
--- a/livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
+++ b/livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
@@ -1088,7 +1088,11 @@ class TTS(tts.TTS):
         return self._session
 
     def prewarm(self) -> None:
-        self._prewarm_task = asyncio.create_task(self._prewarm_impl())
+        # Don't replace a prewarm still in flight: the old task would lose its only
+        # reference, and aclose() cancels just the latest one, so it could build a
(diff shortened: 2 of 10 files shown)


livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/experimental/realtime/realtime_model.py
```diff
@@ -553,6 +553,17 @@ def __init__(self, realtime_model: RealtimeModel) -> None:
         # Session recycling: proactively restart before credential expiry or 8-min limit
         self._session_start_time: float | None = None
         self._session_recycle_task: asyncio.Task[None] | None = None
+        self._deferred_tool_recycle_task: asyncio.Task[None] | None = None
+        # True once a recycle is past its debounce and into teardown/restart, where
+        # cancelling would strand the session; pending coalesces updates arriving then.
+        self._tool_recycle_running = False
+        self._tool_recycle_pending = False
+        # Permanent, unlike _is_sess_active, which a recycle clears and sets again.
+        self._closing = False
```
longcw: Could you cut the aws part of #7054 down to the first commit, which only holds the two task handles. The other commits rewrite session shutdown — shield, idempotency, a re-entrant recycle loop — and that is a behavior change to the plugin, not a lint fix. Keep it in #7052 for its own review.

livekit-plugins/livekit-plugins-inworld/livekit/plugins/inworld/tts.py
```diff
@@ -1087,7 +1088,7 @@ def _ensure_session(self) -> aiohttp.ClientSession:
         return self._session
 
     def prewarm(self) -> None:
-        asyncio.create_task(self._prewarm_impl())
+        self._prewarm_task = asyncio.create_task(self._prewarm_impl())
```
longcw: here still replaces `_prewarm_task` on a second prewarm, soniox has the guard, and inworld needs it too

longcw (changes requested):
Thanks — the rule is right and the survey is accurate. Since #7054 already carries #7050, #7051 and #7053 unchanged, let us review it alone and close those three. and some comments below, main for the aws plugin

Rehansanjay:
Thanks — all three done in 7fe421a.

**Consolidated.** #7050 merged in the meantime, so it has dropped out of the stack on its own. #7051 and #7053 are unchanged here and I am closing both, pointing at this PR.

**aws cut back.** Worth flagging that the first commit was not purely handles — it also added a replace-in-place cancel in `update_tools` and cancellation in `aclose`, both behaviour changes. So rather than cherry-pick it as-is I rewrote the aws part to hold the handles and nothing else: two sets, add on create, discard on completion. No cancellation, no coalescing, no change to shutdown. Everything about their lifecycle stays in #7052 for its own review, which I think is what you were asking for.

**inworld guard.** Correct, and it undercut the point of #7050. Added as a follow-up commit here, matching the soniox guard and `utils.ConnectionPool.prewarm`.

One consequence to flag: #7052 currently builds on the older aws commit, so once this lands it will need a rebase. Happy to do that as soon as this is settled — I would rather not churn it while you are looking at this one.

Verified after the rework: `ruff check .` clean across the repo with the rule enabled, and it still errors on a reintroduced dangling task, so the rule is enforcing rather than passing vacuously.

Rehansanjay:
One more, in 50289ab — and I want to flag it as a judgment call against the line you drew, so you can push back.

Devin caught that `push_cart` in `examples/drive_thru` gates on `push_running`, which is set *inside* `_push_runner` and so is not set until the loop schedules it. Several order changes in one tick each saw it clear and started a runner of their own, and concurrent runners can deliver a stale cart after a newer one.

I fixed it rather than only holding the handle, because here the two are the same thing: the task retained for RUF006 exists as soon as `create_task` returns, so using it as the guard removes `push_running` entirely — the commit is a net deletion. `push_pending` still carries the trailing edge.

It is still a behaviour change, so if you would rather this PR only hold the reference and leave the race for its own change, say so and I will cut it back. My reasoning for treating it differently from the aws commits: that was a rewrite of session shutdown in a plugin, this is a one-line guard in an example, and examples get copied.

`ruff check .` is clean repo-wide with the rule on, and still errors on a reintroduced dangling task.

longcw (approved):

PR #7056, opened 2026-08-31
Sunset Sarvam TTS model bulbul:v2 and remove bulbul:v3-beta

dhruvladia-sarvam:
### Summary

Removes the deprecated bulbul:v2 and bulbul:v3-beta models from the Sarvam TTS plugin, leaving bulbul:v3 as the sole supported model. This simplifies the codebase by eliminating model-conditional branching and dead v2-only parameters.

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
@@ -69,7 +69,7 @@ _WS_HEARTBEAT_INTERVAL: float = 20.0
 _KEEPALIVE_INTERVAL: float = 30.0
 
 # Sarvam TTS specific models and speakers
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
```diff
@@ -465,16 +381,12 @@ def __init__(
         speaker: SarvamTTSSpeakers | str | None = None,
         speech_sample_rate: int = 22050,
         num_channels: int = 1,  # Sarvam output is mono WAV
-        pitch: float = 0.0,
```
tinalenguyen, line 468: to maintain backward compatibility, let's accept kwargs and throw a warning when pitch, loudness, enable_preprocessing, and enable_cached_responses are set

dhruvladia-sarvam: @tinalenguyen not required as the bulbul:v2 model is deprecated and moved on from completely

PR #7059, opened 2026-08-31
fix(voice): clear committed turn on resumed speech

### LINKED ISSUE #7010: Manual mode: VAD START_OF_SPEECH cancels a committed user turn and strands _user_turn_committed

Shivakushwah143:
## Summary

Fixes #7010.

In manual turn-detection mode, `commit_user_turn()` sets
`_user_turn_committed = True` while the end-of-turn task is running.

If VAD emits `START_OF_SPEECH` before that task completes, the task is
cancelled but `_user_turn_committed` remains `True`.

Because the cancelled task never reaches its normal cleanup path, the
manual-mode guard in `_on_stt_event()` then drops subsequent transcript
events.

This change clears `_user_turn_committed` when `START_OF_SPEECH`
cancels the pending end-of-turn task.

## Changes

- Clear `_user_turn_committed` when resumed speech cancels the pending EOU task.
- Add regression coverage verifying the committed flag is cleared.
- Add regression coverage verifying subsequent `FINAL_TRANSCRIPT` events are not dropped.

## Testing

Added focused regression tests in
`tests/test_audio_recognition_turn_detection.py`.

Local execution was blocked by the current development environment /
dependency setup, but the modified Python files pass syntax validation and
the patch is intentionally limited to the affected state transition.


Rehansanjay (commented):
Not a maintainer — flagging something mechanical before anyone spends review time on it, because it is easy to fix and would otherwise be the first thing anyone says.

Your actual change is small and readable: **one line in `voice/audio_recognition.py` and 57 lines of test.** But the PR carries twelve other files:

```
examples/drive_thru/bg_noise.mp3                              +0/-0
livekit-agents/livekit/agents/resources/city-ambience.ogg     +0/-0
livekit-agents/livekit/agents/resources/crowded-room.ogg      +0/-0
livekit-agents/livekit/agents/resources/forest-ambience.ogg   +0/-0
livekit-agents/livekit/agents/resources/hold_music.ogg        +0/-0
livekit-agents/livekit/agents/resources/keyboard-typing.ogg   +0/-0
livekit-agents/livekit/agents/resources/keyboard-typing2.ogg  +0/-0
livekit-agents/livekit/agents/resources/office-ambience.ogg   +0/-0
livekit-plugins/.../silero/resources/silero_vad.onnx          +0/-0
tests/change-sophie.opus                                      +0/-0
tests/change-sophie.wav                                       +0/-0
tests/long.mp3                                                +0/-0
```

All `+0/-0`, all binary. `.gitattributes` in this repo is:

```
*.mp3 filter=lfs diff=lfs merge=lfs -text
*.ogg filter=lfs diff=lfs merge=lfs -text
```

so `.mp3` and `.ogg` are Git LFS. That set of files showing up with no line changes is the usual signature of a clone made without `git-lfs` installed: checkout writes the pointer text into the working tree, and the next commit records it as a change to the file.

Fix is `git lfs install`, then re-clone or `git lfs pull`, then drop those paths from the branch — `git checkout origin/main -- <paths>` and amend. After that the diff should be the two files you actually touched.

Worth doing before review rather than after: right now the file list makes this look like a 14-file change to core resources, which is not what it is, and the one-line fix plus its test is the part that deserves attention.

PR #7060, opened 2026-08-31
Add Arize AX to OpenTelemetry examples

arizedatngo:
## Summary
- Add Arize AX to the list of OTLP-compatible backends in the voice agent OpenTelemetry example.
- Include the Arize AX OTLP/HTTP endpoint, required headers, and project metadata example.
- Update the agent observability knowledge-base text to name multiple OTLP-compatible backends.

## Notes
This is a docs/example-only update. The existing example already supports arbitrary OTLP/HTTP endpoints and headers through `setup_otel`; this adds provider-specific guidance for users sending LiveKit agent traces to Arize AX.

## Testing
- `git diff --check`


arizedatngo:
Addressed the Devin review item by setting `openinference.project.name` as an OpenTelemetry resource attribute instead of LiveKit metadata. That keeps Arize project selection on the tracer provider/resource path, while session metadata remains reserved for LiveKit job/session attributes.

arizedatngo:
@chenghao-mou hi, looks like you recently worked on the OpenTelemetry example and reviewed the homepage voice-agent example. If you are the right reviewer for this area, could you take a look when you have a chance? CLA, ruff, type-checks, unit tests, and Devin Review are passing now; the earlier Devin concern was addressed by moving the Arize project name to an OpenTelemetry resource attribute.

arizedatngo:
@tinalenguyen could you review or route this OpenTelemetry example update? I saw you merged the recent change to this exact example in #6773 and recent docs PRs. The earlier project-metadata review thread is resolved, and the reported checks pass/skip. The AX example now sets `openinference.project.name` on the tracer resource while keeping LiveKit session metadata separate.

PR #7066, opened 2026-08-31
fix(voice): drop the recognition turn a false interruption resumed over

### LINKED ISSUE #7063: False interruption keeps stale started_speaking_at for next user turn
### Bug Description

### Deterministic timeline

On 1.7.1:

```text
```

### Expected Behavior

### Actual Behavior

### Proposed Solution

### Package Versions

```text
livekit-agents==1.7.1
```

salvatorebottiglieri:
## Summary

Fixes #7063.

When a false interruption is confirmed and the paused agent speech resumes, `_on_false_interruption` never closes the `AudioRecognition` turn that raised the interruption. The turn stays open, and `_on_vad_event` only takes a fresh anchor while `_vad_speech_started` is still `False`:

```python
# audio_recognition.py, _on_vad_event, START_OF_SPEECH
if not self._vad_speech_started:
    self._speech_start_time = speech_start_time
    self._vad_speech_started = True
```

So the *next real utterance* inherits the abandoned turn's `_speech_start_time`, and `ChatMessage.metrics.started_speaking_at` can predate agent speech that actually came first. The guard from #6093 / #6098 doesn't catch it because the resulting metrics stay internally consistent — the anchor is stale, not impossible.

This clears the turn in the `resumed` branch of `_on_false_interruption`, in the order the issue asks for: clear → `_on_start_of_agent_speech` → `audio_output.resume()`.

## Why here and not in the drop path

`_bounce_eou_task` resets the anchors only when the turn *commits*. On a drop it resets just `_turn_backchannel_over_agent`, `_overlap_in_current_turn` and `_user_turn_committed` — which is right in general, because a dropped turn usually means the user isn't finished and the turn should keep accumulating. A confirmed false interruption is the one case where a drop means the opposite: the agent is resuming *over* that turn, so it has been decided to be noise.

## Notes for review

- **Only the `resumed` branch.** `_pause_enabled()` gates `resume_false_interruption` / `can_pause` / `audio_enabled` before `_paused_speech` is set, and the timer is only armed when `_paused_speech` is set — so with resume disabled this code is never reached.
- **Two entry paths, two levels of reset.** `_on_turn_settled` runs after a decision dropped the turn, so nothing is in flight and the full `_clear_user_turn()` applies — it is also needed there, since a confirmed backchannel's transcript would otherwise prepend itself to the next utterance. `_on_timeout` can run with no decision ever open (`turn_detection="stt"` starts no bounce on VAD `END_OF_SPEECH`), where a slow stt final for real speech may still be on its way; there only the inherited anchors are released, via `_release_user_turn_anchors()`. The first version of this PR cleared unconditionally and lost that final — see the discussion below.
- **No live turn can be wiped.** `on_start_of_speech` calls `_cancel_false_interruption_timer()`, closing both entry paths. In the sub-tick window where the timer callback wins, the clear is harmless: `_vad_speech_started` goes `False`, so the VAD `START_OF_SPEECH` processed right after takes its own anchor. This is also why the fix doesn't need the `_last_speaking_time == last_speaking_time` guard the commit path uses.
- **`_clear_user_turn()` swaps the STT pipeline, `_release_user_turn_anchors()` doesn't.** The full clear stays as-is because the issue requires the buffered transcript not to survive a confirmed drop, and it's the existing "discard this turn" primitive that the public `clear_user_turn()` already uses. If you'd rather express the narrow reset as a keyword on `_clear_user_turn` than as a second method, say so — it's a shared primitive, so the shape seemed like your call.

## Test

`tests/test_false_interruption_resume.py::test_resume_drops_the_turn_it_resumed_over`, built on the helpers already in that file.

It drives the harder entry path: with `FALSE_INTERRUPTION_TIMEOUT = 0.3` and `MAX_DELAY = 0.5` the timer fires while the end-of-turn decision is still open, so the resume goes through `_false_interruption_pending` → `_on_turn_settled` → `_on_false_interruption`. The assertion is on the observable anchor rather than an internal flag: after the resume a real VAD `START_OF_SPEECH` is delivered, and `_speech_start_time` must equal *that* onset. Without the fix it comes back ~12.7 s stale — the same order of magnitude as the 12.151 s reported in the issue.

`test_resume_without_a_turn_decision_keeps_a_late_transcript_alive` covers the other entry path: no bounce is started, so the timer fires with no decision open, and the test asserts the anchor is released while the transcript and the stt pipeline — everything a late final needs to commit — survive. Against the first commit it fails on `assert '' == 'what the caller actually said'`.

## Testing

- `uv run pytest tests/test_false_interruption_resume.py` → 11 passed (was 9); each new test fails without its fix and passes with it
- `uv run pytest --unit` → 2307 passed, 5 skipped
- `uv run ruff format --check` and `uv run ruff check` → clean
- `uv run python scripts/check_types.py` (mypy strict) → no issues in 644 source files

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -4516,7 +4516,7 @@ class AgentActivity(RecognitionHooks):
     def _start_false_interruption_timer(self, timeout: float) -> None:
         self._cancel_false_interruption_timer()
 
-        def _on_false_interruption() -> None:
+        def _on_false_interruption(*, turn_dropped: bool) -> None:
             if self._paused_speech is None or (
                 self._current_speech and self._current_speech is not self._paused_speech.handle
             ):
@@ -4532,10 +4532,19 @@ class AgentActivity(RecognitionHooks):
                 and not self._paused_speech.handle.done()
             ):
                 if (recognition := self._audio_recognition) is not None:
-                    # the interruption was false, so the turn that raised it is abandoned: drop
-                    # it before the agent's next speech interval opens, or the next real
-                    # utterance inherits its speech-start anchor and buffered transcript
-                    recognition._clear_user_turn()
+                    # the interruption was false, so the turn that raised it is abandoned: let
+                    # it go before the agent's next speech interval opens, or the next real
+                    # utterance inherits its speech-start anchor
+                    if turn_dropped:
+                        # a decision ran and dropped this turn, so nothing is still in flight
+                        # for it: discard the buffered transcript too, or a confirmed
+                        # backchannel prepends itself to the next utterance
+                        recognition._clear_user_turn()
+                    else:
+                        # the timeout fired with no decision open, and turn_detection="stt"
+                        # starts no bounce on VAD END_OF_SPEECH — a slow stt final for real
+                        # speech may still be on its way, so keep whatever can still commit
+                        recognition._release_user_turn_anchors()
 
                 self._session._update_agent_state(
                     self._paused_speech.agent_state,
@@ -4571,7 +4580,7 @@ class AgentActivity(RecognitionHooks):
             if settled.cancelled() or (recognition is not None and recognition._closing.is_set()):
                 return  # torn down instead of decided; closing releases the pause itself
 
-            _on_false_interruption()
+            _on_false_interruption(turn_dropped=True)
 
         def _on_timeout() -> None:
             self._false_interruption_timer = None
@@ -4586,7 +4595,7 @@ class AgentActivity(RecognitionHooks):
                 eot_task.add_done_callback(_on_turn_settled)
                 return
 
-            _on_false_interruption()
+            _on_false_interruption(turn_dropped=False)
 
         self._false_interruption_timer = self._session._loop.call_later(timeout, _on_timeout)
 
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -4531,12 +4531,18 @@ class AgentActivity(RecognitionHooks):
                 and audio_output.can_pause
                 and not self._paused_speech.handle.done()
             ):
+                if (recognition := self._audio_recognition) is not None:
+                    # the interruption was false, so the turn that raised it is abandoned: drop
+                    # it before the agent's next speech interval opens, or the next real
+                    # utterance inherits its speech-start anchor and buffered transcript
+                    recognition._clear_user_turn()
+
                 self._session._update_agent_state(
                     self._paused_speech.agent_state,
                     otel_context=self._paused_speech.handle._agent_turn_context,
                 )
-                if self._audio_recognition and self._paused_speech.agent_state == "speaking":
-                    self._audio_recognition._on_start_of_agent_speech(started_at=time.time())
+                if recognition is not None and self._paused_speech.agent_state == "speaking":
+                    recognition._on_start_of_agent_speech(started_at=time.time())
                 if self.interruption_enabled:
                     self._disable_vad_interruption_soon()
                 audio_output.resume()
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -4516,7 +4516,7 @@ class AgentActivity(RecognitionHooks):
     def _start_false_interruption_timer(self, timeout: float) -> None:
         self._cancel_false_interruption_timer()
 
-        def _on_false_interruption(*, turn_dropped: bool) -> None:
+        def _on_false_interruption() -> None:
             if self._paused_speech is None or (
                 self._current_speech and self._current_speech is not self._paused_speech.handle
             ):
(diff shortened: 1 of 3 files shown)


salvatorebottiglieri:
The review finding is correct, and it caught a real regression in the first version of this PR. Fixed in the push above. Details, since the reasoning is the interesting part of this change.

`_on_false_interruption` has two entry paths, and I had wrongly treated them as equivalent:

- **`_on_turn_settled`** :  a turn decision ran and did not commit. If it had committed, `on_end_of_turn` would have interrupted the paused speech → `_cancel_speech_pause` → `_cancel_false_interruption_timer` clears `_false_interruption_pending`, and the callback returns early. So reaching `_on_false_interruption` here means the turn was decided and dropped: nothing is in flight for it.
- **`_on_timeout` direct** : no decision was ever open. Reachable because the eou bounce only starts on VAD `END_OF_SPEECH` when `_vad_base_turn_detection`, which is `self._turn_detection_mode in ("vad", None)`. With `turn_detection="stt"` the bounce waits for the stt final instead, so `_end_of_turn_task` is `None` and the timer fires straight through. `_pause_enabled()` doesn't exclude that mode.

In the second path the speech may have been real, with a slow stt final still on its way, and `_clear_user_turn()` runs `_update_stt(None)` — closing the pipeline and cancelling the stt consumer. The final is lost. Before this PR that final would have arrived and committed the turn: out-of-order metrics (the bug this PR is about) but not lost audio. Trading a metrics defect for lost speech is strictly worse.

What makes this awkward is that at timeout time you can't distinguish noise from slow stt — both look like "VAD fired, silence, no transcript". That's why the framework resumes the speech in both cases. So the discriminator can't be "is this noise?", it has to be "was a decision made?".

So the reset is now two levels, with the entry path passed in as `_on_false_interruption(*, turn_dropped: bool)`:

- `turn_dropped=True` → `_clear_user_turn()` as before. The full clear is also *needed* here: a confirmed backchannel's transcript sits in `_audio_transcript` (the drop path in `_bounce_eou_task` only clears it on commit) and would otherwise prepend itself to the next real utterance.
- `turn_dropped=False` → a new `AudioRecognition._release_user_turn_anchors()`, which releases only what would be inherited — `_speech_start_time`, `_vad_speech_started`, `_last_emitted_prediction`, `_turn_tracker` and the open `user_turn` span — and leaves everything that could still commit: the transcript, the accumulated confidence, the turn detector's buffer, the stt pipeline.

Both paths still fix the reported bug, which matters because the issue's own reproduction drives `_start_false_interruption_timer(0)` with no decision open — i.e. the direct path. Narrowing the fix to the settled path alone would have left the reported case unfixed.

New test: `test_resume_without_a_turn_decision_keeps_a_late_transcript_alive`. Against the previous commit it fails on `assert '' == 'what the caller actually said'`.

salvatorebottiglieri:
Also correct, and it's the common case rather than an edge: the shipped `false_interruption_timeout` is 2s (`voice/turn.py`), while a confirmed backchannel drops in a few hundred ms, so the decision normally lands well before the resume timer. `_on_timeout` then falls through to the direct path with `eot_task.done()` — and the previous commit read the verdict off *which callback* invoked the resume, so it treated that as "no decision" and kept the transcript. `_audio_transcript` accumulates, so the dropped words prepend themselves to the next real utterance.

`eot_task.done()` can't stand in for the verdict either: `_end_of_turn_task` is never reset to `None` after completion, so a task completed for an *earlier* turn is indistinguishable from a decision about this one — using it would bring back the discarded-final case from the first finding.

Both of my attempts failed the same way: inferring turn-decision state from an indirect signal. So the fix stops inferring and records it where it is known.

`AudioRecognition._user_turn_dropped` is set by `_bounce_eou_task` as `not committed`, once per logical turn, in the section that already resets the turn-scoped barge-in state. A new VAD `START_OF_SPEECH` supersedes it (the branch that takes a fresh `_speech_start_time`), and both `_clear_user_turn` and `_release_user_turn_anchors` clear it. `_on_false_interruption` now reads the verdict instead of deducing it, which also lets the entry-path parameter from the previous commit go — the resume no longer cares how it was reached.

New test: `test_a_backchannel_dropped_before_the_timeout_does_not_leak_forward` arms a resume timer that outlives the decision and asserts the verdict survives the gap and the transcript ends up empty. Without the recorded verdict it fails on `assert False is True`.

12 tests in the file now (9 before this PR), each red without its own fix; `uv run pytest --unit` → 2308 passed, 5 skipped; ruff and mypy strict clean.

salvatorebottiglieri:
Right again, and this one is squarely on me: I had spotted this exact staleness risk while writing the previous commit and then closed it for the VAD path only, assuming a new turn always begins at a VAD `START_OF_SPEECH`. It doesn't — the resume timer is armed from the stt hooks too (`on_interim_transcript` / `on_final_transcript` when `speaking is False`), so a session can anchor its next turn through `_on_stt_event`, where the verdict from the turn that just ended survived and the resume erased a real transcript. The first finding, by another door.

Three paths take a fresh `_speech_start_time`: `_on_stt_event` START_OF_SPEECH, `_on_vad_event` START_OF_SPEECH, and `_on_vad_event` INFERENCE_DONE. I had covered one. Rather than adding the invalidation to the other two and relying on the next person remembering, they now all go through `_open_user_turn(speech_start_time)`, which takes the anchor and drops the previous verdict together — the invariant is structural instead of remembered, which is precisely why I missed it.

New test: `test_an_stt_anchored_turn_supersedes_the_previous_verdict` anchors the next turn through the stt stream with a dropped verdict standing, and asserts the verdict is superseded and a slow final for the new turn survives the resume. Without the reset on that path it fails on `assert True is False`.

13 tests in the file (9 before this PR), each red without its own fix; `uv run pytest --unit` → 2309 passed, 5 skipped; ruff and mypy strict clean.

Happy to squash the four commits if you'd rather review this as one change — the history is only useful as a record of the review round trips.

Rehansanjay:
Not a maintainer, and nothing here is a criticism of this PR — flagging a collision that is only visible from outside any one of them.

**Three open PRs are fixing this same bug independently**, and all three add a file at the same path, so at most one can merge without the other two rebasing:

| PR | author | size | approach |
| --- | --- | --- | --- |
| **#7066** (this one) | @salvatorebottiglieri | +216/-9 | Adds a `_user_turn_dropped` flag and splits the operation in two: `_clear_user_turn()` for a turn that was dropped, and a new `_release_user_turn_anchors()` for one abandoned before any end-of-turn decision ran — releasing the speech-start anchor while deliberately leaving the transcript, accumulated confidence and turn-detector buffer alone so an stt final still in flight is not lost. Also funnels every path that takes a fresh `_speech_start_time` through a new `_open_user_turn()`. |
| **#7067** | @saime428 | +105/-5 | Adds `reset_stt: bool = True` to `_clear_user_turn`, guarded on `not (recognition._speaking or recognition._vad_speech_started)`. |
| **#7101** | @dorukdumlu | +181/-1 | Same `reset_stt` parameter, with a narrower guard built on a new `_vad_speech_flushed` flag: `not (_speaking or (_vad_speech_started and not _vad_speech_flushed))`. |

All three touch `voice/agent_activity.py` and `voice/audio_recognition.py`, and all three create `tests/test_false_interruption_resume.py`.

Read together the space is roughly: #7067 and #7101 are the same design, with #7101 refining the guard so a flushed-but-still-set `_vad_speech_started` does not block the discard; #7066 is a larger change that treats "dropped" and "abandoned before a verdict" as genuinely different states rather than one call with a flag.

I have no view on which is right — that is a maintainer's call about how much structure this area should carry. Posting it so whoever triages does not have to discover the overlap three times, and so the two authors whose PRs are not taken can rebase early rather than after a conflict.

Same family as #7011 (manual-mode VAD cancelling a committed turn), which touches `audio_recognition.py` too but a different handler.

PR #7067, opened 2026-08-31
fix(voice): discard the stale user turn when resuming a false interruption

### LINKED ISSUE #7063: False interruption keeps stale started_speaking_at for next user turn

saime428:
Fixes #7063.

A user message can be timestamped seconds before the user actually spoke, which sorts it
ahead of the agent reply it was answering. From the production timeline in the report:

| time (UTC) | what happened |
| --- | --- |
| 19:20:06.481 | VAD opens a user turn, no STT ever commits |
| 19:20:10.114 | the interruption is confirmed false, paused agent speech resumes |
| 19:20:18.632 | the user actually starts the next utterance |
| | that message commits with `started_speaking_at = 19:20:06.481` |

12.151 seconds early, taken from a turn the system had already decided was not real.

The ordering guard from #6093 does not catch this, because the metrics stay internally
consistent: `stopped_speaking_at >= started_speaking_at` still holds. The anchor is stale but still a possible value, so the guard has nothing to reject.

The cause is that the resume abandons the recognition turn without clearing it, so its
speech anchors survive into the next real utterance. The discard goes in the resume branch,
before the state restore marks the agent speaking again.

Reproducing it without a live call, on a paused speech whose `agent_state == "speaking"`:
set `_speech_start_time` on the active recognition turn, let the false-interruption timer
fire, and the anchor is still set afterwards. That is
`test_resume_discards_the_stale_recognition_turn` below, which fails on a clean checkout at
`assert recognition._speech_start_time is None`.

## Updated after the review on #7063

@dorukdumlu [compared the three open attempts at this
bug](https://github.com/livekit/agents/issues/7063#issuecomment-5517383726) and raised two
problems with this PR's original version. Both were fair, and this PR now addresses them.

The STT stream is no longer torn down. The original called `_clear_user_turn()`, which
ends by recreating the provider stream (`_update_stt(None)` and back). That is costly on
every false interruption, and it drops audio the provider is still decoding. When the
barge-in was real and VAD missed it, that late final is the only thing left that can
interrupt the resumed speech. `_clear_user_turn` now takes `reset_stt`, and the resume
passes `reset_stt=False`.

The discard is skipped while speech is still tracked. The original cleared
unconditionally, which could take a live utterance's anchors.

On the second point I could confirm the problem but not the path given for it. A VAD segment
below `min_duration` does not leave the timer armed with live anchors: `_on_vad_event` sets
`_speech_start_time` and `_vad_speech_started` and then calls `on_start_of_speech` in the
same synchronous block, and `on_start_of_speech` cancels the timer.

The path that does reach it is STT turn detection. An STT `END_OF_SPEECH` calls
`on_end_of_speech`, which arms this timer, while VAD is still inside its speech segment.
`audio_recognition` already knows about that state and warns on it:

```
stt end of speech received while vad is still in a speech segment, flushing vad
```

Those anchors belong to the utterance in progress, so the resume now leaves them alone. The
guard is `_speaking or _vad_speech_started`, and in the ordinary false-interruption case both
are already `False`: the VAD `END_OF_SPEECH` that armed the timer released them, leaving only
the anchor behind.

## Relationship to #7066 and #7101

All three PRs fix the reported bug. #7066 and #7101 are more thorough; this one is the
smallest change that covers the same ground, at 32 lines of source. It has no
decided/undecided split and adds no new state, so if either of the others is the preferred
base, take it and close this.

## Tests

`test_resume_discards_the_stale_recognition_turn` is the regression test for the report. Its
original version set `_vad_speech_started = True`, which is not the state the timer actually
fires in, so it now models the real one: the VAD `END_OF_SPEECH` has already released
`_speaking` and `_vad_speech_started`, and only the anchor is left.

Two tests are added for the changes above: `test_resume_keeps_the_anchors_of_a_live_utterance`
covers the STT case where the discard must be skipped, and
`test_resume_keeps_the_stt_stream_alive` fails if the discard goes back to recreating the
provider stream.

## Validation

Rebased onto `4d03f50`.

- `uv run pytest tests/test_false_interruption_resume.py --unit`: 12 passed
- The regression test fails on a clean `4d03f50` checkout without this change, at
  `assert recognition._speech_start_time is None`
- `uv run pytest tests/ --unit`, same machine, run back to back:

  | | failed | passed | errors |
  |---|---|---|---|
  | clean `4d03f50` | 2 | 2083 | 9 |
  | this branch | 2 | 2086 | 9 |

  The three extra passes are the tests above. `tests/` has pre-existing flaky failures and
  teardown errors on my machine, plus a pytest `INTERNALERROR: Event loop is closed` raised
  from the concurrency plugin's teardown, so I ran a clean checkout side by side rather than
  reading those as regressions. That `INTERNALERROR` aborts the run before pytest prints its
  summary, which is why I can give the counts but not the names of the two failures.


Rehansanjay:
Heads-up rather than a review: **#7066, #7067 and this PR are all fixing the same bug independently**, and all three create `tests/test_false_interruption_resume.py`, so at most one merges without the others rebasing.

I put a side-by-side of the three approaches on the earliest of them so a maintainer does not have to work it out three times: https://github.com/livekit/agents/pull/7066#issuecomment-5581774489

No view here on which design is right — just so nobody discovers the collision after a conflict.

PR #7068, opened 2026-08-31
feat(metrics): expose reasoning_tokens in LLM metrics and traces

s-hamdananwar:
Carries `reasoning_tokens` from OpenAI-compatible usage payloads through to `LLMMetrics`, `LLMModelUsage`, the `log_metrics` line, and a new `gen_ai.usage.reasoning_tokens` span attribute. Reasoning is a **subset** of `completion_tokens`, never additive.

The inference gateway already returns `completion_tokens_details.reasoning_tokens`, but `inference/llm.py` never read it, so the value was dropped before reaching metrics or traces.

### DIFF
--- a/livekit-agents/livekit/agents/inference/llm.py
+++ b/livekit-agents/livekit/agents/inference/llm.py
@@ -474,12 +474,17 @@ class LLMStream(llm.LLMStream):
                     if chunk.usage is not None:
                         tokens_details = chunk.usage.prompt_tokens_details
                         cached_tokens = tokens_details.cached_tokens if tokens_details else 0
+                        completion_details = chunk.usage.completion_tokens_details
+                        reasoning_tokens = (
+                            completion_details.reasoning_tokens if completion_details else 0
+                        )
                         usage_chunk = llm.ChatChunk(
                             id=chunk.id,
                             usage=llm.CompletionUsage(
                                 completion_tokens=chunk.usage.completion_tokens or 0,
                                 prompt_tokens=chunk.usage.prompt_tokens or 0,
                                 prompt_cached_tokens=cached_tokens or 0,
+                                reasoning_tokens=reasoning_tokens or 0,
                                 total_tokens=chunk.usage.total_tokens or 0,
                                 service_tier=getattr(chunk, "service_tier", None),
                             ),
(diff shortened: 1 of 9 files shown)


shawnfeldman (approved):
Clean additive change — every new field defaults to 0, so nothing that doesn't report reasoning shifts.

Two things I checked rather than assumed:

**The test really does exercise the changed code.** `tests/test_inference_llm_usage.py` imports `livekit.plugins.openai`, while the wiring lands in `livekit-agents/.../inference/llm.py` — which reads like the test covering a different path. It isn't: the plugin's `LLMStream` (llm.py:1041) subclasses `inference.llm.LLMStream` and overrides only `__init__`, so it inherits the changed `_run`. Mutation-confirmed — dropping `reasoning_tokens=reasoning_tokens or 0,` from the `CompletionUsage` construction fails `test_reasoning_tokens_are_reported_from_completion_tokens_details` with `assert 0 == 32`, baseline 3 passed. `reasoning_tokens` appears in exactly the nine files this PR touches, so that ctor is the only producer.

**Field insertion order is safe.** `reasoning_tokens: int = 0` goes in ahead of `total_tokens: int` (no default) in both `LLMMetrics` and `CompletionUsage`. That would be a `TypeError` at import in a dataclass; both are Pydantic `BaseModel`s, and `cache_creation_tokens: int = 0` already sits in the same position, so the shape is pre-existing.

The subset-not-additive invariant is stated in all three docstrings and pinned by `test_collector_aggregates_reasoning_tokens` asserting `output_tokens == 150` alongside `output_reasoning_tokens == 72`. Adding the new span attribute to the `test_trace_types_pii.py` allowlist is right — a token count carries no PII. Gating the span `set_attribute` on truthiness keeps non-reasoning spans unchanged.

25/25 checks green, including `unit-tests`.

Two notes, neither blocking and both arguably out of scope: `RealtimeModelMetrics` in `ModelUsageCollector.collect` has no equivalent branch, so realtime usage will report `output_reasoning_tokens == 0` even for a reasoning model; and plugins that build `CompletionUsage` outside the `inference.llm.LLMStream` base class won't populate it.

theomonnom (approved):

PR #7074, opened 2026-09-01
fix(aws-stt): support aws-sdk-transcribe-streaming >=0.11.0

chakshu-dhannawat:
`aws-sdk-transcribe-streaming` 0.11.0 renamed the client/config exports to `AsyncTranscribeStreamingClient`/`AsyncTranscribeStreamingConfig`, stopped installing `awscrt`, and requires `await Config.resolve(...)` instead of direct construction.\n\nThis change makes the AWS STT plugin compatible with both the pre-0.11 and 0.11+ APIs:\n- Try the new async names first, falling back to the legacy `TranscribeStreamingClient`/`Config` (and the 0.10 `config` alias).\n- Fall back from `AWSCRTHTTPClient` to `AIOHTTPClient` when `awscrt` is not installed.\n- Use `Config.resolve(...)` when available, otherwise construct directly.\n- Pass `transport=` to the config so the same HTTP client is reused and closed cleanly in `SpeechStream.aclose()`.\n\nVerified `tests/test_plugin_aws_stt.py` passes with `aws-sdk-transcribe-streaming==0.7.0` and `0.11.0`.\n\nFixes livekit/agents#7072

### DIFF
--- a/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
+++ b/livekit-plugins/livekit-plugins-aws/livekit/plugins/aws/stt.py
@@ -36,13 +36,22 @@ from .log import logger
 from .utils import DEFAULT_REGION
 
 try:
-    from aws_sdk_transcribe_streaming.client import TranscribeStreamingClient
-
     try:
-        from aws_sdk_transcribe_streaming.config import Config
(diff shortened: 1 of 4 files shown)


longcw:
closing in favor of https://github.com/livekit/agents/pull/7094

PR #7080, opened 2026-09-01
fix(google): support Gemma 4 thinking levels

### LINKED ISSUE #7071: google plugin: thinking_config gate misclassifies Gemma 4, making thinking uncontrollable
## Description

`livekit-plugins-google` routes `thinking_config` by a model-name heuristic that predates Gemma 4, so it is impossible to control thinking for Gemma 4 models through the plugin:

- `_is_gemini_3_model()` in `livekit-plugins-google/livekit/plugins/google/llm.py` (line ~44) returns True only for names containing `"gemini-3"`.
- In `LLM.chat()` (line ~400), every other model is treated as "Gemini 2.5 and earlier": passing a `thinking_config` with `thinking_level` raises `ValueError: Model gemma-4-31b-it does not support thinking_level. Please use thinking_budget (int) instead for Gemini 2.5 and earlier models.`

But Gemma 4 is a `thinking_level` model, not a `thinking_budget` model:

- The Gemini API rejects `thinking_budget` for Gemma 4 with HTTP 400 ("thinking_budget flag isn't supported on Gemma 4 models").
- `thinkingConfig.includeThoughts: false` is silently ignored for Gemma 4 (google-gemini/cookbook#1198).
- `thinkingConfig.thinkingLevel: "MINIMAL"` works and produces zero thought tokens.

So for Gemma 4 the plugin refuses the only parameter the API honors, and the parameter it demands instead is rejected by the API. There is no valid way to disable thinking, which matters for voice agents where thought tokens land ahead of the first spoken token.

## Reproduction

```python
from google.genai import types

llm = google.LLM(
    model="gemma-4-31b-it",
    thinking_config=types.ThinkingConfig(thinking_level=types.ThinkingLevel.MINIMAL),
)
llm.chat(chat_ctx=...)  # raises ValueError from the thinking gate
```

Versions: livekit-agents 1.7.0, livekit-plugins-google 1.7.0 (gate is unchanged in 1.7.1).

## Suggested fix

Teach the gate that Gemma 4 models (`gemma-4-*`) take `thinking_level`, e.g. include them in the level-capable branch alongside `gemini-3*`, or probe by capability rather than name. The current workaround is to construct the LLM without a `thinking_config` and inject `{"thinking_config": {"thinking_level": "minimal"}}` through `chat(extra_kwargs=...)`, which bypasses the gate.

dawNotPoi:
## Summary

- route Gemma 4 through the thinking-level configuration path without enabling Gemini-3-only behavior
- preserve include_thoughts and reject unsupported budget-only configurations with an actionable error
- validate the Gemma 4 minimal/high levels and cover the final generated request configuration

## Verification

- uv run pytest -q tests/test_google_thought_signatures.py tests/test_plugin_google_llm.py (76 passed)
- uv run ruff format --check on changed files
- uv run ruff check on changed files
- uv run --group typing mypy --platform linux -p livekit.plugins.google

Fixes #7071

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
@@ -46,6 +46,16 @@ def _is_gemini_3_model(model: str) -> bool:
     return "gemini-3" in model.lower() or model.lower().startswith("gemini-3")
 
 
+def _is_gemma_4_model(model: str) -> bool:
+    """Check if model is Gemma 4 series."""
+    return model.lower().startswith("gemma-4-")
+
+
+def _supports_thinking_level(model: str) -> bool:
+    """Check if model configures thinking with a level rather than a token budget."""
+    return _is_gemini_3_model(model) or _is_gemma_4_model(model)
+
+
 def _function_calling_config(
     tool_choice: NotGivenOr[ToolChoice], function_tools: Iterable[str]
 ) -> types.FunctionCallingConfig | None:
@@ -364,7 +374,8 @@ class LLM(llm.LLM):
 
         # Handle thinking_config based on model version
         if is_given(self._opts.thinking_config):
-            is_gemini_3 = _is_gemini_3_model(self._opts.model)
+            supports_thinking_level = _supports_thinking_level(self._opts.model)
+            is_gemma_4 = _is_gemma_4_model(self._opts.model)
             is_gemini_3_flash = _is_gemini_3_flash_model(self._opts.model)
             thinking_cfg = self._opts.thinking_config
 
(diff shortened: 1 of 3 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/llm.py
```diff
@@ -378,21 +389,54 @@ def chat(
...
+                        raise ValueError(
+                            f"Model {self._opts.model} does not support thinking_budget. "
+                            "Please use thinking_level ('minimal' or 'high') instead."
+                        )
+
+                    if _level is not None:
+                        level_value = (
+                            _level.value if isinstance(_level, types.ThinkingLevel) else _level
+                        )
+                        if level_value.lower() not in {"minimal", "high"}:
```
longcw: could you add a ref about what level it supports? and we should use `types.ThinkingLevel` instead of text comparision.

longcw:
closing in favor of #7095 with some clean up and validation, you're credited there as co-author

PR #7089, opened 2026-09-01
feat(phonic): add update_options for mid-session config changes

arunwpm-work:
## Motivation

The Phonic realtime plugin sends its config **once** and has no way to change config fields mid-session. A multi-step agent (e.g. an interview that advances topics via handoff) can't, for example, switch `default_language` to Spanish for the next section: the Agent handoff (`_update_session`) only threads **instructions / chat_ctx / tools** into the `reset`, everything else is fixed at `RealtimeModel` construction, and `update_options` was a no-op.

Phonic's `reset` message already accepts a full config, so the capability exists on the wire — this just exposes it via the standard `update_options` method. (Companion to the same change in `agents-js`: livekit/agents-js#2407.)

## Change

Implement `update_options` (previously a no-op) on the Phonic `RealtimeSession`, and add a forwarding `update_options` on the `RealtimeModel` (apps hold the model). It changes any config-level field mid-session and applies it **immediately** via a Phonic `reset`; unset fields keep their current values. Instructions/tools stay driven by the Agent handoff.

- `update_options` exposes each config-level field as its own keyword-only `NotGivenOr` param (`default_language=…`, `voice=…`, `boosted_keywords=…`, …) — matching the `openai`/`google` realtime plugins — so callers pass only what changes. `tool_choice` (the base param the framework sends each turn) is ignored: Phonic has no `tool_choice` (same as `aws`/`nvidia`).
- The handoff reset and `update_options` share a single `_send_mid_session_reset()` helper. `update_options` is synchronous (like other plugins), so the reset is fired as a background task and coalesced.
- **Language switching:** the API rejects a `default_language` that also appears in `additional_languages`. When the default changes (and `additional_languages` isn't passed), the previous default is rotated into `additional_languages` and the new one dropped — e.g. `{default: en, additional: [es]}` + `update_options(default_language="es")` → `{default: es, additional: [en]}`. Keeps the language set intact.

## Usage

```python
model = phonic.realtime.RealtimeModel(
    default_language="en",
    additional_languages=["es"],
    multilingual_mode="request",
)
session = AgentSession(llm=model)

# e.g. in the next task's on_enter, or the advancing tool handler:
model.update_options(default_language="es")
# -> immediate reset; the next reply is in Spanish
```

## Validation

- `uv run ruff format --check` / `ruff check` pass; `mypy` is clean on the change (the one standalone `phonic.AsyncPhonic` error pre-exists on `main` and resolves in CI's full env).
- Verified end-to-end against the Phonic test env (via the JS companion): a task that calls `update_options(default_language="es")` switches a live session to Spanish — the server accepts the `es` reset and the next reply is in Spanish.


livekit-plugins/livekit-plugins-phonic/livekit/plugins/phonic/realtime/realtime_model.py
```diff
@@ -729,8 +793,73 @@ def _build_config_options(
         # Filter out NOT_GIVEN values
         return {k: v for k, v in options.items() if v is not NOT_GIVEN}
 
-    def update_options(self, *, tool_choice: NotGivenOr[llm.ToolChoice | None] = NOT_GIVEN) -> None:
-        logger.warning("update_options is not supported by the Phonic realtime model.")
+    def update_options(
+        self,
+        *,
+        tool_choice: NotGivenOr[llm.ToolChoice | None] = NOT_GIVEN,
+        config: NotGivenOr[PhonicConfig] = NOT_GIVEN,
```
tinalenguyen: could we unpack the params in the config to match the patterns in other plugins? might also be helpful when it's typed as NotGivenOr[] for each

tinalenguyen (approved):

PR #7097, opened 2026-09-02
feat(meta): add Muse Voice streaming STT plugin

seyeong-han:
## Summary

Adds `livekit-plugins-meta` with streaming speech-to-text support for Meta Muse Voice Transcribe 1.0.

- exposes `livekit.plugins.meta.STT`
- sends 24 kHz mono PCM16 audio in paced 80 ms frames
- supports static keywords and language bias at stream creation
- maps cumulative Muse events into non-overlapping LiveKit speech turns
- emits incremental recognition usage from successfully sent audio
- avoids retrying after audio has been consumed, preventing silent speech loss
- redacts credentials, provider payloads, transcripts, and WebSocket close reasons from errors
- adds the workspace source and `livekit-agents[meta]` extra

The adapter uses LiveKit's shared `aiohttp` context and the framework-managed input channel/resampler. Batch recognition, diarization, detected-language metadata, and active-stream keyterm updates are intentionally not advertised.

## Testing

- `uv run pytest tests/test_meta_stt.py --unit` (`33 passed`)
- `make check` (format, Ruff, and mypy across 648 source files)
- wheel and sdist build, followed by an isolated Python 3.10 wheel import
- real Muse realtime smoke: handshake, paced 24 kHz audio, usage events, `endStream`, and clean close
- full hermetic run: `2462 passed, 5 skipped`; the remaining 9 `tests/test_room.py` cases independently reproduce a local LiveKit server `401 invalid API key: devkey` and are unrelated to this plugin


livekit-plugins/livekit-plugins-meta/livekit/plugins/meta/stt.py
```diff
@@ -0,0 +1,760 @@
...
+        Args:
+            api_key: Meta Model API key. Falls back to ``MODEL_API_KEY``.
+            model: Muse Voice Transcribe model identifier.
+            url: Realtime Muse ASR WebSocket endpoint. Must use ``wss://``.
+            keywords: Static recognition keywords sent when each stream starts.
+            language_bias: Static supported language names sent when each stream starts.
+            http_session: Optional aiohttp session. By default, the LiveKit HTTP
+                context session is used.
+        """
+        resolved_key = (api_key if api_key is not None else os.getenv("MODEL_API_KEY", "")).strip()
```
tinalenguyen: should we rename to `META_API_KEY`?

seyeong-han: Can we use both or them, `META_API_KEY` and `MODEL_API_KEY`?
Our official key name is `MODEL_API_KEY`, so it would be good to follow the the [official docs](https://dev.meta.ai/docs/overview/).

seyeong-han: Implemented both environment names. Resolution order is explicit `api_key`, then the official `MODEL_API_KEY`, then `META_API_KEY`. Added focused precedence/fallback tests and validated a real realtime connection using only `META_API_KEY`.

livekit-plugins/livekit-plugins-meta/livekit/plugins/meta/stt.py
```diff
@@ -0,0 +1,768 @@
...
+        return APIConnectionError(
+            f"Meta Muse realtime ASR protocol error: {detail}",
+            retryable=False,
+        )
+
+    @staticmethod
+    def _close_error(close_code: int | None, *, phase: str) -> APIStatusError:
+        code = close_code or -1
+        if code in _NON_RETRYABLE_CLOSE_CODES:
+            retryable = False
```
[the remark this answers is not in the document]
seyeong-han: Fixed. Unexpected code 1000 closures now remain retryable during handshake and before any audio is consumed, while code 1000 after audio remains terminal to preserve replay safety and code 1000 after `endStream` remains successful. Added coverage for all three states; 51 focused tests and full repository checks pass.

livekit-plugins/livekit-plugins-meta/livekit/plugins/meta/stt.py
```diff
@@ -0,0 +1,785 @@
...
+        self._api_key = self._normalize_access_token(resolved_key)
+        self._model = model.strip()
+        self._url = url
+        self._keywords = self._normalize_hints(keywords, name="keywords")
+        self._language_bias = self._normalize_language_bias(language_bias)
+        self._http_session = http_session
+        self._streams: weakref.WeakSet[SpeechStream] = weakref.WeakSet()
+        self._closed = False
+
+    @staticmethod
```
tinalenguyen: i don't think we need static methods in the STT and SpeechStream classes, could we have them as helper methods instead to match the style of our other plugins?

livekit-plugins/livekit-plugins-meta/livekit/plugins/meta/stt.py
```diff
@@ -0,0 +1,785 @@
...
+            or not isinstance(processed_ms, (int, float))
+            or not math.isfinite(processed_ms)
+            or processed_ms < 0
+        ):
+            raise self._protocol_error("audioProgress event has invalid audioProcessedMs")
+        if processed_ms <= self._last_audio_processed_ms:
+            return
+        delta_seconds = (processed_ms - self._last_audio_processed_ms) / 1000
+        self._last_audio_processed_ms = float(processed_ms)
+        self._emit_usage(delta_seconds)
```
tinalenguyen: this seems to emit metric events every 80ms, could we emit metrics once per turn to reduce noise?

livekit-plugins/livekit-plugins-meta/livekit/plugins/meta/stt.py
seyeong-han: Sure, no problem.

livekit-plugins/livekit-plugins-meta/livekit/plugins/meta/stt.py
seyeong-han: That's valid. I'll do it like that.

livekit-plugins/livekit-plugins-meta/livekit/plugins/meta/stt.py
seyeong-han: Refactored the state-independent STT and SpeechStream utilities into module-level helpers to match the other plugins. The Meta STT classes no longer contain staticmethod/classmethod helpers.

livekit-plugins/livekit-plugins-meta/livekit/plugins/meta/stt.py
seyeong-han: Implemented. Provider `audioProgress` deltas are now accumulated and emitted once per speech turn at the `speechEnd` boundary; overlapping/out-of-order finalization is covered. Any session-level residual usage (including silence or an error/cancellation path) is flushed once when the stream attempt exits. Focused tests are 53 passing, full checks pass, and a live silence smoke emitted one 1.040s usage event.

tinalenguyen (approved):

PR #7101, opened 2026-09-02
fix(voice): discard the uncommitted recognition turn on false-interruption resume

### LINKED ISSUE #7063: False interruption keeps stale started_speaking_at for next user turn

dorukdumlu:
## Summary

A confirmed false interruption resumed agent state and audio without discarding the recognition turn the false barge-in opened. `_ensure_user_turn_span` returns a recording span as-is and `_user_turn_start` is only written when it is None, so the next real utterance silently reused both: its committed ChatMessage carried a `started_speaking_at` seconds before the utterance was spoken (12.1s in the reported production timeline), ordering the user line ahead of agent speech that actually preceded it. The #6093 guard cannot catch this because the stale anchors stay internally consistent; they are just anchored to a dead turn.

## What changed

- `_on_false_interruption` calls `_clear_user_turn(reset_stt=False)` before restoring agent state, `_on_start_of_agent_speech`, and `audio_output.resume()`: the span is ended, and the timing anchors, buffered transcript, and turn bookkeeping are dropped.
- The clear is skipped when the user is speaking at the moment the timer fires. Those anchors belong to a live utterance, and the timer can genuinely fire mid-speech: a below-min_duration VAD segment never reaches the timer reset in `_interrupt_by_audio_activity`.
- `reset_stt=False` keeps the live STT stream. Recreating the provider stream on every false interruption is costly, and it drops audio the provider is still decoding: if the barge-in was real after all (a VAD miss), that late final is the only thing left that can interrupt the resumed speech. The manual `clear_user_turn()` API keeps its full-teardown behavior (default `reset_stt=True`).

Behavior note: a final transcript arriving after the resume now commits as a fresh turn without `started_speaking_at`, since the anchors are gone. Unknown is more honest than an anchor the system already declared false, matching what #6093 does for inconsistent anchors.

## Test plan

- [x] `pytest tests/test_false_interruption_resume.py` (11 passed)
  - the resume discards the dead turn before audio resumes (order pinned: clear, then resume), runs the real `_clear_user_turn`, and asserts the span is ended, the anchor is gone, and the STT pipeline is untouched
  - live user speech at timer fire keeps its anchors
  - existing resume-ordering cases unchanged
- [x] Verified the discard test fails without the fix.

Fixes #7063

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -4534,10 +4534,12 @@ class AgentActivity(RecognitionHooks):
             # survive, and the next real utterance reuses them
             # (_ensure_user_turn_span returns a recording span as-is), committing
             # with a started_speaking_at that predates the resumed agent speech.
-            # Skip while the user is speaking right now — those anchors belong
-            # to a live utterance, not to the discarded turn.
+            # Skip while the user is speaking right now: those anchors belong
+            # to a live utterance, not to the discarded turn. reset_stt=False
+            # keeps the live STT stream so audio still being decoded can deliver
+            # a late final and interrupt the resumed speech.
             if self._audio_recognition is not None and not self._audio_recognition._speaking:
-                self._audio_recognition._clear_user_turn()
+                self._audio_recognition._clear_user_turn(reset_stt=False)
 
             resumed = False
             if (
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -4538,16 +4538,17 @@ class AgentActivity(RecognitionHooks):
             # to a live utterance, not to the discarded turn. A live VAD segment
             # counts as speaking even when _speaking is already False, because a
             # premature STT END_OF_SPEECH clears _speaking while VAD is still
-            # mid-segment (the flushed VAD is expected to correct it with a new
-            # SOS). reset_stt=False keeps the live STT stream so audio still
-            # being decoded can deliver a late final and interrupt the resumed
-            # speech.
-            if (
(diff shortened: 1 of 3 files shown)


dorukdumlu:
Addressed Devin's yellow: it was real. A premature STT END_OF_SPEECH sets `_speaking` False while VAD is still mid-segment (the EOS handler even flushes VAD expecting a corrective SOS), so a timer settling in that window passed the guard and the discard wiped live speech state. The guard now also requires `_vad_speech_started` to be False. Regression test: `test_resume_keeps_the_anchors_of_a_live_vad_segment`.

dorukdumlu:
The second yellow is the mirror of the first and also real: silero's flush only resets state and emits no END_OF_SPEECH, so after a correct STT ending `_vad_speech_started` stayed True with nothing left to clear it, and the previous guard suppressed the discard forever on that path. Fixed in bf488aa: the flush marks the segment abandoned (`_vad_speech_flushed`), a fresh VAD SOS marks it live again, and the guard counts a VAD segment as speech only while it is unflushed. Regression test: `test_resume_discards_a_flush_abandoned_vad_segment`.

Rehansanjay:
Heads-up rather than a review: **#7066, #7067 and this PR are all fixing the same bug independently**, and all three create `tests/test_false_interruption_resume.py`, so at most one merges without the others rebasing.

I put a side-by-side of the three approaches on the earliest of them so a maintainer does not have to work it out three times: https://github.com/livekit/agents/pull/7066#issuecomment-5581774489

No view here on which design is right — just so nobody discovers the collision after a conflict.

PR #7103, opened 2026-09-03
fix(hw): discard impossible cgroup v2 CPU samples

### LINKED ISSUE #7102: CGroupV2CPUMonitor trusts a non-monotonic root-cgroup cpu.stat → worker flips to FULL on an idle host and the server refuses room jobs (lost calls)
### Bug Description

On a Xen HVM guest running the worker as a plain systemd service (no container, no CPU quota, cgroup v2), `_DefaultLoadCalc` periodically reports a **bogus load** — typically exactly `~0.80`, sometimes `1.0`, and right afterwards **large negative values** (`-8`, `-93`, we have seen `-9891`) — while the machine is idle (`load average 0.02`). Each spike flips the worker to `WS_FULL` for ~2.5 s:

```
"worker is at full capacity, marking as unavailable", "load": 0.8010850500002334, "threshold": 0.7
"worker is below capacity, marking as available",     "load": -93.70400344999976, "threshold": 0.7   (2 s later)
```

Frequency on our fleet: ~1–2.5 events/hour per worker (7 workers on one host, ~35 events in 8 h after a reboot, 130–180 per worker over 3 days before it). Workers with **zero** traffic are affected exactly like busy ones, and events are not correlated across workers.

**Impact — a real lost call.** The SIP participant joined while the worker was in its 2.5 s FULL window; the server refused the room job and the caller listened to silence until hanging up:

```
22:41:51.267  worker "flyc": worker is at full capacity, marking as unavailable  (load 0.8014)
22:41:51.317  livekit-server: starting RTC session  participant=sip_+40…  room=call_+40…
22:41:51.519  livekit-server agent/client.go:181  failed to send job request  {"error": "no servers available (received 1 responses)", "jobType": "JT_ROOM", "agentName": "flyc"}
22:41:53.769  worker "flyc": worker is below capacity, marking as available  (load 0.0165)
22:42:30      caller hangs up (39 s of silence, no agent ever joined the room)
```

1 lost call out of 623 in 3 days (~0.16 %). Related: #2130 ("Server reports no available agents despite CPU utilization being under 20%").

### Root cause: the root-cgroup `cpu.stat` counter is not monotonic on this guest, and the monitor trusts it blindly

`CGroupV2CPUMonitor.cpu_percent()` reads `usage_usec` from **`/sys/fs/cgroup/cpu.stat` (the root cgroup, i.e. the whole machine)** twice, 0.5 s apart, and divides the delta by `interval × cpu_count`. We instrumented `_read_cpu_usage` inside an unmodified `_DefaultLoadCalc` (patched at class level, nothing else changed) running as a transient systemd service on the affected host. Every anomalous sample is a pair of reads where one of the two values is wildly off — a *partial* or *inflated* sum, while the counter in the file is otherwise ~5.27e9 µs and growing at ~0.23 CPU:

```
22:18:35 [worker_cpu_load_monitor] pct=-677.98  start=5260817467 end=2548880500 diff=-2711936967 dt=0.500 ncpu=8
22:19:07 [worker_cpu_load_monitor] pct=1.00     start=5268770613 end=5860175581 diff=+591404968  dt=0.500 ncpu=8
22:19:26 [worker_cpu_load_monitor] pct=-154.75  start=5889766280 end=5270782758 diff=-618983522  dt=0.500 ncpu=8
22:19:29 [worker_cpu_load_monitor] pct=1.00     start=1210759425 end=5271033390 diff=+4060273965 dt=0.500 ncpu=8
22:19:34 [worker_cpu_load_monitor] pct=1.00     start=5271381786 end=5798443673 diff=+527061887  dt=0.500 ncpu=8
22:20:12 [worker_cpu_load_monitor] pct=-1018.20 start=5288672329 end=1215868554 diff=-4072803775 dt=0.500 ncpu=8
```

The bad values (1.21e9, 2.55e9 ≈ 2 and 4 CPUs' worth; 5.80–5.89e9 ≈ 9 CPUs' worth, on an 8-vCPU guest) look like a torn per-CPU summation of the root cgroup's cputime. The service's **own** cgroup `cpu.stat` (`/sys/fs/cgroup/system.slice/<unit>/cpu.stat`, computed through rstat) never showed a single anomaly in the same windows (0/1200 samples). Whether the root-counter glitch is a kernel or hypervisor issue (kernel 6.12, `CONFIG_VIRT_CPU_ACCOUNTING_GEN=y`, `CONFIG_PARAVIRT_TIME_ACCOUNTING=y`, Xen HVM 4.17, clocksource `xen`) is beside the point for the framework: the monitor turns one impossible sample into refused jobs because it

1. reads the **whole-machine** counter although the decision is per worker, instead of the process's own cgroup (`/proc/self/cgroup`);
2. divides by the **nominal** `interval` instead of the measured elapsed time;
3. clamps only the upper bound (`min(pct, 1)`) — `CGroupV1CPUMonitor` clamps to `[0, 1]`, v2 does not — so `-1018` goes straight into the moving average;
4. feeds it into a 5-sample `MovingAverage` with no plausibility check: four samples clamped at `1.0` give the recurring `0.80` signature, one huge negative sample gives the negative "load".

Numbers from the same host: two `_DefaultLoadCalc` instances run as services for 45 min each saw 44 and 50 raw anomalies out of 5396 samples (0.9 %), and their moving average crossed the threshold (≥ 0.7 or < 0) 138 and 54 times. The same code run from an interactive session (`user.slice`) saw 0 anomalies in 5000+ samples on two hosts — the read is context-dependent, which is one more reason not to build admission control on it.

### Reproduction

Unmodified library, Rocky Linux 10.2 / kernel 6.12.0-211.49.1 / 8 vCPU Xen HVM guest, cgroup2, run as a transient service:

```bash
systemd-run --unit=repro-load --property=CPUAccounting=yes --collect \
  /path/to/venv/bin/python repro_loadcalc.py /root/repro.log 2700
```

`repro_loadcalc.py` (below) just instantiates `livekit.agents.worker._DefaultLoadCalc`, polls `_get_avg()` every 0.5 s and logs anomalies. We are happy to run any branch or extra diagnostics on the host that reproduces this.

### Environment

- livekit-agents **1.6.10 and 1.7.1** (identical `utils/hw/cpu.py`; reproduced on both), Python 3.12.14, psutil 7.2.2
- Rocky Linux 10.2, kernel 6.12.0-211.49.1.el10_2, 8 vCPU Xen HVM 4.17 guest (XCP-ng host at < 3 % CPU, steal 0.08 %), `cgroup2fs`, workers as systemd services (no Docker, no `cpu.max` anywhere)
- `get_cpu_monitor()` → `CGroupV2CPUMonitor`, `cpu_count()` → 8; default `load_threshold` (0.7) and `load_fnc`

### Expected behavior / suggested fix

A worker on an idle host should never mark itself FULL. Concretely:

1. read the **process's own cgroup** (`/proc/self/cgroup` → `/sys/fs/cgroup/<path>/cpu.stat`) and fall back to the root cgroup only when that is unavailable;
2. divide by the **measured** elapsed time between the two reads;
3. clamp v2 results to `[0, 1]` like v1, and **discard** samples whose raw delta is negative or larger than `elapsed × ncpu` (physically impossible) instead of averaging them;
4. optionally log the raw samples when the status flips, so operators can tell real saturation from a measurement artefact.

Any of 1 or 3 alone would have prevented every lost-call window we observed.

<details><summary><code>repro_loadcalc.py</code></summary>

```python
"""Reproducător: EXACT clasa din bibliotecă (_DefaultLoadCalc + CGroupV2CPUMonitor), într-un proces separat, cu log pe fiecare eșantion brut."""
import sys, time, threading
sys.argv[1:] or sys.exit("out dur")
OUT, DUR = sys.argv[1], float(sys.argv[2])
from livekit.agents.utils.hw import get_cpu_monitor
from livekit.agents.worker import _DefaultLoadCalc
import livekit.agents as A
m = get_cpu_monitor()
# 1) monitorul brut al bibliotecii, cu log complet al eșantioanelor anormale
raw_bad = 0; raw_n = 0
def raw_loop():
    global raw_bad, raw_n
    t_end = time.time() + DUR
    while time.time() < t_end:
        t0 = time.monotonic(); v = m.cpu_percent(interval=0.5); dt = time.monotonic() - t0; raw_n += 1
        if v < 0 or v >= 1 or dt > 0.7:
            raw_bad += 1
            with open(OUT, "a") as f: f.write(f"{time.strftime('%H:%M:%S')} RAW cpu_percent={v:.4f} wall={dt:.3f}s\n")
# 2) media mobilă a bibliotecii (firul ei intern), citită la 0.5s ca în worker
calc = _DefaultLoadCalc()
th = threading.Thread(target=raw_loop, daemon=True); th.start()
t_end = time.time() + DUR; avg_bad = 0; avg_n = 0
while time.time() < t_end:
    time.sleep(0.5); a = calc._get_avg(); avg_n += 1
    if a >= 0.7 or a < 0:
        avg_bad += 1
        with open(OUT, "a") as f: f.write(f"{time.strftime('%H:%M:%S')} AVG get_avg={a:.4f} hist={[round(x,3) for x in calc._m_avg._hist]}\n")
th.join(5)
with open(OUT, "a") as f: f.write(f"DONE agents={A.__version__} monitor={type(m).__name__} ncpu={m.cpu_count()} raw_samples={raw_n} raw_bad={raw_bad} avg_samples={avg_n} avg_bad={avg_bad}\n")
```
</details>

### Workaround we are running (in case it helps)

Until this is fixed upstream we monkeypatch `CGroupV2CPUMonitor` at class level from our worker's import path: `_read_cpu_usage` reads the process's own cgroup (`/proc/self/cgroup` → `/sys/fs/cgroup/<path>/cpu.stat`, falling back to the root), `cpu_percent` divides by the measured elapsed time, drops samples with a negative delta or a delta larger than `elapsed × ncpu` (and logs them), and clamps to `[0, 1]`. A regression test asserts the upstream code is still unpatched, so we know when to remove it. Overnight the unpatched workers logged 104 FULL/AVAILABLE flips on 7 idle-ish workers and lost one more call (astral, 01:39:03, `no workers with sufficient capacity`); we will report the counts after a day on the workaround if useful.

Side note, since it is the same kind of "one bad sample kills a turn" class and we carry a monkeypatch for it as well: the Python google provider format still has the guard bug reported for agents-js in livekit/agents-js#2108 (`to_chat_ctx` checks `current_role` instead of the role of the last *produced* turn, so an empty user message after a barge-in yields a context ending in a `model` turn → Vertex 400 "Requests ending with a model turn"). We patch the result to always end with a user turn; happy to open a separate issue with the repro if you want it tracked here.

Alphaxiaoteng:
## Summary

Fixes #7102.

- Prefer the process's cgroup CPU usage, falling back to the mounted root when its path is unavailable or escapes the mount.
- Read effective `cpu.max` limits along that same cgroup's ancestor chain. A child with `max` can still inherit a restricted parent.
- Use measured elapsed time and normalize against effective capacity, clamping to `[0, 1]`.
- Start the elapsed-time window after the first counter read. Cap effective capacity by known host CPUs and the Linux scheduler affinity mask (which already reflects cpuset restrictions), preserving `NUM_CPUS` override precedence and unavailable-affinity fallbacks.
- Reject negative or host-capacity-impossible deltas. Quota-period/burst usage is not a physically impossible sample: applying a quota-rate ceiling to every short window incorrectly reported a saturated worker as idle.
- Give the regression file its required `pytest.mark.unit` category, addressing the failed hosted collection gate.

## Verification

- Archived the previously submitted `099dcbb` with only the new tests retained: **4 failed, 4 passed**. The failures reproduce both full/fractional quota bursts, nested cgroup normalization, and escaping-path behavior.
- The first follow-up `2e81a70`, archived with only the first additional edge-case tests overlaid, produced **4 failed, 9 passed**. These reproduce delayed first reads, quotas exceeding host capacity, tighter affinity, and unknown host capacity with known affinity.
- Current CPU + worker configuration + connection tests, with normal imports, conftest, and repository `--unit` selection: **30 passed, 1 warning**. This includes 19 CPU regression cases. The test fixture also emitted an unretrieved `SystemExit(0)` task diagnostic. Unchanged baseline worker tests passed 11 tests with the same synthetic short-JWT-key warning; the output is not claimed to be warning-free.
- Repository-pinned Ruff formatting/lint passed, including the full repository checks (976 files); `git diff --check` passed. Focused Linux-target mypy passed for the changed module (`--follow-imports=silent`).
- `make check` in the isolated environment passed formatting/lint but could not complete all-plugin type checks: optional plugin packages/typing dependencies are absent. Fresh hosted [CI](https://github.com/livekit/agents/actions/runs/33724885566) and [tests](https://github.com/livekit/agents/actions/runs/33724885501) passed for final commit `6f14da7`: full Python 3.10/3.13 type checks, all three blockguard platforms, and **2457 passed / 5 skipped / 34 warnings** in the unit-test job. This hosted result was verified from the actual job log.

## Limits and disclosure

The tests use tiny local file/time mocks plus actual package imports and repository test configuration. I have not run the reporter's Xen guest or an overnight production experiment. CLA confirmation still belongs to the account owner.

The initial patch was made with Cursor. Codex performed the follow-up repairs and tests, with independent agent review and main-agent reruns. This updates the same PR rather than creating a duplicate.

### DIFF
--- a/livekit-agents/livekit/agents/utils/hw/cpu.py
+++ b/livekit-agents/livekit/agents/utils/hw/cpu.py
@@ -62,11 +62,24 @@ class CGroupV2CPUMonitor(CPUMonitor):
                 break
             cgroup_path = parent
 
-        return quota_limit if quota_limit is not None else psutil.cpu_count() or 1.0
+        capacity_limits = [quota_limit] if quota_limit is not None else []
+        host_cpus = psutil.cpu_count()
+        if host_cpus is not None and host_cpus > 0:
+            capacity_limits.append(float(host_cpus))
+        # Linux affinity is already intersected with cpuset restrictions.
+        try:
+            affinity_cpus = len(os.sched_getaffinity(0))
+        except (AttributeError, OSError):
+            pass
+        else:
+            if affinity_cpus > 0:
+                capacity_limits.append(float(affinity_cpus))
+
+        return min(capacity_limits) if capacity_limits else 1.0
 
     def cpu_percent(self, interval: float = 0.5) -> float:
-        t0 = time.monotonic()
         cpu_usage_start = self._read_cpu_usage()
+        t0 = time.monotonic()
         time.sleep(interval)
         cpu_usage_end = self._read_cpu_usage()
         elapsed = time.monotonic() - t0
--- a/livekit-agents/livekit/agents/utils/hw/cpu.py
+++ b/livekit-agents/livekit/agents/utils/hw/cpu.py
@@ -47,10 +47,22 @@ class CGroupV2CPUMonitor(CPUMonitor):
         env_cpus = _cpu_count_from_env()
         if env_cpus is not None:
             return env_cpus
-        quota, period = self._read_cpu_max()
-        if quota == "max":
-            return psutil.cpu_count() or 1.0
-        return 1.0 * int(quota) / period
+        cgroup_path = os.path.dirname(self._cpu_stat_path())
+        quota_limit: float | None = None
+        while True:
+            quota, period = self._read_cpu_max(cgroup_path)
+            if quota != "max":
+                limit = int(quota) / period
+                quota_limit = limit if quota_limit is None else min(quota_limit, limit)
+
+            if cgroup_path == "/sys/fs/cgroup":
+                break
+            parent = os.path.dirname(cgroup_path)
+            if parent == cgroup_path:
+                break
+            cgroup_path = parent
+
+        return quota_limit if quota_limit is not None else psutil.cpu_count() or 1.0
 
     def cpu_percent(self, interval: float = 0.5) -> float:
         t0 = time.monotonic()
@@ -62,27 +74,34 @@ class CGroupV2CPUMonitor(CPUMonitor):
             return 0.0
 
         cpu_usage_diff = cpu_usage_end - cpu_usage_start
-        num_cpus = self.cpu_count()
-        # Discard torn/non-monotonic root-cgroup samples (negative or > available CPUs).
-        max_diff_usec = elapsed * num_cpus * 1_000_000
-        if cpu_usage_diff < 0 or cpu_usage_diff > max_diff_usec * 1.05:
+        # cpu.max is an average limit, so a sample can span a quota boundary or use
+        # accumulated burst runtime. Only host CPU capacity is a hard upper bound.
+        host_cpus = psutil.cpu_count()
+        max_diff_usec = (
+            elapsed * host_cpus * 1_000_000 if host_cpus is not None and host_cpus > 0 else None
+        )
+        if cpu_usage_diff < 0 or (
+            max_diff_usec is not None and cpu_usage_diff > max_diff_usec * 1.05
+        ):
             logger.warning(
-                "discarding impossible cgroup v2 cpu sample: start=%s end=%s elapsed=%.3fs ncpu=%s",
+                "discarding impossible cgroup v2 cpu sample: start=%s end=%s elapsed=%.3fs "
+                "host_ncpu=%s",
                 cpu_usage_start,
                 cpu_usage_end,
                 elapsed,
-                num_cpus,
+                host_cpus,
             )
             return 0.0
 
         # microseconds to seconds
         cpu_usage_seconds = cpu_usage_diff / 1_000_000
+        num_cpus = self.cpu_count()
         cpu_usage_percent = cpu_usage_seconds / (elapsed * num_cpus)
         return max(min(cpu_usage_percent, 1.0), 0.0)
 
-    def _read_cpu_max(self) -> tuple[str, int]:
+    def _read_cpu_max(self, cgroup_path: str = "/sys/fs/cgroup") -> tuple[str, int]:
         try:
-            with open("/sys/fs/cgroup/cpu.max") as f:
+            with open(os.path.join(cgroup_path, "cpu.max")) as f:
                 data = f.read().strip().split()
             quota = data[0]
             period = int(data[1]) if len(data) > 1 else 100000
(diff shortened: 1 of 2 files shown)


longcw:
closing in favor of #7113

PR #7104, opened 2026-09-03
feat(telemetry): full OTel GenAI semantic conventions + in-process PII stripping

theomonnom:


livekit-agents/livekit/agents/telemetry/trace_types.py
```diff
@@ -82,18 +82,175 @@
...
+    "gemini": "gcp.gemini",
+    "vertex": "gcp.vertex_ai",
+    "vertexai": "gcp.vertex_ai",
+    "google_vertex": "gcp.vertex_ai",
+    "watsonx": "ibm.watsonx.ai",
+    "xai": "x_ai",
+    "grok": "x_ai",
+    "mistral": "mistral_ai",
+    "moonshot": "moonshot_ai",
+}
```
chenghao-mou, line 266: I don't think the key values match our plugin values. e.g. Mistral LLM reports `MistralAI`:

Image

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -262,15 +269,25 @@ async def _traceable_main_task() -> None:
     @abstractmethod
     async def _run(self) -> None: ...
 
+    def _record_genai_request(self, span: trace.Span) -> None:
```
chenghao-mou, line 271: Codex informed met that [the best practice](https://github.com/open-telemetry/semantic-conventions-genai/blob/fee465db333bdd6a7d2faa320edab5cf3101a4f4/docs/gen-ai/gen-ai-spans.md#capturing-instructions-inputs-and-outputs) for these data:

> Full (buffered) content
Model instructions, user messages, and model outputs are considered sensitive and are often large in size.
>
> Recording large or sensitive content in telemetry may be problematic due to high storage costs, regulatory requirements, or the need to enforce different access models for operational and user data.
>
> OpenTelemetry instrumentations SHOULD NOT capture them by default, but SHOULD provide an option for users to opt in.

so it is an opt-in vs opt-out case, wdyt?

livekit-agents/livekit/agents/llm/llm.py
```diff
@@ -252,6 +258,7 @@ async def _traceable_main_task() -> None:
             ) as span:
                 for name, attributes in _chat_ctx_to_otel_events(self._chat_ctx):
                     span.add_event(name, attributes)
```
chenghao-mou: do we still need the for loop here if we record the request in the following line?

livekit-agents/livekit/agents/telemetry/trace_types.py
theomonnom: This is part of the spec:

> For the 16 enumerated providers (openai, anthropic, mistral_ai, aws.bedrock, gcp.vertex_ai, gcp.gemini, gcp.gen_ai, x_ai, perplexity, groq, cohere, deepseek, azure.ai.*, ibm.watsonx.ai, moonshot_ai) the registry spelling is MUST

chenghao-mou: Yes, I was referring to "key" values. I think we missed `Amazon`.

livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -414,18 +419,97 @@ def _translate(self, record: logging.LogRecord) -> OTelLogRecord:
...
+        pii._PIIFilteringSpanProcessor(allow_pii=allow_pii if allow_pii is not None else True),
+    )
+
+
 def set_tracer_provider(
-    tracer_provider: trace_api.TracerProvider, *, metadata: dict[str, AttributeValue] | None = None
+    tracer_provider: trace_api.TracerProvider,
+    *,
+    metadata: dict[str, AttributeValue] | None = None,
+    allow_pii: bool | None = None,
```
chenghao-mou, line 495: Is `allow_pii` too broad a name? As a user, I read `allow_pii=False` as “PII reaches no exporter,” but LiveKit Cloud still receives the original content unless project redaction is enabled. The docstring also starts with “this provider’s exporters,” which reinforces that interpretation.

livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -336,7 +337,9 @@ def __init__(self, inner: SpanExporter, export_jobs: Mapping[str, _JobTelemetry]
 
     def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
         exportable = [
-            s
+            # PII stripped for third-party exporters is put back here: what LiveKit Cloud
+            # may receive is the project's setting, applied at its collector
+            pii.restore_pii(s)
```
chenghao-mou, line 343: nit: should we use 

```python
s
if telemetry_utils.redaction_enabled(s.attributes)
else pii.restore_pii(s)
```

I know we have the project redaction check upstream, but if we ever change that part, this is an easy to miss leakage point.

chenghao-mou (approved):
LGTM, two nits.

livekit-agents/livekit/agents/telemetry/traces.py
theomonnom: That’s fair but you only use this method when you setup an OTEL endpoint. 

For LiveKit, it’s just there, there is no setup

PR #7116, opened 2026-09-04
fix: keep the gateway routing stamp on tool-only turns

u9g:
## Problem

The inference gateway stamps every streamed chunk with the deployment that served it (`choices[0].delta.extra_content.livekit.inference_deployment`) and, when the stamp comes back on an assistant message, pins the next turn to that deployment so the prompt cache warmed by the previous turn is reused.

The SDK only stored the stamp on the assistant `ChatMessage`, and that message is only created when the turn produced text. A turn that produced tool calls and no text dropped the stamp, so every follow-up turn of a tool-driven agent was re-routed at random across providers.

Measured on the simulation summarizer in production over 7 days: 0 of 28,699 gateway routing decisions for that project carried a preferred deployment, and 36% of consecutive reviewer turns switched deployment, each one missing the prefix it had just cached.

## Fix

- `agent_activity`: when no assistant message is stored, carry the generation's extra on the function calls of that turn.
- OpenAI serializer: for a tool-only group, emit the `livekit` extra on the synthesized assistant message, where the gateway reads it. Tool-level `extra_content` is unchanged.

## Verification

- New unit tests: serializer places the stamp on the tool-only assistant message; a fake-session tool turn with no text keeps the stamp on the recorded function call. Both failed before the change.
- Replayed the summarizer's reviewer path against the production gateway with this branch: the follow-up request now carries `extra_content.livekit.inference_deployment` on the assistant message.
- `make unit-tests`: all pass except `tests/test_room.py`, which fails identically on `main` in my environment (local WebSocket connect error).
- `make type-check`: no new errors.

agents-js has the same gap (`agent_activity.ts` stores `generatedExtra` only under `if (forwardedText)`, and `provider_format/openai.ts` synthesizes a bare `{ role: 'assistant' }` for tool-only groups). Follow-up PR there.

### DIFF
--- a/livekit-agents/livekit/agents/llm/_provider_format/openai.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/openai.py
@@ -24,7 +24,9 @@ def to_chat_ctx(
             continue
 
         # one message can contain zero or more tool calls
-        msg = _to_chat_item(group.message) if group.message else {"role": "assistant"}
+        msg: dict[str, Any] = (
+            _to_chat_item(group.message) if group.message else {"role": "assistant"}
+        )
         tool_calls = []
         for tool_call in group.tool_calls:
             tc: dict[str, Any] = {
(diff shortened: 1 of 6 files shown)


u9g:
Closing in favor of the gateway-side fix, which covers clients already in the field without an SDK upgrade: see the agent-gateway PR "fix(inference): carry the deployment stamp on tool calls so tool-only turns stay pinned".

PR #7118, opened 2026-09-04
fix: keep the forkserver preload shared with job processes (gc.freeze)

### LINKED ISSUE #7117: forkserver preload's COW sharing is destroyed by the GC in each job process (~96MB per idle worker)
## Summary

On Linux the worker preloads every registered plugin package plus
`livekit.agents.inference._warmup` into the forkserver process, so that job
processes inherit the already-imported modules and the warmed native models
through copy-on-write (`worker.py`, `set_forkserver_preload`).

That sharing does not survive. Each job process runs its own cyclic garbage
collector, and the first full collection in the child writes to nearly every
page that holds a preloaded Python object, forcing the kernel to copy it. The
memory the preload was supposed to save comes back, per process.

## Why it happens

CPython puts a `PyGC_Head` (two pointers) immediately *before* every
GC-tracked object. A collection pass rewrites those two words for every object
in the generation being scanned — it is a write, even when nothing is
collected. Preloaded modules, classes, functions, docstring/annotation dicts,
enum members, pydantic models etc. are long-lived, so they all end up in the
oldest generation. The first full collection inside a forked job process
therefore touches the header of essentially every preloaded object and
un-shares the pages holding them.

To confirm the scale, a forkserver preloading just
`livekit.agents.ipc` transitively (i.e. `livekit.agents` + its deps) hands each
forked child **233,838** GC-tracked objects. Those objects are spread across
almost every page of the inherited heap, so a full pass dirties nearly all of
it.

Two things this is *not*:

- It is not the native model weights. `init_vad()` / `init_eot()` allocate
  outside the Python heap and stay shared — that part of the preload works.
- It is not fixed by fewer jobs. It is per process and it happens without any
  job traffic: an idle prewarmed process reaches a full collection during
  `initialize_process_fnc` / prewarm, so the pages are already private before
  the first job is ever assigned.

Observed on a Linux worker: **~96 MB per idle job process** moves from shared
to private. With `num_idle_processes` > 1 that multiplies directly.

## Reproducing

On Linux, for a running worker:

```bash
# forkserver pid, then any job process pid
grep -E '^(Shared_Clean|Private_Dirty)' /proc/<job-pid>/smaps_rollup
```

Sample it right after the process is forked, then again after the process has
run a full collection (`gc.collect(2)`, or just let it idle through a prewarm).
`Shared_Clean` drops and `Private_Dirty` rises by the same amount.

## Fix

`gc.freeze()` in the forkserver process once everything is preloaded. It moves
every currently tracked object into the *permanent generation*, which the
collector never scans, so their GC headers are never rewritten and the pages
stay shared for the life of the child. Frozen objects are still freed normally
by refcounting if they ever die; freezing module-level state costs nothing
because it lives for the process lifetime anyway.

It has to run *inside the forkserver*, not in the worker — job processes fork
from the forkserver, so freezing in the parent has no effect on them. The
codebase already has the mechanism for running code there: a side-effect module
in the preload list (`livekit.agents.inference._warmup`). So the fix is one
more such module, listed last:

```python
# livekit/agents/ipc/_preload_freeze.py
import gc

gc.collect()   # don't make import-time cycle garbage permanent
gc.freeze()
```

```python
plugin_packages = [p.package for p in Plugin.registered_plugins] + [
    "av",
    "livekit.agents.inference._warmup",
    "livekit.agents.ipc._preload_freeze",   # keep last
]
```

Kept import-free on purpose: the forkserver's preload loop swallows
`ImportError`, so putting the `gc.freeze()` call at the end of `_warmup.py`
would silently skip it on any host where `livekit-local-inference` fails to
import — while the plugin preload, and therefore the problem, is still there.

Side benefit: those ~230k objects also stop being walked on every full
collection in the job process, which removes a recurring GC pause from the
audio path.

Alternatives considered and rejected:

- `gc.freeze()` in the worker process — wrong process, job processes fork from
  the forkserver.
- `gc.disable()` or raised thresholds in the job process — leaks real cycles
  for the whole lifetime of a long-running session.

Verified that the freeze survives the fork: a child forked from a forkserver
that preloaded the module reports `gc.get_freeze_count() == 233838`.

No behaviour change on `spawn` (macOS/Windows), where there is no preload to
begin with.

Darshak03:
### Problem

On Linux the worker preloads the registered plugin packages plus
`livekit.agents.inference._warmup` into the forkserver process so job processes
inherit the imported modules and the warmed native models copy-on-write.

The sharing does not last. Every job process runs its own cyclic GC, and a full
collection rewrites the `PyGC_Head` (two pointers stored immediately before each
tracked object) of every object in the oldest generation — a write, even when
nothing is collectable. Preloaded modules, classes, functions and their dicts
are long-lived, so they all live there: the first full collection in a forked
child dirties nearly every page holding a preloaded object and the kernel copies
it. Measured on a Linux worker at ~96MB moving from shared to private per job
process, and it happens with no job traffic at all — an idle prewarmed process
already crosses a full collection during prewarm.

The native weights from `init_vad()` / `init_eot()` are allocated outside the
Python heap and stay shared; it is the Python object graph that is lost.

### Fix

`gc.freeze()` in the forkserver, once everything is preloaded. Frozen objects go
to the permanent generation, which the collector never scans, so their headers
are never rewritten and the pages stay shared for the life of the job process.
They are still freed by refcounting if they die, and module-level state lives for
the whole process anyway, so nothing is retained that would not have been.

It has to run inside the forkserver — job processes fork from there, so freezing
in the worker would not affect them. This reuses the mechanism already in the
tree for that: a side-effect module in the preload list, like
`livekit.agents.inference._warmup`. The new module is listed last so everything
above it is resident by the time it runs.

The module deliberately imports nothing but `gc`. The forkserver's preload loop
swallows `ImportError`, so putting the call at the end of `_warmup.py` would
silently skip the freeze on any host where `livekit-local-inference` fails to
import, even though the plugin preload — and the problem — is still there.

Also drops ~230k objects from every full collection in the job process, removing
a recurring GC pause from the audio path.

### Verification

A child forked from a forkserver that preloads the new module reports
`gc.get_freeze_count() == 233838`, i.e. the freeze survives the fork and covers
the whole preloaded graph.

To see the memory effect on a Linux worker, sample a job process right after it
is forked and again after a full collection:

```bash
grep -E '^(Shared_Clean|Private_Dirty)' /proc/<job-pid>/smaps_rollup
```

`Shared_Clean` drops and `Private_Dirty` rises by the same amount without this
change; with it, both hold.

### Notes

- No behaviour change on `spawn` (macOS/Windows) — there is no preload there.
- Alternatives rejected: `gc.freeze()` in the worker (wrong process), and
  `gc.disable()` / raised thresholds in the job process (leaks real cycles for
  the lifetime of a long session).

Fixes #7117

### DIFF
--- a/livekit-agents/livekit/agents/ipc/_preload_freeze.py
+++ b/livekit-agents/livekit/agents/ipc/_preload_freeze.py
@@ -1,0 +1,17 @@
+"""Side-effect module: imported **last** in the forkserver preload list.
+
+A forked job process shares the forkserver's pages copy-on-write, but the first
+gen2 collection in the child traverses every tracked object and rewrites its GC
+header, dirtying (and so un-sharing) nearly every page holding preloaded
+objects. ``gc.freeze()`` moves them to the permanent generation, which the
+collector never visits, so those pages stay shared for the life of the job.
+
+Kept free of imports on purpose: the forkserver swallows ``ImportError`` from
+preload modules, so anything importable that can fail would silently skip the
+freeze.
+"""
+
+import gc
+
+gc.collect()  # don't make import-time cycle garbage permanent
+gc.freeze()
--- a/livekit-agents/livekit/agents/worker.py
+++ b/livekit-agents/livekit/agents/worker.py
@@ -752,6 +752,9 @@ class AgentServer(utils.EventEmitter[EventTypes]):
                 plugin_packages = [p.package for p in Plugin.registered_plugins] + [
                     "av",
                     "livekit.agents.inference._warmup",
+                    # keep last: freezes everything preloaded above so the job
+                    # processes' GC doesn't un-share those pages (see module docstring)
+                    "livekit.agents.ipc._preload_freeze",
                 ]
                 logger.info("preloading plugins", extra={"packages": plugin_packages})
                 self._mp_ctx.set_forkserver_preload(plugin_packages)


Darshak03:
Hi @chenghao-mou,
Could you please help me understand why my changes haven’t been merged? Is there any specific format or guideline I need to follow? I’ve also signed the CLA.

PR #7121, opened 2026-09-04
plugins/phonic: add enable_watermarking realtime config option

arunwpm-work:
Adds an `enable_watermarking` option to the Phonic realtime plugin, gating Phonic's inaudible provenance watermark on the agent's generated audio. Boolean, off by default; adds a very small amount of latency when enabled.

Mirrors the existing `enable_redaction` option across `RealtimeModel.__init__`, `update_options` (model + session), the config message sent to Phonic, and the README.

Draft — depends on Phonic's `enable_watermarking` per-agent config reaching prod (control plane + engine); harmless to send before then (Phonic ignores unknown config fields).


samanyugoyal2010:
@arunwpm-work 
Could we add a small test covering `enable_watermarking`?

In particular, it'd be useful to verify:

* the constructor value makes it into the config payload
* `RealtimeSession.update_options()` updates the payload/reset behavior
* `NOT_GIVEN` leaves the existing value unchanged

That should help catch future cases where a realtime option gets added to one path but not another.

tinalenguyen (approved):

PR #5861, opened 2026-05-27
fix(llm): preserve Field() constraints on function tool arguments

theomonnom:

PR #7126, opened 2026-09-05
fix(llm): preserve Annotated tool argument metadata

Kuang-xianxin:
Tool argument conversion unwraps `Annotated` and rebuilds only metadata stored inside its first `Field`. This loses constraints from types such as `PositiveInt` and skips `BeforeValidator`/`AfterValidator` callbacks, including when an annotation also contains `Field(description=...)`.

Parse the complete annotation with `FieldInfo.from_annotation`, then set signature-default and docstring-description fallbacks directly on that field. Pass the base annotation and the complete field to `create_model`, preserving the order of validators and constraints. Raise the declared Pydantic lower bound to `>=2.5,<3` and synchronize the lockfile metadata: the existing `pydantic_core.from_json` import already requires that floor. This removes the obsolete compatibility fork and deprecated `merge_field_infos` call.

Tests cover constraints, validators and their ordering, aliases, defaults, default factories, and repeated conversion without changing caller-owned Field metadata. For example, `count: PositiveInt` now advertises and enforces its lower bound, and normalization validators run before the tool receives its arguments.

Validation:

- Four initial regression cases failed before the original metadata fix.
- Pydantic 2.13.4 / Python 3.13.14: all 177 tool/proxy/search tests pass using `--real-time`.
- Pydantic 2.5.0 / Python 3.11.15: the actual SDK imports successfully; 174 related tests pass. Three existing schema assertions for unrestricted dictionaries and literal-string types fail identically on the unmodified upstream source at `d8607e6`. These are not claimed as passing minimum-version checks.
- Built and installed the wheel on Pydantic 2.5.0: all eight added annotation regressions pass, and installed metadata declares `pydantic<3,>=2.5`.
- Scoped Ruff lint/format and source-module mypy pass. Wheel build and `uv lock --check --offline` pass; no locked package versions changed.

Found through source inspection; no matching issue or competing fix was identified. This extends the Field-constraint coverage from #5861.


livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -519,19 +519,13 @@ def function_arguments_to_pydantic_model(func: Callable[..., Any]) -> type[BaseM
         # Annotated[str, Field(description="...")]
         if get_origin(type_hint) is Annotated:
             annotated_args = get_args(type_hint)
-            type_hint = annotated_args[0]
```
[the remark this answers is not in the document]
Kuang-xianxin: Confirmed on Pydantic 2.11.9: the separate Field moved MinLen ahead of the AfterValidator, so the validator could return an empty string without rejection. Fixed in eac9a5583c9f7bdccd228affe8820c87dbbf133b by parsing the annotation with FieldInfo.from_annotation and merging only the default/description fallbacks. create_model now receives the base annotation and a single complete FieldInfo. Added the reported ordering regression plus alias/default/default-factory coverage. All 176 related tests pass on Pydantic 2.13.4 and 2.11.9 (using --real-time for the older version).

livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -513,46 +511,17 @@ def function_arguments_to_pydantic_model(func: Callable[..., Any]) -> type[BaseM
...
 
-        if not field_info:
-            field_info = Field(**field_attrs)
-        else:
-            for k, v in field_attrs.items():
-                setattr(field_info, k, v)
-
-        fields[param_name] = (type_hint, field_info)
+        field_info = FieldInfo.merge_field_infos(field_info, **field_attrs)
+        fields[param_name] = (field_info.annotation, field_info)
```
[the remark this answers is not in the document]
Kuang-xianxin: Addressed in 0cf4910a6e0492ae4765a38f25b4dc9cf9182c42. Pydantic 2.0 actually has no merge_field_infos method; for versions with the method, the conversion now explicitly preserves the parsed annotation and metadata after merging fallback attributes. The 2.0 path sets those fallback attributes on the parsed field without calling the missing method. Added a regression covering reusable Field metadata with signature-default and docstring-description fallbacks.

Validation: 177 tool/proxy/search tests pass on Pydantic 2.13.4; scoped Ruff, formatting and mypy pass. A separate probe executing the conversion helper from the source with real Pydantic dependencies passes constraint, validator, alias/default and docstring cases on 2.0 and 2.1.0. This probe isolates the helper: importing the full SDK with Pydantic 2.0 already fails in the unrelated Metadata.model_name model due to that version's protected namespace rule, so I am not claiming whole-package compatibility with 2.0.

livekit-agents/livekit/agents/llm/utils.py
```diff
@@ -513,46 +511,27 @@ def function_arguments_to_pydantic_model(func: Callable[..., Any]) -> type[BaseM
...
         else:
-            for k, v in field_attrs.items():
-                setattr(field_info, k, v)
-
-        fields[param_name] = (type_hint, field_info)
+            merged_field_info = merge_field_infos(field_info, **field_attrs)
+            # Early Pydantic versions drop these during the merge.
+            merged_field_info.annotation = field_info.annotation
+            merged_field_info.metadata = field_info.metadata
+            field_info = merged_field_info
```
longcw, line 533: the `merge_field_infos` branch is dead on one side (pydantic 2.0.3 already has it) and deprecated on the other, maybe drop the fork and keep only the `setattr` loop, which produces the same model on every supported version.

longcw (approved):
thanks for the pr! something nit:

livekit-agents/livekit/agents/llm/utils.py
Kuang-xianxin: Thanks for the review. I tried the single `setattr` loop: the 177 related tests pass on Pydantic 2.13.4, but it loses signature defaults and docstring descriptions on 2.1.0 (still within `pydantic>=2.0,<3`). Here is a standalone reproduction on 2.1.0:

```python
from typing import Annotated
from pydantic import Field, create_model
from pydantic.fields import FieldInfo

field = FieldInfo.from_annotation(Annotated[int, Field(gt=0)])
field.default = 2
field.description = "Number of items."
model = create_model("Args", count=(field.annotation, field))
print(model.model_json_schema())  # count is required; no default/description
model()  # ValidationError: count / Field required
```

On that version, later field reconstruction uses `_attributes_set`, so direct assignment is not retained. The current PR implementation passes the same default/docstring checks plus constraint and validator checks using the converter extracted from its actual source. This is isolated converter validation, not a whole-SDK compatibility claim for early Pydantic.

I have kept the existing implementation for now to preserve that behavior. The fallback also covers 2.0.0, which has no `merge_field_infos` and remains inside the declared range.

longcw: Thanks for the check. I can reproduce the failure, and it happens on pydantic 2.1.0 only. On 2.0.0, 2.0.3, 2.1.1 and every later release, direct assignment on the `FieldInfo` keeps the default and the description.

But this module cannot load on 2.1.0. It imports `from_json` from `pydantic_core`, and that name exists only from pydantic 2.5:

https://github.com/livekit/agents/blob/0cf4910a6e0492ae4765a38f25b4dc9cf9182c42/livekit-agents/livekit/agents/llm/utils.py#L23

On 2.0.x through 2.4.x the import fails before `function_arguments_to_pydantic_model` runs. So the fork protects a version that cannot reach this code. The `pydantic>=2.0` bound in [pyproject.toml](https://github.com/livekit/agents/blob/c6939a5b86e4e797ab821ebb9bf7a6892071e328/livekit-agents/pyproject.toml#L46) is out of date. The real floor is 2.5, and it moved there with #5409.

I would suggest two changes for this PR:

1. Raise the bound to `pydantic>=2.5,<3` in `livekit-agents/pyproject.toml`. This changes nothing for users, because the package does not import below 2.5 today.
2. Remove the fork and the `field_attrs` dict. Assign the two fallbacks directly on the parsed `FieldInfo`:

https://github.com/livekit/agents/blob/0cf4910a6e0492ae4765a38f25b4dc9cf9182c42/livekit-agents/livekit/agents/llm/utils.py#L523-L533

```python
field_info = FieldInfo.from_annotation(type_hint)
if default_value is not ... and field_info.default is PydanticUndefined:
    field_info.default = default_value
if field_info.description is None:
    field_info.description = param_docs.get(param_name)
fields[param_name] = (field_info.annotation, field_info)
```

I ran this version and your fork against your eight new test cases plus a nine-parameter legacy signature on pydantic 2.5.3, 2.10.6, 2.11.9, 2.12.5 and 2.13.4. Both pass every case on every version, and both produce a schema identical to `main` for the legacy shapes. This version does not call `merge_field_infos`, which is deprecated since 2.12, and does not need to restore `annotation` and `metadata` by hand.

Kuang-xianxin: Addressed in a6b99fd. You are right: the isolated 2.1.0 probe did not establish a reachable SDK compatibility requirement. The manifest now declares pydantic>=2.5,<3, the lockfile metadata is synchronized, and the converter assigns both fallbacks directly without field_attrs or merge_field_infos.

Validation: 177 related tests pass on 2.13.4. On 2.5.0 the real SDK imports and all eight added annotation regressions pass, including after installing the built wheel. The broader minimum-version run has 174 passes and three existing schema-assertion failures reproduced on unmodified d8607e6 (unrestricted dictionaries/literal-string types); these are documented in the PR body. Scoped Ruff/format, source mypy, wheel build, and lock consistency checks pass.

PR #7127, opened 2026-09-05
fix(telemetry): describe option objects in the session report

davidzhao:
## What

`session.options` in the session report carried the turn detector as its default repr:

```
turn_detection: "<livekit.agents.inference.eot.detector.TurnDetector object at 0x7fc5518a2a50>"
```

The OTel exporter stringifies any non-primitive log attribute, so anything left as an object in the options dict ends up like this.

## How

- `_serialize_session_options` now serializes explicitly: primitives pass through, mappings recurse, any non-string `Sequence` or `Set` is serialized elementwise (`tts_text_transforms` accepts any `Sequence`), and any other object is reported by class, `module.Class`. An object that wants its configuration shown implements the `telemetry.DescribesOptions` protocol, a `describe_options()` method returning the options worth reporting; it is then rendered as `module.Class(k=v, ...)`. The object decides what is safe to show, so nothing is guessed from public attributes and credentials and endpoints stay out by construction (test included; a failing `describe_options()` falls back to the class name).
- Customer-authored prompt text is never written to the report: `instructions` keys (the keyterm-detection prompt override) are dropped at any depth rather than emitted under an unmarked key.
- `TurnDetector` implements `describe_options()`: model, provider, sample rate, local fallback, and the threshold overrides only when the user set any. Mode strings (`"vad"`, `"manual"`, ...) pass through unchanged. Third-party detectors and models opt in the same way.

Result:

```
livekit.agents.inference.eot.detector.TurnDetector(model=turn-detector-v1-mini, provider=livekit, sample_rate=16000, local_fallback=True)
```

with `threshold_overrides` / `backchannel_threshold_overrides` appended only when the user set them.

## Tests

`tests/test_session_options_report.py` (unit): default detector description, threshold overrides, no credential leakage, mode strings, plugin-like objects, custom `Sequence` and `Set` values, omitted prompt text, nested containers and the `keyterms` PII alias, and a JSON-safety invariant on the whole serialized dict.

Bottom layer of the tracing coverage stack.

### DIFF
--- a/livekit-agents/livekit/agents/inference/eot/detector.py
+++ b/livekit-agents/livekit/agents/inference/eot/detector.py
@@ -130,6 +130,24 @@ class TurnDetector(_BaseStreamingTurnDetector):
     def model(self) -> TurnDetectorModels:
         return self._model
 
+    @property
+    def sample_rate(self) -> int:
+        return self._opts.sample_rate
+
+    @property
+    def local_fallback(self) -> bool:
+        return self._local_fallback
+
+    @property
+    def threshold_overrides(self) -> NotGivenOr[float | dict[str, float]]:
+        """User ``unlikely_threshold`` override(s); NOT_GIVEN when server defaults apply."""
+        return self._opts.thresholds.overrides
+
+    @property
+    def backchannel_threshold_overrides(self) -> NotGivenOr[float | dict[str, float]]:
+        """User ``backchannel_threshold`` override(s); NOT_GIVEN when server defaults apply."""
+        return self._opts.thresholds.backchannel_overrides
+
     def _warn_threshold_override(self) -> None:
         thresholds = self._opts.thresholds
         if is_given(overrides := thresholds.overrides):
--- a/livekit-agents/livekit/agents/inference/eot/detector.py
+++ b/livekit-agents/livekit/agents/inference/eot/detector.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import replace
+from typing import Any
 
 import aiohttp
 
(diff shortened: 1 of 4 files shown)


livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -84,17 +85,75 @@
...
+# internals never end up in the report.
+_OPTION_OBJECT_DESCRIPTOR_ATTRS = (
+    "model",
+    "provider",
+    "label",
+    "sample_rate",
+    "local_fallback",
+    "threshold_overrides",
+    "backchannel_threshold_overrides",
+)
```
chenghao-mou: I wonder if it is cleaner to have a serialization/reporting protocol for those classes.

davidzhao: yeah that's true.. I've thought about that too. but thought it would be complicated to enforce (esp for existing items).. but let me give that a shot

davidzhao: done, using a `describe_options` method as part of a protocol. this feels cleaner.

chenghao-mou (approved):

livekit-agents/livekit/agents/inference/eot/detector.py
```diff
@@ -130,6 +131,23 @@ def __init__(
...
+            "model": self.model,
+            "provider": self.provider,
+            "sample_rate": self._opts.sample_rate,
+            "local_fallback": self._local_fallback,
+        }
+        thresholds = self._opts.thresholds
+        if is_given(thresholds.overrides):
+            options["threshold_overrides"] = thresholds.overrides
+        if is_given(thresholds.backchannel_overrides):
+            options["backchannel_threshold_overrides"] = thresholds.backchannel_overrides
```
u9g, line 148: I think it would have made sense to call this `threshold_backchannel_overrides`, in the style of the rest of the names.

PR #7130, opened 2026-09-05
feat(telemetry): trace the end-of-turn wait and the turn handoff

davidzhao:
## What

`eou_detection` only covered the turn-detector inference, so the endpointing delay itself (often seconds) showed up as an empty gap between `user_turn` and `agent_turn`. This makes the wait, the user hook, and the speech queue visible.

## How

**`eou_wait`** span, one per user turn, child of `user_turn`:
- back-dated to `last_speaking_time`, so the bar starts where the user stopped talking;
- ends on the turn decision with `lk.eou.outcome` (`committed` / `user_resumed` / `dropped`), `lk.eou.wait_duration`, `lk.eou.rearm_count`, the final `lk.eou.endpointing_delay`, and the last `lk.eou.source`;
- a later trigger for the same turn (late STT final, another VAD end of speech) **re-arms** the wait and records a `rearmed` event instead of starting a new span, so the bar stays whole and late transcripts are readable off it;
- a rejected commit (`min_words`, realtime backchannel) records a `lk.eou.not_committed_count` and keeps waiting; resumed speech ends the span at the VAD/STT speech start; teardown drops it.
- resumed speech ends the span at the VAD/STT speech start, kept as a `user_resumed` event; when that start is reported after the fact and a detection or re-arm was recorded in between, the bar runs to the last of those so it always contains its children (an `eou_detection` still running is closed with a `superseded` event). A turn-detection mode change to or from `manual` drops the pending wait.

`eou_detection` keeps its name and now nests under `eou_wait`; the span names and the `lk.eou.*` attribute keys stay on the same vocabulary, so nothing on the cloud side needs to change.

**`on_user_turn_completed`** span around the user hook that gates the reply, nested under the `user_turn` it completes: recognition hands the open `user_turn` span to the activity when it schedules the hook, and the activity ends it after the hook (or when the turn is skipped), so the turn's duration covers the wait for the hook. Exceptions are recorded redaction-aware, honouring `record={"redaction": True}` on the session as well as the job's flag; `StopResponse` is an event.

**`lk.speech.queue_wait`** on `agent_turn`: seconds between scheduling the speech and its first generation authorization.

```
user_turn
├─ user_speaking
└─ eou_wait                 ← new, 2.5s, rearmed ×1
   └─ eou_detection         ← now nested under the wait
on_user_turn_completed      ← new
agent_turn  (lk.speech.queue_wait=0.01)
├─ llm_node
└─ tts_node
```

## Tests

`tests/test_eou_wait_span.py` drives `_run_eou_detection` with real spans: commit (parent, back-dated start, attributes), re-arm keeps one span, resumed speech ends at speech start and leaves `user_turn` open, teardown drops, `not_committed` keeps waiting, `eou_detection` nests under the wait and stays inside it when the user resumes mid-inference, a mode change drops the wait; plus a full fake session asserting `eou_wait`, the hook span, the queue-wait attribute, and a session-only redaction keeping a hook exception's message out of the trace.

Stacked on #7128.

- **Counts, not events.** `lk.eou.not_committed_count` on `eou_wait` (turn decisions the detector rejected before the wait ended) replaces the `not_committed` event; `lk.eou.resume_count` on `user_turn` counts the waits the user cut short by speaking again.
- **Stages on the reply.** The reply's `agent_turn` carries `lk.end_of_turn_delay`, `lk.transcription_delay` and `lk.on_user_turn_completed_delay` next to `lk.e2e_latency`, so a per-turn breakdown reads off one span.
- A `user_turn` created lazily from an STT final (no VAD opened it) starts at the earliest anchor known, so its back-dated `eou_wait` child never starts before it.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -215,6 +215,17 @@ class _PausedSpeechInfo:
 
 
 # NOTE: AgentActivity isn't exposed to the public API
+def _record_queue_wait(speech_handle: SpeechHandle) -> None:
+    """Stamp how long the speech sat in the queue on its agent_turn span.
+
+    Module-level on purpose: the reply tasks are also driven with lightweight stand-ins for
+    the activity in tests, which must not need to know about telemetry helpers."""
+    if (queue_wait := speech_handle._queue_wait()) is None:
+        return
+    span = trace.get_current_span(context=speech_handle._agent_turn_context)
+    span.set_attribute(trace_types.ATTR_SPEECH_QUEUE_WAIT, queue_wait)
+
+
 class AgentActivity(RecognitionHooks):
     def __init__(self, agent: Agent, sess: AgentSession) -> None:
         self._agent, self._session = agent, sess
@@ -1830,13 +1841,6 @@ class AgentActivity(RecognitionHooks):
             skip_reply=skip_reply,
         )
 
-    def _record_queue_wait(self, speech_handle: SpeechHandle) -> None:
-        """Stamp how long the speech sat in the queue on its agent_turn span."""
-        if (queue_wait := speech_handle._queue_wait()) is None:
-            return
-        span = trace.get_current_span(context=speech_handle._agent_turn_context)
-        span.set_attribute(trace_types.ATTR_SPEECH_QUEUE_WAIT, queue_wait)
-
     def _schedule_speech(self, speech: SpeechHandle, priority: int, force: bool = False) -> None:
         # when force=True, we still allow to schedule a new speech even if
         # `pause_speech_scheduling` is waiting for the schedule_task to drain.
@@ -2999,7 +3003,7 @@ class AgentActivity(RecognitionHooks):
             authorization_tasks.append(asyncio.ensure_future(self._user_silence_event.wait()))
         await speech_handle.wait_if_not_interrupted(authorization_tasks)
         speech_handle._clear_authorization()
-        self._record_queue_wait(speech_handle)
+        _record_queue_wait(speech_handle)
 
         if speech_handle.interrupted:
             current_span.set_attribute(trace_types.ATTR_SPEECH_INTERRUPTED, True)
@@ -3494,7 +3498,7 @@ class AgentActivity(RecognitionHooks):
             authorization_tasks.append(asyncio.ensure_future(self._user_silence_event.wait()))
         await speech_handle.wait_if_not_interrupted(authorization_tasks)
         speech_handle._clear_authorization()
-        self._record_queue_wait(speech_handle)
+        _record_queue_wait(speech_handle)
 
         if speech_handle.interrupted:
             current_span.set_attribute(trace_types.ATTR_SPEECH_INTERRUPTED, True)
@@ -3896,7 +3900,7 @@ class AgentActivity(RecognitionHooks):
         if speech_handle.allow_interruptions and not self._rt_overlapping_speech_enabled:
             authorization_tasks.append(asyncio.ensure_future(self._user_silence_event.wait()))
         await speech_handle.wait_if_not_interrupted(authorization_tasks)
-        self._record_queue_wait(speech_handle)
+        _record_queue_wait(speech_handle)
         if speech_handle.interrupted:
             await utils.aio.cancel_and_wait(*authorization_tasks)
             return
@@ -4138,7 +4142,7 @@ class AgentActivity(RecognitionHooks):
             authorization_tasks.append(asyncio.ensure_future(self._user_silence_event.wait()))
         await speech_handle.wait_if_not_interrupted(authorization_tasks)
         speech_handle._clear_authorization()
-        self._record_queue_wait(speech_handle)
+        _record_queue_wait(speech_handle)
 
         if speech_handle.interrupted:
             # nothing was played, but the response may still be generating server-side
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -228,8 +228,10 @@ def _record_queue_wait(speech_handle: SpeechHandle) -> None:
     """Stamp how long the speech sat in the queue on its agent_turn span.
 
     Module-level: tests drive the reply tasks with stand-in activities."""
-    if (queue_wait := speech_handle._queue_wait()) is None:
-        return
+    if (
+        queue_wait := speech_handle._queue_wait()
+    ) is None or speech_handle._agent_turn_context is None:
+        return  # no agent_turn span yet: never fall back to whatever span is current
     span = trace.get_current_span(context=speech_handle._agent_turn_context)
     span.set_attribute(trace_types.ATTR_SPEECH_QUEUE_WAIT, queue_wait)
 
@@ -3927,7 +3929,7 @@ class AgentActivity(RecognitionHooks):
         if speech_handle.allow_interruptions and not self._rt_overlapping_speech_enabled:
             authorization_tasks.append(asyncio.ensure_future(self._user_silence_event.wait()))
         await speech_handle.wait_if_not_interrupted(authorization_tasks)
-        _record_queue_wait(speech_handle)
+        # the queue wait is recorded by _realtime_generation_task, which owns the agent_turn span
         if speech_handle.interrupted:
             await utils.aio.cancel_and_wait(*authorization_tasks)
             return
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -224,6 +224,20 @@ def _end_user_turn_span(info: _EndOfTurnInfo) -> None:
         info.user_turn_span_adopted = False
 
 
+def _record_user_turn_stages(span: trace.Span, user_metrics: llm.MetricsReport) -> None:
+    """The stages between the user stopping and the reply starting, next to lk.e2e_latency on
+    the reply's agent_turn: a per-turn breakdown readable off one span."""
+    attrs: dict[str, float] = {}
+    if (v := user_metrics.get("end_of_turn_delay")) is not None:
+        attrs[trace_types.ATTR_END_OF_TURN_DELAY] = v
+    if (v := user_metrics.get("transcription_delay")) is not None:
+        attrs[trace_types.ATTR_TRANSCRIPTION_DELAY] = v
+    if (v := user_metrics.get("on_user_turn_completed_delay")) is not None:
+        attrs[trace_types.ATTR_ON_USER_TURN_COMPLETED_DELAY] = v
+    if attrs:
+        span.set_attributes(attrs)
+
+
 def _record_queue_wait(speech_handle: SpeechHandle) -> None:
     """Stamp how long the speech sat in the queue on its agent_turn span.
 
@@ -3205,6 +3219,7 @@ class AgentActivity(RecognitionHooks):
                 e2e_latency = started_speaking_at - _previous_user_metrics["stopped_speaking_at"]
                 assistant_metrics["e2e_latency"] = e2e_latency
                 current_span.set_attribute(trace_types.ATTR_E2E_LATENCY, e2e_latency)
+                _record_user_turn_stages(current_span, _previous_user_metrics)
 
         if forwarded_text and add_to_chat_ctx:
             msg = self._agent._chat_ctx.add_message(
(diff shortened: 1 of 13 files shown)


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1957,7 +1974,72 @@ def _ensure_user_turn_span(self, start_time: float | None = None) -> trace.Span:
...
+        span, self._eou_wait_span = self._eou_wait_span, None
+        started_at_ns, self._eou_wait_started_at_ns = self._eou_wait_started_at_ns, None
+        rearms, self._eou_wait_rearms = self._eou_wait_rearms, 0
+        if span is None or not span.is_recording():
+            return
+
+        ended_at_ns = int((end_time if end_time is not None else time.time()) * 1_000_000_000)
+        if started_at_ns is not None:
+            # resumed speech can carry a VAD timestamp from before the anchor; never negative
+            ended_at_ns = max(ended_at_ns, started_at_ns)
```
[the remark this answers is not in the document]
davidzhao: Fixed in 938d9442c: the wait now ends no earlier than what it contains. A still-running eou_detection is closed first with a superseded event and the wait ends at the same instant; a re-arm recorded after the requested end bounds it too. The real resume time is kept as a user_resumed event inside the bar, and lk.eou.wait_duration still equals the span length exactly. Covered by test_resumed_speech_during_detection_keeps_the_child_inside.

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1957,7 +1974,72 @@ def _ensure_user_turn_span(self, start_time: float | None = None) -> trace.Span:
...
+        started_at_ns = int(started_at * 1_000_000_000)
+        with tracer.use_span(user_turn_span):
+            span = tracer.start_span(
+                "eou_wait",
+                start_time=started_at_ns,
+                attributes={trace_types.ATTR_EOU_SOURCE: trigger},
+            )
+        self._eou_wait_span = span
+        self._eou_wait_started_at_ns = started_at_ns
+        self._eou_wait_rearms = 0
```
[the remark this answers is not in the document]
davidzhao: Fixed in 938d9442c: switching turn detection to or from manual now drops the pending wait (outcome dropped) alongside cancelling the endpointing task; the user_turn span stays open for the speech that follows. Covered by test_turn_mode_change_drops_an_open_wait.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2601,15 +2612,24 @@ async def _user_turn_completed_task(
...
+        ) as hook_span:
+            try:
+                await self._agent.on_user_turn_completed(
+                    temp_mutable_chat_ctx, new_message=user_message
+                )
+            except StopResponse:
+                hook_span.add_event("stop_response")
+                return  # ignore this turn
+            except Exception as e:
+                trace_utils.record_exception(hook_span, e)
```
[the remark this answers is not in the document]
davidzhao: Fixed in 938d9442c: record_exception for the hook now takes redacted=session._redaction_enabled or the job flag, so record={"redaction": True} on the session alone redacts the message. Covered by test_hook_exception_honours_session_redaction.

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2601,15 +2612,30 @@ async def _user_turn_completed_task(
...
+                return  # ignore this turn
+            except Exception as e:
+                # the hook is user code and its message can quote the transcript; honour a
+                # redaction switched on for this session alone as well as the job's
+                trace_utils.record_exception(
+                    hook_span,
+                    e,
+                    redacted=self._session._redaction_enabled or trace_utils.redaction_enabled(),
+                )
+                logger.exception("error occurred during on_user_turn_completed")
```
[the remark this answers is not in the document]
davidzhao: exception does not contain PII

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -214,6 +214,28 @@ class _PausedSpeechInfo:
 
 
 # NOTE: AgentActivity isn't exposed to the public API
```
chenghao-mou, line 216: nit: this comment should be moved

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -214,6 +214,28 @@ class _PausedSpeechInfo:
...
+def _end_user_turn(info: _EndOfTurnInfo) -> None:
```
chenghao-mou, line 217: nit: maybe `_end_user_turn_span`

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2601,16 +2633,36 @@ async def _user_turn_completed_task(
...
+                # redaction switched on for this session alone as well as the job's
+                trace_utils.record_exception(
+                    hook_span,
+                    e,
+                    redacted=self._session._redaction_enabled or trace_utils.redaction_enabled(),
+                )
+                logger.exception("error occurred during on_user_turn_completed")
+                return
 
+        _end_user_turn(info)
```
chenghao-mou, line 2665: do we still need this if `_end_user_turn(info)` is called in finally in `_user_turn_completed_task`?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1957,7 +1992,108 @@ def _ensure_user_turn_span(self, start_time: float | None = None) -> trace.Span:
...
+            span.add_event(
+                "rearmed",
+                {trace_types.ATTR_EOU_SOURCE: trigger},
+                timestamp=self._eou_wait_floor_ns,
+            )
+            span.set_attribute(trace_types.ATTR_EOU_SOURCE, trigger)
+            return span
+
+        now = time.time()
+        started_at = min(last_speaking_time, now) if last_speaking_time is not None else now
```
chenghao-mou, line 2029: should we also include the user_turn_span start time here?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1520,9 +1543,14 @@ async def _bounce_eou_task(
...
+                    # ended by hand: when the user resumes mid-inference the wait closes
+                    # first (see _end_eou_wait_span) and the child must not outlive it
                     with (
-                        tracer.use_span(user_turn_span),
-                        tracer.start_as_current_span("eou_detection") as eou_detection_span,
+                        tracer.use_span(eou_wait_span),
+                        tracer.start_as_current_span(
+                            "eou_detection", end_on_exit=False
+                        ) as eou_detection_span,
+                        self._track_eou_detection_span(eou_detection_span),
```
chenghao-mou, line 1553: I am not sure this would be accurate for eot detection. Because we trigger the prediction 200ms silence after the VAD speech (before EOS), the prediction could be from cache or already in flight. So the span duration != detection delay `prediction_event.detection_delay`

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1957,7 +1992,108 @@ def _ensure_user_turn_span(self, start_time: float | None = None) -> trace.Span:
...
+        if started_at_ns is not None:
+            # resumed speech can carry a VAD timestamp from before the anchor; never negative
+            ended_at_ns = max(ended_at_ns, started_at_ns)
+
+        detection, self._eou_detection_span = self._eou_detection_span, None
+        if detection is not None and detection.is_recording():
+            # the user resumed while the detector was still running: the inference is moot,
+            # and the wait ends when it does so the child stays inside the parent
+            self._eou_wait_floor_ns = time.time_ns()
+            detection.add_event("superseded", {trace_types.ATTR_EOU_OUTCOME: outcome})
```
chenghao-mou, line 2076: nit: we should reuse the timestamp captured above for the event timestamp `add_event(.., timestamp=...)`.

livekit-agents/livekit/agents/voice/agent_activity.py
chenghao-mou, line 3873: realtime only assigns `agent_turn_context ` needed for the queue wait in  `_realtime_generation_task`. The above `_record_queue_wait` could be attached to a wrong span.

livekit-agents/livekit/agents/voice/agent_activity.py
davidzhao: nope, you are right this is redundant.

livekit-agents/livekit/agents/voice/audio_recognition.py
davidzhao: I'm not sure what you mean. are you saying we do not have the accurate start time as measured here? I don't think this has changed with my PR though

livekit-agents/livekit/agents/voice/audio_recognition.py
davidzhao: why would we want to relate it to user_turn _span? I think the logic is we start eou_wait from the last the user has spoken

chenghao-mou: `user_turn` can be created lazily from a final STT transcript when no VAD/`START_OF_SPEECH` event created it for the current turn:

- STT updates `_last_speaking_time`.
- STT triggers the EOU task.
- `_ensure_user_turn_span()` uses `time.time()`.
- `eou_wait` is backdated to `_last_speaking_time`.

The child can therefore start before its parent. Should we use min of `(speech_start_time, last_speaking_time, now)` when creating the user turn span instead?

livekit-agents/livekit/agents/voice/audio_recognition.py
chenghao-mou: yeah, nothing's changed. The issue is more about we are reporting a start time (from the new detection span) that doesn't match our actual detection delay (and therefore actual start time).

jmcclanahan (approved):

jmcclanahan (approved):
LGTM from the observability side

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -2601,15 +2633,32 @@ async def _user_turn_completed_task(
...
+                trace_utils.record_exception(
+                    hook_span,
+                    e,
+                    redacted=self._session._redaction_enabled or trace_utils.redaction_enabled(),
+                )
+                logger.exception("error occurred during on_user_turn_completed")
+                return
 
         on_user_turn_completed_delay = time.perf_counter() - start_time
         metrics_report["on_user_turn_completed_delay"] = on_user_turn_completed_delay
```
jmcclanahan, line 2690: The hook delay is computed here but only reaches the metrics event. Could it also land on the agent_turn span, alongside `lk.end_of_turn_delay` and `lk.transcription_delay` copied from `user_metrics`? The reply task already has those in hand where it sets `lk.e2e_latency`. With all stages on one span, "E2E minus stages" per turn is a single-span read downstream. With the way it's currently set up it needs a five-span join by trace and time order.

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1758,6 +1789,9 @@ async def _bounce_eou_task(
                     self._turn_detector_prediction_fut = None
                     self._turn_detector_flushed = True
 
+            elif eou_wait_span.is_recording():
+                eou_wait_span.add_event("not_committed")
```
jmcclanahan: Could this also bump a counter that lands as `lk.eou.not_committed_count` next to `lk.eou.rearm_count`? Rate of rejected turn decisions is a fleet-level tuning signal, and events need the events array scanned, while an attribute is a cheap read.

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -1369,6 +1384,7 @@ async def _on_vad_event(self, ev: vad.VADEvent) -> None:
                 self._vad_speech_started = True
 
             self._cancel_transcription_timeout()
+            self._end_eou_wait_span("user_resumed", end_time=speech_start_time)
```
jmcclanahan, line 1390: Each resume ends one `eou_wait` and the next silence opens another, so one user turn can hold several. Could recognition count the `user_resumed` ends and write `lk.eou.resume_count` onto `user_turn` when it commits? Same applies to the STT path at line 1366.

jmcclanahan (approved):
Couple QoL things for downstream, otherwise lgtm

livekit-agents/livekit/agents/voice/agent_activity.py
davidzhao: done

livekit-agents/livekit/agents/voice/audio_recognition.py
davidzhao: done

livekit-agents/livekit/agents/voice/audio_recognition.py
davidzhao: yup, done in https://github.com/livekit/agents/pull/7130/changes/a8af38e827fe7fa00dc2f1a5519d753ffc18ffe3#diff-6ebfa91a7614432d570af64493a27554b30cc7cf93875abab1ae52aa354397caR1535

chenghao-mou (approved):

PR #7131, opened 2026-09-05
feat(telemetry): trace dispatch, startup, and shutdown

davidzhao:
## What

Before the first turn and after the last one the trace had a single span, `job_entrypoint`, so "the agent took 4 s to say hello" and "the job hung on shutdown" could not be split into their steps. This adds the dispatch timeline, the startup spans, the shutdown spans, and the join keys for lining the agent trace up with server, SIP, and client events later.

## Dispatch timeline

- The worker stamps `received_at`, `accepted_at`, `assigned_at` on `RunningJobInfo`; the process pool stamps `launched_at`. `StartJobRequest` carries the four as 64-bit doubles (a 32-bit float only resolves ~128 s at unix-epoch magnitude, which made every latency read 0).
- `job_entrypoint` is **back-dated to `received_at`** and gets one timestamped event per stage (`job_received`, `job_accepted`, `job_assigned`, `process_assigned`, `entrypoint_started`, `job_started_on_server` from `JobState.started_at`). The attributes are the seconds between adjacent stages, so the chain reads off the span without timestamp arithmetic: `lk.job.accept_latency` (request → accept), `assignment_latency` (accept → server assignment), `launch_latency` (assignment → a pool process takes the job), `entrypoint_latency` (process takes the job → user entrypoint runs), and their sum `dispatch_latency`. Raw `*_at` unix timestamps are not set as attributes. Unknown stages (simulation, console, resumed jobs) are skipped, not guessed.
- `agent_session` is a child of `job_entrypoint`, so the job's trace tells the whole story from dispatch to teardown: `job_entrypoint` → (`room_connect`, `agent_session`, …, `job_shutdown`). Nothing is replayed or re-parented for the session's benefit; a viewer keyed to `agent_session` zooms out to the job for the rest.
- **The trace pipeline is up before the job's first span.** Until now the cloud tracer provider was created on the first `init_recording` (inside `session.start()`), so in a process's first job every earlier span was a non-recording stub: `job_entrypoint` and `room_connect` were lost, and `agent_session` had no valid parent and started its own trace (that is why it rendered as a root). `JobContext._prepare_telemetry` now creates the provider and the gated exporter when the job starts, registering nothing; the gate holds the job's spans (oldest first, bounded) until `init_recording` decides, then uploads or drops them, and drops them at cleanup if the job never registers.
- **`job_entrypoint` spans the whole job.** It is created before the user entrypoint runs and ended after `job_shutdown`, just before the telemetry release flushes, so the root covers dispatch, session and teardown. The user function returning is an `entrypoint_returned` event on it (most entrypoints return right after `session.start()`); a raising entrypoint is recorded redaction-aware, a cancelled one as `entrypoint_cancelled`.
- Join keys on `job_entrypoint` and `agent_session`: `lk.room_sid`, `lk.dispatch_id`, `lk.job.worker_id`, `lk.job.agent_id`. A linked SIP participant's `sip.*` attributes are copied onto `agent_session` under `lk.sip.*` (call id, trunk id and number, rule id, status, headers); only the end user's `sip.phoneNumber` is PII and lands as `lk.pii.sip.phoneNumber`.

## Startup

- `room_connect` around `JobContext.connect`: room ids, `auto_subscribe`, e2ee, local participant, remote participant count; error status on failure. Called before `session.start()` it lands under `job_entrypoint`; called while the session is starting it nests under `session_start`.
- `session_start` groups `AgentSession.start()`: `start_agent_activity`, `setup_toolsets`, RoomIO's `wait_for_participant`, `publish_audio_output`, and the room connect. Only startup work is parented there via explicit contexts; `user_turn` / `agent_turn` stay directly under `agent_session` (tested).
- `wait_for_audio_track` on the RoomIO audio input: linked participant → first frame, with `track_subscribed`, `pre_connect_audio`, and `first_frame` events and `lk.first_frame_delay`.
- Events on `agent_session`: `participant_linked`, `participant_disconnected` (reason), `connection_state_changed`, `agent_state_changed`, `user_state_changed`.

## Shutdown

- `session_close` wraps the AgentSession close with `lk.close_reason` (enum) and `lk.close.drain`; `drain_agent_activity` and `on_exit` nest inside it.
- `job_shutdown` wraps the job's shutdown sequence with `lk.shutdown.reason` and children `on_session_end`, `session_end_upload`, `room_disconnect`, and one `shutdown_callback` per **user** callback named by `lk.callback.name` (the session's own close hook runs without a span: it is already `session_close`). It is a child of `job_entrypoint`, like the session.
- `user_turn` is pinned to the session root explicitly. It can be created from any task (a late STT final arriving during `session_close`, for one) and used to nest under whatever was current there.
- **Startup spans are never current.** `session_start`, `room_connect`, `wait_for_participant`, `publish_audio_output` and `setup_toolsets` wrap code that spawns the tasks living for the whole session (the room's event tasks, the published track, MCP servers). A current span there is inherited by those tasks and becomes the accidental parent of unrelated spans they emit later: in a real run `user_speaking` from the room's active-speaker callback landed under `room_connect`. They are now created with `tracer.detached_span(name, context=...)`: parented explicitly (under `session_start` while it runs), never attached to the context.
- One warm-up module, `livekit.agents.ipc._preload`, holds the framework's lazy one-time work: the livekit-rtc native library `dlopen` (otherwise the first FFI call, the `AudioProcessingModule` at session start, took 100–350 ms), the openai SDK's `resources` tree (otherwise the inference LLM prewarm's first client attribute access took ~300 ms), one httpx client (otherwise the first inference client construction took ~40 ms of GIL-held CPU), and the local VAD and end-of-turn models (previously `inference._warmup`). Under `forkserver` the worker lists it in `set_forkserver_preload` ahead of the GC freeze, so it runs once and job processes inherit the result copy-on-write; under `spawn` each job process imports it while warming up, before the user prewarm. The job process always imports it and the import is a no-op when the forkserver already did, so there is no start-method branching. Each of these showed up as `event_loop_blocked` at session start that no user code caused.
- Job cleanup (tempdir removal, the per-job telemetry release that joins exporter flush threads) moved off the event loop into `asyncio.to_thread`, so shutdown no longer stalls the loop the monitor watches, and THREAD-executor jobs sharing the process are not held up.

```
job_entrypoint  (back-dated to job_received; events per stage)
├─ agent_session
│  ├─ session_start
│  │  ├─ room_connect
│  │  ├─ wait_for_participant
│  │  ├─ wait_for_audio_track
│  │  ├─ publish_audio_output
│  │  └─ start_agent_activity
│  │     ├─ setup_toolsets
│  │     └─ on_enter
│  ├─ user_turn / agent_turn ...
│  └─ session_close
│     ├─ drain_agent_activity
│     └─ on_exit
└─ job_shutdown
   ├─ on_session_end
   ├─ session_end_upload
   ├─ room_disconnect
   └─ shutdown_callback (×N)
```

## Tests

`tests/test_startup_spans.py`: `StartJobRequest` round trip, dispatch events and latencies, server timestamp units, `room_connect` under the job span, success and failure, shutdown callback naming, and a full fake session asserting the `session_start` / `session_close` nesting, that turns are not re-parented, and the state events. `tests/test_recording.py`: the gate holds, flushes and drops spans of undecided jobs; the prepared pipeline holds early spans until the job registers.

Not covered by unit tests (needs a real room): `wait_for_audio_track` and `publish_audio_output`.

Stacked on #7130.


livekit-agents/livekit/agents/job.py
```diff
@@ -347,7 +383,7 @@ async def _on_session_end(self) -> None:
             except Exception:
                 logger.exception("failed to upload the session report to LiveKit Cloud")
 
-    def _on_cleanup(self) -> None:
+    async def _on_cleanup(self) -> None:
```
davidzhao, line 386: this was blocking. moving it to a thread.

livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -393,24 +417,63 @@ def _job_export_state(
...
+        with self._lock:
+            self._open_jobs.add(job_id)
+
+    def job_registered(self, job_id: str, *, traces_enabled: bool) -> None:
+        """The job decided: upload what was held if it records traces, else drop it."""
+        with self._lock:
+            self._open_jobs.discard(job_id)
+            held = self._pending.pop(job_id, [])
+        if held and traces_enabled:
+            self._inner.export([pii.restore_pii(s) for s in held])
```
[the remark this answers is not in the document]
davidzhao: Valid, fixed in 0414a36e5. job_registered runs inside init_recording on the event loop, and OTLPSpanExporter.export blocks on the network with retries. The gate now hands the held spans back to the BatchSpanProcessor (on_end), so they are exported from its own thread once the job is registered, in their original order. Nothing is exported inline anymore.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1224,6 +1256,18 @@ async def _aclose_impl(
...
+            close_span = tracer.start_span(
+                "session_close",
+                attributes={
+                    trace_types.ATTR_CLOSE_REASON: reason.value,
+                    trace_types.ATTR_CLOSE_DRAIN: drain,
+                },
+            )
+            if error is not None:
+                close_span.set_attribute(trace_types.ATTR_EXCEPTION_TYPE, error.type)
+            otel_context.attach(trace.set_span_in_context(close_span))
```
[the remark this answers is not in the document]
davidzhao: Valid, fixed in 0414a36e5. Both contexts (root and session_close) are now attached with tokens and detached: session_close is detached right after the span ends, so the close event handlers and room io teardown run under agent_session, and the root context is restored on the way out of aclose(), so a caller awaiting it in its own task (entrypoint, shutdown callback) gets its previous context back. The update_agent span in the coverage layer got the same treatment.

livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
```diff
@@ -382,6 +399,30 @@ def _on_entry_done(t: asyncio.Task[Any]) -> None:
 
         shutdown_info = await self._shutdown_fut
 
+        # a child of job_entrypoint, like the session was: the job's trace tells the whole
+        # story from dispatch to teardown, and a viewer keyed to agent_session zooms out to it
+        with tracer.start_as_current_span(
+            "job_shutdown",
+            context=self._entrypoint_span_context,
+            attributes={
+                trace_types.ATTR_SHUTDOWN_REASON: shutdown_info.reason,
```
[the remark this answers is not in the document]
davidzhao: Keeping this as is. The shutdown reason is an operator-facing string set by the agent code (or by the framework), not user content, so it is intentionally untagged. This matches the position across the stack: identities, task names, and the shutdown reason are not PII; only RPC payloads/responses and sip.phoneNumber are.

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -1925,6 +1973,23 @@ def _on_aec_warmup_expired(self) -> None:
...
+                # trace; only the end user's number is personal data
+                span.set_attributes(
+                    {
+                        (
+                            trace_types.ATTR_SIP_PHONE_NUMBER
+                            if key == "sip.phoneNumber"
+                            else trace_types.ATTR_SIP_PREFIX + key.removeprefix("sip.")
+                        ): value
+                        for key, value in participant.attributes.items()
+                        if key.startswith("sip.")
```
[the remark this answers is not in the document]
davidzhao: Keeping this as is. The SIP attributes are copied under lk.sip.* deliberately with only sip.phoneNumber tagged as PII; trunk, rule, and call ids are routing metadata that operators need for correlation. Same position as on the RPC identity comments in the stack.

livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
```diff
@@ -382,6 +402,30 @@ def _on_entry_done(t: asyncio.Task[Any]) -> None:
...
+                trace_types.ATTR_SHUTDOWN_USER_INITIATED: shutdown_info.user_initiated,
+            },
+        ):
+            await self._shutdown_job(job_entry_task, shutdown_info)
+        # ended before the telemetry release below flushes: the root goes out with the job
+        job_span.end()
```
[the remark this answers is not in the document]
davidzhao: Valid, fixed in 1a9f218c8. The cleanup was not under a finally on main either, but this PR made the root span depend on it, so a shutdown failure (a session or room close raising) would have lost the trace's root as well as leaking the temp dir and, on thread workers, the per-job telemetry state. The shutdown span now runs inside a try whose finally ends the job span, cancels pending tasks and runs _on_cleanup.

livekit-agents/livekit/agents/worker.py
```diff
@@ -1352,6 +1353,7 @@ async def _answer_availability(self, msg: agent.AvailabilityRequest) -> None:
         """Ask the user if they want to accept this job and forward the answer to the server.
         If we get the job assigned, we start a new process."""
 
+        received_at = time.time()
```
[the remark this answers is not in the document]
davidzhao: Valid, fixed in 1a9f218c8. received_at is stamped in _handle_availability when the message arrives and passed into the task, so the loop delay before the task runs counts toward dispatch latency as it should.

livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
```diff
@@ -427,27 +479,117 @@ def _on_entry_done(t: asyncio.Task[Any]) -> None:
...
+        ) -> None:
+            if _is_framework_callback(callback):
+                # the session's own close hook, already covered by session_close; a span here
+                # would read as a second user callback
+                await callback(shutdown_info.reason)
+                return
+            # a hung callback here is why jobs hit the supervisor's shutdown deadline
+            with tracer.start_as_current_span(
+                "shutdown_callback",
+                attributes={trace_types.ATTR_CALLBACK_NAME: _callback_name(callback)},
```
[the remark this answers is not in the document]
davidzhao: Keeping this as is. Callback names are __qualname__ values of the developer's own functions, code identifiers rather than user or conversation content, the same category as task names, source paths and stack frames that we deliberately leave untagged across the stack. Only RPC payloads/responses, transcripts and sip.phoneNumber carry the PII marker.

livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -393,24 +417,83 @@ def _job_export_state(
...
+        with self._lock:
+            for s in spans:
+                state = _job_export_state(self._export_jobs, s.attributes)
+                if state is not None:
+                    if state.traces_enabled:
+                        # PII filtered for third-party exporters is put back here: what
+                        # LiveKit Cloud may receive is the project's setting, applied at its
+                        # collector. restore_pii is a no-op once that setting mandates
+                        # redaction.
+                        exportable.append(pii.restore_pii(s))
```
chenghao-mou, line 491: found by Codex:

```
ctx.connect() finishes
  → redaction processor runs while redaction=False
  → finished span is buffered

session.start() registers redaction=True
  → buffered span is released
  → Cloud exporter receives the unchanged span
```

if customer code calls ctx.connect before session.start, the redaction setting might not be effective during connect.

livekit-agents/livekit/agents/telemetry/traces.py
chenghao-mou: Found by Codex:

> If the entrypoint calls `set_tracer_provider()`, `session.start()` shuts down the old Cloud processor while `job_entrypoint` is still open. When the job ends, that span reaches the stopped processor and gets dropped, leaving `agent_session` with a missing parent. Could we preserve the Cloud processor and pending buffer across provider changes, or keep the old pipeline alive until its spans finish?

livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
```diff
@@ -382,6 +402,34 @@ def _on_entry_done(t: asyncio.Task[Any]) -> None:
 
         shutdown_info = await self._shutdown_fut
```
chenghao-mou: should we move this into the try block so a cancellation doesn't skip clean up?

livekit-agents/livekit/agents/telemetry/traces.py
davidzhao: fixed

livekit-agents/livekit/agents/ipc/job_proc_lazy_main.py
```diff
@@ -380,8 +396,32 @@ def _on_entry_done(t: asyncio.Task[Any]) -> None:
...
+                attributes={
+                    trace_types.ATTR_SHUTDOWN_REASON: shutdown_info.reason,
+                    trace_types.ATTR_SHUTDOWN_USER_INITIATED: shutdown_info.user_initiated,
+                },
+            ):
+                await self._shutdown_job(job_entry_task, shutdown_info)
+        finally:
+            # whatever happened above (a cancel while waiting, a close that raised), the job
+            # ends: root span, temp dir, per-job telemetry state
+            job_span.end()
```
chenghao-mou, line 413: from Codex:

> If this task is cancelled while awaiting `_shutdown_fut`, `_shutdown_job()` never runs. `job_entry_task` is separate and is not in `_pending_tasks`, so it can keep running while this block ends its root span and removes its resources. Can we cancel and await it before ending the span and starting cleanup?

chenghao-mou (approved):
lgtm, one issue from Codex

livekit-agents/livekit/agents/voice/agent_session.py
```diff
@@ -969,6 +977,10 @@ async def start(
             # inference (`chat`) and tool spans (`execute_tool`) nest underneath it
             gen_ai_telemetry.set_workflow_attributes(self._session_span, name="agent_session")
 
+            # startup as one bar: room connect, participant wait, model prewarm, on_enter
+            session_start_span = tracer.start_span("session_start")
```
jmcclanahan, line 983: What do you think about a `lk.session.time_to_first_speech` on `agent_session`, measured from `participant_linked`? I think a "greeting time" KPI would be pretty useful to put in the dashboard

jmcclanahan (approved):
LGTM, nothing blocking

livekit-agents/livekit/agents/voice/agent_session.py
davidzhao: let's think about this.. I think this is dependent if the agent is suppose to speak first or if the user is. how do we make this metric meaningful?

PR #7134, opened 2026-09-05
feat(telemetry): trace RPC calls made and handled

davidzhao:
## What

Neither user-registered RPC methods nor the avatar datastream RPCs appeared in a trace. This installs a tracing interceptor on the job's local participant and turns each RPC into a span.

Depends on the `RpcInterceptor` hook in `livekit-rtc` (livekit/python-sdks#806, released in 1.1.18); the stack pins `livekit==1.1.18` in #7128. `install` still degrades to a no-op with one debug log on an older SDK.

## Spans

| Span | Kind | Parent |
|---|---|---|
| `rpc_call` | CLIENT | the current span, so an RPC issued from a tool nests under `function_tool` |
| `rpc_handler` | SERVER | the primary session's root span, so it lands on the session timeline |

Attributes: `rpc.method`, `lk.rpc.request_id`, `lk.rpc.caller_identity`, `lk.rpc.destination_identity`, `lk.rpc.payload_size`, `lk.rpc.response_size`, `lk.rpc.response_timeout`, `lk.rpc.error_code`, `lk.rpc.handler_registered` (false when a client called a method the agent never registered). Request and response bodies are under PII keys, `lk.pii.rpc.payload` and `lk.pii.rpc.response`, truncated to 1 KiB. `RpcError` and handler exceptions set error status.

## Where

- `telemetry/rpc.py`: `TracingRpcInterceptor` (subclasses `rtc.RpcInterceptor` when present) and `install(local_participant)`.
- Installed from `JobContext.connect()` after the room connects, and from `RoomIO.start()` for a room connected elsewhere. One singleton interceptor; the SDK dedups registrations by identity, so repeated installs are no-ops.

## Tests

`tests/test_rpc_tracing.py` drives the interceptor with fake continuations: attributes and parenting for both directions, request and response truncation, error codes and status, the unregistered-method flag, handler exceptions, and `install` registering once or degrading. Passes against both the pinned SDK and the interceptor-capable SDK source tree.

Stacked on #7131.


livekit-agents/livekit/agents/voice/room_io/room_io.py
```diff
@@ -188,6 +188,8 @@ async def start(self, *, trace_context: otel_context.Context | None = None) -> N
         self._room.on("connection_state_changed", self._on_connection_state_changed)
         self._room.on("participant_disconnected", self._on_participant_disconnected)
         if self._room.isconnected():
+            # a room connected outside JobContext.connect() (which installs it itself)
+            rpc_tracing.install(self._room.local_participant)
```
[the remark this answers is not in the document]
davidzhao: Valid, fixed in d2fae93eb. The install moved into the connection-state handler, so it runs whenever the room is or becomes connected (already up at start, connected later by JobContext.connect() or by the user, reconnects). install is idempotent (one interceptor instance, deduped by identity in the SDK), so a room JobContext.connect() already instrumented is unaffected. Test added for a room that connects after RoomIO is created.

livekit-agents/livekit/agents/telemetry/rpc.py
```diff
@@ -0,0 +1,141 @@
...
+MAX_PAYLOAD_ATTR_LEN = 1024
+"""Request and response payloads longer than this many characters are truncated in span
+attributes."""
+
+_RpcInterceptorBase: type = getattr(rtc, "RpcInterceptor", object)
+_warned_unsupported = False
+
+
+def _truncate(payload: str) -> str:
+    return payload if len(payload) <= MAX_PAYLOAD_ATTR_LEN else payload[:MAX_PAYLOAD_ATTR_LEN]
```
chenghao-mou: `payload[:MAX_PAYLOAD_ATTR_LEN]` alone should be enough.

livekit-agents/livekit/agents/telemetry/rpc.py
```diff
@@ -0,0 +1,141 @@
...
+
+def _response_attributes(response: str | None) -> dict[str, Any]:
+    response = response or ""
+    attrs: dict[str, Any] = {trace_types.ATTR_RPC_RESPONSE_SIZE: len(response.encode("utf-8"))}
+    if response:
+        attrs[trace_types.ATTR_RPC_RESPONSE] = _truncate(response)
+    return attrs
+
+
+def _handler_parent_context() -> otel_context.Context | None:
```
chenghao-mou: nit: we should reuse session_context.session_root_context

livekit-agents/livekit/agents/telemetry/rpc.py
chenghao-mou: also, should we use byte length check here?

livekit-agents/livekit/agents/telemetry/rpc.py
```diff
@@ -0,0 +1,141 @@
...
+        }
+        with tracer.start_as_current_span(
+            "rpc_handler",
+            context=_handler_parent_context(),
+            kind=trace.SpanKind.SERVER,
+            attributes=attributes,
+        ) as span:
+            try:
+                response = await next(invocation)
+            except rtc.RpcError as e:
```
chenghao-mou, line 122: found by Codex:

> The SDK enforces the response deadline by cancelling the incoming interceptor chain. This produces `CancelledError` here; the SDK converts it to `RpcError(RESPONSE_TIMEOUT)` only after this interceptor has unwound. Since `CancelledError` also bypasses OpenTelemetry’s `except Exception`, the handler span ends with `UNSET` status and no error code even though the caller receives `RESPONSE_TIMEOUT`.

livekit-agents/livekit/agents/telemetry/rpc.py
davidzhao: IMO truncation is a character limit since it's more for display purposes.. byte limiting may cut things off in the middle of a wide char.

chenghao-mou (approved):

jmcclanahan (approved):

PR #7138, opened 2026-09-05
feat: add livekit-plugins-thegrid

CastilloLuis:
## What

Adds `livekit-plugins-thegrid`, an LLM plugin for [The Grid](https://thegrid.ai/) — an OpenAI-compatible inference marketplace.

Modeled on `livekit-plugins-perplexity`. The plugin subclasses the OpenAI plugin's `LLM` and sets the base URL and `THEGRID_API_KEY` handling; there's no bespoke client.

Registration points, matching how perplexity/cerebras are wired:
- `livekit-plugins/livekit-plugins-thegrid/` (new package)
- `livekit-plugins-thegrid = { workspace = true }` in the root `pyproject.toml`
- `thegrid = ["livekit-plugins-thegrid>=1.8.0"]` extra in `livekit-agents/pyproject.toml`
- `tests/test_plugin_thegrid.py`

Left `CHANGELOG.md` and the version manifests alone per CONTRIBUTING.

## One deliberate difference from the perplexity template

Perplexity passes `_strict_tool_schema=False`. I tested The Grid with `strict: true` tool schemas (and `additionalProperties: false`) and it returns well-formed tool calls, so I did **not** carry that flag over — the default applies.

## Verification

Run against the live API, not mocked:

```
$ pytest tests/test_plugin_thegrid.py --plugin thegrid -q
...                                                                      [100%]
3 passed in 6.34s
```

That covers chat, a tool call round-trip on `agent-standard`, and streaming. Also checked `ruff format --check` and `ruff check` clean on the new files, that the missing-key path raises, and that `GET /v1/models` returns 200 directly with no redirect (relevant because `_prewarm_impl()` calls it with a 5s timeout).

## Note on models

The Grid addresses capability tiers — `text-standard`, `code-prime`, `agent-max` each route to a current model for that tier, so an agent keeps working when the underlying model is swapped. `models.py` lists the tiers plus the lab-pinned `*-latest` ids; `model` still accepts any `str`, and `GET /v1/models` is authoritative.

Disclosure: I work on The Grid. Happy to adjust anything to house style.


CastilloLuis:
Good catch, and your suggestion was right. Applied, with one extra step it needed.

Guarding the raise on `client is None` alone was not sufficient: this class still forwarded `api_key=""` to the OpenAI plugin, and `llm.py:152` rejects an explicitly-given empty string with "OpenAI API key is required". So an injected client failed one line further down instead.

It now resolves the key into a local and passes `resolved_key or NOT_GIVEN`, which leaves the OpenAI plugin free to use the supplied client's own authentication.

Verified all three paths:

```
supplied client without key: OK, model = text-standard
no client, no key: still raises ValueError OK
env key path still works: text-standard
```

`pytest tests/test_plugin_thegrid.py --plugin thegrid` still passes 3/3 against the live API, and `ruff check` / `ruff format --check` are clean on 0.15.22.

Separately, I have added the missing `uv.lock` entry for the new workspace member in this branch, which is what was failing the ruff job (it died at `uv sync` before lint ran).

CastilloLuis:
Merged `main` in; the only conflict was `uv.lock`.

Resolved by taking upstream's lockfile wholesale and re-running `uv lock`, so the delta against main is just the new workspace member rather than a hand-merged lockfile:

```
Resolved 328 packages
Added livekit-plugins-thegrid (dynamic)
```

Verified after the merge:

- `uv run ruff check` on the plugin and its test: all checks passed
- `uv run ruff format --check`: 6 files already formatted
- `uv run pytest tests/test_plugin_thegrid.py`: 3 passed, run against the live API with a real `THEGRID_API_KEY`

Worth flagging for whoever reviews: those three tests are live integration tests, so they need a `THEGRID_API_KEY` secret in CI to run there. Without one they fail at construction with `THEGRID_API_KEY is required`. Happy to mark them `skipif` on a missing key if you would rather they no-op in CI, or to supply a key if you have a way to take one.

PR #7143, opened 2026-09-06
feat(telemetry): one agent_turn span per speech handle

davidzhao:
## What

A reply that calls a tool runs two generations (LLM steps) in two tasks. They were two `agent_turn` spans linked only by `lk.parent_generation_id`, so one response rendered as two turns. A speech handle is now exactly one `agent_turn`.

## How

- The speech handle owns the span. The first reply task (pipeline, realtime, or `say`) opens `agent_turn` under `agent_session`; the follow-up generation after a tool call runs in a new task but continues the open span instead of opening a second one. It ends with the speech in `SpeechHandle._mark_done`, whichever step that was on, recording the speech's error (redaction-aware) and the `gen_ai.invoke_agent.duration` metric for the whole turn.
- Each generation is a `generation` event on the span with its `lk.generation_id` and `lk.parent_generation_id`. On the span itself `lk.generation_id` names the latest generation and the new `lk.generation_count` how many there were (more than one means tools ran before the final reply). `lk.speech_id` is set at creation.
- Every step's `llm_node`, `function_tool`, `tts_node`, `realtime_inference` and `agent_speaking` nest under the one turn.
- A preemptive generation discarded for a successor answering the same user turn (a newer attempt, or the real reply after `on_user_turn_completed` changed the transcript) hands its open `agent_turn` to that successor: the wasted generation stays visible under the one turn with a `preemptive_generation_discarded` event, and `lk.speech_id` follows the speech that answered. An attempt cancelled with no successor yet (the user resumed) still ends as its own turn. The queue-wait and interruption helpers tolerate a span already ended with the speech.

```
agent_turn  (lk.speech_id, lk.generation_count=2; events: generation ×2)
├─ llm_node          generation 1: the tool call
├─ function_tool
├─ llm_node          generation 2: the reply
├─ tts_node
└─ agent_speaking
```

## Tests

`tests/test_agent_turn_span.py`: a tool-calling reply through the fake session yields one `agent_turn` with two `generation` events, both `llm_node`s, the tool and the speech inside it, and the turn outlasting the last child; a plain reply is one generation. Existing coverage (`test_coverage_spans`, `test_eou_wait_span`, `test_agent_session`) passes unchanged.

Stacked on #7137.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -303,6 +303,16 @@ def _agent_turn(
         yield span
 
 
+def _continue_discarded_turn(discarded: SpeechHandle | None, successor: SpeechHandle) -> None:
+    """A preemptive generation discarded for ``successor`` (a newer attempt, or the real reply
+    after the transcript changed) hands its open ``agent_turn`` over, so one turn shows the
+    wasted generation and the one that answered. Module-level like ``_record_queue_wait``."""
+    if discarded is None or discarded is successor:
+        return
+    if (carry := discarded._take_agent_turn()) is not None:
+        successor._continue_agent_turn(carry, discarded=discarded)
+
+
 def _record_queue_wait(speech_handle: SpeechHandle) -> None:
     """Stamp how long the speech sat in the queue on its agent_turn span.
 
@@ -2558,6 +2568,13 @@ class AgentActivity(RecognitionHooks):
         ):
             return
 
+        # a newer attempt supersedes the current one; if one is created below it continues the
+        # discarded attempt's agent_turn (the cancelled speech only ends once the loop runs)
+        discarded = (
+            self._preemptive_generation.speech_handle
+            if self._preemptive_generation is not None
+            else None
+        )
         self._cancel_preemptive_generation()
 
         if (
@@ -2585,6 +2602,7 @@ class AgentActivity(RecognitionHooks):
             schedule_speech=False,
             input_details=InputDetails(modality="audio"),
         )
+        _continue_discarded_turn(discarded, speech_handle)
 
         self._preemptive_generation = _PreemptiveGeneration(
             speech_handle=speech_handle,
@@ -2822,6 +2840,7 @@ class AgentActivity(RecognitionHooks):
             return
 
         speech_handle: SpeechHandle | None = None
+        discarded_preemptive: SpeechHandle | None = None
         if preemptive := self._preemptive_generation:
             # make sure the on_user_turn_completed didn't change some request parameters
             # otherwise invalidate the preemptive generation
@@ -2851,6 +2870,7 @@ class AgentActivity(RecognitionHooks):
                     "preemptive generation invalidated after `on_user_turn_completed` because "
                     "the transcript, chat context, tools, or tool choice changed",
                 )
+                discarded_preemptive = preemptive.speech_handle
                 preemptive.speech_handle._cancel()
 
             self._preemptive_generation = None
@@ -2863,6 +2883,8 @@ class AgentActivity(RecognitionHooks):
                 chat_ctx=temp_mutable_chat_ctx,
                 input_details=InputDetails(modality="audio"),
             )
+            # the invalidated preemptive attempt answered this same turn: one agent_turn
+            _continue_discarded_turn(discarded_preemptive, speech_handle)
 
         if self._user_turn_completed_atask != asyncio.current_task():
             # If a new user turn has already started, interrupt this one since it's now outdated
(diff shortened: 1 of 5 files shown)


livekit-agents/livekit/agents/voice/speech_handle.py
```diff
@@ -345,6 +352,54 @@ def _mark_done(self, error: BaseException | None = None) -> None:
...
+
+    def _end_agent_turn(self, error: BaseException | None) -> None:
+        """Close the speech's ``agent_turn`` span: the speech is done, whatever step it was on."""
+        span, self._agent_turn_span = self._agent_turn_span, None
+        if span is None or not span.is_recording():
+            return
+        from ..telemetry import otel_metrics, utils as trace_utils
+
+        if isinstance(error, Exception):
+            trace_utils.record_exception(span, error)
```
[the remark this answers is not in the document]
davidzhao: Keeping this as is. A session's record={"redaction": True} is not session-only: session.start() passes its recording options to job_ctx.init_recording(), which sets the job-level flag from job.enable_redaction or options.redaction, the same expression the session stores. So when _end_agent_turn calls record_exception with the default, redaction_enabled() already reflects that session's setting. This is also exactly how the agent_turn span recorded errors before this PR (the tracer wrapper uses the same default) and how every other span in the pipeline does. The on_user_turn_completed hook additionally passes the session flag because its message can quote the transcript; that is belt-and-braces, not a different model. The only divergence is a session with no job context at all, where nothing uploads to Cloud anyway.

livekit-agents/livekit/agents/voice/agent_activity.py
chenghao-mou, line 2244: found by Codex:

> [P1] Continue server-generated realtime tool replies on the existing turn. this function creates a new `SpeechHandle` for automatic tool replies, so one tool-calling response still produces two `agent_turn` spans.

livekit-agents/livekit/agents/voice/speech_handle.py
```diff
@@ -345,6 +352,54 @@ def _mark_done(self, error: BaseException | None = None) -> None:
...
+        span.set_attribute(trace_types.ATTR_SPEECH_ID, self.id)
+        self._agent_turn_span = span
+        self._agent_turn_context = trace.set_span_in_context(span)
+        self._agent_turn_started_at = started_at
+        self._agent_turn_agent_name = agent_name
+
+    def _end_agent_turn(self, error: BaseException | None) -> None:
+        """Close the speech's ``agent_turn`` span: the speech is done, whatever step it was on."""
+        span, self._agent_turn_span = self._agent_turn_span, None
+        if span is None or not span.is_recording():
```
chenghao-mou, line 388: Q: do we want to gate otel metrics behind traces?

livekit-agents/livekit/agents/voice/agent_activity.py
```diff
@@ -231,11 +231,74 @@ def _record_interruption(speech_handle: SpeechHandle) -> None:
...
+
+
+def _continue_discarded_turn(discarded: SpeechHandle | None, successor: SpeechHandle) -> None:
+    """A preemptive generation discarded for ``successor`` (a newer attempt, or the real reply
+    after the transcript changed) hands its open ``agent_turn`` over, so one turn shows the
+    wasted generation and the one that answered. Module-level like ``_record_queue_wait``."""
+    if discarded is None or discarded is successor:
+        return
+    if (carry := discarded._take_agent_turn()) is not None:
+        successor._continue_agent_turn(carry, discarded=discarded)
```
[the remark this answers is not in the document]
chenghao-mou: This seems valid.

chenghao-mou (approved):
lgtm. One of the Devin comment seems valid when the task is only cancelled later:

```
`SpeechHandle._cancel()` does not immediately cancel the task. It:

1. Sets `_interrupt_fut`.
2. Leaves the task runnable.
3. Calls `task.cancel()` only after the five-second interruption timeout.
```

livekit-agents/livekit/agents/voice/speech_handle.py
davidzhao: the current decision is yes.. to be revisited in the future

davidzhao:
> lgtm. One of the Devin comment seems valid when the task is only cancelled later:
> 
> ```
> `SpeechHandle._cancel()` does not immediately cancel the task. It:
> 
> 1. Sets `_interrupt_fut`.
> 2. Leaves the task runnable.
> 3. Calls `task.cancel()` only after the five-second interruption timeout.
> ```

looked at resolving this and the code would seem a bit complicated.. so I think it's fine to leave as is.. lmk if you disagree

PR #7154, opened 2026-09-07
fix: close temporary speech stream adapters

chenghao-mou:
Automatic stream adaptation retained wrappers through their providers' metrics listener sets. Close owner-created wrappers when their turn, activity, or fallback lifecycle ends. Listener counts now return to baseline after success, failure, recovery, cancellation, and fallback shutdown. Configured providers remain reusable.

Addresses AGT-3451

<details>

**Model:** GPT-5.6

> please investigate AGT-3451
>
> yes, let's close that temp wrapper and create a draft PR
>
> we don't have to do this for stt fallback adapter?

</details>


longcw (approved):
nice catch! maybe also fix it for `tts.FallbackAdapter` and `Agent.default.stt_node` with `stt.StreamAdapter`

PR #7159, opened 2026-09-07
feat(soniox): map TTS character timestamps to aligned transcripts

### LINKED ISSUE #6727: soniox TTS: support aligned transcripts via return_timestamps (character-level timing)
**Feature request**

The Soniox TTS websocket API supports `return_timestamps: true`, returning character-level timing per audio frame (`timestamps.characters`, `character_start_times_seconds`, `character_end_times_seconds` — [docs](https://soniox.com/docs/api-reference/tts/websocket-api)). The plugin currently declares `TTSCapabilities(streaming=True)` with no `aligned_transcript`, and neither requests nor maps the timestamps.

This means sessions running `use_tts_aligned_transcript=True` fall back to the SDK's `TranscriptSynchronizer` speaking-rate estimation for Soniox TTS, so interrupted utterances record imprecise spoken text — while the provider actually exposes the same character-level data the ElevenLabs plugin already maps to aligned transcripts.

**Proposal:** request `return_timestamps: true` in the websocket config, concatenate the per-frame character arrays, and emit them through the SDK's aligned-transcript path, mirroring `livekit-plugins-elevenlabs`' character-alignment handling. Happy to contribute a PR if the approach sounds right.

acharan-tech-200037:
## Summary

Closes #6727.

The Soniox TTS websocket returns character-level audio timings when the start config asks for
`return_timestamps`, but the plugin neither requested them nor read them. It declared
`TTSCapabilities(streaming=True)` with no `aligned_transcript`, so sessions running
`use_tts_aligned_transcript=True` fell back to the SDK's constant speaking-rate estimate — and an
interrupted utterance recorded imprecise spoken text, even though the provider exposes the same
character-level data `livekit-plugins-elevenlabs` already maps.

This requests `return_timestamps`, folds the per-frame character arrays into whole words, and pushes
them through the emitter's aligned-transcript path.

Word durations in one 6.3s reply measured live ranged from 45ms (`'the'`) to 562ms
(`'interrupts.'`), with a 549ms pause between sentences — spread a constant rate cannot represent.

## Changes

- **`return_timestamps: bool = True`** on `TTS`, wired to both the start config and
  `TTSCapabilities.aligned_transcript`, so the plugin can never claim a capability it did not
  request. `return_timestamps=False` opts out.
- **`_recv_loop` maps each frame's `timestamps` before pushing that frame's audio**, since the
  emitter attaches pending timed words to the next frame it emits.
- **`_to_timed_words`** splits the character buffer with the SDK's `split_words`
  (`split_character=True`, so CJK and Thai split correctly) and holds the trailing word back until
  more characters arrive — a frame can end mid-word.
- **The emitted chunks tile the text.** Each runs to the start of the next word, so separators ride
  with the word before them. `perform_text_forwarding` builds the reply with `out.text += delta`, so
  bare word slices would store `"Hellothere,this"` as what the agent said. Timings still describe
  the word alone, not the separator.
- **Malformed frames** (missing or mismatched arrays) are dropped rather than mis-aligned.

### Where this plugin differs from ElevenLabs

ElevenLabs uses one websocket context per utterance. This plugin does not: a single
`SynthesizeStream` spans several Soniox `stream_id`s, rotating on LLM idle gaps and on stream age so
the server's per-stream timeout cannot kill synthesis mid-sentence. Three consequences:

1. **Timings restart per stream.** Soniox times characters from each stream's own first sample, so a
   straight port would send `TranscriptSynchronizer` (`dt = start_time - pushed_duration`) backwards
   partway through a reply. Each stream is offset by the audio already emitted when it opened.

2. **The offset cannot come from `pushed_duration()` alone.** That counts neither audio still queued
   in the emitter nor the tail frame it holds back, so it lags. A `_Timeline` shared across the
   segment tracks the furthest timestamp seen, and a rotated stream starts from whichever is
   further along. Measured live, this moved a rotation from 2.090s to 2.134s — exactly the end of
   the previous stream's last word.

3. **The separator between streams is lost.** The sentence tokenizer hands each stream its leading
   whitespace (`" Then, after a long pause..."`) and Soniox normalises that away before timing the
   text, so two streams concatenated to `"one sentence.Then"`. It is taken back from the text the
   stream was actually given, leaving scripts that do not space their sentences alone.

### Replay safety

A stream that fails transiently is replayed on a fresh `stream_id` so the reply is not cut short.
That is only safe while the stream has produced nothing, and `pushed_duration()` answers that
question late: audio handed to the emitter is invisible until it becomes frames, so a spent stream
could look replayable and have its audio — and its already-published words — spoken twice.

`_StreamData.produced_output` records the fact directly, set the moment audio or timed words are
handed over, and gates the replay. Words are published as soon as they are complete, so nothing is
withheld and nothing can be lost. A stream that will not be replayed flushes the word it was
holding back, since no more characters are coming for it — on failure, and in `_run`'s cleanup,
which is the path an interruption takes.

## Testing

`tests/test_plugin_soniox_tts.py` — 23 tests, 19 new:

- the real `_send_loop` sends `return_timestamps`, and omits it when disabled
- the real `_recv_loop` reassembles `"Hi the"` + `"re"` into `["Hi ", "there"]`, shifted onto the
  segment timeline, and marks audio-only frames as produced output
- **the chunks rebuild the reply exactly** — on one stream, across a rotation, and through
  `_recv_loop`
- timed words reach `frame.userdata[USERDATA_TIMED_TRANSCRIPT]` end-to-end through the real
  `SynthesizeStream`
- a reply split across two streams stays on one non-decreasing timeline
- a rotated stream starts from the timeline rather than from a lagging `pushed_duration()`
- a stream that published words, or merely pushed audio, is never replayed
- **an interruption through `SynthesizeStream.aclose()` keeps the last spoken word**
- mismatched timing arrays are ignored; `_to_timed_words` hold-back, flush and leading-separator
  behaviour

Every fix was mutation-tested: reverting it fails a test, and only the tests that should fail.

`ruff format`, `ruff check` and `mypy` are clean on the touched files.

### Verified against the live API

Checked end-to-end against a real Soniox key, including the rotation path:

```
stream 1 opened at offset 0.000s
stream 2 opened at offset 2.134s
    1.659 ->   2.134  'sentence.'     <- last word of stream 1
    2.416 ->   2.561  'Then, '        <- stream 2 continues the timeline
    ...
    6.607 ->   6.998  'third.'
rebuilt: 'First the agent says one sentence. Then, after a long pause, it says another one. And finally a third.'
```

The server restarts its own clock for stream 2, and `'Then, '` still lands after the previous
stream's last word rather than back near 0.28s — with the chunks rebuilding the spoken text exactly.

### DIFF
--- a/livekit-plugins/livekit-plugins-soniox/README.md
+++ b/livekit-plugins/livekit-plugins-soniox/README.md
@@ -54,6 +54,8 @@ session = AgentSession(
 
 The TTS supports real-time streaming from LLM - text chunks are tokenized and sent to Soniox as words are formed, enabling low-latency speech synthesis.
 
+Soniox also returns character-level audio timestamps, which the plugin forwards as an aligned transcript so an interrupted turn records the text that was actually spoken rather than a speaking-rate estimate. It is on by default; pass `return_timestamps=False` to turn it off.
+
 ## More information and reference
 
 Explore integration details and find comprehensive examples:
--- a/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
+++ b/livekit-plugins/livekit-plugins-soniox/livekit/plugins/soniox/tts.py
@@ -674,18 +674,15 @@ def _accumulate_timestamps(stream: _StreamData, timestamps: dict[str, Any]) -> N
     starts: list[float] | None = timestamps.get("character_start_times_seconds")
     ends: list[float] | None = timestamps.get("character_end_times_seconds")
     if not (chars and starts and ends and len(chars) == len(starts) == len(ends)):
-        # These characters cannot be trusted, so they are skipped - but the gap
-        # falls in the middle of a word. Publish what is already whole, drop the
-        # prefix left in the buffer, and wait for a word boundary: joining the
-        # two sides would invent a word ("bro" + "ps" reads as "brops"), and
-        # keeping either side alone would publish a fragment as a word of its
-        # own. The word the gap swallowed is lost either way.
+        # Only the timings are unusable here - these characters were still
+        # spoken, so none of the text is dropped. They are kept with the word
+        # they interrupted and handed over untimed once a boundary closes the
+        # region: joining the two sides of the gap would invent a word ("bro" +
+        # "ps" reads as "brops") and publishing either side alone would make a
+        # word of a fragment, but carrying the whole stretch untimed does
+        # neither. The synchronizer estimates across untimed text, so this
+        # stretch loses its precision and the reply keeps its words.
         logger.warning("Soniox TTS sent malformed timestamps for a frame, timing its text by rate")
-        # Only the timings are unusable - the characters were still spoken. Keep
-        # them, along with the word they were breaking, and hand them over
-        # untimed once the region closes: the synchronizer estimates across
-        # untimed text, so the reply stays whole and only this stretch of it
-        # loses precision.
         _emit_timed_words(stream)
         stream.pending_untimed += stream.char_text + "".join(chars or [])
         stream.char_text = ""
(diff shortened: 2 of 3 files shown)


acharan-tech-200037:
The tests / unit-tests failure was test_false_interruption_resume.py::test_teardown_does_not_resume_a_deferred_pause — unrelated to this PR (2573 passed, 1 failed). That test asserts on a 50ms wall-clock margin (asyncio.sleep(FALSE_INTERRUPTION_TIMEOUT + 0.05)) under pytest.mark.unit rather than virtual_time, so it is sensitive to runner load. This branch only touches livekit-plugins-soniox, and the failing module imports no TTS plugin. Reopening to re-run CI.

acharan-tech-200037:
@longcw @chenghao-mou  can you please review and tell me if any changes required

PR #7164, opened 2026-09-08
fix(google): deliver tool calls before combined turn completion

### LINKED ISSUE #7151: Google realtime: a response carrying both a terminal `server_content` and a `tool_call` kills the receive task
### What happens

In `livekit-plugins-google`, when a single `LiveServerMessage` carries both a terminal `server_content` (one with `turn_complete`) and a `tool_call`, the receive task raises and the session goes down.

The order is what does it. In the receive loop (`realtime_api.py`, v1.7.1):

```python
if response.server_content:
    self._handle_server_content(response.server_content)   # L1159
if response.tool_call:
    self._handle_tool_calls(response.tool_call)            # L1161
```

1. `_handle_server_content` sees `turn_complete` and calls `_mark_current_generation_done()` (L1382-1383).
2. That closes the function channel: `gen.function_ch.close()` (L1424).
3. `_handle_tool_calls` then does an unguarded `gen.function_ch.send_nowait(...)` (L1491) on the closed channel.

The exception propagates out of the receive task, so the whole realtime session is torn down rather than a single turn being lost.

### Why the obvious fix doesn't work

Swapping the two dispatch calls moves the problem instead of solving it. `_mark_current_generation_done` sets `_done = True` but does not clear `self._current_generation`, so `_handle_server_content` would still find a generation and write into the streams `_handle_tool_calls` had just closed through its own trailing `_mark_current_generation_done()` (L1498).

### Suggested direction

Hold the finalization back for the message that carries both, so the tool call is delivered on an open channel and the generation is finalized once afterwards. This also keeps the invariant #6785 introduced — that every tool call gets answered — which a message like this currently breaks by dropping the call on the floor (or, without a guard, by killing the session).

### Notes

We currently carry a local guard that returns early when `function_ch.closed`, which converts the crash into a dropped tool call plus a warning. We would rather drop the guard than maintain it — happy to send a PR if the direction above looks right to you.

noooooooookro:
Fixes #7151

## What changed

Google Realtime can send a single response containing both terminal server content and a tool call. The receive loop now defers generation finalization for that server content, keeping the function channel open until the tool call is delivered. Existing responses without a tool call retain their current completion behavior.

## Validation

- `python -m pytest tests/test_plugin_google_realtime.py -q` (19 passed)
- `ruff check livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py tests/test_plugin_google_realtime.py`
- `ruff format --check livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py tests/test_plugin_google_realtime.py`

### DIFF
--- a/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
+++ b/livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
@@ -1156,7 +1156,10 @@ class RealtimeSession(llm.RealtimeSession):
                             )
 
                     if response.server_content:
-                        self._handle_server_content(response.server_content)
+                        self._handle_server_content(
+                            response.server_content,
+                            defer_completion=response.tool_call is not None,
+                        )
                     if response.tool_call:
                         self._handle_tool_calls(response.tool_call)
                     if response.tool_call_cancellation:
@@ -1290,7 +1293,12 @@ class RealtimeSession(llm.RealtimeSession):
 
         self.emit("generation_created", generation_event)
 
-    def _handle_server_content(self, server_content: types.LiveServerContent) -> None:
+    def _handle_server_content(
+        self,
+        server_content: types.LiveServerContent,
+        *,
+        defer_completion: bool = False,
+    ) -> None:
         current_gen = self._current_generation
         if not current_gen:
             if self._rejected_tool_calls:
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py
```diff
@@ -1156,7 +1156,10 @@ async def _recv_task(self, session: AsyncSession) -> None:
                             )
 
                     if response.server_content:
-                        self._handle_server_content(response.server_content)
+                        self._handle_server_content(
+                            response.server_content,
+                            defer_completion=response.tool_call is not None,
```
longcw, line 1161: The behaviour is right. Maybe a cleaner shape: finalize here instead of in the two handlers. Both `_handle_server_content` and `_handle_tool_calls` are only called from this loop, so the loop can own the decision and the `defer_completion` flag goes away.

```python
if response.server_content:
    self._handle_server_content(response.server_content)
if response.tool_call:
    self._handle_tool_calls(response.tool_call)
if ((sc := response.server_content) and sc.turn_complete) or response.tool_call:
```

Then drop the `turn_complete` finalize at the end of `_handle_server_content` and the one at the end of `_handle_tool_calls`. Same order as now: content, then the calls, then close, in one iteration.

PR #7183, opened 2026-09-09
feat(elevenlabs): make realtime STT audio chunk duration configurable

### LINKED ISSUE #7182: Allow configuring ElevenLabs realtime STT audio chunk duration
### Feature Type

Would make my life easier

### Feature Description

The ElevenLabs realtime STT plugin currently repacks outgoing PCM into hardcoded 50 ms chunks. Expose a constructor option, `audio_chunk_duration_ms`, so applications can choose the tradeoff between WebSocket message frequency and buffering latency without maintaining a transport wrapper.

Proposed scope:
- Retain the existing 50 ms default.
- Accept positive integer milliseconds at construction time.
- Apply the option only to the `scribe_v2_realtime` send loop.
- Preserve partial-chunk flushing and send remaining audio before the commit message.

For example, `elevenlabs.STT(model="scribe_v2_realtime", audio_chunk_duration_ms=100)` would produce approximately 10 audio messages per second during continuous audio instead of 20, at the cost of up to 50 ms additional buffering.

This is a configuration capability, not a claim that larger chunks resolve provider queue errors. It does not pace reconnect backlogs.

A downstream application can patch the plugin or wrap its transport to combine audio messages, but that duplicates buffering behavior and requires preserving context and commit ordering. Changing the hardcoded default globally would also change latency for existing users.

### Additional Context

An implementation is available in my fork with a documented option and tests exercising the real send loop against a fake WebSocket. The tests cover multiple sample rates and durations, byte-for-byte audio preservation, partial tails, and commit ordering. Live provider validation has not been performed.

Opening this issue before the upstream PR as requested by CONTRIBUTING.md. Maintainer feedback on the API name and scope is welcome.

stevensukma:
ElevenLabs realtime STT currently sends audio in hardcoded 50 ms chunks. Add `audio_chunk_duration_ms` to the STT constructor so applications can tune message frequency and buffering latency while preserving the existing default.

Closes #7182. The feature issue is open for discussion; feedback on the API name and scope is welcome.

```python
from livekit.plugins import elevenlabs

stt = elevenlabs.STT(
    model="scribe_v2_realtime",
    audio_chunk_duration_ms=100,
)
```

The option accepts positive integer milliseconds, applies only to realtime STT, and is set at construction time. At 100 ms, continuous audio produces approximately 10 audio messages per second instead of 20, with up to 50 ms additional buffering. Partial chunks are still sent before the commit. This does not add reconnect pacing or establish that provider queue errors are resolved.

### Validation

- 68 ElevenLabs STT tests passed, including send-loop tests against a fake WebSocket at 8, 16, and 48 kHz with 1, 50, 75, 100, and 200 ms chunks. Tests cover release before flush, byte-for-byte audio preservation, partial tails, and commit ordering.
- Repository-wide `ruff check .` and `ruff format --check .` passed using Ruff 0.15.22 from the upstream lockfile.
- Targeted mypy for the modified plugin source and `git diff --check` passed. Full repository type checking and the full test suite have not been run.
- Live provider validation has not been run.

The constructor option is documented in its API docstring and the plugin README. No changelog or package version changes are included, per CONTRIBUTING.md.

### DIFF
--- a/livekit-plugins/livekit-plugins-elevenlabs/README.md
+++ b/livekit-plugins/livekit-plugins-elevenlabs/README.md
@@ -13,3 +13,24 @@ pip install livekit-plugins-elevenlabs
 ## Pre-requisites
 
 You'll need an API key from ElevenLabs. It can be set as an environment variable: `ELEVEN_API_KEY`
+
+## Realtime speech-to-text audio chunks
+
+Configure the outgoing audio chunk duration when creating the STT instance:
+
+```python
+from livekit.plugins import elevenlabs
+
+stt = elevenlabs.STT(
+    model="scribe_v2_realtime",
+    audio_chunk_duration_ms=100,
+)
+```
+
+`audio_chunk_duration_ms` accepts a positive integer in milliseconds and defaults
+to 50. At 100 ms, continuous audio produces approximately 10 audio messages per
+second instead of 20, at the cost of up to 50 ms additional buffering. Larger
+chunks do not pace reconnect backlogs or guarantee that provider queue errors
+are avoided. Flushes send any remaining partial chunk before the commit.
+
+This option applies only to realtime STT and is set at construction time.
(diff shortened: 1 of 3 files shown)


tinalenguyen (approved):
thanks for the PR!

Issue #7198, opened 2026-09-09
PREFLIGHT_TRANSCRIPT/INTERIM_TRANSCRIPT should clear false interruption timer


superMDguy:
### Bug Description

Currently, the false-interruption timer doesn't get cleared until a turn is committed, which can often longer than the false interruption timeout in some scenarios.

### Expected Behavior

A `PREFLIGHT_TRANSCRIPT`/`INTERIM_TRANSCRIPT` event should clear the false interruption timer.

### Reproduction Steps

```bash
Interrupt the AI, stop in the middle of a sentence, and pause.
This causes the interim transcript to get totally lost, and the AI resumes.
```

### Operating System

macOS Tahoe

### Models Used

Deepgram FLux

### Package Versions

```bash
livekit-agents==1.7.1
```

devtechedge:
Working on a focused fix so PREFLIGHT_TRANSCRIPT / INTERIM_TRANSCRIPT clear the false-interruption timer (same as a committed turn), plus a regression test. Will open a PR linking this issue.

longcw:
are you using `turn_detection="stt"`, could you share a reproducible example?

it's intended to not use `INTERIM_TRANSCRIPT` to clear the timer since usually a noise (false interruption) can generate an interim transcript. If the VAD enabled, and turn detection is the default, the VAD should pause the timer when user is speaking.

devtechedge:
Hi @longcw, thanks for the context.

To answer the repro question: I hit this with Deepgram FLux as the STT (same as reported by @superMDguy), with VAD enabled and turn detection left at the default.

The sequence is: the user interrupts the agent mid-sentence, VAD fires end_of_speech during the pause, and the false-interruption resume timer starts.

STT then keeps producing interim/preflight transcripts for the same utterance, but those do not clear the timer, so the agent resumes before the turn commits and the partial transcript is lost. @superMDguy, if your setup differs (e.g. `turn_detection="stt"`), please share it.

On the design point: I agree interim transcripts alone are not reliable evidence of real speech, since noise can generate them, so clearing the timer on any interim would be wrong.

PR #7199 narrows this: the timer is only cancelled while a false-interruption pause is already active (`_paused_speech` set) and the interim/preflight is non-empty, and it re-arms when `speaking is False`.

With default turn detection, VAD still controls the pause itself; the change only stops the resume timer from firing over an STT stream that is still producing text.

[...]

PR #7199, opened 2026-09-09
fix(voice): re-arm false-interruption timer on interim when VAD is quiet

devtechedge:
## Summary

Refs #7198. This covers the narrow window that is actually broken, not the whole issue.

With VAD configured, `on_interim_transcript` never touched the false-interruption timer at all: the branch that re-arms it is gated on `self.vad is None`. When there is no VAD `START_OF_SPEECH` but STT is still producing the turn's text, nothing cancels or re-arms the timer. That is the reported Deepgram Flux case: the far end goes quiet while Flux keeps sending interims and never sends an EOT.

This change re-arms the timer on a non-empty interim or preflight transcript while the pause is active and VAD reports no speech. The re-arm is bounded: it draws on a deferral budget of `_FALSE_INTERRUPTION_MAX_DEFERRAL_FACTOR` (3) times `false_interruption_timeout`, and stops once that budget is spent.

Only the extension actually granted is charged against that budget, and the armed deadline never moves in. An STT revises one transcript in bursts, so charging a whole timeout per revision would spend the entire budget while shifting the resume once, and a later interim from a still-live turn would be ignored.

It deliberately does not clear the timer. Interims are also produced by noise, as @longcw pointed out on the issue, so clearing would let a noisy room hold the pause open for as long as it keeps producing text. Re-arming fails in the right direction: a live turn keeps deferring until its final lands, and pure noise still resumes once the interims stop.

The bound is what keeps that true. An unbounded re-arm only moves the dead air: pipecat-ai/pipecat#5735 replayed a call recording into Flux and got interims with no EOT for the whole file, so the pause would stay open until the caller hangs up. With the cap the worst case is the agent resuming, which the caller can react to.

Kept separate from #7101 as requested on the issue: same lines, but a different decision. #7101 is about what the resume discards, this is about when it fires.

Two things worth stating plainly.

First, sustained VAD already resets the timer independently of `interruption.min_duration`, because `on_start_of_speech` cancels it on every `START_OF_SPEECH`. `min_duration` only gates whether `on_vad_inference_done` reaches `_interrupt_by_audio_activity`, which is the pause. The gap is narrower than "any `turn_detection="stt"` utterance".

Second, this does not by itself save the buffered transcript. Once a stalled provider stops emitting interims entirely, the timer still expires and the agent resumes. What this fixes is the case where interims are still arriving, which is the common one. Preserving the turn across the resume is #7101's territory, and the two are complementary rather than overlapping.

## Changes

- `livekit-agents/livekit/agents/voice/agent_activity.py`
  - `_defer_false_interruption_timer()`: re-arm the timer within a bounded budget, never clear it. Charges only the granted extension and never pulls the deadline earlier than it already is.
  - `_cancel_false_interruption_timer()` resets the budget, so it is per pause
  - `on_interim_transcript` calls the bounded variant instead of `_start_false_interruption_timer`
- `tests/test_false_interruption_resume.py`
  - interim while speaking leaves the timer alone, interim and preflight re-arm when VAD is quiet, an empty interim does not re-arm
  - a stream of interims past the cap still resumes, and the budget resets for the next pause
  - clustered interims do not spend the budget, so a later interim still defers

## Test plan

- [x] `pytest tests/test_false_interruption_resume.py -q` (16 passed)
- [x] `ruff check` / `ruff format --check` on touched files

## Notes

The cap is a module constant, not a new option. Promoting it to `TurnHandlingOptions` is easy if maintainers want it configurable; I did not add public API surface without agreement on the semantics.


devtechedge:
Thanks for the context @longcw.

That matches what we saw in the code: interim transcripts intentionally don't clear the false-interruption timer since noise can also produce interims, and VAD start-of-speech cancels the timer when the user actually speaks, regardless of turn detection.

Given that, this PR's approach of clearing the timer on interim/preflight conflicts with the intended semantics. We're happy to close it unless you see a narrower fix worth keeping here, for example if sustained VAD speech still fails to cancel the timer in some `turn_detection="stt"` path.

We'll also follow your answer to @superMDguy's question about sustained VAD under `turn_detection="stt"`.

devtechedge:
This is now the narrower fix rather than the clear-on-interim approach that comment describes.

With VAD configured, the re-arm branch in `on_interim_transcript` was gated on `vad is None`, so interims never touched the timer at all.

Instead of clearing, it now re-arms on a non-empty interim or preflight when VAD reports no speech and the pause is active, which is what finals already do.

Noise still resumes once the interims stop, so the semantics @longcw described on #7198 are preserved.

Leaving this open rather than closing, since the narrower window is real, but happy to close it if you would rather handle it in #7101.

devtechedge:
Hi - soft check-in on this false-interruption timer fix whenever you have a moment.

With VAD configured, `on_interim_transcript` never re-armed the false-interruption timer because that branch was gated on `self.vad is None`, so quiet far-end audio with ongoing STT interims (the Deepgram Flux case) could fire a false resume (`Refs #7198`).

This re-arms on a non-empty interim / preflight while the pause is active and VAD reports no speech, bounded by a deferral budget of `_FALSE_INTERRUPTION_MAX_DEFERRAL_FACTOR` (3) times `false_interruption_timeout`, charging only the extension actually granted.

There is focused coverage in `test_false_interruption_resume.py`, and the change stays in `agent_activity.py` without touching the separate hold-interruptions work in #7201.

Happy to rebase, retune the budget, or adjust naming if that helps review - no rush from my side.

Issue #7204, opened 2026-09-10
Expose configurable outgoing audio chunk duration for Deepgram Flux


dnaumenko:
## Proposal

Expose an optional `chunk_size_ms` argument on `deepgram.STTv2` so applications can configure outgoing PCM WebSocket batching. Keep the existing 50 ms default; allow opting into 80 ms without replacing the plugin's send loop.

In the tested Deepgram plugin version (1.6.1), `SpeechStreamv2._run` sets `samples_per_channel` to `sample_rate // 20`. [Deepgram's Flux quickstart](https://developers.deepgram.com/docs/flux/quickstart) recommends 80 ms chunks (2,560 bytes for mono linear16 at 16 kHz).

## Updated evidence: isolated Flux benchmark

We repeated the comparison with **24 direct WebSocket trials from a LiveKit Cloud worker in `us-east`**, removing RTC media, VAD, LiveKit endpointing, LLM, and TTS from the measured path. This follow-up tests the effect of PCM batching on observed Flux EOT latency.

The primary metric is **the interval from acoustic speech end on a real-time-paced PCM timeline to receipt of Flux `EndOfTurn` by the worker**. Acoustic speech end is the end of the last 10 ms window with normalized RMS above 0.025. Both timestamps use the same worker-side monotonic clock.

| Synthetic utterance | Mean EOT latency, 50 ms chunks | Mean EOT latency, 80 ms chunks | Difference, 80 minus 50 |
| --- | ---: | ---: | ---: |
| “Please tell me where to put the blue notebook.” | 437.5 ms | 400.7 ms | −36.8 ms |
| “Please put the blue notebook.” | 446.2 ms | 418.8 ms | −27.5 ms |
| “on the wooden table next to the lamp.” | 532.4 ms | 523.5 ms | −8.9 ms |
| **All trials** | **472.0 ms** | **447.7 ms** | **−24.4 ms (about 5.2%)** |

[...]

PR #7205, opened 2026-09-10
feat(deepgram): make Flux audio chunk duration configurable

dnaumenko:
Applications currently have to replace or patch `SpeechStreamv2._run` to evaluate Flux with an outgoing audio chunk size other than the hardcoded 50 ms. This adds a constructor-only `chunk_size_ms` option to `deepgram.STTv2`, retaining 50 ms as the default.

```python

stt = deepgram.STTv2(
    model="flux-general-multi",
    chunk_size_ms=80,
)
```

[Deepgram recommends 80 ms audio chunks for Flux](https://developers.deepgram.com/docs/flux/quickstart). The setting controls outgoing mono PCM WebSocket batching.

The constructor rejects non-positive/non-integer durations and durations that yield fewer than one sample. The existing send loop still handles partial-chunk flushing and stream closure.

### Validation

- `uv run --no-sync pytest tests/test_plugin_deepgram_stt.py -q`: 38 passed.
- New fake-WebSocket tests exercise the real public stream/send loop at 8/16/48 kHz, default 50 ms and explicit 20/80 ms chunks, byte-for-byte PCM preservation, partial tails, and closure ordering. Constructor validation is also covered.
- `make fix` and `make check` passed with `UV_NO_SYNC=1`: repository formatting/lint and mypy over 657 source files. The isolated environment includes the locked dev/typing groups plus optional boto3 and missing types-jsonschema/types-nanoid stubs.
- `git diff --check` passed. No live-provider performance test was run against this new patch.

Related proposal: #7204. Opening as a draft for API feedback.

### DIFF
--- a/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt_v2.py
+++ b/livekit-plugins/livekit-plugins-deepgram/livekit/plugins/deepgram/stt_v2.py
@@ -54,6 +54,7 @@ class STTOptions:
     sample_rate: int
     keyterm: str | Sequence[str]
     endpoint_url: str
+    chunk_size_ms: int = 50
     language: str = "en"
     eager_eot_threshold: NotGivenOr[float] = NOT_GIVEN
     eot_threshold: NotGivenOr[float] = NOT_GIVEN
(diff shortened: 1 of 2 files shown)


rchengLK:
@dnaumenko could you resolve the merge conflicts first please?

PR #7207, opened 2026-09-10
(stt): reset the retry budget on any event, not only on a final transcript

GregHolmes:
## Summary

`RecognizeStream._main_task` gives a stream `max_retry` attempts (3 by default). `_num_retries` only ever increments, and its single reset lives in `_metrics_monitor_task`, on `FINAL_TRANSCRIPT`:

https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/stt/stt.py#L541-L543

That makes the budget a **stream-lifetime counter replenished by the caller speaking**, rather than a **consecutive-failure counter replenished by the connection recovering**.

A stream that reconnects cleanly but sits through silence never earns its budget back. `max_retry` drops spread over one long call then kill it permanently, with `recoverable=False` and no further reconnect, even though every one of those reconnects succeeded. The drops need not be consecutive or close together: a blip at minute 1, 10 and 40, then a fourth, is enough.

## Why silence is the ordinary case, not an edge case

The gaps where a voice agent has no final transcript are exactly the gaps where it is most exposed: the agent is talking, the caller is on hold, an IVR is playing, someone stepped away. That is also when an idle socket is most likely to be reaped by a NAT or a load balancer. The condition that spends the budget and the condition that blocks its reset are the same condition.

## Reproduction

Three streams whose connections all succeed and then drop. Ten drops each. The only difference is what they deliver in between.

| what the stream delivers between drops | connections made | outcome |
|---|---|---|
| nothing at all | 4 | dies: `failed to recognize speech after 3 attempts` |
| a `FINAL_TRANSCRIPT` each time | 11 | survives |
| `RECOGNITION_USAGE` only | 4 | **dies** |

The third row is the ordinary case: the Deepgram plugin emits a usage event every 5s while audio flows, speech or no speech. Same failures, opposite outcome, decided purely by whether somebody was talking.

Instrumenting `_num_retries` shows the mechanism. With transcripts it sits at a steady 1 forever, reset by the event and incremented by the failure. Without them it climbs 0, 1, 2, 3 and the stream is gone.

A runnable script is in the PR below as two tests, but the shape is just:

```python
class _FlappingStream(RecognizeStream):
    async def _run(self) -> None:
        self.runs += 1
        if self._emit is not None:
            self._event_ch.send_nowait(self._emit)   # usage event, no transcript
            await asyncio.sleep(0)
        if self.runs <= self._drops:
            raise APIConnectionError("socket dropped")
        await asyncio.sleep(3600)
```

## Is this intended?

I want to flag that it might be. "A stream that has never been understood should not retry forever" is a defensible rule, and resetting on `FINAL_TRANSCRIPT` implements it. If that is the intent, then the bug is narrower: the budget should still reset once a *reconnect* succeeds, rather than only once recognition succeeds.

## Suggested fix

Reset on any event from the provider rather than only on `FINAL_TRANSCRIPT`. Any event proves the connection came up and delivered. A connection that has never delivered anything at all still gives up as before, so this does not become an unbounded retry.

Reported as #7208.

Happy to take it a different way if you would rather the reset hung off a successful connection explicitly, though that needs a signal out of `_run()` that the base class does not have today.

### DIFF
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -356,6 +356,26 @@ class SpeakerContext(Protocol):
     def to_instructions(self) -> str: ...
 
 
+class _HealthySignallingChan(aio.Chan[SpeechEvent]):
+    """Event channel that clears the retry budget the moment an event is published.
+
+    The reset has to happen in the producer's own turn. `send_nowait` only wakes the
+    metrics consumer, it does not run it, so an attempt that publishes an event and
+    then raises in the same event-loop turn would reach the terminal branch of
+    `RecognizeStream._main_task` with the budget still exhausted. Signalling here
+    also ties the reset to the attempt that produced the event, rather than to
+    whenever a consumer happens to drain it.
+    """
+
+    def __init__(self, stream: RecognizeStream) -> None:
+        super().__init__()
+        self._stream = stream
+
+    def send_nowait(self, value: SpeechEvent) -> None:
+        self._stream._num_retries = 0
+        super().send_nowait(value)
+
+
 class RecognizeStream(ABC):
     class _FlushSentinel:
         """Sentinel to mark when it was flushed"""
@@ -380,7 +400,7 @@ class RecognizeStream(ABC):
         self._stt = stt
         self._conn_options = conn_options
         self._input_ch = aio.Chan[rtc.AudioFrame | RecognizeStream._FlushSentinel]()
-        self._event_ch = aio.Chan[SpeechEvent]()
+        self._event_ch = _HealthySignallingChan(self)
 
         self._tee = aio.itertools.tee(self._event_ch, 2)
         self._event_aiter, monitor_aiter = self._tee
@@ -518,14 +538,6 @@ class RecognizeStream(ABC):
         """Task used to collect metrics"""
 
         async for ev in event_aiter:
-            # Any event proves the connection came up and delivered something, so the
-            # budget in _main_task is for consecutive failures rather than for the
-            # lifetime of the stream. Resetting only on FINAL_TRANSCRIPT tied it to
-            # the caller speaking: a stream that reconnected cleanly but sat through
-            # silence never got its budget back, and max_retry drops spread over a
-            # long call would kill it for good.
-            self._num_retries = 0
-
             if ev.type == SpeechEventType.RECOGNITION_USAGE:
                 assert ev.recognition_usage is not None, (
                     "recognition_usage must be provided for RECOGNITION_USAGE event"
--- a/livekit-agents/livekit/agents/stt/stt.py
+++ b/livekit-agents/livekit/agents/stt/stt.py
@@ -518,6 +518,14 @@ class RecognizeStream(ABC):
         """Task used to collect metrics"""
 
         async for ev in event_aiter:
+            # Any event proves the connection came up and delivered something, so the
+            # budget in _main_task is for consecutive failures rather than for the
+            # lifetime of the stream. Resetting only on FINAL_TRANSCRIPT tied it to
+            # the caller speaking: a stream that reconnected cleanly but sat through
+            # silence never got its budget back, and max_retry drops spread over a
+            # long call would kill it for good.
+            self._num_retries = 0
+
             if ev.type == SpeechEventType.RECOGNITION_USAGE:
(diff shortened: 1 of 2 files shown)


livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -518,6 +518,14 @@ async def _metrics_monitor_task(self, event_aiter: AsyncIterable[SpeechEvent]) -
         """Task used to collect metrics"""
 
         async for ev in event_aiter:
+            # Any event proves the connection came up and delivered something, so the
+            # budget in _main_task is for consecutive failures rather than for the
+            # lifetime of the stream. Resetting only on FINAL_TRANSCRIPT tied it to
+            # the caller speaking: a stream that reconnected cleanly but sat through
+            # silence never got its budget back, and max_retry drops spread over a
+            # long call would kill it for good.
+            self._num_retries = 0
```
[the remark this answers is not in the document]
GregHolmes: Confirmed, and fixed in 6070582. Thank you, this was a real gap and my test was hiding it exactly as you describe.

Reproducing it directly: a stream that enters the event-delivering attempt already at `max_retry` and raises in the same turn as the send dies at run 4, while the identical stream with a `sleep(0)` after the send survives all twelve drops. So the consumer-side reset only worked because the test yielded.

The reset now happens at publication rather than on consumption, in a small `Chan` subclass:

```python
def send_nowait(self, value: SpeechEvent) -> None:
    self._stream._num_retries = 0
    super().send_nowait(value)
```

That puts it in the producer's own turn, so a plugin emitting its last usage event during teardown counts before the error propagates. It also handles your second point: because the signal fires at send time rather than at drain time, a buffered event from an older attempt can no longer reset a newer attempt's budget.

Test added as you suggested, entering at `max_retry` with no `await` after the send. Three-way check:

| | `...survives_drops_while_the_caller_is_silent` | `...resets_even_when_the_producer_never_yields` |
|---|---|---|
| `main` | fail | fail |
| e62c5ad (consumer-side reset) | pass | **fail** |
| 6070582 (publication-time reset) | pass | pass |

livekit-agents/livekit/agents/stt/stt.py
```diff
@@ -538,9 +558,6 @@ async def _metrics_monitor_task(self, event_aiter: AsyncIterable[SpeechEvent]) -
                 )
 
                 self._stt.emit("metrics_collected", stt_metrics)
-            elif ev.type == SpeechEventType.FINAL_TRANSCRIPT:
```
longcw, line 541: I'd like to put the reset back in `_metrics_monitor_task` and widen it from `FINAL_TRANSCRIPT` to other events that verify the connection is health , instead of overriding `send_nowait` in `_HealthySignallingChan`.

longcw: also need to add that for `MultiSpeakerAdapter`  https://github.com/livekit/agents/blob/4de62322fa84b7c1736370ff3ebbd66b5d0ffe6f/livekit-agents/livekit/agents/stt/multi_speaker_adapter.py#L143-L145

longcw (changes requested):
thanks for the pr, I think the analysis is right but need some changes:

GregHolmes:
Addressed in e633a3f36.

- Removed the custom event-channel override and restored the reset to `_metrics_monitor_task`.
- The budget now resets after provider-originated `INTERIM_TRANSCRIPT`, `PREFLIGHT_TRANSCRIPT`, `FINAL_TRANSCRIPT`, or `RECOGNITION_USAGE` events. `START_OF_SPEECH` and `END_OF_SPEECH` stay excluded because adapters can synthesize them locally.
- Applied the same reset conditions in `MultiSpeakerAdapterWrapper`.
- Updated regression coverage for all health-event types and the multi-speaker adapter usage-event path.

Verified with `uv run pytest tests/test_stt_base.py tests/test_multi_speaker_adapter.py` and Ruff.

livekit-agents/livekit/agents/stt/stt.py
GregHolmes: Devin raised two valid edge cases: usage events can be generated locally, and the monitor can run too late. Would you be comfortable with a small per-attempt recovery marker, set only after a real provider response and checked by the retry logic? That avoids both endless retries and missed recoveries.

PR #7209, opened 2026-09-10
feat(plugins-tavus): default to the stock Lucy face when no face or pal is given

carolin-tavus:
## Summary

When neither `face_id` nor `pal_id` is supplied, the plugin falls back to the stock pal (`DEFAULT_PAL_ID`) and omitted `face_id`, so the pal's own default face was used — currently Luna (`r9d30b0e55ac`, phoenix-3). This adds `DEFAULT_FACE_ID = "r4067604db72"` (stock "Lucy - Home", phoenix-4.5) and sends it explicitly in that fully-default case.

Scoped to the default-pal branch only, so existing behaviour is preserved everywhere else:
- a user-supplied `pal_id` still carries its own default face (`face_id` stays omitted)
- an explicit `face_id`, or `TAVUS_FACE_ID` / deprecated `TAVUS_REPLICA_ID`, still wins

## Test plan

- `test_defaults_to_stock_pal_and_face_when_neither_given` (renamed from `test_defaults_to_stock_pal_when_neither_given`) now asserts `face_id == DEFAULT_FACE_ID`
- new `test_env_face_wins_over_default_face` covers `TAVUS_FACE_ID` overriding the default with the default pal
- existing `test_pal_id_only_skips_pal_creation_and_omits_face` still passes, guarding the user-pal case
- `pytest livekit-plugins/livekit-plugins-tavus/tests/test_tavus.py`: 7 passed, 3 failed — the 3 failures (`*_still_work_and_warn`, `test_avatar_session_resolves_new_and_deprecated_args`) are pre-existing on `main` (same 3 fail on an untouched checkout) and unrelated: they expect `DeprecationWarning` while the code emits `logger.warning`
- `ruff check` / `ruff format --check` clean on both files

### DIFF
--- a/livekit-plugins/livekit-plugins-tavus/livekit/plugins/tavus/api.py
+++ b/livekit-plugins/livekit-plugins-tavus/livekit/plugins/tavus/api.py
@@ -98,7 +98,10 @@ class TavusAPI:
             or NOT_GIVEN
         )
 
-        if not pal_id:
+        # extra_payload overrides the payload wholesale, so a pal named there counts
+        # as user-supplied when deciding whether to fill in the stock face.
+        extra = dict(extra_payload) if utils.is_given(extra_payload) else {}
+        if not pal_id and not extra.get("pal_id"):
             # no pal supplied — use the default stock pal and, unless overridden, its stock face
             pal_id = DEFAULT_PAL_ID
             face_id = face_id or DEFAULT_FACE_ID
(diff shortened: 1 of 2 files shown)


carolin-tavus:
Superseded by #7211, which has the same changes (identical tree) authored under the correct git identity so the CLA check can pass. Thanks @devin-ai-integration for the extra_payload catch — that fix is included there.

PR #7228, opened 2026-09-11
fix(voice): finish background audio cleanup when the room disconnects

kiwijuicer:
`BackgroundAudioPlayer.aclose()` can wait indefinitely for `unpublish_track()` when the room has already disconnected or disconnects before the acknowledgement arrives. This can hold up agent shutdown after a call ends.

Race unpublishing against the room's `disconnected` event, skip the request when already disconnected, and remove the listener and settle the owned task on every exit. Connected rooms still wait for the acknowledgement; existing exception suppression and caller cancellation behavior are preserved.

Regression coverage exercises normal acknowledgement, both disconnect orderings, cancellation, and an unpublish error. The two disconnect cases fail on unmodified upstream and all five pass with this fix. `make check` passes.


longcw (approved):
thanks for the pr, looks good to me!

PR #7230, opened 2026-09-11
fix(openai): ignore incomplete and duplicate GPT Live backend tool calls

kiwijuicer:
A wrapped Responses `response.output_item.done` event does not establish that its function-call item completed. GPT Live currently emits the call even when its status is incomplete, and emits a second call when the same completed item is delivered again.

Declare the output item's optional status, require `completed` before recording or dispatching a function call, and use the current response's existing call-ID set to suppress duplicate dispatch. Deduplication stays scoped to the tracked response; this does not claim exactly-once execution across reconnects or responses.

Regression coverage includes missing/null/incomplete/in-progress/failed status, a later completed event, duplicate calls, distinct calls, and response continuation after all results arrive. The GPT Live and duplex adapter suites pass (103 tests), as does `make check`.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
@@ -734,6 +734,8 @@ class GPTLiveSession(
             item = event.item
             if item is None or item.type != "function_call":
                 return
+            if item.status != "completed":
+                return
             if not item.call_id or not item.name or item.arguments is None:
                 logger.warning(
                     "gpt-live dropping function call with missing fields",
@@ -746,6 +748,8 @@ class GPTLiveSession(
                     extra={"call_id": item.call_id, "delegation_id": d_id},
                 )
                 pending = self._delegated_responses[d_id] = _DelegatedResponse(completed=True)
+            if item.call_id in pending.call_ids:
+                return
             pending.call_ids.add(item.call_id)
             self._fnc_call_to_delegation[item.call_id] = d_id
 
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_types.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_types.py
@@ -260,10 +260,11 @@ class ResponseSnapshot(BaseModel):
 
 
 class OutputItem(BaseModel):
-    """Only a completed function-call item carries all of name, call id and arguments."""
+    """Fields needed to dispatch a completed backend function call."""
 
     id: str | None = None
     type: str | None = None
+    status: str | None = None
     call_id: str | None = None
     name: str | None = None
     arguments: str | None = None
--- a/tests/test_gpt_live_model.py
+++ b/tests/test_gpt_live_model.py
@@ -987,6 +987,79 @@ async def test_a_response_continues_only_once_every_call_has_its_answer(
         await model.aclose()
 
 
+@pytest.mark.parametrize("status", ["missing", None, "incomplete", "in_progress", "failed"])
+async def test_noncompleted_backend_calls_are_not_dispatched(
+    monkeypatch: pytest.MonkeyPatch, status: str | None
+) -> None:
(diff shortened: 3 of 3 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_types.py
```diff
@@ -260,10 +260,11 @@ class ResponseSnapshot(BaseModel):
 
 
 class OutputItem(BaseModel):
-    """Only a completed function-call item carries all of name, call id and arguments."""
+    """Fields needed to dispatch a completed backend function call."""
 
     id: str | None = None
     type: str | None = None
+    status: str | None = None
```
longcw:
```suggestion
    status: status: Literal["in_progress", "completed", "incomplete"] | None = None
```

longcw (approved):

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
```diff
@@ -734,6 +734,8 @@ def _handle_response_event(self, envelope: types.ResponseEventEnvelope) -> None:
             item = event.item
             if item is None or item.type != "function_call":
                 return
+            if item.status != "completed":
+                return
```
longcw, line 742:
```suggestion
            if item.status != "completed":
                logger.debug(
                    "gpt-live ignoring incomplete function call",
                    extra={"name": item.name, "status": item.status},
                )
                return
```

kiwijuicer: Added in 624a5f204, using `function_name` instead of `name` in `extra`. `name` is a reserved Python `LogRecord` attribute, so the suggested form raises `KeyError: "Attempt to overwrite 'name' in LogRecord"` when debug logging is enabled.

I also corrected the same key in the adjacent missing-fields warning. The new regression tests reproduced that error before the fix; all 106 GPT Live/duplex tests now pass with debug logging enabled, and `make check` passes.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
```diff
@@ -734,10 +734,16 @@ def _handle_response_event(self, envelope: types.ResponseEventEnvelope) -> None:
             item = event.item
             if item is None or item.type != "function_call":
                 return
+            if item.status != "completed":
+                logger.debug(
+                    "gpt-live ignoring incomplete function call",
+                    extra={"function_name": item.name, "status": item.status},
+                )
```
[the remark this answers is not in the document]
kiwijuicer: Fixed in 143d85b7d. The debug log now uses `lk.pii.function_name`, so the existing redaction filter removes the provider-supplied name when redaction is enabled. I also marked the raw status as `lk.pii.status`: the lenient event parser can retain unexpected strings despite the Literal annotation.

Extended the existing provider-content redaction test to cover rejected calls and verify that filtering removes the private values. All 118 GPT Live, duplex, and telemetry redaction tests pass with debug logging enabled; `make check` also passes.

livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
```diff
@@ -734,10 +734,16 @@ def _handle_response_event(self, envelope: types.ResponseEventEnvelope) -> None:
...
+                return
             if not item.call_id or not item.name or item.arguments is None:
                 logger.warning(
                     "gpt-live dropping function call with missing fields",
-                    extra={"call_id": item.call_id, "name": item.name},
+                    extra={"call_id": item.call_id, "function_name": item.name},
                 )
```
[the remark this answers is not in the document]
kiwijuicer: Fixed in 143d85b7d. The missing-fields warning now uses `lk.pii.call_id` and `lk.pii.function_name`, keeping the static warning message while allowing the existing redaction filter to remove both provider-supplied values.

The extended redaction test covers a malformed completed call with private values in both fields and verifies that neither survives filtering.

PR #7231, opened 2026-09-11
fix(openai): preserve the GPT Live cumulative usage watermark

kiwijuicer:
GPT Live usage is cumulative, but every report currently replaces the watermark on upstream. Reports of 10, 8, and 12 seconds produce deltas of 10, 0, and 4 seconds, overcounting usage. Invalid values can also contaminate later metrics or prevent the final close event from acknowledging shutdown.

Advance the watermark only for finite, nonnegative numeric durations strictly greater than the previous total. Validate raw usage before OpenAI's event constructor coerces values: oversized integers otherwise raise during conversion, and booleans can become numeric durations. Invalid usage is ignored while the enclosing `session.closed` event still completes its acknowledgement. The original event payload is preserved.

Regression coverage includes duplicate/regressive reports, malformed and non-finite values, oversized positive and negative integers, boolean coercion, and final-close usage. The GPT Live and duplex adapter suites pass (114 tests), as does `make check`.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
@@ -4,6 +4,7 @@ import asyncio
 import base64
 import contextlib
 import json
+import math
 import os
 import time
 from collections.abc import AsyncIterable
@@ -837,6 +838,14 @@ class GPTLiveSession(
 
     def _handle_usage(self, usage: types.Usage) -> None:
         # reported cumulatively for the whole session, so only the delta goes to the collectors
+        seconds = usage.seconds
+        if (
+            isinstance(seconds, bool)
+            or not isinstance(seconds, (int, float))
+            or not math.isfinite(seconds)
+            or seconds <= self._usage_total.seconds
+        ):
+            return
         previous, self._usage_total = self._usage_total, usage
         self.emit(
             "metrics_collected",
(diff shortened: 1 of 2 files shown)


livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
```diff
@@ -837,6 +853,9 @@ def _handle_session_closed(self, event: types.SessionClosedEvent) -> None:
 
     def _handle_usage(self, usage: types.Usage) -> None:
         # reported cumulatively for the whole session, so only the delta goes to the collectors
+        seconds = usage.seconds
+        if not _valid_usage_seconds(seconds) or seconds <= self._usage_total.seconds:
```
longcw, line 857: ideally, the usage total on the wire is monotonic: `session.usage.updated` arrives about every 15 s with a growing `seconds`, and `session.closed` repeats the last value.

can you share the logs that showing a different case before we moving forward on this pr?

kiwijuicer: Thanks for checking. I don’t have provider logs showing decreasing usage. The `10 → 8 → 12` example comes from a synthetic regression test added during SDK review; it wasn’t observed in a live session.

The change provides defensive handling if an unexpected report arrives. The malformed-value cases are synthetic too. I should have made that clearer in the description.

Would you consider this useful as defensive hardening? If the SDK should rely on the provider’s monotonicity guarantee, I’m happy to withdraw it.

PR #7233, opened 2026-09-11
fix: prevent llm fallback retries after output

chenghao-mou:
Child and outer retries could replay text or tool calls after provider output. Both now use the shared retry guard when `retry_on_chunk_sent=False`. Preserve pre-output, metadata-only, and explicit opt-in retries.

Addresses [AGT-3492](https://linear.app/livekit/issue/AGT-3492). Builds on @dtran26's diagnosis and reproducers in [agents-js#2477](https://github.com/livekit/agents-js/issues/2477) and the outer-retry fix in [agents-js#2480](https://github.com/livekit/agents-js/pull/2480).

<details>

**Model:** GPT-6

> Fix the Python LLM fallback adapter’s outer retry bug in livekit/agents.
>
> We investigated livekit/agents-js#2477 and confirmed that Python shares one part of the bug: when a provider emits text or tool calls, then raises a retryable APIError, retry_on_chunk_sent=False stops provider fallback but does not stop LLMStream’s outer retry loop. Generation restarts and appends duplicate output.
>
> Verified Python revision: 4de62322fa84b7c1736370ff3ebbd66b5d0ffe6f. Fetch current main and check whether this still applies.
>
> Relevant paths:
> - livekit-agents/livekit/agents/llm/fallback_adapter.py: _run re-raises the original error after output without changing retryability.
> - livekit-agents/livekit/agents/llm/llm.py: _main_task retries that error.
> - AgentSession passes connection options with max_retry=3 by default. Direct adapter.chat() defaults to zero outer retries.
>
> A controlled provider reproduces this without credentials or network calls:
> 1. First request emits “The answer.”, then raises APITimeoutError(retryable=True).
> 2. Next request succeeds with “The answer.”
> 3. With retry_on_chunk_sent=False and outer max_retry=3, output becomes “The answer.The answer.”
> Tool-call emissions duplicate through the same path.
>
> Python already isolates errors by stream, accepts successful child retries, and cancels active children correctly. Those JS-specific failures do not need porting. Python’s Inference provider marks errors non-retryable after output, but custom providers can expose the adapter bug.
>
> Please:
> - Find or create a Linear ticket before creating the branch.
> - Reproduce the bug, then make the smallest coherent fix.
> - Cover text and tool-call replay, preserve pre-output retries and metadata-only behavior, and preserve explicit retry_on_chunk_sent=True.
> - Run relevant tests and required checks.
> - Leave unrelated working-tree changes untouched.
>
> Local investigation:
> - /tmp/agents-js-2477-investigation/python-parity.md
> - /tmp/agents-js-2477-investigation/python_parity.py
> The harness asserts the observed bug; update expectations for regression tests.
>
> Related Node fix: https://github.com/livekit/agents-js/pull/2480
> Original diagnosis and reproducer: https://github.com/livekit/agents-js/issues/2477 by [dtran26.](dtran26.) Preserve contributor credit if that work shapes the fix.
>
> Do not post comments, replies, or resolve threads on my behalf.

> let's create a draft PR.

</details>

### DIFF
--- a/livekit-agents/livekit/agents/llm/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/fallback_adapter.py
@@ -197,6 +197,8 @@ class FallbackLLMStream(LLMStream):
                     retry_interval=self._fallback_adapter._retry_interval,
                 ),
             ) as stream:
+                if not check_recovery:
+                    stream._retry_on_chunk_sent = self._fallback_adapter._retry_on_chunk_sent
                 should_set_current = not check_recovery
                 async for chunk in stream:
                     if should_set_current:
--- a/livekit-agents/livekit/agents/llm/fallback_adapter.py
+++ b/livekit-agents/livekit/agents/llm/fallback_adapter.py
@@ -287,7 +287,7 @@ class FallbackLLMStream(LLMStream):
                         self._event_ch.send_nowait(result)
 
                     return
-                except Exception:  # exceptions already logged inside _try_generate
+                except Exception as e:  # exceptions already logged inside _try_generate
                     if llm_status.available:
                         llm_status.available = False
                         self._fallback_adapter.emit(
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
lg

theomonnom (approved):

PR #7234, opened 2026-09-11
fix(openai): continue a GPT Live backend only once every call in the conversation is answered

dmt219:
GPT Live refuses `response.create` with `function_call_outputs_required` while any function call in the backend conversation is still unanswered, across delegations (the event carries no delegation id). `_maybe_continue_response` gated the continuation per response, so with two responses in flight the one whose tool finished first sent its continuation while the other's tool was still running; the service refused it, and since that response's bookkeeping had already been dropped, nothing ever continued it. The backend never answered and the voice model waited for good. Seen in production today with a slow `lookup_preferences` overlapping a fast `lookup_plan`:

```
code: function_call_outputs_required
message: Submit the pending function call outputs before response.create.
client_event_id: response_create_d003aedf7086
```

Make the barrier conversation-wide: a continuation goes out only when no tracked response has an open call, and every response that has finished asking then continues, oldest first. Remember each continuation until its `response.created` arrives; when the service refuses one anyway, put the response back behind the barrier so the next output re-sends it, bounded so a call this session never tracked cannot loop.

Regression coverage: two delegations with one slow call, a text-only response that neither blocks nor continues, a refused continuation re-armed and re-sent, and a repeatedly refused one given up on. The GPT Live and duplex adapter suites pass (101 tests); `ruff check` and `ruff format --check` are clean. I could not run `make type-check` locally (no uv).

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/gpt_live_model.py
@@ -748,7 +748,13 @@ class GPTLiveSession(
             for event_id, (sent_d_id, _) in list(self._continuations_sent.items()):
                 if sent_d_id == d_id:
                     del self._continuations_sent[event_id]
-            self._delegated_responses[d_id] = _DelegatedResponse()
+            created = _DelegatedResponse()
+            if (previous := self._delegated_responses.get(d_id)) is not None:
+                # the calls the replaced response still waits on are open on the backend all
+                # the same, so the new response waits on them too
+                created.call_ids |= previous.call_ids - previous.returned
+                created.rearms = previous.rearms
+            self._delegated_responses[d_id] = created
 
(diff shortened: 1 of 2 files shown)


dmt219:
Follow-up commit addresses the three review points: a failed or incomplete blocker re-runs the barrier so the responses waiting behind it continue; a refusal that lands after the output that opened the barrier is retried at once from the re-arm; call ids in both log lines moved under `lk.pii.` keys. Also closes a gap found while testing: `response.created` replaced a delegation's entry outright, so calls the replaced response still waited on dropped out of the barrier's view. The new response now inherits them and the re-arm budget. Five regression tests added; GPT Live and duplex suites at 105 passed, ruff check and format clean.

dmt219:
Applied: a replaced response's answered calls now drop their routing entry (the suggested pop), with the assertion added to the inheritance test. 105 passed, ruff clean.

longcw:
thanks for the pr! closing in favor of https://github.com/livekit/agents/pull/7266

PR #7236, opened 2026-09-11
fix: refuse to start a DuplexModel under a text simulation

u9g:
## Problem

A duplex model such as `GPTLiveModel` hears and speaks only audio, and `DuplexRealtimeAdapter` resolves each reply from the audio the model produces. A text simulation (`lk agent simulate` without `audio`) disables audio I/O, so the first reply never opens and every scenario fails on turn one with:

```
failed to generate a reply: the model did not start speaking when asked
```

That message arrives after a 10 second timeout and says nothing about the cause. Reproduced with a GPT-Live agent: all scenarios failed identically in text mode and passed in `lk agent simulate audio`.

## Fix

`AgentActivity` raises at construction when the session is a text simulation and the resolved llm is a `DuplexRealtimeAdapter`:

```
a DuplexModel speaks only through audio, so it cannot run under a text simulation; run `lk agent simulate audio` instead
```

No new capability flag: every duplex model is audio-only by definition, so the type is the signal.

## Tests

Two tests in `tests/test_duplex_adapter.py`: the session refuses to start with `_text_only` patched on, and starts normally otherwise.

agents-js counterpart: https://github.com/livekit/agents-js/pull/2481

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -308,6 +308,14 @@ class AgentActivity(RecognitionHooks):
             )
             self._session._warned_realtime_audio_redaction = True
 
+        # a duplex model has no text modality, and the adapter resolves each reply from the audio
+        # the model produces; without audio a text simulation would only time out on turn one
+        if self._text_only and isinstance(self.llm, llm.DuplexRealtimeAdapter):
+            raise RuntimeError(
+                "a DuplexModel speaks only through audio, so it cannot run under a text "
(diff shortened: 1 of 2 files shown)


u9g:
Before and after under `lk agent simulate` (text mode) with a GPT-Live agent: the released SDK times out on turn one with "the model did not start speaking when asked" and no hint why; this branch refuses at session start and names `lk agent simulate audio` as the mode to use.

https://github.com/user-attachments/assets/5e7958ff-2149-4ebe-b86e-c561c2cb4df9

chenghao-mou (approved):

PR #7237, opened 2026-09-11
fix(respeecher): invalidate the pool on model change instead of retiring it

Rehansanjay:
## Problem

Respeecher bakes the model into the websocket URL:

```python
full_ws_url = f"{ws_url}{self._opts.model}/tts/websocket?api_key=...&source=...&version=..."
```

so `update_options(model=...)` can't reuse pooled connections. It handles that by setting the whole pool aside and building a new one:

```python
self._retired_pools.append(self._pool)
self._pool = utils.ConnectionPool[aiohttp.ClientWebSocketResponse](
    connect_cb=self._connect_ws,
    close_cb=self._close_ws,
)
```

The retired pools are only closed in `aclose()`. Nothing drains them in between, which leaks in two ways:

1. **Idle sockets opened for the old model stay open** until the TTS is closed. Every model change adds another pool's worth.
2. **A stream in flight during the change returns its socket to the retired pool.** `SynthesizeStream` acquires through `self._tts._pool.connection(...)`, so its `put()` lands on the pool that was just set aside — where it becomes idle and is never closed either.

## Why this shape

The retire-the-pool approach was the right call when it was written: at the time `invalidate()` swept checked-out connections into `_to_close`, so it would have severed a stream in flight. #7139 changed that — `invalidate()` now retires checked-out connections and closes them when they are returned, and closes idle ones on the next drain.

That is exactly the behaviour the comment here describes wanting ("letting any in-flight stream finish using its connection"), so the plugin can use it directly:

```python
if is_given(model) and model != self._opts.model:
    self._opts.model = model
    self._pool.invalidate()
```

and `_retired_pools` goes away. One pool for the life of the TTS, so a stream always returns its socket to the pool that will close it. Same pattern as asyncai (#7132), neuphonic (#7133) and cartesia (#7140).

The `model != self._opts.model` guard is kept, so re-setting the current model still keeps a warm connection.

## Tests

New `tests/test_plugin_respeecher_tts.py`. The network is replaced by fakes patched onto the `TTS` class, so every pool it builds — including one created after construction — uses them:

- `test_changing_model_closes_the_idle_pooled_connection`
- `test_changing_model_lets_an_in_use_connection_finish`
- `test_same_model_keeps_the_pooled_connection`

Against the original `tts.py`, the first two fail with the leak they describe and the third still passes (it guards against over-invalidating):

```
FAILED test_changing_model_closes_the_idle_pooled_connection
    AssertionError: the old model's idle socket was left open
FAILED test_changing_model_lets_an_in_use_connection_finish
    AssertionError: the returned old-model socket was never closed
2 failed, 1 passed
```

With the change: `3 passed`. `ruff check` and `ruff format --check` clean.

Note for anyone running these locally: they rely on the post-#7139 `ConnectionPool`, so make sure `livekit-agents` resolves to the working tree rather than an installed release.

---

### DIFF
--- a/livekit-plugins/livekit-plugins-respeecher/livekit/plugins/respeecher/tts.py
+++ b/livekit-plugins/livekit-plugins-respeecher/livekit/plugins/respeecher/tts.py
@@ -163,7 +163,6 @@ class TTS(tts.TTS):
             connect_cb=self._connect_ws,
             close_cb=self._close_ws,
         )
-        self._retired_pools: list[utils.ConnectionPool[aiohttp.ClientWebSocketResponse]] = []
 
     @property
     def model(self) -> str:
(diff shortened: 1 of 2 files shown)


Rehansanjay:
@longcw this is the respeecher follow-up to #7139. Now that `invalidate()` leaves in-flight connections alone, respeecher's retired-pool workaround isn't needed, and it was leaking: idle sockets for the old model stayed open, and a stream in flight during a model change returned its socket to the retired pool, where nothing closed it. One file plus tests, CI green, Devin clean, no open threads.

longcw (approved):

PR #7239, opened 2026-09-11
openai: keep GPT-Live's input clock running when no audio is pushed

u9g:
## Problem

GPT-Live speaks only while its input clock runs, and that clock is the audio the client appends. A session with no microphone never appends anything, so every ask times out after 10 seconds with:

```
failed to generate a reply: the model did not start speaking when asked
```

That is what happens under a text simulation, in `lk agent console --text`, and whenever input audio is disabled. #7236 papered over the first case by refusing to start; the other two still hang.

Measured against the live service with the adapter's `generate_reply`: with nothing pushed the ask times out; with 100 ms silence frames pushed at 10 Hz the reply opens in 1.6 s with transcript and audio. `session.input_audio.mute` alone does not do it.

## Change

- `GPTLiveSession` appends 100 ms of silence whenever nothing has been pushed for 200 ms, once the session has started. A live microphone is never padded: the feeder only fills gaps.
- The #7236 guard is removed. A duplex model now runs under a text simulation like any other model.
- Pure refactor first: `push_audio` split into the timestamp and a shared `_append_audio`.

## Tests

`tests/test_gpt_live_model.py`: idle session appends silence; a live microphone is not padded. The fake sockets ignore silent appends by default so the existing sequence assertions stay exact; the two tests that assert on silent input opt back in. The two #7236 tests are removed with the guard. 99 tests pass across the GPT-Live and duplex adapter files.

End to end: a GPT-Live agent under `lk agent simulate` (text mode) now holds full conversations on this branch. Same run on main fails every scenario on turn one.

## What text mode still cannot do well

GPT-Live has no typed user-input channel; the plugin delivers typed turns as `session.commentary.append` with an instruction to reply. In the end-to-end run the voice model delegated to the backend (and so reached the tools) in one of three scenarios, versus three of three in audio mode. Text mode is now usable for exercising conversation flow, not for asserting tool behavior. That limit is the protocol's, not this change's.

agents-js counterpart: https://github.com/livekit/agents-js/pull/2484.

### DIFF
--- a/livekit-agents/livekit/agents/voice/agent_activity.py
+++ b/livekit-agents/livekit/agents/voice/agent_activity.py
@@ -308,14 +308,6 @@ class AgentActivity(RecognitionHooks):
             )
             self._session._warned_realtime_audio_redaction = True
 
-        # a duplex model has no text modality, and the adapter resolves each reply from the audio
-        # the model produces; without audio a text simulation would only time out on turn one
-        if self._text_only and isinstance(self.llm, llm.DuplexRealtimeAdapter):
-            raise RuntimeError(
-                "a DuplexModel speaks only through audio, so it cannot run under a text "
-                "simulation; run `lk agent simulate audio` instead"
-            )
-
         if self._rt_turn_detection_enabled and not self.allow_interruptions:
             raise ValueError(
                 "the RealtimeModel uses a server-side turn detection, "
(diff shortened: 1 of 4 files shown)


longcw:
closing since `generate_reply` cannot really simulate audio input in production

PR #7240, opened 2026-09-11
Only send modern transcriptions to client sdks which can auto back convert modern -> legacy transcriptions internally

1egoman:
The agents part of https://github.com/livekit/client-sdk-js/pull/2093. A high level summary of the transcription deduplication project:
> Today all transcriptions are sent twice over the reliable data channel, once in legacy format, and once in modern data streams format. Because they are sent twice, the reliable data channel is clogged with data and when an end user is on a poor bandwidth network connection (ie, mobile) this can result in a very poor user experience - RPCs get missed, data stream are only partially delivered, etc.

In https://github.com/livekit/client-sdk-js/pull/2093, I added a new client protocol of 3. Any client which advertises this new client protocol version ignores all legacy transcriptions, and back-converts modern transcriptions into in memory legacy transcriptions.

This pull request implements the converse of this behavior on the agent end: if any connected client sdk advertises support for the client protocol of 3, then the agents sdk now will skip sending legacy transcriptions to all participants.

A few important caveats:
- **Note the phrase "all participants" here, rather than just "one participant"** - this was purposeful because filtering by setting `destination_identities` wouldn't only deliver the packet to the listed participants.
  - Reason behind this change: packet destination identity filtering happens client sdk side, so it's impossible today to deliver a packet to only a subset of participants.
  - This does mean this will take slightly longer to roll out (because all clients in a room will need to be updated for this to be enabled) but I think this should happen in practice in fairly short order as users update their sdks across all platforms, especially for those that are performance concious
- **This "all participants" list intentionally excludes a few types of participants, namely SIP, ingress, and egress.** I've looked through the sip, ingress, and egress codebases and none use legacy (or modern for that matter) transcriptions.
  - Reason behind this change: The go sdk still advertises a client protocol of 0, so without this special case any room with (for example) ingress or egress enabled would not pass this check. We're working on changing this in the go sdk but it's a longer lead project (in concert with core services) which shouldn't be expected to be done anytime soon. 
  - This wouldn't block sip, ingress, or egress from using transcriptions in the future, and if they were going to use transcriptions, they'd want to consume modern transcriptions anyway.

### DIFF
--- a/livekit-agents/livekit/agents/utils/participant.py
+++ b/livekit-agents/livekit/agents/utils/participant.py
@@ -8,6 +8,17 @@ from livekit import rtc
 from ..types import ATTRIBUTE_AGENT_NAME
 
 
+def _client_protocol(participant: rtc.Participant) -> int:
+    """``ParticipantInfo.client_protocol``, or 0 when unknown (legacy client).
+
+    Read off the private ``_info`` deliberately: the client protocol is a LiveKit-internal
+    capability signal used for feature detection between participants, not part of the
+    public participant API.
+    """
+    value = getattr(getattr(participant, "_info", None), "client_protocol", 0)
+    return value if isinstance(value, int) else 0
+
+
 async def wait_for_agent(
     room: rtc.Room,
     *,
--- a/livekit-agents/livekit/agents/voice/room_io/_output.py
+++ b/livekit-agents/livekit/agents/voice/room_io/_output.py
@@ -23,9 +23,11 @@ from ...types import (
     ATTRIBUTE_TRANSCRIPTION_FINAL,
     ATTRIBUTE_TRANSCRIPTION_SEGMENT_ID,
     ATTRIBUTE_TRANSCRIPTION_TRACK_ID,
+    CLIENT_PROTOCOL_TRANSCRIPTION_STREAMS,
     TOPIC_TRANSCRIPTION,
     TimedString,
 )
+from ...utils.participant import _client_protocol
 from .. import io
 from ..transcription import find_micro_track_id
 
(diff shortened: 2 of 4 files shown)


livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -366,10 +405,30 @@ async def aclose(self) -> None:
...
 
+    def _should_publish(self) -> bool:
+        needed = _legacy_transcription_needed(self._room)
+        if needed != self._legacy_status_logged:
+            self._legacy_status_logged = needed
+            logger.debug(
+                "legacy transcription publishing %s",
+                "enabled" if needed else "disabled",
+                extra={"participant": self._participant_identity},
+            )
```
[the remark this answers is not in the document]
1egoman: Ah, interesting - should I get rid of this logging behavior all together? Or if not - I'm assuming that in other places participant ids are logged within the framework's logs. Are there any templates which I can follow on how best to handle this?

livekit-agents/livekit/agents/voice/room_io/_output.py
```diff
@@ -275,6 +266,40 @@ async def _forward_audio(self) -> None:
...
+    Only STANDARD participants -- user-created client SDK instances -- are considered. SIP,
+    INGRESS, AGENT (including avatar workers), CONNECTOR and BRIDGE participants never render
+    legacy transcripts, so their client protocol tells us nothing about whether the legacy
+    packet is still necessary. Counting them would also keep legacy publishing alive in every
+    telephony room for good: the Go SDK does not send a client protocol at all, so SIP and
+    INGRESS participants report 0 permanently. EGRESS participants join hidden and never
+    reach ``remote_participants``.
+    """
+    local_identity = room.local_participant.identity
+    for p in room.remote_participants.values():
```
xianshijing-lk, line 292: how often this _legacy_transcription_needed will be called ? 

If it is called frequently, can we improve the code to reduce the overhead ?

xianshijing-lk (commented):
one question, lgtm if you address it.

livekit-agents/livekit/agents/utils/participant.py
```diff
@@ -8,6 +8,17 @@
...
 
 
+def _client_protocol(participant: rtc.Participant) -> int:
+    """``ParticipantInfo.client_protocol``, or 0 when unknown (legacy client).
+
+    Read off the private ``_info`` deliberately: the client protocol is a LiveKit-internal
+    capability signal used for feature detection between participants, not part of the
+    public participant API.
+    """
+    value = getattr(getattr(participant, "_info", None), "client_protocol", 0)
```
xianshijing-lk: any chance that we can add some tests to make sure _info and proto filed never gets renamed or changed ? 
otherwise the value will default to 0 and fail silently.

livekit-agents/livekit/agents/voice/room_io/_output.py
1egoman: This is called on every transcription event, which in practice for most agent interactions will be under 10 times a second. This function does have an `O(n)` loop, but `n` is fairly small (`n` is ~number of remote participants, so for most agent interactions, this will probably be 1), and the loop has an early bail out if client protocol is under 3 for any participant so it's likely `n` will be smaller in practice for larger rooms.

Given this context - Is there something you have in mind here to reduce overhead? I'm not sure there's an obvious lever I am seeing which would cause a significant impact.

livekit-agents/livekit/agents/utils/participant.py
1egoman: I did some thinking about this, and I think in practice actually this was more complicated than it needs to be. All modern participant implementations should contain `._info.client_protocol` as this generated directly from protobuf. So this "fallback to zero of no field exists" logic makes more sense to be handled at the protobuf level.

The minor risk worth mentioning: in theory if somebody were to use an older [`livekit` package version than is mentioned in pyproject.toml](https://github.com/livekit/agents/blob/f69c55c07488fa31640769c272961e19826d02a8/livekit-agents/pyproject.toml#L30), this would throw a hard error. But this seems like a pretty unlikely thing to occur, and even if so, the `client_protocol` field has been around for almost a year at this point.

Given this I've dropped this `_client_protocol` function and inlined all `participant._info.client_protocol` accesses where each previous call site was located in 970231ab5.

PR #7243, opened 2026-09-11
Add Vui TTS plugin (local streaming TTS on CUDA / Apple Silicon)

mogwai:
## What

A `livekit-plugins-vui` TTS plugin for [Vui Nano](https://github.com/fluxions-ai/vui) — a small, context-aware text-to-speech model trained on real conversations (219M active / 305M total parameters, Apache 2.0). It runs **in-process** through the [`vui-tts`](https://pypi.org/project/vui-tts/) package: on an NVIDIA GPU, or on MLX on Apple Silicon. No API key; weights and the shipped voice prompts download from Hugging Face on first use (`prewarm()` does it ahead of the first request).

```python
from livekit.plugins import vui
tts = vui.TTS(voice="maeve")   # maeve / abraham / rhian / harry, a prompt .safetensors, or a .wav to clone
```

## Design

- `synthesize()` renders one text and rewinds the model to the voice prompt.
- `stream()` renders each sentence as the LLM produces it (via `tokenize.basic.SentenceTokenizer`) **in one conversation row**, so the model conditions each sentence on what it just said and prosody carries across the reply. The row is rewound when input ends.
- Everything that touches the engine — load, prefill, decode, rewind — runs on a single worker thread; CUDA graphs and MLX streams are bound to the thread that created them. Renders serialise behind a lock.
- Interruption sets a cancel event the decode loop checks at every frame.

## Verified

On an M4 (MLX, int8): one-shot 5.5 s of audio in 3.1 s (1.8×); streamed input 5.5 s in 2.3 s (2.4×), first audio at 407 ms. Output transcribed with moonshine ASR. `ruff check` / `ruff format` clean; `uv lock` regenerated.

I'm the author of the model (Fluxions AI) and will maintain the plugin.

**Update — CUDA verified too** (RTX 5090, torch 2.11, plain pip install without flash-attn, sharing the GPU with another job): one-shot 6.6 s of audio in 2.7 s (2.4×); streamed 8.2 s in 3.3 s (2.5×) with **first audio at 48 ms**; both 0.0% WER by moonshine. Devin Review flags addressed — see the comment below.

**Scenario verification (exact PR head, MLX on M4 and CUDA on a 5090):** one-shot; a three-sentence streamed reply in one row; a stream interrupted after 6 frames (`aclose()`) followed by a one-shot on the same row; two concurrent `synthesize()` calls (serialised, both complete); a voice switch via `update_options`; and `aclose()` cancelled mid-render then retried (engine released, executor shut down, no live streams). Every rendered wav transcribed by moonshine: 0.0% WER on both backends apart from one ASR-normalisation "OK"/"okay" and one single-word slip in the 40-word streamed reply on CUDA. First audio 48 ms on CUDA, ~400 ms on M4.


mogwai:
Addressed the three Devin Review flags (they're hidden on the PR by the org settings, so summarising here):

- **CUDA rewind retained prior speech** — real, and an engine-level issue: CUDA `Row.rewind()` restored the KV length but left the previous turn in the codec's streaming context (the MLX engine already re-seeded). Fixed upstream in `vui-tts` 1.1.3 (fluxions-ai/vui#38), verified on a 5090: after a rewind the codec context holds exactly the prompt frames and the next turn renders at 0.0% WER. The plugin now requires `vui-tts>=1.1.3`.
- **Concurrent requests shared the row** — real. A `SynthesizeStream` now owns the engine lock from its first sentence to its end, so streams can't interleave sentences or rewind each other's context; `ChunkedStream` takes the lock per call.
- **Transcriptless .wav cloning** — real. It now raises a clear error pointing at a sibling `.txt`, a baked prompt `.safetensors`, or `vui-tts[server]` for automatic transcription, instead of failing on the missing whisper import.

Rebased on current `main` (lock regenerated). Re-verified on an M4: one-shot and streamed synthesis both 0.0% WER by moonshine. CLA: signing.

mogwai:
Two more updates:

- **Shutdown** (the new Devin flag — real): `aclose()` now cancels live streams (tracked in a `WeakSet` like the other streaming plugins), waits for the worker thread off the event loop, closes the row and drops the engine reference, so a session closing mid-sentence no longer leaves a render running or the model reachable. Idempotent.
- **type-check CI**: the failures were untyped-dependency noise (`vui.*` had no `py.typed` — fixed at the source in `vui-tts` 1.1.4 — and `torchcodec.decoders` re-exports `AudioDecoder` without declaring it). Added a mypy override for `vui.*` / `torchcodec.*` alongside the existing ones for `mistralai`, `smithy`, etc. `mypy -p livekit.plugins.vui` passes locally with the CI invocation.

Re-verified after the change: MLX one-shot and streamed synthesis 0.0% WER with `aclose()` at the end of the run; CUDA numbers are in the description.

mogwai:
Fixed the cancelled-shutdown case: `aclose()` no longer sets `_closed` up front. Every step (close streams, close row, shut the executor) is idempotent and the flag is set only once all of them complete, under a lock, so a call cancelled part-way is finished by the next one. Verified with a test that cancels `aclose()` mid-render and retries: the retry completes with the engine released, the executor shut down and no live streams; a third call is a no-op. mypy and ruff clean; normal synthesis re-verified after the change.

PR #7244, opened 2026-09-12
fix: keep delayed duplex transcripts on the original audio burst

### LINKED ISSUE #7227: DuplexRealtimeAdapter loses delayed transcripts or attaches them to the next audio burst
### Bug Description

`DuplexRealtimeAdapter` only attaches queued transcript fragments from `_on_audio_frame()` while the audio gate is open. A fragment describing audio that has already arrived can therefore miss its original burst:

- If the last audible frame arrives, then its transcript, then the burst closes, the original burst closes without that text or an assistant chat item.
- If the burst closes before its transcript arrives, the next audible burst can claim the old text. Its first fragment establishes a fresh anchor, even though that fragment describes the previous burst.

Both event orders reproduce on upstream `main` at `bbfcbcebe9eeac16120ddffdc166d55f59295211`. The relevant paths are [`_on_audio_frame`, `_close_burst`, and `_on_transcript_delta`](https://github.com/livekit/agents/blob/bbfcbcebe9eeac16120ddffdc166d55f59295211/livekit-agents/livekit/agents/llm/duplex_adapter.py#L375-L470).

### Expected Behavior

A delayed transcript for received audio should remain associated with the original utterance/message ID. It should not require another audible frame and must not be reassigned to a subsequent utterance. Handling late text must also preserve the framework's rule that interrupted/unplayed speech is not published as spoken text.

### Reproduction Steps

Save this as `tests/test_duplex_transcript_issue.py` in the Agents repository. It reuses the existing hermetic duplex fixtures and drives the adapter's handlers directly so the event order is deterministic; no provider connection is involved.

```python
import pytest

from livekit.agents.llm.duplex_adapter import FixedGate

from .test_duplex_adapter import _FakeDuplexModel, _frame

pytestmark = pytest.mark.unit


@pytest.mark.parametrize("next_burst", [False, True])
async def test_delayed_transcript_keeps_its_original_audio(next_burst):
    session = llm.DuplexRealtimeAdapter(
        _FakeDuplexModel(), gate=lambda: FixedGate(0.001)
    ).session()
    try:
        session._on_audio_frame(llm.DuplexAudioFrame(_frame(0.3), start_ms=0))
        first_id = session._burst.id
        if next_burst:
            session._close_burst()
        session._on_transcript_delta(
            llm.DuplexOutputTranscriptDelta("First utterance.", start_ms=0, end_ms=100)
        )
        if next_burst:
            session._on_audio_frame(llm.DuplexAudioFrame(_frame(0.3), start_ms=1000))
            assert session._burst.transcript == "", "old text was attached to the next burst"
        else:
            session._close_burst()
        first = session.chat_ctx.get_by_id(first_id)
        assert first is not None and first.text_content == "First utterance."
    finally:
        await session.aclose()
```

Run `uv run pytest --unit -q tests/test_duplex_transcript_issue.py`.

Observed: **2 failed**. With `next_burst=False`, `first` is `None`. With `next_burst=True`, the new burst's transcript is `"First utterance."`.

### Environment

- macOS 26.6.2, Python 3.13.12
- Agents core loaded directly from the upstream commit above (1.8.1 source)
- `livekit==1.1.18`, `livekit-api==1.2.0`
- Model: the repository's `_FakeDuplexModel`; applicable to asynchronous duplex audio/transcript delivery, including GPT Live

### Proposed Direction

Retain enough segment identity and timing to associate delayed text with its original audio, and attach eligible text when the transcript arrives if its audio is already known. Closed bursts need an explicit finalization policy so late text cannot silently migrate to a new message. The exact policy for updating a closed text stream/chat item should be agreed with the framework's playout and interruption handling before implementing the fix.

cpruijsen:
## Summary

Delayed duplex transcripts attach to the burst that already played that audio, including after the gate has closed. Late text is recorded under the original message id and is not used to open the next burst.

`DuplexRealtimeAdapter` only attached queued transcript fragments from `_on_audio_frame` while the audio gate was open. A fragment describing audio that had already arrived could miss its utterance:

- Last audible frame, then transcript, then close: the original burst closed with no text and no assistant chat item.
- Burst closed before the transcript arrived: the next audible burst claimed the old text as its first fragment and established a fresh span/audio offset from it.

Both orders are in #7227. A third order (next burst already open, then the delayed first fragment) also attached the old text to the new utterance.

The adapter attaches due fragments when they arrive and when the burst closes, same due/ahead rule as before. A closed burst is remembered so late text that still sits on that audio range can be recorded on it.

A closed burst with no offset yet only claims a span in `[audio_start - lead, audio_end]`. The extra lead past `audio_end` would include the next utterance's onset after a normal gated pause. Unplayed spans (audio this burst never reached) are still not published on that item.

## Decision

Late eligible text is attached to the closed burst's chat item; `text_ch` is not kept open after audio ends. Keeping the text channel open until a grace timeout or the next burst would leave `forward_generation` waiting on a stream that usually never gets another fragment. Can switch if the stream should stay open instead.

A closed burst without an offset may still take a first fragment after the next burst has opened, but only if that span sits on audio it already forwarded (no extra lead past `audio_end`). Freezing identity once the next burst opens still gives the next burst the previous utterance. A ±300 ms window around the whole closed range would claim the next utterance after a real silence gap.

## Test plan

- [x] `tests/test_duplex_adapter.py`: delayed transcript after the last frame still joins that burst (including mismatched span/audio clocks while the burst is still open).
- [x] Reporter's two orders plus the sibling (`when=before_close/in_gap/after_next`): original id keeps the text; the next burst does not.
- [x] Stream-level: transcript after the burst has closed via silence does not become the next burst's first fragment.
- [x] Leading fragment after an untranscribed burst, stamped at the actual gap clock, still waits for the next utterance.
- [x] Existing adapter suite (54 unit tests), including rollover, unclaimed timeout, reconnect, and function-call generations.

Fixes #7227

### DIFF
--- a/livekit-agents/livekit/agents/llm/duplex_adapter.py
+++ b/livekit-agents/livekit/agents/llm/duplex_adapter.py
@@ -192,6 +192,10 @@ class _Burst:
     audio_ch: aio.Chan[rtc.AudioFrame]
     audio_start_ms: int
     """Where on the adapter's audio clock the burst opened."""
+    audio_end_ms: int
+    """Last audio clock this burst covers, frozen when it closes."""
+    carries_message: bool = True
+    """False for a function-call generation that never spoke."""
     anchor_ms: int | None = None
     """Span clock minus audio clock, fixed by the first fragment: the sound that opened the gate
     and the oldest unclaimed fragment describe the same moment."""
@@ -307,6 +311,7 @@ class _DuplexRealtimeSession(RealtimeSession):
         self._duplex = duplex
         self._gate = gate
         self._burst: _Burst | None = None
+        self._last_burst: _Burst | None = None
         self._audio_timeout = audio_timeout
         # the adapter's clock: output audio heard so far, which arrives at playback pace
         self._audio_ms = 0
@@ -382,17 +387,8 @@ class _DuplexRealtimeSession(RealtimeSession):
             if not burst.audio_ch.closed:
                 burst.audio_ch.send_nowait(f.frame)
             self._audio_ms += round(f.frame.duration * 1000)
-
-            # the first fragment anchors the span clock to the audio clock; a later one is due when
-            # the audio reaches its span, and one this burst never reaches waits for the next
-            while self._fragments:
-                fragment = self._fragments[0]
-                if fragment.start_ms is not None:
-                    if burst.anchor_ms is None:
-                        burst.anchor_ms = fragment.start_ms - burst.audio_start_ms
-                    if fragment.start_ms - burst.anchor_ms > self._audio_ms + _ATTACH_LEAD_MS:
-                        break
-                burst.attach(self._fragments.popleft())
+            burst.audio_end_ms = self._audio_ms
+            self._drain_fragments()
             return
 
         self._audio_ms += round(f.frame.duration * 1000)
@@ -418,6 +414,8 @@ class _DuplexRealtimeSession(RealtimeSession):
             text_ch=aio.Chan(),
             audio_ch=aio.Chan(),
             audio_start_ms=self._audio_ms,
+            audio_end_ms=self._audio_ms,
+            carries_message=message,
         )
         ev = GenerationCreatedEvent(
             message_stream=burst.message_ch,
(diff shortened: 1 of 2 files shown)


longcw:
thanks for the pr! closing in favor of https://github.com/livekit/agents/pull/7264

PR #7245, opened 2026-09-12
fix(liveavatar): honor agent.state_updated for speaking and barge-in

### LINKED ISSUE #7232: liveavatar plugin doesn't recognize agent.audio_buffer_* / agent.state_updated events from the LiveAvatar server (v1.8.0)
### Bug Description

The liveavatar plugin's WebSocket event handler (`avatar.py's _recv_task`) only recognizes 4 event types: `session.state_updated, agent.speak_started, agent.speak_ended, agent.speak_interrupted`. During a real session, the LiveAvatar server sends different event types that all fall through to the "unhandled" branch and are silently ignored:

```
agent.audio_buffer_appended   (repeated many times per response)
agent.audio_buffer_committed
agent.state_updated
agent.audio_buffer_cleared
```
Only the very first "connected" message on session start used the schema the plugin expects (session.state_updated); every event during the actual conversation used this different, unrecognized schema instead.

As a result, _handle_agent_speak_started and _handle_agent_speak_ended never fire during a real conversation, so self._avatar_speaking / self._avatar_interrupted never update, and QueueAudioOutput.notify_playback_started() / notify_playback_finished() are never called from server acknowledgment. This most directly affects _on_clear_buffer's barge-in logic, which checks if self._avatar_speaking: before sending agent.interrupt — since that flag never becomes True, interrupting the avatar mid-response may not correctly signal LiveAvatar's server to stop.

### Expected Behavior

The plugin should recognize whatever event schema the LiveAvatar server currently sends, so _avatar_speaking and playback-start/finish notifications correctly reflect the avatar's real state, and barge-in/interruption works as intended.

### Reproduction Steps

```bash
1. Set up a LiveKit agent using livekit.plugins.liveavatar.AvatarSession (is_sandbox=True), following the standard pattern (AgentSession with any STT/LLM/TTS, await avatar.start(session, room), then await session.start(agent=..., room=...)).
2. Set the livekit.plugins.liveavatar logger to DEBUG.
3. Connect via the Agents Playground and have a short back-and-forth conversation.
4. Observe the log — every event past the initial "connected" message logs as Unhandled LiveAvatar event: agent.audio_buffer_appended / agent.audio_buffer_committed / agent.state_updated / agent.audio_buffer_cleared.
```

### Operating System

Operating System: macOS 26.6.2 (Darwin 25.6.0, arm64)

### Models Used

STT: Deepgram (default model) LLM: google/gemini-3-flash via LiveKit Inference TTS: cartesia/sonic-3 via LiveKit Inference Avatar: HeyGen LiveAvatar, LITE mode, sandbox

### Package Versions

```bash
- livekit-plugins-liveavatar: 1.8.0
- livekit-agents: 1.8.0
```

### Session/Room/Call IDs

room: console-320ff264
job_id: AJ_ogpRGmyexCVC

### Proposed Solution

```python
Update _recv_task's event dispatch in avatar.py to also handle agent.audio_buffer_appended, agent.audio_buffer_committed, agent.audio_buffer_cleared, and agent.state_updated, mapping them to the equivalent _avatar_speaking/playback-notification behavior currently tied to agent.speak_started/agent.speak_ended. Since the exact intended semantics of the new event names aren't publicly documented, this likely needs input from LiveKit/HeyGen on the current protocol spec before the mapping can be written correctly.
```

### Additional Context

Discovered while investigating a separate, still-unconfirmed report that the avatar appears static (no movement) during a live conversation despite the avatar correctly being told it's in a "listening" state (via agent.start_listening, which is triggered client-side independent of this bug). This event-handling gap doesn't confirm or rule out that issue, but is a real, separate compatibility gap worth fixing regardless.

_No response_

cpruijsen:
## Summary

`agent.state_updated` with `new_state == "talking"` follows the existing speak-started path; leaving talking (`idle` / `listening`) follows speak-ended. `speak_*` still work and are idempotent if a session emits both. Buffer append/commit are treated as command acks, not playback lifecycle. `agent.audio_buffer_cleared` is treated as the interrupt ack. `error` / `warning` are logged at the matching level.

LiveAvatar LITE emits `agent.state_updated` (`idle` / `listening` / `talking`) and `agent.audio_buffer_*` acks during a conversation. The plugin only handled `session.state_updated` and `agent.speak_*`, so `_avatar_speaking` never became true, playback start/finish were never notified from the server, and barge-in skipped `agent.interrupt`.

## Decision

Speaking and playback are driven from `agent.state_updated`, not `audio_buffer_appended` → started and `audio_buffer_committed` → ended (the mapping sketched in the issue). The [LITE events spec](https://docs.liveavatar.com/docs/lite-mode/events) defines buffer events as command acks and `agent.state_updated` as the avatar speaking signal; the reporter also observed `agent.state_updated` on a real sandbox session. Mapping commit to "playback finished" would end playout on the `speak_end` ack, which can fire while the avatar is still talking. Can switch if append/commit should own playback instead.

Fixes #7232

## Test plan

- [x] Unit tests in `tests/test_plugin_liveavatar.py` (`pytest.mark.unit`): `agent.state_updated` talking/listening updates `_avatar_speaking` and playback notifications; buffer append/commit do not; speak_started + talking is one playback-started; barge-in sends `agent.interrupt` after talking and does not after append-only.
- [x] Those tests fail without the dispatch change (5 failed, 4 passed) and pass with it, plus existing `livekit-plugins/livekit-plugins-liveavatar/tests/test_api.py`.
- [ ] Live LITE/sandbox conversation with `LIVEAVATAR_API_KEY` / `LIVEAVATAR_AVATAR_ID`: DEBUG should no longer print Unhandled for `agent.state_updated` / `agent.audio_buffer_*`, and interrupting mid-reply should send `agent.interrupt`.

### DIFF
--- a/livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/avatar.py
+++ b/livekit-plugins/livekit-plugins-liveavatar/livekit/plugins/liveavatar/avatar.py
@@ -329,20 +329,7 @@ class AvatarSession(BaseAvatarSession):
                             message="LiveAvatar connection closed unexpectedly."
                         )
                 event = json.loads(msg.data)
-                event_type = event.get("type")
-                if event_type == "session.state_updated":
-                    state = event.get("state")
-                    logger.debug(f"LiveAvatar session state: {state}")
-                    if state == "connected":
-                        self._session_connected.set()
-                elif event_type == "agent.speak_interrupted":
-                    self._handle_agent_speak_interrupted(event)
-                elif event_type == "agent.speak_ended":
-                    self._handle_agent_speak_ended(event)
-                elif event_type == "agent.speak_started":
-                    self._handle_agent_speak_started(event)
-                else:
-                    logger.debug(f"Unhandled LiveAvatar event: {event_type}")
+                self._handle_server_event(event)
 
         io_tasks = [
             asyncio.create_task(_forward_audio(), name="_forward_audio_task"),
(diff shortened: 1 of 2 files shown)


tinalenguyen (approved):
thanks for the PR!

PR #7246, opened 2026-09-12
fix(anthropic): update prefill regex for Claude 4.7+ and guard trailing tool_use (#7217)

### LINKED ISSUE #7217: Anthropic plugin: hardcoded _NO_PREFILL_PATTERNS misses Claude 4.7/4.8/5, causing HTTP 400 "does not support assistant message prefill"
**Version:** `livekit-agents` 1.8.1, `livekit-plugins-anthropic` 1.8.1 (verified against the released wheels; also present unchanged in 1.5.1, 1.6.8 and 1.7.1)

Related: #6868 (open, covers the model-list half; the tool-pairing caveat below is not addressed there), #6007 (same gap in the AWS plugin), and #5254 / #6213 (both closed, about the content of the injected message rather than which models get one).

### Summary

`livekit-plugins-anthropic` decides whether to append a trailing user message using a hardcoded tuple of two model-name prefixes. Anthropic has since removed assistant-message prefill from a much larger set of models. Any of those newer models therefore takes the "prefill is fine" path, and every request whose context ends on an assistant turn fails with a non-retryable HTTP 400.

### Framework behaviour

`livekit/plugins/anthropic/llm.py:40-46`:

```python
# Claude 4.6+ no longer supports prefilling (trailing assistant messages).
_NO_PREFILL_PATTERNS = ("claude-sonnet-4-6", "claude-opus-4-6")


def _model_disables_prefill(model: str) -> bool:
    """Return True if the model does not support assistant message prefilling."""
    return any(model.startswith(p) for p in _NO_PREFILL_PATTERNS)
```

`livekit/plugins/anthropic/llm.py:220-224`:

```python
# Claude 4.6+ does not support prefilling (trailing assistant messages).
inject_trailing = _model_disables_prefill(self._opts.model)
anthropic_ctx, extra_data = chat_ctx.to_provider_format(
    format="anthropic", inject_trailing_user_message=inject_trailing
)
```

`livekit/agents/llm/_provider_format/anthropic.py:105-108`:

```python
# Claude 4.6+ does not support prefilling (trailing assistant messages).
# Append a dummy user message so the request ends with a user turn.
if inject_trailing_user_message and messages and messages[-1]["role"] == "assistant":
    messages.append({"role": "user", "content": [{"text": ".", "type": "text"}]})
```

The comment says "Claude 4.6+", but the tuple enumerates exactly two 4.6 model ids by literal prefix. Anthropic documents prefill as unsupported on Opus 4.6, Sonnet 4.6, Opus 4.7, Opus 4.8, Opus 5, Sonnet 5, and Fable 5 — the tuple covers two of the seven. `models.py` likewise stops at `claude-opus-4-6`, so newer ids are only reachable via the `str` half of `model: str | ChatModels`, where nothing flags the mismatch.

This tuple is byte-identical in 1.5.1, 1.6.8, 1.7.1 and 1.8.1 — it has not been updated as new models shipped.

### Minimal reproduction

No network call needed to see the wrong branch taken:

```python
from livekit.plugins.anthropic.llm import _model_disables_prefill

print(_model_disables_prefill("claude-opus-4-6"))  # True
print(_model_disables_prefill("claude-opus-4-7"))  # False  <-- should be True
print(_model_disables_prefill("claude-opus-4-8"))  # False  <-- should be True
print(_model_disables_prefill("claude-opus-5"))    # False  <-- should be True
print(_model_disables_prefill("claude-sonnet-5"))  # False  <-- should be True
```

End-to-end, against the API:

```python
from livekit.plugins.anthropic import LLM

ctx = llm.ChatContext()
ctx.add_message(role="user", content="hello")
ctx.add_message(role="assistant", content="I was in the middle of")  # trailing assistant

model = LLM(model="claude-opus-4-7")
async for _ in model.chat(chat_ctx=ctx):   # HTTP 400
    pass
```

Swap the model id to `claude-opus-4-6` and the same context succeeds, because the trailing user message is injected.

### Expected vs actual

**Expected:** for any Anthropic model that does not support assistant prefill, `to_provider_format` is called with `inject_trailing_user_message=True`, so a context ending on an assistant turn is made legal before dispatch.

**Actual:** for every model outside the two hardcoded 4.6 prefixes, `inject_trailing=False`. The array is sent ending on an assistant message and the API returns HTTP 400 (`This model does not support assistant message prefill`). Because it is a 400, the SDK does not retry — the turn is lost. In a realtime voice agent this surfaces as a dropped response mid-conversation.

This is easy to hit in normal `Agent` use: `generate_reply` after a tool call, a greeting emitted after an assistant history item, or any background-triggered reply that copies the context while an assistant turn is trailing.

### Suggested fix

Replace the literal-prefix tuple with a pattern that covers the family, so new point releases do not silently regress:

```python
import re

_NO_PREFILL_RE = re.compile(r"^claude-(?:sonnet|opus|haiku|fable)-(?:4-[6-9]|5)")


def _model_disables_prefill(model: str) -> bool:
    return bool(model) and bool(_NO_PREFILL_RE.match(model))
```

This keeps 4.5 and earlier on the prefill-capable path, where prefill is still valid. Extending `ChatModels` in `models.py` to include the current ids would also make the gap visible at type-check time.

### One caveat on the fix — trailing tool calls need separate handling

`_provider_format/anthropic.py:46-47` maps a `function_call` item to `role="assistant"` and emits it at `:66-74` as a `tool_use` block. So a context whose last item is a `function_call` with no matching `function_call_output` also serialises to an array ending on an assistant message — and `inject_trailing_user_message` will happily append a plain-text user message after it.

That produces a *different* 400. Anthropic's documentation states:

> "Tool result blocks must immediately follow their corresponding tool use blocks in the message history. You cannot include any messages between the assistant's tool use message and the user's tool result message."

and names the resulting error:

> "tool_use ids were found without tool_result blocks immediately after"

So widening the model list alone converts a prefill 400 into a tool-pairing 400 for contexts with an in-flight tool call. Suggested guard, in `to_chat_ctx` before the injection at `:107`:

```python
if inject_trailing_user_message and messages and messages[-1]["role"] == "assistant":
    last_content = messages[-1].get("content") or []
    has_unresolved_tool_use = any(
        isinstance(b, dict) and b.get("type") == "tool_use" for b in last_content
    )
    if not has_unresolved_tool_use:
        messages.append({"role": "user", "content": [{"text": ".", "type": "text"}]})
```

Callers would still need to drop or resolve the orphan `tool_use` themselves, but this at least stops the framework from constructing an array that is invalid for a second, less obvious reason. Happy to open a PR for either half.

rrfunde:
Fixes #7217

### DIFF
--- a/livekit-agents/livekit/agents/llm/_provider_format/utils.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/utils.py
@@ -144,10 +144,14 @@ def group_tool_calls(chat_ctx: llm.ChatContext) -> list[_ChatItemGroup]:
         call_id_to_group[tool_output.call_id].add(tool_output)
 
     # validate that each group and remove invalid tool calls and tool outputs
-    for group in item_groups.values():
+    groups = list(item_groups.values())
+    for i, group in enumerate(groups):
+        # Trailing tool calls in the last group without outputs are unresolved/pending
+        if i == len(groups) - 1 and group.tool_calls and not group.tool_outputs:
+            continue
         group.remove_invalid_tool_calls()
 
-    return list(item_groups.values())
+    return groups
 
 
 @dataclass
--- a/livekit-agents/livekit/agents/llm/_provider_format/utils.py
+++ b/livekit-agents/livekit/agents/llm/_provider_format/utils.py
@@ -92,7 +92,9 @@ def convert_mid_conversation_instructions(
     return llm.ChatContext(items)
 
 
-def group_tool_calls(chat_ctx: llm.ChatContext) -> list[_ChatItemGroup]:
+def group_tool_calls(
+    chat_ctx: llm.ChatContext, *, allow_unresolved_tail: bool = False
+) -> list[_ChatItemGroup]:
     """Group chat items (messages, function calls, and function outputs)
     into coherent groups based on their item IDs and call IDs.
 
@@ -143,11 +145,16 @@ def group_tool_calls(chat_ctx: llm.ChatContext) -> list[_ChatItemGroup]:
 
         call_id_to_group[tool_output.call_id].add(tool_output)
 
-    # validate that each group and remove invalid tool calls and tool outputs
+    # validate each group and remove invalid tool calls and tool outputs
     groups = list(item_groups.values())
     for i, group in enumerate(groups):
         # Trailing tool calls in the last group without outputs are unresolved/pending
-        if i == len(groups) - 1 and group.tool_calls and not group.tool_outputs:
+        if (
+            allow_unresolved_tail
+            and i == len(groups) - 1
+            and group.tool_calls
+            and not group.tool_outputs
+        ):
             continue
         group.remove_invalid_tool_calls()
 
--- a/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
+++ b/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 import os
+import re
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
 from typing import Any, Literal, cast
@@ -38,12 +39,12 @@ from .models import ChatModels
 from .utils import CACHE_CONTROL_EPHEMERAL
 
 # Claude 4.6+ no longer supports prefilling (trailing assistant messages).
-_NO_PREFILL_PATTERNS = ("claude-sonnet-4-6", "claude-opus-4-6")
+_NO_PREFILL_PATTERN = re.compile(r"^claude-(?:sonnet|opus|haiku|fable)-(?:4-[6-9]|[5-9])")
 
 
 def _model_disables_prefill(model: str) -> bool:
     """Return True if the model does not support assistant message prefilling."""
-    return any(model.startswith(p) for p in _NO_PREFILL_PATTERNS)
+    return bool(_NO_PREFILL_PATTERN.match(model))
 
 
 @dataclass
--- a/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/models.py
+++ b/livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/models.py
@@ -14,4 +14,13 @@ ChatModels = Literal[
     "claude-opus-4-20250514",
     "claude-opus-4-1-20250805",
     "claude-opus-4-6",
+    "claude-opus-4-7",
+    "claude-opus-4-8",
+    "claude-opus-5",
+    "claude-sonnet-4-7",
+    "claude-sonnet-4-8",
+    "claude-sonnet-5",
+    "claude-haiku-4-6",
+    "claude-haiku-4-7",
+    "claude-haiku-5",
 ]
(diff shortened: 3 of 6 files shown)


rrfunde:
Addressed feedback: scoped unresolved tool call retention to Anthropic via `allow_unresolved_tail=True` in `group_tool_calls`, keeping the default behavior unchanged for all other providers.

livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/models.py
```diff
@@ -14,4 +14,13 @@
     "claude-opus-4-20250514",
     "claude-opus-4-1-20250805",
     "claude-opus-4-6",
+    "claude-opus-4-7",
+    "claude-opus-4-8",
+    "claude-opus-5",
+    "claude-sonnet-4-7",
+    "claude-sonnet-4-8",
+    "claude-sonnet-5",
+    "claude-haiku-4-6",
```
longcw: no all these models are real, could you clean them up?

livekit-agents/livekit/agents/llm/_provider_format/utils.py
```diff
@@ -92,7 +92,9 @@ def convert_mid_conversation_instructions(
     return llm.ChatContext(items)
 
 
-def group_tool_calls(chat_ctx: llm.ChatContext) -> list[_ChatItemGroup]:
+def group_tool_calls(
+    chat_ctx: llm.ChatContext, *, allow_unresolved_tail: bool = False
```
longcw: I don't think `allow_unresolved_tail` is needed.

`group_tool_calls` drops an unresolved function call with a warning, so the array never ends on a tool_use block.

livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/models.py
rrfunde: Cleaned up in 2a287f94a: reverted ChatModels to only include released models, relying on _NO_PREFILL_PATTERN to dynamically match newer models passed as strings.

livekit-agents/livekit/agents/llm/_provider_format/utils.py
rrfunde: Addressed in 2a287f94a: removed allow_unresolved_tail and reverted group_tool_calls back to its original implementation, relying on standard unresolved tool call pruning.

livekit-plugins/livekit-plugins-anthropic/livekit/plugins/anthropic/llm.py
```diff
@@ -38,12 +39,12 @@
 from .utils import CACHE_CONTROL_EPHEMERAL
 
 # Claude 4.6+ no longer supports prefilling (trailing assistant messages).
-_NO_PREFILL_PATTERNS = ("claude-sonnet-4-6", "claude-opus-4-6")
+_NO_PREFILL_PATTERN = re.compile(r"^claude-(?:sonnet|opus|haiku|fable)-(?:4-[6-9]|[5-9])")
```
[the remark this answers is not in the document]
rrfunde: Addressed in f81ea2f6d:
- Updated `_model_disables_prefill` to parse major and minor version numbers as integers and compare against `(4, 6)`.
- Handled multi-digit versions (`claude-sonnet-4-10`, `claude-sonnet-10`) as well as 8-digit date suffixes (`20260101`).
- Added multi-digit version test cases to `tests/test_plugin_anthropic.py`.

PR #7247, opened 2026-09-12
test(google): disarm GenAI client finalizers to prevent task leak across unit tests (#6881)

### LINKED ISSUE #6881: Flaky unit-tests: a genai client finalizer lands in an unrelated test's leak check

rrfunde:
## Summary
- Disarm `google.genai.client.AsyncClient.__del__` and `BaseApiClient.__del__` in `tests/test_plugin_google_realtime.py` so that asynchronous client finalizers do not schedule `aclose()` tasks on subsequent unrelated tests' event loops during GC.
- Add `gc.collect()` and task drain in `_make_session` and `_make_configured_session`.
- Fix flag parsing in `tests/conftest.py` so boolean category flags (like `--unit`) do not swallow subsequent test file arguments.

Fixes #6881

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -57,19 +57,36 @@ _CATEGORY_HINT = (
 )
 
 
+def _parameterized_categories() -> frozenset[str]:
+    block = re.search(r"(?ms)^markers\s*=\s*\[(.*?)^\s*\]", _PYPROJECT.read_text(encoding="utf-8"))
+    names = re.findall(r'^\s*"(\w+)\(', block.group(1), re.MULTILINE) if block else []
+    return frozenset(names)
+
+
+PARAMETERIZED_CATEGORIES = _parameterized_categories()
+
+
 def pytest_addoption(parser: pytest.Parser) -> None:
     _add_realtime_option(parser)
     group = parser.getgroup("categories", "test category selection")
     for category in CATEGORIES:
-        group.addoption(
-            f"--{category}",
-            nargs="?",
-            const=True,
-            default=None,
-            metavar="PROVIDER",
-            help=f"select only `{category}` tests; optionally narrow to a single "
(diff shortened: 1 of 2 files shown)


rrfunde:
Addressed feedback: reverted the global GenAI check in `conftest.py` and scoped finalizer disarming to an autouse pytest fixture using `monkeypatch` in `test_plugin_google_realtime.py` so that other tests and global leak checks are completely unaffected.

PR #7248, opened 2026-09-12
fix(sarvam): map wav output codec to audio/pcm to prevent missing RIFF error (#5267)

### LINKED ISSUE #5267: [sarvam tts] output_audio_codec="wav" causes "Invalid WAV file: missing RIFF/WAVE" error
### Bug Description

_Note: This occurred a few times and I have not been able to reproduce it consistently. Filing anyway as the root cause is clear from the traceback and source code inspection._

When using the Sarvam TTS plugin with output_audio_codec="wav", the Sarvam API returns raw PCM bytes instead of a proper WAV container file. The plugin sets mime_type to audio/wav, which causes LiveKit's internal decoder to expect a RIFF/WAVE header — but since none is present, it crashes immediately on the first audio chunk.

### Expected Behavior

Either the plugin should prepend a valid RIFF/WAVE header to the raw PCM bytes returned by Sarvam before pushing to the emitter, or wav should be removed from ALLOWED_OUTPUT_AUDIO_CODECS / raise a clear error explaining it is unsupported in this context.

### Reproduction Steps

```bash
import sarvam

OurTTS = sarvam.TTS(
    target_language_code="hi-IN",
    model="bulbul:v3",
    speaker="ritu",
    pace=1.0,
    temperature=0.8,
    output_audio_codec="wav"  # triggers the bug
)


Run any agent that synthesizes speech using the above TTS instance.
```

### Operating System

Fedora Linux

### Models Used

TTS: Sarvam bulbul:v3

### Package Versions

```bash
livekit-agents — latest from GitHub
livekit-plugins-sarvam — latest from GitHub
Python 3.13
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python
Prepend a RIFF/WAVE header to the raw PCM data when output_audio_codec == "wav" before pushing to the emitter. This needs to be done in two places:

ChunkedStream._run (non-streaming path)
SynthesizeStream._handle_audio_message (WebSocket streaming path)


Workaround: Use output_audio_codec="mp3" — Sarvam returns a proper mp3 bitstream for this codec and it works correctly end-to-end.
```

### Additional Context

Additional Context
The issue is that wav is listed in ALLOWED_OUTPUT_AUDIO_CODECS in tts.py and accepted without error, giving users no indication it will fail at runtime. The Sarvam API itself returns raw PCM (no container) for the wav format. This is a Sarvam API behaviour that the plugin needs to account for.

Image

rrfunde:
## Summary
When using `output_audio_codec="wav"`, the Sarvam TTS API returns raw PCM bytes rather than a RIFF/WAVE container file.
Mapping `"wav"` to `"audio/pcm"` in `_CODEC_TO_MIME` allows `AudioEmitter` to ingest the raw PCM directly without failing with `ValueError: Invalid WAV file: missing RIFF/WAVE`.

Fixes #5267

### DIFF
--- a/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
+++ b/livekit-plugins/livekit-plugins-sarvam/livekit/plugins/sarvam/tts.py
@@ -26,6 +26,7 @@ import enum
 import json
 import os
 import platform
+import struct
 import weakref
 from dataclasses import dataclass, replace
 from typing import Literal
@@ -155,6 +156,28 @@ def _decode_telephony(codec: str, data: bytes) -> bytes:
         raise ValueError(f"_decode_telephony does not support codec: {codec}")
     pcm = table[np.frombuffer(data, dtype=np.uint8)]
     return pcm.astype("<i2").tobytes()
+
+
+def _extract_wav_pcm(audio_bytes: bytes) -> bytes:
+    """If audio_bytes starts with a RIFF/WAVE container, extract the raw PCM data.
+
+    Parses the RIFF chunk structure to locate the 'data' chunk instead of
+    assuming a fixed 44-byte header, handling metadata or extended format chunks.
+    """
+    if len(audio_bytes) < 12 or not audio_bytes.startswith(b"RIFF") or audio_bytes[8:12] != b"WAVE":
+        return audio_bytes
+
+    pos = 12
+    while pos + 8 <= len(audio_bytes):
+        chunk_id = audio_bytes[pos : pos + 4]
+        chunk_size = struct.unpack("<I", audio_bytes[pos + 4 : pos + 8])[0]
+        pos += 8
+        if chunk_id == b"data":
+            end = pos + chunk_size
+            return audio_bytes[pos:end] if end <= len(audio_bytes) else audio_bytes[pos:]
+        pos += chunk_size + (chunk_size % 2)
+
+    return audio_bytes[44:] if len(audio_bytes) >= 44 else audio_bytes
 
 
 # Supported languages in BCP-47 format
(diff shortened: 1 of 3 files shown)


rrfunde:
Addressed feedback:
- Restored `_CODEC_TO_MIME["wav"] = "audio/wav"` so REST `synthesize()` continues to parse full WAV files with RIFF headers.
- Specifically mapped `output_audio_codec == "wav"` to `audio/pcm` inside `SynthesizeStream._run()` for the WebSocket streaming path.
- In `SynthesizeStream._handle_audio_message()`, added a check to strip the 44-byte RIFF container header if the server happens to send RIFF-wrapped chunks on the WebSocket, ensuring `AudioEmitter(audio/pcm)` always receives clean PCM samples.

rrfunde:
Addressed review feedback: implemented `_extract_wav_pcm()` to parse the RIFF container chunk by chunk and extract the payload of the `data` chunk, handling files with extended `fmt` chunks or metadata (`LIST`, `JUNK`) rather than slicing at a fixed 44-byte offset. Added unit tests for extended chunks.

rrfunde:
Addressed review feedback: preserved WAV parser state across WebSocket chunks in `SynthesizeStream` via `_extract_streaming_wav_pcm()` and `_locate_data_chunk()`, buffering initial headers across messages until the `data` chunk is identified, correctly handling split headers and variable-length metadata before streaming PCM. Added unit test for split headers.

rrfunde:
Addressed review feedback:
- Buffered up to 4 bytes when testing for the `RIFF` signature to avoid classifying split WAV signatures (e.g. `b"RI"`) as raw PCM.
- Preserved the parsed `data` chunk length across WebSocket messages and capped emitted samples to the declared size, discarding trailing metadata chunks (such as `LIST` or `INFO`) and handling transitions to subsequent RIFF containers.
- Added comprehensive unit tests covering partial RIFF buffering, raw PCM with coincident `RI` prefixes, and trailing metadata chunk discarding both within and across WebSocket message boundaries.

rrfunde:
Addressed review feedback:
- Tracked the outer RIFF container size via `_wav_container_remaining` derived from `file_size` in the RIFF header.
- Safely consumed and skipped trailing container metadata (e.g. `LIST`, `JUNK`) up to the exact container boundary without scanning for `b"RIFF"`, preventing payload content containing `b"RIFF"` from triggering false container boundaries.
- Added unit tests verifying that trailing metadata chunks containing `b"RIFF"` are ignored and consecutive containers are properly extracted.

PR #7255, opened 2026-09-12
fix(openai): pass response_id in ResponseCancelEvent on interrupt

### LINKED ISSUE #5564: RealtimeSession.interrupt() is a no-op without response_id under concurrent in-flight responses (OpenAI plugin)

rrfunde:
## Summary

`RealtimeSession.interrupt()` in the OpenAI Realtime plugin was emitting:
```python
self.send_event(ResponseCancelEvent(type="response.cancel"))
```
without specifying `response_id`. Under OpenAI Realtime API semantics, a cancel event without `response_id` is a no-op when multiple responses are in flight.

This PR:
1. Records `response_id` on `_ResponseGeneration` from `ResponseCreatedEvent.response.id`.
2. Updates `RealtimeSession.interrupt()` to pass `response_id=self._current_generation.response_id` when available, cleanly cancelling the targeted in-flight response, while falling back to a general cancel if absent.

Fixes #5564

## Testing
- Added `test_interrupt_includes_response_id_from_current_generation` in `tests/test_realtime/test_openai_realtime_model.py`.
- `uv run pytest tests/test_realtime/test_openai_realtime_model.py` passed (26 tests).
- `uv run pytest tests/test_realtime_interrupt_cancels_generation.py` passed (6 tests).
- `uv run ruff check` passed.

### DIFF
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/inference_realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/inference_realtime_model.py
@@ -137,6 +137,9 @@ class InferenceRealtimeModel(RealtimeModel):
         self._opts.api_version = None
         if is_xai:
             self._capabilities.can_disable_turn_detection = can_disable_turn_detection
+            self._supports_targeted_cancellation = False
+        else:
+            self._supports_targeted_cancellation = True
         self._inference_opts = _InferenceOptions(
             provider=provider,
             api_key=resolved_api_key,
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -1754,7 +1754,10 @@ class RealtimeSession(
     def interrupt(self) -> None:
         if not self.has_active_generation:
             return
-        is_legacy_azure = self._opts.is_azure and self._opts.api_version is not None
+        is_legacy_azure = (
+            getattr(self._opts, "is_azure", False)
+            and getattr(self._opts, "api_version", None) is not None
+        )
         if (
             isinstance(self._current_generation, _ResponseGeneration)
             and self._current_generation.response_id
@@ -1894,7 +1897,11 @@ class RealtimeSession(
             # interrupted or timed out before the server created it: cancel by id and mark it
             # discarded so its trailing events are skipped, instead of surfacing it
             self._discarded_event_ids.discard(client_event_id)
-            if self._opts.is_azure and self._opts.api_version is not None:
+            is_legacy_azure = (
+                getattr(self._opts, "is_azure", False)
+                and getattr(self._opts, "api_version", None) is not None
+            )
+            if is_legacy_azure:
                 self.send_event(ResponseCancelEvent(type="response.cancel"))
             else:
                 self.send_event(
--- a/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
+++ b/livekit-plugins/livekit-plugins-openai/livekit/plugins/openai/realtime/realtime_model.py
@@ -208,14 +208,17 @@ def _normalize_azure_client_event(event: dict[str, Any]) -> None:
     """In-place normalization of client event dicts for legacy Azure compatibility.
 
     The legacy Azure Realtime API uses "text" for assistant content parts,
-    while the newer OpenAI API uses "output_text".
+    while the newer OpenAI API uses "output_text". Additionally, the legacy
+    Azure beta response.cancel schema does not support response_id.
     """
     item = event.get("item")
-    if item is None:
-        return
-    for content_part in item.get("content", ()):
-        if content_part.get("type") == "output_text":
-            content_part["type"] = "text"
+    if item is not None:
+        for content_part in item.get("content", ()):
+            if content_part.get("type") == "output_text":
+                content_part["type"] = "text"
+
+    if event.get("type") == "response.cancel":
+        event.pop("response_id", None)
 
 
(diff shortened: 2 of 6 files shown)


rrfunde:
Fixed: Preserved bare `response.cancel` for legacy Azure sessions (`is_azure` with `api_version`) by omitting `response_id` both in `interrupt()` / discarded response cancellation and during client event normalization in `_normalize_azure_client_event`.

rrfunde:
Fixed: Added `_supports_targeted_cancellation` to gate `response.cancel` with `response_id` strictly to the OpenAI Realtime API (including Azure GA), keeping the bare cancel schema for subclasses like xAI Realtime API as well as legacy Azure.

rrfunde:
Addressed review feedback:
- Represented targeted cancellation capability explicitly via `_supports_targeted_cancellation` on `RealtimeModel`.
- Direct OpenAI models and LiveKit Inference `openai/...` models keep targeted cancellation enabled.
- Legacy Azure and xAI (`livekit-plugins-xai` and `xai/...` inference routes) have targeted cancellation disabled and use the bare `response.cancel` event without `response_id`.
- Added unit tests covering LiveKit Inference OpenAI and xAI routes and xAI model cancellation.

PR #7260, opened 2026-09-13
fix: skip progressive warm-up for buffered target frames

### LINKED ISSUE #7259: RoomIO progressive framing creates a 40 ms startup gap with realtime-paced 100 ms audio packets
## Bug Description

RoomIO's progressive framing retains 40 ms from an already available 100 ms audio packet. With packets arriving every 100 ms, this creates an initial 40 ms source-playout gap under immediate playback.

This is reproducible with generated PCM and the installed `AudioByteStream`; no model, network, or credentials are needed. We found it while investigating short discontinuities in GPT-Live speech.

The relevant configuration is a 50 ms target frame with `progressive=True`. On a newly constructed or cleared stream:

| Input arriving at t=0 | Frames returned immediately | Audio retained | First output | Gap before the next packet at t=100 ms* |
| --- | --- | --- | --- | --- |
| 100 ms, progressive framing | 20 ms + 40 ms | 40 ms | t=0 | 40 ms |
| 100 ms, fixed 50 ms framing | 50 ms + 50 ms | 0 ms | t=0 | 0 ms |

*The gap assumes an initially empty playback queue that starts consuming emitted frames immediately. Receiver buffering can change the audible result.*

Progression grows from 20 to 40 to 50 ms. After emitting 60 ms, the remaining 40 ms is below the next 50 ms target. It stays buffered until another packet arrives, although that audio is already available.

All PCM bytes are preserved after `flush()`. The defect being reported concerns delivery timing, not discarded samples.

`flush()` does not reset the progressive frame size. `clear()` does reset it. Therefore this condition occurs after construction or clearing, including interruption clearing, rather than necessarily on every utterance.

## Expected Behavior

When complete target-sized frames are already available, initial packetization should avoid creating a playback gap by retaining their audio unnecessarily.

The fast initial emission for providers delivering small fragments should remain available. Globally disabling progression changes that behavior: with 20 ms packets every 20 ms, fixed 50 ms framing first emits at t=40 ms, while progressive framing starts at t=0.

## Reproduction Steps

1. Save the Python script below as `repro.py`.
2. Run it with Python 3.13 and `livekit-agents==1.8.1`, `livekit==1.1.18` installed. With uv:

```bash
uv run --python 3.13 --with 'livekit-agents==1.8.1' --with 'livekit==1.1.18' repro.py
```

3. Inspect the printed `results`. The script asserts the initial output sizes, retained audio, virtual playback gaps, and byte-exact PCM preservation.

This uses the real installed utility and generated 24 kHz mono signed 16-bit PCM. Arrival and playback times are virtual, so operating-system scheduling does not affect this minimal reproduction. The script makes no network calls; uv may need to download dependencies first.

<details>
<summary>Complete standalone reproduction</summary>

```python
"""Offline reproduction; requires installed livekit-agents, makes no network calls."""

import json
import math
from array import array
from dataclasses import asdict, dataclass

RATE = 24000
TARGET_SAMPLES = RATE // 20


@dataclass(frozen=True)
class Packet:
    arrives_ms: int
    input_audio_ms: int
    emitted_frames_ms: tuple[float, ...]
    retained_ms: float


@dataclass(frozen=True)
class Result:
    progressive: bool
    provider_packet_ms: int
    first_emission_ms: int
    playback_gaps_ms: tuple[float, ...]
    packets: tuple[Packet, ...]
    exact_pcm_preserved_after_flush: bool


def reproduce(progressive: bool, packet_ms: int) -> Result:
    stream = AudioByteStream(RATE, 1, samples_per_channel=TARGET_SAMPLES, progressive=progressive)
    input_pcm = bytearray()
    emitted_pcm = bytearray()
    timeline: list[Packet] = []
    gaps: list[float] = []
    playout_end_ms: float | None = None
    first_emission_ms: int | None = None
    for packet_index in range(10):
        arrived_ms = packet_index * packet_ms
        sample_count = RATE * packet_ms // 1000
        offset = packet_index * sample_count
        pcm = array('h', [round(4000 * math.sin(2 * math.pi * 220 * (offset + n) / RATE))
                          for n in range(sample_count)]).tobytes()
        input_pcm.extend(pcm)
        frames = stream.push(pcm)
        for frame in frames:
            if first_emission_ms is None:
                first_emission_ms = arrived_ms
            if playout_end_ms is not None and arrived_ms > playout_end_ms:
                gaps.append(round(arrived_ms - playout_end_ms, 6))
            playout_end_ms = max(float(arrived_ms), playout_end_ms or 0) + frame.duration * 1000
            emitted_pcm.extend(bytes(frame.data))
        timeline.append(Packet(arrived_ms, packet_ms, tuple(round(f.duration * 1000, 6) for f in frames),
                               (len(input_pcm) - len(emitted_pcm)) / (RATE * 2) * 1000))
    for frame in stream.flush():
        emitted_pcm.extend(bytes(frame.data))
    assert first_emission_ms is not None
    assert emitted_pcm == input_pcm
    return Result(progressive, packet_ms, first_emission_ms, tuple(gaps), tuple(timeline), True)


results = tuple(reproduce(progressive, packet_ms) for packet_ms in (100, 20) for progressive in (True, False))
assert results[0].packets[0].emitted_frames_ms == (20.0, 40.0)
assert results[0].packets[0].retained_ms == 40.0
assert results[0].playback_gaps_ms == (40.0,)
assert results[1].packets[0].emitted_frames_ms == (50.0, 50.0)
assert results[1].playback_gaps_ms == ()
assert results[1].first_emission_ms == 0
assert results[2].first_emission_ms == 0
assert results[3].first_emission_ms == 40
print(json.dumps({'clock': 'virtual milliseconds; immediate playback of frames emitted by the real installed utility',
                  'results': [asdict(result) for result in results]}, indent=2))
```

</details>

The four conditions produce:

| Packet cadence | Mode | First emission | Modeled gaps between emitted frames |
| --- | --- | --- | --- |
| 100 ms every 100 ms | Progressive, 50 ms target | 0 ms | One 40 ms gap |
| 100 ms every 100 ms | Fixed, 50 ms target | 0 ms | None |
| 20 ms every 20 ms | Progressive, 50 ms target | 0 ms | Two 20 ms gaps |
| 20 ms every 20 ms | Fixed, 50 ms target | 40 ms | None |

The small-packet control demonstrates a startup-latency tradeoff, not universally better continuity from progressive mode.

## Operating System

Reproduced locally on macOS 26.6.2 arm64, Python 3.13.11. The deterministic utility reproduction does not involve an iOS client.

## Models Used

The surrounding investigation used OpenAI GPT-Live-1 through `livekit-plugins-openai`, with LiveKit Cloud publication and a desktop RTC subscriber. Captured provider output contained 100 ms PCM packets at 24 kHz mono. We are describing that observed packet size, not assuming it is a provider API guarantee.

The standalone reproduction above requires no model or API key.

## Package Versions

```text
Python 3.13.11
livekit-agents==1.8.1
livekit-plugins-openai==1.8.1
livekit==1.1.18
```

The relevant behavior is also present in upstream source at commit `f82ba1eb087a38b5d4031726f1a41d2222f7bfae` (main when inspected on 2026-09-13). We inspected that source; the executable reproduction uses the installed versions above.

- [RoomIO enables progressive framing with a 50 ms target](https://github.com/livekit/agents/blob/f82ba1eb087a38b5d4031726f1a41d2222f7bfae/livekit-agents/livekit/agents/voice/room_io/_output.py#L60).
- [AudioByteStream initialization and frame-size progression](https://github.com/livekit/agents/blob/f82ba1eb087a38b5d4031726f1a41d2222f7bfae/livekit-agents/livekit/agents/utils/audio.py#L120).

## Session/Room/Call IDs

Not required for the deterministic reproduction. Supplementary Cloud checks used temporary synthetic rooms, which were deleted after the runs.

## Proposed Solution

One possible direction is to skip progressive warm-up when the initial buffer already contains a complete target frame. Emit target-sized frames immediately in that case; retain progressive startup for smaller incoming fragments.

For the 100 ms / 50 ms case, that would emit 50 + 50 ms without introducing an application prebuffer or delaying the first output. The current RoomIO options do not expose a public selection for progressive framing.

This is a proposed direction, not a tested upstream patch. Useful regression cases would include:

- 100 ms packets at realtime cadence, with no artificial initial underfill.
- Small incoming fragments, preserving their early first emission.
- `clear()` and interruption restart behavior.
- `flush()` and end-of-stream byte preservation, without assuming `flush()` resets progression.

## Additional Context and Verification Limits

We also replayed fictional GPT-Live audio at its recorded arrival times through the installed gate, actual `_ParticipantAudioOutput`, RTC source, and LiveKit Cloud transport. We preserved the installed 200 ms source queue capacity.

Two baseline runs reported first source-playout gaps of **40.421 ms** and **39.431 ms**. Those intervals come from the SDK's playback-run accounting. They corroborate the utility timing behavior, but are not direct measurements of a gap heard on a physical device.

An offline comparison using the same captured frame times changed the first predicted gap from **40.397 ms to 0.397 ms** when using fixed framing. Later gaps remained unchanged, including one of **119.785 ms**. Correcting this initial underfill therefore does not resolve arbitrary later provider-arrival jitter.

The fixed-framing comparison has **not** been validated through Cloud or an iPhone. The Cloud harness uses RoomIO directly and does not reproduce the complete AgentSession speech scheduler. A separate 180 ms prebuffer experiment reduced source gaps but did not consistently improve receiver concealment. We do not claim that either experiment fixes all audible cutouts.

Related work:

- #5292 introduced progressive output to reduce time to first audio. This report concerns input arriving at playback rate in larger complete packets.
- #5158 describes related initial-audio truncation symptoms. This report isolates deterministic buffering before the RTC source; it does not establish a common cause with that issue.

No private conversation content, credentials, customer audio, or production identifiers are included in this report.

swayamg20:
## Summary

Fixes #7259.

With RoomIO's 50 ms target, progressive framing currently emits 20 + 40 ms from an already available initial 100 ms packet and retains 40 ms. At realtime packet cadence, that creates a modeled 40 ms startup gap with immediate playback.

This change skips the progressive ramp when a full target frame is buffered before the first emission. The same packet immediately produces 50 + 50 ms, with no retained tail or additional prebuffer.

- Preserve the existing early 20 ms emission for smaller fragments and the progression once that ramp has started.
- Keep `clear()` resetting startup selection and `flush()` preserving progression.
- Add byte-stream and RoomIO regression coverage for packet cadence, multiple sample rates/channels, interruption restart, memoryview input, and byte-exact flush behavior.
- Make the playback-test fake wait for explicitly driven draining, rather than instantly consuming queued audio during interruption tests. Existing playback assertions are preserved.
- Update the utility's documentation to describe conditional startup sizing.

## Validation

Python 3.13.12, dependencies installed from the existing lockfile with `uv sync --frozen --all-extras --dev` (uv 0.12.10). Cloud credentials were removed from the test environment.

- Utility and playback-progress tests: **44 passed**.
- Focused audio suite, including decoder, sink proxy, emitter, and array buffer: **145 passed, 1 credential-dependent skip**.
- Regression control using the original upstream `AudioByteStream` class loaded in memory: **26 failed, 8 passed**; the patched class passes all 34 cases in that selection.
- Changed-file Ruff formatting/lint and `git diff --check`: pass.
- Core strict mypy check: **215 source files pass**.
- Full `--unit --audio_eot --no-concurrent` run: **3,148 passed, 5 skipped, 9 setup errors**. All nine errors are in `tests/test_room.py` because `livekit-server` is not installed locally; no assertion failures.

Repository-wide checks are **not green on the base commit** `f82ba1eb087a38b5d4031726f1a41d2222f7bfae`: formatting/lint/type checking encounter duplicate `secondary_languages` arguments in `livekit-plugins/livekit-plugins-elevenlabs/livekit/plugins/elevenlabs/stt.py` and `tests/test_plugin_elevenlabs_stt.py`. Neither file is changed here; compiling their exact upstream contents reproduces the syntax errors.

## Scope

This addresses initial framing underfill, not arbitrary provider-arrival jitter. Non-target-aligned packets can still leave partial frames. There are no public API, dependency, or RoomIO queue-capacity changes. This patch has not been validated through Cloud transport or on a physical audio device.


longcw:
thanks for the pr! closing in favor of https://github.com/livekit/agents/pull/7261

PR #7272, opened 2026-09-14
fix(room_io): release a partial frame before the source runs dry

### LINKED ISSUE #7259: RoomIO progressive framing creates a 40 ms startup gap with realtime-paced 100 ms audio packets

longcw:
**Problem:** `RoomIO` frames output audio into 50 ms frames behind a 20, 40 ms ramp, so any packet that ends mid-frame leaves a partial frame in `AudioByteStream` until the next packet arrives. A realtime-paced model such as GPT-Live plays out, runs dry for the length of that partial frame, and then continues, which is audible at the start of each response and after every interruption.

**Fix:** The partial frame is released on a deadline instead of on the next packet: after each push, `capture_frame` arms a timer that flushes the byte stream 20 ms before the audio queued ahead of it, in the source and in the channel, runs out. A source with slack packs full 50 ms frames, a source about to run dry gets its partial frame in time, and the ramp goes: nothing plays before the first frame, so one under 20 ms waits for more instead of going out small.

Fixes #7259. Stacks on #7271, the revert of #7261.

<details>

> **How to see it**
>
> `test_the_tail_follows_before_the_source_runs_dry` in `tests/test_playback_progress.py` pushes one 80 ms packet and expects 50 ms in the source at once, a timer due in 30 ms, and the full 80 ms after 40 ms. On the base commit the 20 ms left after the ramp waits for a second packet that never comes.
>
> **Why a deadline rather than a frame-size floor**
>
> Holding a partial frame costs nothing while it reaches the source before the audio ahead of it plays out, so the deadline is the exact criterion. A size floor, the approach in #7261, decides from the packet size alone and knows nothing about the queue, so a 20 ms source yields 20 ms frames for the whole session, since `capture_frame` never waits between pushes.
>
> **Why the timer is armed in `capture_frame` and counts the channel**
>
> The partial frame is created by a push, and the push is the only point that sees every frame ahead of it: the source queue plus the frames still in the channel, which is `_pushed_duration - _source_pushed_duration`. Arming in `_forward_audio` would see only the source, and a packet that fills no frame never reaches the forwarder, so a source that starts with packets under 40 ms would gap once per response. One timer is armed per push, the previous one is cancelled, and a fire on an empty buffer flushes nothing.
>
> **Why the ramp goes**
>
> The ramp existed so that a first small packet plays at once. With the deadline, the first timer of a segment fires at once, and the flush skips a first frame under 20 ms because nothing plays yet and waiting opens no gap; the segment-end `flush()` passes `force=True` so a segment that is all tail still goes out. The TTS `AudioEmitter` sends its 10 ms head and the 40 ms frame a task hop apart, so the first frame is 50 ms rather than 10. A 100 ms first packet, the #7259 case, becomes two 50 ms frames with no timer involved, where the ramp emitted 20 and 40 ms and left 40 ms to the deadline.
>
> **Where 20 ms comes from**
>
> The callback needs one event loop turn to wake `_forward_audio` plus one FFI request to reach the rust queue, which drains in 10 ms ticks. Two ticks cover that under normal load; a loop stall longer than 20 ms gaps every path regardless of framing.
>
> **Blast radius**
>
> Only `_ParticipantAudioOutput` in `voice/room_io/_output.py` changes. `AudioByteStream` gains only the `buffered_duration` property, and the TTS `AudioEmitter` at `tts/tts.py:1237` and `tts/tts.py:1283` is untouched, since `RoomIO` re-frames its output. The rust `AudioSource` buffers sub-10 ms tails itself, so a small released frame costs one FFI round trip and nothing else.
>
> **Why the playback-progress fake changed**
>
> `_FakeSource.wait_for_playout` zeroed `queued_duration` instantly. Two tests push 500 ms and only passed because the ramp left 40 ms in the byte stream, which delayed the drain until the interruption was seen; with nothing held back the fake now drains only when the test drives it, matching a real source that takes as long as the audio it holds.

</details>


davidzhao (approved):
lg

theomonnom (approved):
ah makes sense!

PR #7273, opened 2026-09-14
fix(tts): forward inner frames as-is in FallbackAdapter and StreamAdapter

theomonnom:
Fixes AGT-3438 (Eightfold: first syllable clipped with `inference.TTS("inworld/inworld-tts-2", voice="Marta")` wrapped in `tts.FallbackAdapter`).

## Problem

`FallbackAdapter` and `StreamAdapter` take the frames the wrapped TTS already emitted, convert them back to bytes, and push them into their own `AudioEmitter`. That emitter re-chunks them through the progressive frame-size ramp (20/40/80/160/200 ms) and holds a tail, so most of the first chunk waits for the provider's next chunk. The slow-generation flush never arms below 150 ms sent, so nothing releases it.

Measured with a fake stream that sends a 220 ms chunk and then pauses:

| Path | Released before the next chunk |
|---|---|
| Wrapped TTS on its own | 130 ms |
| Through `FallbackAdapter` (before) | 50 ms |
| Through `FallbackAdapter` (after) | 120 ms |

The room output plays out as soon as the first frame lands and has a 200 ms queue, so 50 ms of lead runs dry before inworld's second chunk arrives (60–300 ms later) and the gap lands on the first syllable. Cartesia's leading silence hides the same stall, which is why it sounded provider-specific.

## Fix

Add `AudioEmitter.push_frame()` to forward an already-framed chunk as-is, and use it in both adapters. The emitter still holds its usual 10 ms tail to tag the final frame, hence 120 ms rather than 130 ms. Flush and end-of-segment handling no longer requires a byte stream to exist, since a segment fed only via `push_frame()` never creates one.

## Test

`test_push_frame_is_not_rechunked`: a 220 ms frame pushed with `push_frame()` comes out as one 210 ms frame plus the 10 ms tail, where `push()` would ramp from 20 ms and hold the rest.


longcw (approved):
may also fix it in `_ChunkedStreamFromStream` https://github.com/livekit/agents/blob/888db1c2b17dff3646f12605f97c68e6aa55fa1a/livekit-agents/livekit/agents/tts/tts.py#L493

PR #7280, opened 2026-09-14
bithuman: resolve the SDK on Python 3.10 and 3.14 again

theomonnom:
## Summary

The `python_version >= '3.11' and python_version < '3.14'` marker on the `bithuman` dependency (#6561, #6614) was correct when bithuman 2.3.11 through 2.8.0 only shipped cp311–cp313 wheels. bithuman 2.8.1 (2026-08-01) restored cp310 and cp314 wheels and declares `Requires-Python: >=3.10,<3.15`, and every release since (2.9.0, 3.x) does the same. The marker now leaves the plugin installed without its SDK on Python 3.10 and 3.14.

Changes:

- Drop the marker; the dependency is back to `bithuman>=0.5.25,<3`.
- Add the `Programming Language :: Python :: 3.14` classifier, matching `livekit-agents`.
- `uv lock --upgrade-package bithuman`: 2.7.0 → 2.9.0, the newest 2.x with wheels for every supported interpreter. 2.9.0 dropped its `networkx` dependency, which is why that package leaves the lock.
- Remove `print_unsupported_python` and the two `try`/`except ImportError` wrappers that called it. That helper existed only to explain the SDK's absence on the versions the marker excluded; with the marker gone it fires on any `ImportError` and blames the interpreter for unrelated failures.
- Declare `pillow>=10.3.0`.

Not changed: the `<3` upper bound. bithuman 3.0.0 shipped on 2026-09-07 and has not been checked against `avatar.py`, so that belongs in its own PR.

## The pillow fix

Removing the wrapper exposed a pre-existing bug it had been swallowing. `avatar.py` imports `PIL` at module scope, but the package never declared `pillow`, so importing the plugin raised `ModuleNotFoundError: No module named 'PIL'` on **every** Python version unless something else in the environment happened to pull Pillow in. Reproduced against unmodified `main` on CPython 3.13, a fully supported version:

```
File ".../livekit/plugins/bithuman/__init__.py", line 24, in <module>
    from .avatar import AvatarSession, BitHumanException
File ".../livekit/plugins/bithuman/avatar.py", line 16, in <module>
    from PIL import Image
ModuleNotFoundError: No module named 'PIL'
```

The old wrapper caught this, called `print_unsupported_python()`, which printed nothing on 3.13 and re-raised — which is exactly the misdiagnosis the helper's removal is meant to avoid. `livekit-plugins-lemonslice`, the only other plugin importing PIL, declares `pillow>=10.3.0`; this matches it.

## Verification

- `uv pip install` of the plugin on CPython 3.14.7 and 3.10.18: both resolve `bithuman==2.9.0` and `pillow==12.3.0`, and `from livekit.plugins.bithuman import AvatarSession, BitHumanException` succeeds on both.
- `uv lock --check` passes; `ruff check` and `ruff format --check` pass.

Reported in Slack: https://live-kit.slack.com/archives/C05P6DRC6MV/p1789407534492229


livekit-plugins/livekit-plugins-bithuman/pyproject.toml
```diff
@@ -22,17 +22,17 @@ classifiers = [
...
     "Programming Language :: Python :: 3 :: Only",
 ]
 
-# bithuman 2.x builds only for Python 3.11 through 3.13, so the marker keeps
-# 3.10 and 3.14 from resolving it at all rather than backtracking onto 2.3.10,
-# which ships cp310 and cp314 wheels and declares no upper bound. The floor
-# stays wide so platforms 2.x dropped (Windows, x86_64 macOS, macOS < 14) can
-# still resolve the 1.x line, which satisfies everything avatar.py calls.
+# bithuman 2.8.1+ ships wheels for every Python livekit-agents supports (3.10
+# through 3.14), so the dependency no longer needs a python_version marker. The
```
davidzhao: remove bithuman comment here?

chenghao-mou (approved):

PR #7282, opened 2026-09-15
fix(avatar): restore the previous audio route on AvatarSession.aclose()

### LINKED ISSUE #7276: AvatarSession.aclose() doesn't unbind the audio output start() installed, so a failed avatar start silently mutes the agent
### Bug Description

`AvatarSession.start()` points the agent session's audio output at the avatar's datastream. `AvatarSession.aclose()` never puts it back. That asymmetry is invisible on the happy path, but if the avatar fails to start, the agent is then silent (and doesn't degrade to just audio).

### Expected Behavior

`AvatarSession.aclose()` sets audio output back to what it was before.

### Reproduction Steps

```bash
import asyncio
from livekit import rtc
from livekit.agents.voice import AgentSession
from livekit.agents.voice.avatar import AvatarSession, DataStreamAudioOutput


class FakeAvatarSession(AvatarSession):
    """Mimics what an avatar plugin's start() does."""

    @property
    def avatar_identity(self) -> str:
        return "avatar-worker"

    async def start(self, agent_session, room):
        await super().start(agent_session, room)
        agent_session.output.replace_audio_tail(
            DataStreamAudioOutput(room=room, destination_identity=self.avatar_identity)
        )


async def main():
    session = AgentSession()
    room = rtc.Room()
    print("before start:       ", session.output.audio)

    avatar = FakeAvatarSession()
    await avatar.start(session, room=room)
    print("after plugin start: ", type(session.output.audio).__name__)

    # the avatar never joins: wait_for_join(timeout=...) raises, the app gives up
    await avatar.aclose()
    print("after aclose:       ", type(session.output.audio).__name__)


asyncio.run(main())


Output:


before start:        None
after plugin start:  DataStreamAudioOutput
after aclose:        DataStreamAudioOutput
```

### Operating System

All

### Models Used

_No response_

### Package Versions

```bash
livekit-agents==1.8.0
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

Have AvatarSession remember what it replaced in start() and restore it in aclose() — symmetric with how it already unregisters its own event handlers and cancels its own join task. (Or provide a parameter to close to do this)

### Additional Context

_No response_

_No response_

dorukdumlu:
## Summary

Avatar plugins route the agent session's audio to the avatar with `output.replace_audio_tail(...)` in `start()`, and `aclose()` never put it back. When the avatar fails to start (`wait_for_join` timing out, the app giving up), the session stays pointed at a dead datastream sink and the agent goes silent instead of degrading to regular audio. With the canonical ordering (avatar started before `session.start()`), the stale route also makes `session.start()` set `room_options.audio_output = False`, so no RoomIO audio ever exists to fall back to.

Fixes #7276

## What changed

- `AvatarSession` gains `_attach_audio_output(sink)`: it installs the route and remembers what `replace_audio_tail()` swapped out. `aclose()` restores it:
  - the previous tail goes back under any wrapper chain (TranscriptSynchronizer and friends stay attached),
  - a route that was installed over nothing is cleared, so a later `session.start()` sets up room audio normally,
  - a route someone else installed after the avatar is left alone (identity-guarded).
- `AgentOutput.replace_audio_tail()` now returns the sink it replaced, and `AgentOutput.audio_tail` exposes the current tail. Both backward compatible.
- All in-repo avatar plugins are migrated to the helper, a mechanical one-line change each. Out-of-tree plugins that call `replace_audio_tail` directly keep working exactly as before, just without the restore.

One known limit, called out rather than papered over: if the avatar dies after `session.start()` already ran, restoring the route can't bring audio back, because RoomIO audio was disabled at startup and there is no `_ParticipantAudioOutput` to return to. Handling that would mean building the room audio output late, which felt like a separate feature. Happy to take it as a follow-up if there's interest.

## Test plan

- [x] New `tests/test_avatar_session_close.py` (5 tests): restore over nothing, restore of a previous output, restore of the tail under a wrapper proxy, a newer route wins, aclose without an attach is a no-op. The restore tests fail without the fix.
- [x] `pytest tests/test_audio_sink_proxy.py tests/test_inference_avatar.py tests/test_recorder_io.py --unit` (90 passed)
- [x] ruff and mypy clean on the touched files


dorukdumlu:
Both findings were right, addressed in acc2670. The red one was a plain miss on my side: I swept livekit-plugins for `replace_audio_tail` callers and forgot the one avatar that lives in core, so the inference AvatarSession kept bypassing the tracking. It now goes through `_attach_audio_output` like the plugins, with a close regression test on the inference path. For the yellow, the restore is synchronous and doesn't depend on the participant removal at all, so it simply runs first in `aclose()`, before any await; added a test that cancels the close mid-removal and checks the route still came back.

PR #7290, opened 2026-09-15
add pid to log lines from parent process

andrewnitu:
## What
This PR adds 'pid' to log lines from the parent worker process (logs such as `registered worker`, `HTTP server listening`, `preloading plugins`) where the pid is not already present.

Lines from the child processes already carry pid so no need to add it here.

NOTE: This doesn't cover all possible lines, e.g. lines printed via print() without using the logger, lines from imported libraries, etc. Might be ways around this but not sure at the moment.

## JS parity
JS SDK already emits pid from both parent processes and child processes.

## Testing
Verified on a hosted agent in staging: parent lines `pid: 1`, job child `50`, prewarmed child `122`, no JSON line without a pid.


theomonnom:
This isn't going to work for any native libraries, since they don't use the Python logger. We have the same issue on the job processes

andrewnitu:
@theomonnom agreed, I noted that in the description. But it's better than what was before :-)

we have to come up with another solution if we want to capture those.

theomonnom (approved):

PR #7295, opened 2026-09-15
worker: read agent name from livekit.toml as last fallback

u9g:
## Why

The agent name is moving out of code into `livekit.toml`, checked into git, so Cloud, `lk`, simulations and observability can all read the same name without booting the worker. Proposal: https://app.notion.com/p/livekit/Proposal-livekit-toml-layout-for-agent-name-and-regions-3d83c4901a4281beb3b9f1f3e74de44d

This is phase 1: the SDK learns to read the file and warns about names set in code. Phase 2, a few weeks out, removes the code parameter.

## What

Agent name precedence is now `LIVEKIT_AGENT_NAME_OVERRIDE`, then the `agent_name` argument to `rtc_session` (code), then `LIVEKIT_AGENT_NAME`, then `[agent] name` from `./livekit.toml`.

- The toml fallback is applied at the start of `run()` and only in production. `run(devmode=True)` (`lk agent dev`, `dev`, `console`) never takes it, so dev workers stay out of the deployed agent's pool.
- When the name comes from code, `run()` logs one warning pointing at `livekit.toml`. The `agent_name` docstrings mark the parameter deprecated.
- A missing file, a missing `[agent]`/`name`, or a malformed file means no name; malformed files log at debug.
- `tomllib` on 3.11+, `tomli` on 3.10 (new marker dependency, `uv.lock` updated).

JS parity: livekit/agents-js has the matching change.

## Verification

`uv run --project livekit-agents pytest tests/test_worker_config.py -q --unit`: 15 passed (5 new: toml fallback, env beats toml, devmode ignores toml, missing file, code-name warning). `ruff check`, `ruff format --check` and `scripts/check_types.py` clean on touched files.


livekit-agents/livekit/agents/worker.py
```diff
@@ -527,6 +556,24 @@ def decorator(
...
+                "agent_name is set in code; move it to livekit.toml ([agent] name). "
+                "The agent_name parameter will be removed in a future release.",
+                extra={"agent_name": self._agent_name},
+            )
+        if self._agent_name_source != "none" or devmode:
+            return
+        name = _toml_agent_name(os.path.join(os.getcwd(), "livekit.toml"))
+        if name:
+            self._agent_name = name
+            self._agent_name_source = "toml"
```
theomonnom:
```suggestion
            self._agent_name_source = "livekit.toml"
```
nitpicking

theomonnom (approved):

PR #7308, opened 2026-09-16
fix(mistralai): standardize API error handling

jeanprbt:
## Summary

Standardize Mistral error handling across LLM, STT, and TTS while preserving provider error details and retry semantics.

## Motivation

With an invalid LLM model such as `foo`, Mistral returns HTTP 400 with error code 3003.

```text
Model foo is currently not in use.
```

The LLM plugin previously handled Mistral's `SDKError` as a generic exception, turning this non-retryable API response into a retryable `APIConnectionError`. One user turn therefore produced four misleading `Connection error` events, the
initial request plus three retries.

## Changes

- Translate Mistral `SDKError` and `HTTPValidationError` into `APIStatusError` across LLM, STT, and TTS.
- Preserve the provider status, body, and `x-request-id`.
- Translate actual `asyncio`/`httpx` timeouts into `APITimeoutError`.
- Preserve LLM stream retryability so errors after output begins are not retried, avoiding duplicated partial output.

Ordinary 4xx responses are now non-retryable, while transient statuses continue to follow `APIConnectOptions`.

## Tests

- HTTP 400 is attempted once and remains an `APIStatusError`.
- HTTP 422 preserves its status, body, and request ID.
- An error after the first streamed chunk is not retried.

```bash
uv run pytest tests/test_plugin_mistralai_llm.py -q
```


tinalenguyen (approved):
thank you for the PR!

PR #7309, opened 2026-09-16
fix(workflows): forward Twilio CallToken during warm transfers

piyush-gambhir:
Warm transfers normally show the business number to the supervisor. This adds optional per-call caller-ID preservation through Twilio's CallToken, while retaining the business number as the default and fallback.

## Behavior

- `twilio_from_number` remains the business/verified number. Supply `original_caller_number` and a nonempty `twilio_call_token` from the same validated inbound webhook to preserve that caller ID. Keep the pair in server-side state keyed by `CallSid`, outside prompts, logs, and participant attributes.
- Without a token, including an empty token, use the business number. Token-free transfers remain compatible with older Twilio clients.
- Token use validates SDK support at construction and requires `twilio>=6.55.0`. Missing or incompatible SDKs fail before transfer setup.
- Retry once without the token, from the business number, only for HTTP 400 with Twilio error 21210 (unverified From) or 21212 (invalid From, including unavailable caller IDs). Other rejections, transport failures, and failed fallbacks propagate without further dialing.
- Cancellation and ringing-timeout failures propagate promptly without waiting for provider cleanup. A small instance-owned `utils.aio.TaskSet` retains pending creation/cleanup so a late call SID can still be canceled. There is no process-global task registry, five-second teardown deadline, or custom HTTP timeout policy. Cleanup remains best effort while the worker is alive.

## Validation

- `pytest tests/test_twilio_connector_warm_transfer.py -q`: **33 passed**. Covers construction-time compatibility, legacy dialing, token forwarding/omission, XML escaping, both fallback codes, retry exclusions, failed fallback, correct-SID cancellation, and real-thread late creation/cleanup after cancellation.
- `make fix`, repository Ruff formatting/lint, and focused strict mypy: passed.
- Full `make check` reaches typechecking but remains blocked by missing optional-plugin typing dependencies (`livekit.plugins.anam` missing its type marker; the environment also reports missing `types-aiofiles`).
- GitHub CI on commit `4e4268e`: Ruff, typechecking on Python 3.10/3.13, unit tests, and blockguard tests passed. The local optional-plugin typing limitation above did not reproduce in CI.
- Historical live QA on 2026-09-16 verified valid-token caller ID, business-number defaults, and invalid-token fallback in both SDKs. This revision was checked with deterministic tests; no new live PSTN calls were placed.

Feature scope discussion: https://github.com/livekit/agents/issues/7375. Opened after the implementation PRs to request maintainer scope review; this does not represent prior approval.

JavaScript counterpart: https://github.com/livekit/agents-js/pull/2540.

## References

- [Inbound webhook parameters](https://www.twilio.com/docs/voice/twiml#request-parameters)
- [Calls API](https://www.twilio.com/docs/voice/api/call-resource#create-a-call)
- [Error 21210](https://www.twilio.com/docs/api/errors/21210) and [error 21212](https://www.twilio.com/docs/api/errors/21212)
- [Twilio Python 6.55.0](https://github.com/twilio/twilio-python/releases/tag/6.55.0)


piyush-gambhir:
JS counterpart moved from livekit/agents-js#2510 to livekit/agents-js#2540 (same change, rebased as linear commits on main).

Issue #7170, opened 2026-09-08
google: FunctionTool free-form object parameters reach Gemini without additionalProperties and come back empty (text API supports parameters_json_schema)


IngLP:
## Summary

With the Google plugin on the text (`generateContent`) path, a `FunctionTool` whose parameter is a free-form object (`dict[str, str]`, `dict[str, Any]`, a Pydantic model with `extra="allow"`, …) always reaches Gemini as an object with no allowed keys, and the model answers with an empty object `{}`. The same tool declared as a `RawFunctionTool` with `additionalProperties` is filled correctly.

The cause is in `livekit/agents/llm/_provider_format/google.py::to_fnc_ctx`: a `FunctionTool` is always converted through `_GeminiJsonSchema(...).simplify()` into the legacy `parameters` field, and the simplifier pops `additionalProperties` (`livekit/plugins/google/utils.py`, `_simplify`). The `use_parameters_json_schema` flag only affects `RawFunctionTool`, so on the text API — where `parameters_json_schema` *is* supported and does accept `additionalProperties` — ordinary function tools never benefit from it.

## Minimal reproduction

```python
import asyncio, os
from livekit.agents import llm
from livekit.agents.llm import function_tool
from livekit.plugins import google


@function_tool
async def save_contact(fields: dict[str, str]) -> None:
    """Save the contact fields gathered in the conversation (free-form keys)."""


save_contact_raw = function_tool(
    lambda raw_arguments: None,
    raw_schema={
        "name": "save_contact",
        "description": "Save the contact fields gathered in the conversation (free-form keys).",
        "parameters": {
            "type": "object",
            "properties": {
                "fields": {"type": "object", "additionalProperties": {"type": "string"}}
            },
            "required": ["fields"],
        },
    },
)


async def run(tool, label):
    ctx = llm.ChatContext()
    ctx.add_message(role="user", content="Hi, I'm Mario Rossi, my email is mario.rossi@example.com. Please save my contact.")
    model = google.LLM(model="gemini-3.8-flash", api_key=os.environ["GOOGLE_API_KEY"])
    async with model.chat(chat_ctx=ctx, tools=[tool], tool_choice="required") as stream:
        async for chunk in stream:
            for call in (chunk.delta.tool_calls if chunk.delta else []):
                print(f"{label}: {call.name}({call.arguments})")
    await model.aclose()


async def main():
    await run(save_contact, "FunctionTool (dict[str, str])")
[...]

PR #7312, opened 2026-09-16
fix(google): send FunctionTool schemas as parameters_json_schema on the text API

JustinScitech:
This directly fixes issue #7170

### Summary
A free-form object is a JSON object with **any string keys** rather than a fixed set of fields. 

An issue occurring due to this was on the Gemini text API, where a FunctionTool with a free-form object parameter (dict[str, str], dict[str, Any], a Pydantic model with extra="allow") reached Gemini as an object with no allowed keys, so the model answered with an empty dict {}. However, the same tool written as a RawFunctionTool was filled correctly.

### Root cause
Gemini takes tool parameters in two ways: 
1. The legacy parameters field, a Gemini Schema that cannot express free-form objects
2. `parameters_json_schema`, the standard JSON Schema that can express free-form objects. 

Only RawFunctionTool was moved to parameters_json_schema in a previous fix. Ordinary function tools stayed on the legacy field, which is the format that Gemini Live requires, so the Gemini text API never received a schema capable of describing them. Building that legacy field runs _GeminiJsonSchema.simplify(), which drops additionalProperties, and that is where the allowed keys disappeared.

### Fix
The function `to_fnc_ctx` now honours the existing use_parameters_json_schema flag for FunctionTool, the same way it already did for RawFunctionTool:

Text API (which is the default for google.LLM): the Pydantic-generated JSON Schema is sent as-is, so additionalProperties survives.
Live API (which sets `use_parameters_json_schema=False`, set by the realtime model): unchanged, still the simplified legacy parameters, because Gemini Live rejects parameters_json_schema ([googleapis/python-genai#1147](https://github.com/googleapis/python-genai/issues/1147)).

A tool with no parameters sends None, as before. Free-form objects still will not work on the Live API. That is a Gemini limitation rather than something the plugin can work around, and the issue asks for that path to stay as it is. These are all re-verified in the tests added to the test_schema_gemini.py file.

### Review order
1. `livekit-agents/livekit/agents/llm/_provider_format/google.py`: the `FunctionTool` branch of `to_fnc_ctx`, about fifteen lines. The `RawFunctionTool` branch is untouched.
2. `tests/test_schema_gemini.py`: five tests appended to the existing Gemini schema module.

### Testing

The five new tests need no API key and cover both API paths, no-parameter tools, `FunctionTool` and `RawFunctionTool` parity, and the `create_tools_config` output validating as a genai `FunctionDeclaration` that still carries the nested `additionalProperties`.

To rerun everything:

```bash
uv run --frozen pytest tests/test_tools.py tests/test_schema_gemini.py -q
uv run --frozen ruff check livekit-agents/livekit/agents/llm/_provider_format/google.py tests/test_schema_gemini.py
uv run --frozen ruff format --check livekit-agents/livekit/agents/llm/_provider_format/google.py tests/test_schema_gemini.py
uv run --frozen --group typing mypy -p livekit.agents -p livekit.plugins.google
git diff --check
```

Results on this branch:

| Check | Result |
| --- | --- |
| pytest | 131 passed |
| ruff check | All checks passed |
| ruff format --check | 2 files already formatted |
| mypy | no issues in 231 source files |
| git diff --check | clean |

For an end-to-end check, you can run the reproduction from #7170 with `GOOGLE_API_KEY` set. 

```
GOOGLE_API_KEY="PUT_KEY_HERE" uv run --frozen python - <<'PY'
import asyncio, os


@function_tool
async def save_contact(fields: dict[str, str]) -> None:
    """Save the contact fields gathered in the conversation (free-form keys)."""


async def main():
    chat_ctx = llm.ChatContext()
    chat_ctx.add_message(
        role="user",
        content="Hi, I'm Mario Rossi, my email is mario.rossi@example.com. Please save my contact.",
    )
    model = google.LLM(model="gemini-3.6-flash", api_key=os.environ["GOOGLE_API_KEY"])
    async with model.chat(chat_ctx=chat_ctx, tools=[save_contact], tool_choice="required") as stream:
        async for chunk in stream:
            for call in chunk.delta.tool_calls if chunk.delta else []:
                print(f"{call.name}({call.arguments})")
    await model.aclose()


asyncio.run(main())
PY
```

It was verified against `gemini-3.6-flash`, `gemini-3-flash-preview` and `gemini-3.1-flash-lite`: the `dict[str, str]` tool is now filled with the same keys as the raw variant, and nested Pydantic models and no-parameter tools still work. 

Also `gemini-2.5-flash` from the original report now returns a 404 and indicates that it is closed to new users:
image

`This model models/gemini-2.5-flash is no longer available to new users. Please update your code to use models/gemini-3.6-flash for the latest features and improvements. We recommend you to use the Interactions API.`


theomonnom (approved):
lgtm!!

PR #7313, opened 2026-09-16
fix(liveavatar): register the plugin, interruption fix

tinalenguyen:
Register the plugin with Plugin.register_plugin. liveavatar was the only avatar plugin that never did, so its logger stayed at NOTSET and resolved up to the livekit logger, which setup_logging pins to WARN. Every logger.debug in the plugin was dropped before reaching a handler. Also exports __version__.

Keep the interrupt latch across a redundant speak start. _avatar_interrupted suppresses the server's speak_ended for a turn that _on_clear_buffer already finished with notify_playback_finished(interrupted=True). The reset sat outside the already_speaking guard, so a redundant start for the same turn (speak_started plus state_updated -> talking) cleared the latch and let a second notify_playback_finished through.

Stop logging every audio_buffer_appended ack. _forward_audio sends one agent.speak per 600ms-1s chunk, so the ack arrives about once per second for the whole turn and carries nothing playback is not already tracking.


theomonnom (approved):
YEshh, nice catch!

PR #7314, opened 2026-09-16
fix(anam): retry retryable HTTP status failures when starting a session

tinalenguyen:
`AnamAPI._post` raises `APIStatusError` for any non-OK response from
  `/v1/engine/session`, but its retry loop only caught
  `(aiohttp.ClientError, asyncio.TimeoutError)`. The `APIStatusError` therefore
  escaped on the first attempt, so `max_retry` / `retry_interval` were dead
  config for every HTTP-status failure — a single transient 5xx from Anam killed
  avatar startup with no retry.

  This adds an `except APIStatusError` arm that:

  - re-raises immediately when `retryable` is `False` (4xx other than 408/429/499),
  - otherwise logs, sleeps `retry_interval`, and retries within `max_retry`,
  - re-raises the `APIStatusError` once retries are exhausted, so the caller keeps
    `status_code` / `body` instead of a wrapped `APIConnectionError`.

  The transport arm is unchanged: `aiohttp.ClientError` and timeouts still end in
  `APIConnectionError` as before.

  This matches the sibling avatar plugins (`protoface`, `avatario`, `liveavatar`,
  `lemonslice`, `keyframe`, `runway`), which all consult `APIStatusError.retryable`.


livekit-plugins/livekit-plugins-anam/livekit/plugins/anam/api.py
```diff
@@ -143,7 +143,7 @@ async def _post(
         url = f"{self._api_url}{endpoint}"
         session = self._session or aiohttp.ClientSession()
         try:
-            for attempt in range(self._conn_options.max_retry):
+            for attempt in range(self._conn_options.max_retry + 1):
```
theomonnom, line 146: we already have the retry mechanism handled inside the base class of TTS

theomonnom (approved):

livekit-plugins/livekit-plugins-anam/livekit/plugins/anam/api.py
tinalenguyen: hmm i don't think there is for anam avatar API?

PR #7318, opened 2026-09-17
Move OpenAI Realtime inference client into agents core

russellmartin-livekit:
## Summary
- move the OpenAI Realtime protocol engine into a private core module
- add the hosted RealtimeModel under livekit.agents.inference
- retain the direct-provider API and legacy hosted imports through compatibility wrappers
- keep GPT-Live owned by livekit-plugins-openai until LiveKit Inference supports it
- update dependency bounds, documentation, and realtime CI paths

## Compatibility
- preserve the existing direct OpenAI RealtimeModel metrics label
- restore realtime.utils and realtime.gpt_live_types package attributes
- realtime engine logs now use the livekit.agents logger because the implementation lives in core

## Validation
- 126 targeted unit tests passed
- Ruff lint and format checks passed for affected files
- mypy passed for livekit.agents and livekit.plugins.openai


shawnfeldman (approved):
Approving, with the scope of my review stated explicitly: I reviewed this as inference-gateway — the move fidelity and the hosted (LiveKit Inference) realtime client. I did not audit the 2.4k-line realtime protocol engine line by line, because it moves essentially unchanged.

**Move fidelity.** `plugins/openai/realtime/realtime_model.py` → `agents/llm/_realtime/openai.py` differs by **37 lines out of 2449**, and all but three are import re-homing (`..log`/`..models`/`.utils` → absolute `livekit.agents.*` + `.openai_types`/`.openai_utils`). `utils.py` → `openai_utils.py` likewise. Both old paths keep `import *` + `__getattr__` shims, and `test_plugin_import_is_canonical_class` pins that the plugin import resolves to the canonical class.

**The three non-move lines** are the deliberate `if type(self) is RealtimeModel: self._label = "livekit.plugins.openai.realtime.realtime_model.RealtimeModel"` pin.

**Hosted client** (`inference/realtime/openai.py`) is a faithful port of `inference_realtime_model.py` — same three overrides (`_create_ws_url_and_headers` swapping in the inference token + `X-LiveKit-Inference-Provider`, `_wrap_session_update` stripping `model` from the session payload since the gateway takes it from `?model=`, `_is_fatal_error` adding the three `unsupported_*` codes). Model is still required provider-prefixed, and `self._opts.is_azure = False` / `api_version = None` correctly keeps ambient Azure env off the gateway URL and wire format.

Rest of the delta is mechanical: `resolve_credentials` extracted from `inference/llm.py` (same env precedence, same messages), `InferenceClass` and `RealtimeModels` relocated to `_utils`/`openai_types` and re-exported. All 25 checks green including ruff, type-check 3.10/3.13, test-realtime, unit-tests and livekit-plugins-openai.

Two things worth a reply, neither blocking:

1. **Metrics label asymmetry.** `_label` is `f"{type(self).__module__}.{type(self).__name__}"`. You pinned the direct class's pre-move label, but the hosted one moves from `livekit.plugins.openai.realtime.inference_realtime_model.InferenceRealtimeModel` to `livekit.agents.inference.realtime.openai.RealtimeModel`. Presumably intended given the class is being renamed and re-namespaced on purpose — but it will shift `RealtimeModelMetrics` labels for hosted sessions, so anyone grouping dashboards on that string sees a break. Worth a line in the PR body if deliberate.

2. **Devin's second report says "1 new potential issue … not posted on this PR by your GitHub settings."** It's only visible in the Devin UI, so it is outside what I can see from the PR. Please check it before merging.

No Critical or Important issues from my pass.

PR #7325, opened 2026-09-17
feat(openai): add GPTLiveModel.with_azure for Azure OpenAI

acharan-tech-200037:
GPT-Live could only reach api.openai.com. Azure serves it at /openai/v1/live/sessions, authenticates with an api-key header or an Entra bearer token, and resolves the backend Responses model as a deployment of the same resource, so the OpenAI defaults reach nothing there.

with_azure mirrors RealtimeModel.with_azure: it takes the deployment, endpoint and credentials, falls back to AZURE_OPENAI_API_KEY and AZURE_OPENAI_ENDPOINT, and rejects an incomplete configuration up front, including responses delegation with no backend deployment, which otherwise fails once per delegated turn while the voice model promises an answer.

The session URL now swaps only the scheme, instead of replacing the first "http" anywhere in the URL. OpenAI requests are unchanged.


longcw (approved):
looks good to me!

acharan-tech-200037:
> looks good to me!

Thank you @longcw for taking the time to review this! It was great working on this one, and I'm excited to see Azure support for GPT-Live land in LiveKit.

Issue #7302, opened 2026-09-16
Gemini 3.8 Live is async-by-default: plugin ends up in a hybrid blocking/non-blocking state, behavior/scheduling cannot be set per tool, and tool results wait for playout


IngLP:
### Summary

`gemini-3.8-live` (GA, 2026-09-15) makes asynchronous function calling the default: *"Async execution (`behavior: NON_BLOCKING`) is now the default function calling mode. You can still use synchronous blocking mode for backwards compatibility by setting `behavior: BLOCKING` on your tool declarations"* (model page, "Migrating from Gemini 3.1 Flash Live"). `gemini-3.8-live-extended-thinking` accepts **only** `NON_BLOCKING`. Three things in `livekit-agents` / `livekit-plugins-google` (main `fc22671`, identical to 1.8.2 for the files below) get in the way of using that properly. Measured on 1.8.1 with the production-size system instruction of a phone agent, 15 live trials per arm.

### 1. With the new default, the plugin ends up in a hybrid state it does not know about

`tool_behavior` is not set → `to_fnc_ctx` omits `behavior` from every `FunctionDeclaration` (`livekit-agents/livekit/agents/llm/_provider_format/google.py:152`, `:167`: `if tool_behavior is not None: schema["behavior"] = ...`). On 3.8 the server then treats every tool as `NON_BLOCKING`, but `supports_silent_scheduling` is derived from what the *client* set (`livekit-plugins-google/livekit/plugins/google/realtime/realtime_api.py:708-711`), so it stays `False`: a tool result with `reply_required=False` (typical for a fire-and-forget update, or for a tool that completed inside an interrupted turn, `voice/generation.py` marks those) is sent without `scheduling: SILENT`, the model answers anyway, and the warning at `realtime_api.py:718-723` ("a tool result wants no reply … declare the tools NON_BLOCKING on the Gemini API to keep it silent") fires on **every call: 15/15 and 15/15 in two arms with the default, 10/10 with `tool_behavior=BLOCKING` set explicitly, 0/15 with `tool_behavior=NON_BLOCKING` set explicitly**. Suggestion: on models where the documented default is `NON_BLOCKING`, either send `behavior` explicitly (whatever the user chose, defaulting to the model's default) or derive `supports_silent_scheduling` from the effective behavior, not from the option.

### 2. `behavior` and `scheduling` are per session, they need to be per tool

[...]

PR #7329, opened 2026-09-18
fix(google): default tool_behavior to NON_BLOCKING on Gemini 3.8 Live

longcw:
**Problem:** Gemini 3.8 Live declares tools `NON_BLOCKING` unless the client sends a `behavior` field, and the plugin sends none while `tool_behavior` is unset. The server then runs tools async while `supports_silent_scheduling` reads the unset option and stays `False`, so a result with `reply_required=False` goes out without `scheduling: SILENT` and the model answers it anyway.

**Fix:** `RealtimeModel.__init__` resolves `tool_behavior` to `NON_BLOCKING` for 3.8 models and stores it in `_opts`, so the tool declaration and the scheduling decision read one value. Models without the tag still send no `behavior`, and an explicit `tool_behavior` still wins.

Addresses the first item of #7302.

<details>

> **Is the Gemini 3.8 Live default measured?**
>
> No. The claim comes from the model page quoted in #7302. The warning counts reported there (15/15 with the default, 10/10 with `BLOCKING`, 0/15 with `NON_BLOCKING`) restate `supports_silent_scheduling`, which reads no server field, so they are not evidence about the server. One live call settles it: call a tool that sleeps for three seconds with `tool_behavior` unset. Speech during the sleep means the server ran the tool async. Silence means the server declared it `BLOCKING`, and this change must not land.
>
> **How to see the failure**
>
> A tool that returns `None` sets `reply_required=False` (`generation.py:1202`), and that is the only input which reaches the silent path. On `gemini-3.8-live` without this change, the warning at `realtime_api.py:718` fires for such a call and the model speaks after the result. A tool that returns a value keeps `reply_required=True` and never reaches the path, so it cannot show the failure.
>
> **Blast radius**
>
> `_opts.tool_behavior` has two readers: the tool declaration built in `create_tools_config` (`realtime_api.py:1254`) and `supports_silent_scheduling` (`realtime_api.py:710`). Resolving the default once in the constructor keeps them consistent, instead of adding a second derivation at each reader. `update_options` already restarts the session when the value changes (`realtime_api.py:602`), and Vertex keeps its own guard, because it accepts the declaration but drops `scheduling`.
>
> **What this does not fix**
>
> Every tool on 3.8 is now declared `NON_BLOCKING`, a hangup tool included, which the model can talk over. Per-tool `behavior` needs a field on `FunctionToolInfo` and is the second item of #7302. The wait that holds a tool result until playout ends lives in core (`agent_activity.py:4667`) and is #6741.

</details>


davidzhao (approved):
lgtm

PR #7333, opened 2026-09-18
feat(telemetry): record conversation content on realtime spans

### LINKED ISSUE #7304: Realtime models: OTel span misses input/output or usage on `realtime_inference` span, no user input on `agent_turn` span
### Bug Description

With a realtime model (tested with `GPTLiveModel`, same code path as `RealtimeModel`), the spans from livekit-agents 1.8.x are missing most of the conversation content. A GenAI backend shows every turn as an empty generation with no cost.

The cascaded pipeline path is fine. `llm_request` gets `gen_ai.input.messages`, `gen_ai.output.messages` and usage, and `agent_turn` gets `lk.pii.instructions` and `lk.pii.user_input`. The realtime path never calls the content helpers.

## What's missing

| Span | Has today | Missing |
|---|---|---|
| `realtime_inference` (`generate_content`) | provider, model, `gen_ai.output.type`, conversation id | `gen_ai.input.messages`, `gen_ai.output.messages`, `gen_ai.system_instructions`, usage |
| `agent_turn` (`invoke_agent`) | `lk.pii.response.text`, `lk.pii.response.function_calls` | `lk.pii.user_input`, `lk.pii.instructions` |

The user's transcript is on no span at all. Only `user_speaking` debug spans, without text.

Usage: GPT-Live's backend model (`gpt-5.6-luna`, ~1.4k prompt / 74 completion tokens per turn) is emitted as `LLMMetrics` on `metrics_collected` but never attached to a span. The voice model's `RealtimeModelMetrics` reports zero tokens (billed by duration), so `realtime_inference` has no usage even when `record_realtime_metrics` matches.

### Expected Behavior

Parity with the pipeline path: `realtime_inference` carries input/output messages, system instructions and usage; `agent_turn` carries `lk.pii.user_input` and `lk.pii.instructions`.

Happy to send a PR. Rough shape: reuse `to_input_messages` / `to_output_messages` from `telemetry/gen_ai.py` on the realtime chat context, and attach the backend `LLMMetrics` to the pending `realtime_inference` span by `response.id`.

### Reproduction Steps

```bash
1. `AgentSession(llm=GPTLiveModel(...))` with one function tool, `set_tracer_provider(provider)` with any OTLP exporter. `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` unset (defaults on), no redaction.
2. Short conversation with one tool call.
3. Look at the spans.

Real example, the Langfuse voice demo (https://langfuse.com/docs/demo) runs exactly this setup. Public trace: https://cloud.langfuse.com/project/clkpwwm0m000gmm094odg11gi/traces/738d3c599f6071781340ebe0ead4054c?observation=ca5d2ea847dfef01

`realtime_inference` there carries only `gen_ai.operation.name=generate_content`, `gen_ai.provider.name=openai`, `gen_ai.request.model=gpt-live-1`, `gen_ai.request.stream=true`, `gen_ai.output.type=speech`, `gen_ai.conversation.id`. `agent_turn` has `lk.pii.response.text` and nothing about the user.
```

### Operating System

MacOS

### Models Used

_No response_

### Package Versions

```bash
livekit-agents 1.8.1, livekit-plugins-openai 1.8.1, exporter is the Langfuse Python SDK. Same on `main` today.
```

### Session/Room/Call IDs

_No response_

### Proposed Solution

```python

```

### Additional Context

_No response_

Image

longcw:
**Problem:** A session driven by a realtime model records almost nothing of the conversation on its spans: `realtime_inference` carries no messages and no instructions, `agent_turn` carries no user input, and the caller's transcript reaches no span at all. A GenAI backend therefore shows every turn as an empty generation.

**Fix:** The realtime path now fills `gen_ai.system_instructions`, `gen_ai.input.messages`, `gen_ai.tool_definitions` and `gen_ai.output.messages` on `realtime_inference`, and `lk.pii.instructions` and `lk.pii.user_input` on `agent_turn`. The model's own final transcript opens a `user_turn` span with `lk.pii.user_transcript`, back-dated to the provider's `turn_started_at` or marked with `lk.user_turn.start_estimated` when the provider reports none.

close #7304.

<details>

> **How accurate is the `user_turn` start?**
>
> The start is an estimate. `InputTranscriptionCompleted.turn_started_at` is set by the openai, gpt-live and google plugins only, and each derives it differently: openai uses the time its `input_audio_buffer.speech_started` event arrived, google uses the generation's creation time. Every other plugin leaves it `None`, and the span then starts where the transcript arrived and carries `lk.user_turn.start_estimated`.
>
> A caller who needs the real speech boundaries can pass a `vad` to `AgentSession`. The VAD then opens the `user_turn` span at the measured speech onset and drives the `user_speaking` spans and the user state, and this change stamps the transcript onto that span instead of opening its own.
>
> **Why does the transcript open the span, and why in `AudioRecognition`?**
>
> `InputSpeechStartedEvent` is an empty dataclass, so the speech events cannot say which transcript belongs to which speech interval — pairing them would race with a provider that withholds the final transcript until its reply has finished generating. The transcript already carries `item_id` and `turn_started_at`, so it is the only signal with both the text and its start.
>
> `AudioRecognition` stays the one owner of `user_turn`. A user-supplied VAD already opens that span under a realtime model at `audio_recognition.py:1392`, with an empty transcript, because `_audio_transcript` is filled by STT alone. Minting a second span in `agent_activity.py` would leave that case broken.
>
> **Why do the instructions not come from the chat context?**
>
> `update_instructions(..., add_if_missing=True)` runs only in the `elif isinstance(self.llm, llm.LLM)` branch at `agent_activity.py:1278`. A realtime model sends its instructions through `_update_session(instructions=...)`, so `to_system_instructions(chat_ctx)` returned an empty list on this path. It now accepts `ChatContext | str`, and the realtime caller passes `_render_realtime_instructions(...)`, the same string sent to the provider. The pipeline caller at `voice/generation.py:401` is unchanged.
>
> **What this does not carry**
>
> Per-turn usage. `_realtime_spans` is keyed by `generation_ev.response_id`, a client-side burst id of the form `item_<uuid>` for a `DuplexModel`. `GPTLiveModel` reports `LLMMetrics.request_id` as the OpenAI Responses `response.id` and `RealtimeModelMetrics.request_id` as the session id, so `record_realtime_metrics` never matches on that model. Joining the backend tokens to a turn needs the plugin to put the burst or delegation id on the metrics event first.

</details>


livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -2002,6 +2002,28 @@ def _ensure_user_turn_span(self, start_time: float | None = None) -> trace.Span:
...
+        transcript: str,
+        confidence: float | None,
+        turn_started_at: float | None,
+    ) -> None:
+        """Put a realtime model's own transcript on a ``user_turn`` span.
+
+        No VAD or STT event opens a span for a server-detected turn, so one opened here is
+        back-dated to ``turn_started_at`` and closed at once; one a VAD opened is left to it."""
+        already_open = self._user_turn_span is not None and self._user_turn_span.is_recording()
+        span = self._ensure_user_turn_span(start_time=turn_started_at)
```
chenghao-mou, line 2017: it could be the case that the transcript is late and gets attached to a new user span, or create a duplicate user_turn span if the vad one is closed. should we add a field to annotate that this is from realtime model, so at least it is clear when it is misaligned?

livekit-agents/livekit/agents/voice/audio_recognition.py
```diff
@@ -2002,6 +2002,28 @@ def _ensure_user_turn_span(self, start_time: float | None = None) -> trace.Span:
...
+        span.set_attribute(trace_types.ATTR_USER_TRANSCRIPT, transcript)
```
chenghao-mou, line 2018: VAD also ends the span with _audio_transcript, which could be empty with realtime model.

PR #7351, opened 2026-09-19
fix(openai): don't re-dispatch a GPT Live call redelivered after its result

avionicharshit-byte:
#7230 skips a redelivered GPT Live backend call by checking the current response's call id set (`_backend_running_responses[d_id]`, or `_backend_open_calls` once the response has completed). but `_append_items` removes the call id from those sets as soon as its result is sent, so the check only holds while the call is still unanswered.

if a fast tool answers before the service redelivers the same completed `response.output_item.done`, the redelivery gets through and the tool runs a second time. for a tool that books or sends something, that means doing it twice. if the redelivery lands after `response.completed`, the call also goes back into `_backend_open_calls` and holds the continuation until a second output arrives for a call that was already answered.

the fix keeps a per connection `_backend_dispatched_calls` set, checks it before dispatching, and clears it in `_reset_for_reconnect`, since a new connection is a new session. dedup is still per connection like before, it doesnt try to cover reconnects.

added `test_a_call_redelivered_after_its_result_is_not_dispatched_again`, run for a redelivery before and after `response.completed`. both fail on main with the tool dispatched twice and pass with the fix. `tests/test_gpt_live_model.py` passes (86) and ruff is clean.


avionicharshit-byte:
@longcw follow-up to #7230, a redelivered call could still run the tool twice once its result was sent. regression test included, happy to rebase

longcw (commented):
thanks for the pr. could you share the debugging logs when the issue happens? one thing I worried about is when this would happen and when should we clear the set.

avionicharshit-byte:
no live logs, found it reading the code after #7230, so the redelivery-after-result case is inferred from #7230's duplicate delivery, not observed. the set is per connection and cleared on reconnect, one call id per dispatched call, so it grows with tool calls in a session. if you'd rather not take it without a real trace, happy to close it

longcw:
feel free to reopen if you see the same call id actually dispatched twice

PR #7364, opened 2026-09-20
fix(xai): detect a silently dropped STT socket instead of hanging

### LINKED ISSUE #7363: xai STT: a silently dropped WebSocket is never detected
### Bug Description

On a half-open connection (no FIN, no RST) the xAI STT `SpeechStream` sits on a dead socket forever: the session stays open, no transcripts arrive, and nothing is ever logged. Same class of bug as #7206 (Deepgram v1), #7356 (Soniox) and #7358 (Telnyx), still present in `livekit-plugins-xai`.

`_connect_ws` opens the socket without `heartbeat`, so `recv_task` parks on `ws.receive()` and the retry in `_main_task`, which only runs when something raises, never gets a turn. There is no keepalive task in this plugin, so nothing else touches the socket when audio stops. And when a heartbeat is on, the resulting `WSMsgType.ERROR` hits the "unexpected xAI message type" branch and is stepped over; the loop only ends because a `CLOSED` follows, raised as a generic "closed unexpectedly" with the actual reason from `ws.exception()` dropped.

### Expected Behavior

The dropped connection is detected within the heartbeat window and `_main_task` reconnects, with the reason attached to the error it logs.

### Reproduction Steps

Point `XAI_WEBSOCKET_URL` at a local aiohttp WebSocket server with `autoping=False` that sends `transcript.created` on connect, answers audio frames with `transcript.done` messages for 10s, and then stops reading, stops writing, and stops answering pings while keeping the TCP socket open. Run a real `xai.STT` stream pushing 20ms silence frames for 120s.

On `main`: last transcript at 9.2s, no warning or error logged for the remaining 110s, one socket for the whole run.

### Versions

livekit-agents 1.8.2, livekit-plugins-xai at `main` (e5024c76d), aiohttp 3.14.3, Python 3.13, macOS.

sidxh:
On a half-open connection (no FIN, no RST) the xAI STT `SpeechStream` sits on a dead socket forever: the session stays open, no transcripts arrive, and nothing is ever logged.

Two things cause it, same class as Deepgram v1 before #7206, Soniox in #7357 and Telnyx in #7359. `_connect_ws` opens the socket without `heartbeat`, so `recv_task` parks on `ws.receive()` and the retry in `_main_task`, which only runs when something raises, never gets a turn. This plugin has no keepalive task, so when audio stops nothing else touches the socket either. And a `WSMsgType.ERROR` (how a heartbeat timeout arrives) hit the "unexpected xAI message type" branch and was stepped over; the loop only ended because a `CLOSED` followed, raised as a generic "closed unexpectedly" with the reason on `ws.exception()` dropped.

This adds `heartbeat=30.0` (parity with Deepgram and Muse) and raises a retryable `APIConnectionError` from the ERROR frame with `ws.exception()` as the cause, so `_main_task` reconnects deterministically and the log says why.

Verified against a local fake xAI server that goes dead 10s into each connection while keeping the TCP socket open, with a real `SpeechStream` pushing silence for 120s. On `main` the last transcript is at 9.2s, nothing is logged for the remaining 110s, and one socket is opened. With this change the plugin logs `xAI connection lost (caused by ServerTimeoutError: No PONG received after 15.0 seconds)` at 55.6s, reconnects, and transcripts resume at 56.9s; three sockets over the run. Two regression tests added in a new `tests/test_plugin_xai_stt.py`; both fail on `main`.

#7279 handles xAI's application-level `{"type": "error"}` message in the same `recv_task`; that is a different failure (the server tells you it is done) and the two changes are independent.

30s heartbeat means ~46s to detect, which is long for a live call. Kept it for parity; happy to lower it.

Closes #7363.


tinalenguyen (approved):
thank you for the PR!

PR #7367, opened 2026-09-20
fix(llm): make ChatContext.truncate() honor a non-positive max_items

sxh313:
## What

`ChatContext.truncate(max_items=...)` used `self._items[-max_items:]`. In Python `-0` is `0`, and `items[0:]` is the whole list -- so truncating to a budget of **zero items returned every item**, and a **negative** budget silently trimmed from the *front* of the conversation instead of keeping nothing.

The guard now treats any non-positive budget as "no conversational items", which still leaves the preserved instruction message, exactly as the docstring describes.

```
before:  truncate(max_items=0)  -> ['system', 'user', 'assistant', 'user']   (nothing dropped)
after:   truncate(max_items=0)  -> ['system']
before:  truncate(max_items=-2) -> front-trimmed slice
after:   truncate(max_items=-2) -> ['developer']
```

## Why it matters

`truncate` is the documented way to bound a context (`llm/chat_context.py:571`). A budget normally comes from a token/message calculation, and the interesting case is the one where it reaches 0: the caller asks for no history and gets all of it, so the request goes out at full size instead of failing or shrinking -- the opposite of what the bound was for. `#4843` / `#4845` earlier fixed which roles count as an instruction and the duplicate insert, but the slice itself was untouched, so this path is still live.

Valid (positive) budgets are unaffected: `len(self._items) <= max_items` still short-circuits, and `items[-max_items:]` is unchanged for `max_items > 0`.

## Tests

Three regression cases in `tests/test_chat_ctx.py`, next to the existing truncate tests. Differential, with only `chat_context.py` reverted:

```
FAILED tests/test_chat_ctx.py::test_truncate_zero_max_items_keeps_only_instruction
FAILED tests/test_chat_ctx.py::test_truncate_zero_max_items_without_instruction
FAILED tests/test_chat_ctx.py::test_truncate_negative_max_items_keeps_only_instruction
3 failed, 6 passed, 63 deselected

with the fix: 71 passed, 1 skipped   (whole tests/test_chat_ctx.py)
```

`ruff check` and `ruff format --check` are clean on both files.

## AI disclosure

This fix was produced with an AI coding assistant (Claude), reviewed and verified by me locally as described above.


livekit-agents/livekit/agents/llm/chat_context.py
```diff
@@ -588,7 +591,9 @@ def truncate(self, *, max_items: int) -> ChatContext:
             None,
         )
 
-        new_items = self._items[-max_items:]
+        # `-0` is `0` and `items[0:]` is the whole list, so a zero budget would otherwise
+        # keep every item and a negative one would drop items from the front instead.
+        new_items = self._items[-max_items:] if max_items > 0 else []
```
longcw: I think it should raise if max_items < 0
```python
if max_items < 0:
    raise ValueError("max_items must be non-negative")
...
new_items = self._items[-max_items:] if max_items else []
```

sxh313: Adopted in dfc9be76, exactly as you sketched:

```python
if max_items < 0:
    raise ValueError("max_items must be non-negative")
...
new_items = self._items[-max_items:] if max_items else []
```

The guard sits ahead of the `len(self._items) <= max_items` early return so a negative budget raises regardless of how many items the context holds. `test_truncate_negative_max_items_keeps_only_instruction` is now `test_truncate_negative_max_items_raises`, which also asserts the context is left untouched.

Verified locally on the branch: `pytest tests/test_chat_ctx.py` -> 71 passed, 1 skipped; the new test fails with `DID NOT RAISE` against the previous head (b5acab88) with only `chat_context.py` swapped back, so it pins the behaviour rather than the code shape. `ruff check` and `ruff format --check` clean on both files.

Thanks for the direction - raising is a better contract than my guessing an empty budget.

PR #7369, opened 2026-09-20
fix(simplismart): detect a silently dropped STT socket instead of hanging

### LINKED ISSUE #7368: simplismart STT: a silently dropped WebSocket is never detected
### Bug Description

On a half-open connection (no FIN, no RST) the Simplismart STT `SpeechStream` sits on a dead socket forever: the session stays open, no transcripts arrive, and nothing is ever logged. Same class of bug as #7206 (Deepgram v1), #7356 (Soniox), #7358 (Telnyx) and #7363 (xAI), still present in `livekit-plugins-simplismart`.

`_connect_ws` opens the socket without `heartbeat`, so `recv_task` parks on `ws.receive()` and the reconnect in `_run`, which only fires when that task ends, never gets a turn. There is no keepalive task in this plugin, so nothing else touches the socket when audio stops. And when a heartbeat is on, the resulting `WSMsgType.ERROR` hits the "unexpected simplismart message type" branch and is stepped over; the loop only ends because a `CLOSED` follows, and the reason on `ws.exception()` is dropped.

### Expected Behavior

The dropped connection is detected within the heartbeat window, the reason is logged, and the stream reconnects through the plugin's existing `_reconnect_event` path.

### Reproduction Steps

Point the stream's `ws_url` at a local aiohttp WebSocket server with `autoping=False` that accepts the config message, answers audio frames with transcript payloads for 10s, and then stops reading, stops writing, and stops answering pings while keeping the TCP socket open. Run a real `simplismart.STT(streaming=True)` stream pushing 20ms silence frames for 120s.

On `main`: last transcript at 9.2s, no warning or error logged for the remaining 110s, one socket for the whole run.

### Versions

livekit-agents 1.8.2, livekit-plugins-simplismart at `main` (e5024c76d), aiohttp 3.14.3, Python 3.13, macOS.

sidxh:
On a half-open connection (no FIN, no RST) the Simplismart STT `SpeechStream` sits on a dead socket forever: the session stays open, no transcripts arrive, and nothing is ever logged.

Two things cause it, same class as Deepgram v1 before #7206, Soniox in #7357, Telnyx in #7359 and xAI in #7364. `_connect_ws` opens the socket without `heartbeat`, so `recv_task` parks on `ws.receive()` and the reconnect in `_run`, which only fires when that task ends, never gets a turn. This plugin has no keepalive task, so when audio stops nothing else touches the socket either. And a `WSMsgType.ERROR` (how a heartbeat timeout arrives) hit the "unexpected simplismart message type" branch and was stepped over; the loop only ended because a `CLOSED` followed, with the reason on `ws.exception()` dropped.

This adds `heartbeat=30.0` (parity with Deepgram and Muse) and, on the ERROR frame, logs the reason and requests a reconnect through the plugin's existing `_reconnect_event`, the same way it already handles an unexpected `CLOSED`, rather than raising.

Verified against a local fake Simplismart server that goes dead 10s into each connection while keeping the TCP socket open, with a real `SpeechStream` pushing silence for 120s. On `main` the last transcript is at 9.2s, nothing is logged for the remaining 110s, and one socket is opened. With this change the plugin logs `simplismart connection lost: ServerTimeoutError('No PONG received after 15.0 seconds')` at 55.3s, reconnects, and transcripts resume at 56.6s; three sockets over the run. Two regression tests added to `tests/test_plugin_simplismart_stt.py`; both fail on `main`.

This leaves the send path alone. #6474 covers send-side drops across several plugins including this one, and the two changes are independent.

30s heartbeat means ~46s to detect, which is long for a live call. Kept it for parity; happy to lower it.

Closes #7368.


tinalenguyen (approved):
thank you for the pr!

PR #7370, opened 2026-09-20
fix(google): don't resolve the STT project from ADC when credentials are supplied

ShreeBohara:
`_get_recognizer` falls back to `google.auth.default()` whenever the credentials in use don't expose a `project_id`. That's right when nobody passed credentials, but it also fires when you pass `credentials=` yourself, and the credential types that usually carry no project — Workload Identity Federation, impersonated credentials — hit it every time. On a host or pod with unrelated ADC configured, the plugin builds a v2 recognizer path for a project the caller never asked for.

On my machine:

```python
stt = STT(credentials=AnonymousCredentials(), model="chirp_3")
client = await stt._create_client(timeout=1.0)
stt._get_recognizer(client)
# 'projects/<my-unrelated-gcloud-project>/locations/global/recognizers/_'
```

The comment in `_create_client` already says the project is resolved there "so `_get_recognizer` doesn't fall back to Application Default Credentials" — but the `project_id`/`quota_project_id` lookup finds nothing on those credential types, so it does fall back.

This only consults ADC when no credentials were supplied, and otherwise raises the clear error that's already there. It also treats a `None` project from ADC as unresolved: `google.auth.default()` returns `(creds, None)` when no project is configured, which used to produce a literal `projects/None/...` path.

One behaviour change: passing `credentials=` without a project and relying on ADC to fill it in now errors instead. The message points at `project=`.

Verified with `make check` and the full `--unit` selection (3096 passed; the 9 `test_room.py` errors need a local livekit-server and are unrelated). The new test fails on `main` regardless of how the host is configured.


ShreeBohara:
recheck

PR #7372, opened 2026-09-21
fix(soniox): hold the endpoint until the translation arrives

### LINKED ISSUE #6887: soniox: an <end>/<fin> between an original token and its translation drops source_texts from the final transcript

longcw:
**Problem:** In translation mode an `<end>` token can arrive while the target side is still empty, because Soniox finalizes the source tokens before it produces their translation. `send_endpoint_transcript()` then emits no `FINAL_TRANSCRIPT` and no `END_OF_SPEECH`, so the turn stays open until the next utterance ends and two utterances merge into one final with a mispaired source.

**Fix:** The endpoint is held together with its source side, and released as soon as a translation lands. A fresh source buffer takes over while the endpoint is held, so tokens for the next utterance cannot join the held one.

Fixes #6887. Replaces #6950.

<details>

> **How to see it**
>
> `test_endpoint_before_translation_pairs_each_utterance` sends two utterances, each as a source message ending in `<end>` followed by a translation message. On `main` it produces one final for both utterances, with `source_texts=['Good bye.']` against `target_texts=['Hola mundo.']`. The test asserts two finals, each paired with its own source.
>
> **Why #6950 is not enough**
>
> That PR stops the `else` branch from resetting `final_original`, so the source text survives the marker. The endpoint is still discarded, so the same input still yields one final for two utterances, now with `source_texts=['Hello world.Good bye.']` — the source of both turns against the translation of one. Its test passes because frame two appends a second `<end>`, and Soniox sends one `<end>` per endpoint.
>
> **What closes a held endpoint**
>
> Three paths, so the hold cannot outlive the stream: a translation landing in any later message (`stt.py:566`), a new endpoint (`stt.py:503`), and the `finished` or error flush that already calls `send_endpoint_transcript()`. An utterance that never gets a translation is emitted source-only, with its own tokens supplying the text and the timings, rather than dropped as it is today.
>
> **Blast radius**
>
> The three closures live inside `_recv_messages_task` and are reachable from nowhere else. `held_original` is only ever set when `is_translation_mode` is true, so the non-translation path keeps its current behavior; the 28 existing tests in `tests/test_plugin_soniox_stt.py` cover it and are unchanged.
>
> **What is not confirmed**
>
> Soniox documents that translation tokens follow the transcription tokens chunk by chunk, but not whether the complete final translation for a segment always arrives in one message. If a translation splits across messages, the held endpoint closes on the first chunk and the tail joins the next utterance — smaller than the current failure, but still wrong. A live capture of the token order around `<end>` would settle both this and the reachability of the original report.

</details>


longcw:
Closing this.

The source/translation pairing is positional only. The realtime token carries no id and no segment id, and translated tokens carry no `start_ms` or `end_ms`, so the only boundary signal in the stream is `<end>` itself.

That means if `<end>` can arrive before an utterance's translation, the plugin has no way to know when that translation is complete. Every candidate rule is a guess about provider timing with its own failure mode: flush on the first final translation chunk (this PR), gate on pending non-final tokens, or wait for the next endpoint. This PR's guess splits a translation that arrives in two chunks — `Hola.` becomes a final `Ho` and a final `la.` paired with the next utterance's source, where `main` keeps the text whole.

The premise is also still unconfirmed. Neither #6887 nor this branch has a live capture showing `<end>` landing before the translation of the utterance it closes.

If the gap is real, the fix belongs on the Soniox side — emit `<end>` after the utterance's translation tokens, or add a translation-complete marker. Refs #6887, #6950.

PR #7374, opened 2026-09-21
fix: preserve fallback provider attribution on llm node spans

chenghao-mou:
After failover, `llm_node` could report the fallback model with the primary provider. Preserve the serving-provider update from [#7137](https://github.com/livekit/agents/pull/7137) by recording the node's request identity before nested inference starts.

Addresses AGT-3551: [issue](https://linear.app/livekit/issue/AGT-3551/llm-node-overwrites-the-serving-provider-after-fallback).

The full local type check reports existing plugin dependency errors also present on `main`; the core package passes.

<details>

**Model:** GPT-6

> can you create draft PR for the fallback issue?

The requested issue is the `llm_node` provider overwrite found during the preceding telemetry review. The fallback stream records the serving provider, then node completion overwrites it with the original provider. This PR preserves the per-request attribution.

</details>


longcw (commented):
from claude code:

**What it misses.** On the request where the failover happens, `FallbackAdapter.model` still returns the primary: `_next_instance()` marks the primary unavailable only after it fails. `FallbackLLMStream._run` then writes `gen_ai.provider.name` and `gen_ai.response.model` to `_caller_span`, but not `gen_ai.request.model`. The node span ends with the primary's request model beside the fallback's provider — the same mismatched pair, reversed. The

chenghao-mou:
> from claude code:
> 
> **What it misses.** On the request where the failover happens, `FallbackAdapter.model` still returns the primary: `_next_instance()` marks the primary unavailable only after it fails. `FallbackLLMStream._run` then writes `gen_ai.provider.name` and `gen_ai.response.model` to `_caller_span`, but not `gen_ai.request.model`. The node span ends with the primary's request model beside the fallback's provider — the same mismatched pair, reversed. The

Yeah, this is expected from this change:
started:
- gen_ai.request.model: model A
- gen_ai.provider.name: provider A

after fallback:

previously:
- gen_ai.request.model: model A
- gen_ai.response.model: model B
- gen_ai.provider.name: provider A

now:
- gen_ai.request.model: model A
- gen_ai.response.model: model B
- gen_ai.provider.name: provider B

so the fallback (requested A but B responded) is clear.

longcw (approved):

PR #7376, opened 2026-09-21
fix: record llm node exceptions consistently

chenghao-mou:
Exceptions raised when calling or awaiting an `llm_node` miss `error.type`. Project redaction also leaves node failures without an exception event or error status. Record failures through the existing redaction-aware helper across invocation, streaming, and cleanup, with one event per node span. Preserve cancellation and inference classification.

Addresses AGT-3553: [Record llm_node exceptions consistently across node types](https://linear.app/livekit/issue/AGT-3553/record-llm-node-exceptions-consistently-across-node-types).

<details>

**Model:** GPT-6

> okay, so we can have another PR to record the exception.

The discussion identified inconsistent exception recording between coroutine and streaming LLM nodes. This PR records node failures without adding inference metadata to early failures.

</details>


livekit-agents/livekit/agents/voice/generation.py
```diff
@@ -183,6 +189,24 @@ async def _llm_inference_task(
     data: _LLMGenerationData,
     model: str | None = None,
     provider: str | None = None,
+) -> bool:
+    try:
+        return await _llm_inference_impl(
+            node, chat_ctx, tool_ctx, model_settings, data, model, provider
+        )
+    except Exception as exc:
+        telemetry_utils.record_exception(trace.get_current_span(), exc)
```
longcw: Could this move into `_DynamicTracer.start_as_current_span` instead? Pass `record_exception=False, set_status_on_exception=False` unconditionally, then wrap the `yield` in `except Exception as e: telemetry_utils.record_exception(span, e); raise`. `detached_span` in the same class already does exactly this.

longcw: maybe also `use_span`?

chenghao-mou: good point, will do that

longcw (approved):

livekit-agents/livekit/agents/telemetry/traces.py
```diff
@@ -181,14 +186,18 @@ def start_span(self, *args: Any, **kwargs: Any) -> Span:
...
+        record_exception = bound.arguments.get("record_exception", True)
+        set_status_on_exception = bound.arguments.get("set_status_on_exception", True)
+        bound.arguments.update(record_exception=False, set_status_on_exception=False)
+        with trace_api.use_span(*bound.args, **bound.kwargs) as span:
+            try:
+                yield span
+            except Exception as exc:
+                telemetry_utils.record_exception(
+                    span, exc, record_event=record_exception, set_status=set_status_on_exception
+                )
```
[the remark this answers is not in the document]
chenghao-mou: False positive, internally otel catches only Exception as well.

PR #7381, opened 2026-09-21
feat(mistralai): add chat completions API mode

leamoukarzel:
## Why

The Mistral Chat Completions API (`/v1/chat/completions`) delivers significantly lower p50 TTFT than the Conversations API (`/v1/conversations`): **-21% on mistral-small** (334 ms vs 421 ms) and **-36% on mistral-medium** (450 ms vs 707 ms).

However, Chat Completions only supports function tools — provider tools (web search, document library, code interpreter, connectors) require the Conversations API. This PR adds an `api_mode` switch so callers can opt into the faster path when they don't need provider tools, while keeping Conversations as the default.

## Summary

Adds an `api_mode` parameter to the MistralAI LLM plugin, allowing callers to choose between:

- **`ApiMode.CONVERSATIONS`** (default) — uses the Conversations API (`/v1/conversations`), which supports both function tools and provider tools (web search, document library, code interpreter, connectors).
- **`ApiMode.CHAT_COMPLETIONS`** — uses the Chat Completions API (`/v1/chat/completions`), which only supports function tools but has lower latency.

The default remains `CONVERSATIONS`, so this is fully backward-compatible.

### Changes

- New `ApiMode` enum exported from `livekit.plugins.mistralai`
- `LLM.__init__` and `update_options` accept an `api_mode` parameter
- `LLMStream._run` dispatches to `_run_conversations` (existing logic, unchanged) or `_run_chat_completions` (new)
- Chat completions mode validates that no provider tools are passed (raises `ValueError`)
- Streaming delta parsing and tool call accumulation for the chat completions response format
- Comprehensive test coverage for the new mode

## Test plan

- [x] Existing conversations-mode tests still pass
- [x] New `TestChatCompletionsMode` tests cover text streaming, tool call accumulation, usage reporting, error mapping, and provider tool rejection
- [ ] Manual integration test against Mistral API


livekit-plugins/livekit-plugins-mistralai/livekit/plugins/mistralai/llm.py
```diff
@@ -43,9 +46,26 @@
...
+    """
+
+    CONVERSATIONS = "conversations"
+    CHAT_COMPLETIONS = "chat_completions"
+
+
 @dataclass
 class _LLMOptions:
     model: ChatModels | str
+    api_mode: ApiMode
```
tinalenguyen, line 68: should we accept `str` as well so users wouldn't need to import `ApiMode`?

tinalenguyen (approved):

Issue #6399, opened 2026-07-13
Support sequential or dependency-aware execution of parallel tool calls


PrinceGupta1999:
### Feature Type

Would make my life easier

### Feature Description

When an LLM returns multiple tool calls, LiveKit currently starts each call as a separate asyncio task and waits for them together:

https://github.com/livekit/agents/blob/03ab627260539504ea75f37f9636063e0223e415/livekit-agents/livekit/agents/voice/generation.py#L842-L865

This works for independent tools, but in many real life workloads tools depend on others, due to shared dependencies.

For example, in a hotel booking flow, the LLM may return `save_room_preference` and `save_meal_preference` together. However, the runtime must execute `save_room_preference` first because the available meal options may depend on the selected room package.

It would be helpful to support an optional execution policy such as:
- `parallel`: current behavior.
-  `Custom dependency ordering`: run independent calls in parallel while waiting for prerequisites.

The model should still be able to return multiple calls in one response. The ordering should happen locally during tool execution. Existing parallel behavior should remain the default.

### Workarounds / Alternatives
- Setting parallel_tool_calls=False forces separate model turns between tools
- Applications can also build a wrapper tool or coordinate with locks, but this requires custom orchestration and does not have a clean view of all calls in the current batch.
- TaskGroup supports ordered conversational tasks, but not ordering ordinary tool calls returned together.

### Additional Context

A related request exists for conditional TaskGroup workflows: https://github.com/livekit/agents/issues/5697 

This request is specifically about execution ordering for function tool calls within the same model turn.

mastermanas805:
I'd like to take this on, but the streaming shape of tool dispatch changes the design enough that it's worth agreeing on scope first.

[...]

PR #7385, opened 2026-09-21
feat: add dependency-aware tool execution

vznh:
## Problem and behavior

Related to #6399.

When a model returns `save_meal` before `save_room` in one function stream, both currently start immediately. Awaiting the existing executor is insufficient: its first `ctx.update()` returns control before the tool body finishes.

This draft adds optional local ordering:

```python
@function_tool(after=("save_room",))
async def save_meal(ctx: RunContext, meal: str) -> str:
    ...
```

Independent calls start eagerly. Dependents wait until stream EOF fixes batch membership and every matching prerequisite terminates. Tools without dependency metadata retain parallel dispatch.

## Scope and contract

- `after` orders execution; it does not require prerequisite success. A failed or individually cancelled prerequisite still finishes its ordering obligation. There is no configurable failure policy or Agent/AgentSession setting.
- Names refer to calls in one function-stream batch. All matching calls count; an absent prerequisite creates no edge. Provider stream boundaries are not universally whole model turns.
- Cycles in the available tool snapshot are rejected before dispatch. Session close, stream cancellation, and permanent handoff abandon queued work. Already admitted session-scoped tools retain their existing lifetime; inline AgentTask pauses are not abandonment.
- Progress is not termination. Internal pending state and deferred results preserve distinct history IDs and delivery ordering. The public chat/history schema is unchanged.
- No argument regeneration, result injection, entity matching, cross-turn graph, retry policy, or rollback is added. Tools needing earlier results as arguments still require another model turn or a composite tool.

Production changes are confined to tool metadata, the executor terminal signal, response-local scheduling, and activity lifecycle/delivery integration: four production files. Please review the batch-boundary, cycle, and pending-output contract while this remains draft.

## Validation

Base: `c0a8f4467f3cc48298a8fb0c05aca7744be56579`.

- Focused dependency/tool/realtime suite: **172 passed**. This includes **36 dependency cases**, with policy-only cases removed and three provider-commit failure regressions added.
- Ten fresh-process runs with asyncio debugging and varied hash seeds: **360 successful test executions**.
- Randomized local probe: **100 acyclic graphs / 825 calls**, mixing shuffled arrival, progress updates, and errors; no premature admissions, duplicate history item IDs, or orphan tasks.
- Real local LiveKit server/RTC probe: room text input, remote synthetic audio reception, progress before completion, dependent ordering, final context delivery, and clean teardown passed.
- Existing RTC resume and full-reconnect tests against the local server: **2 passed**, including audio continuity/recovery.
- Actual OpenAI plugin and SDK against a local HTTP/SSE endpoint: fragmented streaming data/arguments, reverse-arriving calls, progress continuation, and final tool-output delivery passed. This is adapter coverage, not a live OpenAI service test.
- `make fix` and `make check`: formatting, lint, and typing passed (**682 source files**).
- Full unit/audio suite: **3,632 passed, 5 skipped, 2 failed**. The failures are unchanged macOS sandbox process-memory tests (`task_for_pid` denied / USS unavailable), previously reproduced on baseline.

Fresh GitHub CI passed on `4c3185073`: unit tests, Ruff, Python 3.10/3.13 typing, and blockguard on Ubuntu/macOS/Windows. Provider/evaluation jobs were skipped.

Fault injection covers initial, late, and persistent realtime provider commit failures. A reproduced unhandled exception in the newly introduced deferred-reply task is now explicitly retrieved and logged. The corrected harness resolves follow-up model requests through natural close; these cases settle without orphan tasks. Persistent provider refusal does not imply successful delivery and no retry/recovery policy is introduced.

Live provider services, cloud dispatch/TURN, physical microphone/speaker behavior, and production deployment remain unvalidated: this sandbox has no provider credentials or audio devices. This remains a draft for review, not a claim of complete production certification.


vznh:
Devin approved ^_^

vznh:
@tinalenguyen would it be possible to check this over? thank you :-)

PR #7386, opened 2026-09-21
feat: add TypeSafe reply review plugin

AhmadIbrahiim:
## Summary

Adds `livekit-plugins-typesafe`, which reviews an agent's own replies and nudges it back on course when one goes off track.

After the agent speaks, the plugin sends that reply, the agent's `instructions`, its tool catalog and the recent transcript to [TypeSafe](https://typesafe.ai)'s System One model (Jev). Jev does not generate text. It answers typed yes/no, multiple-choice and rating questions with calibrated probabilities, and it answers every question in a request in parallel against one ingestion of the state. Five checks are therefore a single round trip, which is what makes it usable on a live call. Thresholds are applied in plugin code rather than by the model, and when one is crossed the plugin appends a correction to the chat context so the next reply corrects itself.

Nothing is authored per agent. The checks read `instructions` and the tool catalog on every review, so an agent handoff is picked up automatically.

```python
from livekit.plugins import typesafe

reviewer = typesafe.Reviewer()
reviewer.attach(session)
```

Two placements. Observe is the default and runs off the reply path once the assistant's message is committed, so it adds nothing to time-to-first-token. Gate is opt-in per check, holds a draft until it clears and redrafts once before any audio, at the cost of the full generation plus one evaluation. Gate needs an `llm_node`, so realtime and speech-to-speech models can only be observed.

A review that cannot run is logged and the turn goes unjudged rather than silencing the agent. That result carries `evaluated=False`, and the public flag is `needs_correction` rather than `ok`, so an outage cannot be read as a clean call. The README states plainly that this is a review layer and not an enforcement boundary.

No new dependencies. The API is a single POST, called over the agent's shared aiohttp session using the existing `APIStatusError` and `APIConnectionError` types, rather than pulling in the vendor SDK and a second HTTP stack.

## Installation

```bash
pip install livekit-plugins-typesafe
```

Set `TYPESAFE_API_KEY`. An example agent is at `examples/voice_agents/reviewer.py`.

## Validation

- Plugin tests: 33 passed. 27 unit tests plus 6 that drive a real `AgentSession` through the fake STT/LLM/TTS harness, covering the event subscription, tool bookkeeping, nudge delivery, `detach()` and behaviour during an outage.
- `uv run pytest --unit`: 3546 passed. The one failure, `test_google_credentials.py::test_clear_error_when_project_unresolvable`, is unrelated and environment-dependent; it asserts no application default credentials are present and fails on any machine where `GOOGLE_APPLICATION_CREDENTIALS` is set.
- `ruff format --check`: 1071 files clean.
- `ruff check`: clean.
- `mypy` via `scripts/check_types.py`: no issues in 688 source files, with `livekit.plugins.typesafe` included in the checked packages.

Also exercised against the live TypeSafe endpoint (Jev 1.13) end to end: a real `AgentSession` producing a real reply, reviewed by the real API. Round trips came back in 378 to 884ms for a five-check review at roughly 960 input tokens, which is within the latency budget the design assumes. Unit tests still run against a fake client so CI needs no key; happy to add a `pytest.mark.plugin("typesafe")` integration test if you want one in the suite.

That live run is also what set the thresholds. The first pass shipped guesses, and two of them fired on compliant replies. Measured on a labelled set, a compliant reply scores 0.46 to 0.93 on `follows_instructions` while a violating one scores 0.01 to 0.03, so the original 0.5 sat inside the compliant band. Jev is close to certain about a violation and much less certain that nothing is wrong, since spotting one broken rule is easier than confirming every rule held, so the thresholds belong near the violating end rather than halfway. `follows_instructions` moved 0.5 to 0.25, `advances_task` 0.4 to 0.25, and `severity` 2.0 to 2.2. `unsupported_claim` and the `expected_tool` confidence floor measured correctly and are unchanged. The sample is nine replies under one prompt, which the README says plainly.

## Notes for reviewers

- The default thresholds are starting points rather than calibrated values. TypeSafe's own guidance is that thresholds must be evaluated against your data, and the README says so and shows how to tune them against recorded calls.
- `expected_tool` assumes one tool per turn. An agent that legitimately calls two will show low confidence, and the check only fires above a confidence floor, so it stays quiet instead of firing wrongly.
- Tests live in the repo's `tests/` directory rather than inside the plugin, because they use the shared `fake_session` harness and the `--unit` category marker.


AhmadIbrahiim:
## Second round addressed — 5 of 6 fixed, 1 does not reproduce

Pushed in 5721074. Five findings fixed, one respectfully disputed with evidence below.

### Fixed

- **Default thresholds outlive their model** — agreed, and this was self-inflicted: calibrating the thresholds is what created the coupling, and the README already said "a threshold is only valid against the version it was tuned on" while the code followed the alias anyway. `DEFAULT_MODEL` is now pinned to `jev-1.13.0`, `CALIBRATED_FOR` names it, and the reviewer warns once if another version answers while the shipped thresholds are in use. Verified a pinned version id is accepted by the API even though `GET /v1/models` lists only aliases.
- **Missing answers look fully evaluated** — the sharpest of the set: it defeated the exact flag added last round to fix a different finding. `evaluated` now requires every active check to return a readable answer, so `{"answers": {}}` on a 200 is unjudged rather than clean.
- **Skipped reviews disappear from results** — applied, with one change to your patch: the verdict is recorded with `evaluated=False`, since nothing was evaluated.
- **Handoffs misattribute in-flight reviews** — the producing agent is captured when the message commits, and a correction is dropped rather than written into a different agent. **Partially fixed, and I want to be explicit about the limit.** Your first case, the review resuming after `update_agent`, is fixed and has a test that fails without it. Your second, a handoff landing before the event handler runs, is not: I traced it, and `conversation_item_added` genuinely does not name the author, so the successor still wins that race. Closing it needs agent ownership on the event. Recorded in the code and the README rather than papered over.
- **Identical gated replies get re-evaluated** — took the counted-queue option. Gate verdicts are queued per draft text and drained on commit, bounded at 8 with oldest-first eviction so an abandoned stream cannot leave a marker for a later identical reply.

### Does not reproduce

- **Gate feedback precedes reviewed reply** — I could not reproduce this and believe the analysis assumes append semantics. `ChatContext.insert` (`chat_context.py:473`) inserts *by creation time*, not at the end. The assistant message carries `created_at=reply_started_at`, which precedes a nudge stamped later, so it sorts ahead regardless of when the nudge is applied. I verified by printing the committed order with the nudge applied inside `gate` and again with it deferred: both give `system, user, assistant, feedback`.

  I moved the nudge to commit time anyway, because it does not depend on that ordering invariant and the queue was needed for the counted-drafts fix. The ordering test is kept but labelled an invariant guard, not a regression test, because it passes either way.

### Verification

Each fix has a test, and I checked the load-bearing ones actually fail without their fix rather than trusting a green run. Reverting the agent capture to a late `current_agent` resolve fails the handoff test; forcing `evaluated = True` fails both unjudged tests. That check is also what caught the ordering claim above.

- Plugin tests: 53 passed (10 new)
- `uv run pytest --unit`: 3566 passed. The single failure, `test_google_credentials.py::test_clear_error_when_project_unresolvable`, is pre-existing and environment-dependent.
- `ruff format --check` 1071 files, `ruff check`, and strict `mypy` over 688 files: all clean
- Re-ran end to end against live Jev 1.13: violation nudged, compliant reply clean

PR #7388, opened 2026-09-21
baseten: inline mid-conversation instructions for google/gemma-4-31B-it and Qwen/Qwen3.8-27B

henrytang-baseten:
## Summary

The Baseten plugin inherits the OpenAI serializer, which sends system messages wherever they appear in the chat context. LiveKit appends `generate_reply(instructions=...)` and the expressive TTS guide as a trailing system message, which is fine for models whose chat template renders a system turn anywhere but not for models whose template only accepts a leading one.

This PR makes the Baseten `LLM` inline mid-conversation system messages for two specific Model API ids:

- `google/gemma-4-31B-it`
- `Qwen/Qwen3.8-27B`

For those two models, `chat()` runs the existing `convert_mid_conversation_instructions` helper on a copy of the chat context before delegating to the OpenAI plugin, so later system messages are sent as `<instructions>`-wrapped user messages, the same treatment the Gemini, Anthropic, Bedrock and Mistral serializers already apply. The leading preamble stays a system message, tool calls and tool outputs are preserved, and the caller's `ChatContext` is not mutated.

Every other model id, including other Gemma and Qwen sizes and generations and dedicated deployments, is unaffected and keeps receiving the request exactly as before. The list is an explicit allow-list rather than a family match because template behaviour varies by generation within both families. A new `inline_mid_conversation_instructions` constructor argument overrides the default in either direction.

## Commits

1. **Inline mid-conversation instructions in the Baseten plugin.** Adds the `chat()` override, the constructor argument, and unit tests. Detection was initially a `gemma` / `qwen` substring match.
2. **Scope to an explicit allow-list.** Replaces the substring match with the two ids above and updates the tests to assert that other Gemma and Qwen ids pass through untouched.

## Changes

- `livekit-plugins-baseten/.../llm.py`: `chat()` override, model-id allow-list, `inline_mid_conversation_instructions` argument. Imports `convert_mid_conversation_instructions` from `livekit.agents.llm._provider_format.utils`; happy to move this to a public location or add an OpenAI-format variant in core if maintainers prefer.
- `tests/test_plugin_baseten_llm.py`: unit tests that capture the request body through an `httpx.MockTransport`. They cover allow-list detection, the rewrite for the two listed models, tool history preservation, no mutation of the caller's context, pass-through for all other ids (including other Gemma and Qwen ids), a byte-identical request when there is nothing to rewrite, and both directions of the override.

## Testing

- `uv run pytest tests/test_plugin_baseten_llm.py --unit` and `tests/test_chat_ctx.py` pass.
- `ruff check`, `ruff format --check` and `mypy` are clean on both files.
- Not yet verified against a live Baseten deployment of either listed model.

## Notes for review

- Adding a model to the list is a one-line change once its chat template behaviour is known. Gemma 2 and 3, for example, also reject a later system turn, but Gemma 3 additionally rejects consecutive user turns, so it would need the instruction folded into the preceding user message rather than appended.


livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/llm.py
```diff
@@ -1,21 +1,45 @@
...
+# user/assistant alternation) and Qwen3.5 raises "System message must be at the
+# beginning". Gemma 4 and Qwen3 render a later system turn without error, but the
+# rewrite is applied to the whole family so a per-turn instruction is delivered the
+# same way regardless of which generation a deployment serves. For these, the
+# per-turn instructions LiveKit appends as trailing system messages
+# (``generate_reply(instructions=...)``, expressive TTS guides) are inlined as
+# ``<instructions>``-wrapped user messages instead. Matched case-insensitively against
+# the model id so both Model API ids ("google/gemma-4-31B-it") and dedicated-deployment
+# names work; pass ``inline_mid_conversation_instructions`` to override.
+_INLINE_INSTRUCTIONS_MODEL_FAMILIES = ("gemma", "qwen")
```
peirangong-baseten: Would you please verify this change against both Gemma 2 and 3? My agent called out that with this change it is going to break the both cases. Details below:

For Gemma 3, there's a strict format that alternates between caller and agent. Without your change today it silently accepts. But with the fix now it will appear with two caller messages in a row which breaks in Gemma 3.

For Gemma 2, it refuses any note to the model. Your description implies the fix is applied to Gemma 2. If it is not, it would be good to clarify.

henrytang-baseten: Thanks. As discussed with you and @iancarrasco-b10, I updated the PR to only apply to 

google/gemma-4-31B-it
Qwen/Qwen3.8-27B

livekit-plugins/livekit-plugins-baseten/livekit/plugins/baseten/llm.py
```diff
@@ -1,21 +1,49 @@
...
+_INLINE_INSTRUCTIONS_MODELS = frozenset(
+    {
+        "google/gemma-4-31B-it",
+        "Qwen/Qwen3.8-27B",
+    }
+)
+_INLINE_INSTRUCTIONS_MODELS_LOWER = frozenset(m.lower() for m in _INLINE_INSTRUCTIONS_MODELS)
+
+
+def _needs_inline_instructions(model: str) -> bool:
```
iancarrasco-b10: nit: for better semantics lets name this _supports_inline_instructions

henrytang-baseten: Done

tinalenguyen (commented):
should we also add `"google/gemma-4-31B-it"` and `"Qwen/Qwen3.8-27B"` to `models.py` now?

tinalenguyen (approved):

PR #7391, opened 2026-09-22
fix(core): preserve PCM samples across mid-stream flushes

naszzz:
Mid-stream PCM flushes could discard an odd number of buffered bytes, shifting subsequent samples and causing loud static.

`AudioByteStream.flush()` now emits complete samples and retains trailing bytes. `AudioEmitter` uses the new `reset_progressive()` method to restart frame sizing without losing audio. `clear()` still discards audio for interruptions.


livekit-agents/livekit/agents/utils/audio.py
```diff
@@ -177,48 +176,58 @@ def buffered_duration(self) -> float:
...
-        Use this method when you have no more data to push and want to ensure
-        that all buffered audio data has been processed.
+        Returns:
+            A frame containing the complete samples, or an empty list if none
+            are available.
         """
         if len(self._buf) == 0:
             return []
 
-        if len(self._buf) % self._bytes_per_sample != 0:
```
longcw, line 198: maybe still log a warning when this happens, otherwise a stream that never aligns now loses its tail silently at a terminal flush.

longcw (approved):
lgtm! something nit:

